PR #19, opened 2015-01-25
Fixes for makefiles, GLFW 3.1 library for Linux

procedural:
GLFW needs additional libraries to link on Linux, also version 3.1 came out recently and Debian-like distros won't have it so it's reasonable to keep it locally like for Windows and OSX.

Old `make clean` command deletes every source file in folder since on Linux every file is executable :) New command deletes only compiled files.


raysan5:
Thank you very much for your contribution! :D

PR #61, opened 2015-10-26
Added Alpha Numeric Mappings

kd7tck:
GLFW key mappings where added for ease of reference.


kd7tck:
Sorry will move to develop branch soon.

PR #67, opened 2015-12-30
Fixed Android compilation with 1.3 updates

victorfisac:
- Fixed Android MakeFile adding gestures module.
- Fixed Android project template functions names.
- Updated compiled library and header.


raysan5:
Hi @victorfisac !

Thanks for the review! Just one detail, InitGesturesSystem() is already called internally by InitWindow(), is not required to be called again by the user. I will correct it in an update.

The first idea was decoupling that functionality (gestures system) from main raylib source but I realized that gestures are a must in touchable platforms (Android, HTML5 in tablets); so, by default, I initialize gestures on Android and HTML5.

Actually, I'm still thinking a better way to handle this; ¿should I understand simple touch as a gesture?¿or would it be better to link it directly with mouse button click?

PR #72, opened 2016-01-09
Fixed small glitch on zoom-in

victorfisac:
It needs to be confirmed if commented lines doesn't affect camera behaviour... I think that it doesn't...


raysan5:
Thanks @victorfisac, I'll plan a deep camera testing for next week.

PR #73, opened 2016-01-11
World to screen conversion and little review

victorfisac:
Example file can be deleted if you don't want it to be an official example... It is just the file I used for test purposes.


raysan5:
Great addition @victorfisac!

PR #75, opened 2016-01-19
raymath standalone

kd7tck:
Sorry this diff is so big. It would not let me merge so I simply copied your current develop branch and replaced mine with yours. Then I did a commit with it, it made hundreds of changes to line backup with yours. However the diff demands that all prior changes be included, even ones that are not needed. Not sure why? Otherwise if you don't want to deal with this merge, I could just email you the updated file and you can add it yourself.


kd7tck:
I plan to reset my branch, hold on.

PR #77, opened 2016-01-20
Added ray-sphere collision detection

victorfisac:
Review CalculateBoundingBox() changes. Maybe, these are not the correct values. I just changed to a Vector3 data type to compile successful.


raysan5:
Please @victorfisac, I just added some changes to develop branch and it conflicts with yours...

Could you sync to latest version and add changes again? (remember to make a copy before)

PR #82, opened 2016-01-29
Disable cursor feature

procedural:
Exposes `GLFW_CURSOR_DISABLED` flag that disables cursor so the mouse can't leave the window, essentially a better hide function for 3D cameras than `GLFW_CURSOR_HIDDEN` flag.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1097,6 +1097,24 @@ void ShowCursor()
     cursorHidden = false;
 }
 
+// Disable mouse cursor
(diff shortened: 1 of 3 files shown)


procedural:
:) Feature in action: https://github.com/procedural/quatFirstPersonCamera#example-for-raylib

raysan5:
Thanks for the fixes! The camera example looks fantastic! :D

PR #107, opened 2016-04-03
Cleaning of Makefile files

LelixSuper:
I've removed some useless code (GLEW recipes and duplicated code) and fixed the command for cleaning of `/src/` after you've compiled raylib for GNU/Linux.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -119,28 +119,6 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     endif
 endif
 
-# define library paths containing required libs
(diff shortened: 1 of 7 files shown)


raysan5:
Hi LelixSuper, I think you remove the most of the library paths -L , they could be required on some platform, like RASPBERRY_PI.

LelixSuper:
@raysan5 where? I removed the block about library paths -L, because there is a duplicate.

raysan5:
True, sorry, I didn't check it right the pull-request commit. :P

Issue #99, opened 2016-03-05
[audio] Chiptunes support


kd7tck:
Wondering if there was any support for this. I have been looking into it and am going this route: https://github.com/kd7tck/raylib/commit/8a16d76a6554bbf06969dc9d6bbae715c6f36a75

Wanted a little feedback, wondering if better libraries are out there.

raysan5:
Hi @kd7tck, I really like chiptunes-based module music (MOD, XM, IT, S3M) and I already was looking for some way to add support for it to raylib.

I checked the following libraries:

- libxm (https://github.com/Artefact2/libxm)
- libxmp (http://xmp.sourceforge.net)
- DUMB (http://dumb.sourceforge.net/index.php)
- libmikmod (http://mikmod.sourceforge.net)
- libmodplug (http://modplug-xmms.sourceforge.net)

Actually, all them are quite good and can be integrated with raylib but one of the raylib key design points is keeping dependecies to the minimum, not depend on external libraries unless they can be directly integrated into raylib basecode (like [stb libraries](https://github.com/nothings/stb)). For that reason I didn't integrate any of those libraries into raylib... yet. 

The only one in this list that is quite small and I think could be ported to single-header to be added to raylib is [libxm](https://github.com/Artefact2/libxm).

kd7tck:
Both DUMB and libxm require one external lib. However DUMB is ANSI C(DOS), and libxm is C11. Are you planning to add support for the modern version of C?

I am thinking about changing over to XMP, it seems easier to port into yours. DUMB works too but it does have these weird UNIX, DOS commands bundled in to maximize backwards compatibility.

kd7tck:
I want to make this my next project, I will make a stb_chiptunes header file. I plan to get a proto up and running in the next month or two. I hope to have an in game synth and a player that is incorporated directly into the playmusicstream function. The game I am working on is nearing completion and I totally forgot about music until the last minute.

I intend to make Raylib library the highlight of my game, the raylib logo will even show at the beginning. Maybe you could add a simple chiptune to the logo as it runs for 1.5.

By the way, thank you for making raylib.

raysan5:
Hi @kd7tck, that's fantastic! I have been looking for a single-header chiptunes file for long time but I couldn't find anything! As I told you, the closest approach was porting libxm to single file...

[...]

PR #109, opened 2016-04-24
Added proto version of jar_xm

kd7tck:
This is an early draft, needs lots of work. Still need to figure out way
to calculate total length of song. This is hard because xm tracks stream
out zeros when done, only position in track can be found. Position does
not give any direct value of how much more time is left. I think that by
setting the loop count to 1 and seeking until the end I can total up the
number of samples and come up with a length.


kd7tck:
This is for issue #99

PR #110, opened 2016-04-26
quick fix

kd7tck:
Boolean errors

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -261,8 +261,9 @@
 //----------------------------------------------------------------------------------
 #ifndef __cplusplus
 // Boolean type
-#include <stdbool.h>
-//typedef enum { false, true } bool;
+    #ifndef true
+        typedef enum { false, true } bool;
+    #endif
 #endif
 
 // byte type
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -261,7 +261,8 @@
 //----------------------------------------------------------------------------------
 #ifndef __cplusplus
 // Boolean type
-typedef enum { false, true } bool;
+#include <stdbool.h>
+//typedef enum { false, true } bool;
 #endif
 
 // byte type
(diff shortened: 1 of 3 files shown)


kd7tck:
This final patch fixes some problems from the last one I pushed.

src/raylib.h
```diff
@@ -261,7 +261,8 @@
 //----------------------------------------------------------------------------------
 #ifndef __cplusplus
 // Boolean type
-typedef enum { false, true } bool;
+#include <stdbool.h>
```
raysan5: Hi @kd7tck! Why do you need the stdbool.h? It works good with the typedef...

I did it that way for several reasons:

1) MSVC doesn't support this library until VS 2013 (http://stackoverflow.com/questions/8548521/trying-to-use-include-stdbool-h-in-vs-2010)
2) It's a new library from C99, C89 and C90 didn't support it. Despite raylib uses a minimal set of features of C99, I try to keep the full base-code as portable as possible.
3) It's easy to implement without the need of an external library (http://stackoverflow.com/questions/1921539/using-boolean-values-in-c). In raylib I always try to keep dependencies to the minimum.
4) I try to keep raylib.h clean of includes...

kd7tck:
Is it possible to cut and paste the contents of stdbool.h into that area. I find that defining the value of true and false allows me to have pre-processor branches that check for the existence of bool. enumerating them prevents the pre-processors from checking if it exists.

src/raylib.h
```diff
@@ -261,8 +261,9 @@
 //----------------------------------------------------------------------------------
 #ifndef __cplusplus
 // Boolean type
-#include <stdbool.h>
-//typedef enum { false, true } bool;
+    #ifndef true
+        typedef enum { false, true } bool;
+    #endif
```
raysan5: It's ok, that solution works!

PR #111, opened 2016-04-30
First stage of audio API update

kd7tck:
Look over changes and give feedback please.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -97,9 +97,10 @@ typedef enum { INFO = 0, ERROR, WARNING, DEBUG, OTHER } TraceLogType;
 //----------------------------------------------------------------------------------
 // Global Variables Definition
 //----------------------------------------------------------------------------------
+static bool mixChannelsActive_g[4]; // What mix channels are currently active
 static bool musicEnabled = false;
-static Music currentMusic;      // Current music loaded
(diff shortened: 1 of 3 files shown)


raysan5:
Hi @kd7tck, thanks for the update! Probably `AudioContext` struct could be unified (or combined) with `Music` struct at some point to get a single and generic `AudioContext`... but it's ok for now.

Personally, I would avoid `channel_t` and `mix_t` enums, I'd just use `int channels` and `int numMixChannel` (or `int numStream`).

I'd call new audio device check function: 
```c
bool IsAudioDeviceReady(void);
```
About max num of concurrent audio context, it can be generalized with:
```c
#define MAX_AUDIO_CONTEXTS  4
```
There are some other details on formating but I'll wait for a more complete version; as I said, I think that at some point `AudioContext` could be integrated/unified with `Music` struct.

kd7tck:
Alright I will get rid of the enum types. I figured it would stop errors from unsupported audio setups.  I plan to rewrite the music streaming type after I am done with this. The future version of music will simply be an abstracted system that wraps on top of AudioContext, something like the following:

```C
typedef Audiocontext* Music
```

raysan5:
That looks great to me! Thanks for the effort! :)

PR #115, opened 2016-05-15
Add support for opengl 2.1

dvolk:
Adds the ability to use shaders on older hardware.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -55,7 +55,7 @@
 
 #if defined(PLATFORM_DESKTOP)
     #define GLAD_EXTENSIONS_LOADER
-    #if defined(GLEW_EXTENSIONS_LOADER)
+    #if defined(GLEW_EXTENSIONS_LOADER) || defined(GRAPHICS_API_OPENGL_21)
         #define GLEW_STATIC
         #include <GL/glew.h>        // GLEW extensions loading lib
     #elif defined(GLAD_EXTENSIONS_LOADER)
@@ -1504,7 +1504,7 @@ static void InitDisplay(int width, int height)
 
             // NOTE: GLEW is a big library that loads ALL extensions, we can use some alternative to load only required ones
             // Alternatives: glLoadGen, glad, libepoxy
-        #elif defined(GLAD_EXTENSIONS_LOADER)
+        #elif defined(GLAD_EXTENSIONS_LOADER) && defined(GRAPHICS_API_OPENGL_33)
             // NOTE: glad is generated and contains only required OpenGL version and Core extensions
             //if (!gladLoadGL()) TraceLog(ERROR, "Failed to initialize glad\n");
             if (!gladLoadGLLoader((GLADloadproc)glfwGetProcAddress)) TraceLog(ERROR, "Failed to initialize glad\n"); // No GLFW3 in this module...
@@ -1516,6 +1516,15 @@ static void InitDisplay(int width, int height)
             //if (GLAD_GL_ARB_vertex_array_object) // Use GL_ARB_vertex_array_object
         #endif
     }
+    #if defined(GRAPHICS_API_OPENGL_21)
+    else if(rlGetVersion() == OPENGL_21)
+    {
+        GLenum error = glewInit();
+	if (error != GLEW_OK) TraceLog(ERROR, "Failed to initialize GLEW - Error Code: %s\n", glewGetErrorString(error));
+	if (glewIsSupported("GL_VERSION_2_1")) TraceLog(INFO, "OpenGL 2.1 profile supported");
+	else TraceLog(ERROR, "OpenGL 2.1 profile not supported");
+    }
+    #endif
 #endif
     
     // Enables GPU v-sync, so frames are not limited to screen refresh rate (60Hz -> 60 FPS)
--- a/src/makefile
+++ b/src/makefile
@@ -49,6 +49,7 @@ ifeq ($(PLATFORM),PLATFORM_RPI)
 else
     # define raylib graphics api to use (OpenGL 1.1 by default)
     GRAPHICS ?= GRAPHICS_API_OPENGL_11
+    #GRAPHICS = GRAPHICS_API_OPENGL_21  # Uncomment to use OpenGL 2.1
     #GRAPHICS = GRAPHICS_API_OPENGL_33  # Uncomment to use OpenGL 3.3
 endif
 ifeq ($(PLATFORM),PLATFORM_WEB)
--- a/src/rlgl.c
+++ b/src/rlgl.c
@@ -42,6 +42,15 @@
     #endif
 #endif
 
+#if defined(GRAPHICS_API_OPENGL_21)
+    #ifdef __APPLE__
+        // ?
+    #else
+        #define GLEW_STATIC
+        #include <GL/glew.h>
+    #endif
+#endif
+
 #if defined(GRAPHICS_API_OPENGL_33)
     #ifdef __APPLE__                // OpenGL include for OSX
         #include <OpenGL/gl3.h>
@@ -158,7 +167,7 @@ typedef struct {
     float *vertices;            // 3 components per vertex
     float *texcoords;           // 2 components per vertex
     unsigned char *colors;      // 4 components per vertex
-#if defined(GRAPHICS_API_OPENGL_11) || defined(GRAPHICS_API_OPENGL_33)
+#if defined(GRAPHICS_API_OPENGL_11) || defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_21)
     unsigned int *indices;      // 6 indices per quad (could be int)
 #elif defined(GRAPHICS_API_OPENGL_ES2)
     unsigned short *indices;    // 6 indices per quad (must be short)
@@ -197,7 +206,7 @@ typedef enum { INFO = 0, ERROR, WARNING, DEBUG, OTHER } TraceLogType;
 //----------------------------------------------------------------------------------
 // Global Variables Definition
 //----------------------------------------------------------------------------------
-#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
+#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2) || defined(GRAPHICS_API_OPENGL_21)
 static Matrix stack[MATRIX_STACK_SIZE];
 static int stackCounter = 0;
 
@@ -277,7 +286,7 @@ unsigned int whiteTexture;
 //----------------------------------------------------------------------------------
 // Module specific Functions Declaration
 //----------------------------------------------------------------------------------
-#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
+#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2) || defined(GRAPHICS_API_OPENGL_21)
 static Shader LoadDefaultShader(void);
 static Shader LoadSimpleShader(void);
 static void InitializeBuffers(void);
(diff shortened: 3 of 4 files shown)


raysan5:
Hi @dvolk, thank you very much for the update. Unfortunately, latest raylib version is being developed on [develop](https://github.com/raysan5/raylib/tree/develop) branch and it's quite far ahead `master` branch. GLEW dependency is being completely removed and replaced by GLAD and support for older versions of OpenGL was not planned. Actually, I was planning to remove support for OpenGL 1.1 at some point (current drivers just translate it automatically to latest version supported by the graphic card).

Why do you need OpenGL 2.1? Are you using it for any specific platform? Maybe Raspberry Pi desktop?

dvolk:
Oh, I see. Can GLAD work with 2.1?

I have an old laptop that only supports 2.1 so I thought it'd be neat to have it.

raysan5:
Hi @dvolk, [GLAD](https://github.com/Dav1dde/glad) should work ok with OpenGL 2.1; included version is for OpenGL 3.3 but I think it should work.

On the other hand, raylib only uses a small set of OpenGL 3.3 functions and tracks VAO support to use only VBOs in case no VAOs supported. It should work with OpenGL 2.1 with only some small tweaks...

Supporting old hardware is always tricky because it implies additional conditions to check and additional code to maintain so I decided to just stuck with OpenGL 1.1 for older devices and OpenGL 3.3 for current ones. It's true that shaders support is not available on OpenGL 1.1 but adding support for OpenGL 2.1 also implies adding support for GLSL version 120...

dvolk:
Well if you like I can work on it.

I don't know completely how different GLSL 120 and 330 are. For simple programs it seems it's just the way in/out variables are specified, so it shouldn't be much trouble.

raysan5:
`GLSL 120` is similar to `GLSL 100` used on `GRAPHICS_API_OPENGL_ES2`. The only change required is `#version 100` to `#version 120` and removing the `precision mediump float;` line in the fragment shader.

Supporting OpenGL 2.1 is not much trouble but I feel it adds unnecessarily complexity to the library instead of simplifying it. A really small segment of potential raylib users require this capability.

dvolk:
Raylib with the ES target does not build for me. I assumed it didn't work on desktops.

raysan5:
`GRAPHICS_API_OPENGL_ES2` is intended for Android, Raspberry Pi and HTML5 platforms, it could work on desktop using `WGL_EXT_create_context_es2_profile` extension because OpenGL ES 2.0 is essentially a stripped-down version of OpenGL 2.0 with a few added features... but raylib is not considering that case (again, it implies some tweaks and redesign).

The way you implemented it in this push-request was mostly right, it only required:
 - To be done in current raylib [develop](https://github.com/raysan5/raylib/tree/develop) branch
 - Replace GLEW extensions initialization by GLAD extensions initialization (crossing fingers)
 - Review internal shader conditions

dvolk:
Gotcha. So you would accept that?

raysan5:
Ok, I would accept the update... just wait a bit (one week) because rlgl module is being highly modified right now, to avoid conflicts.

Again, thank you very much for your addition.

dvolk:
Great! I'm travellng anyway at the moment so I don't have much time.

raysan5:
Hi @dvolk, if you can send me the changes commented in this thread I will integrate them in the develop branch. Thanks! :)

dvolk:
I started on this, but the develop branch doesn't build for me

```
dv@laptop:~/src/lib/raylib/src$ git log | head -3
commit 60232810d83c7ea3a52491f0b20444003be53358
Author: raysan5 <raysan5@gmail.com>
Date:   Fri Jun 3 19:00:58 2016 +0200
dv@laptop:~/src/lib/raylib/src$ make clean && make
rm -f *.o libraylib.a
Cleaning done
gcc -c core.c -O1 -Wall -std=gnu99 -fgnu89-inline -Wno-missing-braces -I. -I../src -I../external/glfw3/include -I../external/openal_soft/include -DPLATFORM_DESKTOP -DGRAPHICS_API_OPENGL_11
core.c: In function ‘InitDisplay’:
core.c:1723:28: warning: unused variable ‘modes’ [-Wunused-variable]
         const GLFWvidmode *modes = glfwGetVideoModes(glfwGetPrimaryMonitor(), &count);
                            ^
core.c: In function ‘StorageLoadValue’:
core.c:1091:13: warning: ignoring return value of ‘fread’, declared with attribute warn_unused_result [-Wunused-result]
             fread(&value, 1, 4, storageFile);
             ^
gcc -c rlgl.c -O1 -Wall -std=gnu99 -fgnu89-inline -Wno-missing-braces -I. -I../src -I../external/glfw3/include -I../external/openal_soft/include -DPLATFORM_DESKTOP -DGRAPHICS_API_OPENGL_11
rlgl.c: In function ‘SetMatrixProjection’:
rlgl.c:2256:5: error: ‘projection’ undeclared (first use in this function)
     projection = proj;
     ^
rlgl.c:2256:5: note: each undeclared identifier is reported only once for each function it appears in
rlgl.c: In function ‘SetMatrixModelview’:
rlgl.c:2262:5: error: ‘modelview’ undeclared (first use in this function)
     modelview = view;
     ^
Makefile:122: recipe for target 'rlgl.o' failed
make: *** [rlgl.o] Error 1
```

raysan5:
Ouch! I see, a couple of changes I did two days ago... 
I'll correct it today when I get home (in about 4-5 hours).

Thanks for pointing it out! :)

raysan5:
Ok, develop branch updated.

raysan5:
Checking the required changes to introduce OpenGL 2.1 support, I realized that it can be extremely simplified to the following lines at beginning of rlgl.c:
```c
#if defined(GRAPHICS_API_OPENGL_21)
    #define GRAPHICS_API_OPENGL_33
#endif
```
That way, won't be required to append `|| defined(GRAPHICS_API_OPENGL_21)` everywhere; most of the functionality used for OpenGL 3.3 is actually introduced in OpenGL 2.1.

Obviously, in some code points it would be required to check for `GRAPHICS_API_OPENGL_21`... basically, on default shader definition.

dvolk:
The examples load resources/shaders/standard.vs and resources/shaders/standard.fs, which fall back to the default shaders when they fail to load. The default shaders don't seem to render fonts any more though.

Also I'm not sure what causes the standard shader to be used in the examples.

raysan5:
Ouch! Yes, that part needs redesign... One of the additions for next raylib version i support for materials and lighting; it is accomplished using a standard shader (right now placed in resources/shaders/ folder).

Standard shader loading is done [here](https://github.com/raysan5/raylib/blob/develop/src/rlgl.c#L2576) and this function is called at raylib initialization, 

Standard shader needs to be embed into raylib (similar to default shader) and also adapted for every version of GLSL (100, 120, 330). It's also planned to do a lazy initialization for the shader (only if it is used).

Fallback to default shader was not properly implemented, I just added it right now.

raysan5:
Ok, I just finished everything I post a while ago... Standard shader hasn't been tested yet on OpenGL ES 2.0, most probably some functions are not supported... Need to review it, I'll try to do it tomorrow...

dvolk:
On 2.1 at least, there's no inverse(mat3), so that's going to be interesting.

raysan5:
Here it is some info: 
    http://stackoverflow.com/questions/22643604/glsl-es-2-0-inverse-matrix
    https://github.com/dsheets/gloc/blob/master/stdlib/matrix.glsl
    https://www.opengl.org/discussion_boards/showthread.php/197633-Help-converting-from-glsl-400-to-130

I'll take a look ASAP.

raysan5:
Shader has been adapted by @victorfisac on pull request https://github.com/raysan5/raylib/pull/128

raysan5:
Hi @dvolk, did you take a look to this issue? Does it work for you now?

dvolk:
I gave it a try but I can't get it to load the standard shader. Something's lost in the soup of #defines.

Here's the diff: http://pastebin.com/raw/uhxEyJyT

raysan5:
Ok, I'll try to update my code to support OpenGL 2.1.

raysan5:
Support for OpenGL 2.1 implemented in commit https://github.com/raysan5/raylib/commit/4df7a0f2f8cf5c4dde236bd99d05d83c7b472db5

Please, @dvolk, could you verify it works? In my PC it works ok.

dvolk:
It builds and the examples run, but all of the games are broken (many elements don't show, no errors).

Also is there an example that's supposed to load the standard shader? Because looking at the log they're only loading the default shader (+ some load extra from files).

So this might have the same problem I had.

raysan5:
Ok, now everything should work, I corrected some issues on commit https://github.com/raysan5/raylib/commit/9fdf4420d5354c32394246703015a686df8135ce

To test the standard shader, you should use example: [shaders_standard_lighting](https://github.com/raysan5/raylib/blob/develop/examples/shaders_standard_lighting.c).

dvolk:
Oh, I didn't have that one to test because an example before it fails to build.

```
dv@laptop:~/src/lib/raylib/examples$ make
gcc -o audio_music_stream audio_music_stream.c -O2 -Wall -std=c99 -I. -I../src -I/usr/local/include/raylib/ -L. -L../../src -lraylib -lglfw -lGL -lopenal -lm -pthread -ldl -lX11 -lXrandr -lXinerama -lXi -lXxf86vm -lXcursor -DPLATFORM_DESKTOP 
audio_music_stream.c: In function ‘main’:
audio_music_stream.c:27:21: warning: passing argument 1 of ‘PlayMusicStream’ makes integer from pointer without a cast [-Wint-conversion]
     PlayMusicStream("resources/audio/guitar_noodling.ogg");         // Play music stream
                     ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:879:5: note: expected ‘int’ but argument is of type ‘char *’
 int PlayMusicStream(int index, char *fileName);                 // Start music playing (open stream)
     ^
audio_music_stream.c:27:5: error: too few arguments to function ‘PlayMusicStream’
     PlayMusicStream("resources/audio/guitar_noodling.ogg");         // Play music stream
     ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:879:5: note: declared here
 int PlayMusicStream(int index, char *fileName);                 // Start music playing (open stream)
     ^
audio_music_stream.c:61:34: error: too few arguments to function ‘PauseMusicStream’
         if (IsWindowMinimized()) PauseMusicStream();
                                  ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:882:6: note: declared here
 void PauseMusicStream(int index);                               // Pause music playing
      ^
audio_music_stream.c:62:14: error: too few arguments to function ‘ResumeMusicStream’
         else ResumeMusicStream();
              ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:883:6: note: declared here
 void ResumeMusicStream(int index);                              // Resume playing paused music
      ^
audio_music_stream.c:64:22: error: too few arguments to function ‘GetMusicTimePlayed’
         timePlayed = GetMusicTimePlayed()/GetMusicTimeLength()*100*4; // We scale by 4 to fit 400 pixels
                      ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:887:7: note: declared here
 float GetMusicTimePlayed(int index);                            // Get current music time played (in seconds)
       ^
audio_music_stream.c:64:43: error: too few arguments to function ‘GetMusicTimeLength’
         timePlayed = GetMusicTimePlayed()/GetMusicTimeLength()*100*4; // We scale by 4 to fit 400 pixels
                                           ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:886:7: note: declared here
 float GetMusicTimeLength(int index);                            // Get current music time length (in seconds)
       ^
audio_music_stream.c:66:9: error: too few arguments to function ‘UpdateMusicStream’
         UpdateMusicStream();        // Update music buffer with new stream data
         ^
In file included from audio_music_stream.c:14:0:
../src/raylib.h:880:6: note: declared here
 void UpdateMusicStream(int index);                              // Updates buffers for music streaming
      ^
Makefile:421: recipe for target 'audio_music_stream' failed
make: *** [audio_music_stream] Error 1

```

PR #119, opened 2016-05-20
remove user-specific ant files

gordonmcshane:
This removes local.properties, an ant user-specific file to cache build variables.
They prevented me from building raylib/example for android on OSX.
Removing these fixed the problem.


raysan5:
Hi @gordonmcshane! Thank you very much for your contributions! :)

PR #120, opened 2016-05-21
Standard Lighting (2/3)

victorfisac:
- 3 light types added (point, directional, spot).
- DrawLights() function added using line shapes.
- Standard lighting example added.
- Removed useless struct variables from material and light.
- Fixed light attributes dynamic locations errors.
- Standard vertex and fragment shaders temporally added until rewrite it
as char pointers in rlgl.
- Improved spacing format.
- Fixed some audio module compile warnings.
- Remove deprecated lighting example and shaders.

TODO:
- Add normal and specular maps calculations in standard shader.
- Add control structs to handle which attributes needs to be calculated
(textures, specular...).
- Adapt standard shader to version 110.
- Rewrite standard shader as char pointers in rlgl.

standard_lighting_screenshot


raysan5:
Wow! Amazing update @victorfisac! An excelent work! Thank you very much! :D

PR #121, opened 2016-05-27
I made this for some others I am now working with

kd7tck:
I made this for some of the programmers I am working with right now. Figured you might have some use for it.


LelixSuper:
Does it work also for *nix systems (Linux/OSX)? I don't know cmake, so sorry if I can't understand!

raysan5:
Hi @kd7tck, thank you very much for your addition!

raysan5:
By the way @kd7tck, GLEW shouldn't be included, in last version it was replaced by GLAD.

@LelixSuper, I neither use CMake but I'm pretty sure it supports Linux building...

PR #122, opened 2016-05-30
Standard Lighting (3/3)

victorfisac:
- Added normal and specular map to drawing process and standard shader.
- Changed point light attenuation to radius.
- Improved DrawLights().
- Added Draw3DCircle to draw point lights radius.
- Removed normal depth due a useless result and some specular issues.
- Added full tint color to material data type (ignores lighting).
- Updated rlgl and raylib header files with new changes.

TODO:
- Rewrite standard shader as char pointer with version 110 compatibility. (Ray :P)


raysan5:
Wow! Amazing work Victor! Thank you very much for all your effort and your time invested in this big raylib addition! :D

raysan5:
By the way, I will take care of the standard shader transcription to version 100. Still thinking the best way to embed it into raylib base code...

PR #123, opened 2016-06-02
mod player

kd7tck:
Test this out a little, It works on my end. Finally finished the MOD player, the player was modified extensively from the original version called hxcmodplayer.

I'm starting to wind down my work on the game engine and toolset, and I think this version of the raylib API will end being the one we use in the game.


raysan5:
Wow! A big addition of changes! Just a couple of comments:

 - OpenAL update to latest version (I imagine it's 1.17.2): It's ok but I'm a bit concerned about the fact that version 1.15 was only about 400KB for the .dll and new version 1.17 is about 900KB... just wondering what added to this new version to double it's compiled code size!
 - Why we have `jar_xm_context_t` but not `jar_mod_context_t`?
 - I'm not a fan of prepending (or postpending) 'g_' for global variables...
 - I've seen you return some code errors, it would be great to create an enum with those error codes for better organization (and reference)
```c
typedef enum {
    ERROR_CONTEXT_CREATION = 1,
    ERROR_MIXCHANNEL_CREATION,
    ERROR_LOADING_XM,
    ERROR_LOADING_MOD,
    // ...
} AudioError;
 - Would it be possible to have some simple usage example?

Again, thank you very much for your effort. Great work!

**Comment by kd7tck** [comment]
The new lib does not come from the official OpenAL site. This one is the newer fork that was made, for some unkown reason that fork includes new backends. This I guess increases the size, supposedly it supports the modern directsound system in windows. Where do you want the example posted to?

Here is a very simple one for playing two songs at once:
```C
int intro_load(void)
{
    InitAudioDevice();
    AudioError er = PlayMusicStream(0, "assets/audio/1.mod");
    AudioError er2 = PlayMusicStream(1, "assets/audio/2.xm");
    if(er || er2) handleError();
    return 1;
}

int intro_unload(void)
{
    CloseAudioDevice();
    return 1;
}

int intro_tick(long counter)
{
    UpdateMusicStream(0);
	UpdateMusicStream(1);
    BeginDrawing();
        ClearBackground(RAYWHITE);
    EndDrawing();
    
    return 0;
}
```

raysan5:
Hi @kd7tck, I'm integrating your pull-request but I'll make some review changes...

PR #126, opened 2016-06-08
clean up

kd7tck:
Consolidated allot of stuff.


raysan5:
Thanks for the update @kd7tck !

PR #127, opened 2016-06-09
Fix Android template with 1.5 raylib features (audio streaming, gestures)

victorfisac:


raysan5:
Thanks for the update @victorfisac!

PR #128, opened 2016-06-09
Adapted shaders for OpenGL ES 2.0

victorfisac:
Standard lighting now is compatible with OpenGL ES 2.0. I had to move some calculations from GPU to CPU because some functions doesn't exist in glsl 100: inverse(), transpose(). Besides, the code has been tweaked to avoid implicit casts because GLSL 100 doesn't accept it (these changes are done in glsl 330 code too).

Other shaders have been updated using texture2D() function instead of texture().


raysan5:
Great! :D

PR #129, opened 2016-06-11
Improved physac module (fixed time steps)

victorfisac:
Note: in physac module, I just removed steps for loop and replaced /PHYSAC_STEPS by *deltaTime (git changes colors are so crazy and it is so difficult to understand it).


raysan5:
Wow! That's great Victor! Please, could you update the examples? Is still posible to run an application as single-threaded?

victorfisac:
Yes, I can using the template that I sent you by email (creating PhysicsThread() in each example). 

I tested to calculate delta time and call update physics in main thread and it doesn't work well, collisions are calculated so bad so rigidbodies starts to overlap the floor... 

To avoid that, I can do a for to repeat UpdatePhysics() a number of times and divide deltaTime by that number... But it would be the old method by a static number of steps, so it doesn't have sense. The unique solution is to update physics in a second thread...

victorfisac:
- Updated physac examples.
- Updated core and raylib header with GetTime() changes.

victorfisac:
I close this pull to continue working on module until next update.

PR #131, opened 2016-06-14
Physac 1.0 module completed

victorfisac:
- Physics calculations in secondary thread.
- Define PHYSAC_NO_THREADS to avoid using pthread library functions. So, user must create a thread using its own libraries to call void *PhysicsThread(void *args).
- Physics examples updated with new changes.
- Instructions to compile physics examples added as comments in code header.
- Switched from static physics steps to fixed time steps (as much steps as CPU without v-sync can do)
- Added internal time management GetCurrentTime() and InitTimer() to calculate physics delta time.
- Updated physac header documentation.


raysan5:
Amazing work @victorfisac! Congrats! :)

PR #148, opened 2016-07-22
OS X compatiblity changes and compiled library

sol-prog:
Hi,
I've added some changes in order to be able to build libraylib.a on Mac OS X El Capitan.
Regards,
Sol


raysan5:
Hello @sol-prog! Thank you very much for the update! Unfortunately, I don't have a Mac and I can not try raylib on OSX, all the updated come from people using it! :)

Thanks! :)

PR #172, opened 2016-08-04
Fix install command of src/ makefile

LelixSuper:


raysan5:
Many thanks for finding and correcting this bug! :D

PR #173, opened 2016-08-06
Fixes for some Lua bugs

ghassanpl:
Here are the changes I could make without messing up a lot of stuff. Most of the bugs are fixed. The main one had to do with using relative lua stack indices in GetArgument_* functions. Almost all of the other ones were in Lua files.


raysan5:
Great! So fast! Many thanks for the update! :D

PR #175, opened 2016-08-06
More fixes for Lua

ghassanpl:
GetDroppedFiles and SetShaderValue in Lua working
Exposed Texture2D.id to Lua
Lights now have settable/gettable fields

Forgot to look at CheckCollisionRayBox, gonna do that in next commit.


raysan5:
Many thanks for the update! :)

PR #180, opened 2016-08-12
Small fix for GenMeshCubicmap()

teodor-stoenescu:
This fix allows GenMeshCubicmap() to create cubic maps having cells of
arbitrary sizes.


raysan5:
Great addition @teodor-stoenescu! Thank you very much for the fix! :)

PR #188, opened 2016-08-31
Develop

teodor-stoenescu:
Greater LoadOBJ() flexibility:
- LoadOBJ can load objects having no texture coordinates.
- LoadOBJ can load objects having texture coordinates specified as tuples (the way blender exports OBJs)


raysan5:
Hi @teodor-stoenescu! Thank you very much for the improvement! :D

PR #206, opened 2016-12-15
Added SetupViewport so high-DPI fix applies to EndTextureMode

joeld42:
High-DPI (retina) mode was not working on macbook when EndTextureMode was used. This is because EndTextureMode would call glViewport with the non-retina screen size. There was already a fix for this in init graphics but it wasn't applied to EndTextureMode. You can see this if you run the rendertexture sample on a high-dpi macbook.

This change simply moves the glViewport call (with the apple fix) out into it's own function, SetupViewport, and calls it from both EndTextureMode and InitGraphics.


raysan5:
Thank you very much for the fix @joeld42! Please, could you move this commit into develop branch instead of master?

PR #207, opened 2016-12-15
Add example artifacts to .gitignore

ficoos:
Adding the example binaries to the gitignore file as they should be ignored by git.


ficoos:
Closing as I mistakenly based the PR on master instead of develop

PR #208, opened 2016-12-15
Add example artifacts to .gitignore

ficoos:
Adding the example binaries to the gitignore file as they should be ignored by git.


raysan5:
Hi @ficoos, thanks for the improvement but I think it would be better to use this more generic approximation:

# Ignore all examples files
/examples/*
# Unignore all examples dirs
!/examples/*/
# Unignore all examples files with extension
!/examples/*.*
#Unignore examples Makefile
!/examples/Makefile

raysan5:
Just updated myself on commit https://github.com/raysan5/raylib/commit/060d501cd1a8b591d38b3169ee98c416314261a7

Thanks anyway for pointing this out! :)

PR #211, opened 2016-12-31
Fix vbo indices for rlglUpdateMesh

joeld42:
Fixes vbo indices for normals in rlglUpdateMesh. Also similar fix to color, tangents and texcoord2 but I haven't tested these since I'm not currently using.


raysan5:
Hey @joeld42! Thank you very much for the fix... and happy new year! :)

PR #212, opened 2017-01-03
Raycast Picking Utilities

joeld42:
Adds support for raycasts against a ground plane, a triangle, or a mesh. 
Includes an example/test code using this for mouse picking.
Also included code to evaluate barycentric coordinates to raymath.


raysan5:
Just merging it but as commented by mail I'm doing several changes, mainly naming and coding style related:

1. Renaming functions to be more consistent with `CheckCollisionRay*()`:
```c
RLAPI CollisionRayInfo GetCollisionRayMesh(Ray ray, Mesh mesh);
RLAPI CollisionRayInfo GetCollisionRayTriangle(Ray ray, Vector3 p1, Vector3 p2, Vector3 p3);
RLAPI CollisionRayInfo GetCollisionRayGround(Ray ray, float groundHeight);
```
2. `GetCollisionRayGround()` looks quite specific to Y-axis, could it be replaced by `GetCollisionRayPlane()`? What do you thing? Maybe is over-complicated for the user...
3. Moving all those functions to `models.c`, seems more appropiate to me.
4. `Barycenter()`, not sure where to put that, probably is right being in `raymath.h` but I don't like to force the user to include it when using raylib... 
5. Code formatting to raylib code style (spacing, braquets).

Just another detail, you use `sprintf()` for text drawing but raylib already provides that functionality for you:
```c
// Your code:
char line[1024];
sprintf(line, "Hit Object: %s\n", hitObjectName);
DrawText(line, 10, 30, 15, BLACK);

// Using raylib provided functionality:
DrawText(FormatText("Hit Object: %s", hitObjectName), 10, 30, 15, BLACK);
```

PR #213, opened 2017-01-03
added RPi touchscreen event processing thread

AudioMorphology:
Created touchThread that processes events from /dev/input/event4 on a Raspberry Pi, and updates the existing mouse variables and left mouse button. 

event4 is a virtual uinput device created by the ts_uinput daemon.

Basic usage:
- install tslib on raspberry pi (or clone & make)
- run ts_calibrate to create calibration data for the particular touchscreen
- run ts_uinput as a daemon, and verify the event number for the virtual event stream (event4 in my case)


raysan5:
Hi @AudioMorphology, I'm integrating this feature into raylib. 

Please, could you do the PR to develop branch instead of master and replace all TABS by 4 spaces?

PR #230, opened 2017-02-23
Examples Linux Physics Fix

OdnetninI:
According to Man Pages, timespec with std=c99 needs "_POSIX_C_SOURCE 199309L". Due a some kind of bug, "#define _POSIX_C_SOURCE 199309L" doesn't work, so we must use "-D_POSIX_C_SOURCE=199309L" in compiler options.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -497,6 +497,13 @@ audio_module_playing: audio_module_playing.c
 audio_raw_stream: audio_raw_stream.c
 	$(CC) -o $@$(EXT) $< $(CFLAGS) $(INCLUDES) $(LFLAGS) $(LIBS) -D$(PLATFORM) $(WINFLAGS)
 
+# Linux Fix to timespect from 
+ifeq ($(PLATFORM),PLATFORM_DESKTOP)
+    ifeq ($(PLATFORM_OS),LINUX)
+        CFLAGS += -D_POSIX_C_SOURCE=199309L
+    endif
+endif
+
 # compile [physac] example - physics demo
 physics_demo: physics_demo.c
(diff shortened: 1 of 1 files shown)


OdnetninI:
If "-D_POSIX_C_SOURCE=199309L" is not used, physics examples will not compile.

raysan5:
Have you tried using `std=gnu99` ?

OdnetninI:
Yes, but if the code is written under c99 standar, why include gnu extensions?
In future may be problematic

raysan5:
Sincerely, I'm not an expert on c99 vs gnu99 differences but it seems that GCC lacks some features of c99 and so requires enabling gnu99 in some cases, an example is compilation for RPI, it required gnu99.

But I agree that c99 standard is the way to go (despite CLang uses gnu99 by default)... also, it seems gnu99 would be deprecated at some point...

Reference: http://stackoverflow.com/questions/5313536/whats-the-difference-between-gnu99-and-c99-clang

raysan5:
By the way, the source code solution, did you defined `_POSIX_C_SOURCE` before including `time.h`?
```c
#define _POSIX_C_SOURCE    199309L       // Before including "time.h"
```

OdnetninI:
Yes, I did, but later I can test it again in my main computer.

raysan5:
ok, merging!

OdnetninI:
It doesn't work with 
`#define _POSIX_C_SOURCE    199309L       // Before including "time.h"`

captura de pantalla de 2017-02-23 19-06-40

OdnetninI:
With the Makefile workaround, all works perfectly:

captura de pantalla de 2017-02-23 19-09-55

PR #234, opened 2017-03-02
fix for audio.c(607): error C2036 in Visual Studio.

bugcaptor:
Fix for audio.c(607): error C2036: 'void *': unknown size in Visual Studio 2015.


bugcaptor:
Is it better to separate the code by _MSC_VER definition?

raysan5:
I've been thinking about the best way to write this... I think it's clearer to use `void *` to note type-independence but it's true that it's basically a bytes buffer...

I'm merging your change.

PR #246, opened 2017-03-21
staging-linux-gitignore

RDR8:
Hi Ray, you may have a reason for leaving .a and .so build products in. Is the top line supposed to catch them? If this is helpful, I may comb through the games/$SUBS and list those too.

I have a couple of other branches I'm working on. One is to clean and unify all the makefiles in the repo. Basically, anywhere I find a spot fix and it's helpful elsewhere... The second is related to c99 and __linux, basically copying the recent fix #230 to src/Makefile allows to compile as c99 without tinkering with ifdefs in source files.

Thanks for the opportunity to contribute. I'm learning a lot.


raysan5:
Hi @RDR8! Thanks to you for contributing!

About ignoring .a and .so, those files are required for [release distributions](https://github.com/raysan5/raylib/tree/master/release). I try to keep a stable version of the library on those folders.

About ignoring games folder, I don't understand what's the pourpose of that, additionally, there can be files like .wav, .ogg, .fnt... that shouldn't be ignored.

RDR8:
I understand about the built libs.

With games/, I was imitating what was done with examples/ to keep binaries out of the repo. Maybe not necessary but it looks better on my end.

PR #251, opened 2017-03-29
Restore inadvertant changes to Makefiles

RDR8:
Restore /usr/local/include/raylib to INCLUDES in examples/Makefile. Is in use by at least one user.
Modify CFLAGS for PLATFORM_WEB in examples/Makefile and src/Makefile to reflect emscripten recommended optimizations.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -87,7 +87,7 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     endif
 endif
 ifeq ($(PLATFORM),PLATFORM_WEB)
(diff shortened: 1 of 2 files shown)


raysan5:
Hi @RDR8, about emscripten optimization level, I was using -O1 to avoid an external memory initialization code file (.mem), that data is just embedded as text in the .js.

If -O2 is used, it would be better to also set `--memory-init-file 0` (https://kripken.github.io/emscripten-site/docs/tools_reference/emcc.html#emcc-memory-init-file), just to simplify publishing...

raysan5:
By the way, have you tried latest emscripten version 1.37.x?

RDR8:
I have to admit that this is untested. I was just trying to repair the deletions I made initially. I still need the emscripten SDK. On the list with cmake.

PR #261, opened 2017-04-14
Update .gitignore for new examples/ arrangement.

RDR8:
Descend into new sub-directories to scrub executables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,8 @@ examples/*/*
 !examples/*/*.c
 !examples/*/*.lua
 !examples/*/*.png
+!examples/*/*.html
+!examples/*/*.js
 # Unignore examples Makefile
 !examples/Makefile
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -62,6 +62,8 @@ examples/*/*
 !examples/*/*.c
 !examples/*/*.lua
 !examples/*/*.png
+!examples/*/*.html
```
RDR8: Bad commit message. DON'T ignore html and js files.

raysan5:
Hi RDR8! Thanks again for the update! Actually, at this moment I'm working in the examples, redesigning, reorganizing, adding, removing...

Just a comment, .html and .js can be ignored (they are generated files, like .exe). :)

RDR8:
I understand about the .html and .js. I'm not sure what to do with them. They need to be deleted before each rebuild with emcc. Maybe adding PLATFORM_WEB to the clean: target in examples/Makefile is a better approach. I know you're busy getting ready for 1.7 so I'll pay attention and try to stay out of the way. If there is something specific you need, I'm happy to help.

PR #264, opened 2017-04-21
Update Makefile

Triangle345:
basic_game - Removed glew from linker flags because on windows this fails to compile with the installed raylib package.


raysan5:
Hi Triangle345! Thanks for the fix! Could you push it to develop branch, it's the branch with latest changes. :)

Triangle345:
sure!

PR #266, opened 2017-04-21
basic_game - update makefile

Triangle345:
Removed glew from basic_game makefile as glew is not part of the raylib package (no glew in mingw).


raysan5:
Thanks for the update @Triangle345! Actually, it remind me I need to update all Makefiles for templates... :P

PR #295, opened 2017-05-29
Accept external CFLAGS

jubalh:
Some distributions build the programs with a specific set of CFLAGS.
Lets allow that by adding raylibs flags.


raysan5:
Thank you very much! :)

Issue #176, opened 2016-08-09
[build] CMake support


Leinnan:
Any chance for cmake support? It would make developing on Linux much easier.

raysan5:
Hi @Leinnan, I'm not used to CMake... Someone created a [CMakeList.txt](https://github.com/raysan5/raylib/blob/develop/src/CMakeLists.txt), don't know if its ok...

raysan5:
Any help on this issue is hightly appreciated!

oswjk:
Hello! I've done some work with CMake and I've got an example of what I did with raylib here:

https://github.com/oswjk/raylib/tree/cmake
https://github.com/oswjk/rayskeleton

I took a liberty of not using the binaries bundled in the git repository (in my opinion, they do not belong there). So my approach was to download and build the dependencies with CMake.

I've tested this stuff on Xubuntu 16.04  (GCC) and Windows 10 (using MinGW) at the moment and all the dependency downloading and building seems to work fine. I need to create a little bit bigger skeleton to test all the audio stuff and other functionality so that they work also.

I'm not fully satisfied with my changes yet and there is probably some stuff missing also. If someone wants to adapt this, your welcome :-)

Leinnan:
NICE! I will try check it out in two weeks because now I have to much work
with starting studying on university.

Can't wait to start working with Raylib :D



2016-09-26 20:46 GMT+02:00 Oskari Timperi <notifications@github.com>:

raysan5:
Hi @oswjk! Thank you very much for this great addition! 

Sorry, I'm not used to CMake build system... but I will try it, you can send me a pull-request if you want to integrate it into raylib develop branch.

In terms of build system, on raylib develop branch I just added support for shared library compilation (check [Makefile](https://github.com/raysan5/raylib/blob/develop/src/Makefile#L39)) and [VS2015 preconfigured projects](https://github.com/raysan5/raylib/tree/develop/project/vs2015). Adding support for CMake is great!

rayskeleton looks also very interesting! Don't know if I understand it correctly but it is a kind-of raylib-project CMake template ready to use, am I right?

oswjk:
I'm not comfortable yet with integrating the CMake build system, as there are untested and missing things (RPi for example). I can test the RPi myself as I have one somewhere :-)

I'm also not yet fully satisfied with the bootstrapping part of my additions. I need to think about it a bit more, if I could make it better.

[...]

PR #297, opened 2017-06-05
Initial meson support

jubalh:
I would like to add this initial meson support, so far working for Linux, along the Makefile.
Once the meson support is also done for Windows, Android and Web one might drop the Makefile.

I rebased my commits since in the first one I had a mistake, because of this 1adc0313015b30cf061a3bc9d47292f82dc649f8 has the 'with jubalh'. Sorry for that one.

This commit also regards https://github.com/raysan5/raylib/issues/176
I suggest to close that issue since the initial request was to support CMake to have easier building on Linux, which is now solved through meson.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,5 @@ src/libraylib.bc
 !release/rpi/libraylib.a
 !release/win32/mingw32/raylib.dll
 
+# Meson build system
+builddir/
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ features
    *  Basic 3d support for Geometrics, Models, Billboards, Heightmaps and Cubicmaps
    *  Flexible Materials system, supporting by default diffuse, normal and specular maps
    *  Shaders support, including Model shaders and Postprocessing shaders
-   *  Powerful math module for Vector, Matrix and Quaternion operations: [raymath](https://github.com/raysan5/raylib/blob/master/src/raymath.c) 
+   *  Powerful math module for Vector, Matrix and Quaternion operations: [raymath](https://github.com/raysan5/raylib/blob/master/src/raymath.h)
    *  Audio loading and playing with streaming support (WAV, OGG, FLAC, XM, MOD)
    *  Multiple platforms support: Windows, Linux, Mac, **Android**, **Raspberry Pi** and **HTML5**
    *  VR stereo rendering support with configurable HMD device parameters
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,14 @@
+project('raylib', 'c', version:	'1.7.0',
+		license: 'zlib',
+		meson_version: '>= 0.39.1')
+
+cc = meson.get_compiler('c')
+
+glfw_dep = dependency('glfw3')
+gl_dep = dependency('gl')
+openal_dep = dependency('openal')
+x11_dep = dependency('x11')
+m_dep = cc.find_library('m', required : false)
+
+subdir('src')
+
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,0 +1,19 @@
+install_headers('raylib.h')
+
+source_c = [
+  'audio.c',
+  'core.c',
+  'models.c',
+  'rlgl.c',
+  'shapes.c',
+  'text.c',
+  'textures.c',
+  'utils.c',
+  'external/stb_vorbis.c',
+]
+
+raylib = shared_library('raylib',
+                        source_c,
+                        dependencies : [ glfw_dep, gl_dep, openal_dep, m_dep, x11_dep],
+                        install : true)
+
--- a/src/meson.build
+++ b/src/meson.build
@@ -12,8 +12,10 @@ source_c = [
   'external/stb_vorbis.c',
 ]
 
-raylib = shared_library('raylib',
-                        source_c,
-                        dependencies : [ glfw_dep, gl_dep, openal_dep, m_dep, x11_dep],
-                        install : true)
+# use 'meson --default-library=static builddir' to build as static, if no builddir yet exists
+# use 'mesonconf -Ddefault_library=static builddir' to change the type
+raylib = library('raylib',
+                  source_c,
+                  dependencies : [ glfw_dep, gl_dep, openal_dep, m_dep, x11_dep],
+                  install : true)
 


raysan5:
Thanks for the addition @jubalh! Just wondering... would it be possible to move all building files into a build folder? Not sure if that's possible or appropiate...

jubalh:
I think it is expected that the meson.build is at the root location. And then having more meson.build files in subfolders (for subprojects if there are any).

BTW: meson then can create project files for Visual Studio (etc).

jubalh:
@raysan5 is this commit okay? :)

raysan5:
Just been investigating a bit about meson to see how it works and I've seen it depends on [Python](https://github.com/mesonbuild/meson/blob/master/docs/markdown/Use-of-Python.md) and Ninja, personally I don't like dependencies. In comparison to [Premake](https://premake.github.io/), it only depends on a small C program (a single executable file).

In terms of syntax, Premake syntax is similar to Lua but I really like both.

In terms of performance [both look similar](https://sourceforge.net/p/meson/wiki/Simple%20comparison/).

In terms of target support, [Premake supports multiple targets](https://github.com/premake/premake-core/wiki/What-Is-Premake#key-features) but I couldn't find [detailed info on the targets](http://mesonbuild.com/index.html#features) for Meson, at least regarding [Visual Studio projects](http://mesonbuild.com/Using-with-Visual-Studio.html).

Also seen that big companies like [Blizzard use Premake](https://github.com/premake/premake-core/wiki/Who-Uses-Premake) and [meson is used on more Linux-centric projects](http://mesonbuild.com/Users.html).

Well, that's just a bit of research, before adding anything to raylib I always do that, I try to find the best solution or at least the most simple and functional one.

In any case, I don't have time right now to work on a build system so Meson proposal is right to me, as long as keeping the plain Makefiles, my primary building system for the moment.

jubalh:
Of course.

Honestly I never saw a project using Premake. Will look into it too.

Regarding meson and Visual Studio I saw that there is the following command: `meson --backend=vs`.

RDR8:
I support of this initiative, including the larger goal of extending support to additional distributions.  From my perspective as a beginner, learning to use the basic tools, like make and gcc, is a larger hurdle than learning to read the code in raylib. Throw in git and a full-featured build system and the learning curve steepens. Ray has done a good job presenting instructions so that Debian-based users can get up-and-running in minutes but I spent days learning how to build raylib on Solus Linux. It was just the basics. Spartan programming, as Ray calls it. Kind of like eating my carrots. I enjoy learning about these things but a more casual investigator on a non-Debian distribution may not see a hurdle but a barrier. To the point, anything that can be done to expose more people to raylib itself, in an opaque way, must be a good thing.  
  
Is it possible or even desirable to add another level of if statements to the makefile or build file to detect a particular distribution or file system arrangement so that we know where to place the raylib binary and header? Right now, it's a manual process. For example, if I want to use Solus' GLFW package, I have to know that it's a shared library /usr/lib/libglfw.so and edit the makefile accordingly. Furthermore, the location to install the raylib library after it's built or installed as a package may be different. Since Solus and Debian, at least, prefer to package a shared library and not the default static build, the final location wouldn't be /usr/local/lib since that's for third-party stuff. Solus doesn't even have that directory by default. It's their way of encouraging me to contribute by building a package. (^;  
  
A possibly undesirable consequence of distributing a shared raylib is that now it's a dependency for any games made with it. Seems like a burden for the gamer herself. Please, correct me if I'm wrong. Anyway, it's a big plate of spaghetti that I'm happy to share with you.  
  
As to which build system, if any, to adopt, meson seems to be far more common among major projects than premake. The dependencies should be easily satisfied. Python is there by default on every distro I've tried. Lua, used by premake, is generally available but may not come pre-installed or even be part of development meta-packages. I can't speak to general availability of ninja but I expect it's in every up-to-date distro's repository. The one encounter with premake I've had was as the build system for the tutorials for  [Learning Modern 3D Graphics Programming](http://opengl.datenwolf.net/gltut/html/index.html). I didn't proceed because premake isn't packaged for Solus yet.  
  
Anyhow, I'd like to assist but, as always, I have to learn how first. Thanks @raysan5 for raylib and @jubalh for keeping the ball rolling.

jubalh:
Thanks for your support @RDR8.

> A possibly undesirable consequence of distributing a shared raylib is that now it's a dependency for any games made with it. Seems like a burden for the gamer herself. Please, correct me if I'm wrong. Anyway, it's a big plate of spaghetti that I'm happy to share with you.

No thats not a problem :)
On Linux all libraries are shared libs. Why?
* You save space by only needing it once and not having it in all the programs that use it
* If there is an update/security fix there is only one place where you need to do the update (as a packager, the user doesnt have to care), and you dont have to check all the other packages and update their internal version of the lib

Actually having it as a shared lib makes it much easier for users, when the distros start to package raylib they only need to install it via their package manager. If they install a game that uses raylib it just pulls in the dependency automatically via their package manager (for open source games)

Now of course there will be people who distribute non OSS games or who want to link statically for other reasons still. Also this is no problem since meson allows easily to switch between dynamic and shared library building: `--default-library=static`

> I can't speak to general availability of ninja but I expect it's in every up-to-date distro's repository
Ninja will be in all major distros because some (more and more) GNOME packages depend on it for building.

So far I only created the meson file to build for Linux. I will leave it at that for now since @raysan5 wants to keep Makefiles for now. However it shouldn't be hard to write the corresponding code for the other platforms.

@raysan5 pointed out that this will introduce more dependencies, which is true. Though they are only build dependencies, meaning the developer will need to have them but the user who plays the game will not.
Since I think Windows also doesn't ship with make by default there is not much of a difference of having meson as a dependency or make.

RDR8:
Does this mean that compiler options need to be moved out of the makefiles into source files? Or will meson.build grow to include them?

jubalh:
If you write the options before the meson call they will be used.

Usually distributions have a standard set of options they use, so if packagers grab raylib in its current state this will just happen automatically, depending on their invocation of meson. Usually they have a macro which calls it with the desired options.

There is also the possibility of defining them in the meson build file: http://mesonbuild.com/Adding-arguments.html

RDR8:
> On Linux all libraries are shared libs. Why?  

Just thinking in the context of a game jam or other situation where one might want to quickly share something offline.  
  
I don't know enough to properly discuss software distribution but I do generally understand the importance of shared libs and centralized code. A rolling release like Solus wouldn't exist without them and I love the speed of system software updates. And yet we have Flatpak and and other static distribution methods emerging. May it be appropriate to use them for software, like any particular game, that has a relatively low installation base yet needs wide compatibility?

RDR8:
Thank you. I'm looking forward to testing this out. There's a lot I don't understand about runtime and build dependencies, dev headers and such. Hopefully I can get up to speed and start pushing packages. (^;

raysan5:
Hey @RDR8! Actually, one of the goals of raylib is avoiding any external dependency, I mean, in current raylib Windows distribution, you generate self-contained executables, with no external library dependencies.

I believe in Linux the common usage of libraries is different, you usually install required dependencies as shared libraries that are not compiled into the final game binary.

RDR8:
The convergence of development philosophies and practices around open source software is interesting.  
  
On the one hand, with Solus, I get once or twice weekly updates with required, security, and optional packages. I can pick and choose if I need to and have mechanisms to roll back if something breaks. The whole process never takes more than 5 minutes unless the network is jammed. Yay, Linux!  

On the other hand, also with Solus, I want to use the Solus native Steam client instead of the regular Valve client. Both are available via a toggle app that's installed with the Steam package. Doing this gives me an extra 2 -4 fps on Intel laptop graphics. Cool! Innovative! But wait... Some games require infrequently used libraries. Such a game may or may not launch because Solus is by nature a lean distribution and may not include a needed library by default. Luckily for me, the development team at Solus loves games too and within a day or two after reporting the event, the needed library will come in with a new Steam client update.  That's great and far better than the most recent Ubuntu LTS, which, for example, has a GLFW package that's six! years old, forcing me to procure cmake and trust it to build GLFW if I want to dabble with raylib.  But I still just want to play! What ends up happening is that I toggle over to the Valve client which includes all the libraries any of my games need. Problem solved. So, 2 - 4 fps, or days of my life? IN THIS CASE, the choice is easy.  
  
Speaking of choice, Thank you! @jubalh and @raysan5  for making such choices available. I'm learning a lot and not just about video games. 
  
I realize I'm talking about differences in the perspectives of end users and developers and that most people coming to raylib will be interested and willing to do some work as they explore but I think it's important to remember that raylib is advertised to beginners. Raylib would be a perfect companion to an primer on general development practices. Learn how to develop software on with Unix-style tools and end up with a video game! It's a book I couldn't write today but it's out there somewhere... Ray?  

Apologies for the long ramble but y'all got me to thinking. I'm off to storm the castle. Have a great weekend. (^;

raysan5:
Hey @RDR8, great ramble! :D

 > I realize I'm talking about differences in the perspectives of end users and developers and that most people coming to raylib will be interested and willing to do some work as they explore but I think it's important to remember that raylib is advertised to beginners.

Completely agree with that point. raylib was created for students because I found a need on my course. I've been using it and refining it along four years with the feedback of more than 350 students. I know there are multiple good practices on software development that I have not followed with raylib dev, specially in terms of building and versioning - every new version of raylib introduces multiple API and ABI changes.

raylib was originally developed for Windows, that's the system we use in our institution, the course where I teach is focused on 3D Animation and Videogames (mostly art side) and most of the software is for Windows; my students profile is primary art-oriented, most of them have never written a single line of code. I needed a tool easy to install and easy to use, with minimal dependencies, to start typing code in seconds with no previous knowledge.

From that point, raylib has grown A LOT, mostly because I enjoyed the project and thanks to the continuous students feedback that motivates me to keep improving and adding features (multiplatform, shaders, vr, samples, templates...).

I imagine that if I want to reach a wider audience in the academic field, improving Linux support (at least in terms of availability and building) is A MUST. Unfortunately, I'm not an expert on Linux packaging (just been a simple user of the system, long time ago), so, any help on that direction is highly appreciated.

 > raylib would be a perfect companion to a primer on general development practices. Learn how to develop software with Unix-style tools and end up with a video game you can give to your friends! Market it to teens using the Raspberry Pi and Geany as the IDE. It's a book I couldn't write today but it's out there somewhere... Ray?

Sincerely, it's in my TODO list... but need more time, I even considered dropping my current job to focus on raylib: youtube tutorials, my twitch channel, educational material, complete and guided game samples, teaching book... but it's not an easy decision to take... and current raylib users-base is quite low I think...

RDR8:
> current raylib users-base is still low I think...  
  
How can I help change that?  
  
@jubalh I need to get familiar with meson before I can follow your lead and produce a distributable package for any distribution. I'm thinking about Raspbian first because of it's position in the education and maker markets. It would be nice to get such a thing into official repos but I've obviously never been a maintainer so that's a stretch goal. Thanks for your guidance so far, including the CFLAGS edit to the makefile. What's the state of openSUSE on the Pi or similiar hardware like the Tinker Board? The Rockchip 3288 on the TB absolutely flys for such a low power device.  Wish I knew enough to work at a low level in that space. Anyhow, I'm Rider8 on the raylib forum if you feel like opening a line.

jubalh:
@RDR8 you are in for a lot of work if you want to do that ;)

If you would become a packager for Debian, and package raylib, your package would automatically end up in Ubuntu and Raspbian afaik.

Not really sure about Raspbian but certainly Ubuntu regularly pulls from Debian.
Also once written for Debian the packaging files are the same for Raspian, just built for another architeture.

openSUSE works great on ARM boards. Several people work in this area.
More info can be found on the [wiki](https://en.opensuse.org/Portal:ARM).
Afaik it is mostly used by professionals though not so much for education. I dont know much about boards or that are in general to give a qualified answer.

Info about [openSUSE on the rpi3](https://en.opensuse.org/HCL:Raspberry_Pi3) can be seen also in the wiki.

What I know however, is that we were the first ones to have AARCH64 support ;)
Meaning the first ones who were able to use the whole power of the rpi3.

See [here](https://www.raspberrypi.org/blog/suse-linux-enterprise-server-for-raspberry-pi/) and [here](https://www.suse.com/communities/blog/opensuse-raspberry-pi-3-zero-functional-system-easy-steps/).

Anyways AFAIK openSUSE on rpi is for people who know what they do, who want to create something of their own, and maybe not as easy to use as raspbian. But I am not sure on this one.

jubalh:
@RDR8 By the way, upon your suggesting, I now also enabled the ARMv7 and AARCH64 builds for raylib.
It can be seen at the [OBS devel project](https://build.opensuse.org/package/show/devel:libraries:c_c++/raylib)

RDR8:
So I guess the thing to do is make a deb then follow the submission procedures.  I don't know about PPAs other than as an intermediate step. Worst case would be to host a deb somewhere.  
Is is possible to use or convert your RPM to Fedora? Maybe not in your interest, I'm just curious.  
I'm intrigued by the progress of the ARM platform. Also the price of entry is low.  
  
My actual goal is to investigate and if possible build a simple network of point-of-sale kiosks and technical stations supporting a small business such as a bike shop, or other scenario where goods, services, and money need to be tracked. Square is ok but my customers don't want to be on that list. I've been using Google Apps for basic record keeping but the scripting performance on a typical Chromebook is unsatisfactory. One thing led to another and both of our Chromebooks are running Linux. Much better. I;ve considered something like Nextcloud and Collabora Online but that's technology that strays from my investigation of C and traditional multi-user systems. SUSE offers services in this vein? I can't find the system I was looking at. It seemed targeted at much larger enterprises than mine. In short, I want to hook up some Pis with very simple interfaces for a few specific actions on a local data store. Oh and there's also a bot on wheels that responds to "Hey, Bot!", with "Yeah, Boss?" I want to make sure I don't run out of things to think about. (^;

RDR8:
I'd heard of OBS but never seen it. Looks awesome. I need to get openSUSE for my board and test it out. I was sticking with Raspbian for the video drivers though I understand progress is being made on an open source driver. Thanks!

RDR8:
SLE Point of Service is what I was thinking of.

jubalh:
> So I guess the thing to do is make a deb then follow the submission procedures. I don't know about PPAs other than as an intermediate step. Worst case would be to host a deb somewhere.
Is is possible to use or convert your RPM to Fedora? Maybe not in your interest, I'm just curious.
Yes OBS, can also build Fedora packages and even othe distributions like Arch and Debian.

The latter need extra files though (PKGBUILD and dsc things, meaning their native format).

People can create their own private repos on OBS.
For example there is an editor called [focuswriter](https://gottcode.org/focuswriter/), the author of that chose OBS to build his packages which he then ships to different distros.
If you click download on his site you get refered to the built [packages on OBS](https://software.opensuse.org/download/package?package=focuswriter&amp;project=home:gottcode).
It offers one-click install for openSUSE and instructions for other distros. (adding repo and installing with native distro packaging tool).
In gottcode's [home project on OBS](https://build.opensuse.org/package/show/home:gottcode/focuswriter) you can see the spec file he uses to build for openSUSE and Fedora, and the dsc for Debian.

OBS is good for building packages for many distros. However these packages will be in users home repos, which means people can add the repo and then install.

For raylib I am pushing it to offcial repos so that users can just install it (on openSUSE).

As a site note: if you didnt check out openSUSE yet, it has even more cool features:
There is [SUSE Studio](https://susestudio.com/) whre you can create 'your own distro' with a few clicks.
Its based on openSUSE, you can define which repos to add by default, which packges to install, which not to install, themes etc.
You can create ISOs, virtual machines etc...

Also openSUSE has two distros: Tumbleweed which is rolling release (in my opinion the fastest and most stable rolling release out there), and Leap which is stable/versioned.

We created an use [openQA](http://open.qa/) to test the distribution/programs/packages. Which is the reason for detecting errors often before shipping the new snapshot.

All in all its great stuff, but it needs time to learn all about these things :-)
One step at a time.

RDR8:
Just browsing the Debian packaging splash page, I see what you mean about a lot of work. Yet another set of tools to learn. Well, that's why I'm here. Since it could be an effective way to help grow raylib's user base, it's worth further investigation.   
  
A couple of initial questions to answer are:  
Which version of raylib to package? Just package the initial Release and update every time a new version comes out or follow the master branch on git and push updates more frequently?  
Which release of Debian to target? Stable, testing, or stretch?  
  
SUSE has been recommended to me several times and is interesting in terms of my main project. This seems like a good time to begin exploring it.  
  
Any thoughts on these matters are welcomed and much appreciated. Thanks for your time!

jubalh:
You will only be allowed to ship released things, no git things into the official distribution.
Stable, testing etc. Think about the name :-)

All new packages arrive in unstable, they will there for X days then move into testing, then they will be there until the next release happens (maybe 5 years). Then its in stable.
There is no other way :-)

jubalh:
People will need stable releases, which will only get security updates, not changing any functionality. Until the next release, then the whole process start again.

Thats why its also important to use library versioning, and dont break the ABI without increasing major version number. The library might even fall out of the repo if it violates that, or just doesnt get updated (afaik).

RDR8:
Awesome! Thanks for providing context. It's tremendously helpful.

jubalh:
Different distributions have different guidelines regarding these things, its important to keep that in mind.
And if one settles for a distribution one should not just look at what standard desktop environment and wallpaper they use but what are their rules for packages, does that rule suite your needs, how good is updates and security handled, what tools does the distribution offer, is the package management in good order so i can update for many years without reinstallation etc. There is a lot to take into account.

It also needs some experimentation with different concepts to know the pros and cons, and what one really wants sometimes one only knows when one realizes what one doesnt want hehe.

Take your time, only go as deep as you feel comfortable.

RDR8:
At a minimum, I'll try not to break things publicly. (^;

PR #301, opened 2017-06-11
Add define to have CLOCK_MONOTONIC work in c99

jubalh:
If we compile with c99 without gnu extensions (gnu99) we need this
define, to have CLOCK_MONOTONIC and similar macros available

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -109,6 +109,10 @@
     #include "external/gif.h"   // Support GIF recording
 #endif
 
+#if defined(__linux__) || defined(PLATFORM_WEB)
+    #define _POSIX_C_SOURCE 199309L // Required for CLOCK_MONOTONIC if compiled with c99 without gnu ext.
+#endif
+
 #include <stdio.h>          // Standard input / output lib
 #include <stdlib.h>         // Required for: malloc(), free(), rand(), atexit()
 #include <stdint.h>         // Required for: typedef unsigned long long int uint64_t, used by hi-res timer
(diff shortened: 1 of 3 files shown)


RDR8:
This looks a lot like the problem we dealt with in src/Makefile by defining -D_DEFAULT_SOURCE in CFLAGS. It pulls all the old POSIX as well as _BSD_SOURCE with one define. Since Michael isn't using the Makefile, he's not getting _DEFAULT_SOURCE which may be more appropriate than specific _POSIX_C_SOURCE defines. I ran into this with physac and dr_flac as included with raylib. As always, correct me if necessary. Reference forthcoming.

raysan5:
@RDR8, you're right, it's related... same issue again...

Personally I prefer to find a code solution instead of additional defines passed in Makefile...

RDR8:
So... Moving that to c files as @jubalh has done is the way?

raysan5:
For the moment it seems a good solution to me... but I didn't evaluate it carefully...

RDR8:
@jubalh, would you test this by using _DEFAULT_SOURCE instead? It will be some hours before I can set up for it.

RDR8:
dr_flac carries it's define of _BSD_SOURCE with it, but audio examples were still problematic until we put that or _DEFAULT_SOURCE into the example build commands.  
  
physac was a bit different, failing at raylib build time, while looking for bsd function definitions in time.h. IIRC

RDR8:
So, such a define should be any file which may need it? What about double declarations? Use ifndefs?  
The issue begins with declaring std=c99. That turns off a lot of stuff that we want.

PR #303, opened 2017-06-11
Meson: use gnu99

jubalh:
We use variable init in for loops, so need c99.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,7 @@
 project('raylib', 'c', version:	'1.7.0',
 		license: 'zlib',
-		meson_version: '>= 0.39.1')
+		meson_version: '>= 0.39.1',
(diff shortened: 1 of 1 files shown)


RDR8:
I know I keep harping on it but -std=c99 -D_DEFAULT_SOURCE builds raylib and examples cleanly with gcc and clang. If it's not related or doesn't suit your purposes, I'll shut up.

jubalh:
I have to look more into it, I didnt know -D_DEFAULT_SOURCE before you mentioned it.
Using gnu99 looks cleaner to me, but I have to read up on this :)

RDR8:
`man feature_test_macros`

RDR8:
I don't know much about the differences of gnu99 and c99, or gcc and clang. I just wanted to see if it would run. (^;

PR #309, opened 2017-06-28
Add some image generation functions

nounoursheureux:
Namely:
- `GenImageHorizontalV`
screen0

- `GenImageHorizontalH`
screen1

- `GenImageChecked`
screen2

- `GenImageWhiteNoise`
screen3

- `GenImageCellular`
screen4

The gradient implementation may be a bit naive, for example it doesn't do any gamma correction.

I'll look into implementing `ImageGenSpot` and `ImageGenPerlin` but I don't know if I'll succeed.

Also I think I could implement an `ImageSaveAs` function so that the users can save their procedurally generated images (looking at the `TakeScreenshot` implementation this shouldn't be too hard)

### DIFF
--- a/examples/textures/textures_image_generation.c
+++ b/examples/textures/textures_image_generation.c
@@ -1,0 +1,57 @@
+/*******************************************************************************************
+*
+*   raylib [textures] example - Procedural images generation
+*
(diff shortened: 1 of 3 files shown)


raysan5:
Wow! Great addition! And a very clean coding, I like it! :)

PR #310, opened 2017-06-28
Add the SaveImageAs function

nounoursheureux:
Signature: `SaveImageAs(Image image, const char *filename);`

I think this would be useful for the users who generate procedural textures, it allows them to save them (for example to upload or share them)
Right now the workaround is to display the image on the entire window and to take a screenshot, but I think a dedicated function makes sense


raysan5:
Good addition @nounoursheureux!

Please, do you mind changing parameters order to: 
```c
void SaveImageAs(const char *fileName, Image image);
```
I think is more consistent with other similar functions.

nounoursheureux:
Of course :+1:

raysan5:
Great! Thank you very much!

PR #311, opened 2017-06-28
Add more image generation functions: radial gradient and perlin noise

nounoursheureux:
Signatures:
- `GenImageRadialGradient(int width, int height, Color inner, Color outer);`
radial

- `GenImagePerlinNoise(int width, int height, float scale);`
perlin


raysan5:
Wow! So fast! Thank you for the addition!

Just an idea, for future improvements, it would be great to have a radial gradient density parameter:
image

PR #312, opened 2017-06-28
Add a density parameter to GenImageRadialGradient

nounoursheureux:
And here are some results:
- Density = 0.0
density0
- Density = 0.3
density3
- Density = 0.5
density5
- Density = 0.8
density8
- Density = 1.0
density1

I don't know if this is really what you expected, tell me 😄


raysan5:
That's perfect! 👍

PR #313, opened 2017-06-28
Removed an unwanted file and fixed an example

nounoursheureux:
I accidentally pushed a binary to the repo so I have to delete it
I also forgot to unload the images in textures_image_generation.c

Maybe I tried to go a little too fast 😅


raysan5:
Hey! No worries! Thanks again for all your effort!

PR #315, opened 2017-07-01
Fix for OBJ loading

joeld42:
When there is a newline at the end of an .OBJ file, this can cause a bogus triangle to be added to the mesh. In LoadOBJ(), the "dataType" does not get reset, and the result from fscanf is not checked, so at the end of the file the last triangle is sometime counted twice, but since there isn't a corresponding entry in the vertex data it can create a bogus triangle. Usually the verts are all zero and you don't see it, but this has been causing occasional glitchy rendering for me.

Simply resetting the dataType for each line fixes this, as it won't process the last entry twice.


raysan5:
Hi @joeld42! Good catch! Thanks for correcting this issue! 👍

PR #327, opened 2017-07-15
Implement PBR support for raylib

victorfisac:
- Some last changes of @raysan5 was reverted. For example: shader continues using MaterialProperty struct and useSampler. You should see with your computer if you can run models_material_pbr and models_skybox as expected in the screenshots I recently added.

- If you cannot run it, you can delete again MaterialProperty and write an alternative code by your own.

- Created new elements in location enumerable to avoid using GetShaderValue during rlDrawMesh().

- Implemented lighting (not using memory allocation) in example.

Note: the problem wasn't calculing BRDF or Prefilter textures. The mMatrix (model transform) wasn't being assigned and normal calculations had errors (black shading)...


raysan5:
Amazing work @victorfisac! Just testing it right now!

Issue #328, opened 2017-07-17
[build] Shared library on OS X


define-private-public:
I'm working on some Nim bindings at the moment and I'm trying to test they work on OS X.  I'm trying to work with raylib as a shared library.  So When I clone the source and did `make SHARED_RAYLIB=YES PLATFORM=PLATFORM_DESKTOP` there were no `.so` files in the `release/osx` folder.

I noticed that around line `320` of the `src/Makefile` that there was no `ifeq` clause for `OSX`.  So I added one in (by copy and pasting the one for `LINUX`).  It did get a bit farther, but then this error arose:

```
mac:src me$ make SHARED_RAYLIB=YES PLATFORM=PLATFORM_DESKTOP
gcc -c audio.c -O1 -Wall -std=c99 -D_DEFAULT_SOURCE -fgnu89-inline -Wno-missing-braces -fPIC -I. -Iexternal -Iexternal/openal_soft/include -Iexternal/glfw3/include -DPLATFORM_DESKTOP -DBUILDING_DLL -DAL_LIBTYPE_STATIC
audio.c:878:25: warning: unused variable 'numSamplesOgg' [-Wunused-variable]
                    int numSamplesOgg = stb_vorbis_get_samples_short_interleaved(music->ctxOgg, m...
                        ^
1 warning generated.
gcc -c core.c -O1 -Wall -std=c99 -D_DEFAULT_SOURCE -fgnu89-inline -Wno-missing-braces -fPIC -I. -Iexternal -Iexternal/openal_soft/include -Iexternal/glfw3/include -DPLATFORM_DESKTOP -DBUILDING_DLL
gcc -c models.c -O1 -Wall -std=c99 -D_DEFAULT_SOURCE -fgnu89-inline -Wno-missing-braces -fPIC -I. -Iexternal -Iexternal/openal_soft/include -Iexternal/glfw3/include -DPLATFORM_DESKTOP -DBUILDING_DLL
gcc -c rlgl.c -O1 -Wall -std=c99 -D_DEFAULT_SOURCE -fgnu89-inline -Wno-missing-braces -fPIC -I. -Iexternal -Iexternal/openal_soft/include -Iexternal/glfw3/include -DGRAPHICS_API_OPENGL_33
[...]

PR #329, opened 2017-07-19
Able to make a .dylib for OS X

define-private-public:
This should make the proper `.dylib` file, and set it so that whatever program is trying to be run, the .dylib can be in the same folder.

1. There should be a note of that `same folder` thing somewhere in the docs.
2. I haven't tested how this effects of building the examples.

Discussion should go back to ticket #328 to resolves those two things

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -145,6 +145,14 @@ endif
 # default gcc compiler
 CC = gcc
 
+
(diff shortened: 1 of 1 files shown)


raysan5:
Thank you very much for the update!

PR #333, opened 2017-07-21
Update Makefile in `examples/`

define-private-public:
We need to grab the OS name before determining what the path should be.


raysan5:
Sure! Thank you very much! :)

PR #348, opened 2017-08-27
CMake based build system.

define-private-public:
Some people might find this handy.

Not all OSes and platforms are ready, but a good chunk of it is good to go.


raysan5:
Great work! Thank you very much for your time and your effort! :)

PR #359, opened 2017-09-25
Increment version in meson build file

jubalh:
Increment raylib version and so lib version number.

Actually since this is develop branch I guess incrementing already (so we dont forget) cannot hurt.


raysan5:
Hi @jubalh! Thanks for the update! :D

PR #363, opened 2017-10-05
Remove unportable bool guard check

a3f:


raysan5:
Hi @a3f, I try to avoid any include in `raylib.h` so I used the enum and defines, it should be the same as using `#include <stdbool.h>`...

What conflicts did you find with that approach?

a3f:
Would these old compilers be even able to use raylib? The header currently requires either C99 or C++ and `<stdbool.h>` is a mandatory C99 header, present even in free-standing (barebones) environments.

I noticed it because it broke [this build](http://www.cpantesters.org/cpan/report/77646436-a929-11e7-b1db-c0b5893dd193):

```
# Failed test 'xs'
# at t/init.t line 15.
#   ExtUtils::CBuilder->compile failed
#     error building testalienT4zpy/test.o from 'testalienT4zpy/test.c' at /opt/perl-5.12.5/lib/site_perl/5.12.5/ExtUtils/CBuilder/Base.pm line 183.
#     cc -I/opt/perl-5.12.5/lib/5.12.5/x86_64-linux/CORE -fPIC -I/tmpfs/.cpan-build-cpansand/2017100412/Alien-raylib-0.003-13/blib/lib/auto/share/dist/Alien-raylib/include -c -fno-strict-aliasing -pipe -fstack-protector -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -O2 -o testalienT4zpy/test.o testalienT4zpy/test.c
#     In file included from /opt/perl-5.12.5/lib/5.12.5/x86_64-linux/CORE/perl.h:2429:0,
#                      from testalienT4zpy/test.xs:3:
#     /opt/perl-5.12.5/lib/5.12.5/x86_64-linux/CORE/handy.h:108:16: error: expected Ã¢ÂÂ;Ã¢ÂÂ, identifier or Ã¢ÂÂ(Ã¢ÂÂ before Ã¢ÂÂcharÃ¢ÂÂ
#      #  define bool char
#                     ^
#     /tmpfs/.cpan-build-cpansand/2017100412/Alien-raylib-0.003-13/blib/lib/auto/share/dist/Alien-raylib/include/raylib.h:302:42: note: in expansion of macro Ã¢ÂÂboolÃ¢ÂÂ
#                  typedef enum { false, true } bool;
#                                               ^
#     /opt/perl-5.12.5/lib/5.12.5/x86_64-linux/CORE/handy.h:108:16: warning: useless type name in empty declaration
#      #  define bool char
#                     ^
#     /tmpfs/.cpan-build-cpansand/2017100412/Alien-raylib-0.003-13/blib/lib/auto/share/dist/Alien-raylib/include/raylib.h:302:42: note: in expansion of macro Ã¢ÂÂboolÃ¢ÂÂ
#                  typedef enum { false, true } bool;
#                                               ^
```

I maintain a small raylib wrapper and distribution for Perl ([Alien::raylib ](https://github.com/athreef/Alien-raylib)and [Graphics::Raylib](url)) and I am looking into having them smoke tested by [CPAN testers](http://matrix.cpantesters.org/?dist=Alien-raylib+0.003).

I don't know if that smoke tester can do C99, but it doesn't hurt to eliminate possible causes of problems.

raysan5:
As per the output, it seems the issue is not related to the current implementation of `enum bool` of raylib, actually it works ok in all supported and tested platforms (Windows, Linux, OSX, Android, RPI, HTML5).

Issue seems related to `#define bool char` definition on `handy.h`. Please, could you try this way on `raylib.h`:
```c
#ifndef __cplusplus
// Boolean type
    #if !defined(_STDBOOL_H) || !defined(__STDBOOL_H)   // CLang uses second form
        typedef enum { false, true } bool;
        #define HAS_BOOL
    #endif
#endif
```
As said, I try to avoid any  `#include` in `raylib.h` header but if there is no alternative, I can consider adding `stdbool.h`.

a3f:
I don't think that raylib should define general identifiers like `HAS_BOOL`. There's nothing wrong with including standard headers like `<stddef.h>` or `<stdbool.h>` if we already implicitly require them being available. That's what they're there for after all.

An alternative could be:
```
#ifndef __cplusplus
    #ifndef bool
        typedef enum { false, true } bool;
    #endif
#endif
```

`<stdbool.h>`'s bool is always a macro (ISO C11 §7.18¶2), so that would be the correct way to check if the header has been included. I still think just including the standard header is the better approach.

raysan5:
Just decided to implement the proposed alternative in commit https://github.com/raysan5/raylib/commit/399d4b5f4618ee9147c379364ad69b8c1fbb3002

I'll consider the include in next version... In any case, thanks for the improvement.

a3f:
Broke build for me again, on [this FreeBSD Perl 5.14](http://www.cpantesters.org/cpan/report/bdd1ffc0-0c96-11e8-a1cf-bb670eaac09d) `bool` is defined, but with `TRUE` and `FALSE`, not `true` and `false`…

I can easily work around that, but it would be nice if raylib could do the right thing and just include the header ^^.

raysan5:
> I can easily work around that, but it would be nice if raylib could do the right thing and just include the header ^^.

agree, go for it. :)

a3f:
_it_ = ?

raysan5:
Sorry, I mean, let's `#include <stdbool.h>`

a3f:
Thinking about it again, the _headers_ are C89 actually, so you could build the library with a c99 compiler, but write your application in a C89 compiler, so a simple

```
#ifndef __cplusplus
     #include <stdbool.h>
#endif
```

wouldn't be sufficient. I will think of something.

PR #382, opened 2017-11-05
Have SetShaderValue(i) take a const pointer

nounoursheureux:
Just a tiny little change: the `glUniform*` functions take const pointers, so I don't think there is any reason to disallow const pointers from being used in `SetShaderValue` :)


raysan5:
Thanks for the review @nounoursheureux!

PR #384, opened 2017-11-09
Store the default shaders

### LINKED ISSUE #381: [rlgl] Load one custom shader and one default shader in LoadShader()
The title is not very clear, but what I'd like to be able to do is to call `LoadShader` with one NULL pointer and one valid path, and have it load the default shader for the NULL argument and load the shader at the given path.
An example: say I want to write a custom fragment shader but keep the default vertex shader, I could just write `LoadShader(NULL, "custom.fs")` and have it return a shader program that has the default vertex shader and my custom fragment shader.
Right now AFAIK I have to copy the default vertex shader and load it myself, it isn't a major annoyance but I think this can be improved :)
If you are okay with this proposition I think I can implement it myself and submit a PR.

nounoursheureux:
I did as you suggested: I added a `LoadShaderPartial` function and changed the `LoadShaderProgram` function to take already compiled shaders. Which means we only have to compile the default shaders once now !

Fix #381

### DIFF
--- a/examples/shaders/shaders_postprocessing.c
+++ b/examples/shaders/shaders_postprocessing.c
@@ -20,10 +20,8 @@
 
 #if defined(PLATFORM_DESKTOP)
     #define GLSL_VERSION            330
-    #define DEFAULT_VERTEX_SHADER   "resources/shaders/glsl330/base.vs"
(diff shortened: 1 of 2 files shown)


raysan5:
Great! Thanks for the update @nounoursheureux! 

I'll probably rename that function to `CompileShader()`, it's more descriptive.

PR #392, opened 2017-11-22
 Fix macOS build of new rglfw.c approach

### LINKED ISSUE #391: [build] Travis CI system fails on OSX compilation
After adding GLFW3 library as an additional module ([rglfw](https://github.com/raysan5/raylib/blob/develop/src/rglfw.c)), Travis CI OSX compilation fails, maybe due to the inclusion of .m files and the requirement of clang compiler...

a3f:
There have been two problems:

* GLFW itself was compiled with the definitions for compiling
_against_ GLFW (Fixed by removing requirement for external glfw)

* rglfw.c was being compiled as C code, although it includes
Objective C files.

This _might_ break the Windows build, needs to be checked.

Fixes #391, but as noted I'd prefer though a separate source directory
and build script for GLFW.

The prior commit fixes a Travis CI compile error when compiling for Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,8 @@ before_install:
                               libopenal-dev
                               libxcursor-dev libxinerama-dev
                               mesa-common-dev libx11-dev libxrandr-dev libxi-dev xorg-dev libgl1-mesa-dev libglu1-mesa-dev libglew-dev;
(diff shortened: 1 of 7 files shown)


raysan5:
Many thanks for the update!

PR #395, opened 2017-11-24
Generate and install pkg-config pc file and allow static/shared builds in one go

a3f:
After installation, compiling new user programs is possible with

    $ cc game.c `pkg-config --static --libs --cflags raylib`
or

    $ cc game.c `pkg-config --libs --cflags raylib`
depending on configuration

Also adds following configuration options:
- `WITH_PIC` "Compile static library as position-independent code"
- `STATIC_RAYLIB` "Build raylib as a static library"
- `MACOS_FATLIB`  "Build fat library for both i386 and x86_64 on macOS"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ env:
   global:
     - VERBOSE=1
   matrix:
-    - CFLAGS=-m64
-    - CFLAGS=-m32
-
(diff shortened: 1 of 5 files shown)


raysan5:
Thanks! Great addition! :)

Issue #385, opened 2017-11-12
[audio] Replace OpenAL Soft audio library by mini_al


raysan5:
Since the beginning of raylib I've been looking for a lightweight alternative library to [OpenAL Soft](https://github.com/kcat/openal-soft) (https://github.com/raysan5/raylib/issues/52) and finally, there seems to be a solid candidate for that: [mini_al](https://github.com/dr-soft/mini_al) by @dr-soft and proposed by @gen2brain (https://github.com/raysan5/raylib/issues/52#issuecomment-340394183).

Don't misundertood, **OpenAL Soft is great**, easy-to-use, lots of features... but it's also a big library, sometimes difficult to build, with not-really-zlib-compatible license (https://github.com/raysan5/raylib/issues/372#issuecomment-340219354)... and adds a big dependency to raylib. 

**One of the final goals for raylib is avoiding any external dependency, being a self contained library that you can download, compile and run in any platform.**

It seems **mini_al** fits perfectly on what I was looking for, actually, library is in active development and new features and platforms are being added. Additionally, @mackron started working on [mini_al support for raylib](https://github.com/mackron/raylib/commit/75433a670e0880c4d23d5178b073836de3628547) and also, @gen2brain started porting [mini-al to Go language](https://github.com/gen2brain/malgo), perfect for [raylib-go](https://github.com/gen2brain/raylib-go) binding. Also love [mini_al logo](https://camo.githubusercontent.com/a6cb4cd4f8848986914c8756dfab696017f695be/687474703a2f2f647265642e696f2f696d672f6d696e69616c5f776964652e706e67). :P

Many thanks to @mackron and @gen2brain for their help on this project!

raysan5:
More issue discussion on mini-al github: https://github.com/dr-soft/mini_al/issues/3

raysan5:
Replacement is almost complete, being tested in multiple platforms, but it is a really big library change so, thinking about some transition mechanism. 

Maybe supporting OpenAL Soft and mini_al for some time? Or just move fordward and remove OpenAL Soft completely on next raylib iteration? Thinking about it...

minecraftlog21:
I'd say just make the transition completely, whats the point of having two different sound APIs if anyone needs OpenAl cant they just link in them selves.

raysan5:
[...]

PR #398, opened 2017-11-25
Build examples and games on Travis CI

a3f:
They were disabled because they failed to build,
but this patch set fixes the build on Linux and macOS.

This doesn't apply to the AppVeyor build on Windows yet;
it currently fails at linking with OpenAL.


raysan5:
Thanks for the update! :)

About OpenAL, just note that it will probably be removed on next raylib version, replacing it by mini_al (https://github.com/raysan5/raylib/issues/385, https://github.com/dr-soft/mini_al/issues/3)

PR #413, opened 2017-12-05
mini_al integration

mackron:
This adds mini_al as a backend for audio. I've left the OpenAL implementation in place, and it can be toggled with USE_MINI_AL:
```
#define USE_MINI_AL 1 // Set to 1 to use mini_al; 0 to use OpenAL.
```
A few notes about this:
- I've tried to use a consistent coding style, but I know I have used my own personal style in places, so you may want to clean that up.
- The implementation of mini_al is defined in `external/mini_al.c`. It needs to a separate transaction unit due to naming conflicts with the Win32 API.
- In my testing, I found the first call to `EndFrame()` is slow, and thus causes a pause in the audio at the beginning of playback. Moving `UpdateMusicStream()`, etc to after the drawing section should fix this, I think.
- With OpenAL, the maximum volume for sounds is clamped to 1, but this restriction does not apply to mini_al.
- I changed the implementation of WaveFormat() to use mini_al's built-in format conversion routines. This enables a greater number of formats.
- I've done some testing, but you need to do your own testing. In particular, the part where audio buffers are tracked (`TrackAudioBuffer()` and `UntrackAudioBuffer()`) has not been fully tested. I've done no testing whatsoever on Mac.
- I've changed `void SetMusicLoopCount(Music music, float count)` to `void SetMusicLoopCount(Music music, int count)` (changed `count` from `float` to `int`).
- I've not updated any make files or anything.

Some things I've noticed which I haven't fixed in this pull request, but you should keep in mind:
- There's a bug with your looping logic for music streams with the OpenAL backend. I've not fixed this because I don't know if you'll be keeping the OpenAL backend in the long run.
- I don't mean this to sound like I'm promoting my own library, but your .wav loader isn't the best. Using dr_wav will enable support for a much wider array of .wav files.
- If you are planing on continuing to support FLAC, I highly recommend you upgrade dr_flac - the one you're currently using is way out of date.
- There is a section of code in UpdateMusicStream() that looks like the following. I think the intent is to have audio pause while the window is minimized, and then resumed when it is restored? I'm not sure this code works the way you're intending, but I've left it in place for backwards compatibility. You may want to compare the mini_al and OpenAL backends to see if there's any difference.
```
// NOTE: In case window is minimized, music stream is stopped,
// just make sure to play again on window restore
if (IsMusicPlaying(music)) PlayMusicStream(music);
```
See [issue #3 in mini_al](https://github.com/dr-soft/mini_al/issues/3) for prior discussion on this.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1426,6 +1426,13 @@ void UpdateMusicStream(Music music)
             music->loopCount--;        // Decrease loop count
             PlayMusicStream(music);    // Play again
         }
+        else
+        {
+            if (music->loopCount == -1)
+            {
+                PlayMusicStream(music);
+            }
+        }
     }
     else
     {
@@ -1506,6 +1513,13 @@ void UpdateMusicStream(Music music)
                 music->loopCount--;        // Decrease loop count
                 PlayMusicStream(music);    // Play again
             }
+            else
+            {
+                if (music->loopCount == -1)
+                {
+                    PlayMusicStream(music);
+                }
+            }
         }
         else
         {
@@ -1537,13 +1551,21 @@ bool IsMusicPlaying(Music music)
 // Set volume for music
 void SetMusicVolume(Music music, float volume)
 {
+#if USE_MINI_AL
+    SetAudioStreamVolume(music->stream, volume);
+#else
     alSourcef(music->stream.source, AL_GAIN, volume);
+#endif
 }
 
 // Set pitch for music
 void SetMusicPitch(Music music, float pitch)
 {
+#if USE_MINI_AL
+    SetAudioStreamPitch(music->stream, pitch);
+#else
     alSourcef(music->stream.source, AL_PITCH, pitch);
+#endif
 }
 
 // Set music loop count (loop repeats)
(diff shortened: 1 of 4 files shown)


raysan5:
Hi @mackron! 

Many thanks for all the hard work invested in this implementation and your amazing library!

 > I've tried to use a consistent coding style, but I know I have used my own personal style in places, so you may want to clean that up.

I'll review that.

 > I changed the implementation of `WaveFormat()` to use mini_al's built-in format conversion routines. This enables a greater number of formats.

That's great! Also noticed a question about `sampleCount` vs `frameCount`, what is the difference? For me `sampleCount` is the number of samples, independent of the bit depth of each sample (8-16-24-32 bit).

 > I've done some testing, but you need to do your own testing.

Yes, I'll do.

 > I've not updated any make files or anything.

Already working on that!

 > There's a bug with your looping logic for music streams with the OpenAL backend. I've not fixed this because I don't know if you'll be keeping the OpenAL backend in the long run.

Oh, what is the issue? Just would like to know for learning... I'll keep the OpenAL Soft backend for a while but my plan is to completely remove it in a near future.

 > I don't mean this to sound like I'm promoting my own library, but your .wav loader isn't the best. Using dr_wav will enable support for a much wider array of .wav files.

Agree, my wav loading implementation didn't pretend to be very professional, actually it was a learning exercise for myself. I'll probably switch to dr_wav.

 > If you are planing on continuing to support FLAC, I highly recommend you upgrade dr_flac - the one you're currently using is way out of date.

Actually, I need to update all external libraries!

 > There is a section of code in UpdateMusicStream() that looks like the following. I think the intent is to have audio pause while the window is minimized, and then resumed when it is restored? I'm not sure this code works the way you're intending, but I've left it in place for backwards compatibility. You may want to compare the mini_al and OpenAL backends to see if there's any difference.

Yes, that was the objective. It seemed to work as intended. I'll take a closer look...

I'm integrating it and I'll start reviewing all the commented points and build system configurations to switch to `mini_al` library.

Again, many thanks for all your hard work on this! :D

mackron:
Apple's documentation has a good definition of the difference between samples and frames: https://developer.apple.com/library/content/documentation/MusicAudio/Conceptual/CoreAudioOverview/WhatisCoreAudio/WhatisCoreAudio.html. See the "A Little About Digital Audio and Linear PCM" section. Basically, a frame is 1 sample per channel. So for stereo audio, a frame is made up of 2 samples (a left sample and a right sample).

I never looked into the exact cause of the OpenAL music looping bug, but to reproduce, just enable the OpenAL backend and set the loop count to 1 in your music example and note how it loops forever.

raysan5:
Thanks for the link! Really educative! :)

 > I never looked into the exact cause of the OpenAL music looping bug, but to reproduce, just enable the OpenAL backend and set the loop count to 1 in your music example and note how it loops forever.

Oh! I see... didn't notice that issue...

PR #421, opened 2017-12-14
Build shared libs, games and examples on CI

a3f:
Now with external OpenAL and GLFW dependencies removed, we don't have to worry about installing them in CI.
Shared libraries are now always built along with static libs. Games and examples are built everwhere except for Visual Studio, because Physac needs pthreads, which VS doesn't provide.


raysan5:
Great! Thank you very much for the review! :)

PR #422, opened 2017-12-19
Update CONTRIBUTING.md

RDR8:
Thanks for the nice new introduction.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Use your best judgement, and feel free to propose changes to this document in a 
  - raylib is SIMPLE and EASY-TO-USE, I tried to keep it compact with a small set of functions, if a function is too complex or is not clearly useful, better not to include it.
  - raylib is open source and free; educators and institutions can use this tool to TEACH videogames programming completely by free.
  - raylib is collaborative; contribution of tutorials / code-examples / bugs-solving / code-comments are highly appreciated.
- - raylib's license (and it's external libs respective licenses) allow using it for commercial products.
+ - raylib's license (and its external libs respective licenses) allow using it for commercial products.
 
 ### Some interesting reads to start with
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ Thank you very much for your time! :)
 ----
 
 Here is a list of raylib contributors, these people have invested part of their time 
-contributing (in some way or another) to make raylib project better. Huge thanks to all of them!
+contributing (in some way or another) to make the raylib project better. Huge thanks to all of them!
 
  - [Zopokx](https://github.com/Zopokx) for testing the web.
  - [Elendow](http://www.elendow.com) for testing and helping on web development.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -8,10 +8,10 @@ Use your best judgement, and feel free to propose changes to this document in a
 ### raylib philosophy
 
  - raylib is a tool to LEARN videogames programming, every single function in raylib should be a tutorial on itself.
- - raylib is SIMPLE and EASY-TO-USE, I tried to keep it compact with a small set of functions, if a function is too complex or has not a clear usefulness, better not to include it.
+ - raylib is SIMPLE and EASY-TO-USE, I tried to keep it compact with a small set of functions, if a function is too complex or is not clearly useful, better not to include it.
  - raylib is open source and free; educators and institutions can use this tool to TEACH videogames programming completely by free.
  - raylib is collaborative; contribution of tutorials / code-examples / bugs-solving / code-comments are highly appreciated.
- - raylib's license (and its external libs respective licenses) allow using it for commercial products.
```
a3f: `its` is correct here.

CONTRIBUTING.md
```diff
@@ -23,6 +23,18 @@ Use your best judgement, and feel free to propose changes to this document in a
 [raylib Wiki](https://github.com/raysan5/raylib/wiki) contains some information about the library and is open to anyone for edit. 
 Feel free to review it if required, just take care not to break anything.
 
+### raylib C coding conventions
+
+Despite being written in C, raylib does not follow the standard Hungarian notation for C,
```
a3f: If there's a standard notation for C, it would be what the standard library and by extension Unix uses, which is `snake_case`. Hungarian notation is more of a DOS/Windows thing, I believe.

a3f: Oh, never mind me. You just shifted that text around :D

CONTRIBUTING.md
```diff
@@ -8,10 +8,10 @@ Use your best judgement, and feel free to propose changes to this document in a
 ### raylib philosophy
 
  - raylib is a tool to LEARN videogames programming, every single function in raylib should be a tutorial on itself.
- - raylib is SIMPLE and EASY-TO-USE, I tried to keep it compact with a small set of functions, if a function is too complex or has not a clear usefulness, better not to include it.
+ - raylib is SIMPLE and EASY-TO-USE, I tried to keep it compact with a small set of functions, if a function is too complex or is not clearly useful, better not to include it.
  - raylib is open source and free; educators and institutions can use this tool to TEACH videogames programming completely by free.
  - raylib is collaborative; contribution of tutorials / code-examples / bugs-solving / code-comments are highly appreciated.
- - raylib's license (and its external libs respective licenses) allow using it for commercial products.
+ - raylib's license (and it's external libs respective licenses) allow using it for commercial products.
```
a3f: `its` is correct here.

RDR8: Debatable. Consider the possessive of raylib just prior... Superfluous in any case.

a3f:
Thanks.

CONTRIBUTING.md
RDR8: raysan is pretty much a Windows guy I think, so it makes sense that he would say it that way. Good info though, thanks!

PR #428, opened 2018-01-13
Add mouse input and revert keyboard to callback model

SamNChiet:
Mouse input is implemented, with all bells-and-whistles. This includes cursor locking and scroll wheel support.

Keyboard input is reverted to a callback model to better reflect the existing architecture in "core.c"


raysan5:
Thanks @SamNChiet! Great adddition! :D

PR #432, opened 2018-01-17
fixed cmakefile

questor:
small bugfix for cmakefile in examples-folder


a3f:
Tested this locally on macOS and it builds ok. Thanks for your contribution!

PR #433, opened 2018-01-18
GCC/Clang: Treat void pointer arithmetic as error

a3f:
As an extension, GNU C treats `sizeof(void)` as `1`. MSVC doesn't.
Make it an error on GCC/Clang to avoid accidental MSVC breakage.


raysan5:
Hi @a3f, thanks for providing a  way to detect this issue!

Just corrected a couple of issues lately related to dealing with `void *` pointers on MSVC.

PR #440, opened 2018-01-25
 Make function calls without prior declaration an error

a3f:
which is the default behavior on C99 and up.


raysan5:
Hi @a3f, thanks for the improvements. Does it require some function review? I mean, some function called without previous declaration?

a3f:
Two of them, which are fixed in the latter two commits.

PR #442, opened 2018-01-26
Fix model rendering on framebuffer

### LINKED ISSUE #441: RenderTexture2D not rendering models.
I'm not sure which commit it was but whichever one removed the `rlClearScreenBuffers` from `BeginTextureMode` caused a bug where models won't render. Text and shapes render fine. Compiling the post processing example shows this bug and adding the line back seems to fix the issue.

TheLumaio:
Fixes #441

### DIFF
--- a/examples/shaders/shaders_postprocessing.c
+++ b/examples/shaders/shaders_postprocessing.c
@@ -126,10 +126,11 @@ int main()
         //----------------------------------------------------------------------------------
         BeginDrawing();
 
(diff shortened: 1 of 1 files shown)


raysan5:
Hi @TheLumaio, thanks for the PR but I'll just discard it because I'll correct this issue directly inside raylib as stated in the issue comment: https://github.com/raysan5/raylib/issues/441#issuecomment-360742885

PR #445, opened 2018-01-29
Update Makefile

RDR8:
Causes the example Makefile to prefer src/raylib.h over release/include/raylib.h. It fixes one example needing the newer header.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -165,7 +165,7 @@ endif
 
 # Define include paths for required headers
 # NOTE: Several external required libraries (stb and others)
-INCLUDE_PATHS = -I. -I$(RAYLIB_PATH)/release/include -I$(RAYLIB_PATH)/src -I$(RAYLIB_PATH)/src/external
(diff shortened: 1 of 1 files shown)


RDR8:
Breaking something else, I see. I'd like to override the release/subdirs withing the examples makefile but I can't figure out the precedence. I have everything in /usr/local/lib and /usr/local/include but can't seem to get raylib.h from there since it's finding it first through INCLUDE_PATHS. Maybe just a bad use case for the examples.

a3f:
Is this patch supposed to be against master?

raysan5:
This patch should be against develop branch...

RDR8:
Yes, I did mean to patch develop. Sorry for the silly mistakes. Stayed awake far too long last night really trying to find a good solution to path issues like this. It seems like  a common problem for new people working with the examples. Maybe it's a documentation thing. You try to warn about paths in the docs but questions come up regularly. Looking for way to answer them once and for all.

PR #456, opened 2018-02-03
Return false from InitWindow if glfwInit or glfwCreateWindow fails

a3f:
You can't do much with raylib if glfwInit or glfwCreateWindow fails,
currently it just exits by means of TraceLog(LOG_ERROR.
User code, however, might want to fall back to a text-only UI
or display a warning if raylib can't be used.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -362,7 +362,7 @@ extern void UnloadDefaultFont(void);        // [Module: text] Unloads default fo
 //----------------------------------------------------------------------------------
 // Module specific Functions Declaration
 //----------------------------------------------------------------------------------
-static void InitGraphicsDevice(int width, int height);  // Initialize graphics device
+static bool InitGraphicsDevice(int width, int height);  // Initialize graphics device
 static void SetupFramebufferSize(int displayWidth, int displayHeight);
 static void InitTimer(void);                            // Initialize timer
 static void Wait(float ms);                             // Wait for some milliseconds (stop program execution)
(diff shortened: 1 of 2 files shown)


raysan5:
Need to think about this change. I understand the pourpose but I'm usually not using that approach (returning true if function works) on any raylib function, my approach is usually an automatic fallback to a working alternative... but considering it's InitWindow(), there is not an alternative...

a3f:
I was mistaken about the reason. The assert was triggered because `glfwCreateWindow` failed and the `NULL` it returned was passed to `glfwSetWindowPos`. I have pushed a fix (44cd4fa) to `develop`, so it exits cleanly.

I adjusted this patch appropriately. I would be content with a `TryInitWindow` function too. I have graphical tests in my Perl distribution and I want to skip them if there's no graphic device available.

raysan5:
I've been investigating a bit about this design pattern (returning true on success), I know it's quite common in C language, but if changing `InitWindow()` to follow this convention, I should probably change some other functions for consistency (i.e. `ToggleFullscreen()`, `SetWindow*()`, `DisableCursor()`, `InitVrSimulator()`, `InitAudioDevice()`...).

Checking the API, I noticed I already define `IsVrSimulatorReady()` and `IsAudioDeviceReady()`, I'd propose a similar approach with `IsWindowReady()`, I prefer it than dealing with return values despite being a bit more cumbersome.
```c
InitWindow(800, 450, "try to init window");

if (IsWindowReady())
{
    // continue with window use
}
else
{
    // look for another display option
}
```

`InitGraphicsDevice()` is used internally, so it could return `true` or `false` as proposed but it should be reviewed for all the platforms for correct behaviour.

Some references I checked:
https://en.wikipedia.org/wiki/Return_code
https://stackoverflow.com/questions/385975/error-handling-in-c-code
https://stackoverflow.com/questions/559061/should-i-return-true-false-values-from-a-c-function
http://www.cplusplus.com/forum/beginner/180858/

a3f:
> I'd propose a similar approach with IsWindowReady()

Sounds good to me.

PR #459, opened 2018-02-04
Bump version to 1.9.4

a3f:


raysan5:
Ouch! I miss that! Should I launch a new release? I can cancel previous one...

a3f:
I think that would be the best, to avoid confusion as this release has the version embedded in the libraries and the pkg-config files. I am in favor of a 1.9.4-dev (To avoid breaking people who have pulled 1.9.3-dev already…)

raysan5:
> Should I push a 1.9.4?

Sure. I'm also changing it in raylib internals and integrating to master again.

raysan5:
Just as a note, we should find a way to automatically pick version from somewhere instead of changing version number text manually inside raylib.

a3f:
Suggestion: Check if git is available and use `git describe --tags`. This returns `1.9.3-dev` on `master`. If there are commits since the tag, it appends a commit id after the tag. Pass it as `RAYLIB_VERSION` macro to build process. If there's no git, we'll need to use a hardcoded value as fallback.

PR #463, opened 2018-02-11
mini_al: Support {Net,Open}BSD OSS

a3f:
Fixes this build failure:
http://www.cpantesters.org/cpan/report/a069fade-0e1f-11e8-a1cf-bb670eaac09d


raysan5:
Thanks for the improvement!

PR #464, opened 2018-02-11
Linux make install improvements

RDR8:
Quite a few additions to the makefiles here, mostly limited to LINUX and RAYLIB_LIBTYPE=SHARED, to avoid the dreaded `./core_basic_window: error while loading shared libraries: libraylib.so.1: cannot open shared object file: No such file or directory`  by formally installing libraylib (to the location of your choice) or by customizing RAYLIB_RUNTIME_PATH. This adds control, portability, and a running choice of shared libraries to the developer experience. Now distro-agnostic! 

That is to say, one can have multiple builds of libraylib to choose from at compile and run time, without sudo, using only make, on any ELF-aware system with binutils. FreeBSD and MacOS users might extend these changes if desired. I'm curious if this plays on WSL.

The default `make; make install; cd ../examples; make` behavior is unchanged.

With `make RAYLIB_LIBTYPE=SHARED`, the examples compile but still don't run against our staged .so in RAYLIB_RELEASE_PATH unless that location is provided _in full_ to RAYLIB_RUNTIME_PATH or RAYLIB_INSTALL_PATH. Instructions on how to do this are included in the comments of [examples/Makefile](https://github.com/RDR8/raylib/blob/develop-make-install-linux/examples/Makefile).

The examples DO compile and run against our version after `make install` or `make install RAYLIB_LIBTYPE=SHARED`. More context is provided by the comments in both Makefiles.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -31,6 +31,9 @@ RAYLIB_PATH        ?= ..
 
 # Define default options
 
+# One of PLATFORM_DESKTOP, PLATFORM_RPI, PLATFORM_ANDROID, PLATFORM_WEB
+PLATFORM              ?= PLATFORM_DESKTOP
+
 # Default path for raylib on Raspberry Pi, if installed in different path, update it!
 ifeq ($(PLATFORM),PLATFORM_RPI)
     RAYLIB_PATH       ?= /home/pi/raylib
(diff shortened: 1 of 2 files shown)


raysan5:
Hi RDR8, thanks for the review. Just a detail, on examples/Makefile PLATFORM should be defined before checking its value on line 36.

RDR8:
Plenty o' details to check. It's working well across three different machines for me.

RDR8:
Thank you!

raysan5:
Thanks to you for the improvements! :)

RDR8:
i'm with you that make is preferable to cmake. I've learned a lot and I hope that this will save someone some time down the line.

a3f:
@raysan5 I'd prefer squashing pull requests like these into a single commit (or requesting the author to squash them). Github has a Squash option for that. Makes history easier to follow

raysan5:
@a3f, agree.

RDR8:
I understand. Most of these came in at once in a file, but I couldn't leave it alone.

a3f:
@RDR8 You can use `git commit --amend` for smaller changes, or a `git rebase -i HEAD~5` at the end to squash the last 5 commits.

RDR8:
Ok, thank you. I'll learn for next time.

Issue #466, opened 2018-02-14
[build] Provide pre-configured projects for common IDEs


raysan5:
Prepare several pre-configured projects for common development environments avaliable in the different OS, similar to the already provided solutions for VS2015 and VS2017. The proposed IDEs would be:

- [x] Visual Studio 2015 (`Windows`) -> *raylib/projects/VS2015*
- [x] Visual Studio 2017 (`Windows`) -> *raylib/projects/VS2017*
- [x] Visual Studio Code (`Linux`, `Windows`, `macOS`) -> *raylib/projects/VSCode*
- [x] Geany (`Linux`, `Windows`, `macOS`) -> *raylib/projects/Geany*
- [x] Code::Blocks (`Linux`, `Windows`, `macOS`) -> *raylib/projects/CodeBlocks*
- [x] GNOME Builder (`Linux`) -> *raylib/projects/Builder*
- [ ] KDevelop (`Linux`, `Windows`, `macOS`) -> *raylib/projects/KDevelop*

Provided projects should be configured to compile raylib library (static/dynamic) and a simple example (i.e. `core_basic_window.c`).

OvermindDL1:
KDevelop is a significantly better C/C++ IDE for linux than about anything else listed there (and I think it has a windows port as well but it's still in-dev).  It's native format is just cmake files though (it makes a secondary file for holding workspace settings but that's fairly useless overall).  Code::Blocks is fairly bad considering what is available nowadays, and isn't Geany basically a glorified text editor?  Never heard of GNOME Builder though...  ^.^;

RDR8:
YIf you there are people who would benifit, you're welcome to support it. The more the merrier, just follow the naming convention. Code::Blocks is driven by requests on the forum and elsewhere. Geany isn't too different from Gedit or Kate. Like them, it has plugins to expose a terminal, version control, gdb, scripting support and various other low-level tools. It's included in Raspbian, is available everywhere else I've looked and runs well on every machine and DE I've tried. It is kind of boring but gets the job done in a clear way.

Builder is nice if you're on GNOME. Both GNOME and KDE are too heavy for some kinds of machines raylib runs on, though that is no reason to not support them. If you have a sensible config for KDevelop, it should be just a drop-in PR, though I won't speak for Ray.

raysan5:
Thanks @OvermindDL1 for the point, just added KDevelop to list.

RDR8:
[...]

PR #470, opened 2018-02-16
Add Builder project files

jubalh:
Takes care of Builder for https://github.com/raysan5/raylib/issues/466

### DIFF
--- a/project/Builder/README.md
+++ b/project/Builder/README.md
@@ -18,3 +18,6 @@ We added comments to the file to give you an idea which values you should edit.
 For a full overview of options please check the [meson manual](http://mesonbuild.com/Manual.html).
 
 In the provided file we assume that the build file is located at the root folder of your project, and that all your sources are in a `src` subfolder.
+
+Check out the `examples` directory for a simple example on how to use this template.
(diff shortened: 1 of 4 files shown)


a3f:
Could you remove the openAL and glfw3 dependency?

jubalh:
@a3f why?

a3f:
There's no need to link them in anymore. Audio is now handled by [`src/external/mini_al.h`](https://github.com/raysan5/raylib/blob/master/src/external/mini_al.h). GLFW3 is included in [`src/rglfw.c`](https://github.com/raysan5/raylib/blob/master/src/rglfw.c).

jubalh:
@a3f I believe that is not true for 1.8.0.
openSUSE ships raylib 1.8.0, because that is the latest stable release.

My idea of a workflow is that all packages get intalled from the distribution with as less self compilation as possible. Only the game the 'user' wants to write is what he should compile, the rest should installed via the package manager.

For this reason I prefer having this in there.

jubalh:
Added a note about it, so it can be changed later and people understand. Thanks for the hint @a3f.

raysan5:
ok, I see both points of view, agree with @a3f that it should be removed but also with @jubalh about the latest stable release. Just integrating it (thanks for the note) and it should be updated after next official stable release.

Issue #452, opened 2018-02-02
[build] Examples should build without the project being installed


yurivict:
Currently, it doesn't find the header and the library.
I tried to create a FreeBSD port option that would have built and installed examples, but with the current setup this is impossible.

a3f:
Is CMake an option? You could build the `develop` branch with

https://github.com/raysan5/raylib/blob/d1ef6869a9e9404ce7039ddddad8fa35240d4f42/.travis.yml#L40

raysan5:
Sorry, don't understand this issue, you mean adding [FreeBSD in release folder](https://github.com/raysan5/raylib/tree/develop/release)?

yurivict:
No. Currently examples fail when raylib isn't installed.

raysan5:
> Currently examples fail when raylib isn't installed.

Is it not the expected behaviour? I mean, raylib examples require the library to be installed to be compiled. Excuse me if I'm missing the point...

yurivict:
What I meant is that examples should build as part of the project using locally built library.
For example, if I would like the port to install examples, they should be built as part of the project, and then installed. Currently it expects that the library is installed before it can build examples.

RDR8:
Maybe take at look at #464 and see if it could be tweaked for FreeBSD.

RDR8:
#480 is more progressive. It solves:

>What I meant is that examples should build as part of the project using the locally built library.
For example, if I would like the port to install examples, they should be built as part of the project, and then installed. Currently it expects that the library is installed before it can build examples.

raysan5:
Just closing this issue. Please, feel free to reopen it if still not correct.

PR #480, opened 2018-02-22
Update Makefiles. Automate example usage of staged libraylib.so.

RDR8:
The default setting of `EXAMPLE_RUNTIME_PATH ?= $(RAYLIB_RELEASE_PATH)` in ../examples/Makefile makes it so that the examples run automatically against a freshly built, but not yet installed, libraylib.so in ../release/libs/linux. This is like an implicit USER mode. It does not require root permissions or any special input beyond `RAYLIB_LIBTYPE=SHARED` but is easily customizable if desired. The value of `EXAMPLE_RUNTIME_PATH` will override that of `RAYLIB_INSTALL_PATH` at an example's run time, if libraries exist at both locations. It is recommended to set `EXAMPLE_RUNTIME_PATH = $(RAYLIB_INSTALL_PATH)` if you don't want to pick up fresh versions of libraylib in ../release/libs/linux. If no libraylib exists at `EXAMPLE_RUNTIME_PATH`, ld.so moves on to check the location set by `ldconfig` during `../src$ sudo make install RAYLIB_LIBTYPE=SHARED`, i.e. `RAYLIB_INSTALL_PATH`.

Add `RAYLIB_PREFIX` and some adjustments to `RAYLIB_PATH` on `PLATFORM_OS = LINUX` to acquire the explicit string needed by `ldconfig` and possibly other tools that do not perform path variable expansion. Depending upon the capabilities of each platform's `make`, it may be possible to generalize `realpath` out of the `PLATFORM_OS` switch. I believe this emulates the usage of DESTDIR but with helpfully explicit naming of the path variables.

Additional comments to clarify usage. Miscellaneous cleanup for consistency.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -34,33 +34,19 @@ RAYLIB_PATH        ?= ..
 # One of PLATFORM_DESKTOP, PLATFORM_RPI, PLATFORM_ANDROID, PLATFORM_WEB
 PLATFORM              ?= PLATFORM_DESKTOP
 
-# Default path for raylib on Raspberry Pi, if installed in different path, update it!
-ifeq ($(PLATFORM),PLATFORM_RPI)
-    RAYLIB_PATH       ?= /home/pi/raylib
-endif
-
-# RAYLIB_RELEASE_PATH points to provided binaries or your immediate build of libraylib.
-# See below for additions.
-RAYLIB_RELEASE_PATH   ?= $(RAYLIB_PATH)/release/libs
-
 # Locations of your newly installed library and associated headers. See ../src/Makefile
-# On Linux, if you have installed raylib but cannot compile the examples, check that the
-# *_INSTALL_PATH values here are the same as those in src/Makefile or point to known locations.
-# To enable system-wide runtime linking to libraylib.so, run sudo ldconfig $(RAYLIB_INSTALL_PATH).
-# ldconfig is not necessary if using RAYLIB_RUNTIME_PATH below.
-RAYLIB_INSTALL_PATH   ?= /usr/local/lib/raysan5
-RAYLIB_H_INSTALL_PATH ?= /usr/local/include/raysan5
-
-# Set runtime path to custom location of shared library if desired, avoiding sudo ldconfig.
-# If you have compiled the examples but cannot run them, examine both RAYLIB_INSTALL_PATH and 
-# RAYLIB_RUNTIME_PATH. To see which libraries a built example is using, ldd core/core_basic_window;
-# Look for libraylib.so.1 => $(RAYLIB_INSTALL_PATH)/libraylib.so.1
-# or libraylib.so.1 => $(RAYLIB_RUNTIME_PATH)/libraylib.so.1 or similar listing.
-# Implemented for LINUX below with CFLAGS += -Wl,-rpath,$(RAYLIB_RUNTIME_PATH)
-# This should be a fully qualified path. RAYLIB_RELEASE_PATH doesn't work here
-# because it's a relative path. You must spell it out if needed.
-# To see the result, run readelf -d core/core_basic_window, looking at the RPATH attribute.
-RAYLIB_RUNTIME_PATH   ?= $(RAYLIB_INSTALL_PATH)
+# On Linux, if you have installed raylib but cannot compile the examples, check that
+# the *_INSTALL_PATH values here are the same as those in src/Makefile or point to known locations.
+# To enable system-wide compile-time and runtime linking to libraylib.so, run ../src/$ sudo make install RAYLIB_LIBTYPE_SHARED.
+# To enable compile-time linking to a special version of libraylib.so, change these variables here.
+# To enable runtime linking to a special version of libraylib.so, see EXAMPLE_RUNTIME_PATH below.
+# If there is a libraylib in both EXAMPLE_RUNTIME_PATH and RAYLIB_INSTALL_PATH, at runtime,
+# the library at EXAMPLE_RUNTIME_PATH, if present, will take precedence over the one at RAYLIB_INSTALL_PATH.
+# RAYLIB_INSTALL_PATH should be the desired full path to libraylib. No relative paths.
+DESTDIR ?= /usr/local
+RAYLIB_INSTALL_PATH ?= $(DESTDIR)/lib
+# RAYLIB_H_INSTALL_PATH locates the installed raylib header and associated source files.
+RAYLIB_H_INSTALL_PATH ?= $(DESTDIR)/include
 
 # Library type used for raylib: STATIC (.a) or SHARED (.so/.dll)
 RAYLIB_LIBTYPE        ?= STATIC
@@ -69,6 +55,7 @@ RAYLIB_LIBTYPE        ?= STATIC
 RAYLIB_BUILD_MODE     ?= RELEASE
 
 # Use external GLFW library instead of rglfw module
+# TODO: Review usage on Linux. Target version of choice. Switch on -lglfw or -lglfw3
 USE_EXTERNAL_GLFW     ?= FALSE
 
 # Use Wayland display server protocol on Linux desktop
@@ -103,6 +90,22 @@ ifeq ($(PLATFORM),PLATFORM_RPI)
     endif
 endif
 
+# RAYLIB_PATH adjustment for different platforms.
+# If using GNU make, we can get the full path to the top of the tree. Windows? BSD?
+# Required for ldconfig or other tools that do not perform path expansion.
+ifeq ($(PLATFORM),PLATFORM_DESKTOP)
+    ifeq ($(PLATFORM_OS),LINUX)
+        RAYLIB_PREFIX ?= ..
+        RAYLIB_PATH    = $(realpath $(RAYLIB_PREFIX))
+    endif
+endif
+# Default path for raylib on Raspberry Pi, if installed in different path, update it!
+# This is not currently used by src/Makefile. Not sure of its origin or usage. Refer to wiki.
+# TODO: update install: target in src/Makefile for RPI, consider relation to LINUX.
+ifeq ($(PLATFORM),PLATFORM_RPI)
+    RAYLIB_PATH       ?= /home/pi/raylib
+endif
+
 ifeq ($(PLATFORM),PLATFORM_WEB)
     # Emscripten required variables
     EMSDK_PATH = C:/emsdk
@@ -114,7 +117,8 @@ ifeq ($(PLATFORM),PLATFORM_WEB)
     EMSCRIPTEN=$(EMSDK_PATH)\emscripten\$(EMSCRIPTEN_VERSION)
 endif
 
-# Define raylib release directory for compiled library
+# Define raylib release directory for compiled library.
+# RAYLIB_RELEASE_PATH points to provided binaries or your freshly built version.
 ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     ifeq ($(PLATFORM_OS),WINDOWS)
         RAYLIB_RELEASE_PATH = $(RAYLIB_PATH)/release/libs/win32/mingw32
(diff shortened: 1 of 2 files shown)


examples/Makefile
```diff
@@ -34,41 +34,25 @@ RAYLIB_PATH        ?= ..
...
-# To enable system-wide runtime linking to libraylib.so, run sudo ldconfig $(RAYLIB_INSTALL_PATH).
-# ldconfig is not necessary if using RAYLIB_RUNTIME_PATH below.
+# On Linux, if you have installed raylib but cannot compile the examples, check that
+# the *_INSTALL_PATH values here are the same as those in src/Makefile or point to known locations.
+# To enable system-wide compile-time and runtime linking to libraylib.so, run ../src/$ sudo make install RAYLIB_LIBTYPE_SHARED.
+# To enable compile-time linking to a special version of libraylib.so, change these variables here.
+# To enable runtime linking to a special version of libraylib.so, see EXAMPLE_RUNTIME_PATH below.
+# If there is a libraylib in both EXAMPLE_RUNTIME_PATH and RAYLIB_INSTALL_PATH, at runtime,
+# the library at EXAMPLE_RUNTIME_PATH, if present, will take precedence over the one at RAYLIB_INSTALL_PATH.
 RAYLIB_INSTALL_PATH   ?= /usr/local/lib/raysan5
```
a3f: I would suggest adding a `DESTDIR ?= /usr/local` in the previous line and change this and the following to use `$(DESTDIR)` with lib and include concatenated. Also, why would we expect people to `#include <raysan5/raylib.h>`?  It's `<raylib.h>` now and if it changes (probably to `<raylib/raylib.h>`), it should be changed here as well as CMake. I see it's been only recently added. Why?

RDR8: I hear you. My variable naming scheme follows, more or less, the existing pattern in the makefiles for a couple of reasons. First, I'm not ready to interfere too deeply in @raysan5s workflow. Second, the purpose of these edits is to ease usage of the makefiles for someone who may never have seen any of these tools before. Explicit names aid clarity. Using DESTDIR, while generally correct, says nothing about what is actually there. I feel that generalizations aid experts but impede beginners. Once someone understands what's happening, they can easily generalize up.

My choice of install locations is somewhat more arbitrary. You'll notice that I'm quite greedy when grabbing headers from the raylib source tree. In addition to raylib.h, there are ~20 files in ../raysan5/include after installation. Some of them may not be strictly necessary but I like having them for reference anyway. With this arrangement, one can install raylib, copy examples/ out of the raylib tree, delete the raylib tree, then compile and run the examples like normal with no additional input. Why would someone do that? People like to tinker. This makes the installation much more robust than before. It also provides a central location for other interesting tools from raysan5 such as raygui.h. I also wanted to demonstrate the flexibility of the install target by calling attention to this "non-standard" location. I did a little research and found that such a bundling scheme isn't unusual. Changing it is trivial. How and why is described in the makefiles. 

As it is, ldconfig sets up the runtime links to ../lib/raysan5 automatically in the install target. Compile time links and includes are provided as usual, with -L and -I in the makefile. Usage is transparent and documented.

The biggest thing here is the elimination of recurring runtime linking problems for newcomers working from the wiki.

a3f: You can keep the variables and still provide a `DESTDIR` that overrides them if defined. Makes everyone happy.
Choice of installation by CMake or Makefiles shouldn't affect whether source code compiles or not. So we should have the same default header location for both Makefiles and CMake. As for the libraries, they should be installed to `$(DESTDIR)/lib` not a subdirectory.

RDR8: Ok, I understand your usage of DESTDIR and have no objections there. Thanks

> Choice of installation by CMake or Makefiles shouldn't affect whether source code compiles or not.

I agree. We can use whatever name you like. I chose raysan5 because raylib is not raygui is not raylib-lua etc. I do still need to bundle the headers somewhere. The uninstall target is kind of dumb and dangerous. There are explicit safety checks but it's recursive and leaves crumbs. I need to go ahead and get the full list of installed headers so they can be rm'd one by one. It was a late-night smash and grab. A review is a priority.

raysan5: Is this ready for merging? any other detail to review?

a3f: @raysan5 Would be nice to have the path changes and DESTDIR. @RDR8 You can do the edits, amend the commit with `git commit --amend && git push --force` to update the PR.

RDR8: It will be a few hours before I can sit down with my branch. You can go ahead if you want, but I will be working on it anyway. Thanks very much for your input!

RDR8: If we don't bundle the includes under $(DESTDIR)/include/raylib, I have quite a bit of pruning to do. As I said, this was an indiscriminate bulk grab. If you see anything obviously not needed or tremendously wrong-headed, now's the time. I suppose it's best to have a minimal traditional install but right now my include tree looks like, and in the end I personally want something like, this: (^;
```pi@Z83 /usr/local/include/raysan5 $ tree
.
├── audio.c
├── audio.h
├── camera.h
├── core.c
├── easings.h
├── external
│   ├── AL
│   │   ├── alc.h
│   │   ├── alext.h
│   │   ├── al.h
│   │   ├── efx-creative.h
│   │   ├── efx.h
│   │   └── efx-presets.h
│   ├── android
│   │   └── native_app_glue
│   │       ├── android_native_app_glue.c
│   │       ├── android_native_app_glue.h
│   │       └── NOTICE
│   ├── ANGLE
│   │   ├── angle_windowsstore.h
│   │   ├── EGL
│   │   │   ├── eglext.h
│   │   │   ├── egl.h
│   │   │   └── eglplatform.h
│   │   ├── GLES2
│   │   │   ├── gl2ext.h
│   │   │   ├── gl2.h
│   │   │   └── gl2platform.h
│   │   ├── GLES3
│   │   │   ├── gl3ext.h
│   │   │   ├── gl3.h
│   │   │   └── gl3platform.h
│   │   └── KHR
│   │       └── khrplatform.h
│   ├── dr_flac.h
│   ├── dr_wav.h
│   ├── glad.h
│   ├── glfw
│   │   ├── deps
│   │   │   ├── mingw
│   │   │   │   ├── dinput.h
│   │   │   │   ├── _mingw_dxhelper.h
│   │   │   │   └── xinput.h
│   │   │   └── vs2008
│   │   │       └── stdint.h
│   │   ├── include
│   │   │   └── GLFW
│   │   │       ├── glfw3.h
│   │   │       └── glfw3native.h
│   │   ├── LICENSE.md
│   │   ├── README.md
│   │   └── src
│   │       ├── CMakeLists.txt
│   │       ├── cocoa_init.m
│   │       ├── cocoa_joystick.h
│   │       ├── cocoa_joystick.m
│   │       ├── cocoa_monitor.m
│   │       ├── cocoa_platform.h
│   │       ├── cocoa_time.c
│   │       ├── cocoa_window.m
│   │       ├── context.c
│   │       ├── egl_context.c
│   │       ├── egl_context.h
│   │       ├── glfw3Config.cmake.in
│   │       ├── glfw3.pc.in
│   │       ├── glfw_config.h.in
│   │       ├── glx_context.c
│   │       ├── glx_context.h
│   │       ├── init.c
│   │       ├── input.c
│   │       ├── internal.h
│   │       ├── linux_joystick.c
│   │       ├── linux_joystick.h
│   │       ├── mappings.h
│   │       ├── mappings.h.in
│   │       ├── mir_init.c
│   │       ├── mir_monitor.c
│   │       ├── mir_platform.h
│   │       ├── mir_window.c
│   │       ├── monitor.c
│   │       ├── nsgl_context.h
│   │       ├── nsgl_context.m
│   │       ├── null_init.c
│   │       ├── null_joystick.c
│   │       ├── null_joystick.h
│   │       ├── null_monitor.c
│   │       ├── null_platform.h
│   │       ├── null_window.c
│   │       ├── osmesa_context.c
│   │       ├── osmesa_context.h
│   │       ├── posix_thread.c
│   │       ├── posix_thread.h
│   │       ├── posix_time.c
│   │       ├── posix_time.h
│   │       ├── vulkan.c
│   │       ├── wgl_context.c
│   │       ├── wgl_context.h
│   │       ├── win32_init.c
│   │       ├── win32_joystick.c
│   │       ├── win32_joystick.h
│   │       ├── win32_monitor.c
│   │       ├── win32_platform.h
│   │       ├── win32_thread.c
│   │       ├── win32_time.c
│   │       ├── win32_window.c
│   │       ├── window.c
│   │       ├── wl_init.c
│   │       ├── wl_monitor.c
│   │       ├── wl_platform.h
│   │       ├── wl_window.c
│   │       ├── x11_init.c
│   │       ├── x11_monitor.c
│   │       ├── x11_platform.h
│   │       ├── x11_window.c
│   │       ├── xkb_unicode.c
│   │       └── xkb_unicode.h
│   ├── jar_mod.h
│   ├── jar_xm.h
│   ├── mini_al.c
│   ├── mini_al.h
│   ├── par_shapes.h
│   ├── rgif.h
│   ├── stb_image.h
│   ├── stb_image_resize.h
│   ├── stb_image_write.h
│   ├── stb_perlin.h
│   ├── stb_rect_pack.h
│   ├── stb_truetype.h
│   ├── stb_vorbis.c
│   └── stb_vorbis.h
├── gestures.h
├── models.c
├── physac.h
├── raylib.h
├── raymath.h
├── rglfw.c
├── rlgl.c
├── rlgl.h
├── shader_distortion.h
├── shapes.c
├── text.c
├── textures.c
├── utils.c
└── utils.h

16 directories, 128 files`

Comment (RDR8) [reply examples/Makefile]: It may be viable to add an `EXTRA = FALSE` setting to do the default installation of just raylib.h.
A value of `TRUE`, together with `RAYLIB_INSTALL_MODE = USER` would satisfy me. Anyway, that's next level stuff. For now, I'm just going to remove all the junk, then come back later with a better solution for my playground.

Comment (RDR8) [reply examples/Makefile]: Ok, DESTDIR is all good but this other thing is interesting and extremely annoying. I'm back to the original error compiling textures_image_generation with the old raylib.h in release/include. As far as I can tell, it boils down to this. We are mixing "system" include locations with `#INCLUDE "raylib.h"` and I don't yet understand how to get the behavior I want without using `$(DESTDIR)/include/raylib`. In examples/Makefile, I have the line
 
` INCLUDE_PATHS = -I. -I$(RAYLIB_H_INSTALL_PATH) -I$(RAYLIB_PATH)/release/include -I$(RAYLIB_PATH)/src -I$(RAYLIB_PATH)/src/external `

With `RAYLIB_H_INSTALL_PATH = /usr/local/include/raylib`, the precedence works as expected so that when I compile, I get the newer raylib.h copied to `RAYLIB_H_INSTALL_PATH` from src/raylib.h during installation.

With `RAYLIB_H_INSTALL_PATH = /usr/local/include`, the precedence expected is ignored because, I think, it's a "system" location and we always `#INCLUDE "raylib.h"` not `#INCLUDE <raylib.h>` Now when compiling, I get the older raylib.h still in release/include even though the raylib.h I want was copied to `RAYLIB_H_INSTALL_PATH` from src/raylib.h during installation.
It seems to always prefer any "raylib.h" over the one in /usr/local/include proper.

Adding that one directory to `RAYLIB_H_INSTALL_PATH` makes it all go away. So long as we give `-I$(RAYLIB_H_INSTALL_PATH)` it makes no difference anywhere else.  "raylib.h" will be found just fine, I think. If it's any consolation, peeking into my default Debian derivative /usr/local/include, it's nothing but sub-directories.

Comment (a3f) [reply examples/Makefile]: Why would you install C source files into the include installation directory? Also, I didn't notice that `make install` has become a GNU/Linux only thing. That's not good. People who want to use the Makefile should do so, even while it's worked on.

Comment (a3f) [reply examples/Makefile]: Regarding your question, maybe change it to `-I$(RAYLIB_H_INSTALL_PATH) -isystem $(RAYLIB_PATH)/release/include`?

Comment (RDR8) [reply examples/Makefile]: > Why would you install C source files into the include installation directory?

No good reason. They're gone.

`make install` has been a Linux only target since I started playing with raylib about a year ago. The makefiles have worked with the instructions on the wiki since day one, but variations, such as installing GLFW with a package manager instead of compiling as instructed, brought the workflow to a halt. Things have improved quite a bit with documentation but these two issues of compile-time includes and run-time linking to a free-floating libraylib seem to still recur occasionally.

Comment (RDR8) [reply examples/Makefile]: Pretend I have my uncommitted edits here. Should I `git commit --amend` before `git pull --rebase` or rebase before `git commit --amend`?

Comment (RDR8) [reply examples/Makefile]: Hmm, I'm getting
```
$ git pull --rebase
First, rewinding head to replay your work on top of it...
Applying: Update Makefiles. Enable example usage of staged libraylib.so.
Using index info to reconstruct a base tree...
M	examples/Makefile
M	src/Makefile
Falling back to patching base and 3-way merge...
Auto-merging src/Makefile
Auto-merging examples/Makefile
CONFLICT (content): Merge conflict in examples/Makefile
error: Failed to merge in the changes.
Patch failed at 0001 Update Makefiles. Enable example usage of staged libraylib.so.
The copy of the patch that failed is found in: .git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

a3f: open` examples/Makefile` and fix the conflict manually. The conflicting lines are between `<<<<<<<` and `>>>>>>>`, you can't miss them. After that `git add . && git rebase --continue`.

RDR8: @a3f You are a thinking man's thinking man. Thanks very much for your help!

RDR8:
@Martinfx Would you care to review the usage of DESTDIR, RAYLIB_INSTALL_PATH, RAYLIB_H_INSTALL_PATH, and EXAMPLE_RUNTIME_PATH in src/Makefile and examples/Makefiles? Some of the hard-coded paths under PLATFORM_OS=FREEBSD could be replaced with these variables. I'm happy to do the edits if you feel that it is appropriate. Additionally, we have #452. Do you think is there any use to provide an install: target for FreeBSD? Thanks for your time.

RDR8:
DESTDIR is nice addition. Thanks for putting it out there. -isystem was also a key piece, as was your help with git.

Issue #478, opened 2018-02-21
[build] -fgnu89-inline 


gen2brain:
Go recently blacklisted all CFLAGS and now allows only flags that are whitelisted, see https://github.com/gen2brain/raylib-go/issues/32 . I did remove that flag and it does build without it, but I am not sure what are the consequences.

If this is a must I can probably open issue with them, to try to also whitelist `-fgnu89-inline`.

raysan5:
Right now I can not remember what was the use for that flag... actually, I think it came from a PR from someone...

gen2brain:
Ok, if it is not that important and it builds fine, I will just leave it removed then.

Thanks.

a3f:
I think we should remove it in raylib too. It might lead to build errors down the road, as MSVC doesn't use the gnu89 inline definition. If we got any `extern inline`s, they should be replaced by plain `inline`s and vice versa.

raysan5:
Flag removed in commit https://github.com/raysan5/raylib/commit/97bf2706c497f90ea4e3ead4e69819bc7a1f6083

raysan5:
About `extern inline`, it's just used in [raymath.h](https://github.com/raysan5/raylib/blob/master/src/raymath.h). It should be reviewed just there...

a3f:
The problem with `static inline` is possible duplication of out-of-line functions, so we want `raymath.h` to have inline definitions and one source file to have the only out of line definitions. With C99 inline semantics that means:

- `raymath.h` has plain `inline` definitions of the functions
- `core.c` or some other `.c` file has all prototypes with `extern` in front of them.

e.g.
`raymath.h`:
```
inline float Clamp(float value, float min, float max) 
{
    const float res = value < min ? min : value;
    return res > max ? max : res;
}
```
`core.c`:
```
#include <raymath.h>
extern float Clamp(float value, float min, float max);
```

raysan5:
Closing this issue, just need to review raymath.h, already a PR waiting and a related issue (https://github.com/raysan5/raylib/issues/498)

PR #482, opened 2018-02-22
raymath.h: Use C99 inline semantics

a3f:
RAYMATH_EXTERN_INLINE was renamed to RAYMATH_HEADER_ONLY, which user code
may define if they want to use it as header-only library. If multiple
files in the same project define RAYMATH_HEADER_ONLY, they might each
have duplicate out-of-line definitions of the same functions.

By default, raymath.h exposes inline definitions, which instructs the
compiler _not_ to generate out-of-line definitons, if out-of-line
definitions are required, those of the file defined with
RAYLIB_IMPLEMENTATION are used instead. There may be only one such file.

In C++ mode, the compiler will select only one out-of-line definition
automatically, so no need to define a RAYLIB_IMPLEMENTATION.

Unfortunately, we have to remove raymath function declaration from
raylib.h as those declarations would lead to duplicate out-of-line
definitions which would yield linker errors. This problem didn't
exist with GNU89 or C++, because there multiple defintions are ok,
but in C99 they aren't.

Addresses #478.

### DIFF
--- a/examples/models/models_skybox.c
+++ b/examples/models/models_skybox.c
@@ -62,7 +62,7 @@ int main()
 
             Begin3dMode(camera);
 
-                DrawModel(skybox, Vector3Zero(), 1.0f, WHITE);
+                DrawModel(skybox, (Vector3){0, 0, 0}, 1.0f, WHITE);
                 
                 DrawGrid(10, 1.0f);
 
--- a/src/core.c
+++ b/src/core.c
@@ -98,12 +98,11 @@
     #define _POSIX_C_SOURCE 199309L // Required for CLOCK_MONOTONIC if compiled with c99 without gnu ext.
 #endif
 
+#define RAYMATH_IMPLEMENTATION  // Define external out-of-line implementation of raymath here
+#include "raymath.h"            // Required for: Vector3 and Matrix functions
+
 #include "rlgl.h"           // raylib OpenGL abstraction layer to OpenGL 1.1, 3.3+ or ES2
 #include "utils.h"          // Required for: fopen() Android mapping
-
-#define RAYMATH_IMPLEMENTATION  // Use raymath as a header-only library (includes implementation)
-#define RAYMATH_EXTERN_INLINE   // Compile raymath functions as static inline (remember, it's a compiler hint)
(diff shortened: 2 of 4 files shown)


raysan5:
Need to review this carefully... actually, I tried to follow stb lessons on header-only library implementation: https://github.com/nothings/stb/blob/master/docs/stb_howto.txt

Also, need to see how it works with [rlgl_standalone](https://github.com/raysan5/raylib/blob/master/examples/others/rlgl_standalone.c) (uses [rlgl.h](https://github.com/raysan5/raylib/blob/master/src/rlgl.h)).

About `*ToFloat()` approach seems ok but I'd like to review naming...

a3f:
Interesting read. Thanks. They don't use C99 inline at all though.

raysan5:
This issue make me ask me if inline is really required, actually, despite understanding the concept and use, I don't know what is the real performance benefit... it should probably be profiled... some references I checked:

https://www.greenend.org.uk/rjk/tech/inline.html
https://stackoverflow.com/questions/1932311/when-to-use-inline-function-and-when-not-to-use-it
https://gustedt.wordpress.com/2010/11/29/myth-and-reality-about-inline-in-c99/

Also looked for some reference implementations [here](https://github.com/nothings/single_file_libs#vectors) and found a similar implementation on [HandmadeMath](https://github.com/HandmadeMath/Handmade-Math/blob/master/HandmadeMath.h). It's a very good reference.

a3f:
Of course, it's not required, but if you

* use C99
* want to allow inlining without link-time optimization
* want to avoid multiple static out-of-line copies of the same function
* want function pointers to inline functions to compare equal

That's how you do it.

The `HandmadeMath.h` you linked just defines everything as `static inline`, so if the compiler doesn't inline it, you get a static out-of-line definition, possibly multiple copies of the same code in the same application and increased code size → more to fit into instructions caches → lower performance.

a3f:
I pushed this PR along with a CMake patch by mistake. Sorry about that. I pushed a reverting commit.

raysan5:
Just merging changes to work on it from this point. Thanks for the review @a3f.

PR #495, opened 2018-02-25
Added option  ENABLE_DEBUG_MODE for debug raylib

Martinfx:


a3f:
There's already the standard `cmake -DCMAKE_BUILD_TYPE=Debug`...

a3f:
I would accept a patch that sets the default build type to `RelWithDebInfo` if none was specified though. See here for more information: https://blog.kitware.com/cmake-and-the-default-build-type/

PR #504, opened 2018-03-13
Fix potential bugs from static analysis

Martinfx:


src/raylib.h
```diff
@@ -692,7 +692,7 @@ extern "C" {            // Prevents name mangling of functions
 //------------------------------------------------------------------------------------
 
 // Window-related functions
-RLAPI void InitWindow(int width, int height, void *data);         // Initialize window and OpenGL context
+RLAPI void InitWindow(int width, int height, const char *title);         // Initialize window and OpenGL context
```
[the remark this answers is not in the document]
Martinfx: Maybe write some ifdef for concrete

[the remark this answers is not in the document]
Martinfx: Yes in raylib.h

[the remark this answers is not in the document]
raysan5: I just removed it to avoid PLATFORM definition in raylib.h and make it generic for all platforms. I know that `void *data` it's not the most elegant solution but for now I think is the best approach.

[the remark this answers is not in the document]
Martinfx: I see it as small security hole.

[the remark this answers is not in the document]
Martinfx: You look to commit 80b8bd769faddf0c1381502ac171f8d82d837ccc

src/raylib.h
```diff
@@ -692,7 +700,7 @@ extern "C" {            // Prevents name mangling of functions
 //------------------------------------------------------------------------------------
 
 // Window-related functions
-RLAPI void InitWindow(int width, int height, void *data);         // Initialize window and OpenGL context
+RLAPI void InitWindow(int width, int height, windowTitle name);  // Initialize window and OpenGL context
```
raysan5: Hey @Martinfx, many thanks for the effort on this review but for this time I prefer to leave it a it was, using `void *data` and avoiding any `PLATFORM_` define in `raylib.h`.

Martinfx: or rewrite as https://github.com/gfxprim/gfxprim/blob/master/demos/c_simple/backend_example.c
https://github.com/gfxprim/gfxprim/tree/master/include/backends

raysan5: I know, but all possible solutions make the code more complex than leaving it as it is...

a3f: I didn't dig into the code, why couldn't the platform specific data be stored to a raylib internal `static` buffer?

a3f: Ah, never mind. I thought it was just an output parameter.

a3f: For android, we could define `android_main` in raylib:
```
void android_main(struct android_app* app) {
    InitAndroidApp(app);
    exit(main(1, &(char[]){ "raylib", NULL }));
}
```
And allow use of `InitWindow(.., "windowTitle)` as usual. This is beneficial to users because they don't need #ifdefs in their code. If they require access to the `struct android_app`, we could add a `GetAndroidState` function. Would that be ok? I haven't tried raylib on Android.

Maybe a similar arrangement can be found for UWP, but I haven't seen UWP example code.

Martinfx: Ok

raysan5: Hi @a3f! That would be amazing if we could avoid forcing user to `#ifdef PLATFORM_ANDROID` on their code! Sorry, but I don't understand exactly your proposal... that code would be defined in `core.c`?

a3f: @raysan5 See #508.

raysan5: @a3f @Martinfx #508 solved this issue `InitWindow()` in Android platforms, maybe we could find a similar approach for UWP and then let `void *data` to be just `const char *title`.

UWP example is provided as a [VS2015 project](https://github.com/raysan5/raylib/tree/master/project/vs2015.UWP) and [here it is initialization point](https://github.com/raysan5/raylib/blob/master/project/vs2015.UWP/raylib.App.UWP/App.cpp#L425).

raysan5:
Just reviewed InitWindow() implementation, `void *data`pointer is not used any more. Changes implemented in commit https://github.com/raysan5/raylib/commit/bbdf9f48806acba5cac2480c0f60f1840d322ece

raysan5:
@Martinfx thank you very much for all the code and formatting reviews! Please let me know when I could merge them.

Martinfx:
Please merge. thank you M.

PR #505, opened 2018-03-13
Add memory sanitizer for better debug

Martinfx:


a3f:
Could you make the suggestion not to run ASAN and MSAN together more explicit? e.g.:

```
if (ENABLE_MSAN AND ENABLE_ASAN)
    MESSAGE(WARNING "Compiling with both AddressSanitizer and MemorySanitizer is not recommended")
endif()
```

PR #508, opened 2018-03-16
Allow use of main instead of android_main

a3f:
Inspired by #504.
Instead of requiring the user to do `PLATFORM_ANDROID` `#ifdef`ery,
have the `android_main` entry point exported by raylib and call
the user-defined `main`. This way many games could (in theory)
run unmodified on Android.

**This is untested!**

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -499,17 +499,33 @@ void InitWindow(int width, int height, void *data)
 #endif
 
 #if defined(PLATFORM_ANDROID)
+/* To allow easier porting to android, we allow the user to define a main function
+ * which we call from android_main, that we define ourselves
+ */
+extern int main(int argc, char *argv[]);
+void android_main(struct android_app *_app) {
+    char arg0[] = "raylib"; /* argv[] are mutable */
+    app = _app;
+    /* TODO should we maybe report != 0 return codes somewhere? */
+    (void)main(1, (char*[]){ arg0, NULL });
+}
+/* TODO add this to header, if apps really need it) */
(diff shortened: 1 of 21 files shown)


src/core.c
```diff
@@ -499,17 +499,33 @@ void InitWindow(int width, int height, void *data)
...
+ * which we call from android_main, that we define ourselves
+ */
+extern int main(int argc, char *argv[]);
+void android_main(struct android_app *_app) {
+    char arg0[] = "raylib"; /* argv[] are mutable */
+    app = _app;
+    /* TODO should we maybe report != 0 return codes somewhere? */
+    (void)main(1, (char*[]){ arg0, NULL });
+}
+/* TODO add this to header, if apps really need it) */
```
a3f: Do they need it?

raysan5: Actually, user doesn't need it, android app lifecycle is [managed internally by raylib](https://github.com/raysan5/raylib/blob/master/src/core.c#L2902).

raysan5: [android_native_app_glue.h](https://github.com/raysan5/raylib/blob/master/src/external/android/native_app_glue/android_native_app_glue.h#L111) has more info about that structure.

raysan5:
Just waiting to get some time to test this change properly when merging... I'm closing academic term this week and I'm a bit busy for long implementation/testing sessions...

raysan5:
Merged. Need to test it properly.

raysan5:
Amazing! It works like a charm! :D

Issue #506, opened 2018-03-15
[minial] Heap buffer overflow in audio examples on FreeBSD


Martinfx:
Hello i have problem with heap buffer overflow on FreeBSD 
```
mkdir build
cmake -DENABLE_ASAN=ON -DENABLE_UBSAN=ON ..
make
```

```
=================================================================
==57395==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000025e90 at pc 0x0000007517c2 bp 0x7fffdf9d64a0 sp 0x7fffdf9d6498
WRITE of size 4 at 0x602000025e90 thread T16777215
    #0 0x7517c1 in mal_pcm_f32_to_s32 asan/Code/raylib/src/external/mini_al.h:11520:17
    #1 0x747e48 in mal_pcm_convert asan/Code/raylib/src/external/mini_al.h:10554:38
    #2 0x75722d in mal_dsp_read_frames_ex asan/Code/raylib/src/external/mini_al.h:11110:9
    #3 0x753f84 in mal_dsp_read_frames asan/Code/raylib/src/external/mini_al.h:11045:12
    #4 0x759948 in mal_device__read_frames_from_client asan/Code/raylib/src/external/mini_al.h:2655:29
    #5 0x75bc62 in mal_device__start_backend__oss asan/Code/raylib/src/external/mini_al.h:6997:9
    #6 0x726d35 in mal_device__start_backend asan/Code/raylib/src/external/mini_al.h:9038:18
    #7 0x725f5e in mal_worker_thread asan/Code/raylib/src/external/mini_al.h:9233:31
    #8 0x800c72bc4 in thread_start /usr/src/lib/libthr/thread/thr_create.c:289:16

0x602000025e91 is located 0 bytes to the right of 1-byte region [0x602000025e90,0x602000025e91)
allocated by thread T0 here:
    #0 0x337c0c in __interceptor_malloc (asan/Code/raylib/build/examples/audio_sound_loading+0x337c0c)
    #1 0x7369dc in mal_device_init__oss asan/Code/raylib/src/external/mini_al.h:6977:40
    #2 0x72ffb2 in mal_device_init asan/Code/raylib/src/external/mini_al.h:9770:22
    #3 0x39958e in InitAudioDevice asan/Code/raylib/src/audio.c:509:14
    #4 0x349759 in main asan/Code/raylib/examples/audio/audio_sound_loading.c:25:5
[...]

PR #509, opened 2018-03-18
Fix a buffer overflow in the OSS/BSD backend of mini_al.

mackron:
Related to issue #506.

This should fix a crash on OSS/BSD resulting from a buffer overflow.


raysan5:
Many thanks for the update @mackron!

PR #510, opened 2018-03-20
Fix an infinite loop in ALSA backend of mini_al.

mackron:


raysan5:
Many thanks for the quick fix! :D

Issue #439, opened 2018-01-25
[core] Implement orthographic projection for 3D camera


srseil:
I need an orthographic projection for the game I am currently working on, and as far as I can see this functionality is currently not exposed through the API. I think it would be useful to have this integrated in the current camera system. There could be a separate member in the `Camera` struct to set the projection, defaulting to perspective. Something along those lines.

raysan5:
Don't know if that's what you mean but, have you checked [core_2d_camera](https://github.com/raysan5/raylib/blob/master/examples/core/core_2d_camera.c) example?

Also, maybe related, you can check example [core_world_screen](https://github.com/raysan5/raylib/blob/master/examples/core/core_world_screen.c) example.

victorfisac:
@raysan5 I think he wants a 3d camera but orthographic (discarding multiplying projection matrix). It is useful to make a make like Monument Valley where you need a 3d point of view but orthographic.

Camera2D is not prepared for 3d models drawing.

Gamerfiend:
I second this, and am working on it myself but I haven't gotten very far as I'm a novice graphics programmer. Here is the difference @raysan5 

[http://db-in.com/images/projection_example.gif](url)

raysan5:
@victorfisac thanks for the clarification! You're all right, I'll add support for orthographic projection. 

Need to look for the best way to do that.

victorfisac:
@raysan5 Camera struct can implement this feature with an enum. Also, maybe the first result is not as expected and you have to implement a 'size' field to scale all drawn vertices; Unity has this in Camera component (with a value of '5' by default).

raysan5:
Hi @victorfisac! Thanks for the recommendation! Need to evaluate other options but it sounds lika a good solution to me...

raysan5:
Implemented in PR https://github.com/raysan5/raylib/pull/513

PR #513, opened 2018-03-24
Add orthographic 3d rendering mode

autious:
I needed orthographic rendering for my own things. This solution doesn't modify the existing Camera struct but instead uses a new one called CameraOrtho and a new Begin3dModeOrtho() call. This means that the input systems and picking functions are not compatible, but for that to be possible they would need a lot of internal reimplementation. Some of the semantics might not be compatible either, suggesting that only a partial reimplementation would be necessary, further suggesting that a separate set of functionality is warranted.

This is a (at least partial) solution of issue #439

### DIFF
--- a/examples/models/models_orthographic_projection.c
+++ b/examples/models/models_orthographic_projection.c
@@ -1,0 +1,109 @@
+/*******************************************************************************************
+*
+*   raylib [models] example - Show the difference between perspective and orthographic projection 
+*
+*   This program is heavily based on the geometric objects example
+*
+*   This example has been created using raylib 1.0 (www.raylib.com)
+*   raylib is licensed under an unmodified zlib/libpng license (View raylib.h for details)
+*
+*   Copyright (c) 2014 Ramon Santamaria (@raysan5)
+*
+********************************************************************************************/
+
+#include "raylib.h"
+
+int main()
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    int screenWidth = 800;
+    int screenHeight = 450;
+    bool view_ortho = true;
+
+    InitWindow(screenWidth, screenHeight, "raylib [models] example - geometric shapes");
+
+    // Define the camera to look into our 3d world
+    Camera camera; 
+
+    SetTargetFPS(60);   // Set our game to run at 60 frames-per-second
+    //--------------------------------------------------------------------------------------
+
+    // Main game loop
+    while (!WindowShouldClose())    // Detect window close button or ESC key
+    {
+        // Update
+        //----------------------------------------------------------------------------------
+        // TODO: Update your variables here
+        //----------------------------------------------------------------------------------
+        //
+
+        // Input
+        //----------------------------------------------------------------------------------
+        if(IsKeyPressed(KEY_SPACE)) 
+        {
+            view_ortho = !view_ortho;
+        }
+
+
+        // Draw
+        //----------------------------------------------------------------------------------
+
+        if(view_ortho) 
+        {
+            camera = (Camera){{ 0.0f, 10.0f, 10.0f }, { 0.0f, 0.0f, 0.0f }, { 0.0f, 1.0f, 0.0f }, 10.0f, CAMERA_ORTHOGRAPHIC };
+        } 
+        else 
+        {
+            camera = (Camera){{ 0.0f, 10.0f, 10.0f }, { 0.0f, 0.0f, 0.0f }, { 0.0f, 1.0f, 0.0f }, 45.0f, CAMERA_PERSPECTIVE };
+        }
+
+        BeginDrawing();
+
+            ClearBackground(RAYWHITE);
+
+            Begin3dMode(camera);
+
+                DrawCube((Vector3){-4.0f, 0.0f, 2.0f}, 2.0f, 5.0f, 2.0f, RED);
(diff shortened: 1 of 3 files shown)


victorfisac:
I think it would be better to implement a ENUM like CAMERA_PERSPECTIVE, CAMERA_ORTHOGRAPHIC. And use just Begin3dMode()/Begin3dMode().

Inside Begin3dMode(), check which value has this enum and initialize 'double top' and use rlFrustum or rlOrtho; because the other code inside of Begin3dModeOrtho() is same as Begin3dMode().

What do you think, @autious @raysan5 ?

autious:
I considered that, but it would break all existing code that uses the c-style initializer for the Camera struct and i didn't feel comfortable making a breaking change. This fix will not affect any existing code. I would respect such a change to occur only on the authority of the core team of maintainers.

a3f:
> it would break all existing code that uses the c-style initializer for the Camera struct

It would only break ABI (which we intend to do with raylib 2.0, anyway), but code (API) should stay the same. Just add the new `enum` as the last `struct` member and it will automatically be initialized to `0` (which `CAMERA_PERSPECTIVE` should be defined as).

Minor downside would be overzealous warnings (`-Wmissing-field-initializers`).

raysan5:
Agree with @victorfisac and @a3f.

autious:
Understood, i'll create a variation of the implementation matching the suggestions.

autious:
Modified version uploaded. I haven't tested all the functions that use a Camera as a parameter, but i have modified the Ray cast and world-point projection to mouse coordinate functions, they appear to function correctly in my personal project (with a minor deviation of ~0.1 between each other)

autious:
I also created an extra example program that allows the user to switch between perspective and orthogonal rendering, to demonstrate the visual difference. A better variant for student insight might be to have a camera moving around the object and zooming in/out to fully visualize it.

src/core.c
```diff
@@ -919,13 +919,26 @@ void Begin3dMode(Camera camera)
...
     rlLoadIdentity();                   // Reset current matrix (PROJECTION)
 
-    // Setup perspective projection
-    float aspect = (float)screenWidth/(float)screenHeight;
-    double top = 0.01*tan(camera.fovy*0.5*DEG2RAD);
-    double right = top*aspect;
+    if(camera.type == CAMERA_PERSPECTIVE) 
+    {
+        // Setup perspective projection
+        float aspect = (float)screenWidth/(float)screenHeight;
```
victorfisac: Why you initialize aspect inside if/else. It has the same value in both cases.

src/core.c
```diff
@@ -919,13 +919,26 @@ void Begin3dMode(Camera camera)
...
+        double top = 0.01*tan(camera.fovy*0.5*DEG2RAD);
+        double right = top*aspect;
```
victorfisac: Same with this.

examples/models/models_orthographic_projection.c
```diff
@@ -0,0 +1,109 @@
...
+        if(IsKeyPressed(KEY_SPACE)) 
+        {
+            view_ortho = !view_ortho;
+        }
+
+
+        // Draw
+        //----------------------------------------------------------------------------------
+
+        if(view_ortho)
```
victorfisac: This should be executed inside IsKeyPressed condition statement

examples/models/models_orthographic_projection.c
```diff
@@ -0,0 +1,109 @@
...
+
+#include "raylib.h"
+
+int main()
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    int screenWidth = 800;
+    int screenHeight = 450;
+    bool view_ortho = true;
```
victorfisac: You should be using camera->type instead of a global boolean.

autious:
Some changes have been made based on suggestions made by @victorfisac

examples/models/models_orthographic_projection.c
```diff
@@ -0,0 +1,109 @@
...
+int main()
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    int screenWidth = 800;
+    int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [models] example - geometric shapes");
+
+    const Camera perspective_camera = (Camera){{ 0.0f, 10.0f, 10.0f }, { 0.0f, 0.0f, 0.0f }, { 0.0f, 1.0f, 0.0f }, 45.0f, CAMERA_PERSPECTIVE };
```
victorfisac: You can reduce code amount using just one Camera variable (without const) with perspective values and define values:
```
#define      FOVY_ORTHOGRAPHIC        10.0f
#define      FOVY_PERSPECTIVE         45.0f
```

So, you can then:

```
if(IsKeyPressed(KEY_SPACE)) 
{
     if(camera.type == CAMERA_PERSPECTIVE) 
     {
          camera.fovy = FOVY_ORTHOGRAPHIC;
          camera.type = CAMERA_ORTHOGRAPHIC;
     } 
     else 
     {
          camera.fovy = FOVY_PERSPECTIVE;
          camera.type = CAMERA_PERSPECTIVE;
     }
}
```

raysan5:
Hey @autious! Many thanks for working on this improvement!

@victorfisac thanks for the review and the proposed ideas!

autious:
You're welcome. It's always nice to contribute a little.

PR #516, opened 2018-03-31
Initialize the timer after the graphics device on desktop and web.

jacquesh:
This is already the order that is used for Android (InitGraphicsDevice before InitTimer).

It doesn't appear to make a difference on desktop but on web using the timer before it's been initialized (by glfwInit, inside InitGraphicsDevice) causes the a long (and variable but often several seconds) sleep between the first and second frame.

I've seen many instances of sleeping for between 1 and 3 seconds but also a few of >100 seconds.

I see previousTime isn't used in InitGraphicsDevice so I would expect there to be few negative side-effects of this, but admittedly I haven't looked particularly hard.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -442,13 +442,13 @@ void InitWindow(int width, int height, void *data)
     uwpWindow = (EGLNativeWindowType)data;
 #endif
 
-    // Init hi-res timer
-    InitTimer();
-
     // Init graphics device (display device and OpenGL context)
     // NOTE: returns true if window and graphic device has been initialized successfully
     windowReady = InitGraphicsDevice(width, height);
     if (!windowReady) return;
+
+    // Init hi-res timer
+    InitTimer();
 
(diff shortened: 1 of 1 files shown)


a3f:
It used to be this way before I moved it around in 468309d06c7ac3e3278057cfb1cfc19b332833bf.
> Also moves InitTimer() before InitGraphicsDevice(), to allow it to be
tested even if InitWindow ultimately fails.

```
GLFWAPI double glfwGetTime(void)
{
    _GLFW_REQUIRE_INIT_OR_RETURN(0.0);
    return (double) (_glfwPlatformGetTimerValue() - _glfw.timer.offset) /
        _glfwPlatformGetTimerFrequency();
}
```

shows [glfwGetTime](https://github.com/glfw/glfw/blob/3c9011030f6faffaa9d33639b4139187329fc079/src/input.c#L1276
) should only be called after `glfwInit`...

Thanks for the fix!

Issue #517, opened 2018-04-03
Some CMake issues and additions


gen2brain:
Hello,

I started working on Gentoo ebuild (https://gist.github.com/gen2brain/84e530f6dcb61d9a2eda11decf8d0230) for raylib and I ran into some issues with cmake. Here is the patch for cmake that I am applying currently in the ebuild https://gist.github.com/gen2brain/6af2dd0b69366249e1a8e77ef1fcb48e . It does several things:

- includes GNUInstallDirs so it can be used for multilib builds (i.e. ABI_X86 64, 32, x32) , so `lib/` dir is not hardcoded
- adds USE_AUDIO option so raylib can be built without alsa on Linux, as supported in Makefile
- adds USE_WAYLAND option so raylib can be built with Wayland support

Also, I don't compile examples and games, if USE flags are provided then they are just copied to /usr/share/doc . After that, user can not compile examples that includes `raymath.h` because header file is not installed. Also, if you just use plain cmake to build examples, they fail on `physac.h`.

I don't know much about CMake, so I just wanted to report here for now, maybe there are better ways for this.

I would prefer not to use patch when I open a new bug report for inclusion of new package in repo, it would be much nicer not to patch anything.

a3f:
Patch looks good to me. Could you create a PR?

Installing other headers is discussed in #306.

raysan5:
Hey @gen2brain! Nice addition! I'm not an expert on CMake, maybe @a3f could take a look to it.

Just feel free to send a PR for this update.

gen2brain:
Sure, I will create PR.

raysan5:
Could we close this issue?

PR #518, opened 2018-04-03
Add GNUInstallDirs and USE_AUDIO/USE_WAYLAND options to CMake

gen2brain:
- include GNUInstallDirs so it can be used for multilib builds (i.e. ABI_X86 64, 32, x32) , so lib/ dir is not hardcoded
 - add USE_AUDIO option so raylib can be built without alsa on Linux, as supported in Makefile
 - adds USE_WAYLAND option so raylib can be built with Wayland support

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Setup the project and settings
 project(raylib)
 include("../utils.cmake")
+include(GNUInstallDirs)
 
 set(PROJECT_VERSION 1.9.4)
 set(API_VERSION 1)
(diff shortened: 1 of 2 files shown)


a3f:
Would be nice to have Travis CI check a Wayland configuration. Do you know which packages would be required for Ubuntu?

gen2brain:
Not sure abouto Wayland packages for Ubuntu, `libwayland-client` package is needed for sure, if `libwayland-cursor` library is not there, or a dependency it is also needed. And then, `libwayland-egl` comes from `mesa` compiled with wayland support, not sure how that is in Ubuntu.

raysan5:
I'm linking with `-lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon` but haven't tested myself...

a3f:
This unearthed a small issue: A full CMake build including examples would fail, because it tries to build the `audio_*` examples.

a3f:
@gen2brain Should the pkg-config file be installed to `${CMAKE_INSTALL_LIBDIR}/pkgconfig` as well?

(For future (cross-)reference: This PR relates to issue #517)

PR #521, opened 2018-04-06
Added GetCollisionRayModel

TheLumaio:
An alternative to `GetCollisionRayMesh` which should apply transformations from models.


TheLumaio:
I agree, I'll update the PR.

TheLumaio:
I don't know if I can just remove the librarylib.a file from the pr so I reverted it to the previous commit, not sure if that's what you meant.

PR #522, opened 2018-04-07
Refactor all #define SUPPORT_* into a config.h

a3f:
Instead of requiring users to manually edit or add `#define SUPPORT_*`s to the appropriate source file, centralize all library configuration to an internal `config.h` file.

This file is generated at configuration time from CMake options and hardcoded for the Makefile and other build systems.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,21 +1,25 @@
 # Setup the project and settings
 project(raylib)
-include("../utils.cmake")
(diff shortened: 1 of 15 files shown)


a3f:
Undefining `SUPPORT_SAVE_PNG` is not really supported, but previously, it would have failed to compile anyway, so I think that's ok…

raysan5:
Interesting change! I'll review it ASAP.

PR #524, opened 2018-04-13
Add more key definitions

TheLumaio:
My assumption is that these are the same as the GLFW definitions so that's what I mirrored.


raysan5:
Thanks for the improvement!

PR #531, opened 2018-05-06
Patch implicit type

sherjilozair:
This is a bug waiting to happen, so let's fix it before it does. :)


raysan5:
Thanks for the review @sherjilozair!

Issue #536, opened 2018-05-08
[build] Test Wayland in Travis


gen2brain:
I think that every time I updated GLFW, wayland build was broken, and new protocol files (.c/.h) generated with wayland-scanner needed to be included. It would be good to test build in travis, but I am not sure if it can be just new matrix with some env in .travis.yml, something like that.

I tested in latest Ubuntu and there `apt-get install libgl1-mesa-dev libwayland-dev libxkbcommon-dev` is enough, and I also added this symlink `ln -sf /usr/lib/x86_64-linux-gnu/libwayland-egl.so.1.0.0 /usr/lib/x86_64-linux-gnu/libwayland-egl.so`

For Ubuntu 14, that version is used in CI, `libwayland-egl1-mesa-lts-xenial` must also be included, and again, I had to make a symlink, not sure why just so.1.0.0 is installed without any symlink.

Also, for latest GLFW I had to generate and include more .c files, see https://github.com/gen2brain/raylib-go/blob/master/raylib/cgo_linux.go#L19 (current raylib misses just wayland-idle-inhibit but after you update GLFW more are needed).

I updated GLFW because of this commit, it sounds bad https://github.com/glfw/glfw/commit/8b9221d845e4ae9b314f07375cc3a520ba8d98d7

a3f:
I copied over the Travis CI configuration that GLFW itself uses for wayland, but building the examples fails. Could you take a look?

a3f:
I've updated GLFW and added two Linux configurations to Travis CI: 
* X11+external GLFW
* Wayland+in-tree GLFW.

I didn't have to fixup any symlinks manually.

a3f:
I reverted 2d6fb5c628068757387525e190c3afdbe33ae9c6, because some of my Alien::raylib tests failed. Reopenning this issue as a reminder till it's sorted out...

PR #538, opened 2018-05-09
CMake+CI: Add Wayland configuration

a3f:
Addresses #536.


a3f:
Doesn't work yet; Compiling examples fails because of [X11 symbols left unresolved](https://travis-ci.org/a3f/raylib/jobs/376905087#L1083):
```
../release/libraylib.so.1.9.7: undefined reference to `XAllocClassHint'
../release/libraylib.so.1.9.7: undefined reference to `XSendEvent'
../release/libraylib.so.1.9.7: undefined reference to `XTranslateCoordinates'
../release/libraylib.so.1.9.7: undefined reference to `XFreeColormap'
../release/libraylib.so.1.9.7: undefined reference to `XGetScreenSaver'
...
```
@gen2brain Can you take a look?

gen2brain:
All those symbols should be in libX11.so, not sure why I didn't need it. Final binary does link to libX11.so, but that comes from libGL.so I think.

a3f:
https://github.com/raysan5/raylib/blob/b8ca51fd01a68b9b6040f8c3c631eedefd7d7735/utils.cmake#L20-L24

Shouldn't this have been `add_definitions(-D_GLFW_X11=1)` instead? `set` wouldn't affect the `rglfw` build. I assume you are using an external GLFW?

Changing it to `add_definitions`, gives me:
>
```
[  2%] Building C object release/CMakeFiles/raylib_shared.dir/rglfw.c.o
In file included from /home/a3f/dl/raylib/src/external/glfw/src/internal.h:190:0,
                 from /home/a3f/dl/raylib/src/external/glfw/src/context.c:28,
                 from /home/a3f/dl/raylib/src/rglfw.c:52:
/home/a3f/dl/raylib/src/external/glfw/src/wl_platform.h:55:47: fatal error: wayland-xdg-shell-client-protocol.h: No such file or directory
```

and `apt-file find wayland-xdg-shell-client-protocol.h` on my Ubuntu 16.04.3 tells me that only `libqt5waylandclient5-dev: /usr/include/x86_64-linux-gnu/qt5/QtWaylandClient/5.5.1/QtWaylandClient/private/wayland-xdg-shell-client-protocol.h` provides that header, which seems unrelated.

I think `wayland-scanner` is what generates these headers. Any idea what needs to be done?

gen2brain:
As I understand, it is expected from every Wayland compositor to provide that files, Wayland is just protocol and every compositor must implement everything. Crazy.
You can generate those files with wayland-scanner like this https://github.com/gen2brain/raylib-go/blob/master/raylib/external/scripts/glfw-generate-wayland.sh , not sure what GLFW does in cmake, probably something similar.

a3f:
I will push a patch soon that should fix this.

a3f:
You guys aren't getting notifications every time I force push, are you?

a3f:
@raysan5 I've removed building `rglfw.c` with CMake. Wayland support requires generating headers for GLFW and replicating that in raylib is a bit overkill, so now we use GLFW's CMake setup to build a static GLFW library and link against that. `rglfw.c` remains for use with the Makefile.

a3f:
I just noticed this installs GLFW headers as well now. I am inclined to leave it as is, in case a user needs it.

a3f:
[This broke the `Alien::raylib` build on a couple of machines](http://www.cpantesters.org/distro/A/Alien-raylib.html):

```
#     Can't load 'testalienghM_t/auto/My/Raylib/Raylib.so' for module My::Raylib: testalienghM_t/auto/My/Raylib/Raylib.so: undefined symbol: glfwGetProcAddress at /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.26.1/fc43/lib/5.26.1/x86_64-linux-ld/DynaLoader.pm line 193.
#  at /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.26.1/fc43/lib/site_perl/5.26.1/Test/Alien.pm line 402.
# Compilation failed in require at /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.26.1/fc43/lib/site_perl/5.26.1/Test/Alien.pm line 402.
# BEGIN failed--compilation aborted at /home/sand/src/perl/repoperls/installed-perls/host/k93msid/v5.26.1/fc43/lib/site_perl/5.26.1/Test/Alien.pm line 402.
```

If it takes too long to debug, I'll revert my change.

PR #539, opened 2018-05-10
CMake: Don't create symlinks on unsupporting file systems

a3f:
Panders to the idiosyncrasies of my work flow:
I have my raylib build directory mounted as a VirtualBox vboxfs for use
with my Linux VM, but vboxfs doesn't support symlinks, while raylib shared
library versioning on Unix expects symlinks to work.
If this happens, library versioning is now disabled on Unix with
an error message instead of just failing the build.


raysan5:
Thanks for the improvement! :)

PR #542, opened 2018-05-12
Makefile: Remove unnecessary -no-pie for older GCC support

### LINKED ISSUE #540: Unrecognized flag "-no-pie"
Hello. I tried to compile the examples folder using: make PLATFORM=PLATFORM_DESKTOP (I'm running Debian) and gcc complains about "-no-pie" flag. After removing it (line 218) I could compile the project just fine.

a3f:
Currently, if:
* GCC doesn't supports `-no-pie`: Build error
* GCC supports `-no-pie`
    * GCC is not configured with `--enable-default-pie`: No-op
    * GCC is configured with `--enable-default-pie`:
            Slightly worse performance because we still generate `-fpie` code
            (`-pie` affects linker, `-fpie` affects compiler)

So instead of probing for existence of `-fno-pie -no-pie`, remove it altogether.

Fixes #540: Build breakage on Debian 8 with gcc 4.9.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -215,7 +215,7 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
             #CC = clang
         endif
         ifeq ($(RAYLIB_LIBTYPE),STATIC)
-        CFLAGS += -no-pie -D_DEFAULT_SOURCE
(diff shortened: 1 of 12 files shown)


raysan5:
@a3f thanks for reviewing this issue

PR #545, opened 2018-05-18
Increase API version

jubalh:
Please do this whenever necessary. And adapt it before releasing a new
RC. The RC should already have the correct one.


raysan5:
Thanks for the update!

Issue #551, opened 2018-05-26
[build] Cannot build raylib 2.0.0-dev with -DSHARE on FreeBSD 


Martinfx:
Hello i have problem with compilation on FreeBSD 11 with raylib 2.0.0-dev.

```
#!/bin/sh
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=debug -DSHARED=ON -DBUILD_EXAMPLES=NO -DBUILD_GAMES=NO
gmake -j3
cd ..
```

```
-- The C compiler identification is Clang 4.0.0
-- The CXX compiler identification is Clang 4.0.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test COMPILER_HAS_THOSE_TOGGLES
-- Performing Test COMPILER_HAS_THOSE_TOGGLES - Success
-- Testing if -Werror=pointer-arith can be used -- works
-- Testing if -Werror=implicit-function-declaration can be used -- works
-- Testing if -fno-strict-aliasing can be used -- works
-- Testing if file system supports symlinks
-- Testing if file system supports symlinks -- supported
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Could NOT find Vulkan (missing: VULKAN_LIBRARY VULKAN_INCLUDE_DIR) 
-- Using X11 for window creation
-- Looking for XOpenDisplay in /usr/local/lib/libX11.so;/usr/local/lib/libXext.so
-- Looking for XOpenDisplay in /usr/local/lib/libX11.so;/usr/local/lib/libXext.so - found
-- Looking for gethostbyname
-- Looking for gethostbyname - found
-- Looking for connect
-- Looking for connect - found
-- Looking for remove
[...]

PR #552, opened 2018-05-26
Fix cannot find -lX11

Martinfx:
Fixed bug in utill.cmake on FreeBSD (cannot find -X11 libraries) for issue https://github.com/raysan5/raylib/issues/551

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -53,6 +53,11 @@ if(${PLATFORM} MATCHES "Desktop")
   endif()
 endif()
 
+# Ugly crutch. Temporary workaround for #551
+if("${CMAKE_SYSTEM_NAME}" MATCHES "(Free|Net|Open)BSD")
+    link_directories("/usr/local/lib")
+endif()
+
 if(CMAKE_SYSTEM_NAME STREQUAL Linux)
   set(LINUX TRUE)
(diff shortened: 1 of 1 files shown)


a3f:
Please rebase on top of newest master (`git pull --rebase`).

Martinfx:
Ok, i rebased.

yurivict:
This line https://github.com/raysan5/raylib/pull/552/files#diff-01de9067ac77e00c2f64922cadb78a1dR58 hardcodes /usr/local/ The correct way to do this is to use ```${CMAKE_INSTALL_PREFIX}```.

raysan5:
Hi @yurivict, thank you very much for the review.

Please, @Martinfx could you take a look to this fix, I'm afraid I'm not a CMake expert...

yurivict:
I also think that this patch shouldn't even be needed. This is because this library path ```${CMAKE_INSTALL_PREFIX}/lib``` should always be added anyway.
So the bug is likely somewhere else.

Martinfx:
Hello @raysan5 could you please make re-release with last patches ?

yurivict:
The patch should be:
```
if("${CMAKE_SYSTEM_NAME}" MATCHES "(Free|Net|Open)BSD|DragonFly")
    link_directories("${CMAKE_INSTALL_PREFIX}/lib")
endif()
```

PR #557, opened 2018-06-22
Added support OpenBSD, NetBSD, DragonFly OS

Martinfx:
Hello i bring support OpenBSD, NetBSD, DragonFly operating systems.
M.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -76,7 +76,16 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
             PLATFORM_OS=LINUX
         endif
         ifeq ($(UNAMEOS),FreeBSD)
-            PLATFORM_OS=FREEBSD
+            PLATFORM_OS=BSD
+        endif
+        ifeq ($(UNAMEOS),OpenBSD)
+            PLATFORM_OS=BSD
+        endif
+        ifeq ($(UNAMEOS),NetBSD)
+            PLATFORM_OS=BSD
+        endif
+        ifeq ($(UNAMEOS),DragonFly)
+            PLATFORM_OS=BSD
         endif
         ifeq ($(UNAMEOS),Darwin)
             PLATFORM_OS=OSX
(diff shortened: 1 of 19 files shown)


examples/Makefile
```diff
@@ -328,7 +328,7 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     ifeq ($(PLATFORM_OS),BSD)
         # Libraries for FreeBSD, OpenBSD, NetBSD, DragonFly desktop compiling
         # NOTE: Required packages: mesa-libs
-        LDLIBS = -lraylib -lGL -lpthread -lm
+        LDLIBS = -lraylib -lGL -lpthread -lm -lglfw
```
raysan5: GLFW shouldn't be required, now is compiled together with raylib in module [rglfw](https://github.com/raysan5/raylib/blob/master/src/rglfw.c).

Martinfx: ok

Martinfx:
@raysan5 Its this pull request ok ?

a3f:
@Martinfx Have you tested this on OpenBSD? CPANtesters has yet to report me that the Perl wrapper managed to build on an OpenBSD system. They seem to all fail because they don't find `<sys/soundcard.h>`. see e.g. http://www.cpantesters.org/cpan/report/ba78f6ce-63af-11e8-9a1a-c5a44eb4fde3

The test matrix for the most recent Alien-raylib release is here:
http://matrix.cpantesters.org/?dist=Alien-raylib%200.019

Martinfx:
@a3f No but this bug is in mini_al header and is need #include <soundcard.h> only

mackron:
I was just looking at BSD in mini_al just the other day! :)

I will be improving BSD support soon - it's the next thing on my list. I will be adding support for audioio.h and sndio. I think I'll also be restricting OSS to FreeBSD. Any other suggestions?

Martinfx:
@mackron  no :D

PR #558, opened 2018-06-23
Fisable glfw only when it is not found and not external.

yurivict:


a3f:
What is wrong with the old behavior? This breaks `USE_EXTERNAL_GLFW=IF_POSSIBLE` (Link against GLFW if installed systemwide, otherwise use the version embedded in Repository), which the Travis CI build currently supplies.

`find_package(glfw3 3.2.1 QUIET)` in utils.cmake should set `glfw3_FOUND` if it's available and requested (either by `USE_EXTERNAL_GLFW=YES` or `USE_EXTERNAL_GLFW=IF_POSSIBLE`), so I am not sure what this patch fixes.

yurivict:
The point is that disabling should be in sync wit its presence, and it wasn't.

yurivict:
For external glfw, you disable glfw when it is found.

PR #561, opened 2018-06-28
Fix incorrect log agent names

sir-pinecone:


raysan5:
Hi @sir-pinecone! Thank you very much for the review! Actually, I'm working in improved tracelog messages (more consiste and homogeneus) for all raylib! ;)

sir-pinecone:
No problem, Ray! And that's cool, looking forward to seeing what you come
up with. I recently did some work on my logger (which was inspired by
raylib btw) to have it write all printed output to a file. Is that
something you'd be interested in having?

PR #565, opened 2018-07-03
CMake: Fix regression in USE_EXTERNAL_GLFW=IF_POSSIBLE

a3f:
Fixes: e1e036a6 ("Disable glfw only when it is not found and not external", #558)


raysan5:
@a3f many thanks for the review!

Issue #567, opened 2018-07-05
[core] Support gamepad inputs for Android devices


SethArchambault:
I got the simple app template working on Android, but if I press any gamepad button, I get a segfault:

```
V/threaded_app(12555): New input event: type=1
V/threaded_app(12555): New input event: type=1
F/libc    (12555): Fatal signal 11 (SIGSEGV), code 1, fault addr 0x1b in tid 12574 (om.seth.project)
I/libc    (12555): Suppressing debuggerd output because prctl(PR_GET_DUMPABLE)==0
```

I've used asserts to try and locate the source of the error, but it seems like it's trying to jump to some part of the application that doesn't exist when a button is pressed, because I can't get an assert to be hit after loading, and before the crash.. Any clues here? 

I'm working on getting the output to not be suppressed, but unsure how to use:

```
prctl(PR_SET_DUMPABLE, 1)
```

Thanks!

raysan5:
Gamepad on Android? What device? Probably not implemented... but shouldn't be difficult...

raysan5:
ok, just checked raylib sources and also push a related commit here: https://github.com/raysan5/raylib/commit/1f1d8eeeeb5aa34cfa41784c88cdce11fa525fa1
Right now gamepad support is not implemented, it's not difficult to support it, just need to check input state and pressed buttons/asix but it will require some work.

I leave this issue open in the meantime.

SethArchambault:
K I've gotta take a break, but here's what I see so far.

``` 
line 3080
static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
```

Probably this line is breaking:
```
 gestureEvent.pointerId[0] = AMotionEvent_getPointerId(event, 0);
```

definition:
```
int32_t AMotionEvent_getPointerId(
  const AInputEvent *motion_event,
  size_t pointer_index
)
```

I suspect we'll need to getsource and then depending upon what that returns, we'll handle the type of input in it's own way:

```
int32_t AInputEvent_getSource(
  const AInputEvent *event
)
```
These are the options
```
AINPUT_SOURCE_ANY                   any
AINPUT_SOURCE_BLUETOOTH_STYLUS      bluetooth stylus
AINPUT_SOURCE_DPAD                  dpad
AINPUT_SOURCE_GAMEPAD               gamepad
AINPUT_SOURCE_JOYSTICK              joystick
AINPUT_SOURCE_KEYBOARD              keyboard
[...]

PR #570, opened 2018-07-06
SNES-like GAMEPAD Support added for Android, Makefile has optional support for 32bit

SethArchambault:
#567 

This is my first ever significant pull request! I love that that raylib is so easy to comprehend.

Note that in getting the gamepad feature to work, I had to abort before the gestures system could be handled. I tried some methods for ignoring gestures only when gamepad input was detected, however the system seems really glitchy (pointerCount seems to go all over the place), and I had to give up. Perhaps someone who is doing gesture stuff can step in here and try to re-enable it. 

The gamepad feature is really simple, all I had to do was capture the keycode when motion was detected, no axis needed for this one. 

I've tested this with Android, wasn't able to test elsewhere right now. Won't be offended if this is rejected for that reason! Was a great experience solving this myself, I'm so glad I'm using a code base that I can actually reason about!

Seth

### DIFF
--- a/release/include/raylib.h
+++ b/release/include/raylib.h
@@ -241,6 +241,24 @@
 #define GAMEPAD_XBOX_BUTTON_LEFT    13
 #define GAMEPAD_XBOX_BUTTON_HOME    8
 
+// 8BitDo Gamepad SNES CLASSIC
+#define GAMEPAD_SNES_DPAD_UP        19
+#define GAMEPAD_SNES_DPAD_DOWN      20
+#define GAMEPAD_SNES_DPAD_LEFT      21
+#define GAMEPAD_SNES_DPAD_RIGHT     22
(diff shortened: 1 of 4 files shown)


SethArchambault:
Oh, and I recognize the 32bit support may be unwanted.. Sorry!

SethArchambault:
Thinking on this more, instead of calling these enums like this:
GAMEPAD_SNES_DPAD_DOWN   
They should have the word Android in them like:
GAMEPAD_ANDROID_DPAD_DOWN
Since this is linked to the platform, rather than the specific controller..

raysan5:
Great work @SethArchambault! 

If you want to review some part you can add commits to this PR.

I'll review it more carefully tomorrow but it looks very good to me! :)

raysan5:
ok, I'm merging it and I'll review it.

PR #572, opened 2018-07-08
GAMEPAD Support: glfw silence keycode error

SethArchambault:
#567 
There was an "Invalid key" error showing up when using the gamepad key defines. This fixes that - not sure if there are unintended consequences related to GLFW here..


raysan5:
I think this PR could be more intended for GLFW than for raylib... but actually, standard keyboard keys lower than SPACE are not valid keys in most systems.

Sorry, I'm not integrating those changes.

SethArchambault:
@raysan5 Just to clarify - what this means is that anyone who uses android gamepad bindings will have their terminal made unusable for debuggings (They will be flooded with warning messages at least 60 times per second)

hmm.. I suppose the work around would be a Handmade Hero style separation of osx, android, and game code, which is probably the way to go anyhow - I'll pursue that route.

a3f:
> They will be flooded with warning messages at least 60 times per second

How about overriding the `stderr` output with `glfwSetErrorCallback`?

PR #573, opened 2018-07-09
Added "New" and "NewV" functions

NuklearBomb:
`New` and `NewV` functions have been added for creating `Vector2`, `Vector3`, `Quaternion`, and `Matrix`.

```
// Old way
Vector2 foo={0, 1}, bar={5, 3}, res;
res=Vector2Add(foo, bar);
```
```
// New way
Vector2 res;
res=Vector2Add(Vector2New(0,1), Vector2New(5,3));
```
This example is a bit simple, and more complex code would benefit the most from this update. This will definitely make it easier to write and maintain streamlined code.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -153,6 +153,20 @@ RMDEF float Clamp(float value, float min, float max)
 // Module Functions Definition - Vector2 math
 //----------------------------------------------------------------------------------
 
+// Create a vector
+RMDEF Vector2 Vector2New(float x, float y)
+{
+    Vector2 result = { x, y };
+    return result;
+}
+
+// Create a vector from a float array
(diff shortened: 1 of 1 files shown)


a3f:
What's wrong with
```
// C99 way
Vector2 res;
res=Vector2Add((Vector2){0,1}, (Vector2){5,3});
```

TheLumaio:
I was in the process of asking exactly that.
It's less code and you don't get the function call overhead

TheLumaio:
Also, isn't `Vector2 Vector2NewV(float arr[2])` invalid in C99?

NuklearBomb:
Wow, why didn't I think of that? Very cool, and way better! @a3f

PR #579, opened 2018-07-21
Convert window render sizes to floats

jessp:
GLSL 1.10 is typesafe ([PDF specs](https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.1.10.pdf), page 22), so this shader will not load properly during build triggering a type mismatch error.

It's not a super important change, but I came across it while playing with the examples on my pi.


raysan5:
Hey @jessp! Thank you very much for the improvement! I really care for that kind of details! :)

PR #589, opened 2018-07-24
Fix HTML5 examples and games using CMake

pamarcos:
I found that examples and games using CMake were not compiled correctly. The output was LLVM IR bitcode instead of html + js. I introduced the changes needed for their `CMakeLists.txt` mimicking what's done in the Makefiles.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -35,6 +35,7 @@ endforeach()
 
 include(CheckIncludeFiles)
 list(REMOVE_ITEM example_sources ${CMAKE_CURRENT_SOURCE_DIR}/others/standard_lighting.c)
+set(OUTPUT_EXT)
 
 if(${PLATFORM} MATCHES "Android")
   list(REMOVE_ITEM example_sources ${CMAKE_CURRENT_SOURCE_DIR}/others/rlgl_standalone.c)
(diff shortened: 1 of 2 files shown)


a3f:
Thanks!

a3f:
The `ARCH=html5` configuration takes 12 minutes to build now. Might be worth it to make html + js configurable (but still the default) and disable it from CI.

pamarcos:
Yeah, I also told @raysan5 before creating the PR that on my machine the compile time increased drastically. It makes sense, though. Before it was just running the frontend and generating LLVM IR while now it's actually generating WASM from there (hence, compiling for real).

Just out of curiosity, I set up `ccache` for `emcc` to see whether there was any improvement and indeed there is not. Most of the time is spent on the link phase, where `ccache` can do nothing. So I guess we're out of luck. We either have patience or will have to disable the job.

In any case, I think a 12min CI is still small enough. It does surprise considering this is C and the rest of the jobs take nothing, but it is how it is.

raysan5:
I'm think that's related to compiling code to be consumed by Emterpreter, instead of a plain WASM compilation... In any case, thank you very much for working in this improvement.

pamarcos:
I think it's ok to keep this as it is for the sake of simplicity and to avoid having to change all source code to proper code for Emscripten as per https://github.com/raysan5/raylib/blob/master/examples/core/core_basic_window_web.c

However, I think non-experienced users will have trouble compiling their games to web if we don't provide a CMake or Makefile that compiles using the right flags for production. They may wrongly use also this `Emterpreter`. At least, we could edit the wiki to clarify it.

raysan5:
Makefile that compiles using the right flags for production is provided in [raylib/templates](https://github.com/raysan5/raylib/tree/master/templates)

pamarcos:
My bad, I did not know about this. Indeed these are the right flags: https://github.com/raysan5/raylib/blob/master/templates/simple_game/Makefile#L189-L192

PR #591, opened 2018-07-25
Complete raw_audio_stream example

triplefox:
1. Always synthesize a complete frame of audio, using a second buffer (this prevents gaps in playback)
2. Sine is computed correctly, with an adjustable frequency
3. User can modulate frequency in real-time with mouse
4. Entire audio buffer data is shown, visually demonstrating how sine changes in wavelength

### DIFF
--- a/examples/audio/audio_raw_stream.c
+++ b/examples/audio/audio_raw_stream.c
@@ -15,8 +15,9 @@
 
 #include <stdlib.h>         // Required for: malloc(), free()
 #include <math.h>           // Required for: sinf()
+#include <string.h>         // Required for: memcpy()
 
-#define MAX_SAMPLES             22050
+#define MAX_SAMPLES             512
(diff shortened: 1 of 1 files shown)


raysan5:
I'm reviewing the example...

Just noticed some glitches on the drawing wave when changing between frequencies, maybe wave data should be cleaned when changing frequency... what do you think?

screenshot000

triplefox:
I debated adding a cleaning loop very briefly but thought that doing
without is a little more appealing in terms of  conveying the actual memory
as it gets updated. The five-year-old child in me says "oh, I could make
little pictures with this" - and the teenage student might say "oh, so if I
make it lower pitch there are actually more samples used?"

I have no objection to adding it if some visual polish is desired.

raysan5:
hahahaha... ok, leave it as it is! :D

Issue #595, opened 2018-07-25
[core] Support redirecting TraceLog to custom logging system


pamarcos:
As a user, I'd like to have full control of the logging system. e.g. I may want to log in both console and file. Or use my own format with timestamp, thread id, etc.

At the moment, `TraceLog` simply logs the `msgType` (aka `INFO`, `ERROR`, etc) + the text. While I could implement my own logging system for my stuff, raylib would still output everything just to the console. I'd like to sink everything raylib has to say into my own logging mechanism.

To do such thing, the easiest way I can think is using a callback (if previously set) that `TraceLog` calls instead of using its own implementation.

I have created a possible solution in ~15LOC in `utils.c`. The usage from the user point of view would be as follows:

```
void logCustom(int msgType, const char *text, ...)
{
    va_list args;
    va_start(args, text);
    vprintf(text, args);
    printf("\n");
    va_end(args);
}

 int main()
 {
    SetTraceLogCallback(logCustom);
...
```

raysan5:
Just for reference, I've checked [current Android log data redirection](https://github.com/raysan5/raylib/blob/master/src/utils.c#L126)... it just uses [Android log system](https://developer.android.com/ndk/reference/group/logging), but this system does not offer any redirection mechanism like the proposed one...

My main concern was callback usage... but I imagine it's ok... any opinions in the room?

a3f:
> To do such thing, the easiest way I can think is using a callback (if previously set) that TraceLog calls instead of using its own implementation.

Sounds good to me.

a3f:
Closed with #597.

raysan5:
thanks, I still want ti review some formating details on the PR...

a3f:
Ah, sorry for the premature merge...

raysan5:
hahaha... no worries! :D

PR #597, opened 2018-07-26
Add SetTraceLogCallback to enable users setting custom logging

pamarcos:
This is the PR to address https://github.com/raysan5/raylib/issues/595

The number of changes needed in `utils.c` are quite minimum. The only tricky part is that the callback is defined as: 

`typedef void (*TraceLogCallback)(int msgType, const char *text, void *args);`

where `args` is `void*` instead of `va_list` in order to avoid having to include `stdargs.h` since I saw that `raylib.h` did not previously include any other header and I figured out you wouldn't want to start including one now.

I also added an example called `core_custom_logging` to show users how it may be used. This is what the logger in the example looks like:
```
[2018-07-26 10:34:47] [INFO] : Initializing raylib 2.0
[2018-07-26 10:34:47] [INFO] : Display device initialized successfully
[2018-07-26 10:34:47] [INFO] : Display size: 1920 x 1080
[2018-07-26 10:34:47] [INFO] : Render size: 800 x 450
[2018-07-26 10:34:47] [INFO] : Screen size: 800 x 450
[2018-07-26 10:34:47] [INFO] : Viewport offsets: 0, 0
[2018-07-26 10:34:47] [INFO] : GPU: Vendor:   ATI Technologies Inc.
[2018-07-26 10:34:47] [INFO] : GPU: Renderer: AMD Radeon Pro 555 OpenGL Engine
[2018-07-26 10:34:47] [INFO] : GPU: Version:  4.1 ATI-1.68.20
[2018-07-26 10:34:47] [INFO] : GPU: GLSL:     4.10
[2018-07-26 10:34:47] [INFO] : Number of supported extensions: 46
```

Feel free to provide any feedback 😉


a3f:
> where args is void* instead of va_list in order to avoid having to include stdargs.h since I saw that raylib.h did not previously include any other header and I figured out you wouldn't want to start including one now.

I would be in favor of making it `va_list *` and including `<stdarg.h>` as it's just a standard C89 header and type safety is always good. Thoughts @raysan5?

A bit of bike shedding: `SetTraceLogger` has a nice ring to it.

pamarcos:
I checked with `gcc -E` to how much it would increase the header and at least in macOS the overhead is minimum:
```
# 1 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.1.0/include/stdarg.h" 1 3 4
# 30 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.1.0/include/stdarg.h" 3 4
typedef __builtin_va_list va_list;
# 48 "/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/9.1.0/include/stdarg.h" 3 4
typedef __builtin_va_list __gnuc_va_list;
```
I'd also bet on safety and add it, but since it was the first external header added I was quite reluctant to include it at first. It's up to you @raysan5

raysan5:
Well, as you know I'm a bit reluctant to adding headers to `raylib.h`... is it possible to do something like:
```c
typedef void (*TraceLogCallback)(int msgType, const char *text, ...);
```
If there is no other option I can consider adding it...

pamarcos:
I don't see any way to use `typedef void (*TraceLogCallback)(int msgType, const char *text, ...);` that would make it work. Once you have already created a `va_list` for your extra arguments, you cannot forward all of them to another function with `...`. You need to send the `va_list`. As far as I understand, that's precisely the reasoning behind having the `v` flavours for `vprintf` and the rest. 

In my opinion, if we had any other header that already includes some overhead to `raylib.h`, I'd go with including `stdargs.h` for sately. But since there's none and this is anyway probably intended for users who know what they're doing, I don't thing casting an opaque argument to `va_list*` is such a big issue.

a3f:
One could define `TraceLog` as a global function pointer, which is by default initialized to the normal `traceStdout` function, but can be overridden by the user...

raysan5:
Ok, let's include `stdargs.h`...

a3f:
Merged. Thanks!

PR #603, opened 2018-07-27
 Deal with failed LoadMusicStream

jubalh:
It would be good to have a way to continue execution even when loading a stream didn't work.
 Free music and set it to NULL in case something went wrong, so that users can check for it.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -1321,6 +1321,32 @@ float *GetWaveData(Wave wave)
 // Module Functions Definition - Music loading and stream playing (.OGG)
 //----------------------------------------------------------------------------------
 
+// Set music to NULL in case of error and print warning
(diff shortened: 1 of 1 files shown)


jubalh:
@a3f changed

raysan5:
Merging but I need to review this (maybe change it), I don't like at all that `format` parameter.

raysan5:
@jubalh just reviewed changes in commit https://github.com/raysan5/raylib/commit/dc3327fba8df91b252f911cca8b26621622d3488, finally I removed the additional function and moved code inside `LoadMusicStream()`. I prefer to avoid additional functions for something that specific.

raysan5:
Yeah, it's true but message is generic enough to let the user understand that something is wrong with the file he/she tries to load. I'll assume that minimal lack of information.

PR #604, opened 2018-07-27
Travis CI: Test off-by-default configurations as well

a3f:
Just noticed that `cmake -DUSE_OPENAL_BACKEND=ON` has been broken (by me probably)...
Therefore, test all configuration in CI to avoid this happening in future.


a3f:
Travis CI seems to disregard both Github push and pull request notifications...

Issue #592, opened 2018-07-25
[physac] physics_demo and physics_shatter crash when hitting R to reset example


pamarcos:
Both `physics_demo` and `physics_shatter` crash when hitting `R` to reset example. Sometimes `physics_shatter` gets creative instead of crashing directly. It ends up crashing, eventually

![image](https://user-images.githubusercontent.com/3248441/43196009-ef58af50-9006-11e8-8c6b-421919b58d22.png)

raysan5:
Maybe @victorfisac could take a look to this issue...

drezgames:
This happens because the code as it is, is not thread safe. You hit reset to clear the data while the background thread is updating data, you will get these random crashes. I use Object Pascal so for me, I had to disable threading on the C side and manage threading on the Pascal side. To make it work, I just put some mutex calls around the routines that change data. It seems to work and not crash anymore.

raysan5:
Maybe multithreading logic should be moved out of `physac` and let the user implement it on their end...

drezgames:
yea, maybe best thing because just 1 min ago, I got it working smooth, no crashes and less cpu usage. Up until just now I got the crashing fixed, but it was suffering from a little stuttering and it took a simple change to fix that. So now it seems to working good. In my threading loop, just need to do Sleep(1) to allow other threads to run better and that was it. Now no stuttering as the mutex is locked while destroying bodies.

a3f:
I can't reproduce the issue anymore. Did your PR fix it? Can this issue be closed?

pamarcos:
Yes, it was fixed by https://github.com/raysan5/raylib/pull/609. I will close it now

PR #609, opened 2018-07-29
[physac] Fix Physac examples to be run without creating new thread

pamarcos:
This PR addresses https://github.com/raysan5/raylib/issues/592

To start with, both `physics_demo` and `physics_shatter` had a bug because they were never setting `needsReset` back to `false`.

Then, as discussed in https://github.com/raysan5/raylib/issues/592, I thought it would be much simpler to avoid `physac` creating any threads rather than making the library thread-safe. I feel it's much simpler for beginners and newcomers to simply call `RunPhysicsStep` themselves within their main loop rather than learning concurrency all of the sudden. For this reason, I believe `PHYSAC_NO_THREADS` should be the default. I would actually rename it to do the opposite: `PHYSAC_CREATE_THREAD` or similar.

I had to fix refactor `PhysicsLoop` slightly and created `PhysicsRunStep` to be able to run `PhysicsStep` when being called from the user end since it relied on the timing values collected there.

I'd appreciate very much if @victorfisac reviewed this since it's his creation and I'm also modifying his library within this PR. Should I open a different PR if this gets accepted in https://github.com/victorfisac/Physac to keep them in sync or is @raysan5 doing that from time to time?

### DIFF
--- a/examples/physac/physics_demo.c
+++ b/examples/physac/physics_demo.c
@@ -17,6 +17,7 @@
 #include "raylib.h"
 
 #define PHYSAC_IMPLEMENTATION
+#define PHYSAC_NO_THREADS
 #include "physac.h"
 
 int main()
@@ -54,6 +55,8 @@ int main()
         // Update
         //----------------------------------------------------------------------------------
         // Delay initialization of variables due to physics reset async
+        RunPhysicsStep();
+
         if (needsReset)
         {
             floor = CreatePhysicsBodyRectangle((Vector2){ screenWidth/2, screenHeight }, 500, 100, 10);
@@ -61,6 +64,8 @@ int main()
 
             circle = CreatePhysicsBodyCircle((Vector2){ screenWidth/2, screenHeight/2 }, 45, 10);
             circle->enabled = false;
+
+            needsReset = false;
         }
 
         // Reset physics input
@@ -134,4 +139,3 @@ int main()
 
     return 0;
 }
-
--- a/examples/physac/physics_friction.c
+++ b/examples/physac/physics_friction.c
@@ -17,6 +17,7 @@
 #include "raylib.h"
 
 #define PHYSAC_IMPLEMENTATION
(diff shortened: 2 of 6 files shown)


raysan5:
Merging this PR. @victorfisac could take a look later...

Just notice that avoiding physics to run in a second thread, we loose some resolution on physics because they are only calculated in a per-frame basis.

But I understand we let the user to control any additional thread if required.

victorfisac:
Physics needs to be dependent to framerate in order to achieve it's objective: dynamics simulation. From my opinion, trying to calculate physics from main thread is like driving a car with squared wheels.

But, physac library inside raylib can be modified as anybody want in order to fit raylib style (simple, fast, minimalist); if you feel it's better in main thread let's keep in this way.

I have my own repository with official Physac version where I will implement thread safe.

pamarcos:
Well, the changes done don't imply you are calling it always within your main loop (which makes physac framerate-dependant). It simply allows you to do it from any thread you want, instead of letting physac creating a new one for you. In other words, there's nothing that prevents you from starting a new thread that calls `RunPhysicsStep` 60 times (or whatever value you may deem valid) per second.

In the context of the examples since they are all set to run at 60 FPS, I thought it was worth calling `RunPhysicsStep` from the main loop directly to avoid having to implement thread-safety on physac. I did it after assuming the examples have room enough to run the logic + physics and after seeing that `PHYSAC_DESIRED_DELTATIME` default value was `1/60`.

raysan5:
@victorfisac I agree with your arguments, it's not good to calculate physics in main thread, I noticed it in my previous comment, but I also agree that allowing final user to have control over physics steps is a good option. 

C is not intrinsically a multithreaded language, for that reason, multiple threading libraries could be found depending the underlying hardware, letting the users choose their own threading mechanism could be a good option for the library.

victorfisac:
@raysan5 @pamarcos Ok! Now I understand. I agree with this :)

PR #610, opened 2018-07-29
[games] Prevent GLFW changing working dir to 'Resources'

pamarcos:
GLFW was changing the working directory to `Resources` by default. We should prevent this in order to run successfully the games examples that load stuff from `resources/[whatever]`.


raysan5:
Thanks @pamarcos!

PR #612, opened 2018-07-30
fix problem when compiling with MSVC 2017, error C2177

ethan-tqa:
(I'm not sure if the built-in CMake makes any difference)

The old FLT_MAX constant seems to be too big for MSVC, resulting in error C2177. I could have just used as defined in float.h that comes with MSVC. But I thought why not just go all the way and use the bit-by-bit correct constant as defined here: https://stackoverflow.com/questions/16350955/interpreting-the-bit-pattern-of-flt-max

### DIFF
--- a/examples/models/models_mesh_picking.c
+++ b/examples/models/models_mesh_picking.c
@@ -13,7 +13,7 @@
 #include "raylib.h"
 #include "raymath.h"
 
-#define FLT_MAX     340282346638528859811704183484516925440.0     // Maximum value of a float, from bit pattern 01111111011111111111111111111111
+#define FLT_MAX     340282346638528859811704183484516925440.0f     // Maximum value of a float, from bit pattern 01111111011111111111111111111111
 
 int main()
 {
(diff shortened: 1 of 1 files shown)


a3f:
> I could have just used as defined in float.h that comes with MSVC.

`<float.h>` is guaranteed to exist in any conforming C implementation, so that would be the proper way.

> But I thought why not just go all the way and use the bit-by-bit correct constant as defined here

This only applies to IEEE 754 single-precision binary floating-point `FLT_MAX`. While raylib might never run on a system with another type of `float`, we can at least dream. :-)

Could you change it to `#include <float.h>` instead?

ethan-tqa:
I have converted the constant for float.
I suppose that the author originally didn't want to include the whole float.h just for one constant, so he copy pasted the line there. Although I said that it is bit-by-bit correct, the constant is defined using a regular number in source code so it should work everywhere.

raysan5:
Just merged it.

PR #618, opened 2018-08-05
Fix compiler warnings

kimkulling:
Thanks a lot for the great library. I just fixed some compiler warnings which are caused by forat to int, int to float or double to float casts, which are not explicitely marked as explicit casts.


raysan5:
Thanks for the review @kimkulling! :)

kimkulling:
You are welcome! :-)

PR #620, opened 2018-08-12
Update mini_al to 0.8.5.

mackron:
I just thought I would do a quick update to mini_al because I've just fixed a bad bug in the ALSA backend which should significantly improve things on Raspberry Pi. This should resolve the random crackling issues and significantly improve latency.

It does not fix that glitch at the start of streaming - that's happening because the time it takes to render the first frame is longer than the time it takes to play the audio buffer which results in a situation where the audio buffer is unable to get fed with data quickly enough.

This should also fix the audio_raw_stream example.

You should probably also give this a quick test on the other backends before merging.


raysan5:
Thanks for the update @mackron! I'll check it asap! :)

PR #622, opened 2018-08-13
Shapes: Simplifies CheckCollisionRecs

Joefish:
By comparing the edges of the rectangles relative to each other we can
determine if they intersect or not.


raysan5:
Thanks for the improvement!

PR #623, opened 2018-08-13
Games: Fixes bugs in Snake

Joefish:
This PR fixes two bugs in the game Snake. The fruit spawn position when initially colliding with the snake was causes it to be rendered off-grid and the collision detection of the snake's head and the fruit.

Changes were tested on Linux 4.17.12-arch1-1-ARCH


raysan5:
Thanks for the fix! :)

PR #625, opened 2018-08-16
Raymath dllexport fix

ChrisDill:
- Added __declspec(dllexport) to RMDEF in raymath.h. This allows them to be accessed when importing from raylib.dll.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -60,7 +60,11 @@
 #endif
 
 #ifdef RAYMATH_IMPLEMENTATION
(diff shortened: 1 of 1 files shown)


raysan5:
It seems it fails on CI system...

ChrisDill:
Im using MSVC with the visual studio project for the dll to import into c#.  I added it so I could bind raymath. The same export is used in raylib.h for RLAPI although that also checks for _WIN32 so that might be it.

raysan5:
merged!

Issue #627, opened 2018-08-19
[core] Missing key definitions for plus/minus keys?


justinclift:
Trying out the Go binding for raylib, and hit the problem of there being no definitions for the plus/minus keys on the keypad.

https://github.com/raysan5/raylib/blob/506b7b8d7c2e367d61ee92a77b54fb115a566173/src/raylib.h#L104-L182

It turns out they're not here either. :wink:

I can make a PR adding them (here, then in raylib-go), but I'm not sure where the numeric values are coming from.  They don't seem to be keyboard scan codes, which was my first thought.

Ideas? :smile:

justinclift:
Ahhhh, glfw looks to be the source:
```
$ grep -ri "KEY_SPACE" *
external/glfw/include/GLFW/glfw3.h:#define GLFW_KEY_SPACE              32
external/glfw/src/cocoa_init.m:    _glfw.ns.keycodes[0x31] = GLFW_KEY_SPACE;
external/glfw/src/mir_init.c:    _glfw.mir.keycodes[KEY_SPACE]      = GLFW_KEY_SPACE;
external/glfw/src/input.c:    if (key < GLFW_KEY_SPACE || key > GLFW_KEY_LAST)
external/glfw/src/input.c:    if (key < GLFW_KEY_SPACE || key > GLFW_KEY_LAST)
external/glfw/src/wl_init.c:    _glfw.wl.keycodes[KEY_SPACE]      = GLFW_KEY_SPACE;
external/glfw/src/win32_init.c:    _glfw.win32.keycodes[0x039] = GLFW_KEY_SPACE;
external/glfw/src/win32_init.c:    for (key = GLFW_KEY_SPACE;  key <= GLFW_KEY_LAST;  key++)
external/glfw/src/x11_init.c:        case XK_space:          return GLFW_KEY_SPACE;
raylib.h:#define KEY_SPACE            32
```

So, further looking shows this:

#### external/glfw/src/x11_init.c:
```
        case XK_KP_Subtract:    return GLFW_KEY_KP_SUBTRACT;
        case XK_KP_Add:         return GLFW_KEY_KP_ADD;
        ...
        case XK_minus:          return GLFW_KEY_MINUS;
        case XK_equal:          return GLFW_KEY_EQUAL;
```

and:

#### external/glfw/include/GLFW/glfw3.h
```
#define GLFW_KEY_MINUS              45  /* - */
...
#define GLFW_KEY_EQUAL              61  /* = */
...
[...]

PR #628, opened 2018-08-19
Add the plus and minus keys on both the main keyboard and keypad

justinclift:
For #627. :smile:

Actually... thinking about it more, should this be updated to include all of the keypad keys as well?


raysan5:
Thanks! :)

justinclift:
Whoops.  I was in the middle of updating...

PR #629, opened 2018-08-19
Add the remaining numeric keypad keys, and a few other missing ones

justinclift:
Follows on from #628. :smile:

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -133,21 +133,45 @@
 #define KEY_F10             299
 #define KEY_F11             300
 #define KEY_F12             301
-#define KEY_KP_SUBTRACT     333
-#define KEY_KP_ADD          334
 #define KEY_LEFT_SHIFT      340
 #define KEY_LEFT_CONTROL    341
 #define KEY_LEFT_ALT        342
+#define KEY_LEFT_SUPER      343
 #define KEY_RIGHT_SHIFT     344
 #define KEY_RIGHT_CONTROL   345
 #define KEY_RIGHT_ALT       346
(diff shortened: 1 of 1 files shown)


justinclift:
k, now it's good.  Just had to rename one of the new key definitions (copied from GLFW), as an existing android key already took it's name.  Fixed now. :wink:

justinclift:
Looking at that Travis CI failure, it seems to be a problem with the Travis infrastructure during that run:

```
0.00s$ mkdir build
0.10s$ $RUNNER $CC --version
aarch64-linux-android-clang: command not found
The command "$RUNNER $CC --version" failed and exited with 127 during .
```

Is it feasible to re-run the check?

justinclift:
Ahhh, I can probably just re-push the commit with a slightly changed commit date.  That should trigger a re-run. :smile:

raysan5:
Merged! Thanks for the improvement! :)

@justinclift I've seen you've trying the VR Simulator in Go, do you think same improvements are required here?

justinclift:
Just looked at it... it'd probably make sense to do so.  The screen width/height info is in the `VRDeviceInfo` structure anyway so might as well use it. :smile:

That'd probably also make it easier for people to switch to a new device too.  eg just change the one it's using, which will then automatically grab the relevant bits.

Want me to make a PR?  It's a pretty simple change. :smile:

raysan5:
Let me check it more carefully before the PR... :)

justinclift:
Yeah, it's probably something you could do easily, and likely better. :smile:

PR #633, opened 2018-08-21
lerp for vector2 and float

unequaled86:


raysan5:
@unequaled86 thanks for the improvement!

PR #635, opened 2018-08-25
core: Support slashes as well in GetFileName & GetDirectoryPath

### LINKED ISSUE #634: [core] `GetFileName` and `GetDirectoryPath` do not support Unix paths
When using Unix style paths with `GetFileName` we get incorrect results, and in the case of `GetDirectoryPath` we get a SIGSEGV crash.

After looking at the source I noticed you're doing a search through the string for `\` characters used on Windows. Is it possible to check `defined(_WIN32)` and use the current functionality, but on other platforms search for `/` instead?

Also, the crash in `GetDirectoryPath` will also occur when there are no slashes in the provided path.

Below is example code demonstrating the issue and the crash:
```c
#include <stdio.h>
#include "raylib.h"

int main()
{
    const char *fileName = GetFileName("C:\\hello\\test.c");
    printf(fileName);
    printf("\n");
    // Prints: test.c

    const char *fileNameUnix = GetFileName("/hello/test.c");
    printf(fileNameUnix);
    printf("\n");
    // Prints: /hello/test.c
    // Should print: test.c (on Unix platforms)

    const char *directory = GetDirectoryPath("C:\\hello\\test.c");
    printf(directory);
    printf("\n");
    // Prints: C:\hello

    const char *directoryUnix = GetDirectoryPath("/hello/test.c");
    // Program crashes here with:
    // terminated by signal SIGSEGV (Address boundary error)
    printf(directoryUnix);
    printf("\n");
    // Should print: /hello (on Unix platforms)

    return 0;
}
```

a3f:
Fixes #634.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1346,11 +1346,20 @@ const char *GetExtension(const char *fileName)
     return (dot + 1);
 }
 
+/* "string pointer reverse break": return right-most occurrence of charset in s */
+static const char *strprbrk(const char *s, const char *charset)
+{
+    const char *latest_match = NULL;
+    for(; s = strpbrk(s, charset), s != NULL; latest_match=s++)
+        ;
+    return latest_match;
+}
+
 // Get pointer to filename for a path string
 const char *GetFileName(const char *filePath)
 {
-    const char *fileName = strrchr(filePath, '\\');
-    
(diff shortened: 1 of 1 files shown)


raysan5:
Great! Thanks for the fix! I'm reviewing formating...

PR #637, opened 2018-08-30
Updated VSCode project to support OSX building / debugging.

MurrayIRC:
Previously the project was insufficient for running raylib on OSX. This will be the first official project that actually supports building, running, and debugging on the platform.

There are some things to keep in mind:
- This project requires that the user specify the path to their installation of raylib in a few places. It might be a good course of action to work towards having a similar situation to the Windows side of things, where we can be more certain of the direct path to raylib on OSX.
- The wiki needs to be updated to explain better how to run raylib on OSX. Perhaps directing people to use VSCode and linking that project is the best course of action?


raysan5:
@MurrayIRC many thanks for the update! good job! :)

PR #646, opened 2018-09-21
Linux Support for VSCode Project

MurrayIRC:
I'll need to update the wiki to supplement these changes, so that's on the way :)


raysan5:
Great addition! Thanks! :)

PR #650, opened 2018-09-27
Added monitor functions

ChrisDill:
### Description

I added a few useful functions for managing monitors. Currently they use glfwGetPrimaryMonitor but they could be changed to index the array of monitors by taking in a index instead if that is preferred.

### Example
```
    int count = GetMonitorCount();
    int width = GetMonitorWidth();
    int height = GetMonitorHeight();
    int physicalWidth = GetMonitorPhysicalWidth();
    int physicalHeight = GetMonitorPhysicalHeight();
    const char *name = GetMonitorName();
```

Also does this pull request include my merge commit? I did that to update my branch. Is there a way to remove that and what method would you recommend for syncing a branch?


a3f:
> ### Description
> they could be changed to index the array of monitors by taking in a index instead if that is preferred.

I think that would be worthwhile. Would avoid breaking API in future if it's requested.
 
> Also does this pull request include my merge commit?

Yes, it does.

> Is there a way to remove that

`git rebase -i HEAD~6`, and then you can delete the merge commit. You could also squash the tabs commit into the previous one.

> what method would you recommend for syncing a branch?

You can do so with `git pull --rebase` instead, which stacks your commits on top of the origin's HEAD. I find it to be the saner default, you can configure it as default with `git config --global pull.rebase true`.

ChrisDill:
It would mean the user has to track which monitor to use but it does give more control for when you need it. Could also add a function to get the primary monitor index aswell.

Thanks for the help @a3f. I think I will leave it for now but I will keep that in mind for next time!

a3f:
> It would mean the user has to track which monitor to use but it does give more control for when you need it. Could also add a function to get the primary monitor index aswell.

We could set aside `0` or `PRIMARY_MONITOR` or the like to always mean the default monitor.

> Thanks for the help @a3f. I think I will leave it for now but I will keep that in mind for next time!

I was mistaken. You would have to use something older than HEAD~6, because you would have to reorder the commits manually (There's no merge commit to remove...), I could do it at merge if you'd like if @raysan5 approves the pull request.

ChrisDill:
PRIMARY_MONITOR could work although if the primary monitor changes while running it would be wrong. Unlikely to happen but something to consider.

If you could do it at merge that would be good! Don't want to mess git history with unnecessary merge commits thats all.

a3f:
> PRIMARY_MONITOR could work although if the primary monitor changes while running it would be wrong. 

The GLFW documentation says

> The primary monitor is always first in the array returned by glfwGetMonitors.

So, it's safe to always pass `0` to mean primary monitor. If the primary monitor changes while running, GLFW automatically rearranges them appropriately:

https://github.com/glfw/glfw/blob/0be4f3f75aebd9d24583ee86590a38e741db0904/src/monitor.c#L91

raysan5:
I agree passing an `index` it's the better way, being 0 the primary monitor. Please, could you implement that change? After that we can merge.

raysan5:
Great work! Thanks! :)

PR #653, opened 2018-10-04
[physac] Fix physac's fixed time step

pamarcos:
This PR aims to fix some details that https://github.com/raysan5/raylib/pull/609 left unsolved. When I gave support for running physac from main thread I assumed it already had the logic to handle the physics step being called with different time deltas. This doesn't seem to be the case. Please @victorfisac take a look and correct me if I am wrong. This is what I've found in the original implementation within the physics loop that looks wrong:

```
PHYSACDEF void RunPhysicsStep(void)
{
    // Calculate current time
    currentTime = GetCurrentTime();

    // Calculate current delta time
    deltaTime = currentTime - startTime;

    // Store the time elapsed since the last frame began
    accumulator += deltaTime;

    // Clamp accumulator to max time step to avoid bad performance
    MathClamp(&accumulator, 0.0, PHYSAC_MAX_TIMESTEP);

    // Fixed time stepping loop
    while (accumulator >= PHYSAC_DESIRED_DELTATIME)
    {
        PhysicsStep();
        accumulator -= deltaTime;
    }

    // Record the starting of this frame
    startTime = currentTime;
}
```

1. `deltaTime` which is used for the physics formulas is measured in milliseconds. This comes from the substraction of two values taken from `GetCurrentTime()`, which returns milliseconds. My guess is that this method previously returned seconds and hence the confusion.
2. `accumulator` was increased with `deltaTime` (ms) but compared against `PHYSAC_MAX_TIMESTEP` (1.0/60.0=0.016, hence seconds). So, since `deltaTime` in ms was more than likely to be > 0.016 (seconds), `MathClamp` was basically making sure that `accumulator` always had the `PHYSAC_MAX_TIMESTEP` value (0.02, in seconds?). In the end, this mechanism basically was making that the fixed time step loop was always executed at least once and only once. Only once because even if the `accumulator` value was clamped, we were subtracting `deltaTime`(remember, in ms) from it. So, `accumulator` would take a negative number.
3. The result of this is that `PhysicsStep` was called always once with a `deltaTime` in ms that was quite small since there's a loop in a dedicated physics thread that it just runs this. The easiest way to prove that the fixed time step loop is not fixed at all is by introducing an `usleep` call with different values in the physics thread after each `RunPhysicsStep()` call.

So, these changes just make sure that the fixed time step is actually fixed and it runs in consistent steps of `deltaTime` value. I consolidated all time values to ms since it's what was used the most. I have found that `deltaTime` value needs to be quite small for the physics simulation to be accurate. It even turns out that for one example (`physics_restitution`, we even need to decrease the time step for it to perform well. However, if we used such a tiny value to run physics for every demo, it would perform very badly with multiple objects due to the number of collisions that need to be checked. Hence, I decided to use a default value that works well for all demos and create a new API called `SetPhysicsTimeStep` to set another value for the `physics_restitution` example.

I took some quick gifs so that you can see the difference. Take into account that I exported them with 10 FPS to avoid bloating this page, so the quality is not very good.

Before
demo_bad
After
demo_good

Before
friction_bad
After
friction_good

Before
restitution_bad
After
restitution_good

### DIFF
--- a/examples/physac/physics_restitution.c
+++ b/examples/physac/physics_restitution.c
@@ -51,6 +51,9 @@ int main()
     circleC->restitution = 1;
     
     SetTargetFPS(60);
+
+    // Restitution demo needs a very tiny physics time step for a proper simulation
+    SetPhysicsTimeStep(1.0/60.0/100 * 1000);
     //--------------------------------------------------------------------------------------
 
     // Main game loop
--- a/src/physac.h
+++ b/src/physac.h
@@ -96,8 +96,6 @@
 #define     PHYSAC_MAX_VERTICES             24
 #define     PHYSAC_CIRCLE_VERTICES          24
 
-#define     PHYSAC_DESIRED_DELTATIME        1.0/60.0
-#define     PHYSAC_MAX_TIMESTEP             0.02
 #define     PHYSAC_COLLISION_ITERATIONS     100
 #define     PHYSAC_PENETRATION_ALLOWANCE    0.05f
 #define     PHYSAC_PENETRATION_CORRECTION   0.4f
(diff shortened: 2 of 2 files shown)


victorfisac:
giphy

AMAZING! :D

gen2brain:
Physics in raylib-go behave almost the same as in your screenshots, i.e. restitution amount, so I think I can reuse your fix there. Thanks!

Issue #497, opened 2018-02-26
[models] models_material_pbr less shiny


a3f:
Is this how it's supposed to look like? 
<img width="912" alt="screen shot 2018-02-26 at 13 16 28" src="https://user-images.githubusercontent.com/5204368/36670088-5afc85f4-1af7-11e8-8202-98fc20e76ab1.png">

I compared it with a5881fb9ccaf60ea0ff4f6afd7f3925e24330cac and there the size is still the same but it has that violet shine to it:
<img width="912" alt="screen shot 2018-02-26 at 13 24 17" src="https://user-images.githubusercontent.com/5204368/36670406-84342304-1af8-11e8-8a6f-9aee454e5fd6.png">

This on macOS High Sierra
```
./models_material_pbr
INFO: Initializing raylib (v1.9.4-dev)
INFO: Trying to enable MSAA x4
INFO: Display device initialized successfully
INFO: Display size: 1680 x 1050
INFO: Render size: 800 x 450
INFO: Screen size: 800 x 450
INFO: Viewport offsets: 0, 0
INFO: GPU: Vendor:   Intel Inc.
INFO: GPU: Renderer: Intel Iris OpenGL Engine
INFO: GPU: Version:  4.1 INTEL-10.30.14
INFO: GPU: GLSL:     4.10
INFO: Number of supported extensions: 45
INFO: [EXTENSION] DXT compressed textures supported
INFO: [EXTENSION] Anisotropic textures filtering supported (max: 16X)
INFO: [EXTENSION] Debug Marker supported
INFO: [TEX ID 1] Texture created successfully (1x1 - 1 mipmaps)
INFO: [TEX ID 1] Base white texture loaded successfully
INFO: [SHDR ID 1] Shader compiled successfully
INFO: [SHDR ID 2] Shader compiled successfully
INFO: [SHDR ID 3] Shader program loaded successfully
INFO: [SHDR ID 3] Default shader loaded successfully
INFO: [CPU] Default buffers initialized successfully (lines, triangles, quads)
INFO: [VAO ID 1] Default buffers VAO initialized successfully (lines)
INFO: [VAO ID 2] Default buffers VAO initialized successfully (triangles)
INFO: [VAO ID 3] Default buffers VAO initialized successfully (quads)
INFO: OpenGL default states initialized successfully
INFO: [TEX ID 2] Texture created successfully (128x128 - 1 mipmaps)
INFO: [TEX ID 2] Default font loaded successfully
[...]

PR #655, opened 2018-10-07
core: workaround window not being rendered till moved on macOS Mojave

a3f:
Apple ought to fix their OpenGL implementation, but with OpenGL now
deprecated this might not happen so soon.

This has been reported upstream in GLFW in glfw/glfw#1334.
The workaround comes from kovidgoyal/kitty@b82e74f

This also fixes the HiDPI (Retina) scaling issues reported in #497,
so the workaround is enabled anywhere \_\_APPLE\_\_ is defined.


raysan5:
@a3f thanks for the fix! :)

PR #656, opened 2018-10-10
Physac.h fix for variable array size declaration.

noshbar:
Generating the projects using CMake, targeting VS2017, results in an error when compiling.
This is due to physac.h trying to make a 'vertices' array of size 'int count', making it const does not work, either.

This changes the static declaration to a malloc/free combo.

Tested using the physics-demo.


raysan5:
Hey @noshbar! Many thanks for all the improvements! Amazing! :D

noshbar:
No worries at all!

I'm finally at point where I can contribute to projects I like and use (and
I love raylib), so there will be a few more on their way in the future.

Thanks for everything!

raysan5:
That sounds fantastic! You're welcome! 👍

PR #657, opened 2018-10-10
3 minor sample game cleanups:

noshbar:
* changed bitwise & to logical &&
* removed redundant state check
* moved a strcpy inside a NULL check to prevent a crash.
  NOTE: nothing sensible will happen as a result, but no crash at least.


raysan5:
Actually most of those games deserve a full review! Most of them have been created in GameJams too fast and can have bugs... Many thanks for the reviews! :)

noshbar:
They're useful for showing general use, and I was running static analysis
on a new feature I have planned, and these cropped up.

So not a full review, but the worst of the problems gone, hopefully!

PR #659, opened 2018-10-10
Fixes memory leak in text.c

noshbar:
GenImageFontAtlas() allocates an array of stbrp_rect for the packing functions, but it never frees them.


raysan5:
yes, good catch!

PR #661, opened 2018-10-12
Added GetLastWriteTime to allow for file reloading

ChrisDill:
Added a function to get the last write time of a file. I used this so I can reload files or resources if the time since they were last loaded changes.

### Example
```
char *path1 = "test1.txt";
char *path2 = "test2.txt";

long mod1 = GetLastWriteTime(path1);
long mod2 = GetLastWriteTime(path2);
TraceLog(LOG_INFO, "last write time %ld %ld", mod1, mod2);
```


raysan5:
Added, I'm changing the function name to `GetFileModTime()`.

Issue #667, opened 2018-10-16
Build error on Mac with master


fabslab:
Cloning master and running `make PLATFORM=PLATFORM_DESKTOP` on Mac currently results in:
```
clang -c core.c -O1 -Wall -std=c99 -D_DEFAULT_SOURCE -Wno-missing-braces -Werror=pointer-arith -fno-strict-aliasing -Werror=implicit-function-declaration -DUSE_OPENAL_BACKEND -I. -Iexternal -Iexternal/glfw/include -DPLATFORM_DESKTOP -DGRAPHICS_API_OPENGL_33
In file included from core.c:131:
external/dirent.h:63:10: fatal error: 'io.h' file not found
#include <io.h>         /* _findfirst and _findnext set errno iff they r...
         ^~~~~~
1 error generated.
make: *** [core.o] Error 1
```

dj0nes:
Same. 

I'm running MacOS Mojave, have Xcode, xcode tools installed with `xcode-select --install`, and installed the system headers using 

`open /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg`

as suggested in [this post](https://stackoverflow.com/a/52530212).

Any ideas? I am new to the C ecosystem.

raysan5:
Hey! Actually, `external/dirent.h` include should only be included if compiling with MSVC compiler, it's not required in any other case... it seems checking macros are not working properly, don't know why...

As a temporary solution, you can just remove the include.

jubalh:
The same happens on linux.

This is because of the -I option in the Makefile.

`external/dirent.h` is only included in the C file if MSCV, but `INCLUDE_PATHS = -I. -Iexternal -Iexternal/glfw/include` in the Makefile makes the compiler build everything nevertheless.

dj0nes:
Removing `-Iexternal` from [this line](https://github.com/raysan5/raylib/blob/master/src/Makefile#L362) of the makefile works for me. I can compile and run raylib programs after.

I'd be happy to submit a PR putting that include behind a platform flag, if that sounds correct.

Side note: Running `make` does give a ton of "X is deprecated: first deprecated in macOS 10.14 - OpenGL API deprecated" warnings, as well as some unused functions and variables.

a3f:
> Side note: Running make does give a ton of "X is deprecated: first deprecated in macOS 10.14 - OpenGL API deprecated" warnings

[...]

PR #669, opened 2018-10-18
Remove non-Windows includes from Makefile

dj0nes:
Removed -Iexternal from the base include path, added it back conditionally for Windows only. This allows non-Windows platforms to compile raylib as discussed in https://github.com/raysan5/raylib/issues/667.


raysan5:
Hey @dj0nes! Thanks for the improvement! :D

PR #674, opened 2018-10-21
Added RaspberryPi option to ignore duplicate touchscreens (Default on)

Berni8k:
Found out my mouse/touch overhaul can have issues with users of "tslib" as it can pick up both the virtual and real touchscreen device at the same time.

Raylib now by default only uses a single touchscreen device. If there is more than one then the device with the highest number takes priority (This is where tslibs virtual touch device will end up in most use cases). This functionality can be turned off by setting "#define USE_LAST_TOUCH_DEVICE false"

Mouse and other devices are unaffected by this.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -226,6 +226,7 @@
     #define DEFAULT_KEYBOARD_DEV      STDIN_FILENO              // Standard input
     #define DEFAULT_GAMEPAD_DEV       "/dev/input/js"           // Gamepad input (base dev for all gamepads: js0, js1, ...)
     #define DEFAULT_EVDEV_PATH        "/dev/input/"             // Path to the linux input events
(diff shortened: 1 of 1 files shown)


raysan5:
Hey @Berni8k, added a pair of comments to the code, I can review it myself if you prefer.

raysan5:
@Berni8k Comments are on the changed files tab...

Berni8k:
@raysan5 
Ah yes now i see it. Yes i agree i should have used a macro if. 

I had first intended this to be selectable trough the API but then changed my mind half way. I will fix once i get back home.

Berni8k:
@raysan5 
Ah yes now i see it. Yes i agree i should have used a macro if. 

I had first intended this to be selectable trough the API but then changed my mind half way. I will fix once i get back home.

PR #675, opened 2018-10-24
VSCode Project Updates

MurrayIRC:
Included in this pull request are changes that make the VSCode simpler for people to pick up and work with. I plan on spending some more time with the wiki entry for VSCode as well to make the process for setting up raylib much more straightforward.

Updates:
- consolidated build and run tasks into a singular task for each. this occurred when i realized i didn't need to create a separate task for each platform.
- added support in the makefile for a src/ and obj/ file hierarchy that includes subfolders. i will update the VSCode tutorial to reflect this change.


raysan5:
Thanks for the update @MurrayIRC!

PR #683, opened 2018-11-07
Update core.c

iwoplaza:
Some projects need the buffer to not be empited when drawing to the render texture. I would suggest making maybe a variation of the function if you'd like to keep backwards compatibility, maybe like: BeginTextureModeNoClear() ?

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1170,7 +1170,9 @@ void BeginTextureMode(RenderTexture2D target)
 
     rlEnableRenderTexture(target.id);   // Enable render target
 
-    rlClearScreenBuffers();             // Clear render texture buffers
+    // Some projects need the buffer to not be empited when drawing to
+    // the render texture.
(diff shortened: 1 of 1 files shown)


raysan5:
Need to think how to manage this... probably removing `rlClearScreenBuffers()` is the best option.

Just out of curiosity, are you working on any project that requires avoiding buffers cleaning?

iwoplaza:
My friend started working on a project in C, and wanted to use your raylib library as a simple way to draw pixels on the screen with instructions sent over by serial. He wanted to draw to a RenderTexture, so that he could rescale it dynamically. Because each instruction is NOT a re-draw of the whole screen, we need to keep the last state of the RenderTexture, and draw on top of that :).

raysan5:
@iwoplaza ok, I see, thanks for the feedback!

Just let me think a bit more about it... I also need to check all involved code in the repo for a safe redesign.

Issue #686, opened 2018-11-12
Unsafe OpenURL()


jubalh:
Unfortunately `OpenURL()` can still be misused.
If we call the function with our own string all is fine, but if we let a user input an URL to open there is a risk.
A user could could craft the string to already contain the quoting and then run another command afterwards resulting in:

```
OpenURL("https://www.raylib.com/' ; 'rm -rf /");
```

For this reason we should only call this function if we control the string given (for opening our game website) or we need to parse the URL and escape the quotes.

raysan5:
Added small security check in commit https://github.com/raysan5/raylib/commit/618f220851570f2bb9ea0bb354a65e92c6d06968

jubalh:
A solution would be to replace all `'` in the URL. Then all the characters like '&' etc would be okay to use because they would be part of the URL and one could not end it and add another command after the call of `xdg-open` (etc.)

jubalh:
I think it's good like this.

PR #687, opened 2018-11-12
Add warning to OpenURL()

jubalh:
See https://github.com/raysan5/raylib/issues/686


raysan5:
Thanks for the note @jubalh! :)

PR #688, opened 2018-11-13
Check for single apostrophe in OpenURL()

jubalh:
When doing a8dffc63fbe3926498ecb905428f454d0afbe526 I was not aware that
printing a warning and not executing the code would be an option. I only
learned that through 618f220851570f2bb9ea0bb354a65e92c6d06968.

So I propose that we allow all URLs except if the string contains a `'`.
Which could end the URL and call another command via `system()`.

Related to https://github.com/raysan5/raylib/issues/686


raysan5:
Thanks for the improvement!

PR #703, opened 2018-12-20
Prefixing raylib macros with `RL_`.

MarcoLizza:
At the moment, in my opinion, *raylib* is a bit too "invasive" in not using a proper prefix for the API.

As a first step in this direction, I'm suggesting using the `RL_` prefix for the macros (the most common case of identifier clash).

In the future, I'd like to see that also typedefs and function are prefixed with a proper API prefix (being this `RL_ ` or `RayLib_`).


raysan5:
Sorry but I have no plans to add that prefix.

PR #711, opened 2019-01-02
Added SetMouseOffset

ChrisDill:
Needed mouseScale to be seperate for x and y so it works when the width and height are different.
For example works with a size of 500x500 but not when using 1024x600.

Solved by addding `SetMouseOffset` to set the scale with a Vector2.
Default scale is 1.0f for both x and y and `SetMouseScale` sets both aswell.

### Example
```c
Vector2 size = { GetScreenWidth(), GetScreenHeight() };
Vector2 offset = { 1920.0f / size.x, 1080 / size.y };
SetMouseOffset(offset);
```

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -333,7 +333,7 @@ static int defaultKeyboardMode;                 // Used to store default keyboar
 
 // Mouse states
 static Vector2 mousePosition;                   // Mouse position on screen
-static Vector2 mouseScale = { 1.0f };               // Mouse default scale
+static Vector2 mouseScale = { 1.0f, 1.0f };     // Mouse default scale
 static bool cursorHidden = false;               // Track if cursor is hidden
 static bool cursorOnScreen = false;             // Tracks if cursor is inside client area
 static Vector2 touchPosition[MAX_TOUCH_POINTS]; // Touch position on screen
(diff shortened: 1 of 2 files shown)


raysan5:
Actually, proposed `SetMouseOffset()` was intended to add and additional offset to mouse position reads.

For two dimensions scaling, `SetMouseScale()` could be just changed to receive a `Vector2 scale`.

ChrisDill:
Oh okay!

I could use a mouseOffset global in mouse reads like so
```c
return (Vector2){ mousePosition.x+mouseOffset.x*mouseScale.x, mousePosition.y+mouseOffset.y*mouseScale.y };
```
Is that what you mean?

raysan5:
Yes, that's what I mean! :)

Actually, I'm already using something similar on fullscreen mode.

ChrisDill:
I found `renderOffsetX` and `renderOffsetY`. Can use that and maybe change to a Vector2 if it is used in the same way?

raysan5:
Sorry, I think it's not possible, those are used for letterbox screen rendering when needed.

ChrisDill:
Added and updated my example above. Order of operations needed fixing.

raysan5:
I'm reviewing this change, probably modyfing something...

PR #717, opened 2019-01-08
Add project/scripts

neonmoe:
Some people might want to build projects without make systems or such, so here's three build scripts I wrote to build raylib applications without those. Only the three main desktop platforms sadly, but they were the easiest to write and test. I've ran them all successfully on Windows 8.1, Arch Linux (gcc) and a pretty up-to-date macOS, but additional testing would probably be nice if possible.


raysan5:
Great addition @neonmoe! Thanks! :)

PR #719, opened 2019-01-10
Window visibility

MarcoLizza:
Hello! Just a proposal, since I think it's a quick and nice addition. I've added a visibility control for the main windows. That way, the window can be initially created as "hidden" only to be later shown when repositioned/resized. This would solve the issue I'm encountering when trying to auto-scale a windowed application to fit the display while preserving the required area. That is, you ask for a 320x240 window and it is resized to 960x720 automatically (with a scale factor x3 that will preserve the pixels).

What do you think?


raysan5:
Hey! That's a good addition! Maybe I'd change proposed functions to:
```c
// Window-related functions
RLAPI void ShowWindow(void);                              // Shows window
RLAPI void HideWindow(void);                              // Hides window
RLAPI bool IsWindowHidden(void);                          // Check if window is not visible
```
for consistency with:
```c
// Cursor-related functions
RLAPI void ShowCursor(void);                              // Shows cursor
RLAPI void HideCursor(void);                              // Hides cursor
RLAPI bool IsCursorHidden(void);                          // Check if cursor is not visible
```

MarcoLizza:
That sounds appropriate. I'll review the code and send a new PR...

raysan5:
Thank you very much @MarcoLizza! :)

raysan5:
Nice addition! :D

src/raylib.h
```diff
@@ -415,8 +415,9 @@ typedef enum {
     FLAG_WINDOW_RESIZABLE   = 4,    // Set to allow resizable window
     FLAG_WINDOW_UNDECORATED = 8,    // Set to disable window decoration (frame and buttons)
     FLAG_WINDOW_TRANSPARENT = 16,   // Set to allow transparent window
-    FLAG_MSAA_4X_HINT       = 32,   // Set to try enabling MSAA 4X
-    FLAG_VSYNC_HINT         = 64    // Set to try enabling V-Sync on GPU
+    FLAG_WINDOW_HIDDEN      = 32,   // Set to create the window initially hidden
```
a3f: Some people prefer to use raylib as shared library and to avoid complicating
their raylib experience, one should be wary of subtle ABI breaks.
With this change, an old program linking against a new shared raylib would have the window hidden instead of MSAA 4x!
It would've been better to use `FLAG_WINDOW_HIDDEN = 128` here, you could reorder it to have the FLAG_WINDOW*s grouped together

raysan5: @a3f I agree, despite being a small aesthetic change I could be very annoying, I'll revert it.

raysan5: Reverted in commit https://github.com/raysan5/raylib/commit/25205b5113587af71af0f39549c9bb0a3c1faf49

PR #722, opened 2019-01-14
Added ColorFromHSV()

Demizdor:
Implemented from https://en.wikipedia.org/wiki/HSL_and_HSV#Alternative_HSV_conversion. 

When converting back and forth from **Color** to **HSV** some precision will be lost and the initial color won't be exactly the same but this cannot be avoided. I put a note in the source!

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1445,6 +1445,41 @@ Vector3 ColorToHSV(Color color)
     return hsv;
 }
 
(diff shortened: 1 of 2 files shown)


raysan5:
hey @Demizdor! Thanks for the addition! I'd just remove alpha parameter, it could be set directly by user to returned value...

PR #723, opened 2019-01-14
Trace log failing on other

MarcoLizza:
Here it is, a small fix that issue the `LOG_OTHER` log-type not properly handled.


raysan5:
Thanks for the review! :)

PR #731, opened 2019-01-20
Reorganizing logging

MarcoLizza:
Hello Raysan!

Here's my proposal for the logging facility "rework", in order to be more similar to the standard ones.

I also added an API to enable/disable the "exit-on-error" feature.

Hope this suits you fine!

Marco

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -175,7 +175,7 @@ typedef enum {
     LOG_TRACE,
     LOG_DEBUG,
     LOG_INFO,
-    LOG_WARN,
+    LOG_WARNING,
     LOG_ERROR,
     LOG_FATAL,
     LOG_NONE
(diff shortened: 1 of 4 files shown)


raysan5:
@MarcoLizza Warning tracelog messages seem to fail when additional parameters are required.
The output I get now does not show the additional filename parameter...
```
WARNING: [] Image fileformat not supported
WARNING: [] Image could not be loaded
WARNING: [] Image fileformat not supported
WARNING: [] Image could not be loaded
```

raysan5:
`SetTraceLogLevel(LOG_NONE);` outputs `LOG_WARNING` traces, with no line break...

raysan5:
Pushed some reviews but not working properly yet...

PR #735, opened 2019-01-24
Fix clang++ compilation errors

RobLoach:
clang was complaining about the type conversions. For example...
```
src/core.c:1888:15: error: cannot initialize a variable of type 'char *' with an
      rvalue of type 'void *'
        char *cmd = calloc(strlen(url) + 10, sizeof(char));
              ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -276,7 +276,7 @@ static bool windowMinimized = false;            // Check if window has been mini
 static const char *windowTitle = NULL;          // Window text title...
 
 static unsigned int displayWidth, displayHeight;// Display width and height (monitor, device-screen, LCD, ...)
-static int screenWidth, screenHeight;           // Screen width and height (used render area)
+//static int screenWidth, screenHeight;           // Screen width and height (used render area)
 static int renderWidth, renderHeight;           // Framebuffer width and height (render area, including black bars if required)
 static int renderOffsetX = 0;                   // Offset X from render area (must be divided by 2)
 static int renderOffsetY = 0;                   // Offset Y from render area (must be divided by 2)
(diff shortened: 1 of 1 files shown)


raysan5:
Thanks for the review @RobLoach! :)

a3f:
I don't  agree with this. raylib is a C, not a C++ library. Compile the library as C and use the headers in your C++ program if you want to use it.

In this particular case, explicit casting in C for calloc is frowned upon, because of stdlib weren't included and the compiler considers lack of prototypes just a warning, C assumes the return type an int, which shouldn't automatically convert to a pointer, but now that you explicitly cast it, you've forced it.

And C99, which raylib requires, is non-compatible with C++ anyway, so we won't ever support a standard C++ compilers.

RobLoach:
In C we shouldn't be explicit casting? I've seen warnings in GCC too.

a3f:
We should cast when we need to, not when a C++ compiler requires it because it uses different type conversion rules. No C compiler would've warned about this instance there, because it was perfectly fine idiomatic C code.

raysan5:
Sorry @a3f, I don't see the problem with those casts, actually, I always cast to the specific pointer type... unless the pointer is `void *` by design.

You're right that C standard does not require the cast  and raylib is a C library but supporting it being compiled by a C++ compiler could be useful.

Just as a reference for me, [more info here](https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc).

a3f:
> You're right that C standard does not require the cast and raylib is a C library but supporting it being compiled by a C++ compiler could be useful.

The thing is C99 and C++ are incompatible, raylib won't ever be able to compile with a standard C++ compiler. My opinion is simply that supporting *compilation* of raylib in a non-standard C++ dialect isn't a good reason to modify perfectly fine C code.

If it's C++ code, use a C++compiler. If it's C code, use a C compiler and link your C++ against it.

raysan5:
Ok, I agree, it shouldn't be changed just to support being compiled as C++ (C is not C++, I know)... but at least for keeping consistency with all the other `malloc` used by raylib is a point.

a3f:
>Ok, I agree, it shouldn't be changed just to support being compiled as C++ (C is not C++, I know)...

Which is what I wanted to get at.

> but at least for keeping consistency with all the other malloc used by raylib is a point.

Well, I think those are redundant too but ye, consistency is a valid reason. :)

PR #748, opened 2019-02-10
Added Trim functions

Skabunkel:
Added TextTrim function
Added TextRightTrim function
Added TextLeftTrim function
Fixed spelling mistakes


jubalh:
Usually infinite is to be preferred for commit messages. You did it right the first time ;)

raysan5:
I was thinking about adding those functions and I skipped them... do you have some usage case? could the same be accomplished with some other `Text*()` function?

I'm a bit reticent on adding functions that require manual memory free() from the user... couldn't those functions use an internal static buffer?

PR #749, opened 2019-02-10
LoadBMFont filepath fix

Syphonx:
Fixed as issue where `strrchr` in `LoadBMFont` would only look for forward slashes, instead of backslashes causing `strlen` to fail on a null string formatted like so:

> R"(..\Resources\fonts\example.fnt)"


raysan5:
@Syphonx thanks for the review

Issue #752, opened 2019-02-13
[textures] Disabling image export in config.h fails to compile


neonmoe:
I noticed that if I comment out the line following line in config.h, the compilation will fail with multiple undefined references in the ExportImage function.
```c
#define SUPPORT_IMAGE_EXPORT        1
```

After further investigation, a similar problem arises when disabling the following:
- SUPPORT_FILEFORMAT_KTX
- SUPPORT_IMAGE_MANIPULATION
- SUPPORT_IMAGE_GENERATION
- SUPPORT_MESH_GENERATION

Are these just wip, or am I missing something?

I could write a PR for these as I'll be writing a fix for myself in any case. If that's desired, should the relevant functions (eg. ExportImage) be completely removed when compiled with these settings off, or should the functions just not do anything?

raysan5:
@neonmoe Thanks for reporting! It shouldn't fail but in the last months raylib has changed a lot and those flags were not tested. A PR is welcome. About ExportImage() it should be available but just enable supported file formats.

Issue #555, opened 2018-06-09
[textures] Review/rewrite ImageDraw() function 


raysan5:
[`ImageDraw()`](https://github.com/raysan5/raylib/blob/master/src/textures.c#L1279) function does a blitting of one image over another image. Useful for 2D software rendering.

```c
void ImageDraw(Image *dst, Image src, Rectangle srcRec, Rectangle dstRec);
```
That function is crucial for lots of other functions (`ImageTextEx()`, `ImageDrawTextEx()`, `ImageResizeCanvas()`, `ImageDrawRectangle()`) and it should be **optimized** to consider all side cases correctly:

 - [x] srcRec out of bounds (srcImage)
 - [x] dstRec resizing (if required)
 - [x] dstRec out of bounds (dstImage)
 - [x] dstRec negative positioning -> correct `for` loop instead of cropping

EDIT: Image blitting could be greatly improved using `memcpy()` calls to move chuncks of pixels from one side to another instead of pixel-to-pixel moving but it's ok for now.

raysan5:
Code snippet:
```c 
// Faster Image blitting implementation
void ImageDraw(Image *dst, Image src, Rectangle srcrec, Rectangle dstrec)
{
    int dstBufferWidth = dst->width;
    int srcBufferWidth = src->width;
    bool singleMemcpy = (dstrec.x == 0) && (dstBufferWidth == srcBufferWidth);
    
    dstBufferWidth *= 4;    //dstBufferWidth <<= 2;     // 4 bytes (RGBA)
    srcBufferWidth *= 4;    //srcBufferWidth <<= 2;     // 4 bytes (RGBA)
    
    unsigned char *dstDataPtr = (unsigned char)dst->data;
    dstDataPtr += (dstrec.y*dstBufferWidth);
    unsigned char *srcDataPtr = (unsigned char)src->data;
    
    if (singleMemcpy) memcpy(dstDataPtr, srcDataPtr, src->width*src->height);
    else
    {
        dstDataPtr += (destrec.x*4);
        
        while (srcrec.height--)
        {
            memcpy(dstDataPtr, srcDataPtr, src->width);
            
            dstDataPtr += dstBufferWidth;
            srcDataPtr += srcBufferWidth;
        }
    }
}
```

raysan5:
`ImageDraw()` reviewed on commit https://github.com/raysan5/raylib/commit/2433f30b4b81030c6c3fd6cc17ee394403208fe0

[...]

PR #755, opened 2019-02-14
Fix config.h flags

neonmoe:
Related to issue #752.
This PR currently contains fixes to most of the configuration flags. I'm making this PR now since I need some advice on how to continue.

Issues when `SUPPORT_IMAGE_MANIPULATION` is disabled:
- `LoadTextureCubemap` calls `ImageDraw`, which is now undefined. `ImageDraw` then seems to depend on `ImageCrop` and `ImageResize`, and `ImageResize` depends on `stbir_resize_uint8`. Should `LoadTextureCubemap` depend on `SUPPORT_IMAGE_MANIPULATION`, or should something more sophisticated be done?
- `ImageFormat` calls `ImageMipmaps`, which is now undefined. Is it safe to just not call `ImageMipmaps` if this flag is not enabled?
- `ImageAlphaCrop` calls `ImageCrop`, I'm guessing it should be under this flag as well.

A few notes on the changes I made already:
- If I understood correctly, GenImageFontAtlas is only used internally, and only used with TTF fonts. Based on this, and the fact that it exploded with errors once the flag was disabled, I made it depend on the `SUPPORT_FILEFORMAT_TTF` flag.
- Moved GenImageColor out of the `SUPPORT_IMAGE_GENERATION` block, since it's called from four different places outside of it, so modifying the callers would probably cause more problems.

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -655,12 +655,16 @@ Mesh LoadMesh(const char *fileName)
     TraceLog(LOG_WARNING, "[%s] Mesh fileformat not supported, it can't be loaded", fileName);
 #endif
 
+#if defined(SUPPORT_MESH_GENERATION)
     if (mesh.vertexCount == 0)
     {
         TraceLog(LOG_WARNING, "Mesh could not be loaded! Let's load a cube to replace it!");
         mesh = GenMeshCube(1.0f, 1.0f, 1.0f);
     }
     else rlLoadMesh(&mesh, false);  // Upload vertex data to GPU (static mesh)
+#else
+    rlLoadMesh(&mesh, false);  // Upload vertex data to GPU (static mesh)
+#endif
 
     return mesh;
 }
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -767,7 +767,11 @@ Wave LoadWave(const char *fileName)
 {
     Wave wave = { 0 };
 
+#if defined(SUPPORT_FILEFORMAT_WAV)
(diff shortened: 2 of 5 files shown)


raysan5:
Hi @neonmoe! Thanks for the improvements and the notes! :)

About your comments:
 - I agree that `SUPPORT_IMAGE_MANIPULATION` could be a bit tricky, when first added that set of functions, I wasn't expecting to be using them much around... `LoadTextureCubemap()` was added some weeks ago, not properly tested yet and it's a bit special function because it requires image manipulation to work properly... yeah, it probably should depend on `SUPPORT_IMAGE_MANIPULATION`.
 - About `ImageDraw()`, I really think that function could be improved: https://github.com/raysan5/raylib/issues/555
 - About `ImageMipmaps()`, yes, it's safe to just not call this function if flag `SUPPORT_IMAGE_MANIPULATION` is not enabled.
 - Actually, probably all `ImageAlpha*()` should be under the flag... don't remember now why I left those functions outside... probably there was some reason... never mind...

neonmoe:
Moved `LoadTextureCubemap` and `ImageAlphaCrop` under the `SUPPORT_IMAGE_MANIPULATION` block, and made the `ImageMipmaps` in `ImageFormat` depend on the flag as well. All config flags should be working properly now (at least, when I tried each one individually, they compiled).

raysan5:
@neonmoe Thank you very much for the review! :)

PR #757, opened 2019-02-17
Raw keyboard for PLATFORM_RPI

DarkElvenAngel:
I've already set the define symbol to use the Raw keyboard but this can be removed and moved to the makefile, without the symbol defined we simply use legacy mode and read the keys from STDIN.

I have tested some of the examples and it seems to work well.


raysan5:
@DarkElvenAngel Thank you very much for the improvement, I'll review it as soon as possible.

Do we really need `keycodes.h`? I'd prefer not adding an extra file just for it an RPI.

DarkElvenAngel:
No I'm sure there is a way to avoid it's use. I've use it to do  the key code conversions it's simple a array structure with raw code, scan code, and debug key_names. I'm sure it can be streamlined and included in core.c

raysan5:
@DarkElvenAngel I'm reviewing the cod and I have two questions: 

 - When using `input_event`, why do we need to reed rawcodes/scancodes? Is not possible to directly read keycodes and let the system do the mapping for us? As per my understanding, that mapping would only work for some set of keyboards... doesn't it?
 - Why using a separate `EventKeyBoardThreadSpawn()`, won't it be possible to use the already available `EventThreadSpawn()`? It's ok to have them separated, just wondering...

DarkElvenAngel:
@raysan5 

1. The reason I did that was to match the keycodes to the ones used in the raylib.h.  From what I see in `linux/input.h` is that also loads the defines from `linux/input-event-codes.h` this has all the keycode defines and they did not match the scancodes defined in raylib.h so I had to translate between the two.  The extra information I have added in the structure can be downsized to just an array with each index being the rawcode with the translated scancode as it's value.  As for if this mapping is the same for all keyboards I honestly have no idea, however it would seem likely it should be since I'm getting my keycodes from the linux kernel source libraries. However I'm just not sure if that's the case.

 2. I separated the two functions to make it easier to read, and debug when there was a problem.  `EventThreadSpawn()` could be used alone it should work just fine I don't see why it wouldn't.

Thank you for reviewing my code.

src/core.c
```diff
@@ -25,6 +25,8 @@
 *   #define PLATFORM_RPI
 *       Windowing and input system configured for Raspberry Pi i native mode (no X.org required, tested on Raspbian),
 *       graphic device is managed by EGL and inputs are processed is raw mode, reading from /dev/input/
+*   #define USERAWKEYBOARD
```
RobLoach: The other naming conventions use `PLATFORM_RPI`, with an underscore between the words. Should it be `USE_RAW_KEYBOARD`? Could even go one step further and adopt the `SUPPORT_` prefix.

raysan5: Yes, that's right, actually, still thinking what to do with that flag... considering no exposing it directly to the user and default to the new keyboard input system... actually old one had some bugs...

Berni8k:
Im the original author of the raw event input mode for mouse/touch. I have considered keyboard and gamepad support for it but left it out as i thought the existing keyboard method was good enough. I guess i was wrong and people do want better keyboard support.

Upon discussing it with @raysan5 we came to a decision that i should merge this keyboard code into my event handler since its sharing most of the code anyway. Thanks for the code contribution.

Berni8k:
@raysan5 @DarkElvenAngel 
Merge into the existing event handler done and committed as #785

raysan5:
Input system for RPI has been unified with previous implementation in PR #785. Closing this one.

In any case, thank you @DarkElvenAngel and @Berni8k for looking into it! :)

PR #761, opened 2019-02-22
Fixed height bug in DrawTextRecEx()

Demizdor:
Small error that prevented rendering the text inside a rectangle with sufficient height.
Sorry i didn't catch this earlier!


raysan5:
Hey @Demizdor! Thank you very much! :)

PR #762, opened 2019-02-23
Fix audio pitch

ftk:
When calling:
```
SetSoundPitch(sound, 0.8);
// ...
SetSoundPitch(sound, 0.9);
```
expected pitch of 90% of the original sound, but instead got 0.8*0.9=72%
This PR changes the behavior to set pitch relative to the base sound.


raysan5:
Thanks for the addition @ftk!

PR #763, opened 2019-02-23
add SetSoundLoop

ftk:


raysan5:
Sorry, I'm not a fan of this kind of functions, no function in raylib is used to set a `true/false` value.

PR #767, opened 2019-02-24
Fixed segmentation fult created by quick fix

Skabunkel:
My quick fix introduced errors when unloading fonts not using the data field, Do´h.
This should fix that error.


Skabunkel:
_quietly facepalms in the distance_

Skabunkel:
i'm going to cry myself to sleep now.

raysan5:
@Skabunkel thanks for the fix! no worries! you did a great job detecting this issue! :)

src/text.c
```diff
@@ -686,7 +688,8 @@ void UnloadFont(Font font)
     {
         for (int i = 0; i < font.charsCount; i++)
         {
-            free(font.chars[i].data);
+            if(font.chars[i].data != NULL)
```
a3f: This check isn't necessary. `free(NULL)` is always a no-op.

raysan5:
@a3f Thanks for pointing this out! Actually, I was doing that same check in some places in raylib... just removing it.

PR #785, opened 2019-03-28
RaspberryPi Keyboard input with evdev

Berni8k:
Based on pull request from user "DarkElvenAngel", better integrated with the current event system and enhanced with buffer to help with fast typing at low framerates.


raysan5:
Thanks @Berni8k! Great addition! :)

PR #795, opened 2019-04-09
Fixed incompatible pointer warnings

bjornwah:
Fixed a couple of warnings where pointer to unsigned int is used but pointer to size_t is expected.

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -775,7 +775,7 @@ void ExportMesh(Mesh mesh, const char *fileName)
 Material *LoadMaterials(const char *fileName, int *materialCount)
 {
     Material *materials = NULL;
-    unsigned int count = 0;
+    size_t count = 0;
     
     // TODO: Support IQM and GLTF for materials parsing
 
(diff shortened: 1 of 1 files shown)


raysan5:
Thanks for the PR @bjornwah! Sorry but I'm not merging this change, I tried to avoid using `size_t` in all raylib... In any case, glad to see you managed to send it perfectly!

a3f:
@raysan5 `tinyobj_parse_obj` doesn't care though. It will write a `size_t` anyway regardless of the pointer you pass in and if `size_t` is bigger than `unsigned int`, like on x86_64, it will corrupt the stack.

This needs to be fixed, either by patching `tinyobj_parse_obj` to use unsigned int, or by using the standard `size_t` type where appropriate.

raysan5:
Following `tinyobj_loader` comment:
```c
@todo { Remove stddef dependency. size_t? }
```
Replaced `size_t` by `unsigned int` in commit https://github.com/raysan5/raylib/commit/802afe8fe533bc66a6ebd2810090b08d1f635991.

PR #798, opened 2019-04-12
Use typedef rather than #define in order to avoid issues in applicati…

bjornwah:
The typedef is nicer because it is scope-aware and does not globally deny the use of the 'Camera' typename. User code may very well have a Camera class/struct inside a namespace, for example.


bjornwah:
Failed build seems unrelated?

raysan5:
@bjornwah thanks for the improvement and the explanation!

PR #801, opened 2019-04-13
Complete Hidpi support

Silverbeard00:
This should complete HIDPI support in Raylib with the inclusion of the previously committed GLFW flag.

Tested on Linux and Windows with both HIDPI and Regular monitors.  

Windows actually works really well.  I kept changing the DPI and the Raylib window would automatically update while it was running.


Silverbeard00:
Retest was showing some issue with Windows.  I gotta check if it's because Windows gets all iffy when you change the dpi without relogin in.

PR #812, opened 2019-04-23
Fix for DrawRectangleRounded

Demizdor:
Added required texture coordinates for `DrawRectangleRounded()` and `DrawRectangleRoundedLines()`...also added a screenshot for the `text_unicode` example.


raysan5:
Great @Demizdor! Thanks! :D

PR #817, opened 2019-04-24
Fixed text selection in DrawTextRecEx()

Demizdor:
Since the unicode patch, text selection didn't work as expected (see below), so i fixed it(easy fix :D).
comp

Working on raygui textbox now so i'll probably find all the bugs related to drawing unicode text, if there are any... Lucky me!


raysan5:
Thanks @Demizdor! Great work! 👍

PR #819, opened 2019-04-26
UWP Support Overhaul

Rover656:
I have added better support for UWP in both core, utils and added a c++ header which shows how the new UWP Messages interface works, this can be used for most applications, but the API is there for custom implementations if necessary.

# Complete list of things done
- Keyboard, Mouse and Gamepad Input
- Added a single header for managing connection between C++ and C
- Cursor functions all work like they do on other platforms
- Fixed a bug where UWP content would be stuck to the bottom of the window
- Added a hack which fixes "flickering keys". When holding a key without the hack enabled, it will flicker as if the user was pressing it repeatedly.
- Removed VS2015 support, I cannot update it because I don't have the toolkits

Appologies for the double commits, I had some merge issues when pulling master.


raysan5:
Some points to review:

 - I'm using 4-spaces convention on raylib instead of TAB... here the [code conventions](https://github.com/raysan5/raylib/wiki/raylib-coding-conventions).
 - I try to avoid `PLATFORM` dependant code on `rlgl.h`, maybe we can use znear/zfar to avoid collision and `RL_FREE(extList)` shouldn't be an issue (it probably needs some review)...
 - `examples/core/core_basic_window.cpp` shouldn't be modified

raysan5:
Sorry @Rover656, something happened on `core.c` and all my defines and comments alignments were changed as long as other formatting elements...

Rover656:
Hold up, I shall revert the change and try to fix this.

raysan5:
Thanks @Rover656! Great addition! :D

PR #821, opened 2019-04-28
Gamepad Rework

Rover656:
This reworks gamepads so that all button and axis mappings go through the same enum, instead of controller specific controls.

Desktop and UWP Gamepads have been tested and work correctly. Desktop (on windows, if GLFW mapping works here, it should work everywhere else) has been tested with an XB1 and PS4 controller, and UWP just with the XB1.

Web has not been tested as I am unable to test, but what I have written comes straight from the w3 docs and so should be correctly mapping buttons and axis to the new enum's aswell.


Rover656:
Just noticed all the examples need updating... I shall get on that.

raysan5:
Ok, just answered two of my three questions (TABS, gamepad examples).

Third question: What about Android gamepad? I just been investigating it right now and it seems gamepad inputs on Android come in the form of key inputs... not sure how to deal with that properly... neither have an Android gamepad for testing... probably we can skip that part for now.

Rover656:
I have ignored Android gamepads for now as I did not find them implemented and I am unsure as to how to add Android support to it.

raysan5:
ok, do you want me to wait for the examples review before merging?

Rover656:
Yeah please, I am just in the middle of updating the gamepad example to the correct enums, but nothing more. My aim is for it to build and the example to work, not worrying about controller identification for the moment.

Rover656:
That should be it sorted, I'm going to wait for the build to complete before giving the go-ahead.

raysan5:
take it easy, no hurries... :)

raysan5:
It seems there is some issue on `PLATFORM_WEB`:
```
/src/src/core.c:3488:57: error: use of undeclared identifier 'k'
                const GamepadAxis axis = GetGamepadAxis(k);
```

Rover656:
Oh, I see, 2 seconds

Rover656:
That should fix it

raysan5:
@Rover656 Thnak you very much for another great contribution! :)

PR #824, opened 2019-05-03
Add SUPPORT_HIGH_DPI to CMakeOptions.txt

kawa-yoiko:
Hi! ^ ^

This change allows the recently-added option to be toggled in CMake. The wiki page "CMake Build Options" needs to be updated as well.

On a side note, is it desirable that such options default to different values depending on the platform, e.g. `SUPPORT_HIGH_DPI=ON` on macOS (as almost all recent models have a Retina display), and `SUPPORT_BUSY_WAIT_LOOP=OFF` on systems with `nanosleep()`? Because with current master 0b2bad4 and default value `SUPPORT_HIGH_DPI=OFF`, the macOS build draws on only 1/4 of the window.

asteroids


raysan5:
Thanks @kawa-yoiko! Yes, I agree with forcing HighDPI on macOS, added on commit https://github.com/raysan5/raylib/commit/f44888e466ad63eccec229bdcc091f8f7e2f629e

PR #832, opened 2019-05-08
Remove sqrt() from CheckCollisionSpheres()

ProfJski:
Square root calls are computationally expensive.  In this function, it can be avoided.  Instead of checking whether distance<(radiusA+radiusB), we can check whether distance squared < (radiusA+radiusB) squared.

I used the existing Vector functions, but perhaps the code is less readable that way.  The Vector3DotProduct of centerB-centerA with itself gives the distance squared.


raysan5:
Thanks for the improvement @ProfJski! :)

I'll review it to include previous solution commented (just as reference for learners).

PR #833, opened 2019-05-10
Fixed DrawTextRecEx() selection when wordwrap is ON

Demizdor:
My earlier patch https://github.com/raysan5/raylib/pull/817 failed to account for wordwrap.


raysan5:
Thanks @Demizdor! :)

PR #835, opened 2019-05-12
Added julia set shader example.

eggmund:
Feel free to change anything. I don't usually program in C so some parts are probably a bit gross. I feel line 93 - 121 in `shaders_julia_set.c` could be done a lot better, but I was not sure myself.


raysan5:
@eggmund Thanks for this great addition! 👍

eggmund:
Thank you for making programming it so easy! Glad I could contribute 😃

PR #838, opened 2019-05-15
Update julia set example.

eggmund:
Zoom increase/decrease should now scale depending on the current zoom, and the zoom should also now behave as expected (increased zoom = closer image).


raysan5:
@eggmund nice! thanks! 😀

PR #842, opened 2019-05-18
Load glTF

wbrbr:
This PR adds support for the glTF 2.0 file format.

Some features:
* works with .gltf and .glb
* works with embedded (base64) or external textures
* loads the albedo/diffuse texture
* works with multiple mesh per model and multiple primitives per model

Some restrictions (not exhaustive):
* only triangle meshes
* it doesn't support node hierarchies or transforms
* it only loads the diffuse texture. It wouldn't be too hard to make it load other textures (normal, roughness/metalness) if you want to.
* it only supports unsigned short for indices (no byte/unsigned int) and float for texture coordinates (no byte/unsigned short). All the models I have seen loaded correctly but it might still be an issue

Sponza: (there are some  rendering artifacts but it loads correctly)
screenshot000

Duck:
screenshot001

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -60,6 +60,7 @@
 #if defined(SUPPORT_FILEFORMAT_GLTF)
     #define CGLTF_IMPLEMENTATION
     #include "external/cgltf.h"         // glTF file format loading
+    #include "external/stb_image.h"
 #endif
 
 #if defined(SUPPORT_MESH_GENERATION)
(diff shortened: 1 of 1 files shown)


raysan5:
@wbrbr this is an amazing addition! Thank you vey much! I'm merging it as soon as possible! :D

Leinnan:
How about anims?

raysan5:
@Leinnan supporting animations requires some extra work... maybe in a future...

PR #849, opened 2019-05-22
Fix unitialized struct member bug (#848)

### LINKED ISSUE #848: [examples] models_animation sig fault
when compiled from cmake makefiles it just seg faults without running, however running from a real Makefiles it does run showing just a bunch of animated red boxes and segs on exit

 is the invalid pointer because something couldn't be loaded ? - suggestion have your own version of free that zero's pointers on free and doesn't free null pointers.... (but does log attempt)
You get a less useful and short backtrace if the animation isn't run before quit.

### Environment

```
uname -a
Linux xps-void 4.19.44_1 #1 SMP PREEMPT Fri May 17 11:52:04 UTC 2019 x86_64 GNU/Linux
```
```
cat /proc/cpuinfo | grep "model name" | tail -n1
model name	: Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
```
```
glxinfo | grep OpenGL | grep string
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) HD Graphics 5500 (Broadwell GT2) 
OpenGL core profile version string: 4.5 (Core Profile) Mesa 19.0.4
OpenGL core profile shading language version string: 4.50
OpenGL version string: 3.0 Mesa 19.0.4
OpenGL shading language version string: 1.30
OpenGL ES profile version string: OpenGL ES 3.1 Mesa 19.0.4
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.10
```

### Issue Backtrace
```INFO: [TEX ID 2] Default font loaded successfully
WARNING: Trying to re-load an already loaded mesh
WARNING: [resources/guy/guy.iqm] No materials can be loaded, default to white material
INFO: [resources/guy/guytex.png] Image loaded successfully (1024x1024)
INFO: [TEX ID 3] Texture created successfully (1024x1024 - 1 mipmaps)
WARNING: More than 1 animation in file, only the first one will be loaded
INFO: Target time per frame: 16.667 milliseconds
free(): invalid pointer

Thread 1 "models_animatio" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7ae7535 in __GI_abort () at abort.c:79
#2  0x00007ffff7b3de38 in __libc_message (action=action@entry=do_abort, 
    fmt=fmt@entry=0x7ffff7c4a32b "%s\n") at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007ffff7b4451a in malloc_printerr (
    str=str@entry=0x7ffff7c48491 "free(): invalid pointer") at malloc.c:5352
#4  0x00007ffff7b45c44 in _int_free (av=<optimized out>, p=<optimized out>, 
    have_lock=<optimized out>) at malloc.c:4181
#5  0x000055555556aa3e in ?? ()
#6  0x00005555555c411d in ?? ()
#7  0x000055555555c306 in ?? ()
#8  0x00007ffff7ae8b6b in __libc_start_main (main=0x55555555be35, argc=1, 
    argv=0x7fffffffe6a8, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=0x7fffffffe698)
    at ../csu/libc-start.c:308
#9  0x000055555555bd7a in ?? ()
```

### Code Example

freshly cloned from repo example models_animation

wbrbr:
In the `LoadIQM` function, the `vaoId` of the meshes is never initialized to 0. So when `LoadModel` calls `rlLoadMesh`, if `vaoId` is positive then the mesh is never uploaded to the GPU ([see here](https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L2349)).

The fix is simply to initialize all the struct members to 0.

Fixes #848

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -3044,7 +3044,7 @@ static Model LoadIQM(const char *fileName)
     fread(imesh, sizeof(IQMMesh)*iqm.num_meshes, 1, iqmFile);
 
     model.meshCount = iqm.num_meshes;
-    model.meshes = RL_MALLOC(model.meshCount*sizeof(Mesh));
+    model.meshes = RL_CALLOC(model.meshCount, sizeof(Mesh));
 
     char name[MESH_NAME_LENGTH];
 


raysan5:
@wbrbr nice! thanks!

PR #852, opened 2019-05-24
easings.h refinement and example

flashback-fx:
Hello,

Some days ago I started making an example program for the functions in easings.h to see how they worked and to have visual output of the use of each one. I found out that easings.h produced many compiler warnings. I started modifying easings.h in order to shut those compiler warnings (about sequence points related undefined behavior) and I also made it use the tgmath.h C99 header instead of math.h and replaced the integer constants with proper float constants in order to have the functions do less conversions. I found out that I was working on an older version of easings.h and that the issues with the warnings have been fixed already. I still would like these modifications to be accepted because apparently they make the easings.h functions perform better.

I wrote two test programs: easings_results, to test the output of the functions, and easings_perftest, to test the performance of the functions. I used these programs to test differences between the current easings.h file and the one with my changes. Both programs and also the results of some tests can be found in a repository I made called easings_tools.

The tests showed that the version with my changes has some functions running faster: the EaseExpo functions seem to be at least 3x faster and the EaseElastic ones at least 2x faster. EaseSine and EaseCirc seem to be sightly faster as well, especially with compiler optimizations enabled. The rest seem to be equal. The test calls the functions through pointers so I assume the functions are not being inlined. The output is mostly equal to the one given by the current easings.h files, some values are very close but sightly different in the fractional part sometimes. These tests were made under GNU/Linux 64bit using GCC7. As I said before, the test programs are available for whoever is interested in running them on their systems.

I am also sending an example program for easings.h. I reworked the example program I made for me to fit the style (both in code and output) of the rest of the examples in raylib. It's still maybe a little different from the rest of the examples, if you'd like me to change something in order to fit better with the rest I can do it. I put it in examples/core but it's filename (easings_example) does not start with core_ and easings.h is not actually a part of core. I didn't want to put it in examples/. because there's only the template there. I will also move it to whichever place you think it's better. I didn't touch any of the makefiles for the examples.

Finally, I want to take this opportunity to thank you and congratulate you (author and all contributors) for your work on raylib, I think it is a very nice piece of software and I am enjoying it a lot.

Cheers!

Edit:
Direct link to the repository with the test programs and results:
https://github.com/flashback-fx/easings_tools


raysan5:
@flashback-fx thanks for the review! Please, could you move the example to examples/others directory and copy easings.h in the same directory?

flashback-fx:
Done! I should have seen examples/other I somehow skipped it. Thank you for answering so quickly.

raysan5:
Thanks to you for the quick solution! Merging! :)

flashback-fx:
Thank you for accepting it. I am very glad to be able to contribute to raylib.

raysan5:
@flashback-fx Juat note that I'll review the PR as soon as possible. Example will be probably rewritten to simplify it.

PR #855, opened 2019-05-29
Fix and update projects/scripts

neonmoe:
Multiple small fixes I've written since first sending these scripts, as well as a Raspberry Pi one, thanks to Discord user gururamen for the flags for that one :) I tested the scripts on Linux, Windows, macOS and a RPI 3B+.

### DIFF
--- a/projects/scripts/build-linux.sh
+++ b/projects/scripts/build-linux.sh
@@ -2,13 +2,13 @@
 # Change your executable name here
 GAME_NAME="game"
 
-# Set your sources here (relative to the ./builds/linux directory)
+# Set your sources here (relative paths!)
 # Example with two source folders:
-# SOURCES="../../src/*.c ../../src/submodule/*.c"
-SOURCES="../../core_basic_window.c"
+# SOURCES="src/*.c src/submodule/*.c"
+SOURCES="core_basic_window.c"
 
-# Set your raylib/src location here, relative to the ./temp/x directory
-RAYLIB_SRC="../../../../src"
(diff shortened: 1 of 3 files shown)


raysan5:
@neonmoe Thanks for the scripts review!

a3f:
I didn't know we had a `projects/scripts`, but I can't say I like the idea.
If there are shortcomings with the Makefile or the CMake setup, they should be fixed instead of adding a third build system that might get out-of-sync later on.

raysan5:
@a3f well, it's just another project option in case someone wants to use...

a3f:
@raysan5 Projects in the upstream raylib repo can build _themselves_ however they like. That's what they are there for. But they should either link against a precompiled raylib or build it theirselves using the _existing_ build scripts.

Adding yet another way to build raylib only adds confusion and maintenance burden. If these scripts do something that e.g. the Makefile can't, shouldn't that be added to the Makefile instead of a shell script only a part of raylib users use?

a3f:
I just fail to see the utility. If they are better, the Makefile needs to be updated. If they're just another way of doing what `src/Makefile` does, they shouldn't be in the main repository. If they are just a convenience, they should call make (or CMake) and let them build the library behind the scenes.

raysan5:
I see them like a convenience for some users, I agree they should better call make behind the scenes... but I don't see that much trouble if they don't.

They could also be useful as a reference or just as a multiplatform scripting lesson to learn from.

PR #850, opened 2019-05-22
Add rewind for the XM music format

wbrbr:
Disclaimer: I don't know the XM format (at all) and I don't know jar_xm.h at all
But hey, this seems to work on all the files I have tested on.

Fixes #843

PR #856, opened 2019-05-29
Fix jar_xm_reset

wbrbr:
See #850 
The function didn't work for some .xm files ([one](https://modarchive.org/index.php?request=view_by_moduleid&query=59963) [two](https://modarchive.org/index.php?request=view_by_moduleid&query=43516)) so I fixed it.


raysan5:
@wbrbr thanks for the fix... that module requires some work...

PR #858, opened 2019-06-02
Add MP3 config to config.h.in

Rover656:
config.h.in is missing an MP3 flag which stops MP3 audio from working on cmake projects unless this is overcome/overriden.


raysan5:
Ouch! I didn't notice it! Thanks! 👍

Issue #613, opened 2018-07-30
[build] Make raylib available in package managers


raysan5:
This issue is a follow up of #401 . Just moved to a new issue flor clearness.

Currently supported package managers:

- [x] macOS - [Homebrew](https://brew.sh/)
- [ ] Windows - [Chocolatey](https://chocolatey.org/)
- [x] Windows - [vcpkg](https://github.com/Microsoft/vcpkg)
- [x] Linux - [openSUSE](https://build.opensuse.org/package/show/devel:libraries:c_c++/raylib)
- [x] Linux - [Arch User Repositories (AUR)](https://aur.archlinux.org/packages/raylib/)
- [x] Linux - [Linuxbrew](http://linuxbrew.sh/)
- [ ] Linux - [Gentoo](https://bugs.gentoo.org/652848)
- [ ] Linux - [Debian](https://www.debian.org/distrib/packages)
- [x] BSD - [FreeBSD](https://www.freshports.org/devel/raylib/)

If you are able and willing to help, you're welcome!

raysan5:
Just noticed that openSUSE (@jubalh) and Arch (@nounoursheureux) are not updated to raylib 2.0, still 1.8.

jubalh:
@raysan5 thats wrong. openSUSE Tumbleweed is updated to 2.0 on 24th July in the devel project https://build.opensuse.org/package/show/devel:libraries:c_c++/raylib.  From then it goes through test rounds and landed on 28 July on Factory project. And from there will be shipped in the next snapshot.

raysan5:
@jubalh oh, sorry! just look for it on the net and found that version... Thanks for the update! 👍 😄

define-private-public:
I have experience doing debian packaging; I can offer my assistance for this.  Though, I do want you to know that there are MANY different options to do this.  Though this library is open source, I'd recommend creating a binary package (as they are a tad simplier to work with).

What's the status though on the CMake stuff?  I know I helped with it a while back, but I'm really out of the loop on it.  There does exist a command in CMake to generate the debian stuffs: https://cmake.org/cmake/help/v3.0/module/CPackDeb.html

a3f:
> There does exist a command in CMake to generate the debian stuffs

CI is configured to push a binary package (tarball) to the [Github releases page](https://github.com/raysan5/raylib/releases) whenever a new release is tagged, so we could use that. Anyone interested in trying to get it into the official repositories?

mandeep:
[...]

PR #862, opened 2019-06-04
build: increment API_VERSION after release

a3f:
With v2.5.0 out, increment API_VERSION, so binaries dynamically linked against
the released raylib aren't accidentally paired with a development or later
released raylib that may be incompatible.

---

Issues was discussed multiple time in the past (see #613 for an example).
The intention here is not to change raylib's versioning scheme, just the compatibility/API version evaluated by the dynamic loader.


raysan5:
@a3f Thanks for the improvement and excuse me ignorance on this versioning issue...

a3f:
No worries, it took me quite some time (and one last comment by @jubalh) to wrap my head around the issue ^^'.

PR #869, opened 2019-06-05
Added waving cubes example

codecat:
Added my [waving cubes](https://twitter.com/codecatt/status/1135870039637274624) example.

I hope I got the pull request right! I only tried building this on MSVC. Originally it was built on MSVC C++, so it might need some changes for C still.

models_waving_cubes


raysan5:
Wow! Great addition @codecat! Thanks! 😄

PR #870, opened 2019-06-05
Added GetCurrentMonitor() function

rsachetto:
Get the monitor that contains the greater window area.


rsachetto:
Incomplete code.

PR #873, opened 2019-06-06
Changed ANGLE binaries location

msmshazan:
Moved ANGLE binaries into VS2017.ANGLE folder


raysan5:
@msmshazan thanks for the update! :)

PR #876, opened 2019-06-08
Update Makefile

XiaochuanWang:
corrected file location for "raylib.rc.data", which will cause an error during compilation


raysan5:
@XiaochuanWang Thanks for the reviews! :)

Issue #863, opened 2019-06-04
[core] Keyboard input system not working properly


DarkElvenAngel:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

The input system is not working all key presses are echoed to the terminal the only way to terminate the program is with Ctrl+C 

My older build is not affected by this error.

### Environment

Raspberry Pi Native mode

### Issue Screenshot

![out](https://user-images.githubusercontent.com/28216652/58891392-67802080-86ba-11e9-8705-a848adb69e2f.png)

### Code Example

This was a fresh rebuild of the main branch version 2.5 all examples are affected.

raysan5:
Sorry, it was caused by commenting `stdin` keyboard reading... Please, could you try uncommenting `#define SUPPORT_SSH_KEYBOARD_RPI` on `config.h`?

raysan5:
While investigating why this happens, just reactivated `stdin` redirect code: https://github.com/raysan5/raylib/commit/6f9c176d939c8035b0f14836de80e916826c0325

DarkElvenAngel:
That's working now, the output to the console is the same but now we have successfully uploading output and the examples exit cleanly.

What does this flag do?

raysan5:
@DarkElvenAngel Right now two keyboard input systems coexist in the Raspberry Pi: one reconfigures standard system input to read keycodes (that could be dangerous), the second one just reads pre-processed keyboard inputs using `evdev` system library on a second thread. First one supports SSH inputs but it's dangerous, second ones requires super-user rights and it was added lately.

I should remove first mechanism but I just disabled with a flag for the moment, second mechanism should be the standard.

DarkElvenAngel:
I was running the example from the root user both before and after enabling that flag.  If that helps with the investigation of this issue.

@raysan5 so using the library as I have it compiled now is not recommended/dangerous (potentially)?

raysan5:
@DarkElvenAngel It has been this way since the beginning, second mechanism was added some weeks ago.

It's potentially dangerous in the sense that it locks the device if programs does not exit properly. Also, if you have another program running in background it could make it crash when running raylib example (due to the stdin reconfiguration). For those reasons, second mechanism was added...

DarkElvenAngel:
[...]

PR #879, opened 2019-06-10
Keyboard patch

DarkElvenAngel:
This is a patch for the PLATFORM_RPI Raw keyboard driver issue #863 

Please note that I have bypassed the stock keyboard buffer as it is not used and no data is forwarded to the rest of the library for processing this causes the keyboard to be effectively a dead input.  I've also added some terminal setting that will stop echoing and block Ctrl C and Ctrl Z from sending signals.  As well as processing for Exit Key and Screen Capture.

This code works in my limited tests.


raysan5:
@DarkElvenAngel Thanks for the patch! I'll review the mentioned issues... :)

Issue #891, opened 2019-06-14
[rlgl] Mixing triangle calls with quad calls causes artefacts


kawa-yoiko:
### Issue description

(Updated the previously incorrect assumption, see revision history)

Some raylib functions make use of quads (e.g. `DrawRectangle`), while others use triangles (e.g. `DrawTriangleLines`). Interleaving such calls without explicitly calling `rlglDraw()` seems to be causing unwanted artefacts in the output image.

### Environment

macOS 10.13.5, OpenGL 4.1. I have not tested on other platforms.

```plain
INFO: GPU: Vendor:   Intel Inc.
INFO: GPU: Renderer: Intel(R) Iris(TM) Graphics 6100
INFO: GPU: Version:  4.1 INTEL-10.34.27
INFO: GPU: GLSL:     4.10
```

### Issue Screenshot

![screenshot_rlg](https://user-images.githubusercontent.com/45892908/59520843-7017f880-8efd-11e9-9c1b-f6c601c4dce7.png)

### Code Example

The following code intends to draw several vertically aligned squares and horizontally aligned circles divided into outlined triangles, but produces the image above.

```c
BeginDrawing();

    ClearBackground(RAYWHITE);

    for (int i = 0; i < 20; i++) {
        DrawRectangle(100, 100 + i * 20, 200, 200, (Color){160, 160, 160, 64});
        for (int j = 0; j < 997; j++) {
            // rlglDraw();
            DrawTriangleLines((Vector2){
                200 + i * 40, 200 
            }, (Vector2){
                200 + i * 40 + 100 * sinf(M_PI * 2 * j / 997),
                200 + 100 * cosf(M_PI * 2 * j / 997)
            }, (Vector2){
                200 + i * 40 + 100 * sinf(M_PI * 2 * (j + 1) / 997),
                200 + 100 * cosf(M_PI * 2 * (j + 1) / 997)
            }, LIGHTGRAY);
        }
    }

EndDrawing();
```

If the `rlglDraw()` above is uncommented, the artefact will disappear.

The problem seems to be caused by the the vertex padding being not zero for quad calls.

~~On a side note, `DrawTriangle()` and `DrawTriangleLines()` do not call `rlCheckBufferLimit()` at the beginning, but that is not the cause of this issue.~~ Not a problem since `rlEnd()` already prevents overflow for calls with few vertices.

Thanks in advance!

raysan5:
Thanks for reporting @kawa-yoiko! 

[...]

PR #893, opened 2019-06-15
Fix vertex padding not zeroed for quads

kawa-yoiko:
I've tested this with the example in #891 and my own application which used to break, both work fine now.

Hope it helps!


raysan5:
@kawa-yoiko Thanks for the fix! :)

PR #898, opened 2019-06-24
Submitting rmem memory and object pool module

assyrianic:
memory and object pool for quick memory/object allocation and reduce possible memory leaks.

### DIFF
--- a/mempool README.txt
+++ b/mempool README.txt
@@ -1,0 +1,163 @@
+Raylib Memory Pool
+By Kevin 'Assyrianic' Yonan @ https://github.com/assyrianic
+
+About:
+	The Raylib Memory Pool is a quick, efficient, and minimal free list & stack-based allocator.
+
+Purpose:
+	Raylib Memory Pool's purpose is the following list:
+		* A quicker, efficient memory allocator alternative to `malloc` and friends.
+		* Reduce the possibilities of memory leaks for beginner developers using Raylib.
+		* Being able to flexibly range check memory if necessary.
+
+Data Implementation:
+	the memory pool encapsulates two public structs:
+	`freeList` which is an abstracted doubly linked list consisting of a `head` and `tail` pointer to `struct MemNode`, a `len` that tracks the amount of nodes the linked list holds, `maxNodes` which is used for auto-defragging (explained below), and `autoDefrag` which controls whether auto-defragging will execute or not.
+	```c
+	typedef struct MemNode {
+		size_t size;
+		struct MemNode *next, *prev;
+	} MemNode;
+
+	typedef struct AllocList {
+		struct MemNode *head, *tail;
+		size_t len, maxNodes;
+		bool autoDefrag : 1;
+	} AllocList;
+	```
+	
+	`Stack` which is an array-based stack consisting of a byte pointer to the entire array `mem`, the base pointer `base` which changes position when allocating, deallocating, or defragging, and `size` which tracks how large the entire stack buffer is.
+	```c
+	typedef struct Stack {
+		uint8_t *mem, *base;
+		size_t size;
+	} Stack;
+	```
+	
+	```c
+	typedef struct MemPool {
+		struct AllocList freeList;
+		struct Stack stack;
+	} MemPool;
+	```
+
+Explanation & Usage:
+	The memory pool is designed to be used as a direct object.
+	We have two constructor functions:
+	```c
+	struct MemPool MemPool_Create(size_t bytes);
+	struct MemPool MemPool_FromBuffer(size_t bytes, void *buf);
+	```
+	
+	To which you create a `struct MemPool` instance and give the function a max amount of memory you wish or require for your data.
+	Remember not to exceed that memory amount or the allocation functions of the allocator will give you a NULL pointer.
+	
+	
+	So we create a pool that will malloc 10K bytes.
+	```c
+	struct MemPool pool = MemPool_Create(10000);
+	```
+	
+	When we finish using the pool's memory, we clean it up by using `MemPool_Destroy`.
+	```c
+	MemPool_Destroy(&pool);
+	```
+	
+	Alternatively, if you're not in a position to use any kind of dynamic allocation from the operating system, you have the option to utilize an existing buffer as memory for the mempool:
+	```c
+	char mem[64000];
+	struct MemPool pool = MemPool_FromBuffer(sizeof mem, &mem[0]);
+	```
+	
+	To allocate from the pool, we have two functions:
+	```c
+	void *MemPool_Alloc(struct MemPool *mempool, size_t bytes);
+	void *MemPool_Realloc(struct MemPool *mempool, void *ptr, size_t bytes);
+	```
+	
+	`MemPool_Alloc` returns a (zeroed) pointer to a memory block.
+	```c
+	// allocate an int pointer.
+	int *i = MemPool_Alloc(&pool, sizeof *i);
+	```
+	
+	`MemPool_Realloc` works similar but it takes an existing pointers and resizes its data, it does NOT zero the memory as it exists for resizing existing data. Please note that if you resize a smaller size, the data WILL BE TRUNCATED/CUT OFF.
+	If the `ptr` argument for `MemPool_Realloc`, it will work just like a call to `MemPool_Alloc`.
+	```c
+	// allocate an int pointer.
+	int *i = MemPool_Realloc(&pool, NULL, sizeof *i);
+	
+	// resize the pointer into an int array of 10 cells!
+	i = MemPool_Realloc(&pool, i, sizeof *i * 10);
+	```
+	
+	To deallocate memory back to the pool, there's also two functions:
+	```c
+	void MemPool_Free(struct MemPool *mempool, void *ptr);
+	void MemPool_CleanUp(struct MemPool *mempool, void *ptrref);
+	```
+	
+	`MemPool_Free` will deallocate the pointer data back to the memory pool.
+	```c
+	// i is now deallocated! Remember that i is NOT NULL!
+	MemPool_Free(&pool, i);
+	```
+	
+	`MemPool_CleanUp` instead takes a pointer to an allocated pointer and then calls `MemPool_Free` for that pointer and then sets it to NULL.
+	```c
+	// deallocates i and sets the pointer to NULL.
+	MemPool_CleanUp(&pool, &i);
+	// i is now NULL.
+	
+	```
+	
+	Using `MemPool_CleanUp` is basically a shorthand way of doing this code:
+	```c
+	// i is now deallocated! Remember that i is NOT NULL!
+	MemPool_Free(&pool, i);
+	
+	// i is now NULL obviously.
+	i = NULL;
+	```
+	
+	
+	Given that the memory pool is based on a stack allocator, that means to refill the stack, you must free in the opposite order you allocate memory.
+	You can still free memory out of order but that'll create memory block fragments.
+	
+	Freed memory blocks that are not absorbed into the stack allocator are placed into the allocator's free list. If you have too many nodes, the allocation functions might take a while to calculate as the allocating functions first check the free list for any memory blocks it can give you before falling back to the stack allocator.
+	
+	If you have too many nodes (You can check by accessing the `freeList` struct and its `len` member like: `pool.freeList.len`), then you'll have to defragment the free list.
+	Defragmenting consists of merging together memory blocks that are physically near one another.
+	
+	you have two options in terms of defragging:
+	You can manually call the defrag function:
+	```c
+	bool MemPool_DeFrag(struct MemPool *mempool);
+	```
+	which will return true or false depending if it was able to merge any nodes.
+	
+	or you can set a node limit and enable auto defragging.
+	Auto defragging is disabled by default, you can turn it on or off either by calling:
+	```c
+	void MemPool_ToggleAutoDefrag(struct MemPool *mempool);
+	```
+	or set it directly from the `freeList` struct member:
+	```c
+	pool.freeList.autoDefrag = true; // set to on.
+	pool.freeList.autoDefrag = false; // set to off.
+	pool.freeList.autoDefrag ^= true; // toggle on or off.
+	```
+	
+	For auto defragging to work, you must also set a maximum memory block node limit.
+	you can set it directly with the `maxNodes` member in the `freeList` struct member:
+	```c
+	// set free memory block node limit to 100!
+	pool.freeList.maxNodes = 100UL;
+	```
+	If auto defragging is not enabled, then the node limit is ignored of course.
+	
+	Finally, to get the total amount of memory remaining (to make sure you don't accidentally over-allocate), you utilize this function:
+	```c
+	size_t MemPool_MemoryRemaining(const struct MemPool mempool);
+	```
+	
--- a/objpool README.txt
+++ b/objpool README.txt
@@ -1,0 +1,76 @@
+Raylib Object Pool
+By Kevin 'Assyrianic' Yonan @ https://github.com/assyrianic
+
+About:
+	The Raylib Object Pool is a fast and minimal fixed-size allocator.
+
+Purpose:
+	Raylib Object Pool was created as a complement to the Raylib Memory Pool.
+	Due to the general purpose nature of Raylib Memory Pool, memory block fragmentations can affect allocation and deallocation speeds.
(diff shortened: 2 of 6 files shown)


mempool README.txt
```diff
@@ -0,0 +1,163 @@
...
+Purpose:
+	Raylib Memory Pool's purpose is the following list:
+		* A quicker, efficient memory allocator alternative to `malloc` and friends.
+		* Reduce the possibilities of memory leaks for beginner developers using Raylib.
+		* Being able to flexibly range check memory if necessary.
+
+Data Implementation:
+	the memory pool encapsulates two public structs:
+	`freeList` which is an abstracted doubly linked list consisting of a `head` and `tail` pointer to `struct MemNode`, a `len` that tracks the amount of nodes the linked list holds, `maxNodes` which is used for auto-defragging (explained below), and `autoDefrag` which controls whether auto-defragging will execute or not.
+	```c
```
a3f: Use a `.md` file extension for markdown files. While at it, drop the space from the filename

mempool README.txt
```diff
@@ -0,0 +1,163 @@
...
+	
+	When we finish using the pool's memory, we clean it up by using `MemPool_Destroy`.
+	```c
+	MemPool_Destroy(&pool);
+	```
+	
+	Alternatively, if you're not in a position to use any kind of dynamic allocation from the operating system, you have the option to utilize an existing buffer as memory for the mempool:
+	```c
+	char mem[64000];
+	struct MemPool pool = MemPool_FromBuffer(sizeof mem, &mem[0]);
```
a3f: `&mem[0]` => `mem`. The first variant is unnecessarily verbose.

objpool README.txt
```diff
@@ -0,0 +1,76 @@
...
+	```c
+	ObjPool_CleanUp(&vector_pool, &origin);
+	```
+	
+	Which of course is equivalent to:
+	```c
+	ObjPool_Free(&vector_pool, origin), origin = NULL;
+	```
+	
+	You may also optionally back your
```
a3f: Line is incomplete

src/rmem.c
```diff
@@ -0,0 +1,470 @@
+#include "rmem.h"
+
+
+typedef uintptr_t ptrcmp;
```
a3f: Drop the unnecessary custom type.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+
+static size_t _AlignSize(const size_t size, const size_t align)
```
a3f: * `/^_[A-Z]/` identifiers are reserved for the implementation (i.e. compiler and system libraries), don't use them in user code.
* Add an `inline` as an extra hint to the compiler.
* I am not a fan of const parameters. I'd drop them, but that's IMO.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+{
+    return (size + (align-1)) & -align;
```
a3f: s/`-align`/`~(align - 1)`/ is clearer.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+}
+
+
+/************* Memory Pool *************/
+
+static void _RemoveNode(struct MemNode **const node)
+{
+    ((*node)->prev != NULL)? ((*node)->prev->next = (*node)->next) : (*node = (*node)->next);
```
a3f: Some compilers might warn about the expression result being unused. Use an `if` statement instead.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+    ((*node)->next != NULL)? ((*node)->next->prev = (*node)->prev) : (*node = (*node)->prev);
+}
+
+struct MemPool MemPool_Create(const size_t size)
+{
+    struct MemPool mempool = {0};
+    if( size==0UL )
```
a3f: Please keep the coding style coherent. This should read `if (size == 0)`, same for the rest of the commits.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+        return mempool;
+    else {
```
a3f: `else` clause unnecessary. drop it.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+        // align the mempool size to at least the size of an alloc node.
+        mempool.stack.size = _AlignSize(size, sizeof(struct MemNode));
+        mempool.stack.mem = malloc(1 + mempool.stack.size*sizeof *mempool.stack.mem);
+        if( mempool.stack.mem==NULL ) {
+            mempool.stack.size = 0UL;
+            return mempool;
```
a3f: factor out the `return mempool` instead of repeating it.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+        } else {
+            mempool.stack.base = mempool.stack.mem + mempool.stack.size;
+            return mempool;
+        }
+    }
+}
+
+struct MemPool MemPool_FromBuffer(const size_t size, void *buf)
```
a3f: Having the `buf` preceded by the `size` is non-idiomatic. Normally, it's the other way round.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+{
+    struct MemPool mempool = {0};
+    if( size==0UL || buf==NULL || size<=sizeof(struct MemNode) )
+        return mempool;
+    else {
```
a3f: Drop the else, when early exiting.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+    }
+}
+
+void *MemPool_Alloc(struct MemPool *const mempool, const size_t size)
+{
+    if( mempool==NULL || size==0UL || size > mempool->stack.size )
+        return NULL;
+    else {
+        struct MemNode *new_mem = NULL;
+        const size_t ALLOC_SIZE = size + sizeof *new_mem;
```
a3f: keep ALL_CAPS identifier usage for macros.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+            for( struct MemNode **inode = &mempool->freeList.head; *inode != NULL; inode = &(*inode)->next ) {
+                if( (*inode)->size < ALLOC_SIZE )
+                    continue;
+                else if( (*inode)->size <= ALLOC_SIZE + MEM_SPLIT_THRESHOLD ) {
+                    // close in size - reduce fragmentation by not splitting.
+                    new_mem = *inode;
+                    _RemoveNode(inode);
+                    mempool->freeList.len--;
+                    new_mem->next = new_mem->prev = NULL;
+                    break;
```
a3f: Useless break.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+                } else {
+                    // split the memory chunk.
+                    new_mem = (struct MemNode *)( (uint8_t *)*inode + ((*inode)->size - ALLOC_SIZE) );
+                    (*inode)->size -= ALLOC_SIZE;
+                    new_mem->size = ALLOC_SIZE;
+                    new_mem->next = new_mem->prev = NULL;
+                    break;
```
a3f: Ergo

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+    else {
+        struct MemNode *new_mem = NULL;
+        const size_t ALLOC_SIZE = size + sizeof *new_mem;
+        if( mempool->freeList.head != NULL ) {
+            const size_t MEM_SPLIT_THRESHOLD = sizeof(intptr_t);
+            // if the freelist is valid, let's allocate FROM the freelist then!
+            for( struct MemNode **inode = &mempool->freeList.head; *inode != NULL; inode = &(*inode)->next ) {
+                if( (*inode)->size < ALLOC_SIZE )
+                    continue;
+                else if( (*inode)->size <= ALLOC_SIZE + MEM_SPLIT_THRESHOLD ) {
```
a3f: Useless else

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+            // not enough memory to support the size!
+            if( mempool->stack.base - ALLOC_SIZE < mempool->stack.mem )
+                return NULL;
+            else {
+                // couldn't allocate from a freelist, allocate from available mempool.
+                // subtract allocation size from the mempool.
+                mempool->stack.base -= ALLOC_SIZE;
+                
+                // use the available mempool space as the new node.
+                new_mem = (struct MemNode *)mempool->stack.base;
```
a3f: Potential non-aligned address stored. Results in undefined behavior here and in successive lines.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+                mempool->freeList.len++;
+            }
+            
+            if( mempool->freeList.autoDefrag && mempool->freeList.maxNodes != 0UL && mempool->freeList.len > mempool->freeList.maxNodes )
+                MemPool_DeFrag(mempool);
+        }
+    }
+}
+
+void MemPool_CleanUp(struct MemPool *const __restrict mempool, void *ptrref)
```
a3f: * We presume C99, thus use restrict. If MSVC doesn't like it, add some #ifdefery on-top of the file that defines restrict to __restrict if necessary
* use double indirect pointer for the second argument, to make misuse harder?

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+{
+    if( mempool==NULL || ptrref==NULL )
+        return;
+    else {
+        void *__restrict *p = ptrref;
+        if( *p==NULL ) {
+            return;
+        } else {
+            MemPool_Free(mempool, *p), *p = NULL;
```
a3f: Split it unto two lines.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+
+
+void MemPool_ToggleAutoDefrag(struct MemPool *const mempool)
+{
+    if( mempool==NULL )
+        return;
+    else mempool->freeList.autoDefrag ^= true;
+}
+
+#if 0
```
a3f: No dead code. Either leave them in or drop them.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+    if( mempool==NULL || mempool->stack.mem==NULL )
+        return -1;
+    else if( ptr==NULL || (ptrcmp)ptr < (ptrcmp)mempool->stack.mem || (ptrcmp)ptr > (ptrcmp)mempool->stack.mem + mempool->stack.size || (ptrcmp)ptr < (ptrcmp)mempool->stack.base )
+        return false;
+    else if( mempool->freeList.len != 0UL ) {
+        struct MemNode *mem = (struct MemNode *)((uint8_t *)ptr - sizeof *mem);
+        for( struct MemNode *n=mempool.freeList.head; n != NULL; n = n->next )
+            if( mem==n )
+                return false;
+        return true;
```
a3f: drop it here.

src/rmem.c
```diff
@@ -0,0 +1,470 @@
...
+    } else return true;
```
a3f: drop the `else`

mempool README.txt
assyrianic: I did that to be intentionally verbose, I'll remove it since you asked.

src/rmem.c
assyrianic: it's not unnecessary, it's for branch prediction optimization.

src/rmem.c
assyrianic: again, this is for branch prediction optimization.
I discovered this when I was working on my virtual machine project named Tagha.

src/rmem.c
assyrianic: Ummm, the coding conventions specifically said ALL_CAPS for macros AND constants...

https://github.com/raysan5/raylib/wiki/raylib-coding-conventions

```
Code element 	Convention 	Example
Macros 	ALL_CAPS 	#define MIN(a,b) (((a)<(b))?(a):(b))
Defines 	ALL_CAPS 	#define PLATFORM_DESKTOP
Constants 	ALL_CAPS 	const int NUMBER = 8
```

src/rmem.c
assyrianic: it is not useless.

src/rmem.c
assyrianic: Problem with double indirect pointer is that you constantly need to cast the pointer type to `void**` to where that gets as verbose as `reinterpret_cast<void**>()`

src/rmem.c
assyrianic: need it for branch prediction optimization.

src/rmem.c
assyrianic: branch predict optimization.

src/rmem.c
assyrianic: the only compilers that have good C99 support is the variants of GCC and Clang, both support this operation.

src/rmem.c
assyrianic: the `malloc`d memory is aligned to the size of a `struct MemNode *`.
I've tested this memory pool on a Raspberry Pi 2 which is ARM based, there was no misalignment errors which triggers a bus error on ARM processors.

assyrianic: Confirmed, no misalignment or bus errors on my Pi 2. If you have an ARM processor, I suggest testing this on that.

src/rmem.c
a3f: What? Both code versions with or without `else` should give the exact same object code output. The only difference is that you have an extra unnecessary level of indentation with the readability decrease that comes with it.

src/rmem.c
a3f: Still, this is unnecessary crowded. Why not just use `if` statements like everyone else does?

src/rmem.c
a3f: See above

src/rmem.c
a3f: That might have been the case in your other project. I fail to see how this would affect branch prediction _here_ anyhow. If you absolutely want to mark branches, use a macro around `__builtin_expect` (check `<linux/kernel.h>` `likely` and `unlikely` macros).

src/rmem.c
a3f: Didn't know about that one tbh. Still I don't think all caps is appropriate here, it's `const`, but it's not constant (it's variable differed depending on `size`).

src/rmem.c
a3f: Ah, sorry my bad. Same comment about `else` before applies to `else`s here.

src/rmem.c
a3f: Ye, but on the other hand, a plain `void *` parameter is easily passed the pointer, not a ref. Both are suboptimal, but I'd argue better have the user be aware while casting, instead of erroneously accepting wrong pointers.

a3f:
I am still interested in seeing a real-world example where rmem outperforms a stock `malloc`. I can see the utility of an object pool/arena allocator, but not really of rmem.

src/rmem.c
assyrianic: because this would become an `if`-`else` statement like the rest of the code :)

src/rmem.c
assyrianic: What compiler? What version? What flags?

assyrianic:
@a3f with or without auto-defragging?

assyrianic:
The mempool is slightly faster than `malloc` but, from my research, it's because the mempool is not thread safe.

I should also say (which it does in my readme) that the mempool also reduces the chances of memory leaks since a single destruction function call to the mempool frees all memory.

You can also range check memory easier.

raysan5:
@a3f @assyrianic This memory allocation was not intended to be included by raylib, just intended to be used on a code example, illustrating raylib support for a custom memory pool... I though placing it on src directory (similar to `physac.h` or `easings.h`) would be a good idea.

In any case, it still requires a complete review. Functions will probably be renamed and I'd like to move to a header-only library...

assyrianic:
I'd like to say that I'm considering replacing the doubly linked list based Free list with a binary, ordered tree so that finding a good allocation size would be O(_log_ n) rather than O(n) where _n_ is the number of free list nodes.

assyrianic:
@a3f [`malloc` slowness source](https://github.com/mtrebi/memory-allocators)

* **General purpose**: Being a general purpose operation means that it must work in all cases (from 1byte to 1GB or more...). For this reason the implementation is not as efficient as it could be if the needs were more specific.
* **Slow**: Sometimes, when allocating memory, malloc needs to change from user to kernel mode to get more memory from the system. When this happens, malloc turns out to be super slow!

This is the reason for providing an object and a memory pool for raylib, so that developers (beginner or experienced) do not have to rely on `malloc` and potentially hurt performance of their games. Not to mention so that they don't have to implement their own memory or object pools when they need it.

raysan5:
@assyrianic Thanks for your hard work on this module, merging it.

Just note that all names and coding style is being reviewed to follow raylib conventions.

PR #900, opened 2019-07-01
Adding solar system examples.

aldrinmartoq:
Code isn't perfect, but it works ;-)


raysan5:
@aldrinmartoq Thank you very much for providing this example for testing.

Just note this example adds lots of resources that increase a lot raylib size, it won't probably be included with raylib installer in a future, I need to decide it...

PR #914, opened 2019-07-22
Increasing threshold size for traversing memory blocks

assyrianic:
L235 - Increased the memory threshold size so that we can further reduce fragmentation by allocating memory blocks that is close to the requested size by 16 bytes.


raysan5:
Thanks for the improvement!

PR #915, opened 2019-07-26
Should fix RL_WRAP_MIRROR_CLAMP issue

Rover656:
This should fix a logic error which would mean that, even if it was supported, texture wrapping mode RL_WRAP_MIRROR_CLAMP would never be applied. Below is an image of where the issue occurred.

Issue

This PR assumes that OpenGl 1.1 is incapable of using this extension regardless of whether or not the supported flag is true. If this is wrong, please correct me.


Rover656:
Closing due to IDEA folders, will upload clean copy

Issue #917, opened 2019-07-26
Error compiling for web , broken path for shell.html


anatagawa:
Error when compiling for web.

In the makefile , cant find the path for **shell.html**

before:
`    # Define a custom shell .html and output extension
    CFLAGS += --shell-file $(RAYLIB_PATH)\src\shell.html`

after:
`    # Define a custom shell .html and output extension
    CFLAGS += --shell-file $(RAYLIB_PATH)/src/shell.html`

PR #918, opened 2019-07-27
#917 This commit changes the path to 'shell.html' in all Makefiles

niorad:
This fixes the a bug while building for HTML5 by changing the mentioned backslashes to forward-slashes.
https://github.com/raysan5/raylib/issues/917


raysan5:
@niorad Thank you very much for the fix! :)

PR #920, opened 2019-07-28
Patched potential bug when defragging.

assyrianic:
Patched a potential bug concerning the '__RemoveNode' function as, when removing certain nodes that are either at the head or tail, it can yield a free list node of size 0 bug as the previous node was removed and its size was set to 0 but not fully removed from the other nodes.


assyrianic:
Patched another part where the contiguous blocks of a lower size aren't being replaced.

raysan5:
@assyrianic thanks for the fix! :)

PR #923, opened 2019-07-31
End of July rmem Update

assyrianic:
Object Pool Changes:
-- changed 'size' member name of 'union ObjInfo' to 'index' to better name its purpose.
Memory Pool Changes:
-- Added memory node buckets to store and allocate smaller, more frequent byte sizes.
-- Replaced 'memset' call to deinitialize free list data with NULL and 0 assignments.
-- Removed some no-longer-needed commented-out code.
-- Changed insertion sort code to put the largest size at the tail rather than the head.
-- Made certain pointer variables as constant pointers.

### DIFF
--- a/src/rmem.h
+++ b/src/rmem.h
@@ -77,9 +77,13 @@ typedef struct Stack {
     size_t size;
 } Stack;
 
+#define MEMPOOL_BUCKET_SIZE    8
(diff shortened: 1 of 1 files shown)


raysan5:
@assyrianic Thanks for the update! :)

PR #927, opened 2019-08-03
Replace tabs with spaces and update year of copyright notices

VelocityIsNotSpeed:
Tabs in GitHub show as 8 spaces, so it looks like this PR de-indents, but it doesn't.

There are also some minor changes, such as fixing of typos and style, and removal of spaces at ends of lines.

There are 100+ other occurrences of outdated copyright notices in the examples, which i didn't update.

Note commit 9ac2fb44f09de5370c65fdb0aa17480bd308056f: I'm not 100% sure if it affects functionality (it shouldn't).


raysan5:
@VelocityIsNotSpeed thank you very much for this review! I try to do it from time to time but I could miss some files...

About examples, I like to keep the year when they get created and just update it if they are reviewed.

PR #928, opened 2019-08-06
Fix loading multiple animations from .iqm file

culacant:
Like it says on the box, multiple animations can be loaded from a single file.
Commmit 2 fixes an non-freed pointer in the animation example.
Commit 3 adds instructions to not break your model when exporting from blender, this might be fixed by not precalculating frameposes at model.c:1056 and model.c:3217 but this would add a lot of overhead each time a frame is calculated.


raysan5:
@culacant thanks for the improvements! Just a detail, could you replace TABS by spaces?

culacant:
@raysan5 tabs should be fixed.
I had to include stdlib.h in the example to make appveyor and travis CI happy, that's where the other 3 commits come from.

raysan5:
@culacant great! thanks for the improvement! :)

PR #929, opened 2019-08-06
Fix memory leak in LoadGLTF()

jubalh:
Free texturePath in LoadGLTF()


raysan5:
@jubalh good catch! thanks! :D

PR #930, opened 2019-08-06
Improved Android Building

NimbusFox:
Added in a check for x86 and x86_64 branches of android

Added in $(OS) checks to see if we're running on windows or *nix

Added a default location for android ndks and sdks at /usr/lib/android


raysan5:
@NimbusFox thanks for the addition!

PR #931, opened 2019-08-07
Fix build caused by GetImageAlphaBorder()

jubalh:
Fix build failure introduced in 2a913b6587d60dab64c3b81019990e6872a9ac75


raysan5:
@jubalh thanks!

PR #932, opened 2019-08-07
 Fix memleak in physics_restitution example

jubalh:


raysan5:
@jubalh I think ClosePhysics() already dies the trick but better to list it there

PR #933, opened 2019-08-07
fixed xmloader bug, user must free model shaders and textures as they…

chriscamacho:
… might be shared


raysan5:
@chriscamacho thanks for the review! :)

PR #935, opened 2019-08-08
Removing '__RemoveNode'

assyrianic:
Replaced '__RemoveNode' as it was causing invalid memory accesses with regular doubly linked list deletion algorithm.
Replaced double pointer iterator in 'MemPoolAlloc' with single pointer iterator.
Fixed undefined variables errors in 'MemPoolFree' for the freelist bucket.


raysan5:
@assyrianic thanks for the review!

Issue #936, opened 2019-08-08
Two examples fail to compile, version 2.6-dev


simpsond:
Please, before submitting a new issue verify and check:

 - [X ] I tested it on latest raylib version from master branch
 - [ X] I checked there is no similar issue already reported
 - [ X] My code has no errors or misuse of raylib

### Issue description

models_mesh_picking.c fails to compile with:
raylib/examples/models/models_mesh_picking.c:108:53: error: passing 'Model' (aka 'struct Model') to parameter of incompatible type 'Model *' (aka 'struct Model *'); take the address with &
            meshHitInfo = GetCollisionRayModel(ray, tower);

Similar failure in the models_obj_viewer.c:
raylib/examples/models/models_obj_viewer.c:57:74: error: passing 'Mesh' (aka 'struct Mesh') to parameter of incompatible type 'Mesh *' (aka 'struct Mesh *'); take the address with &
                    for (int i = 0; i < model.meshCount; i++) UnloadMesh(model.meshes[i]);
                                                                         ^~~~~~~~~~~~~~~

### Environment

macOS, cmake in clion

PR #937, opened 2019-08-08
updated examples to fix mesh reference

simpsond:
fixes https://github.com/raysan5/raylib/issues/936


raysan5:
@simpsond actually, those functions were reviewed on commit https://github.com/raysan5/raylib/commit/3d5fa81bf2753badacf9e8c01c51943a0c3acf85.

Current implementation should be ok, you probably tested it on a previous version.

PR #938, opened 2019-08-09
addition to raylib to create matrix from 3 euler angles

chriscamacho:


raysan5:
@chriscamacho thanks for the addition!

PR #940, opened 2019-08-11
Small fix in GetMouseY()

yaram:
This is a single small fix to an apparent copy-paste error in the `GetMouseY()` function present when building for the Android platform.


raysan5:
@yaram wow! good catch! thanks for the fix! 😄

PR #945, opened 2019-08-14
A more intuitive Camera2D behavior

arvyy:
Currently when working with Camera2D, you have to constantly change both offset and target when following the player. This isn't intuitive; when following player with camera, generally only one vector - player's position - changes every frame, and thus to update camera one would expect to also only change one field.

This pull request removes camera transformation by target at the end of camera world-space transformation, which means user doesn't have to add this value himself to the offset anymore. In effect, this makes offset to be not relative to world coordinates, but to screen, and thus the value doesn't have to be updated when the position of the followed object changes.

Illustration of new camera behavior

image

### DIFF
--- a/examples/core/core_2d_camera.c
+++ b/examples/core/core_2d_camera.c
@@ -42,7 +42,7 @@ int main(void)
 
     Camera2D camera = { 0 };
     camera.target = (Vector2){ player.x + 20, player.y + 20 };
-    camera.offset = (Vector2){ 0, 0 };
+    camera.offset = (Vector2){ screenWidth / 2, screenHeight / 2 };
     camera.rotation = 0.0f;
     camera.zoom = 1.0f;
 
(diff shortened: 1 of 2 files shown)


raysan5:
@arvyy thanks for the review, code is cleaner now. When I designed previous implementation, I [used this article](https://www.gamasutra.com/blogs/ItayKeren/20150511/243083/Scroll_Back_The_Theory_and_Practice_of_Cameras_in_SideScrollers.php) as reference and tried to provide a camera 2d to allow all those camera types, is it possible with current review?

arvyy:
well since the only place that changed was where offset is used, it's fairly trivial to convert between the current and proposed camera config: `offset_new = offset_current + target`, thus capabilities should be the same.

Regarding the link -- these all seem relatively simple if there were inbuilt functions for converting between screen coords and world coords (to compare player position defined in terms of world coordinates, with bounding box position defined in terms of screen coordinates), ie analogous to `GetWorldToScreen` / `GetMouseRay`, but for 2D. Perhaps could be a subject for another pull request

raysan5:
@arvyy I'm merging the change, still requires some testing...

PR #947, opened 2019-08-18
World to screen and screen to world transformations for Camera2D; extended Camera2D example

arvyy:
Referring to comment here: https://github.com/raysan5/raylib/pull/945#issuecomment-521766590

I went ahead and made an example of handful of camera logic approaches. To do this though, I also needed world to screen and screen to world functions, which I implemented as well. Names are placeholderish, as I didn't try to guess what you'd want them to be named as.


raysan5:
It looks good, example requires some review but I'm merging it.

raysan5:
@arvyy thank you very much for the contribution!

PR #949, opened 2019-08-19
Fix uninitialized vboId in GenMeshHeightmap

oraoto:
`mesh.vboId` is  not uninitialized in `GenMeshHeightmap`, which cause segfault in the models_heightmap example.


raysan5:
@oraoto thanks for the fix! :)

PR #951, opened 2019-08-25
IsAudioBufferProcessed -> IsAudioStreamProcessed

zatherz:
It appears that you've missed renaming this function here, which makes trying to compile standalone raudio.c error out.


raysan5:
Ouch! Good catch @zatherz! :D

Issue #942, opened 2019-08-12
[audio] Music struct requires being updated inside functions


raysan5:
Some weeks ago `raudio` module was completely reviewed and `Music` struct redesigned, it was changed from an opaque pointer to a public struct, like all the other structs available in raylib.

It was long time I wanted to do this change, not only to expose functions parameters or make its bindings easier but also to make it consistent with all the other structures in raylib. Most raylib functions receive data by value, it was a design decision and I try to stuck to it.

Unfortunately, `Music` struct includes `sampleLeft` parameter that requires being updated as long music is being played.

Functions that update `sampleLeft` parameter:
```c
void UpdateMusicStream(Music music);      // Updates buffers for music streaming
void StopMusicStream(Music music);        // Stop music playing
```
Here we have two possible solutions:
1. Require those functions to receive `music`by reference:
```c
void UpdateMusicStream(Music *music);     // Updates buffers for music streaming
void StopMusicStream(Music *music);       // Stop music playing
```
This change will break all code-bases using previous versions of raylib... I use to review functions without worrying much about "small" breaks but in this case I think it really affects simplicity and ease-of-use...

2. Move `sampleLeft` somewhere else.
`Music` struct includes `stream.buffer` pointer that it's actually the audio buffer containing and processing the data to be played, maybe `samplesProcessed` could be moved there... I need to investigate that `AudioBuffer` more in depth...

Feedback and opinions are welcome!

chriscamacho:
Music is highlevel, samples and audio buffers are lower level, so even if samples left is only used when an audio buffer is playing music (as opposed to samples and effects) I think it makes sense to put it in the audio buffer, and it will mean that you don't have to break the API.

Where sampleleft needs to be accessed, will there need to be extra API calls to access this, or will audiobuffer struct become public ?

RobLoach:
One of raylib's benefits is the simplicity of its API. Which option would keep the API simple to use? Bindings can fix/update as needed.

Makes sense to have `Update` use pointers, similar with `UpdateCamera(Camera *camera)`. If `sampleLeft` moved elsewhere, that'd be cool too.

a3f:
I haven't looked, but does user code mess with Music's contents? If not, we could do `typedef struct Music Music[1]`. Maintains source level compatibility, but each use of Music turns into a &Music.

(IMHO, I don't think passing `struct`s by value is something that should be encouraged anyway, but I just wanted to note this as an extra possibility)

raysan5:
[...]

PR #953, opened 2019-08-26
UpdateMusicStream takes pointer

zatherz:
`GetMusicTimePlayed(music)` always returns 0. This is reproducible with the `audio_music_stream` example.

I've investigated this and found out that while `GetMusicTimePlayed` makes use of `music.sampleLeft`, that field itself is supposed to be counted down by `UpdateMusicStream` as the samples are played. However, `UpdateMusicStream`, like all the other related functions, takes the `MusicStream` struct as an argument and not a pointer to it. Of course, the struct ends up being passed by value and the modifications made to the `sampleLeft` never leaves the function, causing `sampleLeft` to never tick down, `IsMusicPlaying` to never return false once `PlayMusicStream` has been called, and other problems.

This PR **changes the API** to make `UpdateMusicStream` take a pointer instead, and modify the underlying struct, making the aforementioned functionality work again. There are probably ways to solve this without having to change the API, but I wanted to make this PR to bring this issue to your attention more than anything.

I've also modified the audio-related examples to work properly.

### DIFF
--- a/examples/audio/audio_module_playing.c
+++ b/examples/audio/audio_module_playing.c
@@ -70,7 +70,7 @@ int main(void)
         // Restart music playing (stop and play)
         if (IsKeyPressed(KEY_SPACE))
         {
-            StopMusicStream(xm);
+            StopMusicStream(&xm);
             PlayMusicStream(xm);
         }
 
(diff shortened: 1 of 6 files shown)


zatherz:
Looks like I didn't test enough. With just the first commit, the time played is never reset as that is done in `StopMusicStream` which also suffers from a struct argument. However, works fine with the small change of making the argument a pointer.

raysan5:
@zatherz thanks for the review but I'm afraid this is not the desired solution for this issue: #942

PR #955, opened 2019-08-27
examples makefile not deleting examples in Linux

chriscamacho:
I'm not sure why the clean command is so complex (especially given that find -type f -executable | xargs rm -fv works!)
but I fixed the version as is, to support x-pie-executable which was preventing it deleting anything on my system...


raysan5:
@chriscamacho thanks! 😄

chriscamacho:
my pleasure

PR #963, opened 2019-09-11
fixes for switching full-screen and v-sync

MasterZean:


raysan5:
Sorry, calling `glfwSwapInternal()` every frame does not look as a good solution, it should be called only once or on context changes.

MasterZean:
I'm setting requireVSync = true on any configFlag change or ToggleFullscreen. You'll se in the new PR.

Who knows why glfwMakeContextCurrent is needed. Have you been able to reproduce the issue?

MasterZean:
Still won't work when starting off in fullscreen mode, only when switching from window to fullscreen.

A call to  glfwSetWindowMonitor after glfwSetWindowMonitor.

The actual framerate with v-sync is random without and sometimes v-sync is actually ignored.

When switching form window to fullscreen I actually can't reach 144 HZ, only 143.

MasterZean:
Yeah, yesterday's positive tests are no longer reproducible. I've been literally stuck for many years now on just mode switching in DirectX. Rendering goes perfectly, great graphics and performance, but switching mode doesn't work. Last time I could make it work was in DirectX 9.

Sad to see that OpenGL is just as unreliable. I'll continue testing but stuck for now.

chriscamacho:
I found that to get switching to work reliably (on Linux) I had to do...

        if (toggleScreen!=0) {
            toggleScreen--;
            if (toggleScreen==0) {
                ToggleFullscreen();
                if (!fullScreen) {
                    SetWindowSize(GetScreenWidth()/2, GetScreenHeight()/2);
                    SetWindowPosition(GetScreenWidth()/4, GetScreenHeight()/4);
                }
            }
        }

when I want to actually activate the change.....
        toggleScreen = 4;
        fullScreen = !fullScreen;
        if (fullScreen) {
            SetWindowSize(gameScreenWidth, gameScreenHeight);

otherwise it would do things like give me a low resolution fullscreen and generally behave unpredictably

not sure if this helps or muddies the waters!

MasterZean:
Hi chriscamacho!

This might be similar to my solution of calling glfwSwapInternal for 5-8 frames.

Something is definitely weird here.

I will rebuild the Raylib test app in pure GLFX and see if the swapping issues still reproduce.

chriscamacho:
hi MasterZean,
note I'm only delaying change of screen change pre entering fullscreen when if falls back to windowed, it doesn't seem as finicky!

raysan5:
Excuse me but I'm not convinced by the `SwapBuffers()` solution... I'll try to review this issue more carefully at some point...

raysan5:
Merging for further review...

PR #964, opened 2019-09-13
fixed some memory leaks, still issue with material index allocation l…

chriscamacho:
…leaking 4 bytes (possibly double allocation)

fixes non freed OBJ file data
fixes double allocation with shader locations in LoadShader

Some how I think the material index's are getting allocated twice, (and so only free'd once with one allocation never used and unfreeable) but I'm still trying to figure it out.....

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -296,7 +296,7 @@ void DrawCubeTexture(Texture2D texture, Vector3 position, float width, float hei
     float x = position.x;
     float y = position.y;
     float z = position.z;
-    
+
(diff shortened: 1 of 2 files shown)


raysan5:
@chriscamacho Great! Thanks for the review! And thanks for the comments and spacing removal... :P

chriscamacho:
well the space removal you get for free because geany
the comment you got because you deserved it :-p
I think I might have found the meshmaterial issue...

raysan5:
hahaha... nice! :D

PR #969, opened 2019-09-19
fixs issue mem leak with LoadModel / OBJ and issue with -1 material i…

chriscamacho:
…ndex

as promised....


raysan5:
@chriscamacho thanks! :D

PR #971, opened 2019-09-19
fixed leak with PBR example

chriscamacho:
as LoadModel now expects user to unload shaders and textures (as they could be used elsewhere) the example now does this...

just before loading the PBR material the default material (from loadOBJ/ LoadModel) must be unloaded, to avoid a leak....

phew... finally got there!


raysan5:
@chriscamacho nice! :D

PR #972, opened 2019-09-21
Fixed regular polygon shape rendering

AlexHCC:
Hi, I think I have fixed a bug.
In the moudle: _shapes_, the function: _void DrawPoly(Vector2 center, int sides, float radius, float rotation, Color color);_ Its implementation is wrong.

polygons2

There are problems with rendering polygons, with number of sides not divisible by 360.
The 360/sides is integer division, and as a result a  non divisible by 360 number will render sides+x sides. Sub-cases:
- If the lower bound of (360/sides) is not divisible by 360 there will be an overlap and it will create an artifact (more visible when rendering transparent). (Example sides: 7, 11, 13, 14...)

polygons1

- If the lower bound of (360/sides) is divisible by 360 there will be no overlap but there are still going to be rendered more edges. For example if you request a polygon with 23 sides, there will be rendered one with 24, or if you want one with 28 or 29 sides there will be rendered 30!
- If the lower bound of (360/sides) is 0 (side count higher than 360), the function is going to create infinitely many polygons (quads/triangles) since it will be stuck in an infinite loop, until there will be a buffer/memory overflow.

Fix: Have a separate from the side counting variable "i" floating point value, that represents the internal angle of the polygon.

In case my explanation was bad, here is some sample code: (controls Z, X, C)
```C
#include"raylib.h"

#define WIDTH  1000
#define HEIGHT 1000

int main(void)
{
    InitWindow(WIDTH, HEIGHT, "Regular polygons bug");
    SetTargetFPS(60);

    /* Default case */
    BeginDrawing();
    ClearBackground((Color){ 20, 20, 20, 255 });
    DrawPoly((Vector2){ WIDTH/2, HEIGHT/2 }, 7, 470.0f, 0.0f, (Color){ 30, 255, 30, 120 });
    EndDrawing();

    while (!WindowShouldClose())
    {
        BeginDrawing();
        if (IsKeyPressed(KEY_Z))
        {
            ClearBackground((Color){ 20, 20, 20, 255 });
            DrawPoly((Vector2){ WIDTH/2, HEIGHT/2 }, 7, 470.0f, 0.0f, (Color){ 30, 255, 30, 120 });   /* Has artifact (better visible with alpha) */
        }
        else if (IsKeyPressed(KEY_X))
        {
            ClearBackground((Color){ 20, 20, 20, 255 });
            DrawPoly((Vector2){ WIDTH/2, HEIGHT/2 }, 13, 470.0f, 0.0f, (Color){ 30, 255, 30, 120 });  /* Has artifact (better visible with alpha) */
        }
        else if (IsKeyPressed(KEY_C))
        {
            ClearBackground((Color){ 20, 20, 20, 255 });
            DrawPoly((Vector2){ WIDTH/2, HEIGHT/2 }, 23, 470.0f, 0.0f, (Color){ 30, 255, 30, 120 });  /* Has 24 instead of 23 sides */
        }
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### DIFF
--- a/src/shapes.c
+++ b/src/shapes.c
@@ -1183,7 +1183,7 @@ void DrawRectangleRoundedLines(Rectangle rec, float roundness, int segments, int
 void DrawTriangle(Vector2 v1, Vector2 v2, Vector2 v3, Color color)
 {
     if (rlCheckBufferLimit(4)) rlglDraw();
-    
+
 #if defined(SUPPORT_QUADS_DRAW_MODE)
     rlEnableTexture(GetShapesTexture().id);
 
@@ -1219,7 +1219,7 @@ void DrawTriangle(Vector2 v1, Vector2 v2, Vector2 v3, Color color)
 void DrawTriangleLines(Vector2 v1, Vector2 v2, Vector2 v3, Color color)
 {
     if (rlCheckBufferLimit(6)) rlglDraw();
-    
+
     rlBegin(RL_LINES);
         rlColor4ub(color.r, color.g, color.b, color.a);
         rlVertex2f(v1.x, v1.y);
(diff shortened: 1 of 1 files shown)


raysan5:
Hey @AlexHCC! Good catch! Thanks for the fix! :)

PR #973, opened 2019-09-21
contributed simple shader example

chriscamacho:
as title...


raysan5:
@chriscamacho thanks for the shader! :D

PR #976, opened 2019-09-24
Improve code readability

AlexHCC:
Thanks ray for accepting my bug fix! ([#972](https://github.com/raysan5/raylib/pull/972))
I would like to make some minor changes, mostly to improve code legibility.
In the src/shapes.c the DrawPoly() function:

- Change the variable name from internalAngle to centralAngle, since technically, the geometric definition of internal angle is [this](https://en.wikipedia.org/wiki/Internal_and_external_angles), and of central angle is [this](https://en.wikipedia.org/wiki/Central_angle).
- The centralAngle can be initialized with the value of zero. Although the functionality remains the same (as long as it is initialized with a 360/sides multiple), it makes more sense this way, and it also minimizes unnecessary computations.

### DIFF
--- a/src/shapes.c
+++ b/src/shapes.c
@@ -1298,7 +1298,7 @@ void DrawTriangleStrip(Vector2 *points, int pointsCount, Color color)
 void DrawPoly(Vector2 center, int sides, float radius, float rotation, Color color)
 {
     if (sides < 3) sides = 3;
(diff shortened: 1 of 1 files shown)


raysan5:
@AlexHCC thanks for the improvement! 😄

Issue #975, opened 2019-09-23
[question] LoadShaderCode non const parameters


heretique:
Hi, I'm wondering if there's a particular reason why the parameters to LoadShaderCode function are not const? Is this an oversight?

PR #977, opened 2019-09-25
Const corectness for LoadShaderCode

heretique:
This is related to issue #975


raysan5:
@heretique Thanks for the review!

PR #979, opened 2019-10-02
Add multi texture support for materials in GLTF format

Gamerfiend:


Gamerfiend:
This branch isn't ready to be merged in, but I would appreciate feedback

PR #980, opened 2019-10-03
Remove dead assignments

jubalh:
The result of `success` is actually never used. Either we should check
for it and return or remove it. I assume just checking the last one is
okay.


raysan5:
@jubalh oh! thanks for the review! that code is quite old! :D

jubalh:
:-)

PR #983, opened 2019-10-05
Update CMakeLists.txt

elitepleb:
CMake project version bump

### DIFF
--- a/projects/CMake/CMakeLists.txt
+++ b/projects/CMake/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.11) # FetchContent is available in 3.11+
 project(example)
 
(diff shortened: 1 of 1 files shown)


a3f:
Hello @elitepleb,

Thanks for the contribution! Could you add a comment why you think this is necessary?

elitepleb:
This change is not necessary or meaningful in any way.
I just noticed the old version number in the CMake project and incremented it.
If there is a problem with this pull request, please tell me so and i will close it.

a3f:
Hmm, could you change the line's comment to "Set this to the minimal version you want to support" or something along these lines? That would work for me.

a3f:
Hmm, seems the Github UI doesn't like it when you manually push pull requests. I've added a space, changed the commit message and pushed it. Thanks!

PR #990, opened 2019-10-21
changed Color struct to 4 floats

chriscamacho:
changed Color struct to 4 floats
GetRandomValue returns floats
added GetRandomInt
added Color4ub rgba 0-255
edited examples to work
colour palette values could bare a double check!

bit of an epic one this, there is could be some breakage!

### DIFF
--- a/examples/audio/audio_module_playing.c
+++ b/examples/audio/audio_module_playing.c
@@ -44,10 +44,10 @@ int main(void)
     {
         circles[i].alpha = 0.0f;
         circles[i].radius = GetRandomValue(10, 40);
-        circles[i].position.x = GetRandomValue(circles[i].radius, screenWidth - circles[i].radius);
-        circles[i].position.y = GetRandomValue(circles[i].radius, screenHeight - circles[i].radius);
-        circles[i].speed = (float)GetRandomValue(1, 100)/2000.0f;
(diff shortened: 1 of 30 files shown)


jamesl-github:
aren't you using British spelling instead of American ?
typedef struct Colour { 

and what is the reasoning for changing r, g, b, a to normalized floats ?
I MUCH prefer 0 to 255  
its the way I specify colors in html 
its the way SFML does it https://www.sfml-dev.org/documentation/2.5.0/classsf_1_1Color.php 
its the way SDL does it https://wiki.libsdl.org/SDL_Color?highlight=%28%5CbCategoryStruct%5Cb%29%7C%28SDLStructTemplate%29

chriscamacho:
its closed (by me) for a reason, it was a misunderstanding about how the internals of raylib should work after I got the wrong end of the stick during a chat with ray.

OpenGL (which raylib uses) uses normalized floats btw.... using colours 0-255 feels weird to anyone who's been used to using just GL and say GLFW on their own, just because other libs do something a particular way doesn't mean its the only way (look at setting shader colours in raylib for example)

I adopted the incorrect spelling to follow ray convention add
```      typedef Color Colour ```
in your own code if it pains you that much....

PR #997, opened 2019-10-25
updated RayMarching Demo

IndustriousNomad:
Added   - #define PLATFORM_DESKTOP line for desktop users.
            This now will correctly find the proper glsl version for the raymarching.fs file.
  Removed - Uniforms --> deltaTime and viewUp. Including the code that was setting them.
            They were never used and they were triggering a log warning.
  Removed - The const from both screenWidth and screenHeight.
            Now they can be used to update the shader resolution when screen is resized.
            NOTE : This is a quick fix and probably not the best idea.
  Added   - IsWindowResized() to check if screen is resized.
            If window is resized then width, height and shader resolution are updated.
  Changed - MIT tag at bottom right color value to BLACK. Now it's easier to see.

### DIFF
--- a/examples/shaders/resources/shaders/glsl100/raymarching.fs
+++ b/examples/shaders/resources/shaders/glsl100/raymarching.fs
@@ -8,8 +8,6 @@ varying vec4 fragColor;
 
 uniform vec3 viewEye;
 uniform vec3 viewCenter; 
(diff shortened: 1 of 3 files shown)


raysan5:
@IndustriousNomad thanks for the review!

PR #1002, opened 2019-10-27
new linux makefile for Android portages

jeangit:
It's for compiling/creating apk from linux.
The principles are the same than the windows Makefile, as it's based on this.


.gitignore
```diff
@@ -1,6 +1,8 @@
 # Ignore generated files
 # ...
 
+.gitignore
```
raysan5: I think it shouldn't be here...

raysan5:
@jeangit thanks! great addition, I added a minor review and for the moment I prefer to keep Windows named as `Makefile.Android`.

.gitignore
jeangit: Can you remove it, or do you need i do another PR with this file removed ?

raysan5: send another PR please, also revert the Makefile.Android rename...

jeangit: well i tried to do another PR, but git is not my cup of tea. Feel free to grab Makefile.Android.linux directly, if you want.

PR #1007, opened 2019-11-01
Fix generation of web examples

pamarcos:
At least on Linux, CMake seems to add automatically
the '-rdynamic' flag. As a result, Emscripten
generates a fat HTML file that does not work.
Removing it allows HTML+JS+WASM generation.


raysan5:
@pamarcos hey! thanks for the review! 😄

PR #1009, opened 2019-11-04
[rlgl] rlUpdateMesh now supports updating indices

brankoku:


raysan5:
@brankoku thanks for the review.

Why did you rename `numVertex` to `num`?

raysan5:
@brankoku thanks for the answer, makes sense. Merging!

PR #1010, opened 2019-11-07
[rlgl] Create function `rlUpdateMeshAt`

brankoku:
`rlUpdateMeshAt` allows the user to update a subset of a buffer, starting at the nth (`index`th) element.

I find myself frequently updating only certain regions of a buffer at a time, and having this function available helps performance tremendously.


raysan5:
@brankoku Nice addition, probably rlUpdateMesh() can use this one?

brankoku:
@raysan5 I updated the code. ;) `rlUpdateMesh` now calls `rlUpdateMeshAt` with an `index` of `0`. Additionally, `rlUpdateMeshAt` has been adapted so that during the case `index == 0`, `rlUpdateMesh`'s original bounds checking and `glBufferData` invocations are preserved.

raysan5:
@brankoku Thank you very much for the improvements! Merging! :)

PR #1011, opened 2019-11-08
[Core] Added ColorFromNormalized which is the reverse of ColorNormalize.

ChrisDill:
I based the name from `ColorFromHSV`. Not sure if it is the best name for it. 

Example usage
```c
Color color = ColorFromNormalized((Vector4){0.5f, 0.1f, 0.1f, 1.0f});
```

I am not sure if a cast should be added for converting the float to unsigned char.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1616,6 +1616,19 @@ Vector4 ColorNormalize(Color color)
     return result;
 }
 
+// Returns color from normalized values [0..1]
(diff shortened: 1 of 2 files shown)


raysan5:
@ChrisDill thanks for the contribution... I'll think about the name and also the consistency with the other functions... it's ok for now...

Issue #924, opened 2019-07-31
[camera] 3D camera not working correctly (maths)


Dacode45:
So I have my camera at 0,0,0 pointing to  a unit cube at 0,0,10 and the cube doesn't render on my mac. Ive found weirder versions of this bug, but this is the simplest I can replicate.
https://pastebin.com/FTRGTAQR

raysan5:
Issue is related to camera matrix maths, probably hitting some division by 0...

Using `camera.position = (Vector3){0.001f, 0.0f, 0.0f};`, it works ok.

yummydev:
So how do I fix this without checking that values are non zero every frame?

raysan5:
@yummydev I don't know, that module should be reviewed. Contributors are welcome!

Ushio:
@raysan5 
Hi, I also curious about the issue.
So I create the fix.

https://github.com/raysan5/raylib/pull/1013

raysan5:
@Ushio thank you very much for the fix!

PR #1013, opened 2019-11-10
Fix camera not working problem at align z direction. #924

Ushio:
Hi, about https://github.com/raysan5/raylib/issues/924
I fix the camera behavior. I noticed additional camera issues.
I think that there are 3 problems.

1. z-direction camera at initialization is not working. it is caused by SetCameraMode(camera, CAMERA_FREE); Because "cameraAngle.y" was nan after this function.

```
	Camera camera = { 0 };
	camera.position = Vector3 { 0.0f, 0.0f, 0.0f };
	camera.target = Vector3 { 0.0f, 0.0f, 10.0f };
	camera.up = Vector3 { 0.0f, 1.0f, 0.0f };
	camera.fovy = 45.0f;
	camera.type = CAMERA_PERSPECTIVE;
```

2. Camera rotation by mouse.y movement has a non-linear behavior when a camera angle close to align xz plane.

3. camera.position at start is not zero. it was flip 180 degrees unexpectedly.

```
	Camera camera = { 0 };
	camera.position = Vector3 { 0.0f, 0.0f, 0.0f };
	camera.target = Vector3 { 0.0f, 0.0f, 10.0f };
	camera.up = Vector3 { 0.0f, 1.0f, 0.0f };
	camera.fovy = 45.0f;
	camera.type = CAMERA_PERSPECTIVE;
```

So I fixed these issues.
https://pastebin.com/FTRGTAQR
by Dacode45 
start working by this fix.

# additional tips
For camera consistent behavior, the following code should work. 
It is good for debugging.

```
	// Main game loop
	while (!WindowShouldClose()) // Detect window close button or ESC key
	{
		// SetCameraMode(camera, CAMERA_FREE);
		// Update
		SetCameraMode(camera, CAMERA_FREE);
		UpdateCamera(&camera); // Update camera
        ....
```

### DIFF
--- a/src/camera.h
+++ b/src/camera.h
@@ -243,9 +243,9 @@ void SetCameraMode(Camera camera, int mode)
 
     cameraTargetDistance = sqrtf(dx*dx + dy*dy + dz*dz);
 
-	// Camera angle calculation
-	cameraAngle.x = atan2f(dx, dz);                   // Camera angle in plane XZ (0 aligned with Z, move positive CCW)
-	cameraAngle.y = atan2f(dy, sqrtf(dx*dx + dz*dz)); // Camera angle in plane XY (0 aligned with X, move positive CW)
+    // Camera angle calculation
+    cameraAngle.x = atan2f(dx, dz);                   // Camera angle in plane XZ (0 aligned with Z, move positive CCW)
+    cameraAngle.y = atan2f(dy, sqrtf(dx*dx + dz*dz)); // Camera angle in plane XY (0 aligned with X, move positive CW)
 
     playerEyesPosition = camera.position.y;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Ushio thanks for the fix! I haven't tested it yet... Just one concern, `SetCameraMode()` is expected to be called once at initialization, not every frame, I imagine it still works that way...

Oh, a small detail, please, could you replace TABS by 4 spaces? Sorry, it's a raylib convention...

Ushio:
Ah, Sorry, I replaced just now.

I know that SetCameraMode() is called at once usually.
But SetCameraMode(camera, CAMERA_FREE); do the following tasks. (I ignore playerEyesPosition. it seems isn't used at CAMERA_FREE )

- Get cameraTargetDistance from position and target
- Get cameraAngle from view direction.

and UpdateCamera() do

- reconstruct position and target from cameraTargetDistance and cameraAngle 

So I think it is an almost inverse operation. 
It is the reason that "SetCameraMode(camera, CAMERA_FREE); "should be able to be called multiple times. If the symmetry behavior is not work, It might be bug. 

From another perspective, SetCameraMode(camera, CAMERA_FREE); 
I think it is expected that the camera properties like position, direction..etc should be kept because of the name SetCameraMode(camera, CAMERA_FREE).

But actually other camera mode like CAMERA_THIRD_PERSON don't have consistent and symmetry behavior. 

I think it sounds great if the following code works well.

```
        if (IsKeyPressed(KEY_ONE)) {
            SetCameraMode(camera, CAMERA_FREE);
        }
        if (IsKeyPressed(KEY_TWO)) {
            SetCameraMode(camera, CAMERA_ORBITAL);
        }
        if (IsKeyPressed(KEY_THREE)) {
            SetCameraMode(camera, CAMERA_FIRST_PERSON);
        }
        if (IsKeyPressed(KEY_FOUR)) {
            SetCameraMode(camera, CAMERA_THIRD_PERSON);
        }
```

But it looks hard work. This update is the first step of this.

raysan5:
@Ushio Thank you very much for the review! :D

PR #1018, opened 2019-11-15
Fix Makefile missing X11 lib when building as a shared library on Linux

JuDelCo:


raysan5:
@JuDelCo thanks for the fix! 😄

PR #1019, opened 2019-11-17
Add DrawPoint3D() function to models.c

ProfJski:
Using DrawCube() or DrawSphere() to represent points in 3D uses many vertexes, which limits FPS when plotting many points (100K-1M).  DrawPoint3D() renders a "point" using a small line, on analogy to how the 2D function DrawPoint() is written.  Uses only 2 vertexes, so much more efficient than DrawCube() which uses 36, or DrawSphere which uses (rings+2)*slices*6 vertexes.


ProfJski:
Two downsides to this approach:
1) The size of the small line is arbitrarily small (0.1), so it can be seen as a line on closer zooms.
2) When the camera is oriented parallel to the small line, along the x axis, the points which lie directly in front of the camera and far from it tend to fade because they are viewed "end on."  A possible remedy is to use 4 vertexes so that the point is equally visible at all orientations, drawing three lines: {0,0,0} to {.1,0,0} to {.1,.1,0} to {.1,.1.1}, which is still cheaper than drawing a cube or a sphere.

For reasons I don't understand, if I write 
    if (rlCheckBufferLimit(8)) rlglDraw();
the points plot appropriately, but if the value of rlCheckBufferLimit() is 2 or 4, undesired scaling is introduced, even though the primitive uses just two rlVertexf() calls.

chriscamacho:
just a thought, could you use a billboard approach and vary the distance given the camera distance ?

ProfJski:
A good idea.  It would certainly solve the visibility of the line at close distances, and it might help with the points vanishing when viewed parallel to the camera because distant lines would be slightly bigger, and thus they would need to be much closer to zero degrees to the camera to vanish.  I'd like to benchmark the performance impact of the distance calculation, however, since I proposed this function to increase performance in rendering about 1 million objects.  I don't know if there's a way to avoid taking a square root in the distance function, since I'm guessing the size scales with distance linearly, so distance squared by itself wouldn't help.

Unfortunately, I have to work tonight and this week will be busy, but I tend to try things out at odd hours to wind down before bed.  I'll see what I can benchmark with your approach -- maybe it won't be an issue.

Other than that, if this were integer math, could one approximate the distance factor well enough by calculating distance squared, then bit shift right by half the value of the msb?  I presume that won't work on float, so I'd try casting it to int and back again.

chriscamacho:
It might be better to do this in the shader, and transform the two verts there, as you'll have the info you need to hand...

ProfJski:
If you have some code to suggest, I would welcome it.  I only have a basic idea of the OpenGL pipeline and haven't coded beyond fragment shaders, so I don't know the appropriate matrix functions to call on the vertexes to do the scaling.

ProfJski:
This is all I'll be able to do tonight, but I thought about using camera info both to scale the size of the small line AND to orient it so that it never is parallel to the angle of view of the camera.  Downside of the following code is that it has two square roots (Length function and Normalize function).  I can figure out how to economize more later.  For now, the resulting frame rate is 1/3 of the pull request code.  I can do 1,000,000 "points" at 20FPS in the proposed code above, and only about 6 FPS with the following kludgy code:

```
void DrawPoint3D(Camera camera, Vector3 pos, Color color) {
    float scale=0.05*Vector3Distance(camera.position,pos);
    scale=(scale<0.05)?0.05:scale;
    Vector3 normal=Vector3Normalize(Vector3Perpendicular(Vector3Subtract(camera.position,camera.target)));
    normal=Vector3Scale(normal,scale);
    if (rlCheckBufferLimit(8)) rlglDraw();
    rlPushMatrix();
    rlTranslatef(pos.x,pos.y,pos.z);
    rlBegin(RL_LINES);
        rlColor4ub(color.r, color.g, color.b, color.a);
        rlVertex3f(0.0,0.0,0.0);
        rlVertex3f(normal.x,normal.y,normal.z);
    rlEnd();
    rlPopMatrix();

return;
}

```
The result looks decent enough.  I don't know the comparative performance of atan2 vs. sqrt, but an alternative way might be to take the angle between camera and an axis and use that to tilt the line.  That leaves only 1 sqrt for scaling.

chriscamacho:
alas I too am pushed for time at the moment, all I can do is post advise and help maybe if you get really stuck, have a look at the shader sample I contributed (shader_mask) 

also take a look at this https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming
I'd have a buffer of 3d coordinates (two per 3d point - both set the same) for GL_LINES

In the vertex shader take into account the Z and push one vert up and the other down

I do know that some implementations of GL_LINES don't respect point size so you'll still need to effect the Y axis to make sure they are all visible (afaik)

using this method would also allow you to have a separate colour buffer that you set but never change, so each point can have its own different colour....

oh just one point if you do use shaders then that means it won't work on GL1.0 but I'm not sure there is any hardware out there that can't do GLES2.0 at least, but could still do 1,000,000 points.....

so a GLES2.0 and GL3.0 shader should hopefully cover everyone ???

raysan5:
I'm thinking about this approach... not sure if using LINES is the best option... but probably the easiest solution for the moment. In any case, merging it as a possible solution.

Just note that function is no yet exposed on `raylib.h`

ProfJski:
Hey @raysan5 ,

I did some further exploration this week with an eye to keeping it maximally performant, and here was my best compromise solution between visibility at all angles and FPS: drawing two small lines, at an angle to each other that spans all three axes, so that no matter what the view, no line is end-on to the camera.  All I did was add 2 more vertex calls.  Timing it with 1M calls of DrawPoint3D using two vertexes (=original code) I get about 20 FPS, and adding two more vertexes in the following code only brings me down to about 15FPS, on an Intel Skylake with Radeon RX580.  Looks good at all angles.
```
void DrawPoint3D(Vector3 pos, Color color) {
    if (rlCheckBufferLimit(8)) rlglDraw();
    rlPushMatrix();
    rlTranslatef(pos.x,pos.y,pos.z);
    rlBegin(RL_LINES);
        rlColor4ub(color.r, color.g, color.b, color.a);
        rlVertex3f(0.0,0.0,0.0);
        rlVertex3f(0.1,0.0,0.1);
        rlVertex3f(0.0,0.0,0.0);
        rlVertex3f(0.1,0.1,0.0);
    rlEnd();

    rlPopMatrix();

return;
}
```
Here are the other methods that I explored, which involved rotating the line so that it always faces the camera:

Code in comments above: too many vector ops, including two square roots.  Looks good, but FPS drop from 20 to 6.  Too expensive.

Compute angle between camera and object projected onto XZ plane using dot(A,B) = Len(A)Len(B)cos(theta).  As slow because of two square roots.  

Take simpler computation of theta from theta=atan2((b.z-a.z),(b.x-a.x)) or using trigonometry identities to eliminate square root from dot product formula squared.  Faster at 9FPS, but still a small blind spot I presume because I'm projecting onto XZ plane rather than getting the true angle.

So that's as far as I got with alternatives.  Nothing worked as fast as simply adding two more vertexes to the call, which still keeps the total number of vertexes pushed much lower than using cubes or spheres, and gives decent "point clouds" for plots.

So the end user is not mislead by the function name into thinking they will get a true point, I wondered if calling the function "DrawHashMark3D" or leaving a comment in function declaration would be useful.

I don't know enough about underlying possibilities in OpenGL primitives, so I just stuck with RL_LINES because using triangles or quads would still result in a plane figure that would vanish when viewed edge-on.  I'm not knowledgeable enough to implement what @chriscamacho proposed, and it seemed to require going around the framework that most shapes.c or models.c functions used, so I didn't try it.

Best,
ProfJski

PR #1022, opened 2019-11-22
fix example build on macos

electronstudio:


raysan5:
As per the CI system, macOS building was working ok without those libraries linking.

What version of macOS are you using?
What linkage errors do you get?

electronstudio:
MacOS 10.14.6

[error.txt](https://github.com/raysan5/raylib/files/3883488/error.txt)

raysan5:
ok, thanks for the info! Merging.

PR #1026, opened 2019-11-24
Comments

piecedigital:
While setting up input recognition I realized that the comments for the face buttons were flipped. So I flipped them around.

Also reorganized the PS3 and XBOX symbols to be in the same order, corresponding with the order of the enums.


raysan5:
@piecedigital thanks for the review!

PR #1029, opened 2019-12-02
Add GetRaylibVersionString() to core.c and raylib.h

aldrinmartoq:
While writing my [raylib-ruby](https://github.com/a0/raylib-ruby) wrapper, I was thinking it would be useful to programatically get the current raylib version, but there was no function like that in the API. So I wrote this one.

```c
RLAPI const char*GetRaylibVersionString();
// returns "2.5" from RAYLIB_VERSION #define
```

I'm not sure the name of functions is the best, but I guess is enough to start. There is a sample in examples/core_basic_window.c

### DIFF
--- a/examples/core/core_basic_window.c
+++ b/examples/core/core_basic_window.c
@@ -20,6 +20,7 @@
 ********************************************************************************************/
 
 #include "raylib.h"
+#include <stdio.h>
 
 int main(void)
 {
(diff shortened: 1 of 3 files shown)


raysan5:
@aldrinmartoq thanks for the proposal but no plans to add that kind of function for the moment.

PR #1034, opened 2019-12-04
CMake - Fix usage of GLFW. Use static library instead of object files.

las3rlars:
Should fix the: $<TARGET_OBJECTS:glfw_objlib> issue

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -35,9 +35,6 @@ if(NOT glfw3_FOUND AND NOT USE_EXTERNAL_GLFW STREQUAL "ON" AND "${PLATFORM}" MAT
   set(GLFW_USE_WAYLAND ${USE_WAYLAND} CACHE BOOL "" FORCE)
 
   add_subdirectory(external/glfw)
-
-  list(APPEND raylib_sources $<TARGET_OBJECTS:glfw_objlib>)
-  include_directories(BEFORE SYSTEM external/glfw/include)
 else()
   MESSAGE(STATUS "Using external GLFW")
   set(GLFW_PKG_DEPS glfw3)


a3f:
Hello @las3rlars,

Thanks for the pull request, but this is the wrong way to go about it. We're (or rather were) using objlib for good reason: to be able to reuse the GLFW build system to build a static GLFW that can be linked into a static raylib. The Makefile sidesteps the GLFW build system and builds it itself, but for the CMake setup I chose to reuse the GLFW subsystem, so we gain automatic support for stuff like Wayland.
 
The missing bits should be reinstated. Deleting references to objlib doesn't help. I'll draft up a pull request shortly and ping the GLFW pull request, that has been sitting unmerged for 1,5 year so far.

a3f:
I've reinstated the deleted GLFW changes (again) and CI is green again now.
@raysan5, the way I'd update the `src/external/glfw` directory:

* check out the branch at https://github.com/glfw/glfw/pull/1307
* git rebase master
* fix merge conflicts
* build GLFW on its own
* copy it over to `src/external/glfw`

This way you don't delete the changes inadvertently. I've repinged my GLFW pull request.

raysan5:
@a3f thank you very much for fixing it... again! I'll take more care in the future! 👍😊

las3rlars:
Hi @a3f

I'm a little confused. The change I suggested would still use the cmake build system.

'add_subdirectory(external/glfw)'

'target_link_libraries(raylib_static glfw ${GLFW_LIBRARIES} ${LIBS_PRIVATE})'

The above lines will compile glfw as a static library and then use that library when linking the raylib_static library.

I might have missed something. And I'm not sure if emscripten and those things requires things done differently. @raysan5 had a valid point if the same symbols are included when linking with the static library. Checking with objdump or something could maybe be valuable.

Anyway, was just curious

Best regards

a3f:
AFAIR there were three problems:

- We don't want to probe for libraries that are already GLFW dependencies, so we need GLFW to tell us about the dependencies it build system had determined
- We want to reexport GLFW symbols from raylib, same as the Makefile does.
- We don't want to build GLFW ourselves, but rather use the GLFW build system, unlike the Makefile

The first one is solved by the first  patch in the referenced GLFW pull request. The second one is solved by the second patch in the referenced GLFW pull request. Your patch only attempted solving the third point.

las3rlars:
Ah, thanks! Now I'm following!

Issue #1035, opened 2019-12-09
[core] SetWindowMonitor does nothing.


DropsOfSerenity:
The current code for `SetWindowMonitor` only `TraceLog`'s out some stuff, why are we not using `glfwSetWindowMonitor`? It seems to be commented out.

The minimal implementation imo would be, if we are setting monitor to a different screen, obtain the screen mode/resolution, and pass them into `glfwSetWindowMonitor`.

E.g something like...

```c
const GLFWvidmode* mode = glfwGetVideoMode(monitors[monitor]);
glfwSetWindowMonitor(window, monitors[monitor], 0, 0, mode->width, mode->height, mode->refreshRate);

```

raysan5:
@DropsOfSerenity Ouch! Good catch! I probably left it commented because it was not working properly and I forgot about it...

raysan5:
@DropsOfSerenity Feel free to send a PR if you want

DropsOfSerenity:
Will do @raysan5 my custom raylib build currently works for me on linux, but I'll want to test it on windows as well later today.

DropsOfSerenity:
Pull request #1036 merged, closing issue.

PR #1036, opened 2019-12-09
SetWindowMonitor made functioning again.

DropsOfSerenity:
* We get the video mode from the target monitor and use that to set, therefore
keeping windowed-fullscreen
* Added a GLFW_AUTO_ICONIFY 0 hint so that glfw does not minimize the window
when changing focus from a windowed fullscreen window. This is more expected
behavior for windowed full screen, when a user alt-tabs or clicks on a window in
the other monitor, they expect their windowed-fullscreen window to remaining up.

This would fix Issue #1035 

Points worth thinking about for the future:

* If the monitors are of different resolutions, can the user manage the framebuffer in a sensible way? E.g letterboxing/scaling

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -933,8 +933,9 @@ void SetWindowMonitor(int monitor)
 
     if ((monitor >= 0) && (monitor < monitorCount))
     {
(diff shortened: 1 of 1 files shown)


raysan5:
@DropsOfSerenity Thank you very much for the improvements! :D

PR #1040, opened 2019-12-18
fixed free camera pitch rotation

chriscamacho:
pitch rotation was only being done on one axis
this allows you to look straight up and down, other camera modes will need this method implementing...

### DIFF
--- a/src/camera.h
+++ b/src/camera.h
@@ -167,8 +167,8 @@ void SetCameraMoveControls(int frontKey, int backKey,
 // FIRST_PERSON
 //#define CAMERA_FIRST_PERSON_MOUSE_SENSITIVITY           0.003f
 #define CAMERA_FIRST_PERSON_FOCUS_DISTANCE              25.0f
-#define CAMERA_FIRST_PERSON_MIN_CLAMP                   85.0f
-#define CAMERA_FIRST_PERSON_MAX_CLAMP                  -85.0f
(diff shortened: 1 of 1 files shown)


raysan5:
Hi @chriscamacho not sure if I understand the issue... please, could you provide an example? or can I check the issue in some of the existing examples?

chriscamacho:
On Sun, 29 Dec 2019 03:23:14 -0800
Ray <notifications@github.com> wrote:

> Hi @chriscamacho not sure if I understand the issue... please, could
> you provide an example? or can I check the issue in some of the
> existing examples?
> 

in order to rotate through one axis your calculation must effect the
other two axes.

an X axis rotation effects Y & Z for example

Your 2 axis rotations are only effecting 3 axes not 4 

Z heading should effect X & Y 
Y pitch should effect X(again) and Z

You can see the current calculation is incorrect if you take off the
pitch limit, instead of the pitch going around a full 360 it pitches up
and down in a sine wave like motion, this means you can only look down
45 degrees (not straight down or even 5 degrees off).  You need to
either add an additional calculation OR just pick the result out of a
matrix like the PR shows you.

raysan5:
@chriscamacho ok, I understand, previous implementation limited the pitch when looking around, probably related to the Euler math. Merging.

chriscamacho:
please note I only had chance to implement it on one mode, you'll need to review the other places with the same issue....

PR #1045, opened 2019-12-20
Forcibly ensure .xm playback starts in the right place; fixes #1043.

### LINKED ISSUE #1043: [audio] .xm song playback starts in the wrong place for modules that loop on a pattern other than the first.
### Issue description

When starting playback of an .xm-format module where the restart point is anything other than the zeroth entry in the pattern table, the song incorrectly starts playing at the restart point, rather than the beginning of the song.

### Environment

- **Platform:** Linux (4.15.0-65-generic #74~16.04.1-Ubuntu SMP Wed Sep 18 09:51:44 UTC 2019 x86_64)
- **OpenGL & GPU version (as detected by Raylib):** INFO: GPU: Renderer: NVS 4200M/PCIe/SSE2 , INFO: GPU: Version:  3.3.0 NVIDIA 384.130 

### Issue Screenshot

n/a (audio-only problem)

### Steps To Reproduce:

[XM audio file that triggers the problem](https://github.com/raysan5/raylib/files/3989983/raylib-xm-playback-testcase.zip)

1. Open and play the attached file in MilkyTracker; carefully observe the sound of the introduction and where playback begins.
2. Again, in MilkyTracker, notice that the restart point is set to something other than the zeroth pattern.
3. Using [audio_module_playing.c](https://github.com/raysan5/raylib/blob/master/examples/audio/audio_module_playing.c), load and play the same file; again, carefully observe where playback begins.

Notice that the intro to the song is not played.

illegalinstruction:


illegalinstruction:
Note that even with this patch in place, some files can still exhibit a 'chirp' of the wrong pattern playing right at the beginning; further examination and modifications to `jar_xm` will be needed to fix that.

raysan5:
@illegalinstruction thanks for the review! 👍😄

PR #1052, opened 2019-12-28
Normals was not correctly updated when using animated meshes

las3rlars:
* Normal vbo not correctly updated


Gamerfiend:
Glad you found a fix for this :)

raysan5:
@las3rlars Nice catch! Thanks for the fix! :D

PR #1057, opened 2020-01-04
add TextIsEqualCase and use in IsFileExtension

MyLegGuy:
this is important because LoadImage uses IsFileExtension.


raysan5:
@MyLegGuy Thanks for the review but I prefer avoiding a new function.

I just modified `IsFileExtension()` to use `TextToLower()` on checking. Commit https://github.com/raysan5/raylib/commit/358ca7595dabf2e29733ffd8b7931319ca44d20e

PR #1058, opened 2020-01-05
Changed minimum vertices for circle shapes from 4 to 3 to allow drawing of triangles

MikeDX:
This PR reduces the minimum vertices of circles from 4 (square) to 3 (triangle) to allow drawing of triangle primitives (filled and outline)

### DIFF
--- a/src/shapes.c
+++ b/src/shapes.c
@@ -215,7 +215,7 @@ void DrawCircleSector(Vector2 center, float radius, int startAngle, int endAngle
         endAngle = tmp;
     }
 
-    if (segments < 4)
+    if (segments < 3)
     {
         // Calculate how many segments we need to draw a smooth circle, taken from https://stackoverflow.com/a/2244088
         #define CIRCLE_ERROR_RATE  0.5f
@@ -307,7 +307,7 @@ void DrawCircleSectorLines(Vector2 center, float radius, int startAngle, int end
         endAngle = tmp;
     }
 
-    if (segments < 4)
+    if (segments < 3)
     {
         // Calculate how many segments we need to draw a smooth circle, taken from https://stackoverflow.com/a/2244088
         #ifndef CIRCLE_ERROR_RATE


raysan5:
@MikeDX I'm not implementing this change, actually triangles can be already drawn with `DrawCircle()`. This function is specific for sector drawing, not intended for triangles.

MikeDX:
That is a real shame, since it is really useful and costs nothing to implement and is useful 
Screenshot 2020-01-02 at 09 52 07 for instance for this

raysan5:
@MikeDX Why not using `DrawCircle()` to do the same?

MikeDX:
I don't believe drawcircle will do just outlines nor will it do circles of varying certices

If I'm wrong though, I'm happy to use that!

Otherwise my project will have to use a forked raylib :(

raysan5:
@MikeDX ok, I see. What about `DrawRing()`, `DrawRingLines()` or `DrawPoly()`?

MikeDX:
My current implementation is this:

```c
int F_circle(int x, int y, float radius, int points, Flua_Colour col, int outline) {
	Vector2 center = {x, y};
	
	if(outline) {
		DrawCircleSectorLines(center, radius, 0, 360, points, col);      // Draw a piece of a circle
	} else {
		DrawCircleSector(center, radius, 0, 360, points, col);      // Draw a circle sector outline
	}

	return 0;
}
```

Does Draw* give me the same ability in the same simple way?

raysan5:
@MikeDX I think `DrawPoly()` let you draw that kind of shapes, actually it's intended for that. `DrawCircleSector()` was intended for other kind of shapes.

MikeDX:
DrawPoly works great for a filled shape but drawRing falls back to circle at 3 points

raysan5:
Ok, so a `DrawPolyLines()` is required.

I checked `DrawRing()` and it uses `DrawCircleSector()` internally.

`DrawRing()` and `DrawCircleSector()` were added by a raylib user, I will review them again.

MikeDX:
If you can add a DrawPolyLines, that will make me happy

I still think reducing the minimum sectors is a zero cost addition, since 3 is no less valid than 4

PR #1067, opened 2020-01-14
Fix for short non-looping sounds

jbosh:
Short non-looping sounds can sometimes think they need to keep playing and will output their first few frames again. This helps to break out of all the loops instead of just this one.


raysan5:
@jbosh thanks for the fix! merging!

PR #1071, opened 2020-01-15
Makes tracker music files (mod, xm) play to completion and loop in the right spot, fixes #989 

### LINKED ISSUE #989: [audio] Playing only half of module (bug)

### Issue description
After simple playing example `audio_module_playing` realizing that only half of music module is playing (30s). Original module in sources resources/mini1111.xm is 62 seconds long. 

When changing examples/audio/audio_module_playing to play old module:

    Music music = LoadMusicStream("resources/chiptun1.mod");

Realized that GetMusicTimeLength(music) returns 3.78 
but should be 7.56 (Total samples/Sample rate)

INFO: [resources/chiptun1.mod] Music file successfully loaded:
INFO:    Total samples: 363007
INFO:    Sample rate: 48000 Hz
INFO:    Sample size: 16 bits
INFO:    Channels: 2 (Stereo)

### Environment
Linux

illegalinstruction:
This fixes #989 by changing the end-of-stream logic for tracker music formats. 

Because the samples are generated on the fly, rather than coming from a stream, we don't need stop and reset playing audio if we've hit what we think is the end and want to loop; if we 'blindly' keep playing, the module player will continue to generate the samples one would expect to hear, no buffer underruns or crashing occur, etc.

This seems like a hacky way of fixing it (and I welcome any suggestions on how to improve it) - there might be a more elegant solution - but it does stop the problem from happening.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1481,8 +1481,19 @@ void UpdateMusicStream(Music music)
 
     // Reset audio stream for looping
     if (streamEnding)
-    {
-        StopMusicStream(music);        // Stop music (and reset)
+    {	
+		// is this a tracker format or a streamed one (ogg, wav, etc.)	
+		if (!((music.ctxType == MUSIC_MODULE_XM) || (music.ctxType == MUSIC_MODULE_MOD)))
+		{
+			// streamed - stop the music and reset it.
+			StopMusicStream(music);
+		}
+		else
+		{
+			// tracker - don't reset it, but remember that we have hit 
+			// the loop point and are playing normally again.
+			streamEnding = false;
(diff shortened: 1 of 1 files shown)


illegalinstruction:
**Caveat:**
Note that `GetMusicTimeLength()` will still return half the expected length - this patch doesn't attempt to fix that.

raysan5:
@illegalinstruction thanks for the review but it is a hacky solution, sample count has to be properly processed to avoid the issue and keep all the functionality working.

illegalinstruction:
Understood; I'll try again if I have some time this afternoon/weekend.

PR #1075, opened 2020-01-17
 Fix compilation when SUPPORT_MESH_GENERATION not supported

Elkantor:
The loadModel function used the GenMeshCube to generate a default cube mesh, if no model could be load.
Now, if the SUPPORT_MESH_GENERATION feature is not supported, we don't load a default cube mesh, but we simply warn the user that no meshes can be loaded, and we can't create a default mesh because the feature is not enabled.

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -664,11 +664,14 @@ Model LoadModel(const char *fileName)
 
     if (model.meshCount == 0)
     {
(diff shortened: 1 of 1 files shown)


raysan5:
@Elkantor thanks for the fix! 😄

PR #1092, opened 2020-02-04
Update to miniaudio 0.10

mackron:
This pull request contains updates to support version 0.10 of miniaudio which includes some breaking API changes affecting raylib's audio system. The only mandatory changes required here is to replace `ma_pcm_converter` with `ma_data_converter`. This is used to convert `AudioBuffer` data to a format suitable for mixing.

I've also removed the manual master volume control and replaced it with miniaudio's built-in volume control (`ma_device_set_master_volume()`).

You will notice I moved the windows.h hacks out of your modified miniaudio.h and put it into raudio.c. I've just done this to make it easier to update miniaudio.h while this testing phase is in progress. I was wondering if perhaps it would be better to place this into a file called `raudio_windows.h` or something just to keep it isolated? I didn't do this - just a thought.

I've only done some very quick tests so you may want to give this a bit of testing yourself before merging.


raysan5:
Hi @mackron! Thank you very much for the review! I'm merging and doing some testing! :)

About that window.h stuff... I'll try to find a place for it... for the moment it's ok placed in the `raudio.c`.

PR #1097, opened 2020-02-11
Fixed memory leaks in textures.c and added jetbrains/clion project-files to gitignore.

SasLuca:
Pretty simple changes, usually involved just adding a call to `RL_FREE`.

For `ImageToPOT` I put the allocation of the `pixels` buffer in the body of the if-statement because if the condition was false the array was allocated for nothing and it would also leak.


raysan5:
@SasLuca thanks for the review!

Issue #1025, opened 2019-11-24
[models] glTF material colours?


electronstudio:
Create object in Blender with material base colour set to red:

![image](https://user-images.githubusercontent.com/7192452/69494558-0f7de600-0eb5-11ea-9c09-a33d486e9f2c.png)

Export to glTF: 
[coloured.gltf.zip](https://github.com/raysan5/raylib/files/3883539/coloured.gltf.zip)

This how the material shows up in glTF file:

```
"materials" : [
        {
            "doubleSided" : true,
            "name" : "Material",
            "pbrMetallicRoughness" : {
                "baseColorFactor" : [
                    0.8000000715255737,
                    0,
                    0.00022257522505242378,
                    1
                ],
                "metallicFactor" : 0,
                "roughnessFactor" : 0.4000000059604645
            }
        }
    ],
```

Raylib displays this model as white (if no texture is applied).

![image](https://user-images.githubusercontent.com/7192452/69494861-794bbf00-0eb8-11ea-81c1-3c202171b575.png)

The colour data is loaded, but it is into model.materials.maps[MAP_ROUGHNESS ] rather than model.materials.maps[MAP_DIFFUSE]. 

Should I be using a shader to get it to display the colour?

electronstudio:
@Gamerfiend

Gamerfiend:
So the color data is loaded into Raylib? But it's in the wrong map? Hmm.. I can take a look at the code and see if I can repro.

raysan5:
@electronstudio @Gamerfiend glTF support still requires some work.

electronstudio:
Any progress on this?  I don't really understand why it was closed if it still requires work?

raysan5:
@electronstudio I think nobody is working on this issue, no intention to work on it from my part. I'm reopening it, feel free to contribute.

electronstudio:
I understand that it's an open source community project so people will only work on what they need, and if I had time I would like to help, and if I had money I would like to donate... but I'm still confused by the model loading issues.

I think people are using Raylib to make games?  And doesn't every game require the ability to import models?  It seems like this should break the workflow of everyone who uses Raylib?

Sorry for complaining when I have no right to complain!

raysan5:
@raysan5 raylib was primary intended for education, on a second place intended for small-medium 2D games (that do not require 3D models), never intended to be a full featured 3D engine, for that kind of projects I recommend professional engines like Unity or Unreal.

electronstudio:
[...]

PR #1108, opened 2020-02-22
Expand GLTF Model support

Gamerfiend:
This PR adds a number of changes to the code of GLTF model loading, as well as a number of features to the support list.

- Models that use a base color instead of an albedo texture, should load that color value in the color field of the MaterialMap struct. Should address https://github.com/raysan5/raylib/issues/1025

- Occulsion texture are loaded and placed into their proper struct field

- Emissive texture and color value (factor) are loaded and placed into their proper struct field


raysan5:
@Gamerfiend thank you very much for the improvement! Maybe I'll review some variables names... :)

PR #1113, opened 2020-02-28
[rlgl] `LoadText()` fix

brankoku:
This fix guarantees the string loaded from a file is zero-terminated. The old code:
```C
// WARNING: \r\n is converted to \n on reading, so,
// read bytes count gets reduced by the number of lines
if (count < size)
{
    text = RL_REALLOC(text, count + 1);
    text[count] = '\0';
}
```
does not zero-terminate the string if `count == size`, which can very well be the case if the text file was written in an editor that uses `\n` instead of `\r\n` (basically every text editor on Linux).

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -3013,8 +3013,10 @@ char *LoadText(const char *fileName)
                 if (count < size)
                 {
                     text = RL_REALLOC(text, count + 1);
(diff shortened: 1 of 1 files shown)


raysan5:
@brankoku oh! true! thanks for the fix!

PR #1118, opened 2020-03-01
Update raymath.h

i-right-i:
Added Vector2Rotate Function.


raysan5:
@i-right-i thanks for the contribution! :)

PR #1120, opened 2020-03-03
add `LoadImageBuffer`

brankoku:
`LoadImageBuffer` allows the user to load an image from data that already exists in memory, such as raw PNG data they have embedded into their application.

I also have a version of this for raylib 2.6, but is there any reason to backport it? Does raylib 2.6 support any platforms that raylib 3.0 will not?


raysan5:
@brankoku Actually, right now I'm working on that direction, it requires lot of redesign so not merging for the moment. Still, I'll keep this PR open for reference.

brankoku:
@raysan5 Very good to hear! If I can make a suggestion, it would be ideal if `LoadDDS` and friends were all adapted to work from memory. Then, `LoadImage` can simply load the file into memory and pass the data into `LoadImageBuffer`. This reduces redundancy and makes all formats usable as both external files and pre-loaded buffers.

raysan5:
This improvement is on the plan. Closing PR.

PR #1122, opened 2020-03-08
Changed `if` to `else if`.

SasLuca:
I think that the `if` statement on line `3121` should be `else if` because it would cause a memory leak if both branches could be hit in the same call because they both do an allocation stored in `image.data`.


raysan5:
@SasLuca thanks for the review! 😄

PR #1123, opened 2020-03-08
Always use latest raylib version for cmake

G3bE:
Before all CMakeLists.txt in examples and tests used a specific version of raylib, which would have been needed to be upgraded. This is fixed in this PR. Also sorry for three commits instead of one

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -96,7 +96,7 @@ endif()
 include_directories(BEFORE SYSTEM others/external/include)
 
 if (NOT TARGET raylib)
-  find_package(raylib 2.0 REQUIRED)
+  find_package(raylib REQUIRED)
 endif()
 
 # Do each example
--- a/games/CMakeLists.txt
+++ b/games/CMakeLists.txt
@@ -18,7 +18,7 @@ if(${PLATFORM} MATCHES "Web")
 endif()
 
 if (NOT TARGET raylib)
-  find_package(raylib 2.0 REQUIRED)
+  find_package(raylib REQUIRED)
 endif()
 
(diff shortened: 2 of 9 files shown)


raysan5:
@G3bE merged! thanks for the review! 😄

a3f:
This doesn't make sense. We don't use the `EXACT` option, so this only species a minimal version. What problem are you trying to solve?

raysan5:
@a3f thanks for commenting, I probably merged it too quickly... probably it requires further review...

G3bE:
The cmake documentation says that `The [version] argument requests a version with which the package found should be compatible`, which doesn't quite sound like minimal version to me. If I'm wrong, feel free to revert.

a3f:
> The cmake documentation says that The [version] argument requests a version with which the package found should be compatible, which doesn't quite sound like minimal version to me.

Check [`raylib-config-version.cmake`](https://github.com/raysan5/raylib/blob/master/cmake/raylib-config-version.cmake), which establishes the compatibility rules for the package.

Does this not work for you? Your commit message sounds like you run into a problem. If so, which?

G3bE:
No, I didn't run into any problem. As I said: if I'm wrong feel free to revert. I'm sorry for any inconvinience caused.

a3f:
> No, I didn't run into any problem. As I said: if I'm wrong feel free to revert.

You said something conflicting in your commit message, that's why I asked. 

> I'm sorry for any inconvinience caused.

No worries. Don't hesitate to open new PRs/issues in future if you run into anything looking fishy :)

raysan5:
@G3bE @a3f thanks to you both for the review.

PR #1126, opened 2020-03-09
Revert "Always use latest raylib version for cmake"

a3f:
We didn't use the latest version, but a compatible version, which is established by  [`raylib-config-version.cmake`](https://github.com/raysan5/raylib/blob/master/cmake/raylib-config-version.cmake).

It's ok to have a minimal version, so revert the raysan5/raylib#1123.


raysan5:
@a3f ok, thanks for the review

PR #1132, opened 2020-03-15
Android: Set CORE.display.[width, height] before screen size check

Doy-lee:
In `InitGraphicsDevice(...)`, the Android section has a screen security check
like other platforms- but CORE.display.width, CORE.display.height are
not set yet, so the security check sets it to 0. So ensure we query the
device's screen width and height before the screen size security check.

This also gives you the ability to run a proper full-screen application
on Android without any scaling and guess work on the target device by
setting screen width and height to 0 in `InitWindow(...)` and using
`GetScreen[Width|Height]()` to get the actual values.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -3191,9 +3191,6 @@ static bool InitGraphicsDevice(int width, int height)
 #if defined(PLATFORM_ANDROID)
     EGLint displayFormat;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Doy-lee Ok, it looks nice! :)

PR #1133, opened 2020-03-16
build-windows.bat fixes

henriquel1997:
- Adding the option to change the target platform(x86/x64)
- Generating the .pdb file on debug builds
- Correcting comments that called the script 'windows-build.bat' instead of 'build-windows.bat'


raysan5:
@henriquel1997 thanks for the improvement!

PR #1137, opened 2020-03-19
Android: Track touch motion across screen

Doy-lee:
This was working in 2.6 but no longer does in current git tree.
It appears touch position is only tracked on
AMOTION_EVENT_ACTION_[DOWN|UP], which only registers the initial touch
on the screen. Subsequent movement is not tracked into CORE.

Touch position and the Gesture System appears to be updated twice in
AndroidInputCallback in what looks like perhaps a copy paste error (code
is identical) with the exception of tracking AMOTION_EVENT_ACTION_UP in
the 2nd copy of the code (but this is not necessary to track).

If you need to track the first touch or release touch position, you can
do so with checking IsMouseButton[Pressed|Released] on the same frame.

This patch makes it so the touch position is always updated, and merges the
duplicated code into 1 singular code path.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -4272,54 +4272,6 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
             // Stop processing gamepad buttons
             return 1;
         }
-
-        int32_t action = AMotionEvent_getAction(event);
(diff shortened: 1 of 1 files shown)


raysan5:
@Doy-lee thanks for the review.

PR #1142, opened 2020-03-23
fog/lighting shaders (for Rpi)

chriscamacho:
ended up using the 100 version shaders as off the top of my 
head can't think of any improvement I could use on the shaders for
version 120

This has been tested on Linux (Desktop) requires testing on a Rpi


raysan5:
@chriscamacho ok, it should work...

PR #1144, opened 2020-03-23
Android: Better track touch input returned from IsMouse*()

Doy-lee:
Switch to actually tracking touch input to use for "mouse" input rather
than the gestures system. The gesture system as an abstraction ontop of
raw touch input loses some information needed to map to "mouse"
input.

Before,
- IsMouseButtonReleased() triggers immediately after the initial touch
(because GESTURE_TAP activates immediately on touch) instead of waiting for the
touch to be released.
- IsMouseButtonUp() returns false, when it should just be the opposite
of IsMouseButtonDown().
- IsMouseButtonDown() returns true only after GESTURE_HOLD (which
activates after some period of time after GESTURE_TAP), when instead it
should just be true whenever there is touch input i.e. gesture !=
GESTURE_NONE or alternatively when any input is received on the screen.

After this PR, touches map closer to mouse input.
- IsMouseButtonReleased() triggers when touch is released (last frame
was touched, this frame not touched).
- IsMouseButtonUp() returns the opposite of IsMouseButtonDown()
- IsMouseButtonDown() is true when
(AMOTION_EVENT_ACTION_DOWN || AMOTION_EVENT_ACTION_MOVE) and false when
(AMOTION_EVENT_ACTION_UP)

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2648,7 +2648,7 @@ Vector2 GetTouchPosition(int index)
 {
     Vector2 position = { -1.0f, -1.0f };
 
-#if defined(PLATFORM_ANDROID) || defined(PLATFORM_WEB)
+#if defined(PLATFORM_ANDROID) || defined(PLATFORM_WEB) || defined(PLATFORM_RPI)
     if (index < MAX_TOUCH_POINTS) position = CORE.Input.Touch.position[index];
     else TRACELOG(LOG_WARNING, "Required touch point out of range (Max touch points: %i)", MAX_TOUCH_POINTS);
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Doy-lee wow! very nice improvement! I added a comment relative to web platform...

Doy-lee:
Oops yeah that's not meant to be there. I assume you mean the #define with `PLATFORM_RPI` instead of `PLATFORM_WEB`. I've updated.

raysan5:
@Doy-lee Thanks for the review! Also note that touch can be supported on RPI through an external touch screen device, some users have used it in the past...

raysan5:
@Doy-lee thank you very much for the improvements! Great work! :)

Doy-lee:
My pleasure

PR #1146, opened 2020-03-24
as per request spotlight example

chriscamacho:
NB GLSL v100 shader compiles and works on desktop but had NOT been tested elsewhere


raysan5:
@chriscamacho thanks! 😊

chriscamacho:
no es nada...

PR #1148, opened 2020-03-24
spotlight example, each spot has own radius, mouse countrol

chriscamacho:
NB glsl100 shader needs testing on "bare metal"

### DIFF
--- a/examples/shaders/resources/shaders/glsl100/spotlight.fs
+++ b/examples/shaders/resources/shaders/glsl100/spotlight.fs
@@ -2,50 +2,67 @@
 
 precision mediump float;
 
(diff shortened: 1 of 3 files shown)


raysan5:
@chriscamacho thanks for improvement! :D

chriscamacho:
Your suggestions not only made for a better example (as usual!)
but the task of finding the nearest spot to a point where all spots can be different sizes ended up being an interesting challenge! (see the shader)
oh btw just a reminder the GLSL v100 shader still needs testing!

raysan5:
@chriscamacho it can be tested on web platform, it uses GLSL 100, I'll try to test it today!

PR #1158, opened 2020-03-29
Fix "variable is reassigned before use" linter warning on raymath.h

ibilon:
Running cppcheck on my project which include raymath.h showed me these two warnings:
```
[raylib/src/raymath.h:557] -> [raylib/src/raymath.h:565]: (style) Variable 'result' is reassigned a value before the old one has been used.
[raylib/src/raymath.h:1326] -> [raylib/src/raymath.h:1328]: (style) Variable 'resAngle' is reassigned a value before the old one has been used.
```

This is mostly a style fix, but since the first value is never used might as well not assign it.


raysan5:
@ibilon thanks for the review.

PR #1162, opened 2020-03-30
[cppcheck] Fix minor warnings in models.c and raudio.c

danimartin82:
Errores fixes:

- models.c,2843 Either the condition 'fileData!=NULL' is redundant or there is possible null pointer dereference: fileData.

- raudio.c,805, 806, 807, 808, %d in format string (no. 2) requires 'int' but the argument type is 'unsigned int'.


raysan5:
@danimartin82 Nice! Thank yopu very much for the improvement! :)

PR #1170, opened 2020-04-02
Potential bug in GenImageFontAtlas

SasLuca:
## Problem Description
If you generate a font with all the default characters except for the whitespace this will cause a character to wrap and not be packed correctly. To test this I added a simple example that loads a font with all the default characters except the whitespace and renders the atlas on screen. This the result:

Notice the `u` character on the second to last row, right at the end some of its pixels wrap and are placed next to the `b` character.

## Potential fix
While examining the code I found this potential issue that might cause this:
```c
// Move atlas position X for next character drawing
offsetX += (chars[i].image.width + 2*padding);
```

I think the issue here is that we are advancing the `offsetX` by too much, I think it should be `offsetX += (chars[i].image.width + padding);`. 

I removed the instances of `2*padding` that I could find which seemed like bugs and this seems to have fixed the issue. Here is an image showing the result after the changes are applied:

## Testing
I also ran all the other text examples to make sure this didn't break anything and they all seem to work fine.

### DIFF
--- a/examples/text/text_font_custom_characters.c
+++ b/examples/text/text_font_custom_characters.c
@@ -1,0 +1,24 @@
+#include "raylib.h"
+
+int main(void)
+{
+    const int screenWidth = 800;
+    const int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [text] example - font custom characters");
+    SetTargetFPS(60);
+
+    int chars[126 - 33];
+    for (int i = 0; i < 126 - 33; i++) chars[i] = 33 + i;
+
+    Font font = LoadFontEx("resources/KAISG.ttf", 64, chars, 126 - 33);
+
+    while (!WindowShouldClose())
+    {
+        BeginDrawing();
+        ClearBackground(WHITE);
+
+        DrawTexture(font.texture, 0, 0, BLACK);
+        EndDrawing();
+    }
+}
--- a/src/text.c
+++ b/src/text.c
@@ -613,7 +613,7 @@ Image GenImageFontAtlas(const CharInfo *chars, Rectangle **charRecs, int charsCo
     // NOTE 2: SDF font characters already contain an internal padding,
     // so image size would result bigger than default font type
     float requiredArea = 0;
-    for (int i = 0; i < charsCount; i++) requiredArea += ((chars[i].image.width + 2*padding)*(chars[i].image.height + 2*padding));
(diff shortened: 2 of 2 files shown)


raysan5:
@SasLuca please, could you try with a couple of other ttf fonts?

SasLuca:
I tried with 2 other fonts: `AnonymousPro-Bold` and `pixantiqua`. Both were fine both using the current algorithm and my modified one.

My guess would be that with the current algorithm this issue may happen sometimes in edge cases and my hope would by that my fix solves those edge cases.

Even logically, it was pointed out to me a by @mugurelenache, that advancing `offsetX` by `2*padding` doesn't seem to make sense, and since the algorithm for deciding the size of the atlas works by guessing a good enough size if the offset is advancing too fast it is reasonable to assume there might be edge cases where this would cause and issue.

raysan5:
@SasLuca just thinking about this, padding is multiplied by two because it considers characters padding in all sides of the character.

It's possible to see this more clearly in my atlas generation tool:

raylib_atlas_text_padding

**KAISG.ttf, 256x256, font size: 32px, padding 4px**

As you can see the 4px padding is applied left-right-top-bottom of the character.

I think the issue is not related to that `2*padding` but the condition to jump to next line... as you said it's probably hitting some limit case so not working properly.

SasLuca:
Aha, i knew we talked about this once and I understood the reason for `2*padding`. I guess then maybe we should increase the guess size by a bit?

raysan5:
@SasLuca yes, I remember now! 

About increasing the guess size, that's actually tricky, I mean it's a guess... I already reviewed it at least two times... and there are always some limit cases (characters get cut somewhere).

In this specific case, problem is not the guessed size (you can see that we still have room in the last line for additional characters) the problem seems to be related in the `nextLine` condition because it fills one line but does not jump to the next.

SasLuca:
Based on my debugging the problem seems to actually be that on the line of that character there is not enough space left at the end for the letter `u`. What happens is that the target index `(offsetY + y)*atlas.width + (offsetX + x)` overflows into the next line of pixels. I tested this by taking the index and diving by the atlas width and found that even on `y = 0` after `x = 18` the line will move to the next one, from line `274` to `275`. Now the question is how does this happen?

SasLuca:
I found the problem!

This line should be changed:
```c
if (offsetX >= (atlas.width - chars[i].image.width - padding))
```
to this:
```c
if (offsetX >= (atlas.width - chars[i].image.width - 2*padding))
```

We needed to take into account the padding on both sides.
I will close this PR and do a new one with my example and this fix. Is that ok?

PR #1171, opened 2020-04-02
Fixed offset check in GenImageFontAtlas

SasLuca:
A small fix to a subtle bug in `GenImageFontAtlas` that we previously discussed in this thread: https://github.com/raysan5/raylib/pull/1170

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -640,7 +640,7 @@ Image GenImageFontAtlas(const CharInfo *chars, Rectangle **charRecs, int charsCo
             {
                 for (int x = 0; x < chars[i].image.width; x++)
                 {
-                    ((unsigned char *)atlas.data)[(offsetY + y)*atlas.width + (offsetX + x)] = ((unsigned char *)chars[i].image.data)[y*chars[i].image.width + x];
+                    ((unsigned char*)atlas.data)[(offsetY + y)*atlas.width + (offsetX + x)] = ((unsigned char*)chars[i].image.data)[y*chars[i].image.width + x];
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/text.c
```diff
@@ -640,7 +640,7 @@ Image GenImageFontAtlas(const CharInfo *chars, Rectangle **charRecs, int charsCo
             {
                 for (int x = 0; x < chars[i].image.width; x++)
                 {
-                    ((unsigned char *)atlas.data)[(offsetY + y)*atlas.width + (offsetX + x)] = ((unsigned char *)chars[i].image.data)[y*chars[i].image.width + x];
+                    ((unsigned char*)atlas.data)[(offsetY + y)*atlas.width + (offsetX + x)] = ((unsigned char*)chars[i].image.data)[y*chars[i].image.width + x];
```
raysan5, line 643: this change is not required

src/text.c
```diff
@@ -653,7 +653,7 @@ Image GenImageFontAtlas(const CharInfo *chars, Rectangle **charRecs, int charsCo
             // Move atlas position X for next character drawing
             offsetX += (chars[i].image.width + 2*padding);
 
-            if (offsetX >= (atlas.width - chars[i].image.width - padding))
+            if (offsetX >= (atlas.width - chars[i].image.width - 2 * padding))
```
raysan5, line 656: please, could you avoid spaces in multiplication `2*padding`? Just ot follow raylib notation rules...

raysan5:
@SasLuca thanks for the review and excuse me for being picky! :P

SasLuca:
Done :D
I realized I formatted those wrong after I already pushed the code, my bad :))

raysan5:
@SasLuca thanks for the fix! Those are the kind of details I really care about and that make raylib what it is! :D

PR #1172, opened 2020-04-04
TextReplace changes

BenStigsen:
- Added NULL return if the replacement string (by) is empty
- Reordered sanity checks since there's no need to initialize variables if the strings are invalid.


raysan5:
@BenStigsen thanks for the review! 😄

ibilon:
Replacing by an empty string could be useful though.

BenStigsen:
Oh wait that's true, not sure why I didn't think this PR through.
It should still check `!by`, but it should not be checking for `by[0] == '\0'`

raysan5:
@BenStigsen feel free to send another PR. it also seems there are some problems with TCC compiler with current implementation...

Issue #1175, opened 2020-04-04
[audio] Please inform users what types can be opened as Music


PapyChacal:
- [X] I tested it on latest raylib version from master branch
  Used both 3.0.0 and master.
 - [X] I checked there is no similar issue already reported
  I feel it could be related to #957 ?
 - [X] My code has no errors or misuse of raylib

### Issue description

  Okay, I was wondering why I can't open my WAV files as Music objects.
  After some searching, building this minimal example and everything, I checked for similar issues to be sure and found #48.
  It would be nice to inform the user of this specificity !
  Beyond that, your lib looks really nice, I'm using raudio only as I'm stuck with SDL2 for a student's project, but I'm looking forward to try the rest of it ! 

### Environment

  I'm using Ubuntu 18.04, GCC 7.5.0 targetting x86_64-linux-gnu
  I don't know what else is relevant to the issue ?

### Issue Screenshot
```
INFO: AUDIO: Device initialized successfully
INFO:     > Backend:      miniaudio / PulseAudio
INFO:     > Format:       32-bit IEEE Floating Point -> 16-bit Signed Integer
INFO:     > Channels:     2 -> 2
INFO:     > Sample rate:  44100 -> 44100
INFO:     > Periods size: 1320
INFO: AUDIO: Multichannel pool size: 16

Try to open as a Sound
INFO: WAVE: [music.wav] File loaded successfully (44100 Hz, 16 bit, Stereo)
INFO: WAVE: Unloaded wave data from RAM
INFO: WAVE: Unloaded sound data from RAM

Try to open as a Music
WARNING: FILEIO: [music.wav] Music file could not be opened
INFO: STREAM: Unloaded audio stream data from RAM

Try to open as a Sound
INFO: WAVE: [shoot.wav] File loaded successfully (44100 Hz, 16 bit, Stereo)
INFO: WAVE: Unloaded wave data from RAM
INFO: WAVE: Unloaded sound data from RAM

Try to open as a Music
WARNING: FILEIO: [shoot.wav] Music file could not be opened
INFO: STREAM: Unloaded audio stream data from RAM
INFO: AUDIO: Device closed successfully
```

### Code Example

```
#include "stdio.h"
#include "raudio.h"

int main()
{
    InitAudioDevice();
    printf("\nTry to open as a Sound\n");
    Sound sound = LoadSound("music.wav");
    UnloadSound(sound);

    printf("\nTry to open as a Music\n");
[...]

PR #1176, opened 2020-04-05
added log warning for unsupported file extension for music stream

louisgjohnson:
related to this issue [#1175](https://github.com/raysan5/raylib/issues/1175) I added a warning if the file extension isn't supported

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1139,8 +1139,8 @@ Music LoadMusicStream(const char *fileName)
         }
     }
 #endif
-    else TraceLog(LOG_WARNING, "[%s] Audio fileformat not supported, it can't be loaded", fileName);
-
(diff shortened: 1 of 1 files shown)


src/raudio.c
```diff
@@ -1139,6 +1139,7 @@ Music LoadMusicStream(const char *fileName)
         }
     }
 #endif
+    else TraceLog(LOG_WARNING, "[%s] Audio fileformat not supported, it can't be loaded", fileName);
```
raysan5, line 1142: it's ok but it should be:
```c
else TRACELOG(LOG_WARNING, "STREAM: [%s] Fileformat not supported", fileName);
```
to be consistent with similar traces.

louisgjohnson: Thanks, updated.

raysan5:
@louisgjohnson Thanks for the review! :)

PR #1177, opened 2020-04-05
added DrawTextPro function to be able to modify line spacing

louisgjohnson:


raysan5:
@louisgjohnson Thanks for the addition but I prefer not to add this function at this moment...

Actually, that's not the better way to draw multiline text, I prefer to let the user manage the multi-line on user code side, calling multiple DrawTextEx() for every line.

louisgjohnson:
@raysan5 okay no worries, I was just trying to work out how to adjust the line height with multi line text and saw that there was a todo to add this ability but having multiple drawText functions is also an easy way to do it

PR #1178, opened 2020-04-05
Allow for empty replacement string

BenStigsen:


BenStigsen:
Allows for empty replacement string, in case the user wants to replace something by nothing. Not sure why I removed it in the last PR, but thanks @ibilon for telling me. Fix for #1172

raysan5:
@BenStigsen thanks for the patch.

PR #1181, opened 2020-04-07
ignore symlink creation failure error

Shylie:


a3f:
Didn't know  about that one. Thanks!

PR #1190, opened 2020-04-10
Add filesystem "override" API.

TSnake41:
This is not a true "override", it fallbacks to default behavior if load fails using override.

This API is very useful for raylua "embedding" as it permit raylib to load file from payload (and avoid the need of external files).
It may be also useful for rres ?


raysan5:
@TSnake41 Thanks for improvement but not plans to expose the API through function pointers at this moment... I tried to avoid function pointers in raylib, also considering removing the tracelog callback... I need to think about it carefully...

PR #1193, opened 2020-04-11
Fix VSCode

stefanerwinmayer:
- Fixes the built executable not launching when debugging from VSCode because of misconfiguration
- Fixes the build error on macOS due to calling the wrong make version

### DIFF
--- a/projects/VSCode/.vscode/launch.json
+++ b/projects/VSCode/.vscode/launch.json
@@ -8,7 +8,7 @@
             "name": "Debug",
             "type": "cppdbg",
             "request": "launch",
-            "program": "${workspaceFolder}/main",
+            "program": "${workspaceFolder}/game",
             "args": [],
             "stopAtEntry": false,
             "cwd": "${workspaceFolder}",
--- a/projects/VSCode/Makefile
+++ b/projects/VSCode/Makefile
@@ -180,6 +180,9 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     ifeq ($(PLATFORM_OS),LINUX)
         MAKE = make
     endif
+    ifeq ($(PLATFORM_OS),OSX)
+        MAKE = make
+    endif
 endif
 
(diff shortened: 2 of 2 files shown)


projects/VSCode/.vscode/launch.json
```diff
@@ -8,7 +8,7 @@
             "name": "Debug",
             "type": "cppdbg",
             "request": "launch",
-            "program": "${workspaceFolder}/main",
+            "program": "${workspaceFolder}/game",
```
raysan5, line 11: this change shouldn't be here, I corrected it some commits ago...

stefanerwinmayer: Why is that though? When i debug it produces an executable named 'game'. Why do you think it should be 'main'? If you look at Windows it also says `"${workspaceFolder}/game.exe"`

raysan5: @stefanerwinmayer On Windows it generates `main.exe`, checking `tasks.json`, it's configured this way on `build release windows`. 

I think that's the right way because compiled file name can change.

raysan5: Ok, checking a bit further it seems `main.c` is hardcoded in `Makefile`...

stefanerwinmayer: Ok I can revert this one. Do you want the other change or should we close this and I apply the make fix change to any makefile that needs it and open a new PR?

raysan5: Just revert this change and I'll merge the other one. Yes, it should be applied to all Makefiles, it can be done in another PR.

raysan5:
@stefanerwinmayer I'm merging and further editing on my side (some new changes required).

stefanerwinmayer:
> @stefanerwinmayer I'm merging and further editing on my side (some new changes required).

Thanks. Sorry for the late response.

PR #1197, opened 2020-04-12
Make raymath more consistent

G3bE:
To quote my self from the #raylib channel on the raylib discord:
Alright, current raymath isn't consistent with functions for adding, subtracting, multiplying and dividing of  vectors. We have:
```
- Vector2Add(Vector2 v1, Vector2 v2), but not Vector2Add(Vector2 v, float add) (same for Vector3 respectively)
- Vector2Subtract(Vector2 v1, Vector2 v2), but not Vector2Subtract(Vector2 v, float sub)(same for Vector3 respectively)
- Vector2MultiplyV(Vector2 v1, Vector2 v2), but not Vector2Multiply(Vector2 v, float mul)
- Vector3Multiply(Vector3 v1, Vector3 v2) (see missing V here), but not Vector3Multiply(Vector3 v, float mul)
- Vector2DivideV(Vector2 v1, Vector2 v2) **AND** Vector2Divide(Vector2 v, float div) (same for Vector3 respectively)
```
Changes in this PR address these issues. Here is a brief overview:
### Rename
| Before                       | After |
|--------------------|------|
| Vector2Add             | Vector2AddV
| Vector2Subtract     | Vector2SubtractV
| Vector3Add             | Vector3AddV
| Vector3Subtract     | Vector3SubtractV
| Vector3Multiply      | Vector3MultiplyV
| QuaternionMultiply | QuaternionMultiplyQ
### Add functions
- `Vector2Add(Vector2 v, float add)`
- `Vector2Subtract(Vector2 v, float sub)`
- `Vector2Multiply(Vector2 v, float mul)`
- `Vector3Add(Vector3 v, float add)`
- `Vector3Subtract(Vector3 v, float sub)`
- `Vector3Multiply(Vector3 v, float mul)`
- `QuaternionAddQ(Quaternion q1, Quaternion q2)`
- `QuaternionAdd(Quaternion q, float add)`
- `QuaternionSubtractQ(Quaternion q1, Quaternion q2)`
- `QuaternionSubtract(Quaternion q, float sub)`
- `QuaternionMultiply(Quaternion q, float sub)`
- `QuaternionDivideQ(Quaternion q1, Quaternion q2)`
- `QuaternionDivide(Quaternion q, float sub)`

Especially for these last Quaternion function I'd like review on necessity and implementation. Thanks in advance

### DIFF
--- a/examples/core/core_2d_camera_platformer.c
+++ b/examples/core/core_2d_camera_platformer.c
@@ -226,13 +226,13 @@ void UpdateCameraCenterSmoothFollow(Camera2D *camera, Player *player, EnvItem *e
     static float fractionSpeed = 0.8f;
     
     camera->offset = (Vector2){ width/2, height/2 };
-    Vector2 diff = Vector2Subtract(player->position, camera->target);
+    Vector2 diff = Vector2SubtractV(player->position, camera->target);
     float length = Vector2Length(diff);
     
     if (length > minEffectLength)
     {
         float speed = fmaxf(fractionSpeed*length, minSpeed);
-        camera->target = Vector2Add(camera->target, Vector2Scale(diff, speed*delta/length));
+        camera->target = Vector2AddV(camera->target, Vector2Scale(diff, speed*delta/length));
     }
 }
 
--- a/examples/core/core_2d_camera_platformer.c
+++ b/examples/core/core_2d_camera_platformer.c
@@ -226,13 +226,13 @@ void UpdateCameraCenterSmoothFollow(Camera2D *camera, Player *player, EnvItem *e
     static float fractionSpeed = 0.8f;
     
     camera->offset = (Vector2){ width/2, height/2 };
-    Vector2 diff = Vector2SubtractV(player->position, camera->target);
+    Vector2 diff = Vector2Subtract(player->position, camera->target);
     float length = Vector2Length(diff);
     
     if (length > minEffectLength)
     {
         float speed = fmaxf(fractionSpeed*length, minSpeed);
-        camera->target = Vector2AddV(camera->target, Vector2Scale(diff, speed*delta/length));
+        camera->target = Vector2Add(camera->target, Vector2Scale(diff, speed*delta/length));
     }
 }
 
--- a/examples/shaders/shaders_spotlight.c
+++ b/examples/shaders/shaders_spotlight.c
@@ -245,12 +245,12 @@ void ResetStar(Star *s)
         
     } while (!(fabs(s->vel.x) + fabs(s->vel.y) > 1));
     
-    s->pos = Vector2Add(s->pos, Vector2MultiplyV(s->vel, (Vector2){ 8, 8 }));
+    s->pos = Vector2AddV(s->pos, Vector2MultiplyV(s->vel, (Vector2){ 8, 8 }));
 }
 
 void UpdateStar(Star *s)
 {
-    s->pos = Vector2Add(s->pos, s->vel);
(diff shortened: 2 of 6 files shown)


G3bE:
DO NOT MERGE. Changes in physics examples needed. Those were not tested since they are not included in the make build.

G3bE:
As far as I'm concerned this PR is good to go :)

raysan5:
@G3bE This is a big breaking change, I need to think carefully about it... Also note that we have `Vector2Scale()` and `Vector3Scale()` that with this new implementation should be removed...

Undoubtely this is a naming consistency change I like but I still have some concerns about if new names would be intuitive enough for new users...

G3bE:
You are totally right about the scale functions. Another possible way of naming is to add a suffix for functions that use values instead of vectors. That way the only function that is getting renamed is `Vector2MultiplyV` (-> `Vector2Multiply`) and the rest is just added functions.

raysan5:
@G3bE I checked several other libraries for reference: XNA, libGDX, netDXF, OpenTK, Unity... All those libraries are object-oriented and allow operators overloading and polymorphysm but from a user point of view it seems a function like `Vector3.Add` usually refers to adding two vectors.

Also checked other references for a pure C implementation of math functionality: [Quake src](https://github.com/id-Software/Quake-2/blob/372afde46e7defc9dd2d719a1732b8ace1fa096e/game/q_shared.h), [single-file-libs](https://github.com/nothings/single_file_libs#vectors), in those cases you can see all kind of function naming and lot of macro usage but most of them agree that `Vector*Add` equivalents refer to pure Vector addition, most even ignore the `float` to Vector operation. [HandmadeMath](https://github.com/HandmadeMath/Handmade-Math/blob/master/HandmadeMath.h) uses a `MultiplyVec2f()` aproximation, adding that `f` at the end to differenciate with vector version.

So, my conclusion is:

 1. Keep `Vector*Add()`, `Vector*Subtract()`, `Vector*Multiply()` and `Vector*Divide()` as the vector form.
 2. Use `Vector*Scale()` to _scale_ vector size by a float (equivalent to multiply/divide).
 3. Add `Vector*AddValue()` to refer to float value addition to vector.

I think this is the most comprehensive solution for most users and also keeps the library quite consistent in naming conventions.

G3bE:
I recon the same-ish thing for quaternions? As in QuaternionAdd for pure quaternion addition and QuaternionAddValue for adding a float value, or should I remove them completely?

raysan5:
@G3bE Yes, same way for quaternions

G3bE:
`Vector3DivideValue` is required by `Vector2Normalize` and it might be worth keeping to help show what you really mean with certain code e.g. you don't want to multiply by 0.125 but split in 8th.

raysan5:
@G3bE what is the difference between multiply by 0.125 and split by 8, actually for me seems the same to scale a vector to 1/2 or 1/8 than divide it by 2 or by 8. 

Reducing it to one function also reduces the number of functions in the API. I prefer to go for that solution.

G3bE:
Alright, your call :)

raysan5:
@G3bE Nice! Thank you very much for this big update of raymath! :D

jamesl-github:
> Also checked other references for a pure C implementation of math functionality: Quake src, single-file-libs. 

I suppose you looked at linmath.h 
its used in most of GLFW's examples 
https://github.com/glfw/glfw/blob/master/deps/linmath.h 
its pure C and also uses at lot of macros

raysan5:
@jamesl-github yes, it also uses a similar convention (despite the macros system).

PR #1202, opened 2020-04-14
Set the correct version of make on macos

stefanerwinmayer:
Currently mingw32-make is set but it should be regular make


raysan5:
@stefanerwinmayer thanks! 👍😄

PR #1205, opened 2020-04-17
Added method to load TTF font from memory

scottmudge:
I had a few fonts I wanted to embed into my executable, but didn't want to cache to disk before loading them.

Added a method to load TTF fonts directly from memory, thought some might find it useful.

Would also be nice to do so with images, but in the meanwhile I'm just using stb_image as a bridge to decode the raw pixel data.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1263,7 +1263,7 @@ void BeginDrawing(void)
 }
 
 // End canvas drawing and swap buffers (double buffering)
-void EndDrawing(void)
+void EndDrawingEx(void(*fn)())
 {
 #if defined(PLATFORM_RPI) && defined(SUPPORT_MOUSE_CURSOR_RPI)
     // On RPI native mode we have no system mouse cursor, so,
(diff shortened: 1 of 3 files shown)


raysan5:
Hi @scottmudge! Thank you very much for the improvement! I'm also working on memory buffers data loading but still deciding the best approach to it, not merging for the moment but I keep it open.

RobLoach:
Wanted to chip in here and say that I really appreciate the loading from memory functions. While I understand it can pose some conflict with the rest of raylib's design (ease of use, concern about memory loss, etc), it does give raylib a lot more legs on how it can be used.

[rres](https://github.com/raysan5/rres) has some of this functionality, but I would really love for more of the `Load*FromMemory()` functions to be brought into raylib core. Thanks!

raysan5:
No plan for those additions at the moment. Internal buffers drawing can be forced with `rlglDraw()`, loading files from memory is still under consideration...

PR #1210, opened 2020-04-21
Update easings.h

flashback-fx:
Make easings.h use powf and sqrtf instead of pow and sqrt.

Hello,

Almost one year ago I made a pull request that made the easings.h file use the tgmath.h header and specified all values as floats (by appending the character f to the numbers). This made the functions in easings.h faster since it eliminated type conversions.

I found out that later the patch was revised and you reverted tgmath.h to math.h and changed the sin and cos functions to their float equivalents (sinf and cosf), but didn't change sqrt and pow to their float equivalents (sqrtf and powf). Was there a reason for omitting these two or was it a mistake?

I've run the benchmarks I prepared last time against the current easings.h and one that uses sqrtf and powf and the performance indeed increases. The affected functions are the ones in the EaseCirc, EaseExpo and EaseElastic and the results range from around 1.2 to 5.5 speed increase depending on the function and test in my machine (I ran six tests, three short ones and three long ones, with no optimization, O1 and O2 respectively). You can see the test code (and run it) as well as my results here https://github.com/flashback-fx/easings_tools (the benchmark is a bit messy to set up, if someone wants to run it but does not know how please ask).

Finally, I made the easings_testbed but I found out I am not credited in examples/README.md (I am in the actual file tho)
Could I please get credit?

Thanks again to ray and all contributors for this amazing library.

### DIFF
--- a/src/easings.h
+++ b/src/easings.h
@@ -90,7 +90,7 @@
     #define EASEDEF extern
 #endif
 
-#include <math.h>       // Required for: sinf(), cosf(), sqrt(), pow()
+#include <math.h>       // Required for: sinf(), cosf(), sqrtf(), powf()
 
 #ifndef PI
     #define PI 3.14159265358979323846f //Required as PI is not always defined in math.h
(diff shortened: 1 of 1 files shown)


raysan5:
@flashback-fx Thanks for the improvement! Sorry, my bad! I updated the README! :)

PR #1211, opened 2020-04-22
Mempool and other Allocators Optimization

assyrianic:
Optimizations of allocators.
Renamed 'Stack' to 'Arena'.
Replaced certain define constants with an anonymous enum.
Refactored MemPool to no longer require active or deferred defragging.


raysan5:
@assyrianic thanks for the update! Actually, there is a pending raylib example showcasing the use of the library!

PR #1212, opened 2020-04-22
[android] Extend validity of the certificate

danimartin82:
For uploading a apk to Google Play a minimun of 25 years (9125 days) is needed for the validity of the key.

More info:
https://developer.android.com/studio/publish/app-signing

Pending to improve the makefiles: the actual makefiles for android in raylib generate a new key in a new keystore each time it is compiled. This is not valid for Google Play, they request to use always the same keysotre.


raysan5:
Thanks @danimartin82! Actually, the key for GooglePlay needs to be configured with the credentials of the user. The generated key is just for testing.

PR #1222, opened 2020-04-26
Added checks to fix compat with switch homebrew

NimbusFox:
Make file will need some changes
```
if PLATFORM=Switch(May need to change to Switch_HB)
    Check if ENV:DEVKITPRO exists
    // This is due to switch not liking miniaudio
    Disable raudio.c

    // Switch homebrew compiling does not like dynamic linking
    compile as static

    // The switch uses most of the code from the desktop platform core
    // And has precompiler flags setup to block the code that breaks
    // on switch
    define both PLATFORM_SWITCH and PLATFORM_DESKTOP

    // Make sure -lGL is not being assigned anywhere as devkitpro does
    // not supply a direct OpenGL library
    Use external GLFW3 from devkitpro
    Add linking flags:
     -lnx -lm -lpthread -lEGL -lglfw3 -ldrm_nouveau -lglapi
     -lGLESv1_CM -lGLESv2 -lnxd -ldeko3dd -ldeko3d
endif
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -45,6 +45,10 @@ endif()
 
 add_definitions("-DRAYLIB_CMAKE=1")
 
+if(${PLATFORM} MATCHES "Switch")
+  set(USE_AUDIO FALSE)
+endif()
+
 if(USE_AUDIO)
   MESSAGE(STATUS "Audio Backend: miniaudio")
   set(sources ${raylib_sources})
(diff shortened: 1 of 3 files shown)


raysan5:
Hi @NimbusFox! Thanks for the improvement but I can not officially support Switch as a platform. It requires a developer license and related code should be closed-source.

It's great to see that raylib could works in Switch with minimal changes but I'm not adding them to the repo.

PR #1227, opened 2020-04-27
Fix materials loading for IQM models

sikor666:
I have problem when loading IQM models with more than one mesh because I can't assign different textures for other meshes, for example I want do like code bellow and without materials is very uncomfortable.

```
// Load mrfixit model
Model mrfixitModel = LoadModel("../resources/mrfixit.iqm");               // Load the animated model mesh and basic data
Texture2D mrfixitBody = LoadTexture("../resources/mrfixit_body.png");    // Load model texture and set material
Texture2D mrfixitHead = LoadTexture("../resources/mrfixit_head.png");    // Load model texture and set material
mrfixitModel.materials[0].maps[MAP_DIFFUSE].texture = mrfixitBody;             // Set map diffuse texture
mrfixitModel.materials[1].maps[MAP_DIFFUSE].texture = mrfixitHead;             // Set map diffuse texture
SetModelMeshMaterial(&mrfixitModel, 0, 0); // body
SetModelMeshMaterial(&mrfixitModel, 1, 1); // head
```

[mrfixit.zip](https://github.com/raysan5/raylib/files/4541110/mrfixit.zip)

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -3006,6 +3006,7 @@ static Model LoadIQM(const char *fileName)
 
     #define BONE_NAME_LENGTH    32          // BoneInfo name string length
     #define MESH_NAME_LENGTH    32          // Mesh name string length
+    #define MATERIAL_NAME_LENGTH 32         // Material name string length
 
     // IQM file structs
     //-----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


raysan5:
@sikor666 thanks for the improvement.

PR #1228, opened 2020-04-27
Gamepad axis bug fixes and improvement

mmalecot:
**Fixes GetGamepadAxisCount:**

**GLFW** defines the number of axis from 0 to 5 which means that **GLFW_GAMEPAD_AXIS_LAST** is equal to **count - 1** and not **count**.

**Fixes core_input_gamepad example:**

For an Xbox gamepad, the axes are not correctly drawn (inverted axes).

**Fixes inaccurate axis values:**

Joysticks axis are really inaccurate between -0.2 and 0.2.
I suggest to return 0.0 for this slice.

### DIFF
--- a/examples/core/core_input_gamepad.c
+++ b/examples/core/core_input_gamepad.c
@@ -92,13 +92,13 @@ int main(void)
                     DrawCircle(259, 152, 39, BLACK);
                     DrawCircle(259, 152, 34, LIGHTGRAY);
                     DrawCircle(259 + (GetGamepadAxisMovement(GAMEPAD_PLAYER1, GAMEPAD_AXIS_LEFT_X)*20),
-                               152 - (GetGamepadAxisMovement(GAMEPAD_PLAYER1, GAMEPAD_AXIS_LEFT_Y)*20), 25, BLACK);
+                               152 + (GetGamepadAxisMovement(GAMEPAD_PLAYER1, GAMEPAD_AXIS_LEFT_Y)*20), 25, BLACK);
 
                     // Draw axis: right joystick
                     DrawCircle(461, 237, 38, BLACK);
                     DrawCircle(461, 237, 33, LIGHTGRAY);
                     DrawCircle(461 + (GetGamepadAxisMovement(GAMEPAD_PLAYER1, GAMEPAD_AXIS_RIGHT_X)*20),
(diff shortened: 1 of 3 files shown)


examples/core/core_input_gamepad.c
```diff
@@ -163,7 +163,7 @@ int main(void)
 
                 DrawText(FormatText("DETECTED AXIS [%i]:", GetGamepadAxisCount(GAMEPAD_PLAYER1)), 10, 50, 10, MAROON);
 
-                for (int i = 0; i < GetGamepadAxisCount(GAMEPAD_PLAYER1); i++)
+                for (int i = 1; i <= GetGamepadAxisCount(GAMEPAD_PLAYER1); i++)
```
raysan5, line 166: This looks really counter-intuitive... it should return the number of axis for indexing starting at 0...

mmalecot: 0 corresponds to GAMEPAD_AXIS_UNKNOWN.
I can leave it like this and do i + 1 inside of the loop. But for me it's not intuitive too...

raysan5: @mmalecot Checking [the documentation](https://www.glfw.org/docs/3.3/group__gamepad__axes.html), gamepad axis indices start at 0 and GLFW_GAMEPAD_AXIS_LAST it's the count.

mmalecot: @raysan5 Yes that's right on the GLFW side. But on [raylib side](https://github.com/raysan5/raylib/blob/master/src/raylib.h#L667), 0 is used for error checking and axes begin at 1.

If you try the core_input_gamepad example without this PR, you will see that Axis 0 has no reality.

raysan5: @mmalecot I see... then issue is probably there, not in the axis count... counting one extra unexistent axis to solve an indexing shift shouldn't be the solution...

mmalecot: @raysan5 OK, see below for details of the last modification.

mmalecot:
@raysan5 

1. I removed **GAMEPAD_AXIS_UNKNOWN** and set all axes variants with the same values as **GLFW**.
2. I removed **GetGamepadAxis** function that became useless as all platforms provides the correct axis (note that GAMEPAD_AXIS_RIGHT_Y was missing from PLATFORM_WEB mapping).

raysan5:
@mmalecot nice! this is a great solution! thank you very much! :)

PR #1229, opened 2020-04-28
Update CMakeLists.txt to fix CMake build for Raspberry Pi

armageddon421:
Tested on a Raspberry Pi 3B+, sucessfully used together with node-raylib

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,6 +117,16 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
   set(PLATFORM_CPP "PLATFORM_RPI")
   set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
 
+  add_definitions(-D_GNU_SOURCE=1)
+  
+  find_library(OPENGL_LIBRARY OpenGL)
+  find_library(GLESV2 brcmGLESv2 HINTS /opt/vc/lib)
+  find_library(EGL brcmEGL HINTS /opt/vc/lib)
+  find_library(BCMHOST bcm_host HINTS /opt/vc/lib)
+  include_directories(/opt/vc/include /opt/vc/include/interface/vmcs_host/linux /opt/vc/include/interface/vcos/pthreads)
+  link_directories(/opt/vc/lib)
+  set(LIBS_PRIVATE ${GLESV2} ${EGL} ${BCMHOST} pthread rt m dl)
+
 endif()
 
 if (${OPENGL_VERSION})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -119,7 +119,6 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
 
   add_definitions(-D_GNU_SOURCE=1)
   
-  find_library(OPENGL_LIBRARY OpenGL)
   find_library(GLESV2 brcmGLESv2 HINTS /opt/vc/lib)
   find_library(EGL brcmEGL HINTS /opt/vc/lib)
   find_library(BCMHOST bcm_host HINTS /opt/vc/lib)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,7 +117,7 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
   set(PLATFORM_CPP "PLATFORM_RPI")
   set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
 
-  add_definitions(-D_GNU_SOURCE=1)
+  add_definitions(-D_DEFAULT_SOURCE)
   
   find_library(GLESV2 brcmGLESv2 HINTS /opt/vc/lib)
   find_library(EGL brcmEGL HINTS /opt/vc/lib)


src/CMakeLists.txt
```diff
@@ -117,6 +117,16 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
   set(PLATFORM_CPP "PLATFORM_RPI")
   set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
 
+  add_definitions(-D_GNU_SOURCE=1)
```
a3f, line 120: What declarations are missing otherwise?

src/CMakeLists.txt
```diff
@@ -117,6 +117,16 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
...
+  
+  find_library(OPENGL_LIBRARY OpenGL)
+  find_library(GLESV2 brcmGLESv2 HINTS /opt/vc/lib)
```
a3f, line 123: Is `/opt/vc` some standard path for the VideoCore? Who installs this?

src/CMakeLists.txt
```diff
@@ -117,6 +117,16 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
...
+  find_library(EGL brcmEGL HINTS /opt/vc/lib)
+  find_library(BCMHOST bcm_host HINTS /opt/vc/lib)
+  include_directories(/opt/vc/include /opt/vc/include/interface/vmcs_host/linux /opt/vc/include/interface/vcos/pthreads)
+  link_directories(/opt/vc/lib)
+  set(LIBS_PRIVATE ${GLESV2} ${EGL} ${BCMHOST} pthread rt m dl)
```
a3f, line 128: Don't you want `${OPENGL_LIBRARIES}` here as well?

src/CMakeLists.txt
armageddon421: I think it comes with every installation of raspbian. I don't know about other distibutions.
I got the paths from the Makefile https://github.com/raysan5/raylib/blob/e65d15f6ad6f9ade05999b66fdd91d3f6eb2a2d0/src/Makefile#L388

src/CMakeLists.txt
armageddon421: I think we could omit OPENGL_LIBRARIES entirely, as it should be replaced by the Raspberry Pi's VideoCode specific libraries.

src/CMakeLists.txt
armageddon421: I'm not sure. It had problems finding some standard headers without this line, if I remember correctly.

raysan5:
@armageddon421 @a3f is this PR ready to merge?

src/CMakeLists.txt
a3f: Please drop the `find_library(OPENGL_LIBRARY OpenGL)` dependency then.

src/CMakeLists.txt
a3f: Would be good to know what exactly failed. Might be just a header we are missing.

src/CMakeLists.txt
a3f: ok

src/CMakeLists.txt
armageddon421: ´´´
[  1%] Building C object src/CMakeFiles/raylib.dir/core.c.o
In file included from /opt/vc/include/interface/vcos/vcos.h:118,
                 from /opt/vc/include/interface/vmcs_host/vc_dispmanx.h:33,
                 from /opt/vc/include/EGL/eglplatform.h:110,
                 from /opt/vc/include/EGL/egl.h:36,
                 from /home/pi/node-raylib/node-raylib/vendor/raylib/src/rlgl.h:666,
                 from /home/pi/node-raylib/node-raylib/vendor/raylib/src/core.c:136:
/opt/vc/include/interface/vcos/pthreads/vcos_platform.h: In function ‘vcos_semaphore_wait_timeout’:
/opt/vc/include/interface/vcos/pthreads/vcos_platform.h:307:13: error: implicit declaration of function ‘sem_timedwait’; did you mean ‘sigtimedwait’? [-Werror=implicit-function-declaration]
       ret = sem_timedwait( sem, &ts );
             ^~~~~~~~~~~~~
             sigtimedwait
/home/pi/node-raylib/node-raylib/vendor/raylib/src/core.c: In function ‘EventThread’:
/home/pi/node-raylib/node-raylib/vendor/raylib/src/core.c:5162:13: error: implicit declaration of function ‘usleep’; did you mean ‘sleep’? [-Werror=implicit-function-declaration]
             usleep(5000); // Sleep for 5ms to avoid hogging CPU time
             ^~~~~~
             sleep
```
must have something to do with the VideoCore library. I'll try to research some more.

Comment (armageddon421) [reply src/CMakeLists.txt:120]: usleep is actually deprecated, I created a separate issue for this.
I have no idea yet what's up with `sem_timedwait`.
Every example - even in the Raspberry Pi userland repo itself - seems really simple. No flags or anything needed, just 2-3 libraries, which I added here.
The Makefile build also runs fine. What's the difference here?

Comment (armageddon421) [reply src/CMakeLists.txt]: I changed it to define `_DEFAULT_SOURCE` instead. That also works, and I got it from the Makefile: https://github.com/raysan5/raylib/blob/72443f8f25b1f8ded3b05da17857b7eeb1556b35/src/Makefile#L275

I hope that settles it for now. Maybe we can improve this situation in the future.

File: src/CMakeLists.txt [hunk 4]
```diff
@@ -117,6 +117,15 @@ elseif(${PLATFORM} MATCHES "Raspberry Pi")
   set(PLATFORM_CPP "PLATFORM_RPI")
   set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
 
+  add_definitions(-D_DEFAULT_SOURCE)
```
a3f, line 120: Why? What's missing if you remove it?

armageddon421: I think I explained my findings in this conversation.
https://github.com/raysan5/raylib/pull/1229#discussion_r420404363

The sleep issue should be resolved by now (see referenced issue) https://github.com/raysan5/raylib/issues/1235

However, I could not find out what's wrong with sem_timedwait. It may be accidentally overridden somewhere.
I could not find any example makefiles for the raspberry pi requiring anything but specifying those libraries and headers to work, so I cant't explain what's wrong.
The solution works, however. That's why I proposed to stick with it for now and try to resolve this later.

a3f: It's not as clean as I'd have liked, because there should be no reason to specify it when we are using a POSIX function like `sem_timedwait`, but ye, works is better than doesn't work.

a3f:
Thanks for your contribution! If you feel your fingers tickling, consider adding it to CI. :-)

Issue #1182, opened 2020-04-07
[build] VS2017.UWP project not working


raysan5:
UWP raylib project is not working (`PLATFORM_UWP`), it should be reviewed.

This project was interesting to develop with raylib for multiple Windows platforms, including `Xbox One`.

raysan5:
I'm investigating this issue. Programs crashes on first `glGetString()` call on `rlglInit()`, it means OpenGL context is not properly created or bound to graphic device. Related function: `InitGraphicsDevice()`

raysan5:
@Rover656 please, could you help with this issue? I tried debugging but it crashes on initialization, it seems issue is on `Windows::UI::Core::CoreWindow^ window` but I can't find it... Thanks! :)

Rover656:
I shall investigate this issue and will get back to you! There's a couple other UWP things I'll also look into to try and polish it up a bit!

raysan5:
@Rover656 Nice! Thank you very much! :D

PR #1231, opened 2020-04-29
UWP rework

Rover656:
Here is the ~~first draft of the~~ UWP rework discussed in discord and relating to issue #1182 

What I have done:
- Replaced the UWP messaging system that I added last time with a function callback system.
- The developer is made to implement all UWP support now (for window and application lifecycle only).
- The new system supports more of raylib's core.

What I have tested:
- Keyboard input
- Mouse input
- Window lifecycle

What I am still to test:
- [x] Gamepad input
- [x] Keyboard character recording
- [x] Whether or not UWP will support GIF recording.

What I am still to add/do:
- [x] Move UWP specifics to a separate header.
- [x] Add keyboard character recording.
- [x] I still need to work out if locking the mouse cursor position is truly possible or if the developer will have to do it themselves. This would be easy as they can set their own functions for this anyway.
- [x] Move BaseApp into the main App. This is now less of a boilerplate and more of an example
- [x] Add full set of keyboard keys to example
- [x] Try to fix the gamepad hack
- [x] Implement UWP lifecycle (Resume and Suspend) properly.
- [x] Ensure code matches raylib standards.


raysan5:
@Rover656 this is a big batch of changes... looking at the code it probably makes sense to move all UWP specific functionality to a self-contained `uwp_events.h` external library, used only on `core.c`.

Also, where is done the events polling? is it done by the App?

Rover656:
Hi ray, I agree with the moving of the UWP functionality to another header. Also, the event polling is only done by the App now, as most if not all of the functionality requires c++ access.

Rover656:
I believe that this is now ready for reviewing.

Rover656:
I'm going to look into UWP touch support, however I'm unsure how compatible it will be with raylib's setup as I've never used the touch APIs.

src/core.c
```diff
@@ -2673,19 +2687,19 @@ int GetMouseWheelMove(void)
 // Returns touch position X for touch point 0 (relative to screen size)
 int GetTouchX(void)
 {
-#if defined(PLATFORM_ANDROID) || defined(PLATFORM_WEB)
+#if defined(PLATFORM_ANDROID) || defined(PLATFORM_WEB) || defined(PLATFORM_UWP)
     return (int)CORE.Input.Touch.position[0].x;
-#else   // PLATFORM_DESKTOP, PLATFORM_RPI
+#else   // PLATFORM_DESKTOP, PLATFORM_RPI, PLATFORM_UWP
```
raysan5, line 2692: Just a typo I think, if PLATFORM_UWP is considered on previous #if, it shouldn't be here in the comment

raysan5:
@Rover656 I looks very nice! Congratulations! I like it! 😄

src/uwp_events.h
```diff
@@ -0,0 +1,119 @@
+/**********************************************************************************************
+*
+*   raylib.uwp_events - Functions for bootstrapping UWP functionality within raylib's core.
+*
+*
+*   LICENSE: zlib/libpng
+*
+*   Copyright (c) 2014-2020 Ramon Santamaria (@raysan5)
```
raysan5, line 8: You must be the one in this copyright notice, you created it!

src/core.c
Rover656: Oh, sorry. I thought that was talking about what the if contained. I shall remove these

src/uwp_events.h
Rover656: Alrighty, I shall add that too

raysan5:
@Rover656 Great improvement! Thank you very much!

PR #1236, opened 2020-05-05
No need for "LOGO_RECS_SIDE" maybe an artifact?

rfaile313:
"Advance Game" will not compile with those four lines of code.


raysan5:
@rfaile313 yeah, probably old code... thanks for the review!

PR #1246, opened 2020-05-12
fix bug in GetPrevDirectoryPath on Unix-like systems

ivan-cx:
How to reproduce a bug:
 1. on Unix-like machine call GetPrevDirectoryPath("/home")
 2. result is empty string 
 3. expected result "/"

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2087,11 +2087,12 @@ const char *GetPrevDirectoryPath(const char *dirPath)
 
     if (pathLen <= 3) strcpy(prevDirPath, dirPath);
 
-    for (int i = (pathLen - 1); (i > 0) && (pathLen > 3); i--)
+    for (int i = (pathLen - 1); (i >= 0) && (pathLen > 3); i--)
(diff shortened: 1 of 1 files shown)


ivan-cx:
BTW, there's probably another bug for GetPrevDirectoryPath("/с/") because of this code:
```
if (pathLen <= 3) strcpy(prevDirPath, dirPath);
```

src/core.c
```diff
@@ -2087,11 +2087,12 @@ const char *GetPrevDirectoryPath(const char *dirPath)
 
     if (pathLen <= 3) strcpy(prevDirPath, dirPath);
```
ivan-cx, line 2088: BTW, there's probably another bug for GetPrevDirectoryPath("/с/") because of this code

raysan5:
@ivan-cx thanks for the review! Actually I was tweaking that code for some time... there is always some case not properly considered! Feel free to send another PR if you feel it can be further improved.

PR #1247, opened 2020-05-13
Adding Normalize and Remap functions

NoorWachid:
// Normalize input value within input range
// Remap input value within input range to output range

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -154,6 +154,17 @@ RMDEF float Lerp(float start, float end, float amount)
     return start + amount*(end - start);
 }
 
+// Normalize input value within input range
+RMDEF float Norm(float value, float start, float end)
+{
+    return (value - start) / (end - start);
+}
+
+// Remap input value within input range to output range
+RMDEF float Remap(float value, float inputStart, float inputEnd, float outputStart, float outputEnd) {
+    return Lerp(outputStart, outputEnd, Norm(value, inputStart, inputEnd));
+}
+
 //----------------------------------------------------------------------------------
 // Module Functions Definition - Vector2 math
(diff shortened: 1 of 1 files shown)


src/raymath.h
```diff
@@ -154,6 +154,17 @@ RMDEF float Lerp(float start, float end, float amount)
     return start + amount*(end - start);
 }
 
+// Normalize input value within input range
+RMDEF float Norm(float value, float start, float end)
```
raysan5, line 158: Is `Norm()` function name standard? I mean, do other math libraries use that naming convention? Why not `Normalize()` for example? I try to name functions with the best recognizable naming...

src/raymath.h
```diff
@@ -154,6 +154,17 @@ RMDEF float Lerp(float start, float end, float amount)
...
+{
+    return (value - start) / (end - start);
+}
+
+// Remap input value within input range to output range
+RMDEF float Remap(float value, float inputStart, float inputEnd, float outputStart, float outputEnd) {
```
raysan5, line 164: Also, is `Remap()` function name standard? Do other math libraries use this naming convention?

NoorWachid: I'm pretty sure they'll never included this function because if they have already norm and lerp. But this name is the most used I believe. The second common name is just "map".

src/raymath.h
NoorWachid: It's like "lerp" they tend to use different name like "interpolate", "lin_inter" etc. So I'm not really sure about it. I'll change it to "Normalize"

raysan5:
@NoorWachid Merged! Thanks for the additions! :)

PR #1248, opened 2020-05-13
Introduced Vector2 and Vector3 Square Lenght.

ThePituLegend:
I would like to add this two functions.
It's simple, but having a length function that doesn't use sqrt() (for comparative purposes) would be nice when you need to calculate lots of distances each frame.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -208,7 +208,7 @@ RMDEF float Vector2Length(Vector2 v)
 }
 
 // Calculate vector square length
-RMDEF float Vector2SqrLength(Vector2 v)
+RMDEF float Vector2LengthSqr(Vector2 v)
 {
     float result = (v.x*v.x) + (v.y*v.y);
     return result;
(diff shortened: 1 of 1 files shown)


src/raymath.h
```diff
@@ -207,6 +207,13 @@ RMDEF float Vector2Length(Vector2 v)
     return result;
 }
 
+// Calculate vector square length
+RMDEF float Vector2SqrLength(Vector2 v)
```
raysan5, line 211: Please, could you review naming to `Vector2LengthSqr()` to better align with `Vector2Length()`?

src/raymath.h
```diff
@@ -401,6 +408,13 @@ RMDEF float Vector3Length(const Vector3 v)
     return result;
 }
 
+// Calculate vector square length
+RMDEF float Vector3SqrLength(const Vector3 v)
```
raysan5, line 412: Please, could you review naming to `Vector3LengthSqr()` to better align with `Vector3Length()`?

raysan5:
@ThePituLegend thanks for the improvement! :)

Issue #1198, opened 2020-04-12
raylib 3.1 roadmap


raysan5:
I started thinking about next raylib release, this time I planned to focus on library polishing and some quality-of-life changes, it will include some naming consistency changes (yeap, some API breaking) and also some new examples. 

Here a list with intended improvements:

 - `core`: Support dynamic memory loading at `InitWindow()` (for global state) -> `config.h` #1240
 - [x] `core`: Support RPI4 platform through DRM #1096 
 - [x] `core`: Redesign window config flags system #1367
 - [x] `rlgl`: Support mesh instancing #1127 
 - [x] `rlgl`: Redesign `GenTexture*()` functions using generic implementations #721 -> Not final yet but ok
 - [x] `raymath`: Review functions naming consistency #1197 
 - [x] `textures`: Unify `ImageDraw*()` functions with `Draw*()` counterpart equivalents https://github.com/raysan5/raylib/commit/b132da0ac2b0777cc965238fc3b011b1fa9b1fe5
 - `text`: Avoid `Image` data storage on `Font` atlas generation -> `config.h`
 - [x] `audio`: Support WAV fileformat for `Music` streaming
 - `utils`: Support VFS on `LoadFileData()` #1327 
 - [x] `build`: Support Continuous Deployment on GitHub Actions #1164
 - [x] `misc`: Add and improve [raylib on Wikipedia](https://en.wikipedia.org/wiki/Raylib)
 - [x] `misc`: Support `UWP` again #1182 **HELP WELCOME!**
 - [x] `misc`: Move `raylib/games` to separate repo (implies web changes)

Some interesting new examples I'd like to add: **HELP WELCOME!**

 - `core_3d_camera_touch` -> Support 3d model visualization with touch/gesture inputs
 - `core_vr_configurator` -> Configure VR parameters in realtime (with `raygui`)
 - `textures_cubemap` -> Load and display a cubemap (new API available)
 - [x] `textures_blending_modes` -> Illustrate multiple blending modes ([reference](https://www.andersriggelsen.dk/glblendfunc.php))
 - [x] `models_skybox` -> Review, not working on OpenGL ES 2.0, related to `Cubemap`
 - [x] `shaders_hot_reloading` -> Track shader on disk with `GetFileModTime()` to reload it

Just note that I'll keep updating this list as required!

chriscamacho:
There could be a lot to do here, might it be a good idea to divide it up amongst interested developers, this might stop people working on PR's doing the same thing...

[...]

PR #1261, opened 2020-05-21
Added Blend Modes example.

accidentalrebel:
Here is a working example showing off the different blend modes available. Pressing any key would toggle to the next blend mode. 

If more blend modes are added in the future, it'll be easy to update this by changing the `blendCountMax` variable.

Also added the example to the list of examples, as well as a preview image similar to what the others have.

This is part of #1198 raylib 3.1roadmap.


raysan5:
@accidentalrebel thanks for the contribution! :D

PR #1263, opened 2020-05-27
fix bug in QuaternionFromVector3ToVector3

jvocaturo:
Things were pointing the wrong way.

Classic XYY copy-paste error.


raysan5:
Good catch! Thanks for the fix! :D

PR #1269, opened 2020-06-05
Fixed buffer overflow in GenMeshPoly

frithrah:
Found and fixed a buffer overflow on the vertices array in GenMeshPoly when using certain numbers of sides - 11 or 50 are good examples.
Changed degree stepping to floats in order to close the poly accurately.
Added check on sides to prevent divide by zero and invalid numbers.

Thanks. :)


raysan5:
@frithrah thanks for the improvement! :)

PR #1271, opened 2020-06-08
Fix extra memory allocated when updating color buffer in mesh

4yn:
Updating the whole color buffer of a mesh would allocate and copy memory for `vertexCount * 4` `float`s instead of `unsigned char`, leading to extra memory allocation as well as data being copied beyond the color buffer.


raysan5:
@4yn thanks for the fix!

PR #1272, opened 2020-06-08
Fix wrong height used when using scissor mode with render texture of different height from window

4yn:
Scissor mode implementation uses global window height instead of the current framebuffer height when calculating scissor mode coordinates, leading to inconsistent scissor mode when drawing to smaller framebuffers.


raysan5:
@4yn thanks!

Issue #1270, opened 2020-06-08
[shaders] Skybox is flipped vertical


peppemas:
Hi, 

I had an issue with the skybox.c example running on my Windows 10 with the following configuration:

INFO: Initializing raylib 3.0
INFO: DISPLAY: Device initialized successfully
INFO:     > Display size: 1920 x 1080
INFO:     > Render size:  1920 x 1080
INFO:     > Screen size:  1920 x 1080
INFO:     > Viewport offsets: 0, 0
INFO: GLAD: OpenGL extensions loaded successfully
INFO: GL: OpenGL 3.3 Core profile supported
INFO: GL: OpenGL device information:
INFO:     > Vendor:   NVIDIA Corporation
INFO:     > Renderer: GeForce GTX 1060/PCIe/SSE2
INFO:     > Version:  3.3.0 NVIDIA 442.94
INFO:     > GLSL:     3.30 NVIDIA via Cg compiler
INFO: GL: Supported extensions count: 388
INFO: GL: DXT compressed textures supported
INFO: GL: ETC2/EAC compressed textures supported
INFO: GL: Anisotropic textures filtering supported (max: 16X)
INFO: GL: Mirror clamp wrap texture mode supported

the skybox appears v-flipped. I added to skybox.fs the support to flip the texture passing a bool. 
I attach a .zip containing the example with shaders and assets. 

[SKYBOX.zip](https://github.com/raysan5/raylib/files/4743887/SKYBOX.zip)

raysan5:
@peppemas Please, could you send a PR with the improvement? Thanks!

PR #1282, opened 2020-06-19
Fix bug #1270

peppemas:
Hi @raysan5 , 

Actually I added a parameter to the shaders in order to flip the cubemap, due to the fact that I don't know if this bug affects other OS (linux, osx, android etc.)


raysan5:
@peppemas Thanks for the improvement!

PR #1289, opened 2020-06-24
[add] GetMonitorRefreshRate(int monitor);

Shylie:
Note: only gets the monitor's current video settings. e.g. a 165 Hz monitor set to 60 Hz will return 60.


raysan5:
@Shylie Great addition! Thank you very much! :)

PR #1291, opened 2020-06-27
Added new function `DrawTextureTiled()`

Demizdor:
Hey @raysan5! I wanted a way to tile parts of a texture, on that point `DrawTextureQuad()` can be used for tilling a texture but only the full texture not parts of it as i wanted so i had to write this. Since it calls `DrawTexturePro()` in a loop i don't know how good it is performance wise but it works (well the rotation is a bit finicky but i left it as it is...maybe it is useful).

Don't know if it's really needed but i also added a example!

draw_texture_tiled


raysan5:
@Demizdor Hey! Very nice addition! Actually, the current implemented could be improved to be more performant. Currently it creates a separate quad for every tile but it's possible to just set the texture as repeat and play with textures coordinates to map into a single quad. `DrawTextureQuad()` does that but it seems it requires some extra parameters to allow that for a part of a texture...

Demizdor:
Hey, thanks for merging @raysan5
>  Actually, the current implemented could be improved to be more performant. Currently it creates a separate quad for every tile but it's possible to just set the texture as repeat and play with textures coordinates to map into a single quad.

AFAIK this is not possible with part of a texture only with the full texture. According to the internet there are 3 ways to tile part of the texture using opengl (using a shader, create a new texture with the part you want to tile and use that full texture, and what i did) but i'm not by any means an opengl master so maybe i'm missing something :)

I know performace isn't that great but i figured with raylib already batching all of the draw calls it won't be that bad, it's a compromise...Now, i'll admit the motive for adding this to the API was purely selfish (still working on https://github.com/raysan5/raygui/issues/82) :P

Issue #1304, opened 2020-07-09
format-security error rlgl_standalone.c


musteresel:
This line:

https://github.com/raysan5/raylib/blob/871cd1a76a41e9e7882ad51dbcd13e5f044a80a1/examples/others/rlgl_standalone.c#L222

causes a build using cmake with no specified options to fail on my system:

~~~
/build/raylib-3.0.0/examples/others/rlgl_standalone.c: In function 'ErrorCallback':
/build/raylib-3.0.0/examples/others/rlgl_standalone.c:222:5: error: format not a string literal and no format arguments [-Werror=format-security]
  222 |     fprintf(stderr, description);
      |     ^~~~~~~
~~~

Compiler:

~~~
-- The C compiler identification is GNU 9.2.0
-- The CXX compiler identification is GNU 9.2.0
~~~

System is NixOS Linux.

Fix:

~~~
fprintf(stderr, "%s", description);
~~~

PR #1305, opened 2020-07-09
Fix format-security error in rlgl_standalone.c

musteresel:
See #1304


raysan5:
@musteresel Thanks for the improvement!

PR #1309, opened 2020-07-14
scale perspective top calc for near culling plane

seanpringle:
I was adjusting `RL_CULL_DISTANCE_NEAR` and found the remaining hard coded `0.01` scaling value used for perspective top/right frustum calculations needed adjustment too. I suggest this change gives more intuitive results from `BeginMode3D` when people first try changing `RL_CULL_DISTANCE_NEAR` and expect only the near plane to move.


raysan5:
@seanpringle Oh! Good catch! Thanks for the improvement!

PR #1311, opened 2020-07-15
clamp ray hit y position to ground plane height

seanpringle:
When intersecting from a `Ray` to a `Y` ground plane, error can accumulate in resulting `RayHitInfo.position.y`, and it may be negative (beneath the supplied positive Y-normal plane).

Example: It's possible to call `GetCollisionRayGround(<mouse ray>, 0.0f)` and intermittently get results that flip flop between `(Vector3){..., 0.0, ...}` and `(Vector3){..., -0.000001, ...}`.

I suggest this should be cleanly clamped to `groundHeight` since we know the hit position must occur on the supplied plane, or not at all.


raysan5:
@seanpringle Agree! Nice improvement! :)

Issue #1127, opened 2020-03-09
[rlgl] Instanced rendering support


ChrisDill:
### Instanced Rendering

> "Instancing is a technique where we draw many objects at once with a single render call"

I learned about this technique from this guide from LearnOpenGL [here](https://learnopengl.com/Advanced-OpenGL/Instancing). I wanted to use it with raylib but I could not find a way to do so without modifications to rlgl. 

The goal is to make it easy and intuitive for a user new to raylib or instanced rendering and take their existing draw code and change it to a instanced version so they can quickly profile the performance difference and see if it makes sense for their program.

### DrawModelsInstanced

I have seen other frameworks have a seperate function for drawing models with instancing. This is a valid approach however it is limited to just models. To support more than that you would need a instanced version for each function and any you add in the future which adds a lot of complexity to the api.

```c
DrawModelsInstanced();
DrawModelsExInstanced();

DrawTexturesInstanced();
DrawTexturesExInstanced();

// And so on for all draw functions...
```

### Proposed API

```c
BeginModeInstanced(int instanceCount);
// Drawing here will draw instanceCount times
EndModeInstanced();
```

I starting with one function called `rlSetInstances(int instanceCount)` and got it to work and then split it into 2 functions once I looked at the Begin and End functions as that felt more consistent and intuitive. It needs a couple of variables added to `RLGL.State` to track if it is used and the instanceCount to draw with.

I also to modified `rlglDraw` and `rlDrawMesh` to use the instanced versions of those functions if instancing is currently being used.

[glDrawArraysInstanced](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDrawArraysInstanced.xhtml)
[glDrawElementsInstanced](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glDrawElementsInstanced.xhtml)

It looks like they are both available in OpenGL 3.1 or higher. I have not looked fully into supporting the feature on older OpenGL versions. It makes sense to put these changes around a `SUPPORT_GPU_INSTANCING` define although I have not tried that yet.

### Feature Screenshots

Note that these are not proper benchmarks. They are to see how many objects I can throw at the GPU using instancing while keeping a constant 60FPS. Proper benchmarks should be done across different platforms and setups.

The original bunnymark example drops FPS around 50000 bunnies. With instancing it can reach 500000+ bunnies.
[...]

PR #1318, opened 2020-07-20
[wip] rlDrawMeshInstanced

seanpringle:
Putting this here for feedback and discussion related to #1127.

`rlDrawMeshInstanced` is a low level approach to instanced mesh rendering. It loads a batch of transform matrices into a VBO, and requires a modified vertex shader using `LOC_MATRIX_MODEL` as an attribute location (not a uniform).

Todo:
- ~~Add a full code example to go with `base_lighting_instanced.vs`.~~
- Figure out if `rlDrawMeshInstanced` fits with the higher level API proposed in #1127. Might not.

### DIFF
--- a/examples/shaders/resources/shaders/glsl330/base_lighting_instanced.vs
+++ b/examples/shaders/resources/shaders/glsl330/base_lighting_instanced.vs
@@ -1,0 +1,36 @@
+#version 330
+
+// Input vertex attributes
+in vec3 vertexPosition;
+in vec2 vertexTexCoord;
+in vec3 vertexNormal;
+in vec4 vertexColor;
+
+layout (location = 12) in mat4 instance;
+
+// Input uniform values
+uniform mat4 mvp;
+
+// Output vertex attributes (to fragment shader)
+out vec3 fragPosition;
+out vec2 fragTexCoord;
+out vec4 fragColor;
+out vec3 fragNormal;
+
+// NOTE: Add here your custom variables
+
+void main()
+{
+    // Send vertex attributes to fragment shader
+    fragPosition = vec3(instance * vec4(vertexPosition, 1.0));
+    fragTexCoord = vertexTexCoord;
+    fragColor = vertexColor;
+
+    mat3 normalMatrix = transpose(inverse(mat3(instance)));
+    fragNormal = normalize(normalMatrix * vertexNormal);
+
+    mat4 mvpi = mvp * instance;
+
+    // Calculate final vertex position
+    gl_Position = mvpi * vec4(vertexPosition, 1.0);
+}
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -123,12 +123,7 @@
     } Vector4;
 
     // Quaternion type
-    typedef struct Quaternion {
-        float x;
-        float y;
-        float z;
-        float w;
-    } Quaternion;
+    typedef Vector4 Quaternion;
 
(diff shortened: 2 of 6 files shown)


src/rlgl.h
```diff
@@ -548,6 +548,7 @@ RLAPI void rlLoadMesh(Mesh *mesh, bool dynamic);                          // Upl
 RLAPI void rlUpdateMesh(Mesh mesh, int buffer, int count);                // Update vertex or index data on GPU (upload new data to one buffer)
 RLAPI void rlUpdateMeshAt(Mesh mesh, int buffer, int count, int index);   // Update vertex or index data on GPU, at index
 RLAPI void rlDrawMesh(Mesh mesh, Material material, Matrix transform);    // Draw a 3d mesh with material and transform
+RLAPI void rlDrawMeshInstanced(Mesh mesh, Material material, int count, Matrix *transforms);    // Draw a 3d mesh with material and transform
```
raysan5, line 551: I like this function, just proposing a change in signature: 
```c
RLAPI void rlDrawMeshInstanced(Mesh mesh, Material material, Matrix *transforms, int count);
```

src/raymath.h
```diff
@@ -114,6 +114,14 @@
         float z;
     } Vector3;
 
+    // Vector4 type
```
raysan5, line 117: Actually, Quaternion could also be changed in this case:
```c
// Vector4 type
typedef struct Vector4 {
    float x;
    float y;
    float z;
    float w;
} Vector4;

// Quaternion type, same as Vector4
typedef Vector4 Quaternion;
```

src/rlgl.h
```diff
@@ -570,6 +571,7 @@ RLAPI Rectangle GetShapesTextureRec(void);                                // Get
 
 // Shader configuration functions
 RLAPI int GetShaderLocation(Shader shader, const char *uniformName);              // Get shader uniform location
+RLAPI int GetShaderAttribLocation(Shader shader, const char *attribName);         // Get shader attribute location
```
raysan5, line 574: I propose a change in name for  a bit more consistency with `GetShaderLocation()`:
```c 
RLAPI int GetShaderLocationAttrib(Shader shader, const char *attribName);
```

examples/shaders/resources/shaders/glsl330/base_lighting_instanced.vs
```diff
@@ -0,0 +1,36 @@
+#version 330
```
raysan5, line 1: Not sure if this file should be here, it would require the instancing example.

raysan5:
@seanpringle Please, could you take a look to the review points? I'm about to merge to merge this PR.

raysan5:
@ChrisDill please, could you also take a look to the PR? Do you think it's fine to merge this lower-level implementation.

src/rlgl.h
seanpringle: Done

src/raymath.h
seanpringle: Done

examples/shaders/resources/shaders/glsl330/base_lighting_instanced.vs
seanpringle: Added an example. Maybe not necessary for something low level, but there it is :)

src/rlgl.h
seanpringle: Done

raysan5:
Thanks for this new feature! Related to issue: #1127

PR #1321, opened 2020-07-24
rlCheckErrors

seanpringle:
Useful to see some log noise when I break rlgl :)

Actually I am surprised not to see `glGetError` anywhere. Are OpenGL errors meant to be caught in some other way? I think at least `GL_OUT_OF_MEMORY` is possible for any raylib user to hit even if they do not touch rlgl directly.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1383,8 +1383,6 @@ void EndDrawing(void)
     }
 #endif
 
-    rlglCheckErrors();
-
     SwapBuffers();                  // Copy back buffer to front buffer
     PollInputEvents();              // Poll user events
 
--- a/src/core.c
+++ b/src/core.c
@@ -1383,6 +1383,8 @@ void EndDrawing(void)
     }
 #endif
 
+    rlglCheckErrors();
+
     SwapBuffers();                  // Copy back buffer to front buffer
     PollInputEvents();              // Poll user events
 
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -519,7 +519,7 @@ RLAPI unsigned int rlLoadAttribBuffer(unsigned int vaoId, int shaderLoc, void *b
 RLAPI void rlglInit(int width, int height);           // Initialize rlgl (buffers, shaders, textures, states)
 RLAPI void rlglClose(void);                           // De-inititialize rlgl (buffers, shaders, textures)
 RLAPI void rlglDraw(void);                            // Update and draw default internal buffers
-RLAPI void rlglCheckErrors(void);                     // Check and log OpenGL error codes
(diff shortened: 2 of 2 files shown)


src/core.c
```diff
@@ -1383,6 +1383,8 @@ void EndDrawing(void)
     }
 #endif
 
+    rlglCheckErrors();
```
raysan5, line 1386: I prefer not forcing the errors check for the cost it can suppose, I'd just leave usage up to the user.

src/rlgl.h
```diff
@@ -519,6 +519,7 @@ RLAPI unsigned int rlLoadAttribBuffer(unsigned int vaoId, int shaderLoc, void *b
 RLAPI void rlglInit(int width, int height);           // Initialize rlgl (buffers, shaders, textures, states)
 RLAPI void rlglClose(void);                           // De-inititialize rlgl (buffers, shaders, textures)
 RLAPI void rlglDraw(void);                            // Update and draw default internal buffers
+RLAPI void rlglCheckErrors(void);                     // Check and log OpenGL error codes
```
raysan5, line 522: About naming, I'd prefer to call it just `rlCheckErrors()`, actually `rlglDraw()` is going to be removed and replaced by `rlDrawRenderBatch()`, added lately, so the library will just keep `rlglInit()` and `rlglClose()` that looks more consistent to me.

src/rlgl.h
```diff
@@ -1797,6 +1798,45 @@ void rlglDraw(void)
...
+        const GLenum err = glGetError();
+        switch (err) {
+            case GL_NO_ERROR:
+                check = 0;
+                break;
+            case 0x0500: // GL_INVALID_ENUM:
+                TRACELOG(LOG_WARNING, "GL_INVALID_ENUM");
+                break;
+            case 0x0501: //GL_INVALID_VALUE:
+                TRACELOG(LOG_WARNING, "GL_INVALID_VALUE");
```
raysan5, line 1815: I try to categorize the TRACELOG() messages, maybe something like:
```c
TRACELOG(LOG_WARNING, "GL: Error detected: GL_INVALID_VALUE");
```

src/rlgl.h
seanpringle: Sounds good. Done.

src/rlgl.h
seanpringle: I feel like this could be better with more detail, but I suppose it should be rare enough and probably relevant only to direct rlgl.h users, so a log notice with error code is enough.

raysan5:
@seanpringle Thanks for the improvement!

PR #1324, opened 2020-07-25
Multiple blendmodes

cedemax:
Adds a new function BeginBlendModeEx that permits the use of any blend factors and blend equation. I have used this in a project (Windows) so think it works as intended. Also adds a default Subtract blend mode for use with the normal BeginBlendMode.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -808,7 +808,8 @@ typedef enum {
     BLEND_ALPHA = 0,        // Blend textures considering alpha (default)
     BLEND_ADDITIVE,         // Blend textures adding colors
     BLEND_MULTIPLIED,       // Blend textures multiplying colors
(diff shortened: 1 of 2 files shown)


raysan5:
@cedemax Still thinking about the best way to implement this. Current implementation with `BeginBlendModeEx()` forces user to include OpenGL header to define the available src-dst modes and I'd like to avoid that.

Also, as commented on Discord, I'm trying to avoid `*Ex()` versions of `Begin*() / End*()` functions. One possible alternative is creating a new function:
```c
SetBlendModeCustom(int srcFactor, int dstFactor, int equation);
```
And then allow:
```c
BeginBlendMode(BLEND_CUSTOM);
```

raysan5:
@cedemax Thanks for the implementation. Despite merging, I'm changing the behaviour as previously commented.

PR #1330, opened 2020-07-30
Fix color of ambient light

xdrie:
The shader as it is right now renders it like this:
image

But with the change, it looks like this:
image

The problem is that the shader currently is not multiplying the ambient color by the texture diffuse color, so it's just using a white texture multiplied by the `ambient` vector as the diffuse layer. We can simply resolve this by just multiplying by the diffuse color.


raysan5:
@xdrie oh! nice! thanks for the improvement!

PR #1346, opened 2020-08-10
Making the windows script faster

henriquel1997:
I got tired of waiting for the build script to do its job so i spent i little time figuring out why it was so slow. 
Turns out it the culprit was the vcvarsall.bat script that was getting called every time. 
So i modified it to only call it if it wasn't called before, so the script is only slow the first time it is called, which isn't the ideal scenario, but it is certainly better than it was before.


raysan5:
@henriquel1997 Nice! Thanks for the improvement! :D

PR #1347, opened 2020-08-11
Change raylib version for VSCode mingw project makefile

Gaetz:
It seems to be the only place where the version was not updated.


raysan5:
@Gaetz thanks for the review! 😄

PR #1353, opened 2020-08-14
Small change to text input box example

someone-existing:
You don't have to spam backspace anymore.


raysan5:
@someone-existing Actually that's not the most convenient way to do this.

It will remove 60 characters per seconds and users won't have fine control over it. The right way to track this issue is using a time counter.

PR #1359, opened 2020-08-22
work on quat and matrix math - deleted multiple copies of raymath.h …

chriscamacho:
…causing issues

-isystem doesn't seem to work for me, is there a specific reason to use it?

its worth noting for quat eular operations that its important to use a ZYX matrix rotation order.


raysan5:
@chriscamacho Thanks for the improvement, I'll review it now.

PR #1361, opened 2020-08-30
Added Never language to the list of bindings.

smaludzi:
Please add information about Never language binding in project documentation.


raysan5:
@smaludzi Oh! Thank you very much for the addition! :D

PR #1375, opened 2020-09-13
Add Gforth bindings and update Factor bindings

ArnautDaniel:
Added Gforth bindings and updated the link to the Factor bindings.  I did the Factor ones too, but changed my name on Github recently.  Now it points to the correct link.


raysan5:
@ArnautDaniel nice! 👍😄

PR #1379, opened 2020-09-17
fixed wrong error message for input device on RPI

kernelkinetic:
error message missed device name and printed file descriptor (which should be always -1 at this line) instead of error number

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -158,6 +158,7 @@
 #include <string.h>                 // Required for: strrchr(), strcmp(), strlen()
 #include <time.h>                   // Required for: time() [Used in InitTimer()]
 #include <math.h>                   // Required for: tan() [Used in BeginMode3D()]
+#include <errno.h>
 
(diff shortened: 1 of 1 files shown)


src/core.c
```diff
@@ -4602,7 +4603,7 @@ static void EventThreadSpawn(char *device)
     fd = open(device, O_RDONLY | O_NONBLOCK);
     if (fd < 0)
     {
-        TRACELOG(LOG_WARNING, "RPI: Failed to open input device (error: %d)", device, worker->fd);
+        TRACELOG(LOG_WARNING, "RPI: Failed to open input device %s (error: %d)", device, errno);
```
raysan5, line 4606: I try to avoid including additional libraries... could we just remove the error number information?

kernelkinetic: The error number gives a hint why opening failed. But of course you can remove the include and write:
```
TRACELOG(LOG_WARNING, "RPI: Failed to open input device %s", device);
```

raysan5:
@kernelkinetic thanks! 👍😄

PR #1381, opened 2020-09-19
Updated joystick mappings

coderoth:
Updated joystick mappings with latest version of gamecontrollerdb (executed GenerateMappings.cmake), so that raylib can identify more joysticks.

To generate the file from raylib repo root:
cd src/external/glfw/CMake
cmake -P GenerateMappings.cmake ../src/mappings.h.in ../src/mappings.h


raysan5:
@coderoth thanks for the improvement!

PR #1382, opened 2020-09-20
update raylib-nelua repository url

Andre-LA:
The repository is not a mirror anymore.


raysan5:
@Andre-LA thanks for the update!

PR #1388, opened 2020-09-24
Platform DRM

kernelkinetic:
- PLATFORM_DRM added to source and examples makefiles and to cmake build system
- SUPPORT_MOUSE_CURSOR_RPI changed to SUPPORT_MOUSE_CURSOR_DRM, mouse cursor code added from @chriscamacho
- public `SetGraphicDeviceName(...)` added to let the user select the graphics device to use (useful only for PLATFORM_DRM)
- Config flag `FLAG_INTERLACED_HINT` added to let the DRM mode selection allow to choose also interlaced graphics mode, otherwise only a progressive mode is choosen

Please tell me if should any comments, changes etc.!

On Raspberry Pi 1-4 `SetGraphicDeviceName(...)` shouldn't be needed, it should select automatically the available one. On other Linux distributions the device name could differ, e. g. depending of the grapics card manufacturer.


kernelkinetic:
The Travis CI needs additional build jobs for PLATFORM_DRM and preinstalled libgbm, libdrm and maybe libmesa or whatever provides EGL and OpenGL ES libraries. It should work on any Linux distribution that provides GBM and DRM which is any modern Linux distribution as far as I know.

examples/Makefile
```diff
@@ -221,6 +230,9 @@ endif
 ifeq ($(PLATFORM),PLATFORM_RPI)
     CFLAGS += -std=gnu99
 endif
+ifeq ($(PLATFORM),PLATFORM_DRM)
+    CFLAGS += -std=gnu99 -DEGL_NO_X11
```
TSnake41, line 234: EGL_NO_X11 could also be added directly to the source code, rlgl doesn't depend on X11 data structures after all.

kernelkinetic: To be honest I'm not sure why this didn't came up earlier. The EGL header `eglplatform.h` by default pulls in `Xlib.h` (which seems wrong for me, but it does anyways) which pulls in another X header that declares a structure `Font` and which in turn conflicts with raylib's structure `Font`. This leads me to add EGL_NO_X11 to prevent EGL from including the X11 headers.
On the other hand it seems that EGL and X11 is possible so maybe PLATFORM_DRM could be merged with PLATFORM_DESKTOP in future, so I decided first to not change/implement any other than the DRM related code.

src/core.c
```diff
@@ -5373,3 +5758,125 @@ void UWPGestureTouch(int pointer, float x, float y, bool touch)
 }
 
 #endif // PLATFORM_UWP
+
+void SetGraphicDeviceName(const char *name)
```
raysan5, line 5762: This function shouldn't be required. 

In any case, it can be used: `#define DEFAULT_GRAPHIC_DEVICE_DRM "/dev/dri/card1"`

or simply let checking `card1` and `card0` (current implementation)

src/core.c
```diff
@@ -337,10 +346,20 @@ typedef struct CoreData {
         GLFWwindow *handle;                 // Native window handle (graphic device)
 #endif
 #if defined(PLATFORM_RPI)
-        // NOTE: RPI4 does not support Dispmanx anymore, system should be redesigned
         EGL_DISPMANX_WINDOW_T handle;       // Native window handle (graphic device)
 #endif
-#if defined(PLATFORM_ANDROID) || defined(PLATFORM_RPI) || defined(PLATFORM_UWP)
+#if defined(PLATFORM_ANDROID) || defined(PLATFORM_RPI) || defined(PLATFORM_DRM) || defined(PLATFORM_UWP)
+#if defined(PLATFORM_DRM)
+        const char *card;                   // /dev/dri/... file name
```
raysan5, line 353: This variable is not required, check next review comment.

raysan5:
@kernelkinetic Added two review comments.

About Travis CI, no worries, that script should be reviewed.

kernelkinetic:
@raysan5 I would prefer the automatic selection if no `DEFAULT_GRAPHIC_DEVICE_DRM` is defined. So one can choose a specific device if necessary and it works out of the box otherwise. I'll also look if I can automatically select by checking all available interfaces if they can render, that's just a flag to check if I'm remembering it correctly. Anyways, I think I'll send a commit tomorrow ☺️

gen2brain:
Go bindings, not that it matters much, would prefer function, nice and clear, and can be accessed easily. Users can also pass `CGO_CFLAGS='-DDEFAULT_GRAPHIC_DEVICE_DRM="/dev/dri/card1"'` if and when they need,  but if it can be done with some detection/checking the best. I guess rare would be the cases where the user really wants to pick.

kernelkinetic:
This commit uses the `DEFAULT_GRAPHIC_DEVICE_DRM` string if it's defined and otherwise tries for `/dev/dri/card1` and if that fails for `/dev/dri/card0` to support Raspberry Pi 0..4.

I've also implemented automatic selection of the render device but it seems that there are some quirks in Raspberry Pi driver implementation which makes it working on Linux but not on Raspberry Pi or vice versa. So at the moment this commit is what gives the most flexibility so far.

raysan5:
Merged! :D

Thanks to all the contributors for this amazing work!

Issue #1393, opened 2020-09-29
No make compile with CMake


paklau99988:
### Issue description
I clone the repo, try build get error. 

### Environment

Win 10, use VC 19 

### Issue Screenshot

![image](https://user-images.githubusercontent.com/65712388/94554364-a26c7700-0262-11eb-91dc-c8233379bfe7.png)

### Code Example

No code, just cmake want make work

kernelkinetic:
This is a duplicate of #1391, please look at https://github.com/raysan5/raylib/issues/1096#issuecomment-699603920 and following comment(s) for the temporary solution.

PR #1395, opened 2020-09-29
Fix 1393

tjammer:
This fixed the cmake issue in #1393.

The problem here is that target_link_libraries for raylib_static is called once with keyword "PUBLIC" (line 189) and once without keyword (line 195). Cmake does not allow mixing these signatures. To fix, either have both calls without keyword or both with.


kernelkinetic:
@tjammer Absolutely, why I didn't see this? :/

raysan5:
@tjammer thanks for the review! 👍😄

PR #1397, opened 2020-09-30
mouse: Return float movement for precise scrolling where possible

Doy-lee:
Making mouse wheel return a float means the threshold between not scrolling and scrolling is more sensitive and allows for finer grained response.

Right now, especially on desktop through GLFW, we truncate the scroll movement from double to int and so we lose quite a bit of data. Implementing for example a window that can be scrolled by the mouse is very choppy because we lose that precision.


raysan5:
@Doy-lee Thanks for the improvement

PR #1398, opened 2020-10-01
Add some bindings' versions

Intasx:
Raylib-forever: 3.1-dev
It generates the c-to-nim bindings straight from the source.

nim-raylib: 3.1-dev
It uses Raylib-forever.

raylib-php: 3.0
Tested it using the provided binaries.

raylib-java: 2.0
Tested it using the provided binaries.


raysan5:
@Intasx thanks for the addition! 👍😄

PR #1400, opened 2020-10-03
Add Vector2Reflect to raymath.h

daniel-junior-dube:
Vector3Reflect exists but not Vector2Reflect. The code is pretty much the same.
I'm not sure what RMDEF does, but I added it to match other function definitions (haven't done much C programming, maybe I'm missing something). Can someone explain to me what it does?


raysan5:
@daniel-junior-dube Thanks for the addition!

`RMDEF` just declares different type of symbols for function depending on the compilation mode of the module (static, inline, as a dll...).

PR #1411, opened 2020-10-17
added GetRandomFloat and SetGridColours

chriscamacho:


raysan5:
@chriscamacho Sorry, no plans to add those additional functions at the moment.

 - `GetRandomValue()` can already be used to get float values with a bit of user math.
 - `SetGridColors()` requires additional global variables that I prefer to avoid. In that line, I could consider redesigning `DrawGrid()` to receive a base color (and add a #define to avoid breaking old code bases).

chriscamacho:
...  used to get float values with a bit of user math.
which is a PITA every time I do it, hence the addition

urg more macro's, I guess you could do it that way, but given core has a big blob of globals why not make them available to the whole of raylib...

PR #1415, opened 2020-10-25
ADDED: GetWindowTitle()

RobLoach:
As a mirror of `SetWindowTitle()`, the `GetWindowTitle()` function will retrieve the title of the window.


raysan5:
@RobLoach Actually, function could make sense for the mirror but... do you know any situation where it is required?

RobLoach:
It's not required. Found I was binding the title. Needed properties to get/set, but it's easy to get around.

Issue #1392, opened 2020-09-27
[core] Mouse movements are bound to the screen resolution


GitSparTV:
### Issue description

On the maze example the mouse movements are bound to the screen resolution I have.
Also the camera movements are shuttering when I hold move key.

### Environment

Raspberry PI 4 no X11, 446f9ff commit. PLATFORM_DRM.

GitSparTV:
Video:
https://imgur.com/R05SyMt

GitSparTV:
Log:
```
INFO: Initializing raylib 3.1-dev
INFO: DISPLAY: no graphic card set, trying card1
INFO: DRM: choosen mode 848x480 (848x480p@60)
INFO: DISPLAY: Upscaling required: Screen size (800x450) smaller than display size (848x480)
INFO: DISPLAY: Device initialized successfully
INFO:    > Display size: 848 x 480
INFO:    > Render size: 800 x 453
INFO:    > Screen size: 800 x 450
INFO:    > Viewport offsets: 0, 3
INFO: GL: OpenGL device information:
INFO:    > Vendor: Broadcom
INFO:    > Renderer: V3D 4.2
INFO:    > Version: OpenGL ES 3.1 Mesa 19.3.2
INFO:    > GLSL: OpenGL ES GLSL ES 3.10
INFO: GL: Supported extensions count: 76
INFO: GL: VAO extension detected, VAO functions initialized successfully
INFO: GL: NPOT textures extension detected, full NPOT textures supported
INFO: GL: ETC1 compressed textures supported
INFO: TEXTURE: [ID 1] Texture created successfully (1x1 - 1 mipmaps)
INFO: TEXTURE: [ID 1] Default texture loaded successfully
INFO: SHADER: [ID 1] Compiled successfully
INFO: SHADER: [ID 2] Compiled successfully
INFO: SHADER: [ID 3] Program loaded successfully
INFO: SHADER: [ID 3] Default shader loaded successfully
INFO: RLGL: Internal vertex buffers initialized successfully in RAM (CPU)
INFO: RLGL: Render batch vertex buffers loaded successfully
INFO: RLGL: Default state initialized successfully
INFO: TEXTURE: [ID 2] Texture created successfully (128x128 - 1 mipmaps)
INFO: FONT: Default font loaded successfully
INFO: RPI: Opening input device: /dev/input/event1 (mouse )
INFO: RPI: Opening input device: /dev/input/event0 (keyboard )
[...]

PR #1422, opened 2020-11-01
Fixed keyboard stuttering but for cmake only

kernelkinetic:
fixed https://github.com/raysan5/raylib/issues/1392 for cmake build system


raysan5:
@kernelkinetic thanks for the fix! :D

PR #1428, opened 2020-11-10
Fixed zero-window upscaling

Ilya3point999K:
There is no zero-check, so window upscales to nothing.
SetupFramebuffer() is kinda wrong, it uses not its params, but global variables. I won't touch it, maybe it has purpose


raysan5:
@Ilya3point999K thanks for the review

PR #1431, opened 2020-11-14
Add RAYMATH_TYPES_DEFINED

maihd:
Add RAYMATH_TYPES_DEFINED to check if Vector*/Matrix* is defined, that we can use RAYMATH_STANDALONE with raylib.h

```C
#define RAYMATH_STANDALONE
#include <raymath.h>
#include <raylib.h> // It should work well
```

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -170,37 +170,41 @@
     typedef enum { false, true } bool;
 #endif
 
-// Vector2 type
-typedef struct Vector2 {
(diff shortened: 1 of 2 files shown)


raysan5:
@maihd Actually you can use raymath with raylib this way:
```c
#include "raylib.h"
#include "raymath.h"
```

maihd:
Thanks, @raysan5.

PR #1437, opened 2020-11-25
Cleanup typecast warnings in visual studio 2019

JeffM2501:
There are a number of type cast warnings when building the library in visual studio. This PR adds casts to those places to prevent the warnings and make the intent of the code more clear.


JeffM2501:
I guess size_t isn't in as many compilers as I hoped. nevermind.

PR #1444, opened 2020-11-29
Added Apple Compilation Instructions

SethArchambault:
Stumbled on this example and I love this! Adding instructions that work for me on MacOS.. Note that glfw3 will need to be statically built locally and copied to this external/libs directory on mac. I can upload my static version, but it probably makes sense to do this only if there's some general naming convention for adding platform specific folders. Like "external_osx/lib" or "external/lib_osx". Then I'll drop my static libs in there.


raysan5:
@SethArchambault nice! 👍😄

PR #1445, opened 2020-11-29
Added GLFW_OPENGL_FORWARD_COMPAT for Apple Support

SethArchambault:
Without GLFW_OPENGL_FORWARD_COMPAT, running this as a standalone will yield the error:
```
NSGL: The targeted version of macOS only supports forward-compatible core profile contexts for OpenGL 3.2 and above
```


raysan5:
@SethArchambault oh! true! Actually I enable it on raylib but I forgot about this example! Nice catch! Thanks! :D

PR #1449, opened 2020-12-05
Add function to get the position of a monitor (and fix some comments)

JeffM2501:
This PR adds a function to the API that returns information about multiple displays on desktop platforms that tells you where a monitor is relative to the primary monitor, so you can move a window onto it. This PR also changes some comments that refer to the primary monitor but are attached to functions that deal with selected monitors.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1437,7 +1437,26 @@ int GetMonitorCount(void)
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


Gamerfiend (approved):

raysan5:

PR #1450, opened 2020-12-05
Fixed an issue where SetAudioBufferPitch could not be dynamically changed

jdeokkim:
I tried making a simple program that changes the pitch of the music dynamically and found out that the pitch was not changing at all. 

```c
#include "raylib.h"

#define TARGET_FPS 60

#define DEFAULT_WIDTH 800
#define DEFAULT_HEIGHT 600

Music music_example;

void LoadResources(void);

void UnloadResources(void);

void UpdateCurrentScreen(void);

int main(void) {
    SetTargetFPS(TARGET_FPS);
        
    InitAudioDevice();
    InitWindow(DEFAULT_WIDTH, DEFAULT_HEIGHT, "some example");
    
    LoadResources();
    
    while (!WindowShouldClose())
        UpdateCurrentScreen();
    
    UnloadResources();

    CloseAudioDevice();
    CloseWindow();
    
    return 0;
}

void LoadResources(void) {
    music_example = LoadMusicStream("res/example.mp3");
    PlayMusicStream(music_example);
}

void UnloadResources(void) {
    StopMusicStream(music_example);
    UnloadMusicStream(music_example);
}

void UpdateCurrentScreen(void) {
    BeginDrawing();
    
    ClearBackground(RAYWHITE);
    
    if (IsKeyDown(KEY_UP))
        SetMusicPitch(music_example, 1.25f);
    else if (IsKeyDown(KEY_DOWN))
        SetMusicPitch(music_example, 0.75f);
    else;
    
    UpdateMusicStream(music_example);
    
    DrawFPS(8, 8);
    
    EndDrawing();
}
```

So I modified the `SetAudioBufferPitch` function to print the value of `pitchMul`, expected value of the changed sample rate and `newOutputSampleRate` and got these results. As you can see, when `buffer->pitch` equals to the second argument of `SetAudioBufferPitch` (`float pitch`), the pitch of the audio won't change compared to what we expected. While this might not be an appropriate solution to this problem, at least adding `buffer->pitch != pitch` to the `if` statement resolved this issue for me. Please let me know if you have any ideas about this problem. Thank you very much!

```c
INFO: AUDIO: Device initialized successfully
INFO:     > Backend:       miniaudio / PulseAudio
INFO:     > Format:        32-bit IEEE Floating Point -> 16-bit Signed Integer
INFO:     > Channels:      2 -> 2
INFO:     > Sample rate:   44100 -> 44100
INFO:     > Periods size:  1320
INFO: AUDIO: Multichannel pool size: 16
INFO: Initializing raylib 3.1-dev
INFO:     > Display size: 1680 x 1050
INFO:     > Render size:  800 x 600
INFO:     > Screen size:  800 x 600
INFO: GL: OpenGL 3.3 Core profile supported
INFO:     > Vendor:   Intel Open Source Technology Center
INFO:     > Renderer: Mesa DRI Intel(R) HD Graphics (Coffeelake 3x8 GT2) 
INFO:     > Version:  4.5 (Core Profile) Mesa 18.0.0-rc5
INFO:     > GLSL:     4.50
INFO: GL: Supported extensions count: 191
INFO: GL: Anisotropic textures filtering supported (max: 16X)
INFO: TEXTURE: [ID 1] Texture created successfully (1x1 - 1 mipmaps)
INFO: SHADER: [ID 1] Compiled successfully
INFO: SHADER: [ID 2] Compiled successfully
INFO: SHADER: [ID 3] Program loaded successfully
INFO: RLGL: Internal vertex buffers initialized successfully in RAM (CPU)
INFO: RLGL: Render batch vertex buffers loaded successfully
INFO: RLGL: Default state initialized successfully
INFO: TEXTURE: [ID 2] Texture created successfully (128x128 - 1 mipmaps)
INFO: FONT: Default font loaded successfully
INFO: STREAM: Initialized successfully (44100 Hz, 32 bit, Stereo)
INFO: FILEIO: [res/example.mp3] Music file successfully loaded:
INFO:     > Total samples: 34205184
INFO:     > Sample rate:   44100 Hz
INFO:     > Sample size:   32 bits
INFO:     > Channels:      2 (Stereo)

pitchMul: 1.250000
target: 35280
result: 35280
pitchMul: 1.000000
target: 35280
result: 44100
pitchMul: 1.000000
target: 35280
result: 44100

...
```

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -632,7 +632,7 @@ void SetAudioBufferVolume(AudioBuffer *buffer, float volume)
 // Set pitch for an audio buffer
 void SetAudioBufferPitch(AudioBuffer *buffer, float pitch)
 {
-    if (buffer != NULL)
+    if (buffer != NULL && buffer->pitch != pitch)
     {
         float pitchMul = pitch/buffer->pitch;
 


raysan5:
@jdeokkim Mmmh... I see... it seems the issue is related to calling the function multiple frames in a row, it should probably only be called once... try with the following approach:
```c
if (IsKeyPressed(KEY_UP)) SetMusicPitch(music_example, 1.25f);
else if (IsKeyPressed(KEY_DOWN)) SetMusicPitch(music_example, 0.75f);
``

**Comment by jdeokkim** [comment]
@raysan5 I tried that approach and it still did not work. That null `else ;` statement was intentional though, I don't know why but I feel uncomfortable when there is no `else` in an `if-else if` statement. 

There is one more reason why I think adding `buffer->pitch != pitch` condition to the if statement is not a bad idea, there is no need to change pitch of the audio buffer when `buffer->pitch` already equals to `pitch`.

**Comment by raysan5** [comment]
@jdeokkim Actually, `SetMusicPitch()` is not intended to be called in a loop but only once, with the proposed condition: `(buffer->pitch != pitch)` it will only be changed the first the first time you call the function inside a loop. Also, you are comparing two `float` values, that's very insecure due to the rounding issues of floats depending on the platform.

Could you try calling `SetMusicPitch(music_example, 1.25f)` on `UnloadResources()`? It should work there...

**Comment by jdeokkim** [comment]
@raysan5 I totally forgot about the internal precision errors when comparing float values... I guess I need to make two variables for storing the previous and current pitch values to check the delta of the pitch and make sure to call `SetMusicPitch` only once per stored pitch values. Thank you for the solution!

**Comment by jdeokkim** [comment]
@raysan5 Another question, if `SetMusicPitch()` is not intended to be called in a loop, how am I supposed to continuously change the pitch of the audio? For example, there might be a situation where you want to increment or decrement pitch based on key events like the example below:

```c
    /* ... */

    if (IsKeyDown(KEY_UP))
        music_pitch += 0.01f;
    else if (IsKeyDown(KEY_DOWN))
        music_pitch -= 0.01f;
    else if (IsKeyDown(KEY_ENTER))
        music_pitch = 1.0f;
    
    if (music_pitch <= 0.5f)
        music_pitch = 0.5f; 
    else if (music_pitch >= 2.0f)
        music_pitch = 2.0f;
    
    /* probably do something like `SetMusicPitch()` here */
    UpdateMusicStream(music_example);
}
```

I tried using `SetMusicPitch()` in this case, but it did not seem to update the pitch of the audio properly. Resetting the pitch of the music to initial value before actually changing the pitch seems to be working though...

```c
UpdateMusicStream(music_example);

SetMusicPitch(*music, 1.0f);  // reset the pitch of the music first
SetMusicPitch(*music, pitch); // then change the pitch to the desired value
```

raysan5:
@jdeokkim You are right. I reviewed the internal implementation to use directly the pitch you set instead of using that value as a multiplier for the current pitch. I also added some example code in `audio_module_playing`.

Now it should work as expected.

jdeokkim:
@raysan5 Thanks a lot! Now I can start working on my cool games again :sunglasses:

PR #1451, opened 2020-12-05
Add enable/disable for depth mask

JeffM2501:
Add functions to enable/disable depth writes to go with functions for depth tests. Since depth test enable/disable is exposed via rlgl, depth writes (mask) should be exposed as well.


raysan5:

PR #1452, opened 2020-12-08
Fixed the build for web using CMake.

object71:
I found that the build for me was failing and I added some if defined checks in the core.c file where the glfwSetWindowAttrib was used. (error: implicit declaration of function 'glfwSetWindowAttrib' is invalid in C99 [-Werror,-Wimplicit-function-declaration])

I also changed some values in the toolchain file so that it correctly uses the .bat files when on windows.


object71:
@raysan5 Hi, I made some improvements on the CMakeLists.txt files. The project wasn't build-able for me for _web on windows_ and I decided to improve some of the cmake configurations.

I edited some values in the cmake/emscripten.cmake toolchain that is local in the project as on windows it didn't correctly execute the .bat compilers.

I added the option to detect EMSCRIPTEN variable set by the emscripten toolchain. The emscripten toolchain is set by default when the emcmake is used before cmake.

I fixed an issue with a replace value by putting quotes around it (the value was empty for some reason for me)

I fixed the extension filename generation for web to properly generate .html files.

raysan5:
@object71 Nice! Thanks for the improvements! I don't maintain the CMake build system... Maybe @a3f could take a look to them?

a3f (approved):
CMake parts look ok to me.

raysan5:
@a3f Nice! Thanks for the review!

object71:
@raysan5 Thanks, I'll see to update the documentation on building for web using cmake now.

PR #1456, opened 2020-12-12
Fixed #1455

### LINKED ISSUE #1455: [examples] text_input_box.c writes out of array bounds
### Issue description

In https://github.com/raysan5/raylib/blob/master/examples/text/text_input_box.c ,
this code:

```
if (IsKeyPressed(KEY_BACKSPACE))
{
  letterCount--;
  name[letterCount] = '\0';

  if (letterCount < 0) letterCount = 0;
}
```

writes at position -1 if we started with a letterCount of 0 (which is the default value). That is tested for, but only after we already wrote there.

### Environment

WIN64

### Issue Screenshot

Unnecessary.

### Code Example

https://github.com/raysan5/raylib/blob/master/examples/text/text_input_box.c
LINE 65

### Fix

Swap IF and array access (line 68 and 70)

Buhlean:
Fixed writing out of array bounds
Adjusted FPS comment to match value
Deleted unused function at the end, which has never been in use in the history of this file

For Issue: https://github.com/raysan5/raylib/issues/1455

### DIFF
--- a/examples/text/text_input_box.c
+++ b/examples/text/text_input_box.c
@@ -112,3 +112,15 @@ int main(void)
 
     return 0;
 }
+
+// Check if any key is pressed
+// NOTE: We limit keys check to keys between 32 (KEY_SPACE) and 126
+bool IsAnyKeyPressed()
+{
+    bool keyPressed = false;
+    int key = GetKeyPressed();
+
+    if ((key >= 32) && (key <= 126)) keyPressed = true;
+
+    return keyPressed;
+}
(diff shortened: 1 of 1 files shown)


examples/text/text_input_box.c
```diff
@@ -113,15 +112,3 @@ int main(void)
 
     return 0;
 }
-
```
raysan5, line 116: Please, do not remove this function, it could be useful as a reference for other users.

raysan5:
@Buhlean thanks!

PR #1457, opened 2020-12-12
Add options to set line width and aliasing to rlGL layer.

JeffM2501:
This PR exposes options to get/set the open GL width and aliasing options for line drawing.


raysan5:
@JeffM2501 thanks for the addition. Just note that exposing some of those functionalities have been skipped on purpose to avoid a bloated API with functions that most users hardly ever would require... In any case, those line functions could be useful at some point.

JeffM2501:
Yes, I thought long and hard if these should be in the API or not before submiting. I agree there are many options that should not be in rlgl. The fact that raylib does expose line drawing features is what made me feel that these should be included, because they are the most common options set when using lines.

PR #1458, opened 2020-12-13
Fix obj/pdb conflicts in Visual Studio Projects

JeffM2501:
This PR moves the 'obj' temp dir for the 3 projects that are part of the visual studio solution. Previously all 3 projects were attempting to share the same temp dir in the solution folder. This causes conflicts in visual studio when multiple projects attempt to write and use the same debug files (such as the runtime PDBs). These conflicts cause warnings that are very confusing to users.

This does not change the bin folder, so the end result is the same build as before, there just is not a conflict when visual studio tries to thread the build process.

### DIFF
--- a/projects/VS2019/examples/core_basic_window.vcxproj
+++ b/projects/VS2019/examples/core_basic_window.vcxproj
@@ -127,42 +127,42 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <LinkIncremental>true</LinkIncremental>
     <OutDir>$(SolutionDir)\bin\$(Platform)\$(Configuration)\</OutDir>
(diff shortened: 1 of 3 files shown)


raysan5:
Hi @JeffM2501! Thanks for the review! Maybe the `bin` dir could also be moved to `$(ProjectDir)` instead of using `$(SolutionDir)`, what do you think?

JeffM2501:
Actually I like all the bins being all in the same place, it provides a nice single location for the DLLs and all the samples. It creates a folder where you can run from. 
image

There are no issues with duplicate files in the bin dir because only the unique files are copied there after linking.

You can move the bin dir if you want, but it will make it harder to run the core basic samples in dll builds because the dll won't be there.

raysan5:
@JeffM2501 Ok, merging!

PR #1459, opened 2020-12-14
Expose mesh collision code

JeffM2501:
This PR breaks the current model ray collision code into two parts, one for meshes and one for models. This allows applications to detect collisions when they are drawing meshes directly using rlDrawMesh.
This PR also changes the model ray collision code to call the mesh version for each mesh, so there is no new code to maintain, just exposing what was already in raylib with a little more granularity.


raysan5:
@JeffM2501 Ok, nice! Adding it!

PR #1460, opened 2020-12-14
Building examples by default only if raylib is standalone.

object71:
This change only modifies the default value. In cmake you can have raylib as a subdirectory (dependency) and when you have it as dependency you wouldn't probably want to use the example by default. You can still use the option to enable them though.


object71:
@raysan5 Hi, I added one quality of life change in this pull request. Cmake can usually be used to include dependent projects as git submodules using add_submodule. In this case your main build target won't be raylib itself and the check I add would limit to building only raylib as a default target leaving us with 10-11 compilation files instead of ~210 .

raysan5:
@object71 Nice! Thanks for the improvement!

PR #1461, opened 2020-12-14
Changed extension for web from .bc to .a

object71:
I did this to support vcpkg expectation. When using the library from vcpkg for web you would install it using vcpkg install raylib:wasm32-emscripten but also vcpkg expects the output lib to be with .a extension instead of .bc

Doesn't make a difference for standalone builds or when raylib is used as a subdirectory dependency.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -173,7 +173,7 @@ library_path_to_linker_flags(__PKG_CONFIG_LIBS_PRIVATE "${LIBS_PRIVATE}")
 if(STATIC)
   MESSAGE(STATUS "Building raylib static library")
   if(${PLATFORM} MATCHES "Web")
-    set(CMAKE_STATIC_LIBRARY_SUFFIX ".bc")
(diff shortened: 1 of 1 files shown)


object71:
@raysan5 The extension in this case is purely cosmetical for other use cases except when using vcpkg as a package manager. Vcpkg expects the output library to end in .a format when using the wasm32-emscripten triplet.

raysan5:
@object71 Thanks! Yeah, it changed in latest emscripten versions, I reviewed it for the `Makefiles`...

object71:
@raysan5 No problem! Its a good library and one of the few that make it that easy to develop with. It does compile to web nicely, just cmake needed a few improvements. I will take a look if I can help with issues next now that I have a pipeline working for me.

raysan5:
@object71 Thank you very much! Actually, I'm not an expert on CMake, it was contributed by a user and I can hardly maintain it... Thanks again for the improvements!

PR #1462, opened 2020-12-15
Add headers to install cmake

object71:
@raysan5 Another small improvement. I really want to get this to work for vcpkg and currently vcpkg only has one of these critical headers.


object71:
By mistake I readed the commit for the lib extension - don't mind that it was just the branch from which I branched wasn't master.

PR #1463, opened 2020-12-16
Changing JPG compression

trikko:
Since I don't think the purpose of raylib is to save space compressing image, I think it could be a good idea to keep quality from 90 to 100.


raysan5:
@trikko Nice! :D

PR #1464, opened 2020-12-17
Skip invalid files when iterating a directory.

JeffM2501:
This PR skips any file that starts with '.'. These are ether a relative link or a hidden file on *nix and should not be iterated.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2421,8 +2421,11 @@ char **GetDirectoryFiles(const char *dirPath, int *fileCount)
 
         while ((entity = readdir(dir)) != NULL)
         {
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 I'm afraid this change could break my [custom `GuiFileDialog()` implementation](https://github.com/raysan5/raygui/blob/master/examples/custom_file_dialog/gui_file_dialog.h)...

raysan5:
Just tested and it skips  the `..`, I prefer to let the user manage it...

JeffM2501:
Yes, that was the point, to skip the . and a .. files. There is no valid use of those entries when doing iteration. Every user would need to skip them or you'll end up iterating the entire file system.

PR #1466, opened 2020-12-21
[Math Feature]: Add two functions for Vector2 to know if two lines intersect and two segments intersect

Elkantor:
This PR adds two function in the [raymath.h](https://github.com/raysan5/raylib/blob/master/src/raymath.h) header:

```c
    bool Vector2LineIntersect(Vector2 p1, Vector2 p2, Vector2 p3, Vector2 p4, Vector2* pointIntersection)
```
```c
    bool Vector2SegmentIntersect(Vector2 p1, Vector2 p2, Vector2 p3, Vector2 p4, Vector2* pointIntersection)
```

If the second one is true, the first one is always true.

### Vector2LineIntersect:
The first one check if two lines A and B defined by their points, so that A is defined by (p1, p2) and B is defined by (p3, p4) intersect. If it's true, it put the intersection point in the *Vector2\* pointIntersection* parameter, if it's not NULL.

Which corresponds to this image for example:
<kbd>

</kbd>
</br>

### Vector2SegmentIntersect:
The second one check if two segments A and B defined by their points, so that A is defined by (p1, p2) and B is defined by (p3, p4) intersect. If it's true, it put the intersection point in the *Vector2\* pointIntersection* parameter, if it's not NULL.

Which corresponds to this image for example:
<kbd>

</kbd>
</br>

I choosed to get back a boolean from these two functions, and get the intersection point as a pointer, because we can't just return an empty Vector2 if the segments, or the lines don't intersect. That would mean that we can't ever have an intersection point which has the coordinates { 0, 0 }.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -337,7 +337,7 @@ RMDEF Vector2 Vector2MoveTowards(Vector2 v, Vector2 target, float maxDistance)
 }
 
 // Get the intersection point of two lines A and B defined by A(p1, p2) and B(p3, p4), return true if it exists, else false
-RMDEF _Bool Vector2LineIntersect(Vector2 p1, Vector2 p2, Vector2 p3, Vector2 p4, Vector2* pointIntersection)
+RMDEF bool Vector2LineIntersect(Vector2 p1, Vector2 p2, Vector2 p3, Vector2 p4, Vector2* pointIntersection)
 {
     const float div = (p4.y - p3.y)*(p2.x - p1.x) - (p4.x -p3.x)*(p2.y - p1.y);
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Elkantor Nice! Thanks for the addition! :D

Demizdor:
Doesn't compile, needs bool type declared...something like in the raylib.h header
```
// Boolean type
#if defined(__STDC__) && __STDC_VERSION__ >= 199901L
    #include <stdbool.h>
#elif !defined(__cplusplus) && !defined(bool)
    typedef enum { false, true } bool;
#endif
```

chriscamacho:
it was years ago but I seem to remember that theres a reason you shouldn't use enum for bool but i really can't remember what it was... proceed with caution when doing stuff with the bool type in particular, especially on different platforms....

PR #1467, opened 2020-12-22
[Fix]: Missing <stdbool.h> 

Elkantor:
This PR adds the missing <stdbool.h> include if we are using *raymath.h* as a standalone.


raysan5:
@Elkantor Looking into this implementation with more detail, I think it would be better to move those functions to `shapes` module,  I'm closing this issue.

PR #1468, opened 2020-12-23
Add Quadratic Bezier drawing

epsilon-phase:
Beziers in most graphics environments take 3 points for a finer control of the direction of the curve.


raysan5:
@epsilon-phase Thanks for the addition! Just note I'm reviewing parameters order...

PR #1474, opened 2020-12-26
Edited version of raylib-mruby to comply with 2.5 (Renamed to 2.4-dev)

Rabios:
As i said, There are functions from raylib 2.5, I assume it's better to rename it to 2.4-dev instead of 2.4


raysan5:
@Rabios Nice! Thanks for caring for that detail!

PR #1475, opened 2020-12-26
Again, Renamed version of raylib-mruby to 2.5-dev

Rabios:
Sorry for that, At least i'm correct now ;)
I found similarity with functions of raylib-cr so i guess it's 2.5-dev now!

Thanks ray!


raysan5:
@Rabios Nice! :D

PR #1477, opened 2020-12-26
Update raylib-cpp to state that it targets 3.5

RobLoach:
[raylib-cpp](https://github.com/robloach/raylib-cpp) now targets raylib 3.5.0 as of [raylib-cpp 3.5.0-alpha1](https://github.com/RobLoach/raylib-cpp/releases/tag/v3.5.0-alpha1).


raysan5:
@RobLoach wow! nice! 👍😄

Issue #1322, opened 2020-07-24
[Core] GetMonitorWidth() and GetMonitorHeight() returns wrong value when in fullscreen


fossegutten:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description
GetMonitorWidth() and GetMonitorHeight() returns wrong value when in fullscreen. When in window mode, they return the correct value!
I want to set the screen size to be the same as the actual monitor size, when in fullscreen. To do this I need the monitor size.
### Environment
Surface Go 2, Windows 10, open GL 330. Integrated intel GPU.

### Code Example

```
            ToggleFullscreen();
            if (IsWindowFullscreen()) { 
                // SetWindowSize(GetMonitorHeight(0), GetMonitorHeight(0)); // returns wrong values
                SetWindowSize(1920, 1200); // hardcoded values work fine!
            } 
            else {
                SetWindowSize(640, 360);
            }
```

raysan5:
@fossegutten What are the values returned? I'm afraid it could be the expected behaviour, when switching to fullscreen monitor resolution is adjusted to fullscreen requested resolution (probably the one returned).

fossegutten:
@raysan5 the values seem to change, depending on the previous window size.
I get wrong values when calling the same two functions before the InitWindow() function too.

Not sure if it is relevant, but I have commented out this in the config `//#define SUPPORT_HIGH_DPI            1`

raysan5:
@fossegutten Those functions can't be called before `InitWindow()` because they actually depend on GLFW initialization. Also returned values when toggling fullscreen are constrained to the closer monitor supported fullscreeen resolution, for example asking for a 800x450 fullscreen will probably return a 800x600 fullscreen because it's the closed fullscreen mode supported by the monitor.

Actually, displaySize is already stored internally but fullscreen monitor mode selected also depends on the screen resolution desired. Easy solution: use always a standard screen/monitor resolution, for example 1280x720 or 1920x1080.

fossegutten:
Oh, right. That is probably what happened! Seems like the values returned was constrained to my monitors supported fullscreen resolutions.
Sorry for reporting this, I thought it was a bug. This is complicated stuff to me!

raysan5:
[...]

Issue #1294, opened 2020-06-29
[core] ToggleFullscreen() could change screen width/height


raysan5:
When calling `ToggleFullscreen()`, it could (wrongly) change requested `screenWidth` or `screenHeight`.

Internal call sequence go as follow: 
```c
ToggleFullScreen() 
    glfwSetWindowMonitor()      // This function calculate the closest monitor-supported width x height
        WindowSizeCallback()    // This function is called with a wrong width or height
            SetupViewport()     // Setup viewport and projection matrix 
```

Maybe issue is GLFW side and `glfwSetWindowMonitor()` shouldn't call `WindowSizeCallback()`?

EDIT 1: One possible solution is just commenting `WindowSizeCallback()` internals.

EDIT 2: `InitGraphicsDevice()` also looks for closest monitor supported `width x height` and modifies it.

funcelot:
ok, i need to test https://github.com/raysan5/raylib/pull/1512 with this one, sadly enougth, i can test on windows only

raysan5:
@object71 Tested after latest PR and it still fails, GLFW picks a valid resolution for the monitor and scales to it ignoring requested width/height... :(

object71:
@raysan5 Sorry, it was working when I submitted it but I haven't tested it in some time, I will take another look.

object71:
@raysan5 I've explored the problem some more. What you probably want to achieve is already done in an example "**core_window_letterbox.c**" .

The main points are that:

- When in fullscreen mode you can only use resolution formats that the monitor supports.
- When toggling to fullscreen mode you would get the resolution closest to the window resolution you specified.
- You can't control that behavior

My proposal:

- **Always set to the highest possible resolution for the current monitor**. This is because you get problems if you scale the whole fullscreen to a lower resolution - all other windows get scaled down which is anoying and if you try to switch with alt+tab everything is in the resolution of your game.
- Leave it to library user to implement it with a render target texture if they want to achieve a more pixelated low-res result.

I will prepare a pull request for that and link it and you can take a look and decide.

raysan5:
@object71 Excuse for the late response! You are right with the main points but I don't think setting the higher possible resolution for fullscreen is the right approach, for example, a game that runs at HD (1280x720) will turn into 4K in a monitor supporting it with a consequent frame drop and possible issues with physics and logic (designed for 1280x720 resolution).

[...]

PR #1479, opened 2020-12-27
Fix wrong values shown for monitor width and height in pixels when fu…

object71:
…llscreen is toggled.

This solves issue #1322 which in my opinion was prematurely closed. As the monitor resolution is expected to be the maximum that the monitor supports. My change is that as per GLFW documentation you can get all current video modes sorted by max resolution. When you toggle fullscreen the first video mode returned would be the current screen resolution setup but that doesn't help if you want to know the maximum supported.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1032,7 +1032,7 @@ void ToggleFullscreen(void)
             glfwSetWindowMonitor(CORE.Window.handle, glfwGetPrimaryMonitor(), 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
             return;
         }
-
+    
         const GLFWvidmode *mode = glfwGetVideoMode(monitor);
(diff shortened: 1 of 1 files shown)


object71:
I'm looking into #1294 but this issue was linked and I saw an opportunity to fix it when fullscreen is toggled.

raysan5:
@object71 I need to review this PR more carefully... as per my understanding, display resolution selected should be the closest to the screen resolution required, not the maximum resolution supported by the display...

object71:
@raysan5 ah, ok, I thought it made more sense to be the max res providedvthe function being called GetMonitorWidth/Height and not screen or window width and height.

object71:
I could then add MaxSupportedMonitorWidth/Height instead because there is no good function providing those at the moment? @raysan5

raysan5:
@object71 I need to think about it...

raysan5:
@object71 I'm merging this improvement, now it will return the maximum width/height supported by monitor, `GetScreenWidth()`/`GetScreenHeight()` still return the current framebuffer resolution than, in most cases, will be the current display resolution used.

object71:
@raysan5 Thanks!

PR #1480, opened 2020-12-27
Disabling WindowSizeCallback around fullscreen toggle.

object71:
#1294 fixed the issue that toggle fullscreen changes the screen size because of the WindowSizeCallback. The proposed edit by @raysan5 was to comment out WindowSizeCallback which I essentially simplified to disable the callback around the set window monitor functions.

The developers using the ToggleFullscreen function should be aware that they need to size the screen correctly on that same frame. Otherwise they should check if fullscreen and size properly using the GetMonitorWidth and GetMonitorHeight functions.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1035,8 +1035,7 @@ void ToggleFullscreen(void)
             return;
         }
     
-        const GLFWvidmode *mode = glfwGetVideoMode(monitor)
-
+        const GLFWvidmode *mode = glfwGetVideoMode(monitor);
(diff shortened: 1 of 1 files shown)


object71:
Resolves #1294

funcelot:
@object71 take a look into https://github.com/raysan5/raylib/pull/1512, can you test fullscreen toggle on pull request https://github.com/raysan5/raylib/pull/1512?

raysan5:
I'm merging for further testing...

PR #1481, opened 2020-12-27
providing an example on how to apply the screen scale to the real mouse

duarm:
Providing a comment to the original example, with the code to make the real mouse take the screen scale of the letterbox into consideration.


raysan5:
@duarm Nice! Thanks for the improvement

Issue #1441, opened 2020-11-28
[models] Models assume that they own meshes


JeffM2501:
If you create multiple Models that use the same manually created Mesh, and then call UnloadModel, each call to Unload will free the same Mesh memory and throw an exception.
Currently any call to a GenMesh function that is then used in a LoadModelFromMesh call, must never call UnloadMesh. 
Basically anytime a Mesh is attached to a Model, it assumes complete ownership of the memory.
This behavior prevents a developer from reusing Meshes in multiple Models.

A simple case that causes the problem would be
```C
Mesh myMesh = GenMeshPlane(1, 1, 1, 1);

Model myModel1 = LoadModelFromMesh(myMesh);
Model myModel2 = LoadModelFromMesh(myMesh2);

UnloadModel(myModel1 );

// both of these will assert/throw if called after the first unload
UnloadModel(myModel2 );
UnloadMesh(myMesh)
```
I have tested this on Windows 10, Visual Studio 2019, OpenGL3, GeForce RTX2080.

Currently the only workaround is to create a new Unload function for meshes that does not deallocate the models.
```C
RLAPI void UnloadModelNoMesh(Model model)
{
	// As the user could be sharing shaders and textures between models,
	// we don't unload the material but just free it's maps, the user
	// is responsible for freeing models shaders and textures
	for (int i = 0; i < model.materialCount; i++) RL_FREE(model.materials[i].maps);

	RL_FREE(model.meshes);
	RL_FREE(model.materials);
	RL_FREE(model.meshMaterial);

	// Unload animation data
	RL_FREE(model.bones);
	RL_FREE(model.bindPose);

	TRACELOG(LOG_INFO, "MODEL: Unloaded model(no meshes) from RAM and VRAM");
}
```

This code must be put inside the Raylib code if raylib is used as a DLL so that the memory is freed in the same module it was allocated in.

raysan5:
Actually, now I remember that issue already pop up in the past: https://github.com/raysan5/raylib/issues/1000

I'll review it again.

raysan5:
Still thinking about this but for the moment added the function, it was the easier solution...

Issue #1000, opened 2019-10-27
[models] 2 models using same mesh causes double-free


chriscamacho:
Please, before submitting a new issue verify and check:

 - [x ] I tested it on latest raylib version from master branch
 - [ x] I checked there is no similar issue already reported
 - [x ] My code has no errors or misuse of raylib

### Issue description
        Mesh mesh = GenMeshTorus(.3, 1, 16, 32);
        models[0] = LoadModelFromMesh(mesh);
        models[4] = LoadModelFromMesh(mesh);
        ....
        UnloadModel(models[0]);
        UnloadModel(models[4]); <--- boom!

### Environment

Linux

### Issue Screenshot
N/A

### Code Example
see above

I you don't want to force the end user to free all meshes before unloading models, how about NULLing the mesh pointer when free-ing and only free when its a NULL pointer ?

chriscamacho:
I have a pending PR, I can add this to it if you like ?

raysan5:
@chriscamacho thanks! you can send a new PR!

chriscamacho:
actually this is not a straight forward as I thought, bit difficult to describe what I mean, can you take a look?

raysan5:
@chriscamacho I see, a bit tricky issue, not easy solution... probably it should be left AS-DESIGN.

chriscamacho:
I think its worth spending some time on, as it is its not really ideal.....

chriscamacho:
one thought, might be to have an array of mesh pointers rather than array of mesh....

Ushio:
Hi, How about the option?

```
typedef struct Model {
    Matrix transform;       // Local transform matrix

    int meshCount;          // Number of meshes
    Mesh *meshes;           // Meshes array
    bool automaticUnloadMesh; // <- New! If the option is ture, UnloadMesh() will called for the meshes

    int materialCount;      // Number of materials
    Material *materials;    // Materials array
    int *meshMaterial;      // Mesh material number

    // Animation data
    int boneCount;          // Number of bones
    BoneInfo *bones;        // Bones information (skeleton)
    Transform *bindPose;    // Bones base transformation (pose)
} Model;

```

And the following code will work.

```

    Mesh mesh = GenMeshTorus(.3, 1, 16, 32);
    models[0] = LoadModelFromMesh(mesh);
    models[4] = LoadModelFromMesh(mesh);
    models[0].automaticUnloadMesh = false;
    models[4].automaticUnloadMesh = false;
    ....
[...]

PR #1482, opened 2020-12-28
Protect against double free() on Meshes?

RobLoach:
Currently, the `UnloadModel()` function will also unload the associated meshes. There is a `UnloadModelKeepMeshes()` function to allow unloading the model without unloading the meshes, but it seems somewhat counter intuitive.

[models_first_person_maze.c](https://github.com/raysan5/raylib/blob/master/examples/models/models_first_person_maze.c) is a good example of what we can run into. In it, we see.....

``` c
// Generate a Cubicmap mesh
Mesh mesh = GenMeshCubicmap(imMap, (Vector3){ 1.0f, 1.0f, 1.0f });
Model model = LoadModelFromMesh(mesh);
```

In normal C logic, one would assume that when you generate an object, you should Unload it. So, at the end of the file, you would assume you would put...

``` c
// De-Initialization
UnloadModel(model);
UnloadMesh(mesh);
```

But this is not the case, because `UnloadModel()` will unload both the Model and the Mesh, so calling `UnloadMesh(mesh)` results in a double free error. There is a `UnloadModelKeepMeshes()` function, but I believe there should be a bit of protection against double free-ing.

**DO NOT MERGE THIS**. This is just a demonstration of how to reproduce the issue through the [models_first_person_maze.c](https://github.com/raysan5/raylib/blob/master/examples/models/models_first_person_maze.c) example. When run, this is what you get when closing the application.

``` log
INFO: FILEIO: [resources/cubicmap.png] File loaded successfully
INFO: IMAGE: [resources/cubicmap.png] Data loaded successfully (32x16)
INFO: TEXTURE: [ID 3] Texture created successfully (32x16 - 1 mipmaps)
INFO: VAO: [ID 2] Mesh uploaded successfully to VRAM (GPU)
INFO: FILEIO: [resources/cubicmap_atlas.png] File loaded successfully
INFO: IMAGE: [resources/cubicmap_atlas.png] Data loaded successfully (256x256)
INFO: TEXTURE: [ID 4] Texture created successfully (256x256 - 1 mipmaps)
INFO: TEXTURE: [ID 3] Unloaded texture data from VRAM (GPU)
INFO: TEXTURE: [ID 4] Unloaded texture data from VRAM (GPU)
INFO: VAO: [ID 2] Unloaded vertex data from VRAM (GPU)
double free or corruption (!prev)
[1]    59167 abort (core dumped)  ./models_first_person_maze
```

Is adding some error correction in the Unload logic a good step forwards to protect against the double-free issue? I assume this also applies to Materials?

If not, feel free to close this issue :wink:

### DIFF
--- a/examples/models/models_first_person_maze.c
+++ b/examples/models/models_first_person_maze.c
@@ -117,6 +117,7 @@ int main(void)
 
     UnloadTexture(cubicmap);    // Unload cubicmap texture
     UnloadTexture(texture);     // Unload map texture
+    UnloadMesh(mesh);           // Unload mesh
     UnloadModel(model);         // Unload map model
 
     CloseWindow();              // Close window and OpenGL context


raysan5:
@RobLoach Yeah, you are right, I agree. I've been working on this issue for some time (#1000, #1441). I haven't found a proper solution yet, `UnloadModelKeepMeshes()` is actually a patch (to be removed in the future).

A possible solution could be letting `LoadModelFromMesh(mesh);` duplicate the mesh internally, to keep its own copy.

RobLoach:
Thanks! Glad it's on your radar and that you've been thinking about it. Should certainly be an issue instead, but thought I'd give a demonstration. Going to close for now. :+1:

CaiusTSM:
Hello, this issue was closed. But what was the final verdict? I'm in favor of UnloadModel not unloading meshes. And something like UnloadModelAll for unloading the model and everything it uses. I think the API should remain consistent, where if you load something, you need to unload it, manually. Or use a more advanced unload function that does more work.

raysan5:
@CaiusTSM `UnloadModelKeepMeshes()` was added to the API. Personally I'm not happy with that oprion, it looks like a patch to me but it was probably the easier solution.

orcmid:
> @CaiusTSM `UnloadModelKeepMeshes()` was added to the API. Personally I'm not happy with that option, it looks like a patch to me but it was probably the easier solution.

@CaiusTSM > Or use a more advanced unload function that does more work.

Smiling as we step closer to the reference-counting chasm.  As valuable (and challenging) that is, it strikes me that raylib avoids user-visible reference-counting by design.  

  It seems 4.0 has important fixes that should not be held up.  The `UnloadModelKeepMeshes()`  hack needn't be broken when there is a better solution later on. Then maybe there is a hidden approach so use/free after free could fail softly.  Does that make any sense?

PR #1485, opened 2020-12-29
Monitor utilites

### LINKED ISSUE #1484: [raylib.h] missing GetWindowMonitor(void) (new feature)
Hey guys,

as I am working on my first RayLib project, I was looking for a way to know which monitor is currently used by the window, and could not find a function in the library that does that.

I tried to include the winuser.h so I can use MonitorFromWindow and that is not possible,.

I found this troubling not to know what monitor the window is currently on as all of the RayLib GetMonitor functionalities all refer to a monitor but we can't even tell which monitor the game was launched at...

Please consider adding this functionality to help figure out which monitor the user is running the game.  Someone might say; it is launched from monitor 1, always.  How do you know if you do not check ?  Specially if it is in windowed mode.

object71:
Resolves #1484

I added this function regarding this issue - We have a bunch of functinos that are for querying monitor information and changing monitors but there is no way currently to know which is the current used monitor. @raysan5

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1452,6 +1452,16 @@ int GetMonitorCount(void)
 #endif
 }
 
+// Get number of monitors
+int GetCurrentMonitor(void)
+{
+#if defined(PLATFORM_DESKTOP)
+    return glfwGetWindowMonitor(CORE.Window.handle);
+#else
+    return 0;
+#endif
+}
+
 // Get selected monitor width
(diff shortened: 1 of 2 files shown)


raysan5:
@object71 Good addition! I'm still thinking about naming... `GetMonitorCurrent()` seems more aligned to the other similar functions... but `GetCurrentMonitor()` is more clear for the user...

PR #1486, opened 2020-12-29
I noticed that toggle fullscreen always gets the primary monitor when…

object71:
… it is more intuitive to use the monitor that the window is currently in.

This monitor should even probably be passed as an id instead to make it more consistent with the rest of the API but this would break existing applications so I am hesitant to change the function signature directly.

@raysan5 What are your thoughts on this one?


raysan5:
@object71 Yeah, absolutely! Good catch!

PR #1487, opened 2020-12-29
Faster version of ImageClearBackground and ImageDrawRectangleRec

JeffM2501:
The use of ImageDrawRectangle to clear an image is very slow. It makes a new image, draws it, and then destroys the image. This is not fast.

This PR changes ImageClearBackground and ImageDrawRectangleRec to fill in the raw image data one pixel at a time (still doing data conversions per pixel), and is an order of magnitude faster.


raysan5:
@JeffM2501 Oh! Nice! Thanks for the review!

PR #1488, opened 2020-12-29
Fix keyboard state change detection on RPI

badlydrawnrod:
## Overview
On Raspberry Pi DRM builds, keyboard state changes are rarely detected (if at all) by `IsKeyPressed(`) / `IsKeyReleased()`.

## Problem
When a key is pressed or released, the input worker thread in `EventThread()` reads the event, maps the scan code to a keycode, then registers the keycode in `CORE.Input.Keyboard.currentKeyState[]`, setting it to 1 for pressed and 0 for released.

When the user’s application calls `EndDrawing()`, that in turn calls `PollInputEvents()`. One of the first things that `PollInputEvents()` does is copy `CORE.Input.Keyboard.currentKeyState[]` to `CORE.Input.Keyboard.previousKeyState[]`.

The input functions `IsKeyPressed()` and `IsKeyReleased()` are supposed to detect key state changes from frame to frame. They do this by comparing `currentKeyState[key]` with `previousKeyState[key]`.

What seems to be happening is that there is a data race in which `PollInputEvents()` can run and update `previousKeyState` with `currentKeyState` before `IsKeyPressed()` / `IsKeyReleased()` has had a chance to see the state change, resulting in nothing being detected and key events being missed. This happens more often than not.

### Code to demonstrate issue (compile for PLATFORM_DRM)
Run the following in a local (non-windowed) tty on a Raspberry Pi. What should happen is that both pressing [space] and pressing [enter] will advance the message. What actually happens is that pressing [space] very rarely advances the message.

```c
#include "raylib.h"

#if defined(PLATFORM_WEB) || defined(EMSCRIPTEN)
#endif

#define UPDATE_FPS 60
#define SCREEN_WIDTH 1280
#define SCREEN_HEIGHT 960

static const char* instruction = "Press [space] or [enter] to advance the message";
static const char* messages[] = {"raylib", "is", "great"};
static int index = 0;
static int numMessages = sizeof(messages) / sizeof(messages[0]);
static bool wasEnterDown = false;

static void Draw(void)
{
    ClearBackground(BLACK);
    BeginDrawing();

    int width = MeasureText(messages[index], 64);
    DrawText(messages[index], (GetScreenWidth() - width) / 2, GetScreenHeight() / 2, 64, GREEN);

    width = MeasureText(instruction, 32);
    DrawText(instruction, (GetScreenWidth() - width) / 2, GetScreenHeight() / 4, 32, RAYWHITE);

    DrawFPS(4, 4);
    EndDrawing();
}

static void UpdateDrawFrame(void)
{
    // Check for [space] changing state to pressed with IsKeyPressed().
    if (IsKeyPressed(KEY_SPACE))
    {
        index = (index + 1) % numMessages;
    }

    // Check for [enter] changing state to pressed by maintaining our own state.
    bool isEnterDown = IsKeyDown(KEY_ENTER);
    if (isEnterDown && !wasEnterDown)
    {
        index = (index + 1) % numMessages;
    }
    wasEnterDown = isEnterDown;

    Draw();
}

int main(void)
{
    InitWindow(SCREEN_WIDTH, SCREEN_HEIGHT, "Empty");
    SetTargetFPS(UPDATE_FPS);

#if defined(PLATFORM_WEB) || defined(EMSCRIPTEN)
    emscripten_set_main_loop(UpdateDrawFrame, 0, 1);
#else
    while (!WindowShouldClose())
    {
        UpdateDrawFrame();
    }
#endif
    CloseWindow();

    return 0;
}
```

## Solution
Make the keyboard state _changes_ persist from frame to frame. This is done by adding another array `CORE.Input.Keyboard.changedKeyState[]`, which is updated in `PollInputEvents()` by XORing the previous state with the current state. Checking `IsKeyPressed(`) / `IsKeyReleased()` then become a matter of checking that `changedKeyState` is non-zero, then looking at `previousKeyState` to see whether it was pressed or released.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -410,7 +410,9 @@ typedef struct CoreData {
             int exitKey;                    // Default exit key
             char currentKeyState[512];      // Registers current frame key state
             char previousKeyState[512];     // Registers previous frame key state
-
+#if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
+            char changedKeyState[512];      // Registers keys whose state changed between frames
+#endif
             int keyPressedQueue[MAX_KEY_PRESSED_QUEUE];     // Input keys queue
             int keyPressedQueueCount;       // Input keys queue count
 
@@ -2681,12 +2683,16 @@ void OpenURL(const char *url)
 // Detect if a key has been pressed once
 bool IsKeyPressed(int key)
 {
+#if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
+    return CORE.Input.Keyboard.changedKeyState[key] != 0 && CORE.Input.Keyboard.previousKeyState[key] == 1;
+#else
     bool pressed = false;
 
     if ((CORE.Input.Keyboard.previousKeyState[key] == 0) && (CORE.Input.Keyboard.currentKeyState[key] == 1)) pressed = true;
     else pressed = false;
 
     return pressed;
+#endif
 }
 
 // Detect if a key is being pressed (key held down)
@@ -2699,12 +2705,16 @@ bool IsKeyDown(int key)
 // Detect if a key has been released once
 bool IsKeyReleased(int key)
 {
+#if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
+    return CORE.Input.Keyboard.changedKeyState[key] != 0 && CORE.Input.Keyboard.previousKeyState[key] == 0;
+#else
     bool released = false;
 
     if ((CORE.Input.Keyboard.previousKeyState[key] == 1) && (CORE.Input.Keyboard.currentKeyState[key] == 0)) released = true;
     else released = false;
 
     return released;
+#endif
 }
 
 // Detect if a key is NOT being pressed (key not held down)
@@ -4224,8 +4234,12 @@ static void PollInputEvents(void)
 #endif
 
 #if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
-    // Register previous keys states
-    for (int i = 0; i < 512; i++) CORE.Input.Keyboard.previousKeyState[i] = CORE.Input.Keyboard.currentKeyState[i];
+    // Register previous keys states and detect changes.
+    for (int i = 0; i < 512; i++)
+    {
+        CORE.Input.Keyboard.changedKeyState[i] = CORE.Input.Keyboard.previousKeyState[i] ^ CORE.Input.Keyboard.currentKeyState[i];
+        CORE.Input.Keyboard.previousKeyState[i] = CORE.Input.Keyboard.currentKeyState[i];
+    }
 
     // Grab a keypress from the evdev fifo if avalable
     if (CORE.Input.Keyboard.lastKeyPressed.head != CORE.Input.Keyboard.lastKeyPressed.tail)
--- a/src/core.c
+++ b/src/core.c
@@ -414,7 +414,7 @@ typedef struct CoreData {
 #if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
             int defaultMode;                // Default keyboard mode
             struct termios defaultSettings; // Default keyboard settings
-			int fd;            				// File descriptor for the evdev keyboard
+            int fd;                         // File descriptor for the evdev keyboard
 #endif
         } Keyboard;
         struct {
--- a/src/core.c
+++ b/src/core.c
@@ -333,12 +333,6 @@ typedef struct {
     bool isKeyboard;                // True if device has letter keycodes
     bool isGamepad;                 // True if device has gamepad buttons
 } InputEventWorker;
-
-typedef struct {
-    int contents[8];                // Key events FIFO contents (8 positions)
-    char head;                      // Key events FIFO head position
-    char tail;                      // Key events FIFO tail position
-} KeyEventFifo;
 #endif
 
 typedef struct { int x; int y; } Point;
@@ -410,9 +404,7 @@ typedef struct CoreData {
             int exitKey;                    // Default exit key
             char currentKeyState[512];      // Registers current frame key state
             char previousKeyState[512];     // Registers previous frame key state
-#if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
-            char changedKeyState[512];      // Registers keys whose state changed between frames
-#endif
+
             int keyPressedQueue[MAX_KEY_PRESSED_QUEUE];     // Input keys queue
             int keyPressedQueueCount;       // Input keys queue count
 
@@ -422,7 +414,7 @@ typedef struct CoreData {
 #if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
             int defaultMode;                // Default keyboard mode
             struct termios defaultSettings; // Default keyboard settings
-            KeyEventFifo lastKeyPressed;    // Buffer for holding keydown events as they arrive (Needed due to multitreading of event workers)
+			int fd;            				// File descriptor for the evdev keyboard
 #endif
         } Keyboard;
         struct {
@@ -561,7 +553,8 @@ static void RestoreTerminal(void);                      // Restore terminal
 #endif
 
 static void InitEvdevInput(void);                       // Evdev inputs initialization
-static void EventThreadSpawn(char *device);             // Identifies a input device and spawns a thread to handle it if needed
+static void ConfigureEvdevDevice(char *device);         // Identifies a input device and configures it for use if appropriate
+static void PollKeyboardEvents(void);                   // Process evdev keyboard events.
 static void *EventThread(void *arg);                    // Input device events reading thread
 
 static void InitGamepad(void);                          // Init raw gamepad input
@@ -901,6 +894,13 @@ void CloseWindow(void)
 
     CORE.Window.shouldClose = true;   // Added to force threads to exit when the close window is called
 
+    // Close the evdev keyboard
+    if (CORE.Input.Keyboard.fd != -1)
+    {
+        close(CORE.Input.Keyboard.fd);
+        CORE.Input.Keyboard.fd = -1;
+    }
+    
     for (int i = 0; i < sizeof(CORE.Input.eventWorker)/sizeof(InputEventWorker); ++i)
     {
         if (CORE.Input.eventWorker[i].threadId)
@@ -908,6 +908,7 @@ void CloseWindow(void)
             pthread_join(CORE.Input.eventWorker[i].threadId, NULL);
         }
     }
+    
 
     if (CORE.Input.Gamepad.threadId) pthread_join(CORE.Input.Gamepad.threadId, NULL);
 #endif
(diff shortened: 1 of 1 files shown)


raysan5:
Mmmh... is not possible to avoid that extra array? I think registering previousState and currentState should be enough...

I need to check it more carefully...

badlydrawnrod:
Thanks for the review. I will investigate and see if I can come up with a better solution.

jmorel33:
I have to agree with badlydrawnrod that there is a serious issue in the usage of detecting key presses.  I noticed that issue when trying a game demo; PAC MAN, which was frustrating and silly.

jmorel33:
I've setup a little program to scan the key presses, if that could be useful for your tests.
Beware, it uses the new GetCurrentMonitor()


`#include "raylib.h"

int kb_layout[168];

static void set_layout(void)
{
kb_layout[0] = KEY_ESCAPE;
kb_layout[3] = KEY_F1;
kb_layout[4] = KEY_F2;
kb_layout[5] = KEY_F3;
kb_layout[6] = KEY_F4;
kb_layout[7] = KEY_F5;
kb_layout[8] = KEY_F6;
kb_layout[9] = KEY_F7;
kb_layout[10] = KEY_F8;
kb_layout[11] = KEY_F9;
kb_layout[12] = KEY_F10;
kb_layout[13] = KEY_F11;
kb_layout[14] = KEY_F12;
kb_layout[16] = KEY_PRINT_SCREEN;
kb_layout[17] = KEY_SCROLL_LOCK;
kb_layout[18] = KEY_PAUSE;

kb_layout[48] = KEY_GRAVE;
kb_layout[49] = KEY_ONE;
kb_layout[50] = KEY_TWO;
kb_layout[51] = KEY_THREE;
kb_layout[52] = KEY_FOUR;
kb_layout[53] = KEY_FIVE;
kb_layout[54] = KEY_SIX;
kb_layout[55] = KEY_SEVEN;
kb_layout[56] = KEY_EIGHT;
kb_layout[57] = KEY_NINE;
kb_layout[58] = KEY_ZERO;
kb_layout[59] = KEY_MINUS;
kb_layout[60] = KEY_EQUAL;
kb_layout[61] = KEY_BACKSPACE;
kb_layout[62] = KEY_BACKSPACE;
kb_layout[64] = KEY_INSERT;
kb_layout[65] = KEY_HOME;
kb_layout[66] = KEY_PAGE_UP;
kb_layout[68] = KEY_NUM_LOCK;
kb_layout[69] = KEY_KP_DIVIDE;
kb_layout[70] = KEY_KP_MULTIPLY;
kb_layout[71] = KEY_KP_SUBTRACT;

kb_layout[72] = KEY_TAB;
kb_layout[73] = KEY_TAB;
kb_layout[74] = KEY_Q;
kb_layout[75] = KEY_W;
kb_layout[76] = KEY_E;
kb_layout[77] = KEY_R;
kb_layout[78] = KEY_T;
kb_layout[79] = KEY_Y;
kb_layout[80] = KEY_U;
kb_layout[81] = KEY_I;
kb_layout[82] = KEY_O;
kb_layout[83] = KEY_P;
kb_layout[84] = KEY_LEFT_BRACKET;
kb_layout[85] = KEY_RIGHT_BRACKET;
kb_layout[86] = KEY_BACKSLASH;
kb_layout[88] = KEY_DELETE;
kb_layout[89] = KEY_END;
kb_layout[90] = KEY_PAGE_DOWN;
kb_layout[92] = KEY_KP_7;
kb_layout[93] = KEY_KP_8;
kb_layout[94] = KEY_KP_9;
kb_layout[95] = KEY_KP_ADD;

kb_layout[96] = KEY_CAPS_LOCK;
kb_layout[97] = KEY_CAPS_LOCK;
kb_layout[98] = KEY_A;
kb_layout[99] = KEY_S;
kb_layout[100] = KEY_D;
kb_layout[101] = KEY_F;
kb_layout[102] = KEY_G;
kb_layout[103] = KEY_H;
kb_layout[104] = KEY_J;
kb_layout[105] = KEY_K;
kb_layout[106] = KEY_L;
kb_layout[107] = KEY_SEMICOLON;
kb_layout[108] = KEY_APOSTROPHE;
kb_layout[109] = KEY_ENTER;
kb_layout[110] = KEY_ENTER;


kb_layout[116] = KEY_KP_4;
kb_layout[117] = KEY_KP_5;
kb_layout[118] = KEY_KP_6;
kb_layout[119] = KEY_KP_ADD;

kb_layout[120] = KEY_LEFT_SHIFT;
kb_layout[121] = KEY_LEFT_SHIFT;
kb_layout[122] = KEY_Z;
kb_layout[123] = KEY_X;
kb_layout[124] = KEY_C;
kb_layout[125] = KEY_V;
kb_layout[126] = KEY_B;
kb_layout[127] = KEY_N;
kb_layout[128] = KEY_M;
kb_layout[129] = KEY_COMMA;
kb_layout[130] = KEY_PERIOD;
kb_layout[131] = KEY_SLASH;
kb_layout[132] = KEY_RIGHT_SHIFT;
kb_layout[133] = KEY_RIGHT_SHIFT;
kb_layout[134] = KEY_RIGHT_SHIFT;

kb_layout[137] = KEY_UP;
kb_layout[140] = KEY_KP_1;
kb_layout[141] = KEY_KP_2;
kb_layout[142] = KEY_KP_3;
kb_layout[143] = KEY_KP_ENTER;

kb_layout[144] = KEY_LEFT_CONTROL;
kb_layout[145] = KEY_LEFT_SUPER;
kb_layout[146] = KEY_LEFT_ALT;
kb_layout[147] = KEY_SPACE;
kb_layout[148] = KEY_SPACE;
kb_layout[149] = KEY_SPACE;
kb_layout[150] = KEY_SPACE;
kb_layout[151] = KEY_SPACE;
kb_layout[152] = KEY_SPACE;
kb_layout[153] = KEY_SPACE;
kb_layout[154] = KEY_SPACE;
kb_layout[155] = KEY_RIGHT_ALT;
kb_layout[157] = KEY_RIGHT_SUPER;
kb_layout[157] = KEY_KB_MENU;
kb_layout[158] = KEY_RIGHT_CONTROL;
kb_layout[160] = KEY_LEFT;
kb_layout[161] = KEY_DOWN;
kb_layout[162] = KEY_RIGHT;
kb_layout[164] = KEY_KP_0;
kb_layout[165] = KEY_KP_0;
kb_layout[166] = KEY_KP_DECIMAL;
kb_layout[167] = KEY_KP_ENTER;

}

static void keyboard_test(void)
{
    ClearBackground(BLACK);
    BeginDrawing();

    int row = 24;
    for (int j = 0; j < 7; j++) {

    for (int i = 0; i < row; i++) {
        int letter = (j*row)+i;
        int key = kb_layout[letter];

        if (key) {
        if (IsKeyDown(key)) {
            DrawRectangle(i*64, 20+j*64, 64, 64, GREEN);
            DrawText(TextFormat("%c", (int)key), i*64, 20+j*64, 64, WHITE);
        }
        else {
            DrawRectangle(i*64, 20+j*64, 64, 64, DARKGRAY);
            DrawText(TextFormat("%c", (int)key), i*64, 20+j*64, 64, GRAY);
        };
        };
        
    };
    };

    DrawFPS(0, 700);
    EndDrawing();
}

int main(void)
{
    set_layout();
    int screenWidth = 1024;
    int screenHeight = 480;

    InitWindow(screenWidth, screenHeight, "I am testing my keyboard");
    int current_monitor = GetCurrentMonitor();
    screenWidth = GetMonitorWidth(current_monitor);
    screenHeight = GetMonitorHeight(current_monitor);
    SetWindowState(FLAG_FULLSCREEN_MODE);
    SetWindowSize(screenWidth,screenHeight);
    SetTargetFPS(GetMonitorRefreshRate(current_monitor));

    while (!WindowShouldClose())
    {
        keyboard_test();
    }
    CloseWindow();

    return 0;
}`

badlydrawnrod:
@jmorel33 thanks. I may adapt that to handle key presses and releases as it's a nice way of visualising the keyboard.

## Possible Approaches
@raysan5 After some research into the [/dev/input/eventX API](https://www.kernel.org/doc/html/latest/input/input.html#event-interface), here are some possible approaches:

1. Check the input in the main thread, rather than in multiple input event worker threads.
2. Check the input in the main thread, using `select()` to determine which file descriptors have data.
3. Check only the keyboard input in the main thread, leaving the remaining input code as it is.
4. Check the input in a single, input event worker thread, using `select()` to determine which devices have data, and adding a mechanism to synchronize with the main thread.

I did a spike to try out option 1, and it ran successfully with the core_input_keys, core_input_mouse and core_input_mouse_wheel examples. It also ran successfully with the code that I used previously to demonstrate the issue (see above). Option 1 looks viable as a first pass, although I suspect option 2 will end up being cleaner.

However, given that the current implementation uses multiple threads, then there is probably a good reason why input needs to be on a separate thread. Thinking about it, if checking input is tied to frame rate, then frequent events such as mouse events or touch events could be dropped due to buffer overrun (there's even a `SYN_DROPPED` for this scenario). The probability of dropped events increases as the frame rate drops.

At this point, option 3 is a reasonable compromise, as keyboard events are relatively infrequent (people don't type that quickly). In the longer term, I suspect that option 4 will end up being cleaner.

I'm going to err on the side of caution and explore option 3 first, with a view to implementing option 4 at a later date.

raysan5:
@badlydrawnrod First of all, thank you very much for such detailed and professional description. 

About the possible solution, actually, **number 3** was the original implementation for Raspberry Pi, I agree that reading keyboard synchronously 30-60 times per second should be more than enough. Mouse/Touch position requires a separate thread because it can really move faster than that. As I said that was the original implementation. It was reworked by a contributor and moved to separate threads, it looked ok to me at that moment... but it's probably time to improve it again. I also think it's the easier solution from a maintenance point of view..

Again, thank you very much for the detailed review of this issue.

badlydrawnrod:
You're welcome. I'll see what I can do with option 3 and update this PR when I have something reviewable.

badlydrawnrod:
Just committed a reworked version of the code, so that keyboard handling is synchronous rather than in a separate thread.

Summary:

- Extracted evdev keyboard handling into `PollKeyboardEvents()`
- Moved keyboard polling to main thread
- Renamed `EventThreadSpawn()` to `ConfigureEvdevDevice()` as its name became misleading (it doesn't necessarily spawn threads)
- Removed unused code (`KeyEventFifo` and `lastKeyPressed`)

src/core.c
```diff
@@ -420,7 +414,7 @@ typedef struct CoreData {
 #if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
             int defaultMode;                // Default keyboard mode
             struct termios defaultSettings; // Default keyboard settings
-            KeyEventFifo lastKeyPressed;    // Buffer for holding keydown events as they arrive (Needed due to multitreading of event workers)
+			int fd;            				// File descriptor for the evdev keyboard
```
raysan5, line 417: Please, could you review TABS by 4 spaces?

raysan5:
@badlydrawnrod Thank you very much for this re-work! :)

src/core.c
badlydrawnrod: Oops - thought I'd got those. Fixing now.

raysan5:
@badlydrawnrod thanks for the improvement on keyboard!

PR #1490, opened 2020-12-29
fix GetCurrentMonitor to correctly lookup the monitor index

JeffM2501:
GetCurrentMonitor calls a function that returns the current monitor handle, not the index. This PR looks up that monitor in the monitor list to get the index so it can be used by other functions.


raysan5:
@JeffM2501 thanks for the quick fix! 👍😄

PR #1492, opened 2020-12-29
Add a current monitor function for window mode windows.

JeffM2501:
This PR adds logic to compute the current monitor for non-full screen windows. The monitor that contains the upper left origin is returned.


jmorel33 (commented):
I tested this and it works fully

raysan5:
@JeffM2501 Great! Thanks for the review!

PR #1493, opened 2020-12-30
added missing constants

funcelot:


funcelot:
added missing definitions for cmake build in vscode

raysan5:
@funcelot Thanks for the review!

PR #1497, opened 2020-12-31
[Cleanup]: Modify check condition less than zero for unsigned integers

Elkantor:
This PR:
- Modify the check condition for the screen window height and width in this function for the *PLATFORM_DESKTOP* macro:
```c
static bool InitGraphicsDevice(int width, int height)
```
These values can't be less than zero because they are unsigned int. So we only need to check if they are equal to 0, not less than 0.


raysan5:

PR #1498, opened 2020-12-31
[Fix]: Avoid dereferencing a null pointer in LoadMaterials function

Elkantor:
This PR:
- Fix a dereferencing of a null pointer in the *LoadMaterials* function inside the *models* module.
(the **materials** pointer array is initialized to NULL, then used in a for loop to access it at *i* index).


raysan5:

PR #1499, opened 2020-12-31
[Fix]: Avoid dereferencing a null pointer in the LoadSoundFromWave function

Elkantor:
This PR:
- Avoid dereferencing the *audioBuffer* pointer if it's NULL (which will crash the app), by returning earlier in the **LoadSoundFromWave** function.


raysan5:

PR #1500, opened 2020-12-31
Update Gforth and Factor bindings

ArnautDaniel:


raysan5:
@ArnautDaniel Very nice! Thanks! 👍😄

PR #1504, opened 2021-01-02
Don't create an ortho matrix when the viewport is 0 in any axis.

JeffM2501:
Not all compilers divide by 0 and return inf, some segfault.
This PR skips the ortho matrix in the resize callback when any axis is 0, to prevent any divide by 0 errors that may come up.
The matrix is not used by anything when minimized, so it just needs to not be called.
The correct matrix will be created when the window is restored.


raysan5:
@JeffM2501 Hey! I think `core_basic_window` shouldn't be in this PR...

JeffM2501:
> @JeffM2501 Hey! I think `core_basic_window` shouldn't be in this PR...

That is correct, that is my test file :( I'll see if I can figure out how to remove that

JeffM2501:
Ok, it should be good now, the example file is not part of the changed files in the PR. I'm very sorry about that.

raysan5:
@JeffM2501 Thanks for the improvement! :)

Issue #1319, opened 2020-07-22
[input] Gamepad buttons and axis not working (linux)


muevoid:
### Issue description

I have tried in both my own project and the gamepad example both in browser and
on my machine compiled. However it just shows up with gamepad not detected.
I have tested with jstest to make sure the joystick is showing up on a system
level and it appears to be fine. (To note i have tried this with both a Xbox
one controller and a steam controller)

### Environment

I am on KISS Linux latest stable kernel.

### Code Example

Just the code in the gamepad_test_example that checks if a gamepad is
available. AKA (IsGamePadAvailable(GAMEPAD_PLAYER1))

raysan5:
@muevoid is the gamepad connected before running the example? It should be at least detected... it could be a KISS Linux specific issue...

muevoid:
Yes i can confirm the gamepad is connected before hand. I'm not sure exactly what on KISS could cause this issue.

raysan5:
@muevoid raylib uses GLFW internally on DESKTOP_PLATFORM (Windows, Linux, macOS) to manage inputs. Every frame [core] `PollInputEvents()` is called that also scans possible gamepads with `glfwJoystickPresent()`.

`IsGamepadAvailable(int gamepad)` just returns the result of that detection (`true` or `false`), you can try with different `int gamepad` values just in case but if the function always returns `false` then the issue is probably on GLFW side and not in raylib.

muevoid:
Kk I am familiar with glfw and will test this tommorow.

raysan5:
Just a reference in case it can be useful in the future for gamepad buttons mapping: https://www.glfw.org/docs/3.3/input_guide.html#gamepad

defutura:
An additional perspective > I just tried this on Ubuntu 20.04 LTS x86_64, kernel: 5.4.0-42-generic. My build of 'core_input_gamepad.c' detects both a cheap Logitech controller and an 8bitdo SNES controller. They do not need to be plugged in before compilation; I can disconnect and reconnect the controllers, and the program will detect them and display their names correctly. However, I can't seem to be able to get any input from any of them, and neither buttons nor axes show up in the interface.

muevoid:
Hmmm I am still doing testing will post here soon.

muevoid:
This is for sure an issue with my configuration and not raylib. So I will close this jist of it was my Xorg configuration and some stuff with evdev.

muevoid:
[...]

PR #1505, opened 2021-01-02
Update mappings.h in glfw using GenerateMappings.cmake

ChrisDill:
I ran into the same issue as #1319 with a ps5 dual sense controller. I used GenerateMappings.cmake to update mappings.h and rebuilt raylib and the examples. I then tested it using core_input_gamepad and the controller now responds as expected.


raysan5:
@ChrisDill Very nice! Thank you very much for the update! 👍😄

PR #1507, opened 2021-01-03
Removed a repeating allocation of memory

### LINKED ISSUE #1495: [models] Double allocation of "vboId" in GenMeshPlane
It seems there is a possible double allocation of the vboId-buffer from the `Mesh` type within the function `GenMeshPlane` as it is first allocated at line 1383:
`1383:    mesh.vboId = (unsigned int *)RL_CALLOC(DEFAULT_MESH_VERTEX_BUFFERS, sizeof(unsigned int));`
before the preprocessor #IF and then again within the preprocessor #ELSE statement in line 1486:
`1486:    mesh.vboId = (unsigned int *)RL_CALLOC(DEFAULT_MESH_VERTEX_BUFFERS, sizeof(unsigned int));`
without freeing the already allocated buffer before.

object71:
Resolves #1495
This line allocated some memory that was already allocated in the beginning of the function and was essentially creating a leak.


raysan5:
@object71 Thanks! Good catch! :)

PR #1512, opened 2021-01-04
Added support for high DPI monitors (windows 10 / multi monitor configuration)

funcelot:
Added support for hi DPI monitors:

Used core_windows_flags app as the only one with internal high DPI flag enabled by default (Windows 10 latest)

Scenario: 

1) add support for moving app window from one hi DPI monitor to another one with different DPI (tested on windows only). need to result in no glitchess, no artefact, correct rescale of app window. i.e. primary monitor (windows 10 latest update) with scale factor 200% will show 800x450 as 1600x900, then, by moving to another monitor, with scale factor 150%, will result in application rescale to 1200x675, with no code change or support from application - Done

2) add support for rescale and app on current window, if scale factor changed by Windows 10, i.e. primary monitor (windows 10 latest update) with scale factor 200% will show 800x450 as 1600x900, then, by changing scale factor on primary monitor, to scale factor 150%, will result in application rescale to 1200x675, with no code change or support fron application - Done

3) add support of correct resizable windowed app resizes with different flags set, i.e. resizing window by dragging size grip will result in correct rescale of high DPI app, then applying scenario 1) or 2) will not change behavior or correctness of scenario 1) or 2) - Done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Ignore generated files
 # ...
 
+.vscode
+
 # Ignore VIM's backup generated files
 *.swp
 *.swo
--- a/examples/models/models_material_pbr.c
+++ b/examples/models/models_material_pbr.c
@@ -24,6 +24,8 @@
 #define IRRADIANCE_SIZE       32        // Irradiance texture size
 #define PREFILTERED_SIZE     256        // Prefiltered HDR environment texture size
 #define BRDF_SIZE            512        // BRDF LUT texture size
+#define LIGHT_DISTANCE 1000.0f
+#define LIGHT_HEIGHT 1.0f
 
 // PBR material loading
 static Material LoadMaterialPBR(Color albedo, float metalness, float roughness);
--- a/projects/Notepad++/raylib_npp_parser/raylib_to_parse.h
+++ b/projects/Notepad++/raylib_npp_parser/raylib_to_parse.h
@@ -16,6 +16,9 @@ RLAPI bool IsWindowResized(void);                                 // Check if wi
 RLAPI bool IsWindowState(unsigned int flag);                      // Check if one specific window flag is enabled
 RLAPI void SetWindowState(unsigned int flags);                    // Set window configuration state using flags
 RLAPI void ClearWindowState(unsigned int flags);                  // Clear window configuration state flags
+RLAPI void SetWindowScale(float scaleFactor);                     // Set window scale factor
+RLAPI void SaveWindowScale(void);                                 // Save window scale to defaults
+RLAPI void RestoreWindowScale(void);                              // Load window scale from defaults
 RLAPI void ToggleFullscreen(void);                                // Toggle window state: fullscreen/windowed (only PLATFORM_DESKTOP)
 RLAPI void MaximizeWindow(void);                                  // Set window state: maximized, if resizable (only PLATFORM_DESKTOP)
 RLAPI void MinimizeWindow(void);                                  // Set window state: minimized, if resizable (only PLATFORM_DESKTOP)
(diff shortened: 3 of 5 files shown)


funcelot:
@raysan5 hi

raysan5:
@funcelot Wow! That's great! I need to review it carefully but getting high-dpi working properly would be really amazing! Just a question, would this PR allow removing `SUPPORT_HIGH_DPI` flag, just supporting it automatically???

funcelot:
> @funcelot Wow! That's great! I need to review it carefully but getting high-dpi working properly would be really amazing! Just a question, would this PR allow removing `SUPPORT_HIGH_DPI` flag, just supporting it automatically???

God new are coming, I'm working on fullscreen mode support from my PR, i added fullscreen more working on current app display, and storing all screenscale parameters structure by default before going fullscreen, so when it will fully work in desired resolution, from screen size 800 450, i think i can take a look into SUPPORT_HIGH_DPI flag support. 

@raysan5 Do you willing to remove it and allow fully automatic high DPI multimonitor support? I just check out flag for high DPI at some points and add more logic, i assume that in no high DPI mode, all scaling will be just disabled, or trivial using identity matrix, or division by 1.0f, so it will be plausible to remove some special cases i suppose.

raysan5:
@funcelot Wow, this is a big change! Great work done here! :D 

Some concerns:

 - Could we avoid `SaveWindowScale()` / `RestoreWindowScale()` to be exposed to the users in the API?
 - Could you try it on the other desktop platforms (Linux X11 / macOS)? Or at least, Linux platform...
 - About removing the `SUPPORT_HIGH_DPI` flag, I think that we should better keep it in case some user prefers to keep the content unscaled independently of the DPI, maybe it can be renamed to something like `SUPPORT_HIGHDPI_AUTO_SCALING`?

funcelot:
> @funcelot Wow, this is a big change! Great work done here! :D
> 
> Some concerns:
> 
> * Could we avoid `SaveWindowScale()` / `RestoreWindowScale()` to be exposed to the users in the API?
> * Could you try it on the other desktop platforms (Linux X11 / macOS)? Or at least, Linux platform...
> * About removing the `SUPPORT_HIGH_DPI` flag, I think that we should better keep it in case some user prefers to keep the content unscaled independently of the DPI, maybe it can be renamed to something like `SUPPORT_HIGHDPI_AUTO_SCALING`?

I need to review code, all my efforts were doomed due to the fact that i cant make a mouse working on a secondary monitor, do you have some advice or samples. glfw is quite unfinished lib for me, cause i spend so much effort to switch fullscreen app from one monitor to another, so may be you know some tricks?

funcelot:
> @funcelot Wow, this is a big change! Great work done here! :D
> 
> Some concerns:
> 
> * Could we avoid `SaveWindowScale()` / `RestoreWindowScale()` to be exposed to the users in the API?
> * Could you try it on the other desktop platforms (Linux X11 / macOS)? Or at least, Linux platform...
> * About removing the `SUPPORT_HIGH_DPI` flag, I think that we should better keep it in case some user prefers to keep the content unscaled independently of the DPI, maybe it can be renamed to something like `SUPPORT_HIGHDPI_AUTO_SCALING`?

I think i will totally review code, it is not usable in practical case, cause switching to fullscreen on a secondary monitor centers mouse in another and there is no any meaningful way i can see to prevent this, so yes, i can remove all functions, but event better to close the issue, until glfw will be more accurate in multimonitor configurations

funcelot:
> @funcelot Wow, this is a big change! Great work done here! :D
> 
> Some concerns:
> 
> * Could we avoid `SaveWindowScale()` / `RestoreWindowScale()` to be exposed to the users in the API?
> * Could you try it on the other desktop platforms (Linux X11 / macOS)? Or at least, Linux platform...
> * About removing the `SUPPORT_HIGH_DPI` flag, I think that we should better keep it in case some user prefers to keep the content unscaled independently of the DPI, maybe it can be renamed to something like `SUPPORT_HIGHDPI_AUTO_SCALING`?

Yes, i will test on X11 and Mac OS X mid 2013 with latest OS update, but i cant find a way to setup mouse pointer correctly

PR #1514, opened 2021-01-06
Big cmake changes

object71:
I refactored a big chunk of the main cmake files. They look more clean now (as most of it is rugged under the couch). One can now reason better with the configurations.


object71:
@raysan5 As we discussed in discord I brought some general improvements on cmake to make it look a bit simpler. Removed duplicating sections and pulled everything more complex into configuration files under <root>/cmake . It is still building normally, install also works

raysan5:
@object71 Wow! What a refactoring! I'm probably not the best person to review it... but it really seems more clear now...

src/config.h
```diff
@@ -29,7 +29,7 @@
 
 // Edit to control what features Makefile'd raylib is compiled with
 #if defined(RAYLIB_CMAKE)
```
raysan5, line 31: I don't like mixing the build system with the library logic... is there any other alternative on CMake side to avoid this `RAYLIB_CMAKE`?

raysan5:
@object71 I'm afraid I merged a small tweak on CMake and some conflics generated... sorry...

cmake/AddIfFlagCompiles.cmake
```diff
@@ -1,4 +1,7 @@
 include(CheckCCompilerFlag)
+
+MESSAGE(STATUS "Pass")
+
```
RobLoach, line 4: Looks like debug information.

``` suggestion
```

CMakeLists.txt
```diff
@@ -1,57 +1,24 @@
...
 
-if (ENABLE_MSAN AND ENABLE_ASAN)
-    MESSAGE(WARNING "Compiling with both AddressSanitizer and MemorySanitizer is not recommended")
-endif()
+# Registers build options that are exposed to cmake
+include(BuildOptions)
 
-add_subdirectory(src)
+# Main sources directory
+add_subdirectory(src raylib)
```
RobLoach, line 21: `add_subdirectory raylib` ? There is no raylib subdirectory?

CMakeLists.txt
```diff
@@ -1,57 +1,24 @@
 cmake_minimum_required(VERSION 3.0)
+project(raylib)
+
+# Directory for easier includes
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
+# TRUE if this project is build as a main project and not as a dependency
```
RobLoach, line 7: Maybe something more descriptive here?
``` suggestion
# RAYLIB_IS_MAIN determines whether the project is being used from root, or as a dependency.
```

cmake/BuildOptions.cmake
```diff
@@ -6,6 +6,19 @@ enum_option(PLATFORM "Desktop;Web;Android;Raspberry Pi;DRM" "Platform to build f
 
 enum_option(OPENGL_VERSION "OFF;3.3;2.1;1.1;ES 2.0" "Force a specific OpenGL Version?")
 
+# This helps support the case where emsdk toolchain file is used
+# either by setting it with -DCMAKE_TOOLCHAIN_FILE=<path_to_emsdk>/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake
+# or by using "emcmake cmake -B build -S ." as described in https://emscripten.org/docs/compiling/Building-Projects.html
+if(EMSCRIPTEN)
+    SET(PLATFORM Web CACHE STRING "Forcing PLATFORM_WEB because EMSCRIPTEN was detected")
+endif()
```
RobLoach, line 14: 1. Most CMake standards I've used have used `CMakeOptions.txt` for this, not `BuildOptions.cmake`... I'm okay with this change, just seems off-standard.
2. You're `SET`ing in an options file. Intentional?

src/CMakeLists.txt
```diff
@@ -1,326 +1,131 @@
...
-
-  add_test("pkg-config" ${PROJECT_SOURCE_DIR}/../cmake/test-pkgconfig.sh)
-else(SHARED)
-  add_library(raylib ALIAS raylib_static)
-endif(SHARED)
 
-if (NOT DEFINED PKG_CONFIG_LIBS_EXTRA)
-  set(PKG_CONFIG_LIBS_EXTRA "${PKG_CONFIG_LIBS_PRIVATE}")
-endif()
+include(GlfwImport)
```
RobLoach, line 39: Would `find_package` work for this? Use it as a dependency rather than an include?

RobLoach (changes requested):
Looks pretty good. A few questions. Happy to see CMake improvements! :+1:

object71:
I'll take a look tomorrow

CMakeLists.txt
object71: this will add the subdirectory but in the build directory it will rename it to raylib

object71: @RobLoach It still adds the src subdirecotyr. The second parameter actually renames it so that when you generate the project with cmake you would end up with /build/raylib/raylib.lib and not /build/src/raylib.lib . Cosmetic change mostly.

src/config.h
object71: @raysan5 There could be but I would suggest it to be in another pull request. This one is quite big and is mostly segmening the cmake files into more reasanable chunks.

src/CMakeLists.txt
object71: @RobLoach GlfwImport is just a new file taht I copied the logic that included the GLFW library. For segmentation purpose only because GLFW logic contained a lot of if... else conditions and polutes this file. Here you only care to know that the GlfwImport.cmake file is taking care of that - makes it cleaner and easier to reason with.

cmake/BuildOptions.cmake
object71: @RobLoach I will return the standard CMakeOptions.txt then to be more inline with other standards then. I though the set belongs nicely there but I guess it doesn't. It forces the PLATFORM option to default to WEB when compiling with Emscripten which is a web compiler.

CMakeLists.txt
RobLoach: Ah, didn't know you could change the binary directory like that! Very cool :+1:

object71:
@raysan5 @RobLoach Did improvements based on the review. Mainly I returned the CMakeOptions.txt file for the options and added 2-3 more comments here and there. This would be the largest pull request on cmake I hope and in the future I would look into clearing up the now separate smaller files and also as suggested see what we can do about the config.h

cmake/InstallConfigurations.cmake
```diff
@@ -0,0 +1,40 @@
...
+            RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
+            PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+    )
+else ()
+    install(
+            TARGETS raylib
+            LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+            PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+    )
+endif ()
```
Neumann-A, line 15:
```suggestion
  install(
          TARGETS raylib EXPORT raylib-targets
          ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
          LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
          RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
          PUBLIC_HEADER DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
  )
```

cmake/InstallConfigurations.cmake
```diff
@@ -0,0 +1,40 @@
...
+if (STATIC)
+    include(LibraryPathToLinkerFlags)
+    library_path_to_linker_flags(__PKG_CONFIG_LIBS_PRIVATE "${LIBS_PRIVATE}")
+    set(PKG_CONFIG_LIBS_PRIVATE ${__PKG_CONFIG_LIBS_PRIVATE} ${GLFW_PKG_LIBS})
+    string(REPLACE ";" " " PKG_CONFIG_LIBS_PRIVATE "${PKG_CONFIG_LIBS_PRIVATE}")
+elseif (SHARED)
+    set(PKG_CONFIG_LIBS_EXTRA "")
+    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_LIBDIR}")
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+    set(CMAKE_MACOSX_RPATH ON)
```
Neumann-A, line 27:
```suggestion
```
According to CMake those values are used to initialize the target properties. Since the target is already created those values do nothing here. (Or they belong above the add_library call. )

cmake/InstallConfigurations.cmake
```diff
@@ -0,0 +1,40 @@
...
+endif ()
+
+join_paths(libdir_for_pc_file "\${exec_prefix}" "${CMAKE_INSTALL_LIBDIR}")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+configure_file(../raylib.pc.in raylib.pc @ONLY)
+configure_file(../cmake/raylib-config-version.cmake raylib-config-version.cmake @ONLY)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/raylib.pc DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/raylib-config-version.cmake DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/raylib")
+install(FILES ${PROJECT_SOURCE_DIR}/../cmake/raylib-config.cmake DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/raylib")
```
Neumann-A, line 36: add an `install(EXPORT raylib-targets DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/raylib" NAMESPACE raylib:: FILE raylibTargets.cmake)` here and include the exported target file in `raylib-config.cmake`

cmake/LibraryConfigurations.cmake
```diff
@@ -0,0 +1,109 @@
...
+    set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
+    set(CMAKE_C_FLAGS "-s USE_GLFW=3 -s ASSERTIONS=1 --profiling")
+    set(CMAKE_STATIC_LIBRARY_SUFFIX ".a")
+
+elseif(${PLATFORM} MATCHES "Android")
+    set(PLATFORM_CPP "PLATFORM_ANDROID")
+    set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
+    include(AddIfFlagCompiles)
+    add_if_flag_compiles(-ffunction-sections CMAKE_C_FLAGS)
+    add_if_flag_compiles(-funwind-tables CMAKE_C_FLAGS)
```
Neumann-A, line 48: Messing around with compiler flags belong into the `CMAKE_TOOLCHAIN_FILE`.

cmake/InstallConfigurations.cmake
object71: @Neumann-A As it was configured before I extracted it into a separate file. I was planning on doing a few more passes after this refactor to clean up those stuff but thanks for the change I will take it.

cmake/LibraryConfigurations.cmake
object71: Out of the scope of this pull request @Neumann-A there will be more refactoring in the future. There is a lot of messing with compiler flags and not only in this file but the current aim is to size down the cmake system into a more consumable chunks.

cmake/InstallConfigurations.cmake
object71: @Neumann-A Thanks for spotting this. They were after the target initialization in the previous version of the file so I would gadly remove them.

object71:
@Neumann-A Thanks for your suggestions. I accepted your improvements but I will not work on big changes for this pull request. This will not be my only addition to improve the build system its just the biggest changeset as of this point.

raysan5:
@object71 @RobLoach @Neumann-A  Wow! Lot's of reviews! Thanks to you all for contributing on this big improvement! Please, let me know when you think it's ready to be merged!

object71:
Well I'm fine with the current state, so if the others are okay.

raysan5:
@a3f you have been in charge of the CMake build system all this time, what do you think about those changes?

examples/CMakeLists.txt
```diff
@@ -28,7 +28,6 @@ if (APPLE AND NOT CMAKE_SYSTEM STRLESS "Darwin-18.0.0")
   add_definitions(-DGL_SILENCE_DEPRECATION)
   MESSAGE(AUTHOR_WARNING "OpenGL is deprecated starting with macOS 10.14 (Mojave)!")
 endif()
-list(REMOVE_ITEM example_sources ${CMAKE_CURRENT_SOURCE_DIR}/others/rlgl_standalone.c)
```
a3f, line 31: Why can this be removed? Does rlgl_standalone no longer clash with raylib proper? The commit message should explain that.

src/CMakeLists.txt
```diff
@@ -1,326 +1,129 @@
...
+set(LIBS_PRIVATE ${LIBS_PRIVATE} ${OPENAL_LIBRARY})
 
-# populates raylib_{FOUND, INCLUDE_DIRS, LIBRARIES, LDFLAGS, DEFINITIONS}
-include(PopulateConfigVariablesLocally)
-populate_config_variables_locally(raylib)
+if (STATIC)
+    MESSAGE(STATUS "Building raylib static library")
+    
+    add_library(raylib STATIC ${raylib_sources})
+    add_library(raylib_static ALIAS raylib)
```
a3f, line 58: raylib_static was a workaround, so when building on Windows the lib files didn't override each other. Does this still work as intended?

src/CMakeLists.txt
```diff
@@ -110,8 +110,8 @@ target_include_directories(raylib
                            ${OPENAL_INCLUDE_DIR}
                            )
 
-# Copy the header files to the build directory
-#file(COPY ${raylib_public_headers} DESTINATION ".")
+# Copy the header files to the build directory for convenience
+file(COPY ${raylib_public_headers} DESTINATION "include")
```
a3f, line 115: You can `git rebase -i` and squash this commit into the original commit btw.

RobLoach:
Pulled the latest changes into [raylib-cpp](https://github.com/robloach/raylib-cpp), which uses CMake with `add_subdirectory("vendor/raylib")` . Running into the following issue, since the platform check had moved:
```
In file included from /home/rob/Documents/raylib-cpp/vendor/raylib/src/core.c:167:
/home/rob/Documents/raylib-cpp/vendor/raylib/src/external/dirent.h:83:10: fatal error: io.h: No such file or directory
   83 | #include <io.h>         // _findfirst and _findnext set errno iff they return -1
      |          ^~~~~~
compilation terminated.
```

a3f (commented):
Looks to me like there is no CMake CI anymore, which is unfortunate as the Travis/AppVeyor were quite varied and tests often spotted CMake regressions. Would be really great to have the Github CI extended by them before this pull request is merged, but that shouldn't be a blocker.

Didn't look too closely at the PR itself, I'd have preferred refactoring into files and changes to the content to happen in separate commits to make review easier, but overall looks fine to me.

examples/CMakeLists.txt
object71: Yes, they no longer crash, the examples build correctly on my machine. @a3f

src/CMakeLists.txt
object71: @a3f I would say **No** but the reason being is that you would be interested in building only one or the other at a given time. It will be more clear and also best if we start relying on the **default provided** [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) variable instead. Using that we can only define one library target and then the user will be responsible on deciding what to build and how to separate the two. At least that is a more clean and clear approach for any modern cmake user -[ read this article that describes why](https://cgold.readthedocs.io/en/latest/tutorials/libraries/static-shared.html).

I also planned on refactoring to the right approach once these improvements get accepted.

object71:
> Pulled the latest changes into [raylib-cpp](https://github.com/robloach/raylib-cpp), which uses CMake with `add_subdirectory("vendor/raylib")` . Running into the following issue, since the platform check had moved:
> 
> ```
> In file included from /home/rob/Documents/raylib-cpp/vendor/raylib/src/core.c:167:
> /home/rob/Documents/raylib-cpp/vendor/raylib/src/external/dirent.h:83:10: fatal error: io.h: No such file or directory
>    83 | #include <io.h>         // _findfirst and _findnext set errno iff they return -1
>       |          ^~~~~~
> compilation terminated.
> ```

@RobLoach I will look into it but the io.h file is part of the system files and shouldn't be affected (at least up to my knowledge) from cmake changes.

src/CMakeLists.txt
a3f: I guess we can drop it if it improves code clarity.

object71:
> > Pulled the latest changes into [raylib-cpp](https://github.com/robloach/raylib-cpp), which uses CMake with `add_subdirectory("vendor/raylib")` . Running into the following issue, since the platform check had moved:
> > ```
> > In file included from /home/rob/Documents/raylib-cpp/vendor/raylib/src/core.c:167:
> > /home/rob/Documents/raylib-cpp/vendor/raylib/src/external/dirent.h:83:10: fatal error: io.h: No such file or directory
> >    83 | #include <io.h>         // _findfirst and _findnext set errno iff they return -1
> >       |          ^~~~~~
> > compilation terminated.
> > ```
> 
> @RobLoach I will look into it but the io.h file is part of the system files and shouldn't be affected (at least up to my knowledge) from cmake changes.

@RobLoach I tried building raylib-cpp with the current version of the pull request and had no problems using mingw on windows. Btw you should know that raylib-cpp is not buildable with msvc because of some other errors. But I couldn't reproduce the problem with the <io.h> file missing. 

Again - it shouldn't be related to the cmake changes as it is one of the language / compiler files and not part of raylib at all! Could you check your system?

RobLoach:
> Btw you should know that raylib-cpp is not buildable with msvc because of some other errors. 

Could you share those errors?

a3f:
@RobLoach The breakage you report, can you reproduce it with current master or does it only occur after merging this pull request?

object71:
> > Btw you should know that raylib-cpp is not buildable with msvc because of some other errors.
> 
> Could you share those errors?

@RobLoach Yes, some of them were from raylib though so I created #1525 and I've written you in discord about the ones in raylib-cpp.

raysan5:
@object71 Changes merged.

object71:
@raysan5 Nice, I'll see to update the parts and also maybe include a cmake github actions (saw it was in the features for 3.6)

PR #1516, opened 2021-01-07
Add info to readme for conan dependency manager

resttime:
Hi, raylib is great. I recently added it to the conan dependency manager and thought it might be nice to add a description in the readme after seeing vcpkg on there.


raysan5:
@resttime Nice! Thank you very much for this addition!

PR #1517, opened 2021-01-07
Fix for PR #1504, orthographic matrix

JeffM2501:
The code in PR #1504 computed the height of the viewport incorrectly so it was always negative, thus causing problems when using the ortho camera. This PR fixes that oversight.


raysan5:
@JeffM2501 Oh! Good catch!

PR #1520, opened 2021-01-08
rlOrtho Divide by zero fix (again)

JeffM2501:
Sometimes top > bottom and sometimes bottom > top depending on drawing mode. This PR changes the ortho div by zero check for a simple equality. We don't care what the numbers are just so long as they are not the same in any axis.


raysan5:
As discussed on Discord, this change is being reverted.

PR #1521, opened 2021-01-08
File Read Callbacks

JeffM2501:
This PR adds in a set of read file callbacks to allow external libraries to provide file reading functionality to raylib. This allows for libraries to provide virtual file systems for raylib that can read from archives or databases with out raylib having to know or care about where the file data comes from while still using all existing file loading capabilities.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2595,7 +2595,7 @@ bool SaveStorageValue(unsigned int position, int value)
 
     unsigned int dataSize = 0;
     unsigned int newDataSize = 0;
-    unsigned char *fileData = LoadFileData(path, &dataSize);
+    unsigned char *fileData = RLLoadFileData(path, &dataSize);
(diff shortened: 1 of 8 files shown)


raysan5:
Hi @JeffM2501! I've been trying to avoid callback mechanisms exposed in raylib... but I'm planning to expose those file-access, together with memory-allocation, I just want to think about it carefully... I keep this PR open for reference.

JeffM2501:
That makes complete sense. I just wanted you to have an implementation for reference. You don't want to confuse people with callbacks when they are just starting out. They should be 100% optional.

Perhaps they should go into separate header that is not referenced by raylib.h to keep them isolated? Maybe like raylib_callbacks.h or something?

raysan5:
I'm already working on this. Closing this PR.

PR #1528, opened 2021-01-12
Reorder typedefs in physac.h to be in header part

greenfork:
Just reorder the code, no changes.

Why:
- It is logical to include typedefs in API part, they are directly used in examples
- It is in line with how raylib.h is
- It makes it easier to parse this file and separate only "header" part (relevant at least for wrappers)

### DIFF
--- a/src/physac.h
+++ b/src/physac.h
@@ -125,10 +125,71 @@
     } Vector2;
 #endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@greenfork Nice! Thanks for the improvement!

PR #1530, opened 2021-01-13
Fixes Android builds on Linux environments

Theldus:
Description
----------------
This PR corrects some build issues on Linux environments for building Android applications in templates/simple_game.

Fixes/improves:
- Automatically detects if host OS is Windows or Linux in the root Makefile, which avoids manual editing of the file to include 'Makefile.Android.linux' instead of Makefile.Android.

- Leave `ANDROID_HOME` and `ANDROID_TOOLCHAIN` with default values only if they do not already exist, which allows the user to add them as environment variables, and thus, Makefile-independent.

- Fixes `lib$(PROJECT_LIBRARY_NAME).so` builds: the generated library was configured to use the path of the NDK libraries instead of the native Android libraries.

### DIFF
--- a/templates/simple_game/Makefile
+++ b/templates/simple_game/Makefile
@@ -101,6 +101,15 @@ ifeq ($(PLATFORM),PLATFORM_DRM)
         PLATFORM_OS=LINUX
     endif
 endif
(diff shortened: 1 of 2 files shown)


raysan5:
@Theldus Thanks for the improvements!

PR #1536, opened 2021-01-16
Fix cmake build error dirent

### LINKED ISSUE #1533: [build] CMake build error on Linux, including "src/external/dirent.h" instead of system's
Building with CMake on Linux errors with the following message:
```
/media/dados/prog/raylib/src/external/dirent.h:83:10: fatal error: io.h: No such file or directory
   83 | #include <io.h>         // _findfirst and _findnext set errno iff they return -1
      |          ^~~~~~
```

The build is including `src/external/dirent.h` instead of the system's default `dirent.h`, it seems the problem is adding `src/external` to import path here: https://github.com/raysan5/raylib/blob/master/src/CMakeLists.txt#L108.

object71:
fixes #1533

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,7 +82,7 @@ DerivedData/
 
 # Jetbrains project
 .idea/
-cmake-build-debug/
+cmake-build-*/
 
 # CMake stuff
 CMakeCache.txt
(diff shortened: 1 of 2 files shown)


object71:
@raysan5 I also added one gitignore edit because I want to fix the following situation in CLion when I have mutliple configurations or want to build release instead of debug.

image

raysan5:
@object71 nice! thanks for the fix

PR #1539, opened 2021-01-17
BINDINGS: Update node-raylib to 3.5

RobLoach:
[`node-raylib`](https://github.com/robloach/node-raylib) is now on raylib 3.5.0


raysan5:
@RobLoach Thanks for the update! 😄

PR #1542, opened 2021-01-21
Remove STATIC and SHARED variables.

object71:
As described in the official documentation https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html this flag is global by default and controls if the library will be built as a shared or a static library allowing us to define only one call to the add_library function (without specifying its type). It is also added as an option to be visible in CMake GUI applications.


raysan5:
@object71 thanks for the improvement! 😄

nawarian:
Hi all! This seems to have create an issue on Homebrew's side.

It previously installed both the shared and static objects but now it seems like this is pretty binary, we can't compile and install both static and shared lib objects at once. Can you confirm? Was this intended?

See https://github.com/Homebrew/homebrew-core/pull/80224 for context.

Sorry about reopening this discussion, I just didn't want to create a new Issue without being sure this is a bug or feature.

Thank you!

raysan5:
@object71 ?

object71:
Was a feature for better general cmake compatibility. But it could be reworked again if there is a case that is better off with having both installed together

nawarian:
The context I brought this up was that homebrew was broken because of it, I can imagine other pipelines got broken too.

But 3.7.0 was a breaking version anyways with lots of API changes, so I think it is alright. I fixed the pipeline there by compiling it twice, with different flags and it is kind of acceptable.

I think we should at least update the documentation to explicitly mention it is necessary to build both static and shared libs at different times.

Let me know if you decide to bring back the old behavior of building both at once so I can come back and update Homebrew again :)

raysan5:
@nawarian personally, I prefer to keep them as individual processes and let the systems that require both libraries run the two processes separately.

PR #1543, opened 2021-01-22
Added a windows and linux simple compile with cmake

object71:
Will expand on the file but it will be nice to know if they are some extremely breaking changes in the future. cmake.yml will run two task one for linux build and one for windows build essentially checking that both are working. We can add all platforms later.


raysan5:
@object71 Wow! Very nice addition! Thank you very much!

PR #1544, opened 2021-01-22
Fix issue when trying to build raylib statically

Not-Nik:
This PR fixes an issue where raylib would build as a shared library regardless of the value in `BUILD_SHARED_LIBS`. This was due to a simple syntax mistake in `cmake/GlfwImport.cmake`.


raysan5:
@Not-Nik 👍

PR #1545, opened 2021-01-24
Removing test file.

object71:
This test file is just testing compilation with the library works correctly but is no longer needed because:
- it is not cross platform
- it taps into the CTest system which is better suited for real unit/integration tests
- it can be incorporated into the pipeline of github actions instead in the future as it tests library installation


raysan5:
@object71 nice! thanks for the review!

PR #1546, opened 2021-01-25
Removing config.h.in file

object71:
CMake relied on this file for configurations and also was interfering in the regular config.h by having a separate definition if building with CMake. This was not entirely correct so instead we will define compile time definitions separately through CMake (CompileDefinitions.cmake) and also will use the provided EXTERNAL_CONFIG_FLAGS that I found that will not use config.h in through the build process.

I also introduced a new compiler option (CUSTOMIZE_BUILD) that when OFF will use the default config.h and when ON will show other options for redefining your own options.

Fixed an error in rlgl.h where if you have both RLGL_STANDALONE and SUPPORT_VR_SIMULATOR you get a compile time error.


raysan5:
@object71 Thanks you very much! Actually it was on my TODO list to remove that CMake check from `config.h`, I think it was the only piece of code coupling raylib sources with an external build system! Thanks! :D

object71:
@raysan5 No problem, You mentioned it in the big changes request. Also makes things more clear for CMake users...

PR #1549, opened 2021-01-26
Documentation cmake

object71:
@raysan5 I added some comments to describe the different sections of CMake. I also made some minor changes (mostly restructuring). I think this is enough refactoring on the build system and I will not do any more changes on it. Will look into the github actions though or some other raylib features.


raysan5:
@object71 Thank you very much! This is truly appreciated! I like reading commenting code, for me is the best way to learn! Thanks! :)

PR #1551, opened 2021-01-27
Gltf animation support

### LINKED ISSUE #1006: [models] glTF animations support
The iqm animator naively "pre-renders" vertex's for each key frame of the animation, and interpolates between them.

GLTF often supplies animations in terms of for example a rotation looking at the example here https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedTriangle you can see there are 5 key frames for this triangle this would be 44*5 bytes just for a single rotation animation of a single triangle!

additionally it is perfectly valid to have multiple animations running concurrently and its equally valid that they need not have the same number of key frames.

As I understand the problem, the only viable solution is to only have one set of vertices and translate them as per the GLTF node hierarchy, this is the only way I could see a separate translation animation channel being able to work with for example a separate rotation channel where they have a different number of key frames.

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#animations see here for an explanation of animation channels

object71:
@raysan5 Hi, I have implemented most of what is required of gltf animation feature and this should resolve #1006 . I map the animations to the current structures provided. I also created two new examples for gltf model loading and gltf model loading with animation.

**What is not implemented**: as discussed in discord the current state of the animation for gltf would require the animation to interpolate between the poses. The rigged figure in the example has only two frames that should be interpolated between so the example would look pretty much glitchy but it is working as a whole. **I will create a new issue for a feature request for animation pose interpolations** and leave the current state of this example as is.

I also modified the general load animation method to now switch between IQM and GLTF based on file extension like it was done in the load model method.

Provided assets are copied from [GLTF-Samples](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0) repository and the licenses for them are provided in a LICENSE file.


object71:
I need to clean up the formatting when I look at it here.

Also keep in mind that this pull request doesn't completly follow the rules as GLTF animations could have different number of frames provided for different parts of the mesh. I will open another issue and work on loading the models like in IQM predefined with some timestep (30 poses per second for example).

Still this could be accepted as a foundation upon which we can improve the animations.

raysan5:
@object71 Just took a quick look, congratulations, very clean and organized code! Just noted some TABS that should be replaced by 4 spaces...

Please, let me know when it's ready to merge!

object71:
@raysan5 Fixed all the identation issues I think

object71:
And thank you! It is still not complete but I wanted to submit the partially working section.

object71:
I think it should be ready for merge

raysan5:
@object71 Merged! Great addition! Thanks!

PR #1557, opened 2021-01-30
Use local includes for GLFW 

JeffM2501:
GLFW was included using #include <GLFW. This tells the compiler to check the system library paths for the header. Raylib includes it's own GLFW so it should be using #include "GLFW.
This PR makes this change, and helps some platforms produce fewer errors/warnings.


raysan5:
@JeffM2501 thanks!

PR #1558, opened 2021-01-31
[Shaders] Fix compiler error in GSSL 3 version of color_mix_shader

JeffM2501:
out is a keyword in shaders and can't be used as a variable name.
This PR changes it to outVal so it compiles.


raysan5:
@JeffM2501 Good catch! 👍😄

PR #1560, opened 2021-02-01
DrawTexturePro vertex label correction

Auios:
I noticed the comments referencing the quad's vertices were incorrect.


raysan5:
@Auios Actually the comment is correct, quad is defined starting by the bottom-right corner: `rlVertex2f(0.0f, dest.height);`

PR #1561, opened 2021-02-01
Update loading of gltf animation.

### LINKED ISSUE #1553: [models][gltf] Similar to IQM on model loading interpolate a const set of poses

### Issue description

Currently when loading a model (look at gltf animation example) you would only get two poses. This is not the correct mapping for the current version of raylib which expects a set of poses targeting 30/60 FPS. 

So it will be good if for each transformation we interpolate the position for a set number of frames.

object71:
This is to account for GLTF info being more like instructions on how to build your animation instead of verbose description of each pose.

Fixes #1553 

example

PS: Also added a few comments to explain the animation loading.

### DIFF
--- a/examples/models/models_gltf_animation.c
+++ b/examples/models/models_gltf_animation.c
@@ -50,10 +50,11 @@ int main(void)
     int animsCount = 0;
     ModelAnimation *anims = LoadModelAnimations("resources/gltf/rigged_figure.glb", &animsCount);
     int animFrameCounter = 0;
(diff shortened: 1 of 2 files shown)


object71:
@raysan5 This fixes the issue where only 2 frames were switching in the example by sampling a constant of 30 poses per second.

raysan5:
@object71 thanks for the update! yeah, I saw the issue! thanks!

PR #1564, opened 2021-02-02
[core] Add functions to set near and far culling planes.

JeffM2501:
Currently the near and far clipping planes are hardcoded in raylib and can only be changed by rebuilding raylib. This PR adds settable near and far plane values to raylib and uses them when creating projection matrices.

By default the same config planes are used, so no code changes are required for existing code to work like it used to.


JeffM2501:
As discussed on discord, this can be handled with custom camera code and custom begin/end mode function.s

PR #1566, opened 2021-02-04
Added CheckCollisionPointPolygon()

Kirottu:
Added a function to check if a point is inside an array of vertices based on code found on [GeeksForGeeks](https://www.geeksforgeeks.org/how-to-check-if-a-given-point-lies-inside-a-polygon/)

The function takes the polygon as an array of Vector2s, the count of the vertices and the point to be checked.

I thought this would be a very nice addition to raylib.


src/raylib.h
```diff
@@ -1130,6 +1130,7 @@ RLAPI bool CheckCollisionPointRec(Vector2 point, Rectangle rec);
 RLAPI bool CheckCollisionPointCircle(Vector2 point, Vector2 center, float radius);                       // Check if point is inside circle
 RLAPI bool CheckCollisionPointTriangle(Vector2 point, Vector2 p1, Vector2 p2, Vector2 p3);               // Check if point is inside a triangle
 RLAPI bool CheckCollisionLines(Vector2 startPos1, Vector2 endPos1, Vector2 startPos2, Vector2 endPos2, Vector2 *collisionPoint); // Check the collision between two lines defined by two points each, returns collision point by reference
+RLAPI bool CheckCollisionPointPolygon(Vector2 polygon[], int vertexCount, Vector2 point);                // Check if point is inside a polygon
```
raysan5, line 1133: To be consistent with the other `CheckCollision*()` and `DrawPoly()` functions, could you change signature to:
```c
RLAPI bool CheckCollisionPointPoly(Vector2 point, Vector2 *points, int pointsCount);
```

raysan5:
@Kirottu I added some review notes... but my main concern is, how is the provided polygon considered? Is it considered as a triangle strip? a triangle fan? should it be closed, could it be just a open set of lines? does it support convex triangles?

Kirottu:
I will look at it, im pretty sure it should support any types of polygons. Look at the article as it goes more in depht about it. But yes, I will look at trying to make those go well with your suggestions.

src/shapes.c
```diff
@@ -1521,6 +1524,40 @@ bool CheckCollisionLines(Vector2 startPos1, Vector2 endPos1, Vector2 startPos2,
...
+{ 
+    // There must be at least 3 vertices in polygon[] 
+    if (vertexCount < 3) return false; 
+ 
+    // Create a point for line segment from p to infinite 
+    Vector2 extreme = {10000, point.y}; 
+ 
+    // Count intersections of the above line with sides of polygon 
+    int count = 0, i = 0; 
+    do
```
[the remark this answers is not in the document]
Kirottu: It would seem like avoiding that would mean unnecessary if statements and variables.

src/raylib.h
Kirottu: That is done now.

raysan5:
@Kirottu how performant is this function? maybe most user could just use `CueckCollisionPointCircle()` instead?

Kirottu:
Basically, what it does is it takes the point and draws a basiaclly infinite (10 000) line from it to the right, then it checks how many times the line has gone thru a polygon boundary and if this value is odd it will return true as the point is inside the polygon. Technically this function could be adapted to use the radius and size variables like rest of raylib for more simplicity. And also in a large scale of polygons `CheckCollisionPointCircle` becomes very imprecise.

raysan5:
@Kirottu I'm afraid I'm not merging this function, it adds a good amount of code and complexity (and maintainability) just for a feature that many users can implement for their respective use cases.

PR #1570, opened 2021-02-06
[CORE] Add GetCurrentFrameBuffertSize

JeffM2501:
This PR adds a function to return the size of the current frame buffer object to the public raylib API.

This is needed to allow external libraries and code to be able to create proper projection matrices when rendering to a texture.
Currently the only thing a library can do is get the screen or monitor size, not the size of the current rendering context. As there are examples that use render textures that are not the same size as the screen, and it is encouraged for people to create external library integrations and camera code, this function is needed.

Integrations of things like Dear ImGui can only set the screen size with functions like
 io->DisplaySize = (ImVec2){(float) GetScreenWidth(), (float) GetScreenHeight()};
If the user wants to render to a texture that is not the same size as the screen, the viewport will be incorrect and there will be no way to resolve the issue.

This also prevents client code from calling rlPerpsective or rlOrtho without knowing if it is rendering to the screen or a texture, making it impossible to create generic external camera code.

The other option would be to have GetScreenWidth and GetScreenHeight return the fbo size instead of the window size, since the "Screen" is the thing that you are currently rendering to.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1429,6 +1429,12 @@ int GetScreenHeight(void)
     return CORE.Window.screen.height;
 }
 
+// Get width and height of the current rendering context (screen or texture)
+Vector2 GetCurrentFrameBuffertSize()
+{
+    return (Vector2) { (float)CORE.Window.currentFbo.width, (float)CORE.Window.currentFbo.height };
+}
+
 // Get native window handle
 void *GetWindowHandle(void)
 {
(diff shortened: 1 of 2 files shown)


raysan5:
@JeffM2501 I'm reviewing this issue and I think `GetScreenWidth()` and `GetScreenHeight()` is the best approach to return current framebuffer size. Actually, screen width/height returned will be window width/height most of the time for most uses, if not, then it's the user who manages the framebuffer.

JeffM2501:
Sounds good to me

raysan5:
Reviewed GetScreenWidth() and GetScreenHeight() to return current fbo.

Issue #1554, opened 2021-01-28
[core] First frame hiccup on Linux when compiling raylib with embedded GLFW


gilzoide:
Hi! I'm on a x86_64 Manjaro Linux on X11 and everytime I build raylib + some project, it gives this super long first frame.
![raylib-first-frame](https://user-images.githubusercontent.com/4603360/106160969-2bd67c00-6165-11eb-932d-53b390563404.gif)

I tracked down that this problem doesn't occur when linking with the raylib installed in my system, which uses external GLFW (https://github.com/archlinux/svntogit-community/blob/packages/raylib/trunk/PKGBUILD#L19-L31), and neither when compiling with external GLFW. Here is how it goes:
![raylib-first-frame-ext-glfw](https://user-images.githubusercontent.com/4603360/106160983-309b3000-6165-11eb-91ef-b7ff7449dff9.gif)

I checked that the GLFW version installed here is 3.3.2 and that raylib's bundled version says 3.4.0, but this by itself doesn't seem to be the problem, I built a minimal GLFW window example linking to raylib's GLFW and the problem doesn't arise.
![glfw-first-frame](https://user-images.githubusercontent.com/4603360/106161001-35f87a80-6165-11eb-995f-872cc8ad6906.gif)

I'm building with CMake, but the same happens when building raylib and the examples using `make`.
![raylib-physics](https://user-images.githubusercontent.com/4603360/106161997-61c83000-6166-11eb-86a2-60ae342227a7.gif)

raysan5:
@gilzoide this is very weird...

gilzoide:
I know, right? Don't know if anyone had this problem before, maybe it's just my build environment making something unexpected...

I'd say this is not a big problem, because it's just me and I guess building executables linking with external GLFW is fine, Linux distributions have good package managers, maybe most users will have GLFW installed already.

gilzoide:
I just now tried rolling back to 3.5.0 and building the physics example again, same problem.
3.0.0, 2.6.0, same thing.
2.5.0 worked! =O

I'll use `git bisect` later to find out what the hell is happening, this is really bugging me out!

gilzoide:
Just FYI, after bisecting, first bad commit is 83916bd3c271d45fd60431b66fca0560cd0c2f93, regarding a GLFW update.
It's a big commit, I'll not go into it today, but will later. I'll keep in touch if I discover more about it.

raysan5:
[...]

PR #1572, opened 2021-02-07
Fix initialize GLFW's Joystick before window is created on Desktop platforms (#1554)

gilzoide:
This PR forces Joystick initialization before the window is created and shown on Desktop platforms, to avoid long first frames as in #1554.


raysan5:
@gilzoide Nice! Thanks for the fix!

PR #1573, opened 2021-02-08
Poll inputs after frame sleep

masterex1000:
This should hopefully slightly reduce input latency when raylib is manually limiting the frame rate of the application. I don't have any concrete numbers, but using the slow mo mode on my camera and running the `core_input_mouse` example it does look like the ball stays slightly closer to my mouse cursor.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1847,7 +1847,6 @@ void EndDrawing(void)
 #endif
 
     SwapBuffers();                  // Copy back buffer to front buffer
-    PollInputEvents();              // Poll user events
 
     // Frame time control system
     CORE.Time.current = GetTime();
@@ -1867,6 +1866,8 @@ void EndDrawing(void)
 
         CORE.Time.frame += waitTime;      // Total frame time: update + draw + wait
     }
+
+    PollInputEvents();              // Poll user events
 }
 
 // Initialize 2D mode with custom camera (2D)


raysan5:
@masterex1000 mmmmh... I'm not sure about the benefits of this change... I need to think about it... do you have some example that can illustrate the difference?

masterex1000:
I believe it will only help when raylib is limiting the framerate and not when vsync is being used, since the sleep wouldn't be taking any time in that situation. Realistically it would only be a few milliseconds of latency, but since its such a small change that shouldn't affect anything else, a free boost (or decline?) of input latency is great, no matter how small.

This is just trying to poll input as "late" as possible so we have the latest input info. I understand why you didn't poll in BeginDrawing(), so the next best place to poll would be at the very end of the `EndDrawing()` function.

raysan5:
@masterex1000 the problem I see with this change is that the time required for polling is not considered for frame synchronization (neither on update or draw time measurement). Maybe it doesn't matter on `PLATFORM_DESKTOP` but it could be a problem for other platforms.

masterex1000:
It won't be considered during the first draw cycle, but timings on those always seem messy anyway. However since raylib times the total time taken between calls to `EndDrawing()` it will still be included if the code is just running in a while loop. In fact, it almost seems like it would stabilize input polling since it would happen at a smooth 60hertz (or whatever framerate your limiting to) since input polling would no longer be tied the the amount of time it took to update and render the scene.

raysan5:
@masterex1000 ok, let's merge it.

PR #1574, opened 2021-02-08
BINDINGS: Remove Harbour

RobLoach:
It looks like the Harbour bindings disappeared? Did they exist at one point?


raysan5:
@RobLoach yeah... it disappeared... :(

PR #1582, opened 2021-02-10
Add methods to transform the VR output

shraiwi:
# This PR:
- Adds a function to the VR simulator called `SetVrOutputTransform()` that allows users to rotate and reflect the VR simulator output
- Adds logic in `EndVrDrawing()` that will transform the VR output according to the transform flags
- Adds `VrOutputTransform` enum
- Updates the `core_vr_simulator` example to include `SetVrOutputTransform()`

# For example:
`VR_OUTPUT_ROTATE_NONE`
VR_OUTPUT_ROTATE_NONE

`VR_OUTPUT_ROTATE_180`
Screenshot from 2021-02-10 10-20-38

`VR_OUTPUT_FLIP_V | VR_OUTPUT_ROTATE_180`
VR_OUTPUT_FLIP_V | VR_OUTPUT_ROTATE_180

# Why?
To my knowledge, raylib is the only library that has a VR simulator that can be used to directly drive a real VR headset. However, not all VR headset displays are mounted vertically. By adding this functionality, it would be easy for users using phones or embedded systems (such as the Pi 4) to rotate and/or flip the display without a window server. Since the transformations just change vertex coordinates, it has a negligible performance impact.

### DIFF
--- a/examples/core/core_vr_simulator.c
+++ b/examples/core/core_vr_simulator.c
@@ -72,6 +72,8 @@ int main(void)
 
     SetCameraMode(camera, CAMERA_FIRST_PERSON);         // Set first person camera mode
 
+    SetVrOutputTransform(VR_OUTPUT_ROTATE_NONE);        // Do not rotate the VR output
+
     SetTargetFPS(90);                   // Set our game to run at 90 frames-per-second
     //--------------------------------------------------------------------------------------
 
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -872,6 +872,15 @@ typedef enum {
     NPT_3PATCH_HORIZONTAL   // Npatch defined by 3x1 tiles
 } NPatchType;
 
+// VR Output Transformations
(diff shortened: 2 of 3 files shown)


raysan5:
@shraiwi Thanks for the review and excuse me for my late response. I'm still thinking about merging those changes or not, I'm considering instead adding an option to get the framebuffer and let the user manage it as desired.

shraiwi:
That might be a better idea, because then they could use the framebuffer for other things as well.

shraiwi:
Should I close this PR?

raysan5:
@shraiwi not yet, I prefer to keep it open for me to remember.

raysan5:
@shraiwi VR simulator has been moved to core module and completely redesigned. Now user is in charge of stereo-render fbo and also in full control of distortion shader. Example code is a bit more complex but better aligned with other raylib examples.

PR #1586, opened 2021-02-14
Fix support of touchscreens for RPI

### LINKED ISSUE #1585: [core] Left Click not registering using a touchscreen on a Raspberry Pi
- [ ✅ ] I tested it on latest raylib version from master branch
 - [ ✅] I checked there is no similar issue already reported
 - [ ✅] My code has no errors or misuse of raylib

### Issue description

I'm using a cheap TFT display on a Raspberry Pi Zero, the cursor works great when using in tandem with tslib, the problem arrises that there's no implementation of a left click, so clickable elements are imposible with the current implementation.

Tslib exposes a ABS_PRESSURE event code, using it to simulate a left click would be a great idea.

### Environment

Raspberry Pi Zero and a cheap TFT display

Jpe230:
The current implementation of the EventThread for the Raspberry Pi doesn't resolve isMouseButtonPressed/Up/Down() neither does it updates the position if using GetTouchPosition(int) if using a touchscreen 

This PR attempts to solve this shortcomings by adding some extra logic: 

-It checks the absolute pressure in the screen in an absolute event to simulate a touch action or a left mouse click.

-Updates the touch position at index 0 on absolute events.

This fixes #1585

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -5752,8 +5752,9 @@ static void *EventThread(void *arg)
                 // Basic movement
                 if (event.code == ABS_X)
                 {
-                    CORE.Input.Mouse.position.x = (event.value - worker->absRange.x)*CORE.Window.screen.width/worker->absRange.width;   // Scale acording to absRange
-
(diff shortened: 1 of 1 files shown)


raysan5:
@Jpe230 Thanks! Nice improvement!

PR #1588, opened 2021-02-14
[raudio] Fix load and unload issues with Music

ChrisDill:
Made changes to raudio to fix issues with `LoadMusicStream` and `UnloadMusicStream`.

Issues fixed:
- If a format is not valid or fails to load it is no longer assumed to be wav. This makes sure the correct unload call is made and makes the format clear to the user.

- Fixed crashing with invalid wav files when unloading. It was trying to use a callback where it shouldn't because it was not initialized to a zero value.

- Fixed crash in jar_xm when unloading a NULL ctxData.


raysan5:
@ChrisDill Nice! Thanks for the improvement!

PR #1589, opened 2021-02-14
Fixed problem related to `IsKeyPressed` function!

Rabios:
There was problem with using `IsKeyPressed`, In example code it uses `char` while `IsKeyPressed` takes an `int`.
However, This fixed!


raysan5:
@Rabios thanks!

PR #1591, opened 2021-02-15
Rename raylib-js to raylib-duktape

RobLoach:
While it is JavaScript, I'd rather this be named raylib-duktape since there are different implementations of javascript available. Like QuickJS, V7, or Node.js.


raysan5:
@RobLoach Good idea! 👍😄

PR #1594, opened 2021-02-16
[models] make sure glTF parent bones are not null before trying to compute the offset from the root

JeffM2501:
The current glFT code assumes that the parent bone is always non null and that it can always compute an offset from it to the root bone. This is not always the case, some bones don't have a parent. If the parent bone node is null, this will compute a very bad offset and cause a segfault later in the code, because the offset is not checked against the bone list bounds.
This change only computes an offset for the parent if the bone is non-null, attaching all null parent bones to the root node.


raysan5:
Hey @JeffM2501! Thanks for the fix!

PR #1595, opened 2021-02-17
fixed the right and left button not working

Redcamellia:
after checking the mouse_right_button down there was an else which prevented any changes to selectedColor that was caused by arrow keys ... 
after some up and running I could fix the problem of arrow keys ...


raysan5:
@Redcamellia thanks for the improvement, I'm reviewing it right now.

PR #1597, opened 2021-02-18
Fixed example doesn't show gamepad for Xbox 360 controllers with different drivers...

Rabios:
Now it should works as well!


raysan5:
@Rabios thanks for the improvement!

PR #1602, opened 2021-02-20
[rnet] fixes for VisualStuido 2019 builds

JeffM2501:
This PR includes fixes to make rnet and it's examples build on Windows in visual studio 2019.
Mostly typecasts, and some int to unsigned short changes, and including the windsock 2 library in the examples.

There are still some warnings in rnet.h but they may be due to the fact that Winsock2 and Berkley sockets are slightly different. I did not want to make any changes that would break Linux/MacOS builds.


raysan5:

PR #1608, opened 2021-02-23
Update RenderBatch to use Mesh

ChrisDill:
Initial work updating the batching system so it uses the `Mesh` internally.

It still uses custom OpenGL to manage it and I put the `Mesh` in `VertexBuffer` as that was the easiest way to get it working. This can maybe be changed in the future.

I have tested this with `OPENGL_33` on Linux by using it with the examples and everything seems to work the same. I could use help testing other configurations such as `OPENGL_ES2` to make sure I have not caused any new OpenGL issues.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -806,16 +806,7 @@ typedef struct VertexBuffer {
     int tcCounter;              // Vertex texcoord counter to process (and draw) from full buffer
     int cCounter;               // Vertex color counter to process (and draw) from full buffer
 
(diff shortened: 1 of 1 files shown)


raysan5:
@ChrisDill Sorry, I'm not merging this change at this moment. Replacing current `VertexBuffer` with minimal required data by a `Mesh` structure with multiple useless data references for the batching mechanism, I think it complicates the code unnecessarily at this moment.

PR #1610, opened 2021-02-23
Patch up GLTF Skeleton loading

chrissinclair:
Few fixes here:
- Add support for u8 bone indicies when loading glTF animations
- Fix a potential segfault when interpolating glTF animations when a given channel has it's first keyframe timed not at 0s.
- Ensure we respect the parent hierarchy for nodes when loading glTF animations, regardless of the order as provided in the nodes list.

I've tested this in a personal project on windows, but no other platforms as yet, using some kenney.nl assets that were exported as glTF from the source blend files. I suspect that there's more that could be done here, but this should help patch up a couple of common issues, glTF seems to support cameras, multiple scenes each with multiple models in, and at the moment we're just reading the raw nodes list which would contain everything. This is probably fine for loading models / animations as we'd not expect to have lots of them in the same glTF, but something to keep in mind.

Where appropriate there's more details on specific changes in the comments.

As an aside - thank you for putting raylib together, I've been using it a bit at work and it's made life significantly easier, very much appreciated!

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -3879,7 +3879,28 @@ static Model LoadGLTF(const char *fileName)
                             short* bones = RL_MALLOC(sizeof(short) * acc->count * 4);
                             
                             LOAD_ACCESSOR(short, 4, acc, bones);
-                            for (unsigned int a = 0; a < acc->count * 4; a ++)
+                            for (unsigned int a = 0; a < acc->count * 4; a++)
+                            {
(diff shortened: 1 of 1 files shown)


raysan5:
@chrissinclair Nice! Thank you very much for the improvements! glTF animations support has been added lately and still requires some work! All contributions are really welcome!

PR #1611, opened 2021-02-24
jar_xm_create_context_safe requesting const char*

nezvers:
Fixing compiler warning for passing (unsigned char*) to (const char*)


raysan5:
@nezvers nice!

PR #1613, opened 2021-02-25
LoadMusicStreamFromMemory OGG support

nezvers:
stb_vorbis.h also has an API for loading from memory.


nezvers:
That would make a feature-complete new API :)

raysan5:
@nezvers Wow! Great work! Thank you very much! :D

PR #1614, opened 2021-02-25
Added raylib-v7 and done edit as comparing between JavaScript binding…

Rabios:


raysan5:
@Rabios Great! Nice addition! 👍😄

Issue #1619, opened 2021-02-28
[Textures] PBR GenTexture functions leave OpenGL state with backface culling off


JeffM2501:
The functions GenTextureCubemap, GenTextureIrradiance, and GenTexturePrefilter all disable backface culling, but never restore it. The code is there but it is commented out. This does not seem like a desired side effect of a function named "GenTexture...". My guess is this was an oversight during development and was not the intended behavior.

This is actually the only reason the skybox example works, because the cube mesh is generated with face winding for 'outside' viewing. This makes it very confusing if you don't want to use a HDR image for a skybox and instead want to call LoadTextureCubemap, as it does NOT disable backface culling. You end up with a case where you change a loading function and don't realize your geometry winding is now a problem.

I think it should be up to the user to set the culling state before drawing anything and the GenTexture functions should fully restore the state.

I think the skybox example should then disable culling and depth writes, so that that it can be an example for both PRB and regular textures in other skybox formats.  I also think the skybox shader should have an optional parameter to not apply gamma correction as it is not needed for non HDR images.

I believe this is one reason why people feel that other skybox formats "don't work". They do work, it is just that the example needs more code that is non obvious to work

PR #1621, opened 2021-02-28
Fix GL state after HDR texture functions and update skybox example for #1619 

JeffM2501:
This is a proposed fix for issue #1619.

It restores the back face culling state after the HDR texture generation functions are called, and updates the skybox tutorial to work with that change.

The skybox tutorial now correctly disables and enables backface culling and depth writes.
The skybox tutorial has also been updated to show an example using a traditional "Cross" skybox in addition to an HDR panorama.
The skybox shader has been updated to make gamma correction optional as well. A sample skybox image (public domain) has been included.

I feel this will provide a better tutorial for people trying to learn the different methods for creating a skybox/cubemap.

image


raysan5:
@JeffM2501 Thank you very much for noticing the issue and the proposed review! I try to abstract as much as possible OpenGL from the user but... there are too many possible situations! It seems all pipeline functionality is required in some or other way sooner or later! :P

JeffM2501:
Yeah the only other way to fully abstract it would be a dedicated DrawSkybox function that ether generated a reversed cube, or did all the state changes required.

PR #1624, opened 2021-03-01
Add check to flip quaternion

masterex1000:
This commit add a check to `QuaternionSlerp()` which inverts one of the quaternions if they are inverted to each other (same rotation, just all components are negative)

The check is outlined here in the other issues section on this page.
https://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/slerp

Three.js has this exact same check in their slerp function, so I'm guessing this is the standard way to do this in graphics focused stuff.


raysan5:
@masterex1000 Thanks for the improvement!

PR #1627, opened 2021-03-02
replace vraylib link

waotzi:
the vraylib binding from  MajorHard doesn't seem to be getting any updates, so maybe replace it with the working fork?


raysan5:
@waotzi ok, nice

PR #1629, opened 2021-03-03
Removing +1 on comparing extesnions because this way it checked the f…

object71:
…ile extension without the . against the file extension with the . resulting in always false.

@raysan5 This is due to recent commit "bcc4418f" which modifies the function for getting the file extension to return the extesnion with the "." For me this resulted in models not being loaded and loading only cubes


raysan5:
@object71 thanks for the fix!

PR #1632, opened 2021-03-04
Optimize DrawTexturePro and DrawRectanglePro transformations

ChrisDill:
- Add check so rotation is only applied if rotation != 0.0f.
- Replace matrix usage by calculating the vertex data directly.

### DIFF
--- a/src/shapes.c
+++ b/src/shapes.c
@@ -119,11 +119,11 @@ void DrawLineEx(Vector2 startPos, Vector2 endPos, float thick, Color color)
     Vector2 delta = {endPos.x-startPos.x, endPos.y-startPos.y};
     float   length = sqrtf(delta.x*delta.x + delta.y*delta.y);
 
-    if (length > 0  &&  thick > 0) 
+    if (length > 0  &&  thick > 0)
     {
         float   scale = thick/(2*length);
         Vector2 radius = {-scale*delta.y, scale*delta.x};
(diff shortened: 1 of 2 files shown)


src/shapes.c
```diff
@@ -619,31 +619,60 @@ void DrawRectanglePro(Rectangle rec, Vector2 origin, float rotation, Color color
...
+    Vector2 br = { 0 };
+    Vector2 tr = { 0 };
+    Vector2 tl = { 0 };
 
-    rlPushMatrix();
-        rlTranslatef(rec.x, rec.y, 0.0f);
-        rlRotatef(rotation, 0.0f, 0.0f, 1.0f);
-        rlTranslatef(-origin.x, -origin.y, 0.0f);
+    // Only calculate rotation if needed
+    if (rotation == 0.0f)
```
raysan5, line 628: Be careful, this check could result in false positives due to float resolution issues.

raysan5: Not sure how to solve it... I did a bit of research and it seems that the specific case for 0.0f should work ok... but no clear answer on this topic.

ChrisDill: It should work fine for functions that pass the rotation as 0.0f internally such as DrawTexture though I am not sure about situations where the user passes in rotation that is very close to 0.0f.

raysan5:
@ChrisDill Excuse me for not merging yet, I'd like to do some tests first.

raysan5:
@ChrisDill Thanks for the improvement! I did some tests and the gain is considerable! :)

PR #1635, opened 2021-03-06
Gltf model and animations exploring problems

### LINKED ISSUE #1623: [Models] GLTF import issues
I have identified a few problems with the GLTF file reader that cause problems when loading animations.

It appears that bone IDs are only assigned when a mesh primitive has the cgltf_attribute_type_joints attribute.
In this file, the primitives do not have this attribute, but instead each node in the bone tree directly references a specific mesh ID.
[PrototypePete.zip](https://github.com/raysan5/raylib/files/6058927/PrototypePete.zip)

The animation system also assumes that all mesh vertices are pre-transformed to align with the binding pose. In this model this is also not true. The meshes are at the origin and are expecting to be transformed by the bone that references them.

Given these assumptions, this means that no bones are ever assigned to the models, nothing is transformed, and the mesh is drawn incorrectly.

A Raylib rendering of a similar model
image

A GLTF Viewer image for reference
image

I do not know a lot about the animation system, but it seems that there needs to be a pass during read that assigns bones and transforms vertices of the meshes that the bones directly reference.

I was able to confirm my suspicions by doing a test.
First I added a quick and dirty bone assignment after each primitive is read in, checking to see if it has no bones, and if a bone references it. I don't believe this is the correct fix, but it at least gets me valid bone IDs for this file.

    if (model.meshes[primitiveIndex].boneIds == NULL && data->nodes_count > 0) 
                    {  
                        for (int nodeId = 0; nodeId < data->nodes_count; nodeId++)  
                        {  
                            if (data->nodes[nodeId].mesh == &(data->meshes[i]))  
                            {  
                                model.meshes[primitiveIndex].boneIds  = RL_CALLOC(4 * model.meshes[primitiveIndex].vertexCount, sizeof(int));  
                            model.meshes[primitiveIndex].boneWeights = RL_CALLOC(4 * model.meshes[primitiveIndex].vertexCount, sizeof(float));  
                            for (int b = 0; b < 4 * model.meshes[primitiveIndex].vertexCount; b += 4)  
                            {  
                                model.meshes[primitiveIndex].boneIds[b] = nodeId;  
                                model.meshes[primitiveIndex].boneWeights[b] = 1.0f;  
                            }  
                        }  
                    }  
                }  

Next I created a fake animation frame that is a copy of the bind pose, and then I cleared the bind pose transformatio, so the bind pose is only applied once.

    Model character = LoadModel("resources/PrototypePete.gltf");
    int animCount = 0;
    ModelAnimation anim;
    anim.boneCount = character.boneCount;
    anim.bones = character.bones;
    anim.frameCount = 1;
    anim.framePoses = MemAlloc(1 * sizeof(Transform*));
    anim.framePoses[0] = MemAlloc(1 * character.boneCount * sizeof(Transform));
    for (int i = 0; i < character.boneCount; i++)
    {
        anim.framePoses[0][i].translation = character.bindPose[i].translation;
        anim.framePoses[0][i].rotation = character.bindPose[i].rotation;
        anim.framePoses[0][i].scale = character.bindPose[i].scale;
        character.bindPose[i].translation = Vector3Zero();
        character.bindPose[i].rotation = QuaternionIdentity();
     //   character.bindPose[i].scale = Vector3Zero();
    }
    UpdateModelAnimation(character, anim, 0);

With this in place, the model does render correctly, so I believe that I have at least identified where the problem is.
image

object71:
@raysan5 @JeffM2501 Please take a look. I've fixed some of the problems and added more models from the open source gltf library for testing. It is working with:

- Prototype Pete (the model Jeff provided)
- Rigged Figure (I noticed that the animation is a bit off compared to other viewers)
- Avocado
- Box Animated

It is **not working** for:

- Animated Morph cube (cube loads morph animation is unsupported)
- Gearbox Assembly (will be left as a todo for now and I will open up a new issue)
- Animated Triangle (will be left as a todo for now and I will open up a new issue)

Fixes #1623


object71:
Fixes #1623

object71:
@raysan5 I also fix 2 issues in this pull request related to animation that I noticed when I loaded more models. One is that I didn't set the animation count out variable and the second one was related to a zero division check.

raysan5:
@object71 thanks for the review! I see there are some empty files in the PR, they probably shouldn't be added...

object71:
@raysan5 I added models that all need to be working for a good gltf support. They can be tested in examples.

raysan5:
@object71 Ok, merging!

PR #1636, opened 2021-03-07
[core] minor fixes to Toggle Fullscreen

JeffM2501:
This PR makes some small fixes to ToggleFullscreen to make it work in a more understandable manner.

- Always try to restore vsync
- Use the internal  GetCurrentMonitor instead of glfw so it picks up the current monitor corner.

Note that this does not change how ToggleFullscreen handles resolution. It will still attempts to change the display resolution to match the window size.

Users that want to toggle fullscren to the monitor size should use code similar to this.

image


JeffM2501:
I have made an example that may go well with these changes to show people how to toggle fullscreen and not reset the monitor resolution.

https://gist.github.com/JeffM2501/6e4630a0e34c0c7dddf066f7192e342d

raysan5:

Issue #1241, opened 2020-05-10
[core] ToggleFullscreen() on Web


raysan5:
Today I've been struggling for hours trying to support `ToggleFullscreen()` on Web, it seems emscripten/browsers mechanisms changed and the previous `Module.requestFullscreen(false, false)` call does not automatically strech the image to fullscreen...

I tried to use the Emscripten provided calls for this:
```c
emscripten_request_fullscreen("#canvas", false);
emscripten_request_fullscreen_strategy("#canvas", false, &strategy);
emscripten_enter_soft_fullscreen("#canvas", &strategy);
```
but despite trying multiple `EmscriptenFullscreenStrategy` parameters... it doesn't work properly... maybe I can't understand how configuration works or it just does not work as expected... I don't know...

I'm going to label this issue asking for help... but the truth is that I don't expect anybody helping with this... it took me hours... long hours on a horrible printf-debug console-web environment... trying to figure out how things work... and getting no result... who could be ready for that?

hbiblia:
In which browsers did you try?

object71:
On web the request fullscreen function is not guaranteed to succeed. It needs to be in response to a UI event directly (not polled at some point).

raysan5:
@object71 Thanks for your answer! Yeah, that's it... I'll try to give it another go...

object71:
Glad I could help. Wasn't sure what your test setup is as in the raylib examples the button was outside the canvas. I am just guessing as this could be the most obvious problem :)

raysan5:
I've been doing some more test and it seems it works on some situation... it really seems browser dependant... I'll investigate it further after raylib 3.5 release...

object71:
I would suggest to drop the feature in raylib itself and leave up to the developers to do it themselves in the browser. You may implement soft fullscreen where you expand the html canvas to full width and height with a high 3nough z index.

funcelot:
> I would suggest to drop the feature in raylib itself and leave up to the developers to do it themselves in the browser. You may implement soft fullscreen where you expand the html canvas to full width and height with a high 3nough z index.

better approach is to warn in console log mechanism as Chrome does.

funcelot:
> I would suggest to drop the feature in raylib itself and leave up to the developers to do it themselves in the browser. You may implement soft fullscreen where you expand the html canvas to full width and height with a high 3nough z index.

[...]

PR #1637, opened 2021-03-07
Fix fullscreen resolution

### LINKED ISSUE #1294: [core] ToggleFullscreen() could change screen width/height

object71:
Fixes #1294
Uses not yet submitted pull request of @JeffM2501 - #1636 

Modified how fullscreen gets toggled.

- Removed the unsetting and setting of the resize callback function. Instead of that I moved the fullscreen flag setting into a more correct place before setting the fullscreen so that this flag can be used in the callback.
- In the resize callback now window size is only set when it is not fullscreen resulting in preserving the window size.
- When going fullscreen the larges resolution is used so that there are no problems of the type when you minimize the window you cannot use anything else in your desktop because the resolution might be too low. If a low res effect is desired one should use render texture (this is the approach all game engines use).

@raysan5 This should yield the required results.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1056,7 +1056,7 @@ void ToggleFullscreen(void)
         CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
         
         const GLFWvidmode *mode = glfwGetVideoMode(monitor);
-        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, mode->width, mode->height, 0);
+        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, mode->width, mode->height, mode->refreshRate);
     }
     else
     {
--- a/src/core.c
+++ b/src/core.c
@@ -1055,8 +1055,7 @@ void ToggleFullscreen(void)
         CORE.Window.fullscreen = true;          // Toggle fullscreen flag
         CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
         
-        const GLFWvidmode *mode = glfwGetVideoMode(monitor);
-        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, mode->width, mode->height, mode->refreshRate);
+        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
     }
     else
     {
--- a/src/core.c
+++ b/src/core.c
@@ -1048,9 +1048,7 @@ void ToggleFullscreen(void)
             CORE.Window.fullscreen = false;          // Toggle fullscreen flag
             CORE.Window.flags &= ~FLAG_FULLSCREEN_MODE;
             
-            int modesCount = 0;
-            const GLFWvidmode *modes = glfwGetVideoModes(glfwGetPrimaryMonitor(), &modesCount);
-            glfwSetWindowMonitor(CORE.Window.handle, glfwGetPrimaryMonitor(), 0, 0, modes[modesCount - 1].width, modes[modesCount - 1].height, GLFW_DONT_CARE);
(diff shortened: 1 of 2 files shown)


JeffM2501:
The main reason I did not do this in my PR was that I was not 100% sure if the intent of the function was to change resolutions or not.

I feel that it is silly to change resolutions on modern displays, so I completely understand the desire for this change.
My issue is that this change removes the ability to change the display resolution. I would add a SetMonitorResolution function to this PR so that the functionality is not lost.

I agree that ToggleFullscreen should work this way, but I would hate for raylib to loose functionality for someone who may have a true need to change monitor resolutions.

object71:
@JeffM2501 Looking at documentation people can just use SetWindowSize after fullscreen is enabled to set the resolution that they want.

object71:
Also if people want to have a low-res effect they can go on for full resolution and just have a texture render target as shown in the example core_window_letterbox which will be rescaled to the new fullscreen size

src/core.c
```diff
@@ -1034,34 +1034,41 @@ void ToggleFullscreen(void)
...
+        
         const GLFWvidmode *mode = glfwGetVideoMode(monitor);
-        glfwSetWindowSizeCallback(CORE.Window.handle, NULL);
-        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
-        glfwSetWindowSizeCallback(CORE.Window.handle, WindowSizeCallback);
-
-        // Try to enable GPU V-Sync, so frames are limited to screen refresh rate (60Hz -> 60 FPS)
-        // NOTE: V-Sync can be enabled by graphic driver configuration
-        if (CORE.Window.flags & FLAG_VSYNC_HINT) glfwSwapInterval(1);
+        glfwSetWindowMonitor(CORE.Window.handle, monitor, 0, 0, mode->width, mode->height, mode->refreshRate);
```
raysan5, line 1059: Actually, fullscreen mode shouldn't change monitor video mode resolution, it should try to find the closest resolution supported by the monitor to the user requested one.

raysan5:
@object71 Added one comment. Fullscreen switch shouldn't change monitor resolution but try to find the closest monitor resolution available for the user desired one. Actually, that's the problematic part, for example:

```
User game resolution:  800x450 (16:9)
Fullscreen toggle resolution selected:  800x600 (4:3) - the closest one to a supported monitor resolution.
```

What should happen in that case is that we keep using a 800x450 "game" resolution but add an offsetY with black bars to the rendering... `SetupFramebuffer()` tries to do that...

For user selected "standard" resolutions (1280x720, 1920x1080...), fullscreen works as expected

Probably the best option at this point is just moving that part of the work to user side and let the user create and manage a custom `RenderTexture` like in `core_window_letterbox` example.

Another issue also related to fullscreen is https://github.com/raysan5/raylib/issues/1241 but that's related to browser behaviour and security concerns while switching fullscreen from application side...

object71:
@raysan5 So do I do something in this PR? Remove the setting fullscreen with the max video mode and use the closest to the initial window size?

raysan5:
@object71 Yes, we can just keep the desired user resolution and let the monitor respond to it.

object71:
Okay, resolving that in a moment

raysan5:
@object71 Nice, keeping an eye! I'll merge when ready! :)

object71:
@object71 Done

raysan5:
@object71 Merged!

Issue #1523, opened 2021-01-10
raylib 3.6 roadmap


raysan5:
Here it is my wish list for next raylib version, this time I'll try to keep it smaller and release it earlier, let's see if possible... Feel free to comment here about proposed improvements or, if you have some specific requirement/feature, feel free to comment on [raylib 2021 wishlist Discussion](https://github.com/raysan5/raylib/discussions/1502).

- [x] `core`: Review: `GetFileExtension()` vs `IsFileExtension()` inconsistencies. **_BREAKING CHANGE_**
- [x] `core`: RPI4: Review `Selected DRM connector mode`, it distorts the screen returning a display-supported resolution.
- [x] `core`: Support gamepad mapping/configuration #1506
- [x] `core`: Review high-dpi support... #1086 #1510
- `rlgl`: Consider removing `Texture2D` struct requirement for the low-level API, probably texture id is enough.
- [x] `rlgl`: Redesigned VR simulator to give more control to the user over the fbo and distortion shader
- [x] `rlgl`: Expose the new custom batching system: `rlLoadRenderBatch()`, `rlUpdateRenderBatch()`, `rlDrawRenderBatch()`...
- [x] `rlgl`: Review shader functions naming to follow `rl*()` convention? **_BREAKING CHANGE_**
- [x] `shapes`: Review `DrawRectangleRounded()` performance
- [x] `textures`: Review `DrawTextureTiled()` performance
- `textures`: Allow `RenderTexture.depth` to be `Texture2D` instead of `RenderBuffer` (default)
- `text`: Improve bitmap fonts loading options: `charOrder`, `margin`, `padding`, `backColor`, `keyColor`...
- `text`: Redesign `DrawFPS()` and `DrawGrid()` to support `Color` parameter like other draw functions. **_BREAKING CHANGE_**
- `models`: Break Model loading API into separate functions... #1501 **_BREAKING CHANGE_**
- [x] `models`: Review: `LoadModelAnimations()` requires `UnloadModelAnimations()`
- [x] `models`: Support GLTF animations! #1006
- `raudio`: Provide ring-buffer system, updatable by user -> raw audio management
- [x] `raudio`: Review/redesign (and maybe remove) multi-channel system
- [x] `raudio`: Support `Music` streaming from memory
- [x] `utils`: Expose API callbacks for memory allocation and file access. #1329 #1501 #1521
- [x] `utils`: Review `TraceLog()`, TRACELOGD macro never runs due to logLevel issues
[...]

PR #1648, opened 2021-03-14
Implement UnloadModelAnimations

object71:
@raysan5 Is this what you meant by UnloadModelAnimations for #1523 . 
We don't have a structure that you can pass in with the count of animations so it must go with a count parameter. Else we would need a breaking change to return a structure with ModelAnimations array and count.


raysan5:
@object71 Yes, that's it. I was just waiting to add it in case of finding a better alternative but it would probably require some internal redesign. It's ok for now.

PR #1654, opened 2021-03-18
raylib-php && raylib-phpcpp support 3.5

joseph-montanez:
Both PHP extensions support RayLib 3.5


raysan5:
@joseph-montanez Great! Thanks for the update! :D

PR #1656, opened 2021-03-19
Improved accuracy for DrawRing and DrawCircleSector angles

frithrah:
Here's a nice simple one. The angle params for DrawRing and DrawCircleSector were ints so precise fitting wasn't possible. I've changed them to floats which solves the problem - they get converted to float in the functions anyway. Related examples updated and checked. Good to go.


raysan5:
@frithrah thanks for the improvement.

PR #1659, opened 2021-03-20
Possible fix for macOS windows opening with the wrong viewport size

BigZaphod:
This appears to fix the issue with macOS windows opening with the wrong viewport size/scale by ensuring that the open gl viewport is set in the coordinate space of the underlying window which may have a different scale factor. I do not know if this works on any other platforms, but it seems like it should.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -4258,10 +4258,12 @@ static void SetupViewport(int width, int height)
     CORE.Window.render.width = width;
     CORE.Window.render.height = height;
 
-    // Set viewport width and height
-    // NOTE: We consider render size and offset in case black bars are required and
-    // render area does not match full display area (this situation is only applicable on fullscreen mode)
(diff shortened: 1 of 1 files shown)


raysan5:
@BigZaphod thanks, I'll review it. Maybe some users don't want the Window scaled on HighDPI...

BigZaphod:
The `FLAG_WINDOW_HIGHDPI` doesn't seem to have any effect for me on macOS when I set it. I'm not sure what the intent of the flag is, really, but setting it or not seems to result in no change that I'm noticing. (I must point out that this is literally the first time I've even used Raylib, so I'm not really in a great position to suggest a change like this and feel a bit nervous about even sending it - but it does seem to fix the problem on my end for the examples, anyway!)

raysan5:
@BigZaphod It breaks the rendering on Windows platforms, just pushed a commit with a fix.

PR #1660, opened 2021-03-21
[AUDIO] Use device native sample rates

JeffM2501:
Currently the audio device is forced to init at 44.1k and is only configuratable by a compile time config flag. This will not work well as not all devices support the same sample rates.

This PR forces MinAudio to init with a sample rate of 0 and then reads the effective sample rate from the device after it has been started.  Then this sample rate is used as the target for all sounds to minimize resampling during playback.

This PR changes no APIs, removes a config define and enhances playback quality.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -195,7 +195,6 @@
 //------------------------------------------------------------------------------------
 #define AUDIO_DEVICE_FORMAT    ma_format_f32    // Device output format (miniaudio: float-32bit)
 #define AUDIO_DEVICE_CHANNELS              2    // Device output channels: stereo
-#define AUDIO_DEVICE_SAMPLE_RATE       44100    // Device output sample rate
 
 #define DEFAULT_AUDIO_BUFFER_SIZE       4096    // Default audio buffer size for streaming
 #define MAX_AUDIO_BUFFER_POOL_CHANNELS    16    // Maximum number of audio pool channels
--- a/src/config.h
+++ b/src/config.h
@@ -197,7 +197,6 @@
 #define AUDIO_DEVICE_CHANNELS              2    // Device output channels: stereo
 #define AUDIO_DEVICE_SAMPLE_RATE           0    // Device sample rate (device default)
 
-#define DEFAULT_AUDIO_BUFFER_SIZE       4096    // Default audio buffer size for streaming
 #define MAX_AUDIO_BUFFER_POOL_CHANNELS    16    // Maximum number of audio pool channels
 
 //------------------------------------------------------------------------------------
--- a/src/config.h
+++ b/src/config.h
@@ -195,6 +195,7 @@
 //------------------------------------------------------------------------------------
 #define AUDIO_DEVICE_FORMAT    ma_format_f32    // Device output format (miniaudio: float-32bit)
 #define AUDIO_DEVICE_CHANNELS              2    // Device output channels: stereo
+#define AUDIO_DEVICE_SAMPLE_RATE           0    // Device sample rate (device default)
 
 #define DEFAULT_AUDIO_BUFFER_SIZE       4096    // Default audio buffer size for streaming
 #define MAX_AUDIO_BUFFER_POOL_CHANNELS    16    // Maximum number of audio pool channels
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -256,9 +256,8 @@ typedef struct tagBITMAPINFOHEADER {
 #ifndef AUDIO_DEVICE_CHANNELS
     #define AUDIO_DEVICE_CHANNELS              2    // Device output channels: stereo
 #endif
-#ifndef AUDIO_DEVICE_SAMPLE_RATE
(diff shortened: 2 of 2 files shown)


jmorel33:
This fix solves the situation with the AUDIO device sample rate stuck at 44.1khz while the hardware is playing at a different sample rate, in my case 96khz.  MiniAudio auto detects sample rate when send a value of 0 as JeffM explained to me, and his fix does exactly let the MiniAudio auto detect the soundcard's sample rate, which is great!

```INFO: FBO: [ID 1] Framebuffer object created successfully
INFO: AUDIO: Device initialized successfully
INFO:     > Backend:       miniaudio / WASAPI
INFO:     > Format:        32-bit IEEE Floating Point -> 32-bit IEEE Floating Point
INFO:     > Channels:      2 -> 2
INFO:     > Sample rate:   96000 -> 96000
INFO:     > Periods size:  2880
INFO: STREAM: Initialized successfully (48000 Hz, 16 bit, Stereo)
INFO: FILEIO: [.xm] Music memory successfully loaded:
INFO:     > Total samples: 3290112
INFO:     > Sample rate:   48000 Hz
INFO:     > Sample size:   16 bits
INFO:     > Channels:      2 (Stereo)
INFO: STREAM: Initialized successfully (48000 Hz, 32 bit, Stereo)
INFO: FILEIO: [.mp3] Music memory successfully loaded:
INFO:     > Total samples: 723456
INFO:     > Sample rate:   48000 Hz
INFO:     > Sample size:   32 bits
INFO:     > Channels:      2 (Stereo)
INFO: FBO: [ID 1] Unloaded framebuffer from VRAM (GPU)```

JeffM2501:
I'm not 100% happy with keeping AUDIO_DEVICE_SAMPLE_RATE, but it was just easy to do.

1) It can be converted to a simple state variable
2) It can be turned back into a define but defaulted to 0, and the resulting device state read and used everywhere. I don't think anyone is going to NEED to force the output sample rate, but this PR does remove that ability. If it is a need then I can make the changes to allow it.

raysan5:
@JeffM2501 If it's kept as state variable (1.) it should be added to `AudioData` context, not sure if anyone would require to change it... if (2.), maybe it can just be removed and  use `AUDIO.System.device.sampleRate` all around.

JeffM2501:
> @JeffM2501 If it's kept as state variable (1.) it should be added to `AudioData` context, not sure if anyone would require to change it... if (2.), maybe it can just be removed and use `AUDIO.System.device.sampleRate` all around.

I'll work on that today, I think that will be for the best.

JeffM2501:
It's cleaned up now.

jmorel33 (changes requested):
This is a nice fix, but the ```LoadMusicStreamFromMemory``` bit still loads XM's at 48khz hard coded.  Could it do the same as ```LoadMusicStream``` for the XM meaning not bound to a specific frequency but instead to the device?

JeffM2501:
> This is a nice fix, but the `LoadMusicStreamFromMemory` bit still loads XM's at 48khz hard coded. Could it do the same as `LoadMusicStream` for the XM meaning not bound to a specific frequency but instead to the device

That has been addressed.

jmorel33:
> > This is a nice fix, but the `LoadMusicStreamFromMemory` bit still loads XM's at 48khz hard coded. Could it do the same as `LoadMusicStream` for the XM meaning not bound to a specific frequency but instead to the device
> 
> That has been addressed.

works all very well 
```INFO: TEXTURE: [ID 21] Texture created successfully (512x216 - 1 mipmaps)
INFO: TEXTURE: [ID 1] Depth renderbuffer loaded successfully (16 bits)
INFO: FBO: [ID 1] Framebuffer object created successfully
INFO: AUDIO: Device initialized successfully
INFO:     > Backend:       miniaudio / WASAPI
INFO:     > Format:        32-bit IEEE Floating Point -> 32-bit IEEE Floating Point
INFO:     > Channels:      2 -> 2
INFO:     > Sample rate:   96000 -> 96000
INFO:     > Periods size:  2880
INFO: STREAM: Initialized successfully (96000 Hz, 16 bit, Stereo)
INFO: FILEIO: [.xm] Music memory successfully loaded:
INFO:     > Total samples: -1
INFO:     > Sample rate:   96000 Hz
INFO:     > Sample size:   16 bits
INFO:     > Channels:      2 (Stereo)
INFO: STREAM: Initialized successfully (48000 Hz, 32 bit, Stereo)
INFO: FILEIO: [.mp3] Music memory successfully loaded:
INFO:     > Total samples: 723456
INFO:     > Sample rate:   48000 Hz
INFO:     > Sample size:   32 bits
INFO:     > Channels:      2 (Stereo)
INFO: FBO: [ID 1] Unloaded framebuffer from VRAM (GPU)
INFO: STREAM: Unloaded audio stream data from RAM
INFO: STREAM: Unloaded audio stream data from RAM
INFO: STREAM: Unloaded audio stream data from RAM
INFO: STREAM: Unloaded audio stream data from RAM```

raysan5:
@JeffM2501 Nice improvement! Thanks! 👍

PR #1665, opened 2021-03-21
[AUDIO] Music Looping enhancements

JeffM2501:
This PR makes some small changes to music looping to make it more consistent across sound types.

It adds functions to get/set the loop state music files, and for XM module files, correctly tells the tracker to loop the sample generator.
XM files are then no longer tracked by sample count, but allowed to loop infinitely using data provided by the tracker.
Because of this the GetMusicTimePlayed function now has special logic for XM files that will use the modulo of the samples played to track the relative time from start regardless of how many loops the tracker has generated.
With this change XM files are no longer brute force restarted for loops, allowing the music file to process the loop as it was written (using a jump table)

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1232,7 +1232,7 @@ Music LoadMusicStream(const char *fileName)
 
             // NOTE: Only stereo is supported for XM
             music.stream = InitAudioStream(AUDIO.System.device.sampleRate, 16, AUDIO_DEVICE_CHANNELS);
-            music.sampleCount = (unsigned int)jar_xm_get_remaining_samples(ctxXm) * 2;    // 2 channels
+            music.sampleCount = (unsigned int)jar_xm_get_remaining_samples(ctxXm)*2;    // 2 channels
             music.looping = true;   // Looping enabled by default
             jar_xm_reset(ctxXm);   // make sure we start at the beginning of the song
             musicLoaded = true;
@@ -1322,8 +1322,7 @@ Music LoadMusicStreamFromMemory(const char *fileType, unsigned char* data, int d
         if (success)
         {
             int sampleSize = ctxWav->bitsPerSample;
-            if (ctxWav->bitsPerSample == 24)
-                sampleSize = 16;   // Forcing conversion to s16 on UpdateMusicStream()
+            if (ctxWav->bitsPerSample == 24) sampleSize = 16;   // Forcing conversion to s16 on UpdateMusicStream()
 
             music.stream = InitAudioStream(ctxWav->sampleRate, sampleSize, ctxWav->channels);
             music.sampleCount = (unsigned int)ctxWav->totalPCMFrameCount*ctxWav->channels;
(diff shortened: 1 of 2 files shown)


src/raudio.c
```diff
@@ -1697,6 +1711,24 @@ void SetMusicPitch(Music music, float pitch)
 {
     SetAudioBufferPitch(music.stream.buffer, pitch);
 }
+// Set the music to loop or not
+void SetMusicLooping(Music *music, bool loop)
+{
+    music->looping = loop;
+
+#if defined(SUPPORT_FILEFORMAT_XM)
```
raysan5, line 1719: I'd prefer to avoid `SetMusicLooping()` and `GetMusicLooping()`, that's the reason to allow `music.looping` direct access, like other data structures. Actually, those two functions are only required for the XM module... I'd prefer some other way to set this, maybe on the `UpdateMusicStream()` call?

src/raudio.c
```diff
@@ -1232,7 +1232,7 @@ Music LoadMusicStream(const char *fileName)
 
             // NOTE: Only stereo is supported for XM
             music.stream = InitAudioStream(AUDIO.System.device.sampleRate, 16, AUDIO_DEVICE_CHANNELS);
-            music.sampleCount = (unsigned int)jar_xm_get_remaining_samples(ctxXm)*2;    // 2 channels
+            music.sampleCount = (unsigned int)jar_xm_get_remaining_samples(ctxXm) * 2;    // 2 channels
```
raysan5, line 1235: Formatting: I don't use spaces between `*` and `/`.

src/raudio.c
```diff
@@ -1322,7 +1322,8 @@ Music LoadMusicStreamFromMemory(const char *fileType, unsigned char* data, int d
         if (success)
         {
             int sampleSize = ctxWav->bitsPerSample;
-            if (ctxWav->bitsPerSample == 24) sampleSize = 16;   // Forcing conversion to s16 on UpdateMusicStream()
+            if (ctxWav->bitsPerSample == 24)
```
raysan5, line 1325: Formatting: I try to keep one-line statements after the check.

src/raudio.c
```diff
@@ -1575,7 +1576,8 @@ void StopMusicStream(Music music)
 // Update (re-fill) music buffers if data already processed
 void UpdateMusicStream(Music music)
 {
-    if (music.stream.buffer == NULL) return;
+    if (music.stream.buffer == NULL)
```
raysan5, line 1579: Formatting... there are several similar issues following...

src/raylib.h
```diff
@@ -1492,8 +1492,10 @@ RLAPI void ResumeMusicStream(Music music);                            // Resume
 RLAPI bool IsMusicPlaying(Music music);                               // Check if music is playing
 RLAPI void SetMusicVolume(Music music, float volume);                 // Set volume for music (1.0 is max level)
 RLAPI void SetMusicPitch(Music music, float pitch);                   // Set pitch for a music (1.0 is base level)
+RLAPI void SetMusicLooping(Music *music, bool loop);                   // Set the music to loop or not
```
raysan5, line 1495: As commented above, I prefer to avoid those functions and allow the direct access to `music.looping`, like other data structures in raylib.

raysan5:
@JeffM2501 Just reviewed the PR, added some comments.

JeffM2501:
I've removed the loop functions and made XM files set the tracker loop function in update.
I've also cleaned up the formatting, that'll teach me to check my IDE settings before using an old computer :)

raysan5:
@JeffM2501 Ok, thanks for the review and excuse me for being so picky.

PR #1669, opened 2021-03-23
Added DrawTextStyle() and DrawTextStyleEx() for Drawing Styled Text with Markdown

OkiStuff:
Added DrawTextStyle() and DrawTextStyleEx() for Drawing Styled Text with Markdown

This is modified code of https://github.com/NightenDushi/Raylib_DrawTextStyle. I feel like it would be nice to have this built-in instead of as a header. But I 100% understand if you don't want to add it in. It would be a little weird for a framework to have a feature like this. Just feel like because one of the goals of raylib is to be easy-to-use library to enjoy making games this would make it easier to do cool stuff with text.

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -58,6 +58,7 @@
 #include <string.h>         // Required for: strcmp(), strstr(), strcpy(), strncpy() [Used in TextReplace()], sscanf() [Used in LoadBMFont()]
 #include <stdarg.h>         // Required for: va_list, va_start(), vsprintf(), va_end() [Used in TextFormat()]
 #include <ctype.h>          // Requried for: toupper(), tolower() [Used in TextToUpper(), TextToLower()]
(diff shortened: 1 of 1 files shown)


OkiStuff:
after getting some feedback from the community
it would be better to take it down

raysan5:
@OkiStuff Maybe it's a bit specific feature to be added to raylib but undoubtedly it should go into a code example!

vaartis:
At least _some_ support for styling not provided by fonts would be nice, like underline and strikethrough. It's already implemented in the PR, maybe something more scaled down from this could be implemented. Some libraries implement routines to create bold and italic fonts themselves, too (sdl_ttf and SFML at least).

raysan5:
@vaartis All those extras can actually be implemented at user level, no need to be provided by raylib. In any case, an example with those features implemented would be nice.

PR #1671, opened 2021-03-23
Makefile: Fix PATH for other PLATFORM_OS's

Theldus:
Description
----------------
The Raylib root Makefile for `PLATFORM=PLATFORM_WEB` is assumed to run in Windows environments, which makes it incorrectly set the PATH variable, so Linux (and possibly other) environments are not able to find binaries from EMSDK_PATH.

Fix this by checking which PLATFORM_OS and configure PATH accordingly.


raysan5:
@Theldus Nice! Thanks for the review!

PR #1672, opened 2021-03-23
Don't call sqrtf when axis already normalized

dbechrd:
Optimization to avoid unnecessary `sqrt` calls in `MatrixRotate` when `axis` vector is already unit length or the zero vector.


raysan5:
@dbechrd Nice! Thanks for the improvement! :)

PR #1673, opened 2021-03-23
Fix 90 degree bug with DrawTexturePro and DrawRectanglePro

ChrisDill:
- The vertices did not map to where I expected causing rotation to be
off by 90 degrees. I reorganized the vertices making it easier to
reason about which fixes this.
- The order for drawing is now topLeft, bottomLeft, bottomRight,
topRight.


raysan5:
@ChrisDill thanks for the review!

PR #1674, opened 2021-03-23
Fix activeTextureId to have MAX_BATCH_ACTIVE_TEXTURES elements instead of the hardcoded 4

REDVM:
Thanks for this nice project!

`RLGL.State.activeTextureId` is used in for-loop like that:
```C
for (int i = 0; i < MAX_BATCH_ACTIVE_TEXTURES; i++)
{
    if (RLGL.State.activeTextureId[i] > 0)
    {
        glActiveTexture(GL_TEXTURE0 + 1 + i);
        glBindTexture(GL_TEXTURE_2D, RLGL.State.activeTextureId[i]);
    }
}
```

So the size of `RLGL.State.activeTextureId` should be MAX_BATCH_ACTIVE_TEXTURES.

I'm using raylib to run a Neural Network on the Raspberry Pi GPU using GLSL shaders. So I needed to increase MAX_BATCH_ACTIVE_TEXTURES.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -825,7 +825,7 @@ typedef struct rlglData {
         Texture2D shapesTexture;            // Texture used on shapes drawing (usually a white pixel)
         Rectangle shapesTextureRec;         // Texture source rectangle used on shapes drawing
         unsigned int defaultTextureId;      // Default texture used on shapes/poly drawing (required by shader)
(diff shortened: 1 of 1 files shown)


raysan5:
@REDVM Nice! Thanks for the fix! 👍😄

PR #1676, opened 2021-03-24
[Text] Have FPS colors change based on good/bad FPS values

JeffM2501:
This PR makes the default FPS display change colors based on the actual FPS values.
FPS >= 30 = Green (good FPS)
FPS < 30 and FPS >= 15  = Orange (not great FPS)
FPS < 15 = Red (Bad FPS)


raysan5:
@JeffM2501 Thanks! This is a great improvement! I like it! :D

PR #1678, opened 2021-03-25
Added animation and keyboard controls to rlgl mesh Instancing example

moliad:
Added a simple jump animation to the instanced mesh.
Added keyboard controls to modify all parts of the animation (speed, height, etc.)
Added text to describe controls.


raysan5:
@moliad Thanks for the improvement!

PR #1683, opened 2021-03-26
Preserve floating point values in DrawRectangleLinesEx

dbechrd:
The `Rectangle` version of `DrawRectangle` is `DrawRectangleRec`, and it preserves non-integral floating point values that are passed to it without truncating them.

The `Rectangle` version of `DrawRectangleLines` is `DrawRectangleLinesEx`, and it casts all of the floating point values passed to it  to integers before forwarding them to `DrawRectangleRec` (which is perfectly capable of handling floats, as noted above).

This causes a bug in the public API where it's impossible to outline a rectangle drawn with `DrawRectangleRec` with lines drawn by `DrawRectangleLinesEx` due to the truncation. For example, the outline will jitter around when drawn on top of a smoothly animated rectangle.

The fix is to simply preserve the floating point precision. I've left the intentional truncation in the case when lineWidth > rectangle width or height for consistency with previous behavior in that degenerate case.

P.S. Apologies for the trailing whitespace noise in the diff (my editor is configured to clean it up automatically), but it was few enough that it seemed fine to let it though. Will remove upon request.

### DIFF
--- a/src/shapes.c
+++ b/src/shapes.c
@@ -760,10 +760,26 @@ void DrawRectangleLinesEx(Rectangle rec, int lineThick, Color color)
         else if (rec.width < rec.height) lineThick = (int)rec.width/2;
     }
 
-    DrawRectangle( (int)rec.x, (int)rec.y, (int)rec.width, lineThick, color);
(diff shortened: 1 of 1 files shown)


raysan5:
@dbechrd thanks! nice improvement! 👍😄

PR #1687, opened 2021-03-27
updates tiny obj loader to latest version, fixed multi materials again...

chriscamacho:


chriscamacho:
the internal buffers have changed in the way they are laid out, I can't see what it is thats changed, but some models will work some won't - I'm withdrawing this PR

PR #1688, opened 2021-03-27
GetCollisionRayMesh makes use of triangle count

car1l2os:
###Reason for the change
The current code is wrong and will cause bugged ray-mesh detections in meshes with shared vertices. 

###Example
If we make use of the GetMeshCube function (where we have 12 triangles but 24 vertices)  there's going to be 4 triangles (2 faces) which are not going to be taken into account for the collision check.

###Comments
I'm not sure on which case the mesh will be intialized without a triangle number but that (in my opinion) should be a problem of the code initializing that mesh and it should not be remedied in the collision check function.


raysan5:
@car1l2os thanks for the improvement! Actually, some days I noticed that issue while drawing indexed meshes... thanks!

PR #1696, opened 2021-04-02
New addition FadeBetween

frithrah:
Simple function FadeBetween to interpolate between two colours. Plus a small demo in an existing example.

### DIFF
--- a/examples/shapes/shapes_rectangle_scaling.c
+++ b/examples/shapes/shapes_rectangle_scaling.c
@@ -12,6 +12,7 @@
 ********************************************************************************************/
 
 #include "raylib.h"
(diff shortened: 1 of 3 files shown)


raysan5:
@frithrah Thanks for the proposed function but I'm not adding it for the moment, I prefer to let the user choose desired interpolation. Actually, raylib already provides [`easings.h`](https://github.com/raysan5/raylib/blob/master/src/easings.h) library to allow custom interpolations between values.

frithrah:
Okay but this is just a helper for the colour blending math, like Fade. You can pass any of the easings results into this new function as the t parameter. I think beginners would appreciate it.

PR #1697, opened 2021-04-02
Make SaveFile* callbacks return a boolean

RobLoach:
The `SaveFileText` and `SaveFileData` callbacks returned `void`, which loses out on whether or not the save was successful. If a save failed, the callback would still return `true`. This change makes it so that those callbacks report the output directly so that you correctly know when a save failed.


raysan5:
@RobLoach thanks for the improvement! 👍😄

PR #1701, opened 2021-04-05
Update to the xm player routine and support functionalities

jmorel33:
Hello RayLib, this is a long in the tooth refresh to the XM module player.

It fixes the majority of all of it's issues,

Some of the changes done were based on original specifications as well as observations made from the ModPlug Library.

It is in no means perfect, but a long overdue refresh to the capability.


raysan5:
@jmorel33 Very nice! Thanks for the review!

PR #1707, opened 2021-04-06
Circle sector minimum segments

frithrah:
There's a hard-coded restriction on the minimum number of segments in DrawRing, DrawCircleSector etc. It's a problem when you need to draw small sectors, e.g. 10 or 20 degrees. These only require 1, 2 or 3 segments (depending on radius) but they're forced to use 4 which is significantly more resource-hungry than necessary.

This change sets the minimum number of segments based on the degree range. I have also updated related examples.

I present this solution but actually prefer having no logic other than the check for zero to trigger the auto calculation. What do you prefer? I'm happy to rework it.


raysan5:
@frithrah thanks for the improvements! 👍😄

PR #1710, opened 2021-04-06
added import in order for alloca to work on macos

Saikyun:
without this modification, I couldn't get raylib running on macos catalina due to alloca being undefined


raysan5:
@Saikyun oh, thanks for the improvement!

Saikyun:
Happy to contribute to the great raylib. :D

PR #1711, opened 2021-04-06
Fix ImageClearBackground

Ergoold:
The `ImageClearBackground` implementation in current master skips over some pixels in the image.
The fix is simple - the y-coordinate of each cleared pixel was calculated in the loop as `i/dst->height`, but needs to be `i/dst->width`.

```c
#include "raylib.h"

int main(void)
{
    InitWindow(640, 480, "Example");
    Image image = GenImageColor(320, 160, BLUE);
    ImageClearBackground(&image, RED);
    Texture texture = LoadTextureFromImage(image);
    while (!WindowShouldClose())
    {
        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    UnloadTexture(texture);
    CloseWindow();
}
```

Current master:
Before

With fix:
After

### DIFF
--- a/src/textures.c
+++ b/src/textures.c
@@ -2347,7 +2347,7 @@ Rectangle GetImageAlphaBorder(Image image, float threshold)
 // Clear image background with given color
 void ImageClearBackground(Image *dst, Color color)
 {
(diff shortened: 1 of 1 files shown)


raysan5:
@Ergoold Good catch! Thanks for the fix! 👍😄

PR #1715, opened 2021-04-08
Fix off by one bug with GetGamepadAxisCount on PLATFORM_DESKTOP

ChrisDill:
Found testing `core_input_gamepad`. The last axis was not drawn.
I found that `GLFW_GAMEPAD_AXIS_LAST` is defined to the last axis which is 5
not the total number of axes which is 6.

See GLFW docs for reference:
https://www.glfw.org/docs/latest/group__gamepad__axes.html


raysan5:
@ChrisDill Thanks for the fix!

PR #1716, opened 2021-04-09
[AUDIO] XM streams use native sample sizes

JeffM2501:
This PR makes XM streams use audio streams in the native device format, and generate the appropriate sample data when possible. This should reduce CPU usage during streaming.


raysan5:
Thanks @JeffM2501! Nice improvement!

PR #1724, opened 2021-04-15
Extended API to read from Screen at custom position and sizes

trbtm:
I extended the API by two functions

- TakeScreenshotAtPosAndSize <=> TakeScreenshot
- GetScreenDataFromPosAndSize <=> GetScreenData

to allow for a custom position and size of the screenshot or data section taken from the screen / window. The changed API is fully backwards compatible since TakeScreenshot and GetScreenData are simply wrappers around TakeScreenshotAtPosAndSize and GetScreenDataFromPosAndSize.

I didn't implement sanity checks for the position and the dimensions.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2555,8 +2555,14 @@ void SetConfigFlags(unsigned int flags)
 // have their own internal file-systems, to dowload image to user file-system some additional mechanism is required
 void TakeScreenshot(const char *fileName)
 {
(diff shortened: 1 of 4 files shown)


raysan5:
@trbtm Thanks for the review but I'm not adding those functions, they are very user specific and, actually, user can do it on its side if required.

PR #1725, opened 2021-04-17
Update and fix Android Linux makefile sample for template project

itarato:
## Tl;dr

This PR attempts to update the Android-Linux makefile on the template project so it runs with the latest state of tooling and version.

## Changes

- added Android NDK reference in order to use their binaries at a few points.

It seems some linked bins and sources were using local or other folders that either do not contain those binaries or moved in recent updates.

- corrected a few architecture related binary file references to the correct files
- added CC/AR definitions to x86 / x86_86
- swapped the order of zipaligning and signing

With the original order I was getting a lot of aligning errors and digging on error reports it seems zipaligning has to be done on the unsigned apk just before signing

- changed signing from jarsigner to apksigner as the former one is not supported anymore
- minor shell command syntax issues

## Test

I've tested it on latest Android version both emulator and real device, all works.

### DIFF
--- a/templates/simple_game/Makefile.Android.linux
+++ b/templates/simple_game/Makefile.Android.linux
@@ -47,15 +47,16 @@ endif
 # Required path variables
 # no need to define JAVA_HOME , JAVA_BIN in linux environment (binaries are in $PATH)
 #JAVA_HOME              ?= C:/JavaJDK/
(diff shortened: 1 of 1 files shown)


raysan5:
@itarato Nice! Thank you very much for the review and the detailed explanation! 👍😄

itarato:
Pleasure :)

PR #1726, opened 2021-04-18
[examples] Added an example for raylib OpenGL interop

arkanis:
Added an example that combines raylib with plain OpenGL to create a GL_POINTS based particle system. Basically what to do when you need a feature that isn't covered by raylib. It covers two important points:

- Flushing the possibly pending rlgl draw batch
- Getting the current model view projection matrix

There are still some open questions, though:

- Did I miss anything? Is there more that needs to be done when switching between raylib and OpenGL?
- The example is made for the current master. The relevant functions changed from raylib 3.5 (rlDrawRenderBatchActive, rlGetMatrixModelview, rlGetMatrixProjection). So what to write in the "created using raylib VERSION" header? For now I added comments with the corresponding function names in raylib 3.5.
- Maybe the example shouldn't use the locations in shader.locs[...] and instead query them itself. It might be easier to understand and apply to new use cases?
- Maybe the vertex shader should use two attributes instead of packing the period into the z coordinate of the position. But this is more of an OpenGL thing and not really related to raylib.

ps.: I really enjoyed working with raylib. Great library. :grinning:


raysan5:
@arkanis Nice example! As it is a bit tangent to raylib itself, could you move it to [`examples/others`](https://github.com/raysan5/raylib/tree/master/examples/others) and rename it to `opengl_interop.c`?

arkanis:
Done. And updated the Makefile accordingly.

Feel free to delete the preview image as no other example in that folder has a preview image.

raysan5:
@arkanis Great! Thanks for the update! :)

Issue #1587, opened 2021-02-14
[models] IQM bone transform not following parent


nonunknown:
### Issue description

Followed the animation at model module example, but the animation is messed!

### Environment
Windows 10 x64
* OpenGL 3.3
* AMD R7 200 series

### Issue Screenshot

https://user-images.githubusercontent.com/58845030/107877003-51bf8880-6ea8-11eb-96bb-b0335632476a.mp4

### Project Files

[ray.zip](https://github.com/raysan5/raylib/files/5977692/ray.zip)

chriscamacho:
does the loader provide all the bone data?

I'd have a stab at making a vertex shader to do this if ....

A) I had 2-3 different examples of bone animation IQM's
b) The loader does in fact provide all the relevant bone data
C) I've I have the time (I've started a new job) fatigue is an issue!

chriscamacho:
I looked at animations with GLTF as well, there is a definite issue there too, so possibly related, as well as messed up transforms there is an issue with missing bones and the anim length

raysan5:
It seems this issue could be solved by previous PR (not tested), I'm closing it. Please, feel free to reopen it if you test it and it does not work yet.

PR #1728, opened 2021-04-18
Use bone weights for animation

nathants:
When using the [iqm](https://github.com/lsalzman/iqm) exporter for Blender, I noticed animations were not using bone weights. 

This adds bone weight support to `UpdateModelAnimation()`.


raysan5:
@nathants Nice! Thanks for the review! By the way, could this improvement be related to https://github.com/raysan5/raylib/issues/1587?

nathants:
@raysan5 it’s possible. Previously the first of four bones was used at max weight with the rest ignored.

PR #1733, opened 2021-04-25
Fixes for 64 bit typecast warnings

JeffM2501:
A pass at fixing typecast warnings for 64 bit builds in Visual Studio. This is not all of them but another attempt to reduce the number.


raysan5:
@JeffM2501 Thanks for the improvements! 👍😄

PR #1734, opened 2021-04-25
[BUILD] Disable examples that won't build by default from 2019 solution configuration

JeffM2501:
This PR disables the example projects in the Visual Studio 2019 soliton that will not build by default, so that users won't be confused when they get errors on build.


raysan5:

PR #1735, opened 2021-04-25
[Examples] Add a custom mesh to mesh generation example

JeffM2501:
This PR adds a custom mesh to the mesh generation example to show users how to define allocate, define, and upload a mesh from pure code.


raysan5:
@JeffM2501 Thank you very much for the quick example PR. I'm merging it, ready for the release, but I'm probably moving it a separate example later on.

PR #1739, opened 2021-04-27
BINDINGS: raylib-cpp targets raylib 3.7.0

RobLoach:
https://github.com/RobLoach/raylib-cpp/releases/tag/v3.7.0


raysan5:
@RobLoach Nice! Thanks for the update! :D

PR #1740, opened 2021-04-27
Fixed definition of UNCOMPRESSED_R8G8B8A8

Rabios:


raysan5:
@Rabios Thanks for the fix!

PR #1741, opened 2021-04-27
Updated raylua and raylib-boo to raylib 3.7!

Rabios:


raysan5:
@Rabios Nice! Thanks for the update! 👍😄

PR #1748, opened 2021-05-04
Fix for newlines in Makefile

IHideMemesInMySourcecode:
modified  templates/simple_game/Makefile.Android.linux
switched out @echo for @printf to make it actually print newlines instead of literal \n's


raysan5:
@IHideMemesInMySourcecode thanks for the review!

PR #1749, opened 2021-05-05
Unset CORE.Window.ready on CloseWindow

TSnake41:
Window shouldn't be considered ready when CloseWindow has been called.


raysan5:
@TSnake41 Thanks for the improvement!

PR #1757, opened 2021-05-08
check for vao extension

psxdev:
I was updating raylib for vita and i found a little problem. 

Raylib had in older versions a check for vao extension in rlUploadMesh now the code is refactorized in uploadMesh at model.c using calls to rlLoadVertexArray and rlEnableVertexArray.

rlEnableVertexArray has the vao extensin check, but rlLoadVertexArray doesn't so to avoid the crash check for vao extension must be added.

Regards


raysan5:
@psxdev thanks for the fix!

PR #1766, opened 2021-05-12
Update raylib-lua and raylib-wren binding versions

TSnake41:
Those bindings got updated recently.


raysan5:
@TSnake41 Nice! Thanks for the update!

PR #1767, opened 2021-05-12
Removes Whole Program Optimization

Bxil:
This PR was prompted by trying to link the x64 Windows release with LLVM's linker.
Using whole program optimization generates lib files with a different file format that is not compatible with any other linker than the one included with the MSVC version used to create them. This means that the library is incompatible even with different MSVC versions. See (specifically the third point): https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-160#restrictions
I removed whole program optimization from all Visual Studio projects, please tell me if I missed any.
I am not sure how the releases are made, if they get their configuration from elsewhere this has to be changed there as well.
Note that this change will result in an ~15% larger library.


raysan5:
@Bxil Thank you very much for investigating this issue! Actually, the releases are generated using that same project, so, if there is any issue the CD system will tell me.

PR #1768, opened 2021-05-12
Fixed GLTF boneWeights uint32 loading

devoil-g:
boneWeights uint32 were loaded in normals instead of boneWeights.


raysan5:
Wow! Good catch! Thanks for the review!

PR #1772, opened 2021-05-14
Fixed over-allocation of buffer

devoil-g:
output->framePoses[frame] is over-allocated.
framePoses is a 2D array:
- first dimension: frames (allocated l. 4717)
- second dimension: nodes (allocated l. 4731)

Second dimension should be allocated of nodes_count transformations only.


raysan5:
@devoil-g nice! good catch! 👍😄

PR #1774, opened 2021-05-16
Added DrawFPSEx() function with custom FPS colors and fontsize.

NotManyIdeasDev:
This function allows the user to choose the colors to indicate good, medium or bad FPS, and also allows to change the font size.


raysan5:
@NotManyIdeasDev I'm afraid I'm not merging this function, `DrawFPS()` itself should be convenient enough.

PR #1780, opened 2021-05-21
Removes debug information from release builds

Bxil:
This halves the size of the distributed binary in addition to removing some linker warnings when the debug symbol database is not distributed with the binary.


raysan5:
@Bxil Nice! Thanks for the review!

PR #1783, opened 2021-05-23
Fixed CMake project template for MacOS

GoldenbergDaniel:
The CMake project template didn't link to the required libraries for macOS, causing the build to fail. I added calls to include the IOKit, Cocoa, and OpenGL libraries in the CMake template for those who are building on macOS.


raysan5:
@GoldenbergDaniel thanks for the review!

Issue #1787, opened 2021-05-25
[rlgl] MacOS binaries error with "Cocoa: Failed to find service port for display" on Big Sur


benbeshara:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

When running a Raylib application on MacOS 11 using the prebuilt binaries, the window routine produces the error `GLFW: Error: 65544 Description: Cocoa: Failed to find service port for display`

This is due to them being built with a version of GLFW < 3.3.3 (see: https://github.com/glfw/glfw/releases/tag/3.3.3)

Libraries built locally with a contemporary version of GLFW do not present this issue

### Environment

macOS Big Sur on an Apple M1 MacBook Pro

PR #1788, opened 2021-05-25
Bumped minimum GLFW version in cmake

benbeshara:
GLFW versions under 3.3.3 do not support MacOS 11, so I've bumped it in the CMake file

See issue #1787


raysan5:
@benbeshara Thanks for the fix!

PR #1794, opened 2021-05-28
GetDeltaMouse() added.

adricoin2010:
Hello, this is my first Pull Request, sorry if it is not well created.

This pull request adds `GetDeltaMouse()`. It is a simple function that returns the movement of the mouse for each frame.

It can be useful for interfaces, cameras, etc. I need it for a project, and seeing that it was not in raylib I decided to add it.

Greetings.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -434,6 +434,9 @@ typedef struct CoreData {
             Vector2 offset;                 // Mouse offset
             Vector2 scale;                  // Mouse scaling
 
+            Vector2 delta;                  // Mouse movement per frame
+            Vector2 previousPosition;       // Previus frame position
+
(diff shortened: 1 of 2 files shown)


raysan5:
@adricoin2010 I'm afraid I'm not adding this functionality, it can be implemented in user side instead of adding two additional global variables to CORE structure.

raysan5:
@adricoin2010 Just registering mouse previous position for a special need, so `GetDeltaMouse()` can be added if you wish.

adricoin2010:
I would love if you could add it. Do I submit a pull request?

raysan5:
@adricoin2010 feel free to send a PR, `GetMouseDelta()` can calculate it directly, no need for extra variables.

PR #1802, opened 2021-06-01
Update raymath.h

isaac553876299:
Function MoveTowards for vectors didn't handle by 0 division properly, think it is better now. NOTE: not sure of the purpose of 2nd argument in the if() statement.


raysan5:
@isaac553876299 thanks for the review!

PR #1806, opened 2021-06-03
[Examples]Add a split screen example 

JeffM2501:
This PR adds a simple example of doing split-screen or multiple viewports using render textures and two cameras and the same scene rendering.

unknown


raysan5:
@JeffM2501 Very nice example! Thanks! :)

PR #1813, opened 2021-06-06
Fix parsing multiple struct fields defined in one line (Matrix struct)

Deathbat2190:
Fixed a bug that was causing the Matrix fields to be of type `float m0, m4, m8,`. Now the type is parsed correctly and the field name is `m0, m4, m8, m12`.


raysan5:
@Deathbat2190 thanks for the fix. Actually, it should probably parse all the fields separately... it's a special case...

PR #1818, opened 2021-06-09
Fix build issue with cmake after glfw upgrade

object71:
In the new upgrade glfw no longer has a target glfw_objlib to take the sources from and build them in raylib. We directly take the sources from glfw now (which should bring the same meaning).


raysan5:
@object71 thank you very much for the fix!

PR #1820, opened 2021-06-10
Added missing null terminator on Input Box example

autopawn:
Added missing null terminator when adding characters to the string, otherwise garbage values are read (often zeros which are equal to '\0', but not every time).

This error results in random characters appearing in the text box every once in a while:
```
asdfasdf??? ll??
```
It is corrected with the proposed fix.

This problem was observed by @griveralazo.


raysan5:
@autopawn thanks for the review!

PR #1821, opened 2021-06-11
Dot in LoadImageFromMemory is now optional

Bxil:
I have found it quite unintuitive that the dot is required when specifying the image format.

### DIFF
--- a/src/textures.c
+++ b/src/textures.c
@@ -294,37 +294,43 @@ Image LoadImageAnim(const char *fileName, int *frames)
     return image;
 }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Bxil I did a long investigation about using dot or not in raylib when dealing with extensions, the final decision, based on many engines and platforms was to keep the dot and make it mandatory across all functions that require it, to make all of them consistent. Actually `GetFileExtension()` returns the dot to be combined with other functions. `IsFileExtension()` also checks for the dot.

So, for the moment I'll keep the dot.

PR #1823, opened 2021-06-12
Make first person camera movement not reliant on framerate

Y5U:
This makes the SetCameraMode(camera, CAMERA_FIRST_PERSON) mode, no longer reliant on frame rate. 10000, and 60 fps now move at the same speed.


raysan5:
@Y5U I'm not merging this PR for the moment, one of the goals of `camera.h` module is to be not dependent of raylib functionality and this change adds a coupling between modules.

Still, I'm thinking about a better approach to this issue.

PR #1826, opened 2021-06-12
updating raylib python cffi version

electronstudio:


raysan5:
@electronstudio thanks for the update! 👍😄

PR #1827, opened 2021-06-13
static libraries linking on windows, without needed config

maihd:
I found it annoying when starting new visual studio and clang on windows project, and need to add too many inputs for linker.


raysan5:
@maihd Sorry, but I prefer to avoid this pragma approach.

maihd:
Okay.

PR #1830, opened 2021-06-15
add rlLoadMatrixf function

RicoP:
The rlgl API has no function to directly set the active matrix. If you want to do so you have to first call loadidentity followed by multmatrix. Having a rlLoadMatrixf  function makes the API a bit more consistent. There are explicit function called rlSetMatrixXXX  outside the rlgl context but they set either the modelview or projection matrix explicit. rlLoadMatrixf set the current active matrix instead.


raysan5:
@RicoP `glLoadMatrixf()` was removed from modern OpenGL and I prefer to avoid it and force the user to use the explicit `rlSetMatrix*()` fuctions.

PR #1831, opened 2021-06-15
Add support for mat3/4 shader uniforms

JeffM2501:
This PR adds support for matrix shader uniforms to rlg.


raysan5:
@JeffM2501 I reviewed this issue and the reason I did not included those Matrix uniform types is because there is a separate function to update a Matrix uniform on the shader: 

```c
// Set shader value uniform matrix
void rlSetUniformMatrix(int locIndex, Matrix mat)
{
#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
    glUniformMatrix4fv(locIndex, 1, false, MatrixToFloat(mat));
#endif
}
```

To keep it simple I just support the standard 4x4 matrix case (the one defined by raylib).

Supporting the additional types would require extra work for the user: convert the 3x3 or 4x4 matrix into a float non-transposed vector to be provided to `rlSetUniform()` function:
```c
// Set shader value uniform
void rlSetUniform(int locIndex, const void *value, int uniformType, int count)
{
#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
    switch (uniformType)
    {
        case SHADER_UNIFORM_FLOAT: glUniform1fv(locIndex, count, (float *)value); break;
        case SHADER_UNIFORM_VEC2: glUniform2fv(locIndex, count, (float *)value); break;
        case SHADER_UNIFORM_VEC3: glUniform3fv(locIndex, count, (float *)value); break;
        case SHADER_UNIFORM_VEC4: glUniform4fv(locIndex, count, (float *)value); break;
        case SHADER_UNIFORM_INT: glUniform1iv(locIndex, count, (int *)value); break;
        case SHADER_UNIFORM_IVEC2: glUniform2iv(locIndex, count, (int *)value); break;
        case SHADER_UNIFORM_IVEC3: glUniform3iv(locIndex, count, (int *)value); break;
        case SHADER_UNIFORM_IVEC4: glUniform4iv(locIndex, count, (int *)value); break;
        case SHADER_UNIFORM_SAMPLER2D: glUniform1iv(locIndex, count, (int *)value); break;
        // TODO: Add the additional matrix types?
        default: TRACELOG(LOG_WARNING, "SHADER: Failed to set uniform value, data type not recognized");
    }
#endif
}
```

I think in that case, probably users should use OpenGL directly.

PR #1835, opened 2021-06-17
Fixed: Binding vertex position twice

Sirvoid:


raysan5:
@Sirvoid thanks for the fix!

PR #1840, opened 2021-06-20
Add support for resizing Emscripten canvas

skylersaleh:
Hi, 

I was trying to use raylib in an application where I wanted the canvas to be able to be dynamically resized (so that it could cover the entire web page), and I noticed that the support for this was only partially implemented in raylib (and commented out). 

So, I went and finished the rest of it and got it working. I hope this is helpful. 

Thanks, for making a great library.


skylersaleh:
You can see an app with this PR applied at the following address: 

https://skylersaleh.github.io/SkyBoy/

raysan5:
@skylersaleh Wow! Amazing! I love that GameBoy emulator! Great work! 

I'm merging it for now, the implementation looks nice. I need to see how it works [with my tools](https://raylibtech.itch.io/), at this moment I was managing the canvas resizing from html size because every tool had some specific needs.

PR #1842, opened 2021-06-21
Fix missing fclose in tinyobj loader

devoil-g:
Missing fclose in tinyobj loader.


raysan5:
@devoil-g I'm merging this fix but actually, `tinyobj_loader_c` must be updated to latest official version that do not use `FILE` anymore.

PR #1843, opened 2021-06-21
Fix DrawTextRec

n67094:
When :

```
Vector2 vec = MeasureTextEx(...);
Rectangle rec = {..., .width: vec.x);
DrawTextRec(...,rect,...);
```
The last letter of the text was not displayed before this fix.


n67094:
arf a probleme remains: 

INFO: textOffsetX + glyphWidth 138.000000 (log put after line 989 of the PR)
INFO: --- text width 133.000000 (the same text measured with MeasureTextEx())

raysan5:
@n67094 Thanks for the review! Actually, that functions should be probably removed from raylib and let the users manage "boxed" text by themselves...

Issue #1838, opened 2021-06-19
[core] Expose `glfwSetWindowSize` on web (or equivalent)


nikki93:
### Issue description

`SetWindowSize` currently [`#if`s-out the call to `glfwSetWindowSize` on web](https://github.com/raysan5/raylib/blob/28093c46a8d264206936d2eb534c56173346020b/src/core.c#L1549-L1555), which makes it hard to create a resizable canvas on web that remains high dpi and unscaled at all sizes. If I remove the `#if` and call it on web it is actually successful at resizing the canvas.

### Environment

Web -- mostly tested on Chrome. Building from Linux, Windows or macOS.

### Issue Screenshot

This is the canvas with the `#if` in place as it is currently in raylib -- you can see all of the rendering getting blurry due to web-level re-scaling:
![raylib-web-resize-orig](https://user-images.githubusercontent.com/49151/122654368-b82b1080-d0ff-11eb-9aa7-96dc01e6a983.gif)

Example of a resizable canvas after I remove the `#if` and add the code from the next section -- the lines and font remain at the same size while the scene itself zooms, with the canvas maintaining dynamic layout vs. the web elements surrounding it:
![raylib-web-resize](https://user-images.githubusercontent.com/49151/122654121-fcb5ac80-d0fd-11eb-8822-dc0e83cbe604.gif)

### Code Example

To achieve the above resizability this is what happens on each frame:
```c
  Vector2 windowSize = { float(GetScreenWidth()), float(GetScreenHeight()) };
  {
#ifdef __EMSCRIPTEN__
    Vector2 canvasSize = {
      float(int(EM_ASM_DOUBLE({
        return window.devicePixelRatio
            * document.querySelector("#canvas").getBoundingClientRect().width;
      }))),
      float(int(EM_ASM_DOUBLE({
        return window.devicePixelRatio
            * document.querySelector("#canvas").getBoundingClientRect().height;
      }))),
    };
    if (windowSize.x != canvasSize.x || windowSize.y != canvasSize.y) {
      SetWindowSize(int(canvasSize.x), int(canvasSize.y));
    }
#endif
  }
  camera.offset = Vector2Scale(windowSize, 0.5);
  camera.zoom = float(windowSize.x) / 800;
```
So -- the scene drawn through `camera` acts as if the screen is 800 units wide at all times. The text is drawn outside camera mode and the line width is divided by the zoom. Exposing `glfwSetWindowSize` allows creating such a dynamically resizable yet non-blurry window.

[...]

PR #1847, opened 2021-06-23
Allow SetWindowSize() on web

nikki93:
This is a simple change to allow calling `SetWindowSize()` on web. The recent change https://github.com/raysan5/raylib/commit/9095dd9e82b6be2b4ab477deaea6fe4634d1583c was a good step forward but in my testing I found that on macOS there were some issues with high-dpi due to not multiplying by `devicePixelRatio`. If `SetWindowSize()` is exposed the user can still directly set the desired canvas size multiplied by `devicePixelRatio`. Here is a comparison (first doesn't use `devicePixelRatio`, second does, and last is a comparison of the two screenshots side-by-side at the same zoom level):
Screen Shot 2021-06-22 at 8 22 30 PM
Screen Shot 2021-06-22 at 8 21 55 PM
image
Regardless of the dpi issues just providing access to `SetWindowSize()` lets the user manage this themselves and also apply other custom size behavior so it seems useful.
This would solve https://github.com/raysan5/raylib/issues/1838.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1560,7 +1560,7 @@ void SetWindowMinSize(int width, int height)
 // TODO: Issues on HighDPI scaling
 void SetWindowSize(int width, int height)
 {
-#if defined(PLATFORM_DESKTOP)
+#if defined(PLATFORM_DESKTOP) || defined(PLATFORM_WEB)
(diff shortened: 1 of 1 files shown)


raysan5:
@nikki93 thanks for the review!

PR #1849, opened 2021-06-23
Major revamp gltf

object71:
@raysan5 
Some major changes on GLTF model loading with the following highlights:

- Removed some previously extracted functions
- Added new GLTFReadValueAs function that will check and read accessors and allocate an array correctly. It is a bit long with a lot of switch cases but really simplifies the rest of the code. (Sadly no templated/generic functions in ANSI C)
- Hierarchical loading - now only the scene is loaded and the scene tree is traversed recursively. This was a requirement because there are some node transforms (node is not to be mistaken with a bone, though bones are also nodes) that are applied on parent and child nodes.
- Now models like the GearboxAssy are loaded correctly. Keep in mind that GearboxAssy is loaded in the example but has an offset from the center of the scene. This is not a bug - Godot loaded it the same way its just how the model is...

PS: Sadly animations are not yet fixed


object71:
Ugh. climits fails. I will replace the values.

examples/core/core_rotate_by_quat.c
```diff
@@ -0,0 +1,83 @@
+#include "raylib.h"
```
raysan5, line 1: Should this file be included with the PR?

src/models.c
```diff
@@ -49,6 +49,7 @@
 #include <stdlib.h>         // Required for: malloc(), free()
 #include <string.h>         // Required for: memcmp(), strlen()
 #include <math.h>           // Required for: sinf(), cosf(), sqrtf(), fabsf()
+#include <climits>
```
raysan5, line 52: This should be `<limits.h>`

raysan5:
@object71 Just took a quick look, it seems ok for the moment. I understand there are some bits missing, right? Please, let me know when it's ready to merge and test.

examples/core/core_rotate_by_quat.c
object71: I don't know, came from the merge. I can remove it.

src/models.c
object71: Actually no, this should be limits.h in all cases except for MSVC based compilers (cl.exe and clang-cl.exe) which require it to be climits.

object71:
@raysan5 Should be ready at this point. I won't solve the animations in this PR I just wanted to commit some simplification of the whole GLTF loading so that it is a bit more accessible to more junior developers reading the code.

src/models.c
raysan5: is it possible to remove this include and just assign some default values?

raysan5: > The `<climits>` wrapper of `<limits.h>` is only available in C++, not C. Also it might be worth noting that in C++ those headers are not guaranteed to contain any declarations specified in the C standard unless explicitly specified in C++ standard.

object71: Of course. Just thought it could be cleaner to have defines for readability of where the values come from. I will just enter the calculations, remove the include and write a comment of what the number is.

raysan5: Nice, thank you very much!

raysan5:
@object71 Thank you very much for this big review! :D

object71:
@raysan5 You didn't wait for me to fix the limit.h files? I will do that later.

raysan5:
@object71 I saw it after merging when reviewing the PR (I usually do a deeper review after merging). Just removed the includes and everything just compiled fine so I thought the usage was already reviewed.

PR #1851, opened 2021-06-24
fix to LoadShaderFromMemory

Ruminant:
Just a simple fix - the LoadShaderFromMemory function almost matches LoadShader, except that the "matNormal" uniform location isn't fetched.

This breaks the lighting in the simple lighting sample (if we switch out the use of LoadShader for LoadShaderFromMemory).


raysan5:
@Ruminant Oh! Good catch! Thanks for the fix! :D

Issue #1854, opened 2021-06-29
Printing pointer instead of pointer value


masterex1000:
In `CompressData()` it looks like a pointer is getting printed instead of the pointer's value. 

Function header:
https://github.com/raysan5/raylib/blob/6f30a19dd7d03a68fc3e39f00e8cd2de0fba95f2/src/core.c#L3023

and a bit later in the function....
https://github.com/raysan5/raylib/blob/6f30a19dd7d03a68fc3e39f00e8cd2de0fba95f2/src/core.c#L3036

Note no `*` in front of `compDataLength`

raysan5:
@masterex1000 Good catch! Feel free to send a PR! 👍😄

raysan5:
Just reviewed it.

PR #1856, opened 2021-06-29
Fix DecompressData() pointer issue

masterex1000:
#1854 This issue also exists in the `DecompressData()` function.


raysan5:
@masterex1000 thanks for the fix! 👍😄

Issue #1816, opened 2021-06-08
[textures] High GPU usage on `textures_image_loading` example


nagy:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib ( i just used the examples )

### Issue description

When starting the example `textures_image_loading` I see about 100% GPU usage. Every other example that I tested worked fine with almost 0% of GPU usage.

### Environment

platform: linux
os: nixos-unstable
opengl-version: ??
GPU details: 
```sh
$ lspci | grep VGA
02:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (rev e7)
```
the gpu driver is `amdgpu`. To be specific, my GPU is Radeon RX 580.

### Issue Screenshot
I see this with `radeontop`

![radeontop](https://user-images.githubusercontent.com/692274/121234524-16e8c280-c894-11eb-98f2-e01240639580.png)

### Code Example

I just executed the example `textures_image_loading` on the todays master branch.

I am happy to provide any more information.

nagy:
Upon investigating this further, i found that the example was just missing a  `SetTargetFPS(60);`. It works fine now.

Sorry for the noise. I can make a pull request for this.

raysan5:
@nagy Ok, feel free to send a PR

PR #1867, opened 2021-07-11
Use 60 FPS for textures_image_loading example

nagy:
Follow up on #1816.


examples/textures/textures_image_loading.c
```diff
@@ -30,6 +30,8 @@ int main(void)
     UnloadImage(image);   // Once image has been converted to texture and uploaded to VRAM, it can be unloaded from RAM
     //---------------------------------------------------------------------------------------
 
+    SetTargetFPS(60);               // Set our game to run at 60 frames-per-second
```
raysan5, line 33: To be consistent with all the other examples, this function call should be placed before the points-line above it.

raysan5 (commented):
Please, could you check the review comment?

examples/textures/textures_image_loading.c
nagy: Updated.

raysan5:
@nagy thanks for the review!

PR #1868, opened 2021-07-11
DrawArc() and Vector3Slerp

ProfJski:
I wrote a Vector3Slerp() function which provides spherical linear interpolation between two points without using quaternions.  It also provides more options for the curve.  

The iradius parameter allows increasing the radius of the arc, so flatter or rounder curves can be drawn.  When iradius=0.0, the curve is a semi-circle with its center on the midpoint of the line between point 1 and point 2. Larger values relocate the center of the curve off the midpoint, producing a flatter curve.

The theta parameter allows the direction of the arc to be rotated in any direction perpendicular to the line between the two points (like a jump rope).

DrawArc() seemed like a natural next step, so I added that too.  The only additional parameters here are Color, and the number of segments to use for the arc, similar to the options you provide for circles.

I offer it to you if you find it useful for inclusion.  Hopefully I did everything right in the pull request! :-)

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -631,19 +631,20 @@ void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, fl
 
 //Draw an arc between two points
 //If iradius=0. the arc is a semi-circle with a center at the midpoint of the line between the two points.  Increase iradius for a flatter arc with a center off the midpoint.  Theta rotates the arc around the line between the two points.
-void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color) {
+void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color) 
+{
     float increment=1.0/(float)steps;
     Vector3 p;
 
     rlCheckRenderBatchLimit(2*(steps+1));
     rlBegin(RL_LINES);
         rlColor4ub(color.r, color.g, color.b, color.a);
-        for (float t=0.0;t<1.0;t+=increment) {
+        for (float t=0.0;t<1.0;t+=increment) 
+        {
             p=Vector3Slerp(v1,v2,t,iradius,theta);
             rlVertex3f(p.x,p.y,p.z);
             p=Vector3Slerp(v1,v2,t+increment,iradius,theta);
             rlVertex3f(p.x,p.y,p.z);
-
         }
     rlEnd();
 return;
--- a/src/models.c
+++ b/src/models.c
@@ -629,6 +629,26 @@ void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, fl
     rlPopMatrix();
 }
 
+//Draw an arc between two points
+//If iradius=0. the arc is a semi-circle with a center at the midpoint of the line between the two points.  Increase iradius for a flatter arc with a center off the midpoint.  Theta rotates the arc around the line between the two points.
+void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color) {
+    float increment=1.0/(float)steps;
+    Vector3 p;
+
+    rlCheckRenderBatchLimit(2*(steps+1));
+    rlBegin(RL_LINES);
+        rlColor4ub(color.r, color.g, color.b, color.a);
+        for (float t=0.0;t<1.0;t+=increment) {
+            p=Vector3Slerp(v1,v2,t,iradius,theta);
+            rlVertex3f(p.x,p.y,p.z);
+            p=Vector3Slerp(v1,v2,t+increment,iradius,theta);
+            rlVertex3f(p.x,p.y,p.z);
+
+        }
+    rlEnd();
+return;
+}
+
 // Draw a plane
 void DrawPlane(Vector3 centerPos, Vector2 size, Color color)
 {
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1370,6 +1370,7 @@ RLAPI void DrawSphereEx(Vector3 centerPos, float radius, int rings, int slices, 
 RLAPI void DrawSphereWires(Vector3 centerPos, float radius, int rings, int slices, Color color);         // Draw sphere wires
 RLAPI void DrawCylinder(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color); // Draw a cylinder/cone
 RLAPI void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color); // Draw a cylinder/cone wires
+RLAPI void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color);          // Draws an arc between two points
 RLAPI void DrawPlane(Vector3 centerPos, Vector2 size, Color color);                                      // Draw a plane XZ
 RLAPI void DrawRay(Ray ray, Color color);                                                                // Draw a ray line
 RLAPI void DrawGrid(int slices, float spacing);                                                          // Draw a grid (centered at (0, 0, 0))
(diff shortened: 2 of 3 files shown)


src/raymath.h
```diff
@@ -621,6 +621,40 @@ RMDEF float3 Vector3ToFloatV(Vector3 v)
...
+    else if (L.y!=0.0) { //Otherwise use y-axis
+        incept=(Vector3){0.0,-D/L.y,0.0};
+    }
+    else { //or z-axis.  All three cannot be zero because of the first if-statement in routine
+        incept=(Vector3){0.0,0.0,-D/L.z};
+    }
+
+    if (Vector3LengthSqr(Vector3Subtract(incept,midpoint))<0.1) {  //If our midpoint and axis intercept are very close together, displace the midpoint slightly so that normalization of F does not result in divide-by-zero or G=0 because F={1,0,0}
+        midpoint=Vector3Add(v1,Vector3Scale(L,0.49));
+        midpoint=Vector3Add(midpoint,Vector3CrossProduct(midpoint,{0.2,0.0,0.0}));
```
ProfJski, line 644: Line 644 should be:
midpoint=Vector3Add(midpoint,Vector3CrossProduct(midpoint,(Vector3){0.2,0.0,0.0}));

raysan5:
@ProfJski Thanks for the addition! Would you mind reviewing the code to follow the [raylib code style conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

ProfJski:
@raysan5 I think I fixed it up according to your style conventions.  I left comments on both files about what I did: single-letter variables to lower-case and curly brackets in "aligned" mode.

src/models.c
```diff
@@ -629,6 +629,27 @@ void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, fl
     rlPopMatrix();
 }
 
+//Draw an arc between two points
+//If iradius=0. the arc is a semi-circle with a center at the midpoint of the line between the two points.  Increase iradius for a flatter arc with a center off the midpoint.  Theta rotates the arc around the line between the two points.
+void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color)
```
raysan5, line 634: I understand how this works but, sincerely, `iradius` and `theta` do not seem intuitive for the users, maybe is the parameter name...

src/raylib.h
```diff
@@ -1370,6 +1370,7 @@ RLAPI void DrawSphereEx(Vector3 centerPos, float radius, int rings, int slices,
 RLAPI void DrawSphereWires(Vector3 centerPos, float radius, int rings, int slices, Color color);         // Draw sphere wires
 RLAPI void DrawCylinder(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color); // Draw a cylinder/cone
 RLAPI void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color); // Draw a cylinder/cone wires
+RLAPI void DrawArc(Vector3 v1, Vector3 v2, float iradius, float theta, int steps, Color color);          // Draws an arc between two points
```
raysan5, line 1373: Also de function name could be confusing between 2D and 3D, maybe it could be renamed to `DrawArc3D()` or `DrawLineBezier3D()` to be aligned with the 2D implementation... but I understand it's not the same...

src/models.c
ProfJski: @raysan5 : That makes sense.  Instead of `theta` would simply `rotation` be better?  And for uniformity, it will be degrees not radians (and use `DEG2RAD` for conversion).

Your critique of `iradius` made me think about what might be most useful.  What about this?

Replace `iradius` with `flatness`.  Make `flatness` relative to the distance between the two points.  A flatness of 1.0 is the minimum radius = half the distance between the points = a semi-circle.  A flatness of 2.0 means twice the minimum radius, so an arc that's about twice as "flat" as a semi-circle.  This has the advantage of providing the same shape of the curve at any scale of distance between the two points.  The user can still find the arc radius easily:  Arc radius= (Half distance between the two points) * flatness

src/raylib.h
ProfJski: Sure I can do that.  I realize now that the other 3D functions above it don't have 3D in the name because a 3D version of them is not possible, whereas an arc can be 2D or 3D.  I didn't do a 2D version precisely because you had 2D curves already.  But if you wanted a 2D version that would be easy to code up.

The thing that made the 3D version lengthy is that `Vector3Perpendicular()` doesn't suit for all orientations of the line between the two points, because of the way that `cardinalAxis` is chosen in that function, and the perpendicular line to a line in 3D is one of infintely many on the perpendicular plane.  For the 2D case, it's just one cross product and done.

src/models.c
raysan5: @ProfJski I decided to investigate how other libraries implement this function, note that most of them refer to `DrawArc()` as a 2D arc, so, `DrawArc3D()` would be a possibility. Here some use cases:

 - [Handles.DrawWireArc](https://docs.unity3d.com/ScriptReference/Handles.DrawWireArc.html) Unity Engine
```cs
void DrawWireArc(Vector3 center, Vector3 normal, Vector3 from, float angle, float radius, float thickness = 0.0f);
```
 - [DrawArc](https://docs.unrealengine.com/4.26/en-US/API/Runtime/Engine/DrawArc/) Unreal Engine
```cpp
void DrawArc(FPrimitiveDrawInterface * PDI, const FVector Base, const FVector X, const FVector Y, const float MinAngle, const float MaxAngle, const float Radius, const int32 Sections, const FLinearColor & Color, uint8 DepthPriority);
```
 - [DrawCircleArc](https://docs.godotengine.org/en/stable/tutorials/2d/custom_drawing_in_2d.html#an-example-drawing-circular-arcs) Godot Engine example
```cs
void DrawCircleArc(Vector2 center, float radius, float angleFrom, float angleTo, Color color);
```
 - [Graphics.DrawArc Method](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.graphics.drawarc?view=net-5.0) .NET System.Drawing:
```cs
void DrawArc(System.Drawing.Pen pen, System.Drawing.Rectangle rect, float startAngle, float sweepAngle);
```
 - [Canvas.drawArc](https://developer.android.com/reference/android/graphics/Canvas#drawArc(android.graphics.RectF,%20float,%20float,%20boolean,%20android.graphics.Paint)) Android Canvas ([+info](https://thoughtbot.com/blog/android-canvas-drawarc-method-a-visual-guide))
```java
void drawArc(RectF oval, float startAngle, float sweepAngle, boolean useCenter, Paint paint);
```

What do you think?

src/raylib.h
raysan5: Probably just a `DrawArc3D()` would be enough.

src/models.c
raysan5: I keep thinking about it... maybe instead of `flatness` we can call it `roundness` so, a `roundness = 0.0` would be a flat line (and `theta` not considered) and a `roundness = 1.0` would be a semi-circle, even allowing greater values for bigger circles. About `theta`, maybe we can call it `swinging` so, with `swinging = 0.0` the arc is balanced and then allow `swinging = -1.0` and `swinging = 1.0` to balance the arc to the left (first vertex) or to the right (second vertex).

What do you think?

630Studios: I would say iRadius and theta are more intuitive than making up new names for things, especially when speaking about well established mathematical concepts like arcs in 2d and 3d space. 

How would "flatness" and "swinging", two terms which I have never heard in relationship to an arc, be more intuitive to and end user?

If you wanted to make it more intuitive to end users, it should probably follow a similar convention to pretty much everything else out there instead of trying to make up something new. Which means you have a radius, fromAngle, and toAngle.

raysan5:
@ProfJski I won't be adding `DrawArc()` for now but I'd like to add `Vector3Slerp()`. Please, could you review the PR? Thanks and sorry for the inconvenience.

PR #1869, opened 2021-07-11
Fix Android multi-touch

humbe:


raysan5:
@humbe thanks for the improvement!

PR #1871, opened 2021-07-12
Fixed normals

GoldenThumbs:
Fixed vertex normals not being loaded and replace by vertex positions.


raysan5:
@GoldenThumbs thanks for the fix!

Issue #1876, opened 2021-07-16
[models] `GenMeshTangent()` should enable vertex attribute


630Studios:
1) According to user Kenneth in discord, Tangents that are calculated on a model after it is loaded are not being passed to sharers and tangents only hold the value of 0.

It would appear that this might be related to the fact the tangents are calculated after the mesh is loaded.

Calling the following code after `GenMeshTangent` fixes the issue.
```C
rlEnableVertexArray(model.meshes[0].vaoId);
rlSetVertexAttribute(4, 4, RL_FLOAT, 0, 0, 0);
rlEnableVertexAttribute(4);
rlDisableVertexArray();
```

2) GenMeshTangents assumes that if tangent data is present on a mesh the length of the tangent data is the length of the vertex count of the mesh. Under certain circumstances this may be true, but there are also a variety of circumstances that can arise where this is not the case.

The code proceeds to use the vertexCount as the length of the tangent data, which can result in either not calculating all the tangents, or attempting to exceed the bounds of the data. 

3) At the end of GenMeshTangents the following is called.
```C
// Load a new tangent attributes buffer
    mesh->vboId[SHADER_LOC_VERTEX_TANGENT] = rlLoadVertexBuffer(mesh->tangents, mesh->vertexCount * 4 * sizeof(float), false);
```
If there was previously tangent data assigned to this mesh, the previous VertexBuffer is never released via `rlUnloadVertexBuffer`

Given issues 2 & 3 it would seem that `GenMeshTangents` should determine if the mesh previous had tangents, if so both the tangents and the previous vertex buffer should be freed, a new list of Tangents should be created, and a new VertexBuffer should be uploaded.

PR #1878, opened 2021-07-16
[Models] Better Fix for GenMeshTangents issue #1876

630Studios:
This is a better fix for issue #1876 and the GenMeshTangents function in general. 

The function was assuming that the mesh had already been uploaded to the GPU and was attempting to access the vboId field without ever verifying that it had actually been allocated.

Additionally if the vboId has been allocated, and the tangents already have a vertex buffer, instead of releasing the old one and creating a new one, this now simply updates the existing buffer.


raysan5:
Nice! Thanks for the review!

PR #1879, opened 2021-07-17
Raspberry RPI/DRM keyboard fix

luizpestana:
Legacy keyboard support was blocking the render loop, also it was keeping the last key status until the next key.


raysan5:
@luizpestana Nice! Thank you very much for reviewing this issue! 👍😄

PR #1883, opened 2021-07-18
Add example shaders_shapes_outline

GoldenThumbs:


raysan5:
@GoldenThumbs thank you very much! This exampke is a nice addition! 👍

jamesl-github:
what does it mean when an example is copyrighted by the submitter ?

raysan5:
@jamesl-github that the submitter is the original creator of the example.

jamesl-github:
> that the submitter is the original creator of the example.

what I meant was how does that affect licensing            
are all examples released under the same license as raylib ? 
is it explicitly stated somewhere that any example you submit to raylib automatically acquires and falls under the raylib license ?
most licenses I see say "attribution required" or "attribution appreciated, but not required"  
they also say things like "can be used in both commercial and non-commercial code with or without modification" 
so I was wondering if the submitter retains all rights to the code or if the code acquires raylib's license

raysan5:
@jamesl-github Afaik, all contributed examples fall under the same license as raylib: zlib/libpng if nothing different is specified

PR #1884, opened 2021-07-19
Ignore generated example binaries

luizpestana:


raysan5:
@luizpestana thanks for the improvement

PR #1885, opened 2021-07-23
Clarify raylib and raylib-cpp in bindings file

JeffM2501:
This PR modifies the bindings file to be less confusing for C++ users.

Raylib is noted as working with C and C++ directly, because it does.

Raylib-cpp has been moved into a Utility Wrapper section since it is not a traditional language binding (it is not required in order to use raylib with C++) but simply an optional conversive wrapper.

### DIFF
--- a/BINDINGS.md
+++ b/BINDINGS.md
@@ -2,12 +2,12 @@
 
 Some people ported raylib to other languages in form of bindings or wrappers to the library.
 
-Here it is a list with the ones I'm aware of:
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 Not sure if this makes it more clear to future users... but I'm merging.

raysan5:
Also note that all "bindings" should be reviewed to verify they are not wrappers and avoid misleading information for the users...

PR #1894, opened 2021-07-26
Remove unused UWP defines

GithubPrankster:
**UWP** is not available on `master` branch anymore so these definitions don't do anything. 
I'm curious if stating *Raylib* supports **UWP** on `raylib.h` is still ok since a branch does exist with the platform supported?


raysan5:
@GithubPrankster Nice! Thanks for the clean up! It should also be removed from the `raylib.h` comments.

PR #1896, opened 2021-07-27
Don't normalize zero length vectors.

JeffM2501:
This PR prevents Vector2Normalize from dividing by zero like Vector3Normalize does.


raysan5:

Issue #1866, opened 2021-07-10
[models] Proposing to add `GetModelBoundingBox()`


630Studios:
I noticed there is no functionality present for getting the BoundingBox of a model, but only a single mesh.
This lack of functionality is also exhibited in the example models_loading.c . Since there is no functionality to get the bounds of a Model, the example simply gets the bounds of the first mesh, which may or may not be the full bounds of the model if additional meshes are present in the model. 

This lack of functionality can for the most part be easily overcome as an individual can write their own GetModelBounds function, however this is functionality that almost anyone dealing with a model will need / end up writing, and as such it would seem to indicate such functionality should be part of the RayLib itself.

Additionally, solving this issue for those working through bindings in languages that used manages memory is a tad more complicated, though doable, without additional underlying functionality. Simply adding this function removes that problem all together. 

```C
BoundingBox GetModelBoundingBox(Model model)
{
    
    int meshCount = model.meshCount;
    BoundingBox bounds = { 0 };
    
    if (meshCount < 1)
    {      
        
        return bounds;
    }else{

        Vector3 tPos;
        bounds = GetMeshBoundingBox(model.meshes[0]);
        for (int i = 1; i < meshCount;i++)
        {
        
            BoundingBox tempBounds = GetMeshBoundingBox(model.meshes[i]);
            
            tPos.x = bounds.min.x < tempBounds.min.x ? bounds.min.x : tempBounds.min.x;
            tPos.y = bounds.min.y < tempBounds.min.y ? bounds.min.y : tempBounds.min.y;
            tPos.z = bounds.min.z < tempBounds.min.z ? bounds.min.z : tempBounds.min.z;
            bounds.min = tPos;

            tPos.x = bounds.max.x > tempBounds.max.x ? bounds.max.x : tempBounds.max.x;
            tPos.y = bounds.max.y > tempBounds.max.y ? bounds.max.y : tempBounds.max.y;
            tPos.z = bounds.max.z > tempBounds.max.z ? bounds.max.z : tempBounds.max.z;
            bounds.max = tPos;
        }
    }
    
    return bounds;
}
```

raysan5:
Ok, feel free to send a PR to add it.

PR #1899, opened 2021-07-28
Add GetModelBoundingBox()

GithubPrankster:
Following #1866. Credits go to [630Studios](https://github.com/630Studios) for the function itself.

The function's prototype has been added in the `Mesh manipulation functions` category in `raylib.h`, should a "`Model manipulation functions`" category be made, or should it wait until more similar functions are created?


raysan5:
@GithubPrankster Thanks for the PR. I'll add it myself while reorganizing models functionality.

PR #1900, opened 2021-07-28
fixed DisableCursor() on web by registering an empty mouse click event

grenappels:
Just a small fix that may be a hack, haven't tested on more than just my machine yet, but it seems harmless enough


raysan5:
@grenappels Thanks for this fix!

PR #1906, opened 2021-08-07
Fix Makefile to build WASM examples.

rcorre:
- Add --preload-file flag before specifying the resource dir
- Add empty resource dir to `shapes/` (otherwise wasm-ld will fail)
- Add wasm outputs to .gitigore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,6 +58,12 @@ packages/
 *.bc
 *.so
 
+# Ignore wasm data in examples/
+examples/**/*.wasm
+examples/**/*.data
+examples/**/*.js
+examples/**/*.html
+
 # Ignore files build by xcode
 *.mode*v*
 *.pbxuser
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -244,7 +244,7 @@ ifeq ($(PLATFORM),PLATFORM_WEB)
     # --memory-init-file 0       # to avoid an external memory initialization code file (.mem)
     # --preload-file resources   # specify a resources folder for data compilation
     # --source-map-base          # allow debugging in browser with source map
-    CFLAGS += -s USE_GLFW=3 -s ASYNCIFY -s TOTAL_MEMORY=67108864 -s FORCE_FILESYSTEM=1 $(dir $<)resources@resources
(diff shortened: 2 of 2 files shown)


raysan5:
@rcorre thanks for the review!

rcorre:
@raysan5 no problem! Just FYI, I added the .gitignore to keep `shapes/resources` in source control, since git ignores empty directories. Removing that removed the directory, so `master` fails with:

```
error: shapes/resources@resources does not exist
emcc: error: '/usr/lib/emscripten/tools/file_packager shapes/shapes_basic_shapes.data --from-emcc --export-name=Module --preload shapes/resources@resources' failed (returned 1)
```

raysan5:
@rcorre actually, I never use this Makefile to compile examples for web, two reasons: 
 - Uses `ASYNCIFY` option, so code is kind-of interpreted at runtime with a big performance penalization. 
 - Packages ALL resource folder with every example while only some of those resources are used by every example.

Web examples are manually converted to use a proper structure (to allow emscripten control the game loop) and a custom Makefile is created to deal with every example properly. Check it here:  https://github.com/raysan5/raylib.com/tree/master/examples/web

I just left the emscripten option in that set of examples as a reference for the users.

rcorre:
Thanks @raysan5! Should we update https://github.com/raysan5/raylib/wiki/Working-for-Web-(HTML5)#compiling-raylib-game to point users at https://github.com/raysan5/raylib.com/tree/master/examples/web as an example? Currently it says:

> To compile your raylib code for web, it's recommended you use an already setup Makefile ready for that, raylib/templates/simple_game/Makefile could be used for reference, just copy it to your code folder.

`raylib/templates` doesn't exist any more, so I checked out https://github.com/raysan5/raylib-game-template but couldn't get it working. The `examples` folder in raylib was actually the most useful thing for learning how to use WASM for me!

> Uses ASYNCIFY option, so code is kind-of interpreted at runtime with a big performance penalization.

Is this no longer true (from https://github.com/raysan5/raylib/wiki/Working-for-Web-(HTML5)#compiling-raylib-game)?

> raylib examples Makefile has been adapted to use ASYNCIFY and, despite the theorical performace lost, the tested results are great.

raysan5:
@rcorre That info on the Wiki should probably updated...

About the `ASYNCIFY` performance lost, that sentence should be reformulated, I just rewrote it to note that, despite they work great (because they are very small samples), there is a performance penalization.

PR #1909, opened 2021-08-09
Document Get{Key,Char}Pressed return value.

rcorre:
I think it is useful to know what the empty behavior is in cases where
you are enumerating key presses (e.g. implementing "press any key to
continue").


raysan5:
@rcorre Thanks for the improvement.

PR #1914, opened 2021-08-12
[rlgl] rlUpdateVertexBuffer

630Studios:
bufferId parameter is supposed to be an unsigned int. This fixes it.


630Studios:
I have absolutely no idea why the CMake windows and linux builds failed....

raysan5:
@630Studios thanks for the fix!

Issue #1809, opened 2021-06-05
[models] LoadOBJ crashes if model has more materials than meshes.


nyeecola:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

On `LoadOBJ()`, raylib does `model.meshCount = materialCount;`, in one of my models the material count is 4, while the mesh count is 1. This ends up causing a segmentation fault on line `3322`: `for (int v = 0; v < 3; v++) { model.meshes[mm].vertices[vCount[mm] + v] = attrib.vertices[idx0.v_idx*3 + v]; } vCount[mm] +=3;` because mm is out of bounds.

### Environment

Windows 10 using Visual Studio 2019.

### Issue Screenshot

https://i.imgur.com/v6viC8r.png

nyeecola:
I think my understanding of this issue is flawed, but I think the screenshot helps understanding the crash better than I can explain why it happens.

nyeecola:
Here is the affected OBJ and material.
[machinegun.zip](https://github.com/raysan5/raylib/files/6603235/machinegun.zip)

PR #1917, opened 2021-08-12
[Models] Fix for issue #1809

630Studios:
All fixed. The model from issue 1809 now now longer crashes raylib and displays.

image


raysan5:
@630Studios thanks for the review!

Issue #1918, opened 2021-08-13
[physac] QueryPerformance... functions are not defined.


JioChoi:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

When I include physac.h, it doesn't build, and it says "LNK2001	unresolved external symbol "int __cdecl QueryPerformanceFrequency(unsigned __int64 *)" (?QueryPerformanceFrequency@@YAHPEA_K@Z)" and "LNK2001	unresolved external symbol "int __cdecl QueryPerformanceCounter(unsigned __int64 *)" (?QueryPerformanceCounter@@YAHPEA_K@Z)".

Surprisingly, it works when I put `#include <extras/physac.h>` inside of `extern "C"` block. However, I heard raylib already has had it in. I checked physac, and it also has had it.

### Environment

 - Windows 10 Home (20H2, 19042.1110).
 - Visual Studio 2019.

### Issue Screenshot

![image](https://user-images.githubusercontent.com/73428860/129333147-aa7066c3-bb82-47a1-84a1-4f90b20f878f.png)
![image](https://user-images.githubusercontent.com/73428860/129333266-a3144fe3-812e-4c00-892c-c8ad55f0fc91.png)

![image](https://user-images.githubusercontent.com/73428860/129333423-17dbbac4-64b7-4f82-852a-f915f7e2ecce.png)
Somewhy it's underlining RAYWHITE... Is my Visual Studio broken?

![image](https://user-images.githubusercontent.com/73428860/129333540-2cd9a7a1-2158-4260-8c38-2e22a5c79760.png)

![image](https://user-images.githubusercontent.com/73428860/129333715-f5fd1465-a387-4155-bbae-0b2094907876.png)
It also works when I only put `extern "C"` on this part.

### Code Example
I don't think we need a code example for this one... but just in case I'll put it.

```
#include <iostream>

#define PHYSAC_IMPLEMENTATION

#include <raylib.h>
#include <extras/physac.h>

int main(int argc, char* argv[]) {
	InitWindow(1920, 1080, "Test");
	SetConfigFlags(FLAG_VSYNC_HINT);
	SetWindowState(FLAG_WINDOW_UNDECORATED);
	SetWindowPosition(0, 0);

	InitPhysics();

	while (!WindowShouldClose()) {
		BeginDrawing();

[...]

PR #1919, opened 2021-08-13
Fix for Issue #1918

JioChoi:
I added PHYSACDEF in front of `QueryPerformanceCounter` and `QueryPerformanceFrequency` in to fix issue #1918.
Those two function declarations are causing name mangling, because they don't have `extern "C"`.


raysan5:
@JioChoi Actually, that's not the proper solution because if `PHYSAC_STATIC` was defined it would generate and issue setting those functions as internal to the module while they must be always external. I'm pushing another solution.

PR #1923, opened 2021-08-15
Correct scripts README to reflect build-windows.bat changes

orcmid:
The place for splicing the location of vcvarsall.bat moved in the last modification of build-windows.bat.  The README still had the old location.


raysan5:
@orcmid thanks for the review! 👍

PR #1934, opened 2021-08-18
add LoadImageAnimDelays() with int **frameDelays parameter

ardputten:
LoadImageAnim() discards the int **delays parameter to stbi_load_gif_from_memory(). I personally needed to use this parameter to determine how long I should display each frame of a gif and thought maybe other people could use it as well.

I added a new function LoadImageAnimDelays(), that replaces LoadImageAnim() (which now calls LoadImageAnimDelays() for backwards compatibility). This function is the same as the old LoadImageAnim(), except that it also returns the frame delays of a gif so it is possible to calculate/know when the next frame should be rendered.

I only tested this change on windows.

This is my first time making a PR so in case I did anything wrong please let me know.


raysan5:
@ardputten I'm afraid I'm not adding this function, `LoadImageAnim()` was added to support multiple image-sequence fileformats and not all the formats support delay data between frames. For this specific use case I recommend using `stbi` library directly.

PR #1941, opened 2021-08-20
Add `up` argument to `DrawBillboardPro`

GithubPrankster:
Based on Discord user `rose 🪐`'s feature request, this allows more flexibility with the Billboard's direction.
`DrawBillboardRec` has been modified accordingly with the original assumption of `up` being locked to Y-axis.
Suggestions & others would be good.


raysan5:
@GithubPrankster I'm merging but I'm still thinking about this function... At this moment it's the only function in raylib with **+8 input parameters**. I try to avoid +6 input parameters in functions, actually there are only 5 functions in raylib with +6 parameters.

I'll check if this function can be simplified or exposed in some other way.

PR #1946, opened 2021-08-23
Revert "Ignore generated example binaries (#1884)"

luizpestana:
We need to revert this as its ignoring changes to Makefiles.


raysan5:
@luizpestana Oh, goid catch! I think I had that same issue in the past... 😄

PR #1947, opened 2021-08-23
Fix tcc not finding emmintrin.h

fuserr7:
This allows to compile raylib with tcc with no errors.

### DIFF
--- a/src/external/stb_image.h
+++ b/src/external/stb_image.h
@@ -688,10 +688,6 @@ typedef unsigned char validate_uint32[sizeof(stbi__uint32)==4 ? 1 : -1];
 #define STBI_NO_SIMD
 #endif
 
-#if defined(__TINYC__)
-  #define STBI_NO_SIMD
-#endif
-
 #if !defined(STBI_NO_SIMD) && (defined(STBI__X86_TARGET) || defined(STBI__X64_TARGET))
 #define STBI_SSE2
 #include <emmintrin.h>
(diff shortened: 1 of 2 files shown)


GithubPrankster (approved):
All good to go. I wonder if this could be passed to the actual `stb` repository, seems like a pretty reasonable thing to add.

raysan5:
@fuserr7 @GithubPrankster I'm afraid this is not the best place to add that check. `stb_image` is exposing `STBI_NO_SIMD` define to allow defining it on user side, not changing the library. It's a good practice to keep the library as compiler-agnostic as possible.

This piece of code should be added in `textures.c`, just before `#include "stb_image.h"`.

fuserr7:
@raysan5 hello, thanks for reviewing. Sure, I'll do the necessary changes later today.

raysan5:
@fuserr7 the PR is empty... are you adding the lines to `textures.c`?

fuserr7:
@raysan5 oh sorry, I didn't think it'd be sent automatically. I am uncertain of where to add the check. Above the stb_image inclusion?

raysan5:
@fuserr7 for example, it can be added here:
```c
// Image fileformats not supported by default
#define STBI_NO_PIC
#define STBI_NO_PNM             // Image format .ppm and .pgm
#if defined(__TINYC__)
    #define STBI_NO_SIMD
#endif

#if (defined(SUPPORT_FILEFORMAT_BMP) || \
...
```

fuserr7:
@raysan5 alright, I'll add it

raysan5:
@fuserr7 Nice! Thank you very much for reviewing and testing it! :D

Issue #1799, opened 2021-05-30
[models] Animated models loaded through Raylib by using GLTF format are broken


object71:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Animated models loaded through Raylib by using GLTF format are broken.
I am also attaching the model for reference:

[girl.zip](https://github.com/raysan5/raylib/files/6566921/girl.zip)

I will try and fix the issue when I can but if someone else is faster to it I won't mind :)

### Environment

I tested on Windows.

### Issue Screenshot

You can see how it is in the provided example here:

![error](https://user-images.githubusercontent.com/19205853/120120486-49b0fd80-c1a6-11eb-9a12-0260f65c438c.gif)

And how it should be when viewed through Windows 3D Viewer:

![correct](https://user-images.githubusercontent.com/19205853/120120501-559cbf80-c1a6-11eb-89f1-7db5de29135d.gif)

### Code Example

Minimal reproduction is to use the samples code for GLTF animations in Raylib and just set the model for loading and animations to use the girl.glb file that I am providing with this issue. There are two animations on that model - idle and run and both seem to be broken.

object71:
When drawing the dots for the bones' position (translated for each animation) the bones' positions seem to be in the right places. This leads me to believe there is no problem with the translation but the problem might be more in the bone weights or rotation values... Or how they are applied in the end.

object71:
I strongly suspect that the issue is related to the skin inverseBoneMatrix not being taken into account on the GLTF model.

More info about the skinning of the mesh: https://github.com/KhronosGroup/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_020_Skins.md

What we need to achieve is to still process animation on all nodes (no matter that not all of them are considered bones - in raylib they are all translated to bones because bones have the Transformation fields related to them). But also check which nodes are part of a skin and take into account that skin's array of inverseBoneMatrix.

Because it is a matrix though we would need a matrix decomposition function so that we can apply its transformations to the bone transformations. I've opened an issue for that: #1850

raysan5:
@object71 Any news on this issue?

object71:
Nothing new that I found out. I just don't have the time to work on it.

MrDiver:
Note: The `LoadModelAnimations` function needs an extra check if the file exists to not segfault.

[...]

PR #1964, opened 2021-08-28
[models] LoadGLTF fixed missing transformations and nonroot skinning problem

MrDiver:
I found the issue with the transforms. `LoadGLTFNode` assumed that every single transformation would automatically be saved in to the matrix, but this wasn't the case. I added the missing transforms with checks if they exist to ensure that all the transforms are applied on Load.

Heres a file that I used for testing which has all things that
[all_test.zip](https://github.com/raysan5/raylib/files/7071277/all_test.zip)

Currently a problem remains with the bones that the vertex data is already transformed on meshes that are a child of a bone structure, which results in the transformation being applied twice.

## Current result
image

## What the scene should look like if the bone error is fixed
image

There is some further information on this transform in the glTF file because sites like https://www.gltfviewer.com/ load the file perfectly fine.

But I really don't know where to find that data right now. But this is the first step in the right direction.

This is the current state of my raylib to figure out where everything goes wrong.
https://github.com/MrDiver/raylib/tree/gltf-recovery

This is a snippet of the loaded data from the glTF file and we can see that the vertices of the cube.004 are already transformed to the right position.
```
============= Model: NODE TRANSFORM pi: 5 ==============
Node Name: Armature
HAS TRANSLATION
[[ 1 0 0 0 ]
 [ 0 1 0 5 ]
 [ 0 0 1 0 ]
 [ 0 0 0 1 ]]

HAS ROTATION
[[ 0.707     0.5       0.5       0         ]
 [ -2.98E-08 0.707     -0.707    0         ]
 [ -0.707    0.5       0.5       0         ]
 [ 0         0         0         1         ]]

Current Transform
[[ 0.707     0.5       0.5       0         ]
 [ -2.98E-08 0.707     -0.707    5         ]
 [ -0.707    0.5       0.5       0         ]
 [ 0         0         0         1         ]]

=== Children of Armature
============= Model: NODE TRANSFORM pi: 5 ==============
Node Name: Cube.004
Current Transform
[[ 0.707     0.5       0.5       0         ]
 [ -2.98E-08 0.707     -0.707    5         ]
 [ -0.707    0.5       0.5       0         ]
 [ 0         0         0         1         ]]

IS MESH
Using Transform on vertex 0 : (-0.707107,4.500000,0.000000) to (1.750000,8.181980,2.750000)
Using Transform on vertex 1 : (-0.707107,4.500000,0.000000) to (1.750000,8.181980,2.750000)
Using Transform on vertex 2 : (-0.707107,4.500000,0.000000) to (1.750000,8.181980,2.750000)
Using Transform on vertex 3 : (0.000000,4.500000,0.707107) to (2.603554,7.681980,2.603554)
Using Transform on vertex 4 : (0.000000,4.500000,0.707107) to (2.603554,7.681980,2.603554)
Using Transform on vertex 5 : (0.000000,4.500000,0.707107) to (2.603554,7.681980,2.603554)
Using Transform on vertex 6 : (-0.707107,5.500000,0.000000) to (2.250001,8.889087,3.250000)
Using Transform on vertex 7 : (-0.707107,5.500000,0.000000) to (2.250001,8.889087,3.250000)
Using Transform on vertex 8 : (-0.707107,5.500000,0.000000) to (2.250001,8.889087,3.250000)
Using Transform on vertex 9 : (0.000000,5.500000,0.707107) to (3.103554,8.389087,3.103554)
Using Transform on vertex 10 : (0.000000,5.500000,0.707107) to (3.103554,8.389087,3.103554)
Using Transform on vertex 11 : (0.000000,5.500000,0.707107) to (3.103554,8.389087,3.103554)
Using Transform on vertex 12 : (0.000000,4.500000,-0.707107) to (1.896447,8.681980,1.896447)
Using Transform on vertex 13 : (0.000000,4.500000,-0.707107) to (1.896447,8.681980,1.896447)
Using Transform on vertex 14 : (0.000000,4.500000,-0.707107) to (1.896447,8.681980,1.896447)
Using Transform on vertex 15 : (0.707107,4.500000,-0.000000) to (2.750000,8.181980,1.750000)
Using Transform on vertex 16 : (0.707107,4.500000,-0.000000) to (2.750000,8.181980,1.750000)
Using Transform on vertex 17 : (0.707107,4.500000,-0.000000) to (2.750000,8.181980,1.750000)
Using Transform on vertex 18 : (0.000000,5.500000,-0.707107) to (2.396447,9.389087,2.396447)
Using Transform on vertex 19 : (0.000000,5.500000,-0.707107) to (2.396447,9.389087,2.396447)
Using Transform on vertex 20 : (0.000000,5.500000,-0.707107) to (2.396447,9.389087,2.396447)
Using Transform on vertex 21 : (0.707107,5.500000,-0.000000) to (3.250001,8.889087,2.250000)
Using Transform on vertex 22 : (0.707107,5.500000,-0.000000) to (3.250001,8.889087,2.250000)
Using Transform on vertex 23 : (0.707107,5.500000,-0.000000) to (3.250001,8.889087,2.250000)
=== Children of Cube.004
================= end node Cube.004 ==================
```

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -5485,7 +5485,7 @@ void LoadGLTFNode(cgltf_data *data, cgltf_node *node, Model *outModel, Matrix cu
     {
         // Check if skinning is enabled and load Mesh accordingly
         Matrix vertexTransform = currentTransform;
-        if(node->skin != NULL && node->parent != NULL)
+        if((node->skin != NULL) && (node->parent != NULL))
         {
             vertexTransform = localTransform;
             TRACELOG(LOG_WARNING,"MODEL: GLTF Node %s is skinned but not root node! Parent transformations will be ignored (NODE_SKINNED_MESH_NON_ROOT)",node->name);
--- a/src/models.c
+++ b/src/models.c
@@ -5446,34 +5446,52 @@ void LoadGLTFMesh(cgltf_data *data, cgltf_mesh *mesh, Model *outModel, Matrix cu
     }
 }
 
-void LoadGLTFNode(cgltf_data *data, cgltf_node *node, Model *outModel, Matrix currentTransform, int *primitiveIndex, const char *fileName)
+static Matrix GetNodeTransformationMatrix(cgltf_node *node, Matrix current)
 {
-    // Apply the transforms if they exist (Will still be applied even if no mesh is present to support emptys and bone structures)
-    if(node->has_matrix){
+    if (node->has_matrix)
+    {
         Matrix nodeTransform = {
         node->matrix[0], node->matrix[4], node->matrix[8], node->matrix[12],
         node->matrix[1], node->matrix[5], node->matrix[9], node->matrix[13],
         node->matrix[2], node->matrix[6], node->matrix[10], node->matrix[14],
         node->matrix[3], node->matrix[7], node->matrix[11], node->matrix[15] };
-        currentTransform = MatrixMultiply(nodeTransform, currentTransform);
+        current= MatrixMultiply(nodeTransform, current);
     }
-    if(node->has_translation){
+    if (node->has_translation)
+    {
         Matrix tl = MatrixTranslate(node->translation[0],node->translation[1],node->translation[2]);
-        currentTransform = MatrixMultiply(tl, currentTransform);
+        current = MatrixMultiply(tl, current);
     }
-    if(node->has_rotation){
+    if (node->has_rotation)
+    {
         Matrix rot = QuaternionToMatrix((Quaternion){node->rotation[0],node->rotation[1],node->rotation[2],node->rotation[3]});
-        currentTransform = MatrixMultiply(rot, currentTransform);
+        current = MatrixMultiply(rot, current);
     }
-    if(node->has_scale){
+    if (node->has_scale)
+    {
         Matrix scale = MatrixScale(node->scale[0],node->scale[1],node->scale[2]);
-        currentTransform = MatrixMultiply(scale, currentTransform);
+        current = MatrixMultiply(scale, current);
     }
-    // Load mesh if it exists
-    if (node->mesh != NULL){
-        LoadGLTFMesh(data, node->mesh, outModel, currentTransform, primitiveIndex, fileName);
-    }
+    return current;
+}
 
+void LoadGLTFNode(cgltf_data *data, cgltf_node *node, Model *outModel, Matrix currentTransform, int *primitiveIndex, const char *fileName)
+{
+    // Apply the transforms if they exist (Will still be applied even if no mesh is present to support emptys and bone structures)
+    Matrix localTransform = GetNodeTransformationMatrix(node, MatrixIdentity());
+    currentTransform = MatrixMultiply(localTransform, currentTransform);
+    // Load mesh if it exists
+    if (node->mesh != NULL)
+    {
+        // Check if skinning is enabled and load Mesh accordingly
+        Matrix vertexTransform = currentTransform;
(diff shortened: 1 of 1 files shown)


raysan5:
@MrDiver Nice! Thanks for the review, it looks good.

MrDiver:
Update on info:

As far as I can tell, the current system is using the `bindPose` and `boneInfo` Attributes as a replacement for the scene tree in the glTF file. But for some reason, there seems to be a few things missing, and I think I have to do a little history digging for this code before this fix is actually useful.

It is hard to decide where the actual problem is with this, because it is so tightly related to the animation problem with the bones. So this might be a quickfix for gltf loading of static models, but it will break the animation part even more because the center of the meshes are now different.

I think it would be useful to have the intended use for the attributes in the Mesh. Because the global transformation is already stored in `bindPose` but it is never used by anything. This might be related to the problem that there is no easy way of telling which `bindPose` is connected to which mesh, except for the `Mesh.boneIds` which should then probably be used to transform the vertex information.

For now i'm stuck at this point with the decision on how to change the values without breaking compatibility of the other two model types.

MrDiver:
## Current State
output_anim

New Update of the Animations. Optimization of the UpdateModelAnimation function, which now only updates the vertices when a change is applied to the vertex data. This results in the ability to combine animations if they are not affecting the same vertex data.

## Wanted State
Swing animation and Side to Side animation applied at the same time.
wanted1
Blue Rotate animation and Side to Side animation applied at the same time.

wanted2

## TODO
- Find and fix mismatch in animation data. Wrong boneId is used for animation data.
- Find a way to undo the transformations done to children of a bone structure

Thoughts: The data loaded in the animation is wrong for the objects that are a child of a bone structure.
The data is currently mismatched with the boneIds (But the right transformations are applied to the wrong object).

New File with Animations -> [all_test_anim.zip](https://github.com/raysan5/raylib/files/7072022/all_test_anim.zip)

~ Maybe someone knows why the data is loaded with a mismatch in Ids.

MrDiver:
### Mini Info Update (Fix for white cube [Skinning])
The problem with the white cube comes from skinning. We have to implement a check if a node is skinned or not and if it is skinned we have to discard the parent transform and replace it with the identity matrix. I don't quite know where to find this data in cgltf right now, but I know it was a part of a struct in the scene tree somewhere.

MrDiver:
# Update Log
### Changed
- NODE_SKINNED_MESH_NON_ROOT will now be checked and print a warning with LOG_WARNING level
- If the above error occurs then the current accumulated transformation is discarded

image

### Todo
- Get pose information and apply to skinned nodes

MrDiver:
The problem with the animations is too big to also pack it in to this issue because there are several functions involved.
The first assumption that I have that there's a mismatch in the `boneIds` when `BindGLTFPrimitiveToBones` is run and when the animation data is loaded. For now, I just added some checks and log messages, so the application doesn't just segfault when there's a problem with the animation Data.

I tested the current stage on the test models I created and also on the girl.gltf from #1799 
Fix list
- All transform information loaded (matrix, translate, rotate, scale)
- Transformation of vertex data on load of gltf files according to gltf standard.
- Skinning meshes handled correctly (static only)
- Correct passing down of transform data to children.

@raysan5 Okay, I think this is as much as I can provide for issue with the missing transforms.

### Example gltf file
image

raysan5:
@MrDiver Is it ready for review and merge?

raysan5:
@MrDiver Great work! Thank you very much for taking care of this issue!

PR #1970, opened 2021-09-02
[models] Animations not updating transforms correctly

MrDiver:
This PR should fix the mismatched transforms with the boneIds.

Current guess is that the calculation
https://github.com/raysan5/raylib/blob/9ff8ad853c77453768c59b5f9ae5206ec63f1bdb/src/models.c#L5168
is causing the mismatch, until now I couldn't verify that.

Checked
- [x] BoneIds are mapped correctly to nodes
- [ ] BoneIds for sampler are correctly calculated e.g. Animation to Bone mapping

### DIFF
--- a/src/models.c
+++ b/src/models.c
@@ -1615,6 +1615,10 @@ void UpdateModelAnimation(Model model, ModelAnimation anim, int frame)
         for (int m = 0; m < model.meshCount; m++)
         {
             Mesh mesh = model.meshes[m];
+            if(mesh.boneIds == NULL || mesh.boneWeights == NULL){
+                TRACELOG(LOG_WARNING, "MODEL: UpdateModelAnimation Mesh %i has no connection to bones",m);
+                continue;
+            }
             bool updated = false; // set to true when anim vertex information is updated
             Vector3 animVertex = { 0 };
             Vector3 animNormal = { 0 };
--- a/src/models.c
+++ b/src/models.c
@@ -1614,12 +1614,14 @@ void UpdateModelAnimation(Model model, ModelAnimation anim, int frame)
 
         for (int m = 0; m < model.meshCount; m++)
         {
+            Mesh mesh = model.meshes[m];
+            bool updated = false; // set to true when anim vertex information is updated
             Vector3 animVertex = { 0 };
             Vector3 animNormal = { 0 };
 
             Vector3 inTranslation = { 0 };
             Quaternion inRotation = { 0 };
-            //Vector3 inScale = { 0 };      // Not used...
+            Vector3 inScale = { 0 };
 
             Vector3 outTranslation = { 0 };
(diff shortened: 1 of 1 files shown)


MrDiver:
After further investigation, I found out that the wrong boneIds are caused by [`BindGLTFPrimitiveToBones`](https://github.com/raysan5/raylib/blob/9ff8ad853c77453768c59b5f9ae5206ec63f1bdb/src/models.c#L5028-L5055)
This is a first lookout on the fix (parenting of animations doesn't work right now for objects that also have an animation, but that is fixable)
animation_system
At least all the correct objects are animated!

MrDiver:
So this gltf file doesn't work correctly. I will look into it later but if someone wants to take a look at it feel free to do so.
[skinned_multi.gltf.zip](https://github.com/raysan5/raylib/files/7097667/skinned_multi.gltf.zip)
The current state is that for simple mesh skin combinations everything works fine but if you have multiple armatures/skins it breaks. For some reason some mesh bone ids are not initialized correctly. This could happen at multiple places.

MrDiver:
I pinned the problem down to being that `bones[a]` returns a too high index value for the actual array of joint nodes.
https://github.com/raysan5/raylib/blob/9ff8ad853c77453768c59b5f9ae5206ec63f1bdb/src/models.c#L5400-L5408
This results in the wrong ID being calculated i changed the calculation from a for loop to a simple
```c
unsigned int skinJointId = skinJoint - data->nodes;
```
which then resulted in the high values seen in the last post. Before it just used 0 as default value which then resulted in everything being affected by the wrong node.

Todo:
- Find a solution and understand what bones[a] is actually used for to then calculate the right id

MrDiver:
Update: https://github.com/raysan5/raylib/blob/9ff8ad853c77453768c59b5f9ae5206ec63f1bdb/src/models.c#L5400
The right skin is not selected before accessing the joints. Because direct access to the array `skins->` causes the index out of bounds behavior

raysan5:
@MrDiver Wow! It looks really nice! Please, let me know when ready! Maybe we can merge current work?

raysan5:
@MrDiver What's the state of this PR? Could we merge it?

raysan5:
@MrDiver I'm afraid I renamed this module and now changes can not be merged due to conflicts... Could you open a new PR with those changes for merging? Sorry for the inconvenience.

rcorre:
Thanks for the work on this! Unfortunately it still struggles with 
[Mech.zip](https://github.com/raysan5/raylib/files/7210123/Mech.zip)

https://user-images.githubusercontent.com/2496231/134333220-0321e921-caf6-4c9d-af02-25d97c92a8df.mp4

That's just a single mesh and armature

raysan5:
@MrDiver What's the state of this PR? Could it be merged? Please, let me know.

PR #1977, opened 2021-09-06
Fixes camera direction for first person camera mode.

GoodNike:
Previous mouse position was initialized to (0, 0) and caused camera direction vector to move first time scene initializes.
Watch the difference on a core_3d_camera_first_person example.


raysan5:
@GoodNike Thanks for the review, I'm merging this change but current camera system should be completely reviewed, probably this module will be completely rewritten (or removed) at some point.

PR #1978, opened 2021-09-06
Fix dynamic library issues on Macos

schveiguy:
On my system, when linking against the 3.7.0 binary downloaded library, the name recorded inside the library is `libraylib..dylib` instead of `libraylib.370.dylib` as the installed symlink is named.

This is due to the use of `$VERSION` which is not set inside the Makefile. I think this was a typo, because there is no symlink built that uses that name.

The binary download for 3.0.0 works fine, but when I build locally 3.0.0, the same problem occurs. I feel the person who generated the raylib binary release must have hand-fixed the issue for the 3.0.0 version, but not the 3.7.0 version.

In addition, adding the `@rpath/` prefix allows you to use DYLIB_LIBRARY_PATH or LD_LIBRARY_PATH to refer to the dynamic library. If you look at the 3.0.0 binary release, it has this prefix in it.

For sure the change from `VERSION` to `RAYLIB_API_VERSION` is correct. I'm not sure what the intent of the original author of this makefile had for the `@rpath` name prefix, but clearly whoever built the binary release thought it should be in there. Without that, you must have the dylib in your local path.

I think the 3.7.0 binary should also be fixed, so that one doesn't have to wait for a new release to have this corrected. Although, this is against master, I'm not sure if you maintain a branch for 3.7.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -496,7 +496,7 @@ else
             endif
             ifeq ($(PLATFORM_OS),OSX)
 				$(CC) -dynamiclib -o $(RAYLIB_RELEASE_PATH)/lib$(RAYLIB_LIB_NAME).$(RAYLIB_VERSION).dylib $(OBJS) $(LDFLAGS) -compatibility_version $(RAYLIB_API_VERSION) -current_version $(RAYLIB_VERSION) -framework OpenGL -framework Cocoa -framework IOKit -framework CoreAudio -framework CoreVideo
-				install_name_tool -id "lib$(RAYLIB_LIB_NAME).$(VERSION).dylib" $(RAYLIB_RELEASE_PATH)/lib$(RAYLIB_LIB_NAME).$(RAYLIB_VERSION).dylib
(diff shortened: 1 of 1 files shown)


raysan5:
@schveiguy thanks for tracking and reviewing this issue! I'm afraid there is no 3.7 branch and current master branch is already set as `4.0-dev`. Hopefully next version will be released soon.

schveiguy:
Well, that's OK. But whomever maintains the macos binary binding may want to manually run that `install-name-tool` command on the dylib in the tarball. Right now, the dynamic lib loader is not going to find it, because you have to name it `libraylib..dylib` which nobody is going to do (I confirmed that if I symlinked a `libraylib..dylib` in my current directory, it would run).

From what I understand, that metadata id tag in the dylib is what tells the dynamic loader where it can find the required dynamic lib. So it links fine, but then goes searching for something that will not exist when trying to run.

Not sure if those release artifacts can be edited after the tag is formed in github...

In my case, I'm the only one who I know uses raylib on macos, so I'm covered ;) All my students are using Windows.

PR #1981, opened 2021-09-08
Vox loaded

procfxgen:
* vox_loader.h -> Removed raylib dependencies
* vox_loader.h -> Reverse Y<>Z for left to right handed system
* models/models_magicavoxel_loading.c -> fix model center


raysan5:
@procfxgen thanks for the review! Still some details:

 - In case no `VOX_MALLOC` was defined, you could add some security checks:
```c
// Allow custom memory allocators
#ifndef VOX_MALLOC
    #define VOX_MALLOC(sz)       malloc(sz)
#endif
#ifndef VOX_CALLOC
    #define VOX_CALLOC(n,sz)     calloc(n,sz)
#endif
#ifndef VOX_REALLOC
    #define VOX_REALLOC(ptr,sz)  realloc(ptr,sz)
#endif
#ifndef VOX_FREE
    #define VOX_FREE(ptr)        free(ptr)
#endif
```

 - There is still a raylib dependency: `LoadFileData()` and `UnloadFileData()`

You can just move those calls to `LoadVOX()` in `models.c` and change `Vox_LoadFileName()` to:
```c 
void Vox_LoadFromMemory(const char* voxData, unsigned int voxDataSize, VoxArray3D* voxarray);
```
That way the library will be completely self-contained and portable.

There are some additional mechanisms to support `Vox_LoadFileName()` and call internally `Vox_LoadFromMemory()` but I think providing the memory loader would be enough for most users, also considering the usual size of .vox files.

orcmid:
I favor the completely-maintained arrangement.

> There are some additional mechanisms to support `Vox_LoadFileName()` and call internally `Vox_LoadFromMemory()` but I think providing the memory loader would be enough for most users, also considering the usual size of .vox files.

In this case, would "Vox_LoadFromFile()' be more parallel with 'Vox_LoadFromMemory()' ?

raysan5:
@orcmid Yes, API could provide both `Vox_LoadFromFile()` and `Vox_LoadFromMemory()`, first one using second one internally and implementing the file loading code, only in case `INCLUDE_STDIO` is supported. It also allows the option for custom file loaders, for example for virtual file systems (i.e. PhysFS).

raysan5:
@procfxgen Thanks for the improvements! I'm merging and reviewing this PR.

PR #1984, opened 2021-09-09
[models] Fixed loading OBJ models with no materials.

MrDiver:
Just quick fix to update `matFace` to the correct value when no materials are present on the mesh.


raysan5:
@MrDiver thanks for the review! 👍😄

630Studios:
Unless I am mistaken this fix is incorrect and will re-introduce issue #1809 where a model has more materials than it does meshes resulting in a crash when attempting to render the mesh.

PR #1985, opened 2021-09-10
Add find opengl for widnows too.

object71:


raysan5:
@object71 Thank you very much for the quick fix! :D

PR #1986, opened 2021-09-11
Implemented GetGamepadName() for emscripten

nbarkhina:
When building for HTML the GetGamepadName() function currently returns null:

ray2

This implements that function by capturing it when the device is detected from the gamepadEvent object. Now the name will return correctly:

ray1

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -461,6 +461,9 @@ typedef struct CoreData {
             int streamId[MAX_GAMEPADS];     // Gamepad device file descriptor
             char name[64];                  // Gamepad name holder
 #endif
+#if defined(PLATFORM_WEB)
+            char gamepadNames[MAX_GAMEPADS][64];
+#endif
         } Gamepad;
     } Input;
     struct {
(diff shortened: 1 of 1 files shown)


src/core.c
```diff
@@ -460,6 +460,9 @@ typedef struct CoreData {
             pthread_t threadId;             // Gamepad reading thread id
             int streamId[MAX_GAMEPADS];     // Gamepad device file descriptor
             char name[64];                  // Gamepad name holder
```
raysan5, line 462: Instead of creating an additional variable, this one can be used just adding the array:
```c
char name[MAX_GAMEPADS][64];
```

nbarkhina: Yes ok I updated it to use the existing variable which is now an array. It exposes these extra properties to the PLATFORM_WEB version is this ok?

image

raysan5: yeah, I think it's ok for now

raysan5:
@nbarkhina Thanks for the improvement! :D

Issue #1982, opened 2021-09-09
[core] Bad GetScreenWidth/GetScreenHeight values after resizing windows with FLAG_WINDOW_HIGHDPI


ArnaudValensi:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Hi,
When using FLAG_WINDOW_HIGHDPI, the GetScreenWidth/GetScreenHeight functions are giving bad values after a window resize. See the video below.
In this example, the GuiStatusBar disappears after a window resize.

### Environment

- Windows 10 pro
- Resolution: 3840x2160 with 150% scaling

### Issue Screenshot

https://user-images.githubusercontent.com/604486/132603229-9664985d-ccb6-4d52-a4aa-bd05bdd98cac.mp4

### Code Example

```
    SetConfigFlags(FLAG_WINDOW_RESIZABLE | FLAG_WINDOW_HIGHDPI);
    InitWindow(1280, 720, "Test");

    while (!WindowShouldClose()) {
        BeginDrawing();
        ClearBackground(DARKPURPLE);

        int width = GetScreenWidth();
        int height = GetScreenHeight();
        Vector2 scale = GetWindowScaleDPI();

        DrawText(TextFormat("screen size: %d, %d\n", width, height), 0, 0, 32, WHITE);
        DrawText(TextFormat("window scale dpi: %f, %f\n", scale.x, scale.y), 0, 40, 32, WHITE);

        GuiStatusBar(
            (Rectangle){0, GetScreenHeight() - 40, GetScreenWidth(), 40},
            "This is a status bar");

        EndDrawing();
    }

    CloseWindow();
```

You can reproduce by doing a `make run` in the following repo https://github.com/ArnaudValensi/raylib-highdpi-issue

ArnaudValensi:
When working with high dpi, we have the concept of physical size and logical size.
The logical size is the virtual size, independent of the dpi scale factor, it doesn't change when the dpi scale factor change.
The physical size is the real number of pixels used by the screen (logical size * dpi scale).

Here it is confusing if GetScreenWidth/GetScreenHeight are supposed to be the logical or the physical size, but since at startup they return the logical size, I guess it is supposed to also return the logical size after a window resize.

A better API would be to have a different set of functions with explicit naming:
- GetLogicalScreenWidth / GetLogicalScreenHeight
[...]

Issue #1931, opened 2021-08-18
[Core] Scissor Mode & High DPI Issue


630Studios:
### Issue description
Per discord user Toadd,  Scissor Mode does not work properly when HIGH_DPI is enabled. It clips the wrong area.

raysan5:
Need more detailed info to review this issue. Please follow the provided template. Closing it.

Issue #1086, opened 2020-01-29
[core] HighDPI not considered while using RenderTexture


raysan5:
This issue is directly related to #837.

When creating a RenderTexture on a HighDPI application:
```c
// Create a RenderTexture2D to be used for render to texture
RenderTexture2D target = LoadRenderTexture(GetScreenWidth(), GetScreenHeight());
```
`GetScreenWidth()` and `GetScreenHeight()` actually return the original screen size requested by the user, despite that everything drawn is scaled to match HighDPI. So, we end up with a RenderTexture smaller than the actual render size.

One possible solution would be to consider the `scaleFactor` directly on `GetScreenWidth()` and `GetScreenHeight()` but if those calls are used in game logic for some maths it will lend to logic errors, because scaling is done automatically only on drawing.

Another solution would be directly exposing the `HighDPI scaleFactor` and let the user manage everything.

Feedback is welcome!

jozanza:
I love the idea of exposing the DPI to the user. I ran into this issue, and am working off a fork that exposes this info via
```c
float GetWindowDPI(void).
``` 
I found it necessary for recreating the render texture when the window is moved to a new monitor with a different DPI or when the window is resized.

raysan5:
Just for reference, [glfwGetMonitorContentScale()](https://www.glfw.org/docs/3.3/monitor_guide.html#monitor_scale) is the related function.

ismyhc:
@raysan5 Im very interested in this feature. I agree with @jozanza . Exposing the dpi seems like a good solution. Im about to do this locally now.

ismyhc:
Hmmm. I was going to try and do this, but I'm getting an error. Here is what I have done to try and expose the dpi.

core.c
```
Vector2 GetWindowDPI(void)
{

#if defined(PLATFORM_DESKTOP)
    float xdpi;
    float ydpi;
    glfwGetMonitorContentScale(CORE.Window.handle, &xdpi, &ydpi);
    return (Vector2){ xdpi, ydpi };
#endif
    return (Vector2){ 1.0f, 1.0f };
}

```

And here is the error I get when calling the function from my code.

```
WARNING: GLFW: Error: 65544 Description: Cocoa: Failed to find a screen for monitor
```

Any ideas?

@jozanza would you mind sharing your implementation in your fork?

ismyhc:
Success! Sorry I figured it out. I was passing the window handle. The glfwGetMonitorContentScale() actually was expecting a monitor pointer.

[...]

Issue #837, opened 2019-05-15
[examples] shaders_julia_set doesn't fill screen on Macbook Retina


a3f:
<img width="912" alt="Screen Shot 2019-05-15 at 18 01 08" src="https://user-images.githubusercontent.com/5204368/57790752-aa924800-773b-11e9-97f7-0405b8dba31e.png">

I am running raylib master with macOS 10.14.4 on a macbook pro retina.

raysan5:
@a3f is high-dpi support flag enabled?

a3f:
`SUPPORT_HIGH_DPI` is defined by default and if I manually undef it, the texture moves:
<img width="912" alt="Screen Shot 2019-05-15 at 23 44 03" src="https://user-images.githubusercontent.com/5204368/57812106-5c496d00-776c-11e9-84a0-a46ca112dd8e.png">

In both cases, the console output is the same:

```
INFO: Display size: 1680 x 1050
INFO: Render size: 800 x 450
INFO: Screen size: 800 x 450
INFO: Viewport offsets: 0, 0
```

raysan5:
@a3f, does it only happen with this example?

raysan5:
Replaced `gl_fragCoord` coordinates by provided `fragTexCoord` (that refer to the full screen texture drawing) on commit https://github.com/raysan5/raylib/commit/9fd410b8a87ede4b40be60ec5043090e2a86c6fe

a3f:
> @a3f, does it only happen with this example?

Just ran them and it happens as well with:

* `shaders_custom_uniform`, `shaders_eratosthenes`, `shaders_postprocessing`

The others are unaffected.

> Replaced gl_fragCoord coordinates by provided fragTexCoord (that refer to the full screen texture drawing) on commit 9fd410b

Still looks the same:
<img width="912" alt="Screen Shot 2019-05-17 at 20 38 34" src="https://user-images.githubusercontent.com/5204368/57949130-d18b7e00-78e3-11e9-8bd4-7a3151907384.png">

raysan5:
ok, I detected the issue, the screenWidth-screenHeight provided to those shaders do not match render size (scaled to HighDPI)

MikeDX:
As an additional feature to this bug, the following code causes the text to be zoomed 2x

```c
BeginTextureMode(F_renderTarget);

    ClearBackground(colour);

    BeginMode3D(camera);
[...]

PR #1987, opened 2021-09-12
Fix rendering issue after window resize #1982

ArnaudValensi:
At the initialization `CORE.Window.currentFbo.width` and `CORE.Window.render.width` was set to the variable passed from InitWindow which is the virtual width, but the framebuffer size created by glfw is `width * dpi scale`.

After resize `CORE.Window.screen.width` was set to the framebuffer size which is the physical width instead of the logical width (on all platform except macos).

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -1537,13 +1537,25 @@ void SetWindowSize(int width, int height)
 // Get current screen width
 int GetScreenWidth(void)
 {
-    return CORE.Window.currentFbo.width;
+    return CORE.Window.screen.width;
 }
 
 // Get current screen height
 int GetScreenHeight(void)
 {
-    return CORE.Window.currentFbo.height;
+    return CORE.Window.screen.height;
+}
+
+// Get current render width which is equal to screen width * dpi scale
+int GetRenderWidth(void)
+{
+    return CORE.Window.render.width;
+}
+
+// Get current screen height which is equal to screen height * dpi scale
+int GetRenderHeight(void)
+{
+    return CORE.Window.render.height;
 }
 
 // Get native window handle
@@ -3820,16 +3832,6 @@ static bool InitGraphicsDevice(int width, int height)
         TRACELOG(LOG_WARNING, "GLFW: Failed to initialize Window");
         return false;
     }
-    else
-    {
-        TRACELOG(LOG_INFO, "DISPLAY: Device initialized successfully");
-#if defined(PLATFORM_DESKTOP)
-        TRACELOG(LOG_INFO, "    > Display size: %i x %i", CORE.Window.display.width, CORE.Window.display.height);
-#endif
-        TRACELOG(LOG_INFO, "    > Render size:  %i x %i", CORE.Window.render.width, CORE.Window.render.height);
-        TRACELOG(LOG_INFO, "    > Screen size:  %i x %i", CORE.Window.screen.width, CORE.Window.screen.height);
-        TRACELOG(LOG_INFO, "    > Viewport offsets: %i, %i", CORE.Window.renderOffset.x, CORE.Window.renderOffset.y);
-    }
 
     // Set window callback events
     glfwSetWindowSizeCallback(CORE.Window.handle, WindowSizeCallback);      // NOTE: Resizing not allowed by default!
@@ -4273,12 +4275,8 @@ static bool InitGraphicsDevice(int width, int height)
     rlLoadExtensions(eglGetProcAddress);
 #endif
 
-    // Initialize OpenGL context (states and resources)
-    // NOTE: CORE.Window.screen.width and CORE.Window.screen.height not used, just stored as globals in rlgl
-    rlglInit(CORE.Window.screen.width, CORE.Window.screen.height);
-
-    int fbWidth = CORE.Window.render.width;
-    int fbHeight = CORE.Window.render.height;
+    int fbWidth = CORE.Window.screen.width;
+    int fbHeight = CORE.Window.screen.height;
 
 #if defined(PLATFORM_DESKTOP)
     if ((CORE.Window.flags & FLAG_WINDOW_HIGHDPI) > 0)
@@ -4297,11 +4295,25 @@ static bool InitGraphicsDevice(int width, int height)
     }
 #endif
 
+    CORE.Window.currentFbo.width = fbWidth;
+    CORE.Window.currentFbo.height = fbHeight;
+    CORE.Window.render.width = CORE.Window.currentFbo.width;
+    CORE.Window.render.height = CORE.Window.currentFbo.height;
+
+    // Initialize OpenGL context (states and resources)
+    // NOTE: CORE.Window.currentFbo.width and CORE.Window.currentFbo.height not used, just stored as globals in rlgl
+    rlglInit(CORE.Window.currentFbo.width, CORE.Window.currentFbo.height);
+
     // Setup default viewport
     SetupViewport(fbWidth, fbHeight);
 
-    CORE.Window.currentFbo.width = CORE.Window.screen.width;
-    CORE.Window.currentFbo.height = CORE.Window.screen.height;
+    TRACELOG(LOG_INFO, "DISPLAY: Device initialized successfully");
+#if defined(PLATFORM_DESKTOP)
+    TRACELOG(LOG_INFO, "    > Display size: %i x %i", CORE.Window.display.width, CORE.Window.display.height);
+#endif
+    TRACELOG(LOG_INFO, "    > Screen size:  %i x %i", CORE.Window.screen.width, CORE.Window.screen.height);
+    TRACELOG(LOG_INFO, "    > Render size:  %i x %i", CORE.Window.render.width, CORE.Window.render.height);
+    TRACELOG(LOG_INFO, "    > Viewport offsets: %i, %i", CORE.Window.renderOffset.x, CORE.Window.renderOffset.y);
 
     ClearBackground(RAYWHITE);      // Default background color for raylib games :P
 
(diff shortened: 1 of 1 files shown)


ArnaudValensi:
Fix issue https://github.com/raysan5/raylib/issues/1982

src/core.c
```diff
@@ -1537,13 +1537,25 @@ void SetWindowSize(int width, int height)
...
-    return CORE.Window.currentFbo.width;
+    return CORE.Window.screen.width;
 }
 
 // Get current screen height
 int GetScreenHeight(void)
 {
-    return CORE.Window.currentFbo.height;
+    return CORE.Window.screen.height;
+}
```
ArnaudValensi, line 1547: I think we expect to get back the logical size from these functions, so it returns the screen size now.
I've created two more functions (GetRenderWidth/GetRenderHeight) in case we want the physical size.
They all return the same size on macos though, maybe I should add this the comment.

src/core.c
```diff
@@ -3820,16 +3832,6 @@ static bool InitGraphicsDevice(int width, int height)
...
     }
-    else
-    {
-        TRACELOG(LOG_INFO, "DISPLAY: Device initialized successfully");
-#if defined(PLATFORM_DESKTOP)
-        TRACELOG(LOG_INFO, "    > Display size: %i x %i", CORE.Window.display.width, CORE.Window.display.height);
-#endif
-        TRACELOG(LOG_INFO, "    > Render size:  %i x %i", CORE.Window.render.width, CORE.Window.render.height);
-        TRACELOG(LOG_INFO, "    > Screen size:  %i x %i", CORE.Window.screen.width, CORE.Window.screen.height);
-        TRACELOG(LOG_INFO, "    > Viewport offsets: %i, %i", CORE.Window.renderOffset.x, CORE.Window.renderOffset.y);
```
ArnaudValensi, line 3831: I moved these logs because we only have the real `CORE.Window.render` size later on.

src/core.c
```diff
@@ -4297,11 +4295,25 @@ static bool InitGraphicsDevice(int width, int height)
     }
 #endif
 
+    CORE.Window.currentFbo.width = fbWidth;
+    CORE.Window.currentFbo.height = fbHeight;
+    CORE.Window.render.width = CORE.Window.currentFbo.width;
+    CORE.Window.render.height = CORE.Window.currentFbo.height;
```
ArnaudValensi, line 4301: These four lines are the most important part. Before this change, `CORE.Window.currentFbo` and `CORE.Window.render.width` were filed with the `width` and `height` passed in parameter by the user to this function. It was not correct because when creating a window at a specific size, glfw create a framebuffer with the size `window size * dpi scale factor` (for all desktop platform except macos).

src/core.c
```diff
@@ -4297,11 +4295,25 @@ static bool InitGraphicsDevice(int width, int height)
...
+
+    // Initialize OpenGL context (states and resources)
+    // NOTE: CORE.Window.currentFbo.width and CORE.Window.currentFbo.height not used, just stored as globals in rlgl
+    rlglInit(CORE.Window.currentFbo.width, CORE.Window.currentFbo.height);
```
ArnaudValensi, line 4305: Moved this line here so we can call it with the correct values.

src/core.c
```diff
@@ -4865,10 +4877,17 @@ static void WindowSizeCallback(GLFWwindow *window, int width, int height)
...
+    Vector2 windowScaleDPI = GetWindowScaleDPI();
+
     // Set current screen size
+#if defined(__APPLE__)
     CORE.Window.screen.width = width;
     CORE.Window.screen.height = height;
-    
+#else
+    CORE.Window.screen.width = width / windowScaleDPI.x;
+    CORE.Window.screen.height = height / windowScaleDPI.y;
```
ArnaudValensi, line 4888: This callback gives the physical size (framebuffer size), so we need to divide by the dpi scale factor.

raysan5:
@ArnaudValensi this PR needs to be reviewed very carefully because it can break other things. Some situations to review:

 - When user do not want to use HighDPI despite the monitor is HighDPI (not enabled `FLAG_WINDOW_HIGHDPI`)
 - When using `BeginTextureMode(RenderTexture)`, currentFBO is stored for further retrieval (and the size is independent of the DPI because is defined by user)
 - When scaling the Window on HTML5 (`EmscriptenResizeCallback`)

ArnaudValensi:
Ok, let me create a checklist:

- [x] When user do not want to use HighDPI despite the monitor is HighDPI (not enabled FLAG_WINDOW_HIGHDPI)
- [ ] When using BeginTextureMode(RenderTexture), currentFBO is stored for further retrieval (and the size is independent of the DPI because is defined by user)
- [ ] When scaling the Window on HTML5 (EmscriptenResizeCallback)
- [ ] Mouse/Touch input interaction when using HighDPI

raysan5:
@ArnaudValensi Just note that those are only some things that should be reviewed for this change, they could be working ok. I reviewed HighDPI several times in the past and I always found some side case where things do not work as expected.

For reference: https://github.com/raysan5/raylib/issues/1086, https://github.com/raysan5/raylib/issues/837, https://github.com/raysan5/raylib/issues/814, https://github.com/raysan5/raylib/issues/1931

When possible, I think it's a good idea to give control of the HighDPI to the user (so `GetWindowScaleDPI()` is provided).

ArnaudValensi:
@raysan5 Ok, thank you for the references.

Maybe these recurring problems/regressions are a good sign that some automated testing could be nice to be implemented. It would make us more confident about the features and the changes we make. I know it is not easy because automated tests often need to be headless, but it is not impossible. We can take a look at how Monogame implemented their tests.
What do you think?

> When possible, I think it's a good idea to give control of the HighDPI to the user (so GetWindowScaleDPI() is provided).

Are you suggesting to not take into account the dpi scaling factor in some cases?
Or are you suggesting that you are not very comfortable with this PR? (I would understand)

raysan5:
Hi @ArnaudValensi, I'm comfortable with this PR, HighDPI has given problems for some time (specially on macOS) but I'm a bit afraid of the side problems that could arise. HighDPI has been reviewed several times and it seems there is always some side case that breaks or does not work as expected.

For example, if HighDPI is enabled and user creates a (800, 600) Window, actually the internal framebuffer created could be (800*1.5, 600*1.5) but a novel user expects to work with the size of (800, 600) for the game logic. An advance user could be creating a RenderTexture of (800, 600) and notice that when drawing it, the Window is bigger (x1.5) than the RenderTexture, in those cases I think it's up to the user to manage RenderTexture drawing considering the `GetWindowScaleDPI()`.

ArnaudValensi:
@raysan5 I have made the following tests on windows/linux/macos/web:

https://user-images.githubusercontent.com/604486/133172822-7134de4a-bfdc-46df-977a-c8c2ba391e54.mp4

Everything is fine. The web platform doesn't seem to use hdpi scale factor at all, so I did not modify its code.
I also tried the web on mobile, but I did not try ios or android builds.

I still need to check `BeginTextureMode`.

All the checks are made using this repository: https://github.com/ArnaudValensi/raylib-highdpi-issue
This raylib libraries corresponding to this PR are directly on the repository, so you can check by doing:
`make run` on window, linux or macos
`make game-web && make run-web` on any platform to build the web version. Docker is needed though.

ArnaudValensi:
I've found a missing modification in EndTextureMode https://github.com/raysan5/raylib/pull/1987/commits/8b1dd7af1ce4ca570f4c1c71a5f577835fd13898

I've checked the example `textures\textures_mouse_painting.c` and it works fine.

`LoadRenderTexture(width, height)` does not use the dpi scale factor, but like you said, we can use `GetWindowScaleDPI()` if needed.

I will check the scissor mode.

ArnaudValensi:
I discovered a bug with the scissor mode, it was here before this PR:

https://user-images.githubusercontent.com/604486/133179103-024ef5a6-93e5-4e5b-a92b-a4e2460aa808.mp4

I fixed it in https://github.com/raysan5/raylib/pull/1987/commits/9187180d55e7414d7438274b6c317df1af1231a6


https://user-images.githubusercontent.com/604486/133179157-34200599-9396-4adc-ac27-3650b8ce4491.mp4

It solves this issue https://github.com/raysan5/raylib/issues/1931

ArnaudValensi:
The julia example works fine with hdpi enabled (https://github.com/raysan5/raylib/issues/837), I've tried on windows and macos.

My only left concern is that I don't know how the hdpi behave on mobile devices, but maybe it can be the scope of another PR if there is some trouble.

So, we might be good to go. 
Also, if there is some other troubles around hdpi in the future, I can be available.

ArnaudValensi:
I think for the scissor, we should only multiply by the scale if the hdpi flag is set. I will make the change Wednesday.

raysan5:
@ArnaudValensi Very nice! Thank you very much for the review of all the HighDPI involved issues and the detailed report! About resize event on mobile using a NativeActivity (Android), I think it's not relevant but it's important to support the touch input properly.

ArnaudValensi:
@raysan5 I fully fixed the scissor, it now works with and without the high dpi flag.

I wanted to try if this PR works on Android but I was not able to compile raylib due to this issue https://github.com/raysan5/raylib/issues/1997
I don't have the time to figure out why. But if compilation on Android works again, I can finalize the checks to validate this PR.

raysan5:
@ArnaudValensi I'm afraid I renamed this module and now changes can not be merged due to conflicts... Could you open a new PR with those changes for merging? By the way, Android compilation should be working again.

Sorry for the inconvenience.

ArnaudValensi:
Ok, I will try to find the time this week.

ArnaudValensi:
I have rebased this PR. 
I still need to check on Android and we should be good.

raysan5:
@ArnaudValensi Nice! Thank you very much for reviewing it again! Please, let me know when ready to merge! :D

PR #1991, opened 2021-09-14
Create CODE_OF_CONDUCT.md

ChakibChemso:


raysan5:
@ChakibChemso No plans to add this kind of document at this moment. Actually, it was added and later removed some years ago.

orcmid:
I have no opinion and defer to raysan5 completely. 

I also note that the document has a direct link to (https://www.raylib.com/index.html). There is not anything there that supports handling of questions or complaints around the proposed code of conduct.  There is also no identification of "community leaders."

ChakibChemso:
ok!

PR #1993, opened 2021-09-16
Prevent INVALID_SOCKET to be defined twice on linux

MrSwedish:
i got an error because of this
so pls


raysan5:
@MrSwedish Thanks for the review, actually this module require a complete review.

PR #1994, opened 2021-09-17
Add SetRandomSeed(unsigned int seed) function

TommiSinivuo:
This is a proposal for a new public API function `SetRandomSeed(unsigned int seed)`.

**Rationale:**

Specifying a fixed seed for the random number generator is often used in games for various reasons, for example when replaying the same randomly generated world or when competing in games that are random by nature (and the random element is undesired).

By adding an API function for seeding the random number generator we solve two different problems regarding the use case:

**1.** The underlying RNG implementation does not leak to client code (as would be the case if we called `srand` directly from the client code). This prevents client code from breaking in the case where the implementation would change.

**2.** Seeding the RNG would be possible and simple from other programming languages (especially in cases where calling libc functions is non-trivial). 

Also, since raylib provides a function for fetching random numbers (`GetRandomValue`) it would feel natural for it to provide a function for seeding the random number generator as well (which would be totally optional to use).

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -2669,6 +2669,12 @@ int GetRandomValue(int min, int max)
     return (rand()%(abs(max - min) + 1) + min);
 }
 
+// Set the seed for the random number generator
+void SetRandomSeed(unsigned int seed)
+{
+    srand(seed);
+}
+
 // Check if the file exists
 bool FileExists(const char *fileName)
 {
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1009,6 +1009,7 @@ RLAPI double GetTime(void);                                       // Get elapsed
 
 // Misc. functions
 RLAPI int GetRandomValue(int min, int max);                       // Get a random value between min and max (both included)
+RLAPI void SetRandomSeed(unsigned int seed);                      // Set the seed for the random number generator
 RLAPI void TakeScreenshot(const char *fileName);                  // Takes a screenshot of current screen (filename extension defines format)
 RLAPI void SetConfigFlags(unsigned int flags);                    // Setup init configuration flags (view FLAGS)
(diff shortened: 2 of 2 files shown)


orcmid:
It seems appropriate to have `void SetRandomSeed(unsigned int)` as a companion to the raylib function, with the caveat that `GetRandomValue` is used internal to some raylib functions.

Considering that `int rand(void)`, `void srand(unsigned seed)`, and `RAND_MAX` are quite well-defined in `<stlib.h>`, with default `srand(1)` understood, I am unclear exactly what leaking is involved.

With `RLAPI int GetRandomValue(int min, int max)` defined using a single `rand()` the issue of leaking is a bit murky.  `GetRandomValue()` is used in some internal raylib functions and `GetRandomValue(0, RAND_MAX)` is  not assured to be the same as `rand()` when `RAND_MAX` is the same as `MAX_INT`.

TommiSinivuo:
@orcmid Yes, you're right that the potential "problem" of using `srand` in client code to seed is rather hypothetical, since I don't see any reason why Raylib would ever need to move away from using `rand` and `srand` from `stdlib.h`. I think I could reword that point to mean something more like "the user of Raylib wouldn't need to know how Raylib produces random values in order to provide a fixed seed". I think that actually makes more sense to myself too.

Good point that Raylib seems to be using `GetRandomValue` internally (apparently in `GenImageWhiteNoise`, `GenImageCellular`). It should probably be noted that providing a fixed seed will affect those functions as well (naturally when either using this potential `SetRandomSeed` function or by calling  `srand` manually in client code).

raysan5:
@TommiSinivuo @orcmid Thank you for the addition and the detailed reasoning about it.

PR #1995, opened 2021-09-18
Fix mappings.h issue from glfw update

ChrisDill:
- Ran GenerateMappings.cmake to restore mapping changes


raysan5:
@ChrisDill Thank you very much for updating the mapping! I always forget about it when updating GLFW!

PR #2000, opened 2021-09-21
Kotlin binding (sort of)

electronstudio:
Not sure if worth adding..

People sometimes ask if there is a Kotlin/native binding so I tried to create one.  Turns out that Kotlin/native binds to C libraries on its own automatically, so I didn't really have to do anything, but this is an example of how to set everything up.  Maybe people looking for Kotlin/native bindings will want to copy this.


raysan5:
@electronstudio Thanks for the addition! In any case, it could be useful as a reference!

PR #2001, opened 2021-09-21
Add function DrawCubeTextureRec

tdgroot:
I am working on a project where I draw cubes with textures from a sprite sheet and had to create a new render function to achieve that.

If this feature is out of scope, I'm okay with that, but I thought it would be worth upstreaming this function!


raysan5:
@tdgroot It seems those changes were done before models.c -> rmodels.c rename, please, could you resubmit them?

tdgroot:
@raysan5 will do whenever I have the time :). I also figured that it'd be nice for the example to render the complete texture in 2D mode to demonstrate what the sheet looks like.

I'll let you know when the branch is updated.

raysan5:
@tdgroot Ok, thank you very much! And sorry for the inconvenience! By the way, I think the example could be skipped for this addition...

tdgroot:
@raysan5 i thought the example would be worth it, because the function doesn't seem so trivial. 

Also in terms of API design, perhaps the following interface would be better?
``` c
RLAPI void DrawCubeTextureRegion(Texture2D texture, Rectangle region, Vector3 position, float width, float height, float length, Color color);
```

raysan5:
@tdgroot About the example, I prefer to avoid it for just one function, it requires maintenance, testing, porting to web, formating... 

About the function formating, I prefer it to be aligned with other similar functions:
```c
void DrawTextureRec(Texture2D texture, Rectangle source, Vector2 position, Color tint);
void DrawBillboardRec(Camera camera, Texture2D texture, Rectangle source, Vector3 position, Vector2 size, Color tint);
```

Also, I think that it can be a bit limiting for some users requiring that kind of functionality... Probably lots of users would require a diferent rectangle per face... that was the reason it was not added in the first place...

tdgroot:
@raysan5 I updated the branch and removed the example. 

> Also, I think that it can be a bit limiting for some users requiring that kind of functionality... Probably lots of users would require a diferent rectangle per face... that was the reason it was not added in the first place...

To me it feels like a nice addition to the library, but again:
>  If this feature is out of scope, I'm okay with that, but I thought it would be worth upstreaming this function!

raysan5:
@tdgroot Thanks for the review! 👍😄

PR #2002, opened 2021-09-21
Use unsigned int for animation count.

rcorre:
LoadModelAnimations takes an `int` for the animation count parameter.
The animation count should never be negative, so it makes sense to
specify it as unsigned in the API. This matches the API for
UnloadModelAnimations, which expects an unsigned int. Both GLTF and IQMM
also store the animation count internally as unsigned, and we were
casting to a signed int for no reason.

GLTF actually uses `size_t` internally, so we're technically risking
overflow, but having 2^32 animations seems unlikely.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1436,7 +1436,7 @@ RLAPI void SetMaterialTexture(Material *material, int mapType, Texture2D texture
 RLAPI void SetModelMeshMaterial(Model *model, int meshId, int materialId);                  // Set material for a mesh
 
 // Model animations loading/unloading functions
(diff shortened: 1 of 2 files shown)


raysan5:
@rcorre It seems those changes were done before `models.c` -> `rmodels.c` rename, please, could you resubmit them?

rcorre:
@raysan5 done!

rcorre:
Build failing for unrelated reason:

```
rmodels.c: In function ‘LoadImageFromCgltfImage’:
rmodels.c:4412:52: error: ‘size’ undeclared (first use in this function)
 4412 |         rimage = LoadImageFromMemory(".png", data, size);
```

raysan5:
@rcorre Thanks for the quick review!

PR #2004, opened 2021-09-22
Fix use of undeclared variable in rmodels.

rcorre:
Broken by de173a93c86fbda92b7d93a5fc202e80d4a16f24.


raysan5:
@rcorre Thanks! Just send a fix before seing this PR!

PR #2006, opened 2021-09-22
`SeekMusicStream` function

GithubPrankster:
This function allows you to seek into a position in seconds on a music stream, something commonly used for music player programs to allow you to get to a certain point of your tune.

Unfortunately it does not support *module-based* formats **(.XM, .MOD)** as they work pretty differently than most others and did not have functions for seeking anyways. (Even the functions for going back to the start are an addition from **Ray** too anyways)

Any _feedback_ (heh!) would be greatly appreciated, specially as this function could prove quite useful to many.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1656,6 +1656,28 @@ void StopMusicStream(Music music)
     }
 }
 
+// Seek music to a certain position (in seconds)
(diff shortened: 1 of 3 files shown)


raysan5:
@GithubPrankster Thanks! Very nice addition! 👍😄

PR #2007, opened 2021-09-22
Modify TODO to NOTE on Makefile

catmanl:
'# TODO: see 'install' target.' (line 655) should be a note rather than something
to do.


raysan5:
@catmanl Thanks for the review!

PR #2008, opened 2021-09-23
Fix + Module format return for `SeekMusicStream`

GithubPrankster:
`SeekMusicStream` did not update the frames being processed, resulting in looping malfunctioning. This PR fixes this + adding a line to return from the function if the format trying to be seeked is a Module one, as they do not have support for seeking.


raysan5:
@GithubPrankster Thanks for the review!

PR #2010, opened 2021-09-23
Implement loop point in `Music` struct

GithubPrankster:
This allows the user to set an optional variable in the `Music` struct with the position in seconds to which the music will immediately seek to after looping. Thus you could have a track with a beginning that plays once and the remainder always looping.

Feedback would be good as always.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1779,6 +1779,7 @@ void UpdateMusicStream(Music music)
     {
         StopMusicStream(music);                     // Stop music (and reset)
         if (music.looping) PlayMusicStream(music);  // Play again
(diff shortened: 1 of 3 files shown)


raysan5:
@GithubPrankster I'm still thinking about merging or not this change... I feel using a defined Music point for loop seems a very specific use case, I think most raylib users won't benefit from it and the addition of an extra variable to the struct has a perfromance hit when passing `Music` struct around. I keep thinking about it...

GithubPrankster:
Honestly it wouldn't be wrong to turn it down as it is a fairly special modification. I'll just leave the branch available regardless in case someone also wants something similar in their fork of raylib.

raysan5:
I'm closing this PR for now, maybe I'm reviewing it in a future. In any case, it can be a point of reference for users looking for this feature.

PR #2014, opened 2021-09-25
Add Zig build file

TommiSinivuo:
@raysan5 as discussed on Discord:

This commit adds a `build.zig` file in `src/` which can be used by the Zig build system to build Raylib. In it's current form it's able to build Raylib into a static library for Windows and Linux.

This has been tested with Zig 0.8.1 in WSL2 on a Windows machine.

How to use:

`cd src/`
`zig build`

This produces a debug build in `src/`. Add one of the following switches to `zig build` command to specify release build type:
`-Drelease-small=true`
`-Drelease-safe=true`
`-Drelease-fast=true`

The Zig compiler has first class support for cross-compilation. Here are some cross-compilation
Cross-compilation examples:

To windows target:
`zig build -Dtarget=native-windows-gnu`

To linux target:
`zig build -Dtarget=native-linux-gnu`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,10 @@ build
 GPATH
 GRTAGS
 GTAGS
+
+# Zig programming language
+zig-cache/
+zig-out/
+build/
+build-*/
+docgen_tmp/
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,0 +1,58 @@
+const std = @import("std");
+
+pub fn build(b: *std.build.Builder) void {
+    // Standard target options allows the person running `zig build` to choose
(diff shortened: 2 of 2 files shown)


raysan5:
@TommiSinivuo This is a GREAT addition! Thank you very much! :D

nektro:
why was this added in the `src/` folder? it should be in root

TommiSinivuo:
> why was this added in the `src/` folder? it should be in root

It's usually in root, yes.

However, in this case the Makefile is in src so it makes some sense to place other build files there as well.

PR #2017, opened 2021-09-27
[Core] Allow PollEvents to use wait logic

JeffM2501:
This PR moves the test to use the glfw wait logic outside of the PollEvents call and into the EndDrawing. This way people who are using the advanced timing feature can choose to use wait or poll depending on program state not a build flag.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -946,7 +946,7 @@ RLAPI const char *GetClipboardText(void);                         // Get clipboa
 // By default EndDrawing() does this job: draws everything + SwapScreenBuffer() + manage frame timming + PollInputEvents()
 // To avoid that behaviour and control frame processes manually, enable in config.h: SUPPORT_CUSTOM_FRAME_CONTROL
 RLAPI void SwapScreenBuffer(void);                                // Swap back buffer with front buffer (screen drawing)
-RLAPI void PollInputEvents(void);                                 // Register all input events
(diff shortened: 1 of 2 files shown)


JeffM2501:
> Is there a way to do this without making a breaking change to the current API?

Of course there is, there is just no point in doing it. Keeping the API simple is more important.

PollEvents is a new function for 4.0 dev, it is not in a released API. There is no legacy code to break or regressions (since it is new). I do not think it is worth adding more code just to prevent a few people who have been using development code from having to add a Boolean to a function they are probably only calling once. PollEvents is only called by applications doing advanced work using the advanced timing flag. It is not something everyone will have to change.

Raylib breaks things that are used by a wider groups than this each release.

raysan5:
@JeffM2501 I'm afraid I'm not merging this change. It implies adding and extra parameter to a function that most of the time will be called with a `false` value, making the function less clear when directly seen in the code (if not checked the function in raylib). Also, I feel it a bit confusing that you are asking to poll events but actually you want to wait for events. If that function is really required by some user, I prefer to just expose the `WaitEvents()`.

JeffM2501:
Fair enough 😀

raysan5:
@JeffM2501 I'm closing this PR. I'll keep the build flag method for the moment. It seems a specific use-case for some raylib applications...

PR #2018, opened 2021-09-29
[rlgl] restore draw batch state after limit check

mausimus:
The intention behind rlCheckRenderBatchLimit is to occasionally call it to ensure buffer space is available for the next set of vertices to be sent, however if that threshold is reached and rlDrawRenderBatch called to draw existing buffers, the state of all draw batches is reset to default (RL_QUADS and default texture).
This means calling code needs to call rlBegin / rlSetTexture after each rlCheckRenderBatchLimit which returned true; proposed fix automatically restores mode and texture of the active batch so calling code won't need to care if batch did overflow or not.


raysan5:
@mausimus Thanks for the improvement! :D

PR #2020, opened 2021-09-29
[web] fix: reset resizedLastFrame

mausimus:
With new emscripten callbacks, resizedLastFrame is now set on web platform but is never cleared. The fix resets the flag in PollInputEvents.


raysan5:
@mausimus Nice! Thanks!

PR #2023, opened 2021-09-30
Cleanup x64/x86 options in scripts/build-windows.bat

orcmid:
/std:c11 is required to allow extended initializers introduced after c99
/utf-8 and validation aligns with the technology direction where VS Build Tools are used for the desktop
/W3 warning flag saves the beginner from puzzling over slack-byte and linker in-lining messages
/sdl causes alerts to uses of library functions that are deprecated in favor of ones that encourage safe memory usage.


raysan5:
@orcmid Thanks for the review!

PR #2025, opened 2021-10-01
Tidying the core_basic_window.c in projects/scripts

orcmid:
Adjustments specific to that location of the file.
1. The press F6 message and the placement of the compiled executable statement are incorrect for the scripts case.
2. <raylib.h> because "raylib.h" is meaningful only when compiling inside raylib/src/ folder.
3. main(void) to get rid of a novice-confusing warning.


projects/scripts/core_basic_window.c
```diff
@@ -4,32 +4,31 @@
...
+*   It has been re-confirmed with raylib 3.7.0
 *   raylib is licensed under an unmodified zlib/libpng license (View raylib.h for details)
 *
 *   Copyright (c) 2013-2016 Ramon Santamaria (@raysan5)
+*   Simple adjustments on 2021-10-01
 *
 ********************************************************************************************/
 
-#include "raylib.h"
+#include <raylib.h>
```
raysan5, line 21: I'd prefer to use quotes for the include, I only use <> for system libraries

raysan5:
@orcmid Just added a small review detail comment that I'm reviewing myself.

PR #2031, opened 2021-10-03
[rcore] fix rpi4 Failed to get DRM resources 

tomarigr:
card1 is not always the the correct card to use on rpi4
rpi os finds the correct card and links it to /dev/dri/by-path/platform-gpu-card during boot.
this fix makes sure that we always point to the correct card on RPI4/CM4


raysan5:
@tomarigr Nice! Thanks for the improvement! 👍😄

PR #2032, opened 2021-10-04
Revert "reset resizedLastFrame on web platform (#2020)"

mausimus:
Apologies @raysan5 , it looks like my fix for web resize has broken desktop resize, the flag needs to be cleared before polling events as it looks like those events actually set it.

Please kindly revert my fix and I will PR a proper one once I had a chance to test properly.


raysan5:
@mausimus ok, no worries! 👍

PR #2033, opened 2021-10-05
added nelua-raylib and removed raylib-nelua

AKDev21:
removed raylib-nelua because it's the same as nelua-raylib but with some significant improvements :smile_cat:

not to mention, nelua-raylib was added by me in the past #1890


raysan5:
@AKDev21 nice! thanks for the improvement!

PR #2040, opened 2021-10-09
[Documentation] Move older and abandoned bindings to new section at the end of the file

JeffM2501:
This PR, reorders the bindings file to move bindings that are abandoned or use older versions down to the bottom of the file, so that people don't have to search though a ton of old bindings to find what they are looking for.
Bindings more than 2 versions old are considered abandoned (older than v3.5), or if there is a binding for the language that targets a newer ersion.

The file can be seen in markdown here.
https://github.com/JeffM2501/raylib/blob/bindings_update/BINDINGS.md


raysan5:

PR #2049, opened 2021-10-11
Add DrawCylinder(Wires)Ex

Horrowind:
These changes implement a variant of DrawCylinder(Wires), which allows to specify a base and a target point from which the cylinder is then drawn. The API is as follows:
```
RLAPI void DrawCylinder(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color);
RLAPI void DrawCylinderEx(Vector3 startPos, Vector3 endPos, float startRadius, float endRadius, int sides, Color color);
RLAPI void DrawCylinderWires(Vector3 position, float radiusTop, float radiusBottom, float height, int slices, Color color);
RLAPI void DrawCylinderWiresEx(Vector3 startPos, Vector3 endPos, float startRadius, float endRadius, int sides, Color color); 
```
I was unsure about the ordering of the parameters. I think this makes the most sense, but note that I had to exchange the order of the radii when implementing DrawCylinder using DrawCylinderEx, which could be a source for errors.

Besides superseding the original DrawCylinder, I found that these helper functions makes it especially useful to draw lines in 3d with a given thickness (set both radii to some small value and sides=3), as well as drawing arrows (draw a cylinder 3/4 of the way and the final 1/4 by a cone).

The first commit is the implementation of both functions. The second commit changes one of the examples to show the usage of the function by flipping the golden cone over, although it might not be the best showcase.

The last commit changes the implementation of DrawCylinder(Wires) to use the new API. During testing, I noticed, that with these changes the orientation of the cylinder is now different than before, see the blue pyramids in the following screenshots.
Before:
old
After:
new

If this is an issue because it might break existing games, I can try to change that, but at the expense of a concise implementation (I would need to change the `Vector3Perpendicular` into something more complicated).

I only tested on Linux, but I fail to see why there should be problems with other platforms.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -777,30 +777,30 @@ void DrawCylinderEx(Vector3 startPos, Vector3 endPos, float startRadius, float e
     rlCheckRenderBatchLimit(numVertex);
     if(sides < 3) sides = 3;
 
-    Vector3 difference = Vector3Subtract(endPos, startPos);
+    Vector3 direction = { endPos.x - startPos.x, endPos.y - startPos.y, endPos.z - startPos.z };
     // Construct a basis of the base and the top face:
-    Vector3 b1 = Vector3Normalize(Vector3Perpendicular(difference));
-    Vector3 b2 = Vector3Normalize(Vector3CrossProduct(b1, difference));
+    Vector3 b1 = Vector3Normalize(Vector3Perpendicular(direction));
+    Vector3 b2 = Vector3Normalize(Vector3CrossProduct(b1, direction));
 
-    float base_angle = (2.0 * PI) / sides;
+    float baseAngle = (2.0*PI)/sides;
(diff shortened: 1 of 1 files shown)


raysan5:
Added some additional review points, mostly formatting related.

One important concern that worth commenting. No `Draw*()` uses `raymath` functionality internally, would it be possible to avoid it? If no possible I can understand.

raysan5:
> Also do you have a good idea for the case of `startPos == endPos`? I could check that and immediately return, otherwise the result might be unexpected.

yeap, that's ok, no need to send useless vertex to GPU.

raysan5:
@Horrowind That's very nice! Thank you very much for the additions! They look very clean and organized! :D

raysan5:
@Horrowind no worries, send a new PR, please

PR #2051, opened 2021-10-13
Add zig buildfile for examples.

rcorre:
- `zig build` to compile all examples
- `zig build [module]` to compile all examples for a module (e.g. `zig build core`)
- `zig build [example]` to compile _and run_ a particular example (e.g. `zig build core_basic_window`)

You can use `-Dtarget=` to compile for a non-native platform, such as
`zig build -Dtarget=x86_64-windows-gnu` to compile from Linux to
Windows.

This is still WIP, as it assumes `../src/libraylib.a` already exists in `src`. This means it won't work out-of-the-box with `src/build.zig` which puts `libraylib.a` in `zig-out/lib`. Questions:

- Should we merge `src/build.zig` and `examples/build.zig` into a single buildfile, so the `examples` can be aware of the main library build step (and not just have to assume a file exists in a particular place)?
- Should `src/build.zig` use `src` as the output directory, to be consistent with `make`?

### DIFF
--- a/examples/build.zig
+++ b/examples/build.zig
@@ -25,7 +25,11 @@ fn add_module(comptime module: []const u8, b: *std.build.Builder, target: std.zi
         exe.setTarget(target);
         exe.setBuildMode(mode);
         exe.linkLibC();
-        exe.addObjectFile("../src/libraylib.a");
+        exe.addObjectFile(switch (target.getOsTag()) {
+            .windows => "../src/raylib.lib",
+            .linux => "../src/libraylib.a",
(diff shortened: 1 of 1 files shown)


raysan5:
@rcorre Thank you very much! This is an amazing improvement! Thanks also for the detailed explanation. I'm merging in current state.

About the issues with `cgltf`, I was reviewing the glTF loading yesterday but everything was compiling fine. I just tested it right now and it works ok (despite some glTF models are not properly loaded).

PR #2063, opened 2021-10-17
REVIEWED: EXAMPLE: textures_bunnymark

lauchimoon:
This example wasn't added until 2.5, but the description said 1.6.

### DIFF
--- a/examples/textures/textures_bunnymark.c
+++ b/examples/textures/textures_bunnymark.c
@@ -2,7 +2,7 @@
 *
 *   raylib [textures] example - Bunnymark
 *
-*   This example has been created using raylib 1.6 (www.raylib.com)
+*   This example has been created using raylib 2.5 (www.raylib.com)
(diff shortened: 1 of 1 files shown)


catmanl:
I was also wondering if the `wabbit_alpha` asset was made by someone working on raylib, since, on further investigation of this asset it was initially made in 2012/2013 ([found here](https://tineye.com/search/64811c617c128e3b151ea430ce2c37c0b6006bbc?sort=score&order=desc&page=1); [original image found at](https://github.com/tanis2000/UnityBunnyMark/blob/master/Assets/wabbit_alpha.png)) so we could add credits as in most examples.

mandar1jn:
Question: why did you review to raylib 2.5? And not something like 3.7

raysan5:
@catmanl Actually, this should probably reviewed in all examples, it's not clear the raylib version every example was created with, the minimal raylib version required and the last version the example was updated to. It should probably listed for every example:

 - This example was created with raylib x.x (minimal version required)
 - This example was reviewed with raylib x.x

PR #2064, opened 2021-10-17
Update SetupViewport so that it doesn't subtract offset from render d…

michaelfiber:
…imensions when calling rlViewport


raysan5:
@michaelfiber thanks for the review!

PR #2065, opened 2021-10-17
Fixing the error of loading VOX models.

warzes:
In the previous change, the offset was skipped when reading the file, so loading the VOX model did not work. Fixed it.


raysan5:
@warzes Nice! Thank you very much for the fix! 👍😄

PR #2066, opened 2021-10-17
build.zig: Use addCSourceFiles

sagehane:


raysan5:
@sagehane thanks for the improvement!

PR #2069, opened 2021-10-18
Allow to run any file

yareky:
Change the default behavior of the "build debug" and "build release" tasks. Now we can run any, currently edited, source file.


raysan5:
@yareky Very nice improvement! Thanks!

PR #2072, opened 2021-10-19
Add D bindings to Raylib

ghost:
These bindings were created for D, as to supersede the Scheme48 bindings (issue #2043) which I'm not going to work on.
Scheme48 has not been developed since April 2014 so it's useless to keep these bindings if that's okay.


rc-05:
> I don't understand what makes a stable language useless? Would you prefer Chez Scheme or Racket for current implementations?
> 
> I am not objecting to this contribution. I am just curious about the remark.
> 
> I know there are c99 discrepancies now, and `/std:c11` is better for initializers that exist in some of the raylib files. I don't here any complaints about obsolescence though.

The problem with Scheme48 is that it doesn't let the programmer access C structs specifically which is sometimes needed.
I've opted for D since it's far easier generating bindings to it

PR #2088, opened 2021-10-29
Add Conway's Game of Life compute shader example.

TSnake41:
Conway's Game of Life compute shader example featuring various ways to use SSBO and Compute Shaders ("transfert" shaders, rendering from a SSBO, ...).


examples/shaders/shaders_compute_gol.c
```diff
@@ -0,0 +1,147 @@
...
+  unsigned int w; // width of the filled zone
+  unsigned int enabled; // whether to enable or disable zone
+};
+
+struct GolUpdateSSBO {
+  unsigned int count;
+  struct GolUpdateCmd commands[MAX_BUFFERED_TRANSFERTS];
+};
+
+int main(void)
```
raysan5, line 44: Please, could you review the example and shaders to follow the [coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)? Thanks!

examples/Makefile
```diff
@@ -489,7 +489,8 @@ SHADERS = \
     shaders/shaders_spotlight \
     shaders/shaders_hot_reloading \
     shaders/shaders_mesh_instancing \
-    shaders/shaders_multi_sample2d
+    shaders/shaders_multi_sample2d \
+    shaders/shaders_compute_gol
```
raysan5, line 493: I prefer to avoid it in the Makefile, this examples is OpenGL 4.3 specific so compilation will fail in some platforms.

examples/shaders/resources/shaders/glsl430/golRender.glsl
```diff
@@ -0,0 +1,24 @@
+// Game of Life rendering shader
+// Just renders the content of the ssbo at binding 1 to screen.
+#version 430
```
raysan5, line 3: Could you rename the shaders to follow `snake_case` convention, raylib uses it for external files/resources. 
In this case it should be `gol_render.glsl`

raysan5:
@TSnake41 Added some points to review.

mandar1jn:
It seems like there are some errors in the cmake run. I can’t see what currently since I’m on mobile

raysan5:
@TSnake41 Thanks for this great example! I'm merging but I'll move the example to others, I think it fits better on that category.

PR #2090, opened 2021-10-30
Add missing file extensions for drag-and-drop support

Peter0x44:
When trying to test some of my models, I noticed that .glb was missing from the extensions that it checks for upon dropping a file on the window. This PR also checks for the missing extensions.


Peter0x44:
These were all the extensions I could think to add - are there potentially any I missed?

raysan5:
@Peter0x44 nice improvement! 👍😄

PR #2094, opened 2021-11-02
added another odin binding

AKDev21:
You can change the name to whatever you want to because the binding doesn't have a name.


raysan5:
@AKDev21 thanks for the addition!

PR #2098, opened 2021-11-05
BINDINGS: raylib-cpp now targets raylib 4.0.0

RobLoach:
https://github.com/RobLoach/raylib-cpp/releases/tag/v4.0.0


raysan5:
@RobLoach thanks for the update! 👍😄

PR #2109, opened 2021-11-06
Add DrawTriangleGradient()

KonPet:
Currently it's only possible to draw rectangles with gradients that aren't rotated. You can get around that by for example rendering to a texture and rotating it, but DrawTriangleGradient() skips that and lets you directly draw a triangle with a gradient. if you want to draw a rotated rectangle with a gradient you can just draw two triangles

(Also drawing Triangles with gradiants was impossible before I think)


KonPet:
Raysan told me he doesn't want new functions, so I closed this PR

mgood7123:
i like this

PR #2111, opened 2021-11-06
Fix Render Batch Limit not being checked in DrawRectangleGradientEx()

KonPet:


raysan5:
@KonPet thanks for the improvement!

PR #2124, opened 2021-11-09
[rlgl] Allow GL4.3 capabilities to be shown

GithubPrankster:
Since 4.3 was added the commented out capabilities can be shown.


raysan5:
@GithubPrankster Nice addition! Actually, I think more OpenGL 4.3 specific info could be exposed, about compute shaders and SSBO.

PR #2127, opened 2021-11-11
Enum count entries for `KeyboardKey`, `MouseButton`, and `GamepadButton`

Andidy:
This allows iterating over all possible inputs from the keyboard, mouse, or gamepad more easily without lots of copy pasting. The `MOUSE_BUTTON_NUM_BUTTONS`, `GAMEPAD_BUTTON_NUM_BUTTONS`, and `GAMEPAD_AXIS_NUM_AXIS` entries are robust to the potential future addition of more entries in these enums because they are placed at the end of the enum declaration.  For the `KeyboardKey` enum I used the value of 512. This enum is out of order and so simply placing the `KEY_NUM_KEYS` enum at the end wouldn't be robust to change. Instead I used the value of the `#define MAX_KEYBOARD_KEYS = 512` found in rcore.c used for the length of the internal arrays that store the Keyboard state. A possible alternative for the MouseButton, GamepadButton and GamepadAxis enums would be to use the `#defines` in rcore.c for their internal states as the value of the _NUM_BUTTONS enum entries.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -631,7 +631,8 @@ typedef enum {
     KEY_BACK            = 4,        // Key: Android back button
     KEY_MENU            = 82,       // Key: Android menu button
     KEY_VOLUME_UP       = 24,       // Key: Android volume up button
-    KEY_VOLUME_DOWN     = 25        // Key: Android volume down button
+    KEY_VOLUME_DOWN     = 25,       // Key: Android volume down button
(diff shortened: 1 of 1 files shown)


raysan5:
@Andidy I'm afraid I'm not merging this PR, raylib does not follow that convention in any enum and if it was added, all enums should be reviewed for consistency. Also note those values could depend on user-customization. In any case, they can be exposed through `config.h` for advance users requiring them.

PR #2129, opened 2021-11-11
Lua output format for parser

iskolbin:
This PR adds LUA output format for parser. It's almost identical to JSON, only notable deviation is different function parameters format, for example in Lua:
```lua
      params = {
        {name = "rec", type = "Rectangle"},
        {name = "color", type = "Color"}
      }
```
while in JSON:
```json
      "params": {
          "rec": "Rectangle",
          "color": "Color"
      }
```
The motivation is quite simple: we cannot assume the order of object/table traversing which is crucial for things like function parameters. 

To speed up things a bit I also added Makefile which by default compiles parser and parses raylib.h in all 4 formats. Also you can call
```
make all
```
which parses raylib, raymath, all extras.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ zig-out/
 build/
 build-*/
 docgen_tmp/
+
+# Parser stuff
+parser/raylib_parser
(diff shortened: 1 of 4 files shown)


raysan5:
@iskolbin Thanks for the improvement! :D

PR #2130, opened 2021-11-11
Add sol-vin/raylib-cr to bindings page. 

sol-vin:
I just updated it to 4.0. :) New version is looking spiffy!


raysan5:
@sol-vin Nice! Thanks! :D

PR #2133, opened 2021-11-11
- Add DrawModelFromTransform() function

warzes:
I am writing my own 3d physics engine. And it needs a Transform matrix. Raylib calculates this matrix only in Draw - which is why I had to duplicate the code.

Of course the Model has a transform field. But in the current DrawModel * () it changes (for example, position), so it cannot be used.

I propose to function that takes this matrix from the user, and does not calculate it in draw.


raysan5:
@warzes I'm afraid I'm not merging this function. You can just use Model.transform to setup a transform matrix.

warzes:
> @warzes I'm afraid I'm not merging this function. You can just use Model.transform to setup a transform matrix.
This will not work as Draw re-sets the transformation matrix. 
Even if I adjust this matrix, it will be incorrect

PR #2136, opened 2021-11-13
Fix memory leak in parser

ronnieholm:
Found and verified fixed with Valgrind.


raysan5:
@ronnieholm Good catch! Thanks!

PR #2138, opened 2021-11-13
ExportMesh only exported 1/3 of the faces

nlapinski:
ExportMesh was only exporting partial face lists for unique faces, should iterate the vertex/3 not facelist/3

This might be the indented functionality for connected meshes? I don't think it would work in that case either though. I only tested with procedural meshes, that seemed to not  return connected faces. Which makes sense for opengl.
like
`GenMeshSphere(1.0, 12,12);`

I'll have to see if there is also a better way to collect a proper export textData size, for now calloc(datasize,4) worked ok, without this it was causing a crash (on windows)

A optimization step to fuse duplicate vertex data would be nice in the future.

Tested the changes on windows/mingw64 seemed to be fine.


raysan5:
@nlapinski Every triangle in a raylib mesh should actually be a face in the OBJ model exported. I'm not merging this code but I reviewed it in referenced commit.

PR #2141, opened 2021-11-14
Fix scissor mode for high DPI screens

lukekras:
I was running the [core scissor test](https://github.com/raysan5/raylib/blob/master/examples/core/core_scissor_test.c) example on a MacBook Pro (macOS Big Sur Intel) which is high DPI (2x). When running the test the scissor square was not lined up with the mouse square outline. I've raised issue (#2142)  and submitted a screenshot. I found that the [rlScissor](https://github.com/raysan5/raylib/blob/6342cf103a4d8b407e84a87fd711e10d8df940e2/src/rcore.c#L2191) call wasn't quite right. Also that line wasn't being called because the [CORE.Window.flags hadn't had the FLAG_WINDOW_HIGHDPI set](https://github.com/raysan5/raylib/blob/6342cf103a4d8b407e84a87fd711e10d8df940e2/src/rcore.c#L2187).

Note that I've added code to set the FLAG_WINDOW_HIGHDPI CORE.Window.flags, but only on desktop (via the PLATFORM_DESKTOP define). The reason for desktop restriction, was that I wasn't sure of the impact of that code for broader deployments other than desktop.

This was only tested on macOS Big Sur with HIGH DPI.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -601,6 +601,8 @@ static bool InitGraphicsDevice(int width, int height);  // Initialize graphics d
 static void SetupFramebuffer(int width, int height);    // Setup main framebuffer
 static void SetupViewport(int width, int height);       // Set viewport for a provided width and height
 
+static void SetHighDPIFromWindowScale(void);            // Helper function to set window high DPI flag from window scale
+
 #if defined(PLATFORM_DESKTOP) || defined(PLATFORM_WEB)
 static void ErrorCallback(int error, const char *description);                             // GLFW3 Error Callback, runs on GLFW3 error
 // Window callbacks events
@@ -805,6 +807,7 @@ void InitWindow(int width, int height, const char *title)
     SetShapesTexture(texture, (Rectangle){ 0.0f, 0.0f, 1.0f, 1.0f });
 #endif
 #if defined(PLATFORM_DESKTOP)
+    SetHighDPIFromWindowScale();
     if ((CORE.Window.flags & FLAG_WINDOW_HIGHDPI) > 0)
     {
         // Set default font texture filter for HighDPI (blurry)
@@ -1452,6 +1455,13 @@ void ClearWindowState(unsigned int flags)
         TRACELOG(LOG_WARNING, "RPI: Interlaced mode can only by configured before window initialization");
     }
 #endif
+}
+
+// Helper function to set window high DPI flag from window scale
+void SetHighDPIFromWindowScale()
+{
+    Vector2 scale = GetWindowScaleDPI();
+    if (scale.x > 1 || scale.y > 1) SetConfigFlags(FLAG_WINDOW_HIGHDPI);
 }
 
 // Set icon for window (only PLATFORM_DESKTOP)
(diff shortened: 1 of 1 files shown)


raysan5:
@lukekras This issue was reviewed recently on this PR: https://github.com/raysan5/raylib/pull/1987

It seems it's platform dependant and also macOS version dependant. Did you verify your solution works for all macOS versions and platforms?

lukekras:
Tested [core_scissor_test](https://github.com/raysan5/raylib/blob/master/examples/core/core_scissor_test.c) on Windows 10, Toshiba Satellite laptop.

Tested on branches master, and Fix_scissor_mode_high_DPI2

INFO: Initializing raylib 4.0
INFO:     > Display size: 1366 x 768
INFO: GL: Supported extensions count: 296
INFO:     > Vendor:   ATI Technologies Inc.
INFO:     > Renderer: AMD Radeon(TM) R2 Graphics
INFO:     > Version:  3.3.13559 Core Profile Context 26.20.12028.2
INFO:     > GLSL:     4.60
INFO: TEXTURE: [ID 1] Texture loaded successfully (1x1 | R8G8B8A8 | 1 mipmaps)
INFO: SHADER: [ID 1] Vertex shader compiled successfully
INFO: SHADER: [ID 2] Fragment shader compiled successfully
INFO: SHADER: [ID 3] Program shader loaded successfully
INFO: RLGL: Render batch vertex buffers loaded successfully in RAM (CPU)
INFO: RLGL: Render batch vertex buffers loaded successfully in VRAM (GPU)
INFO: RLGL: Default OpenGL state initialized successfully
INFO: TEXTURE: [ID 2] Texture loaded successfully (128x128 | GRAY_ALPHA | 1 mipmaps)
INFO: FONT: Default font loaded successfully (224 glyphs)

lukekras:
I tested on a MacBook Pro (Intel) macOS Big Sur 11.6.1. I might have a working old mac that I can test non high DPI.

I couldn't see anywhere in the code where CORE.Window.flags 
 sets the FLAG_WINDOW_HIGHDPI flag, so I set the flag inside InitWindow(). I think this would effect any platform that returns GetWindowScaleDPI() a  value of Vector2 that is not 1. There would be some Windows systems like this, but I don't have one.

raysan5:
@lukekras Note that `FLAG_WINDOW_HIGHDPI` should be set by users requiring calling `SetConfigFlags()` before `InitWindow()`, by default HighDPI support is not enabled.

lukekras:
@raysan5 Ok, so the setting of `FLAG_WINDOW_HIGHDPI` via `SetConfigFlags()` before `InitWindow()` is desired behaviour. Developers' would then need to do some logic similar to `SetHighDPIFromWindowScale()` if they want to handle `FLAG_WINDOW_HIGHDPI`. I'm ok with that. I still think there is an issue in `BeginScissorMode` with high DPI mode. You understandably want lots of testing before merging anything. Would like me to recut the PR with just the `BeginScissorMode` change?

lukekras:
@raysan5 I've created another PR #2160 which has the `BeginScissorMode` change, and puts the `SetConfigFlags(FLAG_WINDOW_HIGHDPI)` as a comment in [core scissor test](https://github.com/raysan5/raylib/blob/master/examples/core/core_scissor_test.c)
I've done some more testing and it gets a little tricky when you have a mixture of monitors attached.

lukekras:
Closing this PR, and raised #2160 with refactored code.

Issue #2145, opened 2021-11-17
[build] Errors building Android Arch x86_64


WilledgeR:
Hello,

Version: 4.0 master branch
Platform: Windows
Target: Android
ANDROID_ARCH           ?= x86_64

`ARM64` works well but:
Im making `x86_64` to test an android app on my emulator. 

`C:\w64devkit\bin\mingw32-make PLATFORM=PLATFORM_ANDROID`
using latest w64devkit
caused:

WilledgeR:
In file included from C:/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/include/direct.h:10,
                 from rcore.c:181:
C:/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/include/io.h:280:15: error: conflicting types for 'mkdir'
   int __cdecl mkdir (const char *) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
               ^~~~~
In file included from rcore.c:169:
C:\Users\Mahdi\AppData\Local\Android\Sdk\ndk\21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/sys/stat.h:156:5: note: previous declaration of 'mkdir' was here 
 int mkdir(const char* __path, mode_t __mode);
     ^~~~~
In file included from C:/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/include/direct.h:10,
                 from rcore.c:181:
C:/mingw-w64/x86_64-8.1.0-win32-seh-rt_v6-rev0/mingw64/x86_64-w64-mingw32/include/io.h:283:15: error: conflicting types for 'chmod'
   int __cdecl chmod (const char *, int) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
               ^~~~~
In file included from rcore.c:169:
C:\Users\Mahdi\AppData\Local\Android\Sdk\ndk\21.4.7075529/toolchains/llvm/prebuilt/windows-x86_64/sysroot/usr/include/sys/stat.h:154:5: note: previous declaration of 'chmod' was here 
 int chmod(const char* __path, mode_t __mode);
     ^~~~~
[...]

PR #2148, opened 2021-11-17
Fixes CC while compiling for ANDROID_ARCH x86

WilledgeR:
forgotten CC overriding when compiling for x86, x86_64

[relative issue](https://github.com/raysan5/raylib/issues/2145)


raysan5:
@WilledgeR Nice! Thanks for the improvement!

PR #2150, opened 2021-11-18
implement transformation and projection stacks

mgood7123:
this allows for separate transformation and projection stacks and addresses the core issue stated in https://github.com/raysan5/raylib/pull/2146

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -888,6 +888,58 @@ RLAPI void rlLoadDrawQuad(void);     // Load and draw a quad
 // Types and Structures Definition
 //----------------------------------------------------------------------------------
 #if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
+
+// A structure to represent a stack
+struct rl_MatrixStackNode {
+    Matrix data;
+    struct rl_MatrixStackNode* next;
+};
+
+typedef struct rl_MatrixStackNode MStack;
+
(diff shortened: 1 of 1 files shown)


raysan5:
This is a good improvement of current implementation but a Matrix stack does seem a feature more aligned with legacy OpenGL style API than new graphics APIs... for the moment, I'm not merging this improvement.

mgood7123:
> This is a good improvement of current implementation but a Matrix stack does seem a feature more aligned with legacy OpenGL style API than new graphics APIs... for the moment, I'm not merging this improvement.

true, tho this is mainly to remain compatible with a core legacy OpenGL feature that is present when using raylib with legacy OpenGL but absent when using raylib with modern OpenGL

simply so RayLib can remain compatible with both legacy OpenGL and modern OpenGL without rewriting code (most modern GL users implement their own matrix stacks anyway so this just makes it easier)

raysan5:
I know... but I don't know how long the OpenGL 1.1 backend will be kept as an option. Current matrix implementation is not perfect it's probably more than enough for users targeting OpenGL 1.1 in 2022.

mgood7123:
> I know... but I don't know how long the OpenGL 1.1 backend will be kept as an option. Current matrix implementation is not perfect but it's probably more than enough for users targeting OpenGL 1.1 in 2022.

true, gl 1.1 is quite old

mgood7123:
maybe the matrix stack can be optional?

PR #2156, opened 2021-11-18
Move raylib-v7 to list of unmaintained bindings

ytrms:
Sadly, [raylib-v7](https://github.com/Rabios/raylib-v7) is no longer being maintained as per their [README](https://github.com/Rabios/raylib-v7/blob/master/README.md).


raysan5:
@ytrms oooh... that's sad... anyway, thanks for the PR!

PR #2157, opened 2021-11-18
improved blending support

tbfts:
Adds more blending options. 
- Doesn't change anything already present, and works exactly the same way. 
- Added matching GL defines for custom modes. (unsure of the names)
- Added support to set factors for colors and alpha separately.

I didn't notice the core had a few blend functions until I was pretty much finished. I tried to think of a way to improve those, but unless we can add the blend defines over there I can't think of any way to make that work. Unless you want to use one of the predefined modes, they're kinda useless atm.

Seems to be working fine for me on vs2020, can't test on anything else.
Also, this is my first contribution ever, so let me know if I'm doing something wrong...

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -839,8 +839,7 @@ typedef enum {
     BLEND_MULTIPLIED,               // Blend textures multiplying colors
     BLEND_ADD_COLORS,               // Blend textures adding colors (alternative)
     BLEND_SUBTRACT_COLORS,          // Blend textures subtracting colors (alternative)
-    BLEND_CUSTOM,                   // Blend textures using custom src/dst factors (use rlSetBlendMode())
-    BLEND_CUSTOM_SEPARATED          // Blend textures using separate custom src/dst factors
+    BLEND_CUSTOM                    // Belnd textures using custom src/dst factors (use rlSetBlendMode())
 } BlendMode;
 
 // Gesture
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -839,7 +839,8 @@ typedef enum {
     BLEND_MULTIPLIED,               // Blend textures multiplying colors
     BLEND_ADD_COLORS,               // Blend textures adding colors (alternative)
     BLEND_SUBTRACT_COLORS,          // Blend textures subtracting colors (alternative)
-    BLEND_CUSTOM                    // Belnd textures using custom src/dst factors (use rlSetBlendMode())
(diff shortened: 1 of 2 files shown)


raysan5:
@tbfts I'm afraid I'm not adding this improvement. It seems intended for a specific use case.

tbfts:
I feel like it could be used for any combination of blending. 
Is it too specific in that way ? 

(not arguing against the decision, just want to understand it more)

But some background; the main reason for this was that I couldn't get glad to work with raylib, in the end it was just easier to implement it directly into raylib. I really needed to separate the color and alpha blend equations for what I was trying to do, and having the defines there are just qol.

raysan5:
It adds 6 new variables to the global data storage that I think most users won't ever need. It also adds a lot of defines that I think are not really needed. Maybe I can consider adding it on a future but at this moment I'm trying to focus efforts for raylib and its libraries in solving issues instead of adding more features.

tbfts:
Ok, I understand that.

What if there were no variables, and just a few functions that directly call the necessary gl blend commands ? 
Say;
`rlSetBlendFactorsSeparated(srcColor, dstColor, eqColor, srcAlpha, dstAlpha, eqAlpha)`
`rlSetBlendConstant(r,g,b,a)`

The only difference is that using `rlSetBlendMode` would only revert the separated command, instead of activate it. But, I feel that difference may go against the design already in place, and might be more of a band aide for the lack of separated blend commands- which I think more of as an issue rather than of adding a new feature, or rather blending is an incomplete feature without it.

One more idea would be to just remove custom modes altogether, and have;
`rlSetBlendModeCustom(srcFactor, dstFactor, equation)` _-removing rlSetBlendFactors_
`rlSetBlendModeCustomSeparated(srcColor, dstColor, eqColor, srcAlpha, dstAlpha, eqAlpha)`
`rlSetBlendConstant(r,g,b,a)`
Again these would directly call the gl commands, and setting a mode would only revert it instead of activate it. This may also have the benefit of allowing us to remove all the blending variables altogether, only needing the `currentBlendMode`. 

I have more ideas for it, but this post is might be getting long.. let me know if I've persuaded you, even if just a little :)
For the future, I will try to keep in mind the overall userbase, instead of the few who might use a feature. 
~thanks for the consideration

raysan5:
@tbfts I'm closing this PR. The improvement in blending is clear but it's out of scope for raylib at this moment. I also think many users today could implement all those specific types of blending using a shader.

Issue #2142, opened 2021-11-14
[core] core_scissor_test not working properly on high DPI screen


lukekras:
- [ x] I tested it on latest raylib version from master branch
 - [ x] I checked there is no similar issue already reported
 - [ x] My code has no errors or misuse of raylib

### Issue description

I was running the [core_scissor_test](https://github.com/raysan5/raylib/blob/master/examples/core/core_scissor_test.c) example program and it wasn't working on a high DPI MacBook Pro. The scissor red box was out of alignment to the mouse outline box.

### Environment

macOS Big Sur Intel MacBook Pro

INFO: GL: OpenGL device information:
INFO:     > Vendor:   Intel Inc.
INFO:     > Renderer: Intel(R) Iris(TM) Plus Graphics 650
INFO:     > Version:  4.1 INTEL-16.5.2
INFO:     > GLSL:     4.10

### Issue Screenshot
<img width="912" alt="Screen Shot 2021-11-14 at 10 22 19 pm" src="https://user-images.githubusercontent.com/1853946/141682682-3abdf343-8ad5-4122-bb86-b588289a8e8a.png">

### Code Example

I've submitted a PR #2141

raysan5:
Related: https://github.com/raysan5/raylib/issues/1982 and https://github.com/raysan5/raylib/pull/1987

lukekras:
Closed PR #2141 and created PR #2160 with refactored code.

raysan5:
@ArnaudValensi pushed a big commit related, maybe he can look into this issue?

lukekras:
@raysan5 @ArnaudValensi 
I've checked out `Fix scissor on macos` #2170 and confirm that it fixes this problem. From my point of view this issue can be closed.

PR #2170 makes PR `Fix scissor mode for high DPI screens (refactored)` #2160 redundant.

raysan5:
@ArnaudValensi @lukekras thanks to you both for working on this issue and the reviews!

lukekras:
No problem @raysan5, happy to help where I can. Cheers.

PR #2160, opened 2021-11-20
Fix scissor mode for high DPI screens (refactored)

lukekras:
This is a refactored version of PR `Fix scissor mode for high DPI screens` #2141

From #2141 :
I was running the core scissor test example on a MacBook Pro (macOS Big Sur Intel) which is high DPI (2x). When running the test the scissor square was not lined up with the mouse square outline. I've raised issue (#2142) and submitted a screenshot. I found that the rlScissor call wasn't quite right. Also that line wasn't being called because the CORE.Window.flags hadn't had the FLAG_WINDOW_HIGHDPI set.

### DIFF
--- a/examples/core/core_scissor_test.c
+++ b/examples/core/core_scissor_test.c
@@ -20,6 +20,9 @@ int main(void)
     const int screenWidth = 800;
     const int screenHeight = 450;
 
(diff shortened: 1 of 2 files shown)


lukekras:
This PR has been made redundant by `Fix scissor on macos` #2170 In view os this I'll close this PR.

PR #2163, opened 2021-11-21
fixed typo: aproximation to approximation

pancakevirus:


raysan5:
@pancakevirus thanks for the fix!

PR #2167, opened 2021-11-23
Android touch fix: check for POINTER_UP and UP events

di-ant:
Implemented checks for `AMOTION_EVENT_ACTION_POINTER_UP` and `AMOTION_EVENT_ACTION_UP` flags to not register  pointers lifted during current event, otherwise they get stuck until next pointer event.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5439,6 +5439,14 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
         return 0;
     }
 
+    int32_t action = AMotionEvent_getAction(event);
+    unsigned int flags = action & AMOTION_EVENT_ACTION_MASK;
+
+    // Don't register pointers lifted during this action,
+    // otherwise they get stuck until next pointer event
+    if (flags == AMOTION_EVENT_ACTION_POINTER_UP || flags == AMOTION_EVENT_ACTION_UP)
+        CORE.Input.Touch.pointCount -= 1;
+
     // Register touch points count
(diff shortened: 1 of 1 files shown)


raysan5:
@di-ant It seems this code resets all touches when one finger is up, is this the expected behaviour?

di-ant:
@raysan5 Is it? It doesn't on my phone, at least.

Right now, on `master`,  as "up" events are coming in, they are counted as touch too, so `int GetTouchPointCount(void);` always returns correct amount of touch points only after following event. Even worse, after a first touch event, it cannot be less than one, because event when last finger left the screen still counts as a touch event, and cannot be changed until next touch event.
Idea behind this fix, is when "up" event is coming in, we just don't count it towards touch points, by subtracting one (only when event is "up" event). (Its `touchPoints -= 1`, not `touchPoints = 1`)
Order of touch points doesn't seem to be affected by this nanopatch, at least with my configuration, `int GetTouchPointId(int index);` works as expected (no misalignment between `index` and `id`), and `int GetTouchPointCount(void);` accurately reflects amount of fingers on my touchscreen.

Maybe it's just a problem with my environment, but i don't have anything else to test with.

My config is:
Build environment: linux, kernel 5.15.3
Android API level: 29
NDK: r23b
jdk: 13.0.2
smartphone: Redmi Note 10 Pro (Android 12)

di-ant:
I am really sorry, somehow i messed up the order. Should be good now.

di-ant:
actually, in specific cases it's not even guaranteed that last event's index is highest, so this is not totally correct, will be back with a real fix.

PR #2168, opened 2021-11-23
allow physac to be built as a shared library

jasonswearingen:
if `BUILD_LIBTYPE_SHARED` is defined allow build as a shared dll

This is as per a discussion on discord yesterday, on how to get physac working with C# bindings.


raysan5:
@jasonswearingen thanks for the improvement!

PR #2181, opened 2021-12-01
ADDED: GetKeyAsText

catmanl:
This can be useful when asking for keyboard input, the user knows what key it's supposed to be pressing or what key has been pressed.


catmanl:
I had also implemented the same functions with their respective type of input such as mouse, gamepad buttons, axis and gestures. I have decided to wait to see whether this would be merged or not and so, add them.

raysan5:
@catmanl thanks but no plans to add that function, I think most users don't need it. It also adds many const string data to the library.

PR #2187, opened 2021-12-05
raylib-cslo switches to MPL-2.0

jasonswearingen:
from LGPL 2.1


raysan5:
@jasonswearingen thanks for the update.

PR #2189, opened 2021-12-06
Fix inverse length in Vector2Normalize

HarriP:
Following the pattern in Vector3Normalize.


raysan5:
@HarriP thanks for the improvement.

PR #2191, opened 2021-12-06
Letterboxing window flag

vb-mich:
Take this with a pinch of salt.

This is a simple change to allow the game to resize with the window without manually having to implement letterboxing, for my applications and I believe many others it's sometimes convenient to let the content automatically resize to fit the window.
This is done simply by NOT changing CORE.Window.render size in SetupViewport if the flag FLAG_FIXED_RENDER_SIZE has been set.

BeginScissorMode has been changed accordingly as well as GetMousePosition.

This is a change that I keep stashing at every update, hope it can get integrated.
Mind that it has been tested only for PLATFORM_WEB and WINDOWS DESKTOP.

Cheers,
Mich


raysan5:
@vb-mich I tested this PR with some examples.
 - It works to scale the 3d context to the window size but it breaks for 2d content (shapes, text...) that gets deformed with scaling.
 - Mouse seems to work ok for 2d, position is correct but not for 3d (tested on `core_3d_picking` example).

I agree that this flag could be very useful but I'm afraid it requires more work to be implemented properly.

PR #2192, opened 2021-12-06
Fix #2190 - Add GetRaylibVersion function

### LINKED ISSUE #2190: Add `GetRaylibVersion` function for bindings
### Issue description

For the benefit of bindings, because raylib does change layouts and semantic meaning of parameters sometimes, it would be nice to have a binary interface to query the current raylib version. I know there is a `#define`, but that doesn't help when you are only using the library.

This would help detect at runtime that you have linked the wrong version, or your dll has changed to an incorrect version.

### Environment

Using raylib from an alternative programming language such as D

### Code Example

proposed implementation:

```C
const char * GetRaylibVersion()
{
   return RAYLIB_VERSION;
}
```

schveiguy:
For bindings other than C which do not read the header file, but may use the compiled binary library.

E.g. in raylib's D bindings, we can link directly against the C library, but because C has no name mangling, changes to function parameters can allow a mismatched library to link, but have disastrous effects (such as memory corruption). Same thing with struct layout changes.

This at least provides a way to verify the binary raylib version matches what the binding expects.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1015,6 +1015,7 @@ RLAPI int GetRandomValue(int min, int max);                       // Get a rando
 RLAPI void SetRandomSeed(unsigned int seed);                      // Set the seed for the random number generator
 RLAPI void TakeScreenshot(const char *fileName);                  // Takes a screenshot of current screen (filename extension defines format)
 RLAPI void SetConfigFlags(unsigned int flags);                    // Setup init configuration flags (view FLAGS)
+RLAPI const char *GetRaylibVersion();                             // Get Raylib version (for non-C bindings)
 
 RLAPI void TraceLog(int logLevel, const char *text, ...);         // Show trace log messages (LOG_DEBUG, LOG_INFO, LOG_WARNING, LOG_ERROR...)
 RLAPI void SetTraceLogLevel(int logLevel);                        // Set the current threshold (minimum) log level
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -6855,3 +6855,10 @@ const char *TextFormat(const char *text, ...)
     return currentBuffer;
 }
 #endif // !SUPPORT_MODULE_RTEXT
+
+// Get the release of raylib that this library is built with, to assist in non-C bindings determining
+// a mismatch of structures and/or parameters.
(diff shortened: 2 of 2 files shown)


raysan5:
@schveiguy I'm afraid I'm not adding this function, the version is already available at the header or at raylib initialization. It seems it's just intended for a very specific use case in a binding.

schveiguy:
It's actually not available to bindings any other way. Let me explain:

The header file can be translated into a binding. However, there is no actual symbol in the library that matches what the `#define` uses. The results of binding against the wrong binary library are disastrous. For example, changes in struct layout will not be detected at link time, and using the functions can result in memory corruption + segfaults.

Even if you translate the header into a value that is readable by the other language, you get the *translated* header, not the original header.

Perhaps this situation is unique to D? I don't see another way around it. For sure, you are storing the string "4.0" in the library, as it is being printed. Perhaps making it a symbol I can use instead of a public function is reasonable?

e.g. inside rcore.c:

```c
const char * _raylibVersion = RAYLIB_VERSION;
...
// later on in InitWindow:
TRACELOG(LOG_INFO, "Initializing raylib %s", _raylibVersion);
```

Would this be allowed? It just adds a symbol to the library that then gives me access to the string you are already storing.

raysan5:
@schveiguy It's not a solution I specially like but I prefer this solution than a new function in the API. I added it in commit https://github.com/raysan5/raylib/commit/ef6959ed540cd650e13611c35ea543922ab83a98.

schveiguy:
Thank you! Just so you can see how I plan to use this:

```d
extern(C) extern const char *raylibVersion; // defined in raylib
/// call this on program start to ensure the raylib binding is correct
void validateRaylib() {
    string rlVersion = raylibVersion[0 .. strlen(raylibVersion)];
    enforce(rlVersion == RAYLIB_VERSION,
        "Error: Raylib binding generated from " ~ RAYLIB_VERSION ~
            ", but binary version is " ~ rlVersion);
}
```

This way, you can at least ensure you aren't running with an incorrect binding. I have had several people run into this when using raylib-d built for 3.7.0 but linking against 4.0.0. This will help starting next release.

Issue #2172, opened 2021-11-25
[raymath] Angle returned by `Vector2Angle()` isn't correct


Knose1:
Like [this](https://githubmemory.com/repo/gen2brain/raylib-go/issues/42) post say, the angle isn't correct.  
I used to multiply one of the vectors by 1000 when calling the function to avoid that problem.   
The problem came back with `Vector2Angle({0,1}, {0,1000})` the result shall be 0 but the returned value is 90
  
I fixed it by replacing this line  
https://github.com/raysan5/raylib/blob/49a5547386a354fc1ca4c6fb6a4c850615b25038/src/raymath.h#L284  
  
By this one in the raymath.h of my project  
```cpp
float result = atan2(v2.y, v2.x) - atan2(v1.y, v1.x) * (180.0f / PI);
```

schveiguy:
This should be closed now.

leowribeiro:
I discovered that if you want to calculate the angle between vectors A and B you can use
Vector2Angle(A, A-B) - 180

raysan5:
@leowribeiro Note that `Vector2Angle()` returns angle in radians so you can not operate with `-180` directly.

leowribeiro:
Ok, I'm confused. I'm using raylib-cpp, in my code I put this code
targetTh = j.Angle(j-v) + 270;
And is working... when I checked the Angle method from Vector2.hpp, it is just a call to Vector2Angle(), look

float Angle(const ::Vector2& vector2) const {
       return Vector2Angle(*this, vector2);
}

My working perfectly, I'm sure.

raysan5:
@leowribeiro that's because you are using an old version of raylib. That was an issue that was fixed some time ago. Now `Vector2Angle()` return radians, not degrees.

leowribeiro:
Understood, thank you.

PR #2193, opened 2021-12-06
Vector2Angle returns degrees instead of radians, but all other raymath

schveiguy:
functions use radians, making this awkward to use.

Per discussion on discord, raymath should always use radians. I found this one was left out by searching for `180` and `360`.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -281,9 +281,9 @@ RMAPI float Vector2Distance(Vector2 v1, Vector2 v2)
 // Calculate angle from two vectors in X-axis
 RMAPI float Vector2Angle(Vector2 v1, Vector2 v2)
 {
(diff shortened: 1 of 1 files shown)


raysan5:
@schveiguy Good catch! I missed that one. Actually, there is an open issue that could be related: https://github.com/raysan5/raylib/issues/2172

jdeokkim:
graph

@raysan5 I believe `atan2f(v2.y, v2.x) - atan2f(v1.y, v1.x)` is the correct way to retrieve an angle between two vectors, based on the defintion of the arc tangent. Here's an article about it from Stack Overflow: https://stackoverflow.com/questions/21483999/using-atan2-to-find-angle-between-two-vectors/21486462

PR #2198, opened 2021-12-07
raylib.h: Add relevant defines to types

greenfork:
Allow order-independent includes so that other library files can
be included before raylib.h.

This fixes https://github.com/greenfork/nimraylib_now/issues/12 and https://github.com/greenfork/nimraylib_now/issues/49 of my Nim bindings. The problem is that Nim doesn't know how to include C files in order. And this seems as a generally good approach to use.

### DIFF
--- a/src/extras/physac.h
+++ b/src/extras/physac.h
@@ -130,6 +130,7 @@ typedef struct Vector2 {
     float x;
     float y;
 } Vector2;
+#define RL_VECTOR2_TYPE
 #endif
 
 // Matrix2x2 type (used for polygon shape rotation matrix)
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -131,18 +131,6 @@
     #define CLITERAL(type)      (type)
 #endif
 
-// NOTE: We set some defines with some data types declared by raylib
-// Other modules (raymath, rlgl) also require some of those types, so,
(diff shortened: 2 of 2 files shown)


raysan5:
Actually those defines were created to allow other files to check for them while keeping the raylib-required types. The problem with this change is that if other files define custom types before raylib, then many raylib internals could broke due to not getting the expected type.

greenfork:
> The problem with this change is that if other files define custom types before raylib, then many raylib internals could broke due to not getting the expected type.

Could you please clarify this part? In order for other files to break raylib code, they need to define `RL_VECTOR2_TYPE` and similar defines. If they do so, they probably know what they do. Current change will only affect code such as
```c
#include "physac.h"
#include "raylib.h"
```
where includes don't have to be in the specific order. In other cases we assume that no other file is going to define `RL_VECTOR2_TYPE` which is a good enough assumption since this define is quite unique.

raysan5:
@greenfork I prefer to force the use of raylib defined types and do not allow other previously included files to redefine them (even they know what are they doing), it could suppose a security concern if that is allowed.

Other files using same types could be included after that and use the defines to check for them if required.

greenfork:
@raysan5 alright, thanks for explaining.

PR #2202, opened 2021-12-09
Add raylib-guile to BINDINGS.md

petelliott:
I wrote a little auto-generated raylib bindings package for [GNU Guile](https://www.gnu.org/software/guile/)!

https://github.com/petelliott/raylib-guile


raysan5:
@petelliott Nice! Thanks for the addition!

PR #2203, opened 2021-12-09
Fixed functions Vector2Angle and Vector3Angle

trikko:


trikko:
Please notice that previously ```Vector2Angle``` tried to make angle positive, but it should return a negative angle if vector are passed in reverse order.

orcmid:
Does this qualify as a breaking change?

orcmid:
In the Vector3Angle replacement, are there any numerical quality differences between the original method and the replacing single expressions?

trikko:
> In the Vector3Angle replacement, are there any numerical quality differences between the original method and the replacing single expression?

The original Vector3Angle returned a Vector2 with two angles on xy e xz. Probably not the expected result by a raylib user.
This implementation instead return a float representing the angle between the two vectors.

orcmid:
> The original Vector3Angle returned a Vector2 with two angles on xy e xz. Probably not the expected result by a raylib user. This implementation instead return a float representing the angle between the two vectors.

Good point.  So it is a breaking change, a different function with the same name, different signature.  I gather these are new functions or at least not on the current cheetsheet.

I see that whatever the advisability of the `Vector3Angle` change, the CONVENTIONS at the head of `raymath.h` are contradicted.  

Do we know where any of these are used in other raylib modules? It does appear that the difference would be detected at compile time and also in intellisense for most usages of `raymath.h`.

trikko:
I wasn't aware of these CONVENTIONS. They sound strange to me. 
Why should we unroll the code if compiler is able to do by itself easily?

https://godbolt.org/z/6n6E6Y8E3

trikko:
> PS: In what revision of the C standard was (passing?) and return of structs by value introduced? I had completely overlooked that. I presume that std:c11 is enough, but maybe std:c17 needed?

It compiles even with ```-std=iso9899:1990``` where single-line comments are not allowed :)

orcmid:
> > PS: In what revision of the C standard was (passing?) and return of structs by value introduced? I had completely overlooked that. I presume that std:c11 is enough, but maybe std:c17 needed?
> 
> It compiles even with `-std=iso9899:1990` where single-line comments are not allowed :)

Wow.  Yes, it is defined for c99.

raysan5:
@trikko Thanks for the review! It seems some example could be affected by this change.

PR #2205, opened 2021-12-12
Add panning to raudio and update audio_raw_stream example.

ptarabbia:


src/raudio.c
```diff
@@ -332,6 +333,8 @@ struct rAudioBuffer {
 
     rAudioBuffer *next;             // Next audio buffer on the list
     rAudioBuffer *prev;             // Previous audio buffer on the list
+
+    float panVolumeSmoothing[2];
```
raysan5, line 337: Could this variable be omitted in some way? If I understand it correctly it's used to make the pan transition smoother, couldn't that be controlled by users? similar to pitch change...

src/raudio.c
```diff
@@ -397,10 +400,18 @@ void StopAudioBuffer(AudioBuffer *buffer);
...
 void ResumeAudioBuffer(AudioBuffer *buffer);
 void SetAudioBufferVolume(AudioBuffer *buffer, float volume);
+void SetAudioBufferPan(AudioBuffer* buffer, float pan);
 void SetAudioBufferPitch(AudioBuffer *buffer, float pitch);
 void TrackAudioBuffer(AudioBuffer *buffer);
 void UntrackAudioBuffer(AudioBuffer *buffer);
 
+
+// pan law
+static float FastSineApproximation(float x)
```
raysan5, line 410: I prefer to avoid adding a new function, code can be implemented directly where required.

raysan5:
@ptarabbia This is a really nice improvement! Thanks!

src/raudio.c
ptarabbia: I have removed smoothing entirely. This feature is now a simple API for statically positioning a sound in the stereo field.

For dynamic pan changes, you need per-sample smoothing, which at the moment, is not feasible by users, as they can only modify things for a whole audio block. However, I'm currently adding a new feature, that will allow users to add audio processing effects - and this will give them full per-sample control if they need it.

raysan5:
@ptarabbia thanks for the improvement!

PR #2208, opened 2021-12-14
Parse enum descriptions and value descriptions

eutro:
This PR updates the API parser to parse the descriptions of enums and enum values. These were marked with TODOs before. Enum value descriptions especially have quite a lot of content that was missing from the parses.

### DIFF
--- a/parser/raylib_parser.c
+++ b/parser/raylib_parser.c
@@ -140,7 +140,7 @@ static void GetDataTypeAndName(const char *typeName, int typeNameLen, char *type
 static unsigned int TextLength(const char *text);           // Get text length in bytes, check for \0 character
 static bool IsTextEqual(const char *text1, const char *text2, unsigned int count);
 static void MemoryCopy(void *dest, const void *src, unsigned int count);
-static char* CharReplace(char* text, char search, char replace);
+static char* EscapeBackslashes(char *text);
 
 static void ExportParsedData(const char *fileName, int format); // Export parsed data in desired format
 
@@ -765,13 +765,19 @@ static bool IsTextEqual(const char *text1, const char *text2, unsigned int count
     return result;
 }
 
-// Search and replace a character within a string.
-static char* CharReplace(char* text, char search, char replace)
+// Escape backslashes in a string into a static buffer
+static char* EscapeBackslashes(char *text)
 {
-    for (int i = 0; text[i] != '\0'; i++)
-        if (text[i] == search)
-            text[i] = replace;
-    return text;
+    static char buf[256];
+    char *a = text;
+    char *b = buf;
+    do
+    {
+        if (*a == '\\') *b++ = '\\';
+        *b++ = *a;
+    }
+    while (*a++);
+    return buf;
 }
 
 /*
@@ -826,44 +832,6 @@ static char *TextReplace(char *text, const char *replace, const char *by)
 }
 */
 
-// Like fprintf, but supports only %i and %S. %S prints an escaped JSON-like string.
-static void FPrintfEscapes(FILE *file, const char *format, ...) {
-    va_list args;
-    va_start(args, format);
-    for (const char *c = format; *c; c++)
-    {
-        if (*c == '%')
-        {
-            c++;
-            switch (*c)
-            {
-                case 'S': {
-                    const char *str = va_arg(args, char *);
-                    putc('"', file);
-                    for (; *str; str++)
-                    {
-                        if ((*str == '\\') || (*str == '"'))
-                        {
-                            putc('\\', file);
-                        }
-                        putc(*str, file);
-                    }
-                    putc('"', file);
-                    break;
-                }
-                case 'i': {
-                    fprintf(file, "%i", va_arg(args, int));
-                    break;
-                }
-                default: {
-                    fprintf(stderr, "WARNING: Unsupported format specifier used: %c\n", *c);
-                }
-            }
-        }
-        else putc(*c, file);
-    }
-}
-
 // Export parsed data in desired format
 static void ExportParsedData(const char *fileName, int format)
 {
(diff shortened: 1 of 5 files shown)


eutro:
Currently introduces a syntax error in outputs with an unescaped backslash.

eutro:
Now also properly escapes _all_ strings. I hope you don't mind the `FPrintfEscapes` function introduced to do this.

raysan5:
@eutro Hey! That's a great improvement! It was on my TODO list!

parser/raylib_parser.c
```diff
@@ -804,6 +826,44 @@ static char *TextReplace(char *text, const char *replace, const char *by)
 }
 */
 
+// Like fprintf, but supports only %i and %S. %S prints an escaped JSON-like string.
+static void FPrintfEscapes(FILE *file, const char *format, ...) {
```
raysan5, line 830: I prefer to avoid this function if possible and just escape them directly on fprintf(). What are the benefits of creating an additional function to do that? It also add an additional library dependency that I'd prefer to avoid.

eutro: Escaping the string requires the insertion of the extra backslashes (or possibly other characters for XML escapes), so it can't be done in-place without awkward shifting of bytes that may overflow buffers. Since they're being output directly, that's unnecessary and we can just insert the extra characters as we output.

raysan5: Sorry, I don't understand the problem, using backslashes is the expected way to print those characters. About the buffers size, it's a specific parser for a specific file (`raylib.h`), buffers could be sized for that with a secure margin.

eutro: To write the raw string of value:
`foo \ bar \ baz`
 in JSON,
`"foo \\ bar \\ baz"`
must be written.

These characters must be inserted into the string at specific indeces, which are the extra backslashes.

Naively, this means shifting the rest of the buffer to the right by a byte every time. Alternatively, counting backslashes ahead of time means shifting each byte at most once. Both of these are slightly more complicated than just outputting the extra backslashes with this function, but not the end of the world.

Additionally, inserting the extra backslashes into the string is destructive, meaning the original strings would be modified. This would be undesirable if we want to output the unescaped strings again. Say, if the program is modified to be able to output JSON _and_ XML without a restart, outputting the JSON would modify the strings before the XML gets output. Again, not the end of the world, but it may be an unintuitive bug.

An alternative function would be one that just outputs an escaped string, then we can split the `printf` calls on those.

Which would be preferrable?
- Using `FPrintfEscapes` (current)
- A `PrintEscaped` function that prints a single escaped string
- An `EscapeString` function that escapes a string in-place and destructively
- Something else

raysan5: Sorry again but I can't see the issue. As per my understanding `foo`, `bar` and `baz` are independently provided so they can just be properly concatenated with required backslashes when writing the required output. Also XML, JSON and others are generated independently from input data. 

Please, could you provide a specific examples from `raylib.h` parsing where that problem happens? Sorry again for not seeing the issue, english is not my first language and I may be missing the point.

eutro: If I want to output the description from this comment here
https://github.com/raysan5/raylib/blob/6342cf103a4d8b407e84a87fd711e10d8df940e2/src/raylib.h#L567
then the string has the raw value of `Key: '\'`. In JSON, I should output this as `"Key: '\\'"`, though, which is a character longer and requires the shifting of the `\'` to the right by a character.

This is how this would be done:

Unescaped string from description:
```
K e y :   ' \ ' 
```

If we want to modify the string in-place to have the escapes we must shift everything after the backslash to the right:
```
K e y :   ' _ \ '
```

and then we can insert the extra backslash:
```
K e y :   ' \ \ '
```
That's all, that's the issue I'm talking about.

---

What is currently being done (in this PR) is that we iterate over the string once, but output an extra backslash before anything that needs to be escaped:
```
K e y :   ' \ '
^ output K
K e y :   ' \ '
  ^ output e
...
K e y :   ' \ '
          ^ output '
K e y :   ' \ '
            ^ output two \s
K e y :   ' \ '
              ^ output '
```

---

So, which _should_ the approach be? Should we escape the string in-place (shifting the string and inserting backslashes, as in the first description) or should we just output an extra backslash as we output the string manually (like the second description)?

raysan5: Ok, now I understand the issue. Checking `raylib.h` there are only 4 cases where that's a problem (always descriptions):
```
Line 567:     KEY_BACKSLASH       = 92,       // Key: '\'
Line 1037: RLAPI char *LoadFileText(const char *fileName);                   // Load text data from file (read), returns a '\0' terminated string
Line 1039: RLAPI bool SaveFileText(const char *fileName, char *text);        // Save text data to file (write), string must be '\0' terminated, returns true on success
Line 1355: RLAPI unsigned int TextLength(const char *text);                  // Get text length, checks for '\0' ending
```
Current solution just replaces it by an empty space:
```c
CharReplace(funcs[i].desc, '\\', ' ')
```
That function could be improved to replace the char by another string, just using an internal static buffer.

eutro: I've replaced it with an `EscapeBackslashes` function now.

parser/raylib_parser.c
```diff
@@ -743,13 +764,19 @@ static bool IsTextEqual(const char *text1, const char *text2, unsigned int count
...
+static char* EscapeBackslashes(char *text)
 {
-    for (int i = 0; text[i] != '\0'; i++)
-        if (text[i] == search)
-            text[i] = replace;
-    return text;
+    static char buf[256];
+    char *a = text;
+    char *b = buf;
+    do
```
raysan5, line 773: I usually try to avoid the `do-while` formats. I can review that function later on.

parser/raylib_parser.c
raysan5: Ok, it looks good to me.

raysan5:
@eutro Thanks for the improvement! I'm merging and reviewing this PR.

PR #2212, opened 2021-12-15
Feature/audio effects

ptarabbia:
This PR adds audio processing to the audiobuffers, making it possible to implement filtering, reverb, ... on a given stream.
It's also includes the changes from my audioCallbacks branch (that simplifies the whole processing system by letting the audio threads request data when needed, instead of polling from the UI thread).

I've modified the audio_music_stream demo to show usage - applying filtering to a music stream, to simulate music played behind a closed door, for example.

### DIFF
--- a/examples/audio/audio_music_stream.c
+++ b/examples/audio/audio_music_stream.c
@@ -14,6 +14,23 @@
 
 #define AUDIO_THREAD_MUSIC_UPDATE true
 
+// a simple lowpass filter applied to the music stream
+static void audioEffectDemo(float* buffer, unsigned int nframes)
+{
+    static float low[2] = { 0.0f, 0.0f };
+    static const float cutoff = 70.0f / 44100.0f; // 70 Hz lowpass filter
+    const float k = cutoff / (cutoff + 0.1591549431f); // RC filter formula
+
+    for (unsigned int i = 0; i < nframes*2; i+=2)
+    {
+        float l = buffer[i], r = buffer[i + 1];
+        low[0] += k * (l - low[0]);
+        low[1] += k * (r - low[1]);
+        buffer[i] = low[0];
+        buffer[i + 1] = low[1];
+    }
+}
+
 int main(void)
 {
     // Initialization
@@ -32,6 +49,7 @@ int main(void)
 
     float timePlayed = 0.0f;
     bool pause = false;
+    bool hasfx = false;
 
     SetTargetFPS(60);               // Set our game to run at 60 frames-per-second
     //--------------------------------------------------------------------------------------
@@ -62,6 +80,20 @@ int main(void)
             else ResumeMusicStream(music);
         }
 
+        // Add/Remove effect
+        if (IsKeyPressed(KEY_F))
+        {
+            hasfx = !hasfx;
+            if (hasfx)
+            {
+                AddAudioStreamProcessor(music.stream, &audioEffectDemo);
+            }
+            else
+            {
+                RemoveAudioStreamProcessor(music.stream, &audioEffectDemo);
+            }
+        }
+
         // Get timePlayed scaled to bar dimensions (400 pixels)
         timePlayed = GetMusicTimePlayed(music)/GetMusicTimeLength(music)*400;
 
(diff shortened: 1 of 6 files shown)


raysan5:
This change is quite big an increases the level of complexity of the library for novice users. I'd prefer to keep the original examples simple and show the advance functionality in separate examples. I also need some time to review it more carefully.

ptarabbia:
> This change is quite big an increases the level of complexity of the library for novice users. I'd prefer to keep the original examples simple and show the advance functionality in separate examples. I also need some time to review it more carefully.

Understood. It's very little code added to the example though.

Also if you prefer me to split this into 2 separate pull requests, one for the threading changes, and another one for audio effects, I can do that too.

src/raylib.h
```diff
@@ -1468,7 +1476,7 @@ RLAPI RayCollision GetRayCollisionQuad(Ray ray, Vector3 p1, Vector3 p2, Vector3
 //------------------------------------------------------------------------------------
 
 // Audio device management functions
-RLAPI void InitAudioDevice(void);                                     // Initialize audio device and context
+RLAPI AudioDeviceInfo InitAudioDevice(void);                          // Initialize audio device and context
```
raysan5, line 1479: All Init*() functions return void, I prefer to keep it that way.

src/raylib.h
```diff
@@ -447,11 +448,18 @@ typedef struct Music {
     AudioStream stream;         // Audio stream
     unsigned int frameCount;    // Total number of frames (considering channels)
     bool looping;               // Music looping enable
-
+    bool background;            // Update in audio thread ?
     int ctxType;                // Type of music context (audio filetype)
     void *ctxData;              // Audio context data, depends on type
 } Music;
 
+typedef struct AudioDeviceInfo
```
raysan5, line 456: It can probably be avoided, no need of a new structure to query that info if required, probably most users won't care about it.

src/raylib.h
```diff
@@ -447,11 +448,18 @@ typedef struct Music {
     AudioStream stream;         // Audio stream
     unsigned int frameCount;    // Total number of frames (considering channels)
     bool looping;               // Music looping enable
-
+    bool background;            // Update in audio thread ?
```
raysan5, line 451: this mechanism looks weird to me, there is probably a better option for that...

examples/audio/audio_music_stream.c
```diff
@@ -23,6 +26,7 @@ int main(void)
     InitAudioDevice();              // Initialize audio device
 
     Music music = LoadMusicStream("resources/country.mp3");
+    music.background = AUDIO_THREAD_MUSIC_UPDATE;
```
raysan5, line 29: This mechanism looks very weird from an user perspective, I'd prefer to avoid it and look for an alternative.

examples/audio/audio_music_stream.c
```diff
@@ -37,7 +41,10 @@ int main(void)
     {
         // Update
         //----------------------------------------------------------------------------------
-        UpdateMusicStream(music);   // Update music buffer with new stream data
+        if (!AUDIO_THREAD_MUSIC_UPDATE)
```
raysan5, line 44: This mechanism looks very weird from an user perspective, I'd prefer to avoid it and look for an alternative.

src/raudio.c
```diff
@@ -476,6 +494,11 @@ void InitAudioDevice(void)
     TRACELOG(LOG_INFO, "    > Periods size:  %d", AUDIO.System.device.playback.internalPeriodSizeInFrames*AUDIO.System.device.playback.internalPeriods);
 
     AUDIO.System.isReady = true;
+
+    deviceInfo.sampleRate = AUDIO.System.device.playback.internalSampleRate;
```
raysan5, line 498: No need to store this info, it can be just queried by the users requiring it with a `GetAudioDeviceInfo()`

raysan5:
@ptarabbia Added some review comments but it would be nice if you could also split the PR into 2-3 smaller ones.

raysan5:
Manually implemented and reviewed in separate PRs.

PR #2214, opened 2021-12-16
Parse struct descriptions

eutro:
This PR updates the API parser to also include struct descriptions. The bulk of the work is rewriting part of the struct parser to read from lines instead of the buffer directly.


raysan5:
@eutro Nice improvement! Thank you very much for working on it!

Issue #2092, opened 2021-10-31
[raudio] `GetMusicTimePlayed()` reports incorrect value after the stream is restarted for XM audio


Peter0x44:
This issue can be reproduced on this example: https://www.raylib.com/examples/audio/loader.html?name=audio_module_playing
The progress bar does not reset upon pressing space, which stops and starts the stream.

This seems to be an issue exclusive to XM, as https://www.raylib.com/examples/audio/loader.html?name=audio_music_stream works correctly, and the progress bar resets when space is pressed.

raysan5:
This is a known issue, it requires some debugging...

ptarabbia:
Just sent a PR fixing this.

PR #2215, opened 2021-12-16
Fix issue #2092

ptarabbia:


raysan5:
@ptarabbia nice! thanks for the review!

PR #2216, opened 2021-12-16
Turn on interpolation for XM playback

ptarabbia:
Default was off, with no way to turn it on.


raysan5:
@ptarabbia thanks for the improvement! :D

Issue #2147, opened 2021-11-17
[build] `make install` doesn't work on RPI3


trikko:
### Issue description

```make PLATFORM=PLATFORM_RPI ```

result:

```
ar rcs ../src/libraylib.a rcore.o rshapes.o rtextures.o rtext.o utils.o rmodels.o raudio.o
raylib static library generated (libraylib.a) in ../src!
```

```sudo make install```

result:

```
mkdir --parents --verbose /usr/local/lib
mkdir --parents --verbose /usr/local/include
cp --update --verbose /home/pi/src/raylib/libraylib.a /usr/local/lib/libraylib.a
cp: impossibile eseguire stat di '/home/pi/src/raylib/libraylib.a': File o directory non esistente
make: *** [Makefile:629: install] Error 1
```

It tries to copy ```/home/pi/src/raylib/libraylib.a``` instead of ```/home/pi/src/raylib/src/libraylib.a```

### Environment

rpi3

**Comment by raysan5** [comment]
This is very weird, I'm looking at the `Makefile` and I can't find the issue. `libraylib.a` is actually generated in `$(RAYLIB_RELEASE_PATH)/libraylib.a`, so, the path should be correct. The only way to modify it is defining a custom `RAYLIB_RELEASE_PATH`...

**Comment by raysan5** [comment]
Just noted this code in line 150:
```make
# RAYLIB_SRC_PATH adjustment for different platforms.
# If using GNU make, we can get the full path to the top of the tree. Windows? BSD?
# Required for ldconfig or other tools that do not perform path expansion.
ifeq ($(PLATFORM),PLATFORM_DESKTOP)
    ifeq ($(PLATFORM_OS),LINUX)
        RAYLIB_PREFIX  ?= ..
        RAYLIB_SRC_PATH = $(realpath $(RAYLIB_PREFIX))
    endif
endif
```
Could it, maybe, be executed before assigning `RAYLIB_RELEASE_PATH  ?= $(RAYLIB_SRC_PATH)`???

colesnicov:
RPI 2 Model B
OS: RapiOS (Buster)
GL Driver: FULL_KMS

I followed the wiki instructions:
cd src && make PLATFORM = PLATFORM_RPI
and i have same errors:
```
sudo make install
mkdir --parents --verbose /usr/local/lib
mkdir --parents --verbose /usr/local/include
cp --update --verbose /home/pi/raylib/libraylib.a /usr/local/lib/libraylib.a
[...]

PR #2218, opened 2021-12-16
Fix build results path on Linux and RPi

wereii:
See #2147


raysan5:
@wereii Thanks again for the review!

PR #2226, opened 2021-12-18
LoadRenderTextureEx with custom pixel format and optional depthbuffer

rmn20:
The default LoadRenderTexture always creates a framebuffer with an alpha channel, which is not usable due to raylib limited blending capabilities.


catmanl:
Small comment, I think that `useDepth` should be of bool type

raysan5:
@rmn20 I'm afraid I'm not adding this function. User needs could vary a lot depending on their _GBuffer_ needs, for that reason `rlgl` provides an API to create custom _GBuffers_ as desired. An usage example is in development but not ready yet.

PR #2230, opened 2021-12-18
Add ImageFloodFill function

Buhlean:
ImageFloodFill fills all pixels adjacent to the given position which have the same colour, recursively.

I did my best to make it pixel format agnostic and style guide conforming. I put it with the manipulation functions (inside their ifdef); if wrong then I'll change that.
The math around the worst case A* memory scenario was weird and inconclusive (as in: there's always a worse case thinkable), so I did the safe thing and allocated the maximum needed. 

I hope it's helpful. 

Here's it in action:
FloodFillTest
I might be slow to respond atm, but I definitely will.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -2074,62 +2074,59 @@ void ImageColorReplace(Image *image, Color color, Color replace)
     ImageFormat(image, format);
 }
 // Flood-fills from the given position
-void ImageFloodFill(Image *dst, int startPosX, int startPosY, Color color){
+void ImageFloodFill(Image *dst, int startPosX, int startPosY, Color color)
+{
     if ((dst->data == NULL) || (startPosX < 0) || (startPosX >= dst->width) || (startPosY < 0) || (startPosY >= dst->height)) return;
-    
+
     int bpp = GetPixelDataSize(1,1,dst->format);
     void *start_pixel_address = (dst->data + bpp*(startPosX + dst->width*startPosY));
-    
+
     Color startColor = GetPixelColor(start_pixel_address, dst->format);
     if (startColor.r == color.r && startColor.g == color.g && startColor.b == color.b && startColor.a == color.a) return;
-    
-    struct flood_fill_pixel_pos {
-        int x, y;
-    };
-    
+
     // x,y array for pixels which we just flood-filled, conservatively sized
     // TODO: research and implement ring buffer, size could maybe be as low as ~5 * smallest image side
-    struct flood_fill_pixel_pos *a_star_edge_list = RL_MALLOC(dst->width*dst->height*sizeof(*a_star_edge_list));
+    int *a_star_edge_list = RL_MALLOC(dst->width*dst->height*2*sizeof(*a_star_edge_list));
     if (a_star_edge_list == NULL) return;
-    a_star_edge_list[0].x = startPosX;
-    a_star_edge_list[0].y = startPosY;
-    
-    for (int flood = 0, pool = 1; flood < pool; flood++)
+    a_star_edge_list[0] = startPosX;
+    a_star_edge_list[1] = startPosY;
+
+    for (int flood = 0, pool = 1; flood < pool; flood+=2)
     {
         int x = a_star_edge_list[flood].x;
         int y = a_star_edge_list[flood].y;
         void *pixel_address = 0;
-        
+
         // clockwise check and color the 4 neighbours, then record
-        struct flood_fill_pixel_pos neighbours[4] = {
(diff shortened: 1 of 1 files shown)


src/rtextures.c
```diff
@@ -2073,6 +2073,65 @@ void ImageColorReplace(Image *image, Color color, Color replace)
 
     ImageFormat(image, format);
 }
+// Flood-fills from the given position
+void ImageFloodFill(Image *dst, int startPosX, int startPosY, Color color){
```
raysan5, line 2077: Please, could you review [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?.

src/rtextures.c
```diff
@@ -2073,6 +2073,65 @@ void ImageColorReplace(Image *image, Color color, Color replace)
...
+    if ((dst->data == NULL) || (startPosX < 0) || (startPosX >= dst->width) || (startPosY < 0) || (startPosY >= dst->height)) return;
+    
+    int bpp = GetPixelDataSize(1,1,dst->format);
+    void *start_pixel_address = (dst->data + bpp*(startPosX + dst->width*startPosY));
+    
+    Color startColor = GetPixelColor(start_pixel_address, dst->format);
+    if (startColor.r == color.r && startColor.g == color.g && startColor.b == color.b && startColor.a == color.a) return;
+    
+    struct flood_fill_pixel_pos {
```
raysan5, line 2086: Please, could you avoid additional structures definition?

Buhlean: I assumed it would be local only. I think I was wrong? 
Changed it.

Buhlean:
Due to changes this is now untested code. Changed to draft.

Buhlean:
Do I change raylib.h to add this function there? Or is that done before releases only?

raysan5:
@Buhlean Function should be added to `raylib.h` but first, it seems it fails while compiling for some platforms, here the errors:
```
rtextures.c: In function ‘ImageFloodFill’:
rtextures.c:2097:37: error: request for member ‘x’ in something not a structure or union
 2097 |         int x = aStarEdgeList[flood].x;
      |                                     ^
rtextures.c:2098:37: error: request for member ‘y’ in something not a structure or union
 2098 |         int y = aStarEdgeList[flood].y;
      |                                     ^
rtextures.c:2117:42: error: ‘pixelAdress’ undeclared (first use in this function); did you mean ‘pixelAddress’?
 2117 |                 Color cc = GetPixelColor(pixelAdress, dst->format);
      |                                          ^~~~~~~~~~~
      |                                          pixelAddress
rtextures.c:2117:42: note: each undeclared identifier is reported only once for each function it appears in
```

orcmid:
> @Buhlean ... here the errors:
> rtextures.c: In function ‘ImageFloodFill’:
> rtextures.c:2097:37: error: request for member ‘x’ in something not a structure or union
>  2097 |         int x = aStarEdgeList[flood].x;
>       |                                     ^
> rtextures.c:2098:37: error: request for member ‘y’ in something not a structure or union
>  2098 |         int y = aStarEdgeList[flood].y;
> ```

aStarEdgList[] is an array of integers.   The statements should be 

'''
int x = aStarEdgeList[flood];
int y = aStarEdgeList[flood+1];
'''

Uncertain why this doesn't fail on all builds.

Also, the use of `2*sizeof(*aStarEdgeList)` might work but `2*sizeof(int)` seems more appropriate.  Either way, a reader/maintainer needs to know something about the left-hand side of the statement in which the RL_MALLOC appears.

Buhlean:
> > @Buhlean ... here the errors:
> > rtextures.c: In function ‘ImageFloodFill’:
> > rtextures.c:2097:37: error: request for member ‘x’ in something not a structure or union
> > 2097 |         int x = aStarEdgeList[flood].x;
> > |                                     ^
> > rtextures.c:2098:37: 
> Also, the use of `2*sizeof(*aStarEdgeList)` might work but `2*sizeof(int)` seems more appropriate. Either way, a reader/maintainer needs to know something about the left-hand side of the statement in which the RL_MALLOC appears.

I was taught to never, ever hardcode types into malloc, that's a class of bugs you can entirely avoid with this technique. And avoiding whole bug classes is a worthy goal imo. Especially ones you cannot auto-detect in any way once they happen. I even refactored the type of that exact line two days ago, knowing full well that I wouldn't have to hunt down all the mallocs, inevitably missing some!

I assumed it fails compilation, I edited it in github on my phone. That's why it's a draft now. As soon as I have a computer at hand I'll correct it.

PR #2232, opened 2021-12-20
Remove TODO on Makefile

catmanl:
Justification: original comment said the following
"TODO: Add other platforms. Remove sudo requirement, i.e. add USER mode."

For the other platforms part, installing is included on unix-like systems, so for example Windows doesn't have a path where to install libraries.

Removing the requirement for sudo is also quite contradictory since we're writing files to directories which require root. Not sure what the original commiter meant by USER mode.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -606,7 +606,6 @@ android_native_app_glue.o : $(NATIVE_APP_GLUE)/android_native_app_glue.c
 # for compilation and enable runtime linking with -rpath, LD_LIBRARY_PATH, or ldconfig.
 # HINT: Add -L$(RAYLIB_INSTALL_PATH) -I$(RAYLIB_H_INSTALL_PATH) to your own makefiles.
 # See below and ../examples/Makefile for more information.
-# TODO: Add other platforms. Remove sudo requirement, i.e. add USER mode.
 
 # RAYLIB_INSTALL_PATH should be the desired full path to libraylib. No relative paths.
 DESTDIR ?= /usr/local


raysan5:
@catmanl thanks for the review.

orcmid:
Not an objection to this change.  Just wondering if that is where "USER mode" came from.

The User Mode install is likely a reference to Windows installs.  There are ways to install at the user level that do not require administrative (i.e., sudo) privileges.  As I recall, the raylib install at `C:\` on Windows achieves that, though not in harmony with the practice for Windows.  (It is essentially an XCopy install, but not a great choice of location when 'C:\' is precious space, such as a SSD, and there are multiple users and/or a business owner.)

There are now better ways to install apps in a sand-boxed folder with their shared libraries (i.e., DLLs) and other resources.  This allows writing in the sandbox, saving configurations, logs, and game-states but limiting access to the file system otherwise.  Good for games.  They can be distributed in the Windows Store also, upon satisfaction of various requirements.  (Win32 apps might not be sandboxes so strictly.  I'm not clear on that.)

Something to look into, but I don't think this is on the fall-line of raylib development.

catmanl:
Hey @orcmid.
With "path to install libraries" I mean something like /usr/lib or /usr/include. Last time I checked Windows didn't have those, and "installing" seems like a Linux-only thing due to Linux actually having places where to locate these resources.

As to the User mode situation, I wasn't aware of this. In any case, a Windows user should review it and implement it :)
Cheers.

orcmid:
On Windows, it is done with environment variables and the setup of a command-line environment.  It looks this way apart from additional ones specified in the command-line session by either additions to the vars inside a session or by options to the compiler, including (my favorite) in @opt files.

VSBuildTools-2021-12-20-1442-x64NativeVars

There is a nice batch file at `raylib/projects/scripts/build-windows.bat` that illustrates some of this.

orcmid:
I also just learned that package managers such as HomeBrew on MacOS tend to have their own substructure for holding installs they do.  My speculation is that it has to do with dependency and update management.

PR #2234, opened 2021-12-21
Added "CheckCollisionRayLine()" function.

lonelyfloat:
It didn't really make much sense to me that there is complete 3D raycasting support, but not 2D raycasting support. So I added a simple raycasting function that users can expand upon to suit whatever they need.

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1714,6 +1714,8 @@ bool CheckCollisionLines(Vector2 startPos1, Vector2 endPos1, Vector2 startPos2, 
     return collision;
 }
 
+
+
 // Check if point belongs to line created between two points [p1] and [p2] with defined margin in pixels [threshold]
 bool CheckCollisionPointLine(Vector2 point, Vector2 p1, Vector2 p2, int threshold)
 {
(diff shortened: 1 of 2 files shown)


src/rshapes.c
```diff
@@ -1714,6 +1714,8 @@ bool CheckCollisionLines(Vector2 startPos1, Vector2 endPos1, Vector2 startPos2,
     return collision;
 }
 
+
+
```
JeffM2501, line 1718: Why two extra newlines?

Peter0x44:
You didn't expose it in raylib.h, so there is no way for users to call it.

src/rshapes.c
```diff
@@ -1734,6 +1736,23 @@ bool CheckCollisionPointLine(Vector2 point, Vector2 p1, Vector2 p2, int threshol
     return collision;
 }
 
+bool CheckCollisionRayLine(Vector2 rayPosition, Vector2 rayDirection, Vector2 startPos, Vector2 endPos, Vector2 *collisionPoint)
+{    
+    bool collision = false;
+    Vector2 offsetPosition;
+    offsetPosition.x = rayPosition.x + rayDirection.x;
+    offsetPosition.y = rayPosition.y + rayDirection.y;
```
JeffM2501, line 1744: This is never used as a vector, always as separate values, it may be best to skip the allocation of a vector and just use x and y values directly like the other collisions do.

src/rshapes.c
```diff
@@ -1734,6 +1736,23 @@ bool CheckCollisionPointLine(Vector2 point, Vector2 p1, Vector2 p2, int threshol
...
+
+    float denom = (startPos.x - endPos.x)*(rayPosition.y-offsetPosition.y) - (startPos.y - endPos.y)*(rayPosition.x-offsetPosition.x);
+    if (denom == 0) return collision;
+    float t = ((startPos.x - rayPosition.x)*(rayPosition.y - offsetPosition.y) - (startPos.y-rayPosition.y)*(rayPosition.x-offsetPosition.x)) / denom;
+    float u = ((startPos.x - rayPosition.x)*(startPos.y - endPos.y) - (startPos.y-rayPosition.y)*(startPos.x - endPos.x)) / denom;
+    collision = (t >= 0 && t <= 1 && u >= 0);
+    if (collision)
+        *collisionPoint = (Vector2){startPos.x + t * (endPos.x-startPos.x),startPos.y + t * (endPos.y-startPos.y)};
```
JeffM2501, line 1752: You need to verify that the passed in collisionPoint is valid also. If NULL is passed in this will crash.
Making a temporary vector2 is also not necessary, you can just directly assign into the pointer's fields. See how CheckCollisionLines does it. This function should be nearly identical to that one.

JeffM2501 (commented):
A few things.
1) You did not change the header file. Just having the code in the .c file is not enough, if it is isn't in the header file nobody can use it.
2) This is very similar to what CheckCollisionLines does, it just uses a ray. This function should look nearly the same as that.
3) The code needs to check for errors
4) The code is making some extra vector2 copies it doesn't need to

src/rshapes.c
lonelyfloat: That's fair. This was adapted from an iteration that used raymath and vector operations, so there are some remnants.

src/rshapes.c
lonelyfloat: I completely forgot about that, thanks!

src/rshapes.c
JeffM2501: I would look over the other collision code and see what you can do to match the style.

lonelyfloat: Good point, doing that rn.

raysan5:
@lonelyfloat As discussed on Discord, I prefer to keep that function out of raylib for now.

PR #2236, opened 2021-12-21
Re-enable QOI support

GithubPrankster:
With the **QOI** format now being solidified, *Raylib* users can begin to utilize it without much worry, specially as plugins for using it are being made for various viewers and photo editors! Hooray to Dominic and the many programmers involved in this project.


raysan5:
@GithubPrankster Nice! Thanks for the review! 👍😄

PR #2238, opened 2021-12-22
add the raylib-scopes binding

salotz:


raysan5:
@salotz thanks for the addition! 👍😄

PR #2240, opened 2021-12-23
fix scopes binding URL

salotz:
Sorry about this, there was a small typo in the URL for the language in the original PR.


raysan5:
@salotz It would be recommended that it was an https for security, instead of http.

salotz:
I'll try to get the maintainer to move to https

PR #2241, opened 2021-12-24
Check before freeing image.data

planetis-m:
Hi, while [trying](https://github.com/greenfork/nimraylib_now/issues/56) to make RAII like resource management for the Nim bindings, I noticed that not all Unload functions check for not nil before freeing. Others that need checks are the ones for Mesh, Material, Model and ModelAnimation structs.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -457,7 +457,7 @@ Image LoadImageFromScreen(void)
 // Unload image from CPU memory (RAM)
 void UnloadImage(Image image)
 {
-    RL_FREE(image.data);
+    if (image.data != NULL) RL_FREE(image.data);
 }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@planetis-m it's not checked because `free(null)` is valid in C and it has no effect. Also note that it uses a macro that users can customize for their needs in case they desire some specific behaviour for a null parameter.

planetis-m:
But... functions in raudio.c do this: https://github.com/raysan5/raylib/blob/4874c45ebff0249d2d1acfb612a7f81638f25238/src/raudio.c#L865

raysan5:
@planetis-m then that's an issue to review on raudio 😄

PR #2243, opened 2021-12-24
Check not null before accessing by index

planetis-m:


raysan5:
@planetis-m thanks for the improvement!

PR #2246, opened 2021-12-26
Added normal support

Fallbork:
I've tried to add normal support to the shaders because I noticed (after a lot of headaches) that all values of vertexNormal were 0. I commented everything that I changed since I'm not the most experienced with OpenGL...
My biggest problem was that "else" after the VAO support check... I couldn't, for the life of me, get the normals to bind to the VAO and appear in the shader; did I miss something? Please do tell me if did so I can add it and revert the commenting I had to do on the else block! :smile:


raysan5:
@Fallbork Sorry, no plans to support normals for default internal buffers, usually shapes/textures do not require them for 2d games and adding them hits performance with an extra buffer to process and upload every draw.

PR #2250, opened 2021-12-27
[Feature] Draw text with shadows

tusharsingh09:
Added new functions:
`DrawTextShadowed()` and `DrawTextShadowedEx()`
Tested on Windows and works fine, further inspection maybe required.
Both take an `offset` value which is a `Vector2` and deviates the shadow from it's original position by the offset.
shadow_text


raysan5:
This effect could be acomplished by users with currently provided functions.

PR #2253, opened 2021-12-29
Added: missing /

shivajiva101:
Specify root as the starting point for the linux compiler path


raysan5:
@shivajiva101 thanks for the improvement!

PR #2254, opened 2021-12-30
[Feature] Added SetWindowOpacity(float opacity)

tusharsingh09:
`SetWindowOpacity()` takes `opacity`(`float`) as a parameter. Value of `opacity` must fall under 0 and 1. If it exceeds 1; it will be equal 1, similarly if it's smaller than 0; opacity is equal to 0.
Random Screenshots (*READ THE VALUE INSIDE SetWindowOpacity()*):
set_window_opacity_0
set_window_opacity_1
set_window_opacity_2
set_window_opacity_3
 `SetWindowOpacity(99999999.0)` goes same as `SetWindowOpacity(1.0)`.

Thank you

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1534,6 +1534,16 @@ void SetWindowSize(int width, int height)
 #endif
 }
 
+// Set window opacity, value opacity is between 0.0 and 1.0
+void SetWindowOpacity (float opacity)
+{
+#if defined(PLATFORM_DESKTOP)
+    if (opacity >= 1.0) opacity = 1.0;
+    else if (opacity <= 0.0) opacity = 0.0;
+    glfwSetWindowOpacity(CORE.Window.handle, opacity);
+#endif
+}
+
 // Get current screen width
 int GetScreenWidth(void)
 {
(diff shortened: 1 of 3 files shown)


src/rcore.c
```diff
@@ -1534,6 +1534,16 @@ void SetWindowSize(int width, int height)
 #endif
 }
 
+// Set window opacity, value opacity is between 0.0 and 1.0
+void SetWindowOpacity (float opacity)
```
raysan5, line 1538: It seems there is an extra space before parenthesis, please review it.

src/rcore.c
```diff
@@ -1534,6 +1534,16 @@ void SetWindowSize(int width, int height)
...
+{
+#if defined(PLATFORM_DESKTOP)
+    if (opacity >= 1.0) opacity = 1.0;
```
raysan5, line 1541: float values should end with `f` --> `1.0f`, `0.0f`

raysan5:
@tusharsingh09 This is a nice addition, added a couple of comments for review.

src/rcore.c
tusharsingh09: I'm really sorry, it was intentional I thought I read it in CONTRIBUTING.md but anyways I'll fix it.

src/rcore.c
tusharsingh09: Yup, okay

raysan5:
@tusharsingh09 thanks for the addition!

PR #2256, opened 2021-12-30
[CORE] Add a function to return the application directory.

JeffM2501:
This PR adds `GetApplicationDirectory `as a function. It uses platform specific code to find the folder that the executable is running in. This is a good path to pass to `ChangeDirectory `so that you ensure you have a good starting working directory.

While this does call platform specific code, it does not include `windows.h`
This is a feature that GLFW should provide, but does not, so we need to provide to to users ourselves.


raysan5:
@JeffM2501 Thanks for addition!

PR #2257, opened 2022-01-01
Update BINDINGS.md

ArnautDaniel:
Adding license for Gforth bindings.


raysan5:
@ArnautDaniel Thanks for the improvement!

PR #2258, opened 2022-01-02
Update Raylib-J version/licence info

CreedVI:


raysan5:
@CreedVI Thanks for the update!

PR #2260, opened 2022-01-02
Change comment about custom blend modes in rlgl.h

glorantq:
Changed a reference to a now non-existent function in rlgl.h to the one that should be used to set custom blend modes. Also fixed the typo.


raysan5:
@glorantq Thanks for the fix! :D

PR #2263, opened 2022-01-03
Register gamepads before first call to EndDrawing

techbech:
__Problem__
Gamepads is not detected by glfw before first call to `EndDrawing` resulting in function like `GetGamepadName` to return null.

__Solution__
Update connection state of joystick in `InitGraphicsDevice` so that the state is ready after calling `InitWindow`.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4507,6 +4507,16 @@ static bool InitGraphicsDevice(int width, int height)
     rlLoadExtensions(eglGetProcAddress);
 #endif
 
+#if defined(PLATFORM_DESKTOP)
+    // Check if gamepads are ready
+    // NOTE: We register gamepads here to make sure that ready state of the gamepads represents the glfw state before the first call to EndDrawing()
+    for (int i = 0; i < MAX_GAMEPADS; i++)
+    {
+        if (glfwJoystickPresent(i)) CORE.Input.Gamepad.ready[i] = true;
+        else CORE.Input.Gamepad.ready[i] = false;
+    }
+#endif
+
     // Initialize OpenGL context (states and resources)
     // NOTE: CORE.Window.currentFbo.width and CORE.Window.currentFbo.height not used, just stored as globals in rlgl
     rlglInit(CORE.Window.currentFbo.width, CORE.Window.currentFbo.height);


raysan5:
@techbech It sounds like an issue with GLFW, not raylib...

raysan5:
@techbech Please note that I try to avoid (as much as possible) adding new functions to the library, I prefer to duplicate code when required. In this case it's only a couple of lines.

Also notes that your code is not following the [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md). It should be reviewed.

techbech:
> @techbech It sounds like an issue with GLFW, not raylib...

No. It's caused by the internal gamepad ready state check at `raylib/src/rcore.c:3454`. And the ready state is first set in PollInputEvents, meaning that the ready state wont be set before the first `EndDrawing` call.

techbech:
@raysan5 I think that is fulfill the convention now

raysan5:
@techbech If I understand it correctly, it only fails before the first frame is finished, right? How dramatic is it for users? Myabe it can just be ignored? In any case, another solution would be just calling `glfwPollEvents()` on `InitWindow()`.

techbech:
> @techbech If I understand it correctly, it only fails before the first frame is finished, right? How dramatic is it for users? Maybe it can just be ignored? In any case, another solution would be just calling `glfwPollEvents()` on `InitWindow()`.

Calling `glfwPollEvents()` will not fix the issue. The issue is the initial raylib gamepad state when entering the main loop. Every gamepad has a `ready` property that is being used in functions `GetGamepadName` to determine if the gamepad is present, if not it then returns NULL. When calling `PollInputEvents` the `ready` property of the gamepads is updated, and this is fine except for the codes that come before the first `EndDrawing` call. So it's not a dramatic issue but I don't see the harm in adding this fix. It's for you to decide.

src/rcore.c
```diff
@@ -4507,6 +4507,16 @@ static bool InitGraphicsDevice(int width, int height)
     rlLoadExtensions(eglGetProcAddress);
 #endif
 
+#if defined(PLATFORM_DESKTOP)
+    // Check if gamepads are ready
+    // NOTE: We register gamepads here to make sure that ready state of the gamepads represents the glfw state before the first call to EndDrawing()
+    for (int i = 0; i < MAX_GAMEPADS; i++)
+    {
+        if (glfwJoystickPresent(i)) CORE.Input.Gamepad.ready[i] = true;
```
raysan5, line 4515: That code shouldn't be in `InitGraphicsDevice()`, it's not related to graphics device.

raysan5:
I'm not merging this PR for now.

Issue #2255, opened 2021-12-30
[parser] raylib `JSON` api generated function parameters order not mandatory


VincentToups:
I'm not sure how serious this is from your point of view, but I think its worth raising the issue.

Raylib includes a raylib_api.json file which encodes the API as JSON objects. The particular issue here is that the JSON objects encoding functions use a key/value JSON object to represent the parameters of the function. While these parameters have a lexical order in the file, a conforming JSON parser and implementation is [not obligated to preserve key order](https://www.json.org/json-en.html). I'm not sure how often it happens that a typical JSON reader fails to preserve key order, however, but I can find a few issues on stack exchange related to this expectation being violated (for [instance](https://stackoverflow.com/questions/4515676/keep-the-order-of-the-json-keys-during-json-conversion-to-csv)).

Here is a snippet from the api json file showing the key/value pairs being used to encode an order.

```
{
      "name": "InitWindow",
      "description": "Initialize window and OpenGL context",
      "returnType": "void",
      "params": {
        "width": "int",
        "height": "int",
        "title": "const char *"
      }
    }
```

I'd propose representing the argument list as a JSON array with objects containing name and type.

PR #2264, opened 2022-01-04
JSON parser: Use array for function params (#2255)

gtrxAC:
Related issue: https://github.com/raysan5/raylib/issues/2255
Uses arrays with objects containing name and type instead of objects with name as key and type as value for function parameters in the JSON raylib parser.

### DIFF
--- a/parser/raylib_api.json
+++ b/parser/raylib_api.json
@@ -2490,11 +2490,20 @@
       "name": "InitWindow",
       "description": "Initialize window and OpenGL context",
       "returnType": "void",
-      "params": {
-        "width": "int",
-        "height": "int",
-        "title": "const char *"
-      }
+      "params": [
+        {
+          "name": "width",
+          "type": "int"
+        },
+        {
+          "name": "height",
+          "type": "int"
+        },
+        {
+          "name": "title",
+          "type": "const char *"
+        }
+      ]
     },
     {
       "name": "WindowShouldClose",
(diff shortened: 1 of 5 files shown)


parser/raylib_parser.c
```diff
@@ -1001,14 +1001,17 @@ static void ExportParsedData(const char *fileName, int format)
...
                 else
                 {
-                    fprintf(outFile, ",\n      \"params\": {\n");
+                    fprintf(outFile, ",\n      \"params\": [\n");
                     for (int p = 0; p < funcs[i].paramCount; p++)
                     {
-                        fprintf(outFile, "        \"%s\": \"%s\"", funcs[i].paramName[p], funcs[i].paramType[p]);
+                        fprintf(outFile, "        {\n");
+                        fprintf(outFile, "          \"name\": \"%s\",\n", funcs[i].paramName[p]);
+                        fprintf(outFile, "          \"type\": \"%s\"\n", funcs[i].paramType[p]);
```
raysan5, line 1009: Maybe we could exchange order of those lines to follow C convention of (`type` `name`).

parser/raylib_api.lua
```diff
@@ -2636,6 +2636,14 @@ return {
         {name = "height", type = "int"}
       }
     },
+    {
+      name = "SetWindowOpacity",
+      description = "Set window opacity [0.0f..1.0f] (only PLATFORM_DESKTOP)",
+      returnType = "void",
+      params = {
+        {name = "opacity", type = "float"}
```
raysan5, line 2644: Maybe we could exchange order of those lines to follow C convention of (`type` `name`).

parser/raylib_api.json
```diff
@@ -2490,11 +2490,20 @@
...
       "description": "Initialize window and OpenGL context",
       "returnType": "void",
-      "params": {
-        "width": "int",
-        "height": "int",
-        "title": "const char *"
-      }
+      "params": [
+        {
+          "name": "width",
```
raysan5: Maybe we could exchange order of those lines to follow C convention of (`type` `name`).

raysan5:
@gtrxAC Thanks for the improvement!

planetis-m:
So this broke my parser as expected, but the issue is the params are now flipped, instead of name, type which what FieldInfo uses by the way, they are now type, name. At least make the consistent and if possible don't follow the C syntax. For example `{"name": "name[10]", "type": "char"` is hard to parse because we need to take into account the field "name" as well to infer the type.

PR #2265, opened 2022-01-04
Simd version of DrawModelEx and DrawMesh and its corresponding math functions

Abdelilah-Majid:
this my attempt at optimizing the single threaded DrawModelEx function, i dont promise any magical speed improvement its just speed it up a little bit about 5 fps depending on the scenario on my old low end machine(this could differ from machine to another)
and again its not a big thing but if you want to look at the perf report of my game engine when using both none simd and simd versions of thees functions i will leave a log file here that include only the top cycles consumers(note my game engine binary is called linux_debug for now thats why you will see a lot of it)
[perf_log.log](https://github.com/raysan5/raylib/files/7808510/perf_log.log)

but if dont like to read 500 lines of log
here is the interesting  part 
DrawModelEx and DrawModel doesnt seem to improve that much but here

(Overhead)      ( sys )     ( usr )     ( Samples )      ( Period  Command )       ( Shared Object )          ( Symbol )
   1.28%     0.00%     1.28%         14540    3167025347  linux_debug     linux_debug              [.] MatrixMultiply                     
   0.03%     0.00%     0.03%           288      62779350  linux_debug     linux_debug              [.] MatrixMultiply_simd                    

you can do the calculation

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -199,9 +199,9 @@ endif
 # Define raylib graphics api depending on selected platform
 ifeq ($(PLATFORM),PLATFORM_DESKTOP)
     # By default use OpenGL 3.3 on desktop platforms
-    GRAPHICS ?= GRAPHICS_API_OPENGL_33
+    #GRAPHICS ?= GRAPHICS_API_OPENGL_33
     #GRAPHICS = GRAPHICS_API_OPENGL_11  # Uncomment to use OpenGL 1.1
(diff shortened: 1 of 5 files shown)


raysan5:
@Abdelilah-Majid Thanks for the PR but I'm afraid I'm not adding it to raylib, this is a big change, very specific for some users needs and very difficult to maintain.

PR #2269, opened 2022-01-08
Added defines to parser

iskolbin:
This PR adds parsing of defines to parser. My main goal was to parse constants and maybe types of them (i.e. stings or numbers). The code parses all defines in file, and put them in the `defs` array which contains `DefInfo`:
```
typedef enum { UNKNOWN = 0, MACRO, GUARD, INT, LONG, FLOAT, DOUBLE, CHAR, STRING, COLOR } DefineType;

typedef struct DefineInfo {
    char name[64];    // Define name
    DefineType type;  // Define type
    char value[256];  // Define value
    char desc[128];   // Define description
} DefineInfo;
```
The parsing process is like: skip leading spaces, skip `#define`, skip spaces, extract `name`. By analysing chars after `name` we can get (or assume) `type`. The `type` is one of:
* UNKNOWN
* MACRO - has closing bracket in name
* GUARD - has nothing after name (or comments), i.e. `#define RAYLIB_H`
* INT - value starts with number and not FLOAT/DOUBLE/LONG
* LONG - value starts with number, doesn't have point, ends on `L`
* FLOAT - value has point and ends on `f`
* DOUBLE - value has point and not ends of `f`
* CHAR - value starts with `'`
* STRING - value starts with `"`
* COLOR - value starts with `CLITERAL(Color)`


mandar1jn:
This is nice and all, but for bindings a value would also be appreciated where applicable

iskolbin:
@mandar1jn totally agree, and this is not that hard but there are some problems rising from the fact that defines are preprocessed:

1. what should we do if we have multiple defines like so:
```
#ifdef SOMECASE
  #define VALUE 1
#else
  #define VALUE 2
#endif
```
2. what should we do if we change define during compilation using `-D` flags?

My guess is just to stick to the first parsed value encountered but probably there are better solutions

iskolbin:
Updated a bit, added values and descriptions

raysan5:
@iskolbin this is a great addition but I'm agree with your concerns about the special situations...

How are the `#defines` managed by raylib bindings?

orcmid:
> #ifdef SOMECASE
>   #define VALUE 1
> #else
>   #define VALUE 2
> #endif

> 2. what should we do if we change define during compilation using `-D` flags?

If your example is in C Language, the `-D` flags are global to the source codes being compiled.  So `-DSOMECASE` has the specified effect.  `-DVALUE=0` might be useless; whether that causes a redefinition warning from the preprocessor of the compiler might be an issue.

If you are talking about CMake, it is more complicated and also depends on whether the compiler also accepts definitions in console-session environment variables.

iskolbin:
@raysan5 @orcmid To be clear the problem arises not in `raylib.h` itself but for instance in `rlgl.h`, i.e. something like https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L192:
```
 191 // Default internal render batch elements limits
 192 #ifndef RL_DEFAULT_BATCH_BUFFER_ELEMENTS
 193     #if defined(GRAPHICS_API_OPENGL_11) || defined(GRAPHICS_API_OPENGL_33)
 194         // This is the maximum amount of elements (quads) per batch
 195         // NOTE: Be careful with text, every letter maps to a quad
 196         #define RL_DEFAULT_BATCH_BUFFER_ELEMENTS  8192
 197     #endif
 198     #if defined(GRAPHICS_API_OPENGL_ES2)
 199         // We reduce memory sizes for embedded systems (RPI and HTML5)
 200         // NOTE: On HTML5 (emscripten) this is allocated on heap,
 201         // by default it's only 16MB!...just take care...
 202         #define RL_DEFAULT_BATCH_BUFFER_ELEMENTS  2048
 203     #endif
 204 #endif
 205 #ifndef RL_DEFAULT_BATCH_BUFFERS
 206     #define RL_DEFAULT_BATCH_BUFFERS                 1      // Default number of batch buffers (multi-buffering)
 207 #endif
```
How bindings solve such things? As I understand: they don't. For instance C# https://github.com/NotNotTech/Raylib-CsLo/blob/main/Raylib-CsLo/autogen/bindings/RlGl.cs#L456:
```
        [NativeTypeName("#define RL_DEFAULT_BATCH_BUFFER_ELEMENTS 8192")]
        public const int RL_DEFAULT_BATCH_BUFFER_ELEMENTS = 8192;

        [NativeTypeName("#define RL_DEFAULT_BATCH_BUFFERS 1")]
        public const int RL_DEFAULT_BATCH_BUFFERS = 1;
```
I.e they are just hardcoded. So I don't think there is *good* solution due to nature of defines.

iskolbin:
@raysan5 Could you please look at this again? I mean should we improve/revisit the changes (probably you have ideas for improvements?) or it's not that useful and we should just close this PR?

orcmid:
> @raysan5 @orcmid To be clear the problem arises not in `raylib.h` itself but for instance in `rlgl.h`, i.e. something like https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L192:
> 
```C
191 // Default internal render batch elements limits
192 #ifndef RL_DEFAULT_BATCH_BUFFER_ELEMENTS
193     #if defined(GRAPHICS_API_OPENGL_11) || defined(GRAPHICS_API_OPENGL_33)
...
196         #define RL_DEFAULT_BATCH_BUFFER_ELEMENTS  8192
197     #endif
198     #if defined(GRAPHICS_API_OPENGL_ES2)
...
202         #define RL_DEFAULT_BATCH_BUFFER_ELEMENTS  2048
203     #endif
204 #endif
> ```
```

It strikes me that a bigger question here is related to the fact that `RL_DEFAULT_BATCH_BUFFER_ELEMENTS` can fail to be defined at all in that sequence of pre-processor directives.  The code seems to depend on one of the three `GRAPHICS_API_OPENGL_mumble` defines always being set and there might need to be some assurance about that.

I fail to understand the problem with the way `RL_DEFAULT_BATCH_BUFFER_ELEMENTS` is set however, unless there is an interference with another header file `#include` that is incorporated in compilation of the same C/C++ language file.

Could there be a misunderstanding about how the preprocessing works to effectively produce a source code that has all preprocessor dependencies removed and is what is actually compiled (apart from file and line number preservations to the original files in diagnostics and the debugger)?

raysan5:
> @raysan5 Could you please look at this again? I mean should we improve/revisit the changes (probably you have ideas for improvements?) or it's not that useful and we should just close this PR?

Actually I have mixed feelings about this PR, from one side I think it's great to be able to parse the `#defines` in raylib, it's the only part missing to parse and this PR adds that extra functionality. On the other side, I think it's difficult to use the parsed data for a binding... Let keep this PR open until while I think about it a bit more...

orcmid:
Oh, now I understand the problem more.

It is necessary to distinguish between independent defines and dependent ones.  Like `RL_DEFAULT_BATCH_BUFFER_ELEMENTS` could of course be defined in a -D or before an #include that hits `rlgl.h` but it matters more that its definition be dependent on `GRAPHICS_API_OPENGL_11`, `GRAPHICS_API_OPENGL_33`, and 'GRAPHICS_API_OPENGL_ES2`.  There is no way to determine the actual case of the matter from inspection of that portion of `rlgl.h` by syntax alone.

I am rather uncertain what it means to derive a binding by parsing in the case of raylib; I do think it takes a different technique, usually one involving annotation of comments in some machine-processable form, as is done for some kinds of embedded documentation.  This has to be done with understanding of the intention behinds the #defines and the conditional use of them.

orcmid:
While accomplishing #define forensics in a way that is not useful for what I think it means to create bindings, you might find the code of [VCshowDefs.c](https://github.com/orcmid/maiko/blob/windev/windev/src/VCshowDefs.c) to be amusing.  The result I got in the particular investigation is at [VCshowDefs.txt](https://github.com/orcmid/maiko/blob/windev/windev/src/VCshowDefs-2021-11-28-1407.txt).

It would be useful to put the #define for SHOW and some supporting items in an #include.  To do that, I would need to get the string.h and stdio.h included safely there.  I have to know some specific things about those files to make that work, unless duplicate #includes of them are already handled properly.

raysan5:
@iskolbin I'm merging the improvement into raylib-parser but as a reference for future users I'd prefer not to export the defines into the .json, .xml and .txt for now. Please, could you review this PR to just keep the `raylib_parser.c` update?

iskolbin:
@raysan5 ok, updated files from current head

raysan5:
@iskolbin Thank you very much for the quick response and all the effort you have put into this new feature. Thanks!

PR #2270, opened 2022-01-08
Fixed Compiling Raylib for android Using cmake

hero2002:


cmake/LibraryConfigurations.cmake
```diff
@@ -40,10 +40,18 @@ elseif (${PLATFORM} MATCHES "Android")
     set(PLATFORM_CPP "PLATFORM_ANDROID")
     set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
-    
+    set(raylib_sources
```
raysan5, line 43: It seems `raudio.c` is missing...

hero2002: > It seems `raudio.c` is missing...
yes it's missing but it's missing in the main cmake where I took so It's not from my side

hero2002: so I will push another pull to fix this and the main one

raysan5: I think it should be added and tested.

hero2002: I will add it as option too

hero2002: wait a minute it's there

hero2002: in src/CMakeLists.txt line 45

raysan5:
@hero2002 thanks for the improvement

PR #2271, opened 2022-01-08
Split build.zig

Not-Nik:
This change enables Zig programs to link raylib by adding the following to its build script:

```zig
const rl = @import("raylib/src/lib.zig").Pkg("raylib/src");

// ...

const exe = b.addExecutable(name, path);

// Add more settings

exe.linkLibrary(rl.addRaylib(exe.builder, exe.target));
```

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,5 +1,77 @@
 const std = @import("std");
 
+pub fn Pkg(srcdir: []const u8) type {
+    return struct {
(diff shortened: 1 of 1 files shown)


raysan5:
@Not-Nik I prefer to keep it in a single file, is it possible?

Not-Nik:
Done! I forced push to keep the tree clean

Not-Nik:
Actualy, please wait before merging, I need to change some things

raysan5:
@Not-Nik Thanks for the improvement and the review!

Issue #2273, opened 2022-01-09
[Documentation/Wiki] Compiling for wasm/html5 on Linux


hidden-r4d:
In the [Wiki,](https://github.com/raysan5/raylib/wiki/Working-for-Web-(HTML5)) it instructs to compile for linux using the command `make PLATFORM=PLATFORM_WEB -B`, however, this does not pass the enviornment variables to `make`, (using GNU Make 4.3 on Ubuntu) resulting in an unsuccessful compilation, with `make` unable to find `emcc` and the other emsdk tools. I've found out (from [this stackoverflow post](https://stackoverflow.com/questions/70141302/compiling-raylib-for-html5-wasm)) that passing `-e` to the options is required, and removes this issue.

RobLoach:
It should likely use `emmake make` rather than simply `make` too, but I'm not entirely sure myself. I've run into similar issues myself with the Makefiles, so had switched over to CMake.

Would you mind trying out the change to the project template I made over at https://github.com/raysan5/raylib/pull/2274 ? It adds some documentation to the CMake project template, and enables web compatibility.

PR #2274, opened 2022-01-10
cmake: Add Web and Docs to CMake project template

RobLoach:
I found the CMake documentation for building for web was lacking (as mentioned in https://github.com/raysan5/raylib/issues/2273), so I've updated the project template to help improve it. This does two things...

1. Adds a Web build to the CMake project template
2. Adds a README.md with documentation on how to use the CMake project template


raysan5:
@RobLoach thanks for the improvement!

PR #2276, opened 2022-01-10
Fixes an error in the spelling of "appearance"

ArchieAtkinson:


raysan5:
@ArchieAtkinson thanks for the fix!

PR #2281, opened 2022-01-13
Restore rlLoadShaderCode behavour

### LINKED ISSUE #2280: [rlgl] LoadShader behaviour change
---

### Issue description

When only a vertex or a fragment shader is passed to `LoadShader`, the shader program fails rather than falling back to the default raylib vertex/fragment shader. E.g. `LoadShader(0, "shader.fs")`. This is a change in behaviour introduced in 7a159664.

### Environment

Platform: Desktop
OS: Linux
OpenGL Version: 460 (shaders are using 330)

### Code Example

Example available at: https://github.com/AbyssalThistle/shader-loading-issue/https://github.com/AbyssalThistle/shader-loading-issue/blob/develop/src/main.c

MatthewOwens:
This would close #2280 by setting vertexShaderId to RLGL.State.defaultVShaderID if vsCode is null, same for the fragment shader.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -3544,8 +3544,16 @@ unsigned int rlLoadShaderCode(const char *vsCode, const char *fsCode)
     unsigned int vertexShaderId = 0;
     unsigned int fragmentShaderId = 0;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@MatthewOwens Nice! Thanks for reviewing this issue! :D

PR #2285, opened 2022-01-14
[CORE] Bug and formating fixes for GetApplicatonDir

JeffM2501:
My initial PR was not set to the correct formatting standards for raylib and there were cases that would not return a valid path. This Update fixes the formatting and should always return a valid path, even if the functions fail to find the exe dir.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -166,29 +166,13 @@
         #ifndef MAX_PATH
             #define MAX_PATH 1025
         #endif
-        void *LoadLibraryA(void *lpLibFileName);
-        void *LoadLibraryW(void *lpLibFileName);
-
-        #ifdef UNICODE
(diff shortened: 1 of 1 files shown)


JeffM2501:
@Peter0x44 can you see if this helps wit the problem?

Peter0x44:
It now returns .\ instead of an empty string, so seems all good to me.
I wasn't really expecting this to work under wine anyway

JeffM2501:
I removed the call to load the DLL with the 'ex' version of the function to get the module path, it now calls the normal kernel32 version. Raylib already includes Kernel32.lib/dll so it links without an new dependencies or DLL loads. This should make it work on a much larger range of windows versions (including wine).

raysan5:
@JeffM2501 Wow! An improvement that removes code! Those are the best ones! 👍😄

PR #2288, opened 2022-01-16
change make for header

GuvaCode:


raysan5:
@GuvaCode I'm afraid this change is not required.

PR #2290, opened 2022-01-17
[CORE] Fixes for GetApplicationDirectory on 32 bit windows builds

JeffM2501:
This PR fixes an issue with the function signatures of the windows calls used by GetApplicationDirectory so that they work on 32 bit builds. No functionality has changed.


JeffM2501:
Note that without this change, 32 bit builds on windows (with Visual Studio, and GCC) are broken.

raysan5:
Thanks for the fix @JeffM2501!

Issue #1680, opened 2021-03-25
[rglfw] Support continue drawing while windows is being resized


lazaroblanc:
I've already asked this question over [on the Discord server](https://discord.com/channels/426912293134270465/427518168995725317/824759090756583485).

I'd like for my windows to continue drawing while they are being resized.
At the moment drawing is interrupted until resizing of the window has stopped.

This issue has been addressed over [on Stack Overflow](https://stackoverflow.com/questions/45880238/how-to-draw-while-resizing-glfw-window).

![image](https://user-images.githubusercontent.com/40198445/112549430-17f27380-8dbe-11eb-9faf-b96f0782bd11.png)

GLFW would allow such a feature but according to @raysan5 this would need a library redesign to be supported in raylib.

Maybe we can get this in a future major version? (v4 or higher?)

lazaroblanc:
Can we update this to "(...) while window is being dragged & resized" ?

raysan5:
I've been investigating if this is feasible and it seems with current raylib implementation it's not possible, users should work at a lower level to accomplish this. In any case, if anyone feels brave enough to investigate it further and find a possible solution, you are welcome.

beetaa:
@efejjota @raysan5 @IanBand  I found this pr in glfw repo. [Win32: Do the event processing during window resize/move](https://github.com/glfw/glfw/pull/1426)

Not-Nik:
This could possibly be resolved by polling GLFW events on a seperate thread and syncing them up with the main thread later. This would require a C version bump to C11, to use `threads.h` and `stdatomic.h` as well as a minor increase in internal complexity from what I can tell, but it would be virtually invisible to the end user. The most challenging part of this, would be the proper handling of resizing, because `WindowSizeCallback` calls `SetupViewport`, which in turn calls `glViewport` via `rlViewport` and OpenGL isn't thread-safe.

Since Microsoft's MSVC doesn't support C11's threading anyways, we could also add a library to handle threading or write some functions into raylib (I'm thinking rthreads).

I'd be willing to type up a prototype, but I can totally understand if this feels like too much of a complexity increase.

orcmid:
Microsoft has announced the addition of C11 atomics, threads, and **`complex`** to Visual Studio C/C++ support.  `stdatomics.h` will be available immediately with [Visual Studio 17.5 Preview 2](https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/).  

[...]

PR #2291, opened 2022-01-17
Support continue drawing while windows is being resized

efejjota:
Solves #1680

The strategy is fully opt-in and fully backward compatible with any existing code without modifications.

The implementation allows calling any custom code while resizing.
However, the most useful things to run while resizing are:

- Running the exact same code used for drawing each frame
- Running some custom code which draws some placeholder screen

I have included two examples showing these two usecases.
Tested on: Windows 10 64bit

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -392,6 +392,8 @@ CORE = \
     core/core_loading_thread \
     core/core_quat_conversion \
     core/core_window_flags \
+    core/core_resize_drawing \
+    core/core_resize_placeholder \
     core/core_split_screen \
     core/core_smooth_pixelperfect \
     core/core_custom_frame_control
--- a/examples/core/core_resize_drawing.c
+++ b/examples/core/core_resize_drawing.c
@@ -1,0 +1,40 @@
+/*******************************************************************************************
+*
+*   raylib [core] example - window redraws while resizing
+*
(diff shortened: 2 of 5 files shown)


efejjota:
Tested on MacOS 12.1
It seems MacOS supports continue drawing while windows is being resized out-of-the-box without this change.

We may want to use conditional compilation to exclude this change for MacOS to not waste computing power by refreshing the window twice.

efejjota:
I'm having issues applying this solution to other examples.
Specifically applying this solution to "core_window_letterbox" doesn't work well.

This change looks like a bad idea for several reasons:
- Calling a draw function from another thread (resize thread) is dangerous and difficult to manage if we use global variables.
- Drawing issues while resizing don't occur in MacOS (not sure if Linux is affected). Adding a new method that is only required in Windows and which adds so much complexity doesn't feel right.
- It forces programmers to structure their program in a specific way, using complex features such as function pointers, etc.

I will try to find another solution which doesn't incur in these issues.

raysan5:
@efejjota I agree with those points, actually it was the reason I did not merged the PR before, I needed to review it more carefully and consider all the side cases, platforms and user implications. I think there was some similar proposal long time ago but I can't find it right now.

efejjota:
I'm no longer investigating this. Will close it for now.

PR #2293, opened 2022-01-18
Review code conventions for specific defines

catmanl:
Unless these imports require to have this format, the code has been adjusted to the conventions


raysan5:
@catmanl thanks for the review! 👍

PR #2297, opened 2022-01-21
Feature: glTF scene support

royqh1979:
Now raylib doesn't support loading/drawing scenes and nodes data in glTF 2.0 files.
This PR adds supports to that.


raysan5:
@royqh1979 This is a big change for the library and I'm afraid it's out of scope for raylib at this moment. It also implies modifying `cglft` library that does not belong to raylib and does not implement the required functionality.
For now, this change is out ouf scope.

PR #2298, opened 2022-01-21
fix: glTF: material color won't be loaded if there's no texture for that material 

royqh1979:
When loading a gltf file, if there's no texture setting for a material, it's material won't be loaded(become WHITE by default).

This pr fixes that.


raysan5:
@royqh1979 Thanks for the review!

PR #2304, opened 2022-01-22
Fix GetApplicationDirectory on macOS

Not-Nik:
Previously failed to build with an implicit declaration of `_NSGetExecutablePath`.


raysan5:
@Not-Nik Thanks for the fix!

PR #2307, opened 2022-01-25
Add Vector2DivideValue to raymath.h

Auios:
I noticed there was no Vector2Divide by value.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -318,6 +318,14 @@ RMAPI Vector2 Vector2Divide(Vector2 v1, Vector2 v2)
     return result;
 }
 
+// Divide vector by value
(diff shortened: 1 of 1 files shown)


Auios:
Oh - I noticed this was discussed in https://github.com/raysan5/raylib/pull/1197
(ops)
Looks like you concluded `Vector2DivideValue` wasnt needed because you had `Vector2Scale` already.
but in order for us to use `Vector2Scale` we need to do `Vector2 value = Vector2Scale(otherVector, a / b)` but maybe I already have the value on hand that I want to divide it by?
`Vector2 value = Vector2DivideValue(otherVector, a)`

raysan5:
This functionality can be aaccomplished with `Vector2Scale()`, no need for an additional function.

PR #2308, opened 2022-01-26
Add DrawTextCodepoints

siddharthroy12:
I've added the DrawTextCodepoints function, please review


src/rtext.c
```diff
@@ -1,4 +1,4 @@
-/**********************************************************************************************
```
raysan5, line 1: why did you delete this line?

raysan5:
@siddharthroy12 thanks, I added a review.

src/rtext.c
siddharthroy12: Oops, didn't notice that

siddharthroy12: Fixed in 13f4bcc

raysan5:
@siddharthroy12 thanks for the addition!

PR #2316, opened 2022-01-29
[CORE] Added `InitWindowAt()`

tusharsingh09:
This function provides same functionality as `InitWindow()` only difference is that `InitWindow()` creates the window in the center of the screen but this function will create the window at the position we give it as extras parameters (an X and Y). It also clamps the window to the top left corner of the screen if the user specifies it to be less than the origin because it is harder to pull out a window from top left without borders. A screen shot is attached below:
InitWindowAt
Thank you


tusharsingh09:
Had issue need to fix it

PR #2317, opened 2022-01-29
[FEATURE] Added `InitWindowAt()`

tusharsingh09:
This function is similar to SDL's window creation API that creates window at a position specified by the user. This functions allows the user to instance the window at a specific position defined by the user, it clamps the window to the top left corner preventing it to go beyond. The clamp is tested on Windows and I'm quite positive the clamp can pose a real problem because different operating systems have different window edge sizes.
An image is attached below
InitWindowAt
:
Thank you


raysan5:
@tusharsingh09 sorry, no plans to add a new function, you can set Window position after Init.

PR #2318, opened 2022-01-29
Add missing defines to CUSTOMIZE_BUILD block

makuto:
This fixes compilation and linking errors when CUSTOMIZE_BUILD is
ON. Any option wasn't actually applied if it wasn't in this list.

I also added a message which makes it a bit easier to confirm you have
successfully enabled an option.


makuto:
I copied the options in the same order they are defined in `CMakeOptions.txt`, which is why the merge is a bit hard to read. Ideally, there would be a way to not require this fragile duplication, but I'm not well-versed enough in CMake to know if that's possible.

raysan5:
@makuto Nice! Thanks for the improvement!

raysan5:
@makuto Sorry, I click the wrong button and closed it without merging! :P 

Reopened and merged!

makuto:
Thanks for the quick response!

PR #2319, opened 2022-01-29
Fix Undefined Symbol `_ftelli64`

audinue:
Fix undefined symbol `_ftelli64` when trying to compile with Raylib's Notepad++ & TCC on Windows.


raysan5:
@audinue Thanks for the fix! Please note that this fix should be submitted to original [`cglft` repo](https://github.com/jkuhlmann/cgltf) or next time I update the library it will break again.

PR #2324, opened 2022-01-31
simplify QuaternionInvert

megagrump:
Just a small mistake that caught my eye.


raysan5:
@megagrump thanks for the improvement!

PR #2329, opened 2022-02-05
Make audio examples compile with external glfw

locriacyber:
raudio needs `dlclose`, which is in `libdl` on Linux. Before this commit the audio examples won't compile with `USE_EXTERNAL_GLFW:STRING=ON` in CMake.


raysan5:
@locriacyber thanks for the improvement!

PR #2340, opened 2022-02-11
Optimize `Vector2Rotate()` function

jdeokkim:
Since `sinf()` and `cosf()` are rather expensive functions compared to basic math operations, we try to avoid them as much as possible.


raysan5:
@jdeokkim Nice tweak! Thanks!

PR #2341, opened 2022-02-12
Properly fix make clean under windows

Peter0x44:
Unfortunately, my previous attempt was broken and didn't work properly
I had to split up make clean into multiple targets by OS, and make sure the Windows one always ran under cmd.exe, so it wouldn't break cross-compilation using the Makefile.

I have tested this as thoroughly as I could, trying compilation on Windows with, and without sh.exe in the PATH, compiling on Linux, and cross compiling from Linux to Windows.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -745,7 +745,7 @@ clean_WINDOWS: SHELL := cmd.exe
 
 clean_WINDOWS:
 	del *.o /s
-	cd $(RAYLIB_RELEASE_PATH)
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 I'd prefer to keep a single `clean` and discretize with #if statements per platform, it's done that way for other targets

raysan5:
Reviewed on commit: https://github.com/raysan5/raylib/commit/ab47c6401e3a7da22a2e9e1f4a2a63144d51e13a

Issue #1121, opened 2020-03-06
[textures] Color faded issue with RenderTexture


gezzuzz:
I draw 2 circles... one with TextureMode one without.. i expected these to give me the same result. 
But when i use Fade they act differently. What am i not understanding here?

Thank you in advance and thank you for creating raylib. 
I hope when i get more experience with it I will because to contribute to the project.

<img width="312" alt="circles" src="https://user-images.githubusercontent.com/46981052/76125209-f14e7880-5fc1-11ea-8424-73eaeb687fe7.png">

```
            Vector2 origin = (Vector2){500,500};
            DrawCircleV(origin,100,BLACK);
            for(int i=80;i<=100;i+=5)
                DrawCircleV(origin,i,Fade(RED,.3));
            
            RenderTexture2D temp = LoadRenderTexture(200, 200);
            BeginTextureMode(temp);
            origin = (Vector2){100,100};
            DrawCircleV(origin,100,BLACK);
            for(int i=80;i<=100;i+=5)
                DrawCircleV(origin,i,Fade(RED,.3));
            EndTextureMode();
            DrawTexture(temp.texture,620,400,WHITE);
```

raysan5:
@gezzuzz Good catch! :D Actually this issue is a tricky one. 

Short answer: main framebuffer is RGB by default while `RenderTexture` framebuffer is RBGA by default. This conditions how alpha is blended with previous colors in the framebuffer.

Related function:
```c
[textures.c]
// Load texture for rendering (framebuffer)
// NOTE: Render texture is loaded by default with RGBA color attachment and depth RenderBuffer
RenderTexture2D LoadRenderTexture(int width, int height)
{
    RenderTexture2D target = rlLoadRenderTexture(width, height, UNCOMPRESSED_R8G8B8A8, 24, false);

    return target;
}
```
You can just change `UNCOMPRESSED_R8G8B8A8` to `UNCOMPRESSED_R8G8B8`. It will work for your case but probably you will have problems for other usage cases... it's a tricky issue.

PR #2342, opened 2022-02-13
add premultiplied alpha blend mode

### LINKED ISSUE #2227: [rlgl] Alpha blending not working as expected for RenderTexture

### Issue description

The current alpha blending is broken, while the RGB channels are fine, the alpha channel results are incorrect.
Here's how it's implemented now:
`case RL_BLEND_ALPHA: glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA); glBlendEquation(GL_FUNC_ADD); break;`

For example, if you have an src alpha 0.5 and dest alpha 1.0, the resulting alpha will be equal to this:
`0.5 * 0.5 + 1.0 * (1.0 - 0.5) = 0.75`
Which is obviously incorrect.

I think proper blending should be implemented like this:
`glBlendFuncSeparate(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA, GL_ONE, GL_ONE_MINUS_SRC_ALPHA)`

### Environment

Windows 8.1 64x
ATI Technologies Inc.
AMD Radeon HD 8330
GLSL 4.60

### Issue Screenshot

I tried to draw a semi-transparent texture on RenderTexture2D cleaned with #463136FF and then draw this RenderTexture2D on screen cleaned with RAYWHITE.
image
I exported my RenderTexture2D to a .png image and those white pixels were transparent.
image

megagrump:
This adds a new blendmode `BLEND_ALPHA_PREMUL`, which is required for correct alpha blending of `RenderTexture`s. 

The same can be achieved in a more complicated way with `BLEND_CUSTOM` and `rlSetBlendFactors`, but that requires a bunch of OpenGL constants which are not readily available in every binding (e. g., Lua).

Fixes #2227, #1121.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -839,6 +839,7 @@ typedef enum {
     BLEND_MULTIPLIED,               // Blend textures multiplying colors
     BLEND_ADD_COLORS,               // Blend textures adding colors (alternative)
     BLEND_SUBTRACT_COLORS,          // Blend textures subtracting colors (alternative)
+    BLEND_ALPHA_PREMUL,             // Blend premultiplied textures considering alpha
     BLEND_CUSTOM                    // Blend textures using custom src/dst factors (use rlSetBlendMode())
 } BlendMode;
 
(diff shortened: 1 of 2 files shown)


raysan5:
@megagrump Ok, thanks for the addition of this new blend mode.

PR #2348, opened 2022-02-19
make const

planetis-m:


planetis-m:
Others that could also be const is UpdateMeshBuffer, rlUpdateVertexBuffer

raysan5:
There are several functions that input parameters can be `const` but I'm only using it for `const char *`. I prefer to keep it that way.

raysan5:
@planetis-m I reviewed this issue more carefully and I'm adding some additional consts.

planetis-m:
So you annotated all const pointers? Thats awesome, hopefully it would help bindings devs figure out if it's an out parameter without looking at the code.

raysan5:
@planetis-m Not all but most of them.

PR #2352, opened 2022-02-22
Update Swift binding version

conifer-dev:
Updated Swift version from 3.7 to 4.0


raysan5:
@conifer-dev thanks for the update! :D

Issue #2361, opened 2022-02-27
[core] `InitWindow()` causes a segmentation fault on android


TheDarkBug:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description
Segfault on android caused by `InitWindow()`

### Environment
Linux 5.14.14-arch1-1, building for Android 12 (linux 4.9.274), sdk 32, ndk 32 r23b, ABI: arm64-v8a.
OpenGL version: 3.2
GPU: Qualcomm Adreno 616

### Reproduce the issue
Build an android app with raylib compiled from the latest master branch commit and run it. It will segfault.

PR #2362, opened 2022-02-27
Fix for #2361

TheDarkBug:
Solves the #2361 issue


raysan5:
Just removed that memset()

PR #2365, opened 2022-02-28
Increase font atlas dimensions guesstimate

### LINKED ISSUE #2356: [rtext] Atlas size too small for font
I have a font ([Orbitron Medium](https://fonts.google.com/specimen/Orbitron?category=Sans+Serif&sidebar.open=true&selection.family=Raleway#standard-styles) at size 64) where the calculated atlas dimensions are to small for all default glyphs. 89/95 glyphs get generated before the atlas is full.

`GenImageFontAtlas` stops making rectangles when it runs out of room:
https://github.com/raysan5/raylib/blob/296b57e1f55f27b5d1f71a5cc8d623a2fbbde20b/src/rtext.c#L733
This leaves a bunch of rectangles at the end of the list uninitialized with garbage coords, which can (and often does) result in a crash in `ImageFromImage`:
https://github.com/raysan5/raylib/blob/296b57e1f55f27b5d1f71a5cc8d623a2fbbde20b/src/rtextures.c#L877-L886

The code to reproduce the problem is basically just this:
```C
Font font = LoadFontEx("Orbitron-Medium.ttf", 64, 0, 0)
```

megagrump:
Increase atlas size guesstimate, print warnings if atlas size is too small, and make sure all rectangles are properly initialized.

Fixes #2356.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -683,7 +683,7 @@ Image GenImageFontAtlas(const GlyphInfo *chars, Rectangle **charRecs, int glyphC
     // so image size would result bigger than default font type
     float requiredArea = 0;
     for (int i = 0; i < glyphCount; i++) requiredArea += ((chars[i].image.width + 2*padding)*(chars[i].image.height + 2*padding));
-    float guessSize = sqrtf(requiredArea)*1.3f;
+    float guessSize = sqrtf(requiredArea)*1.4f;
     int imageSize = (int)powf(2, ceilf(logf((float)guessSize)/logf(2)));  // Calculate next POT
 
     atlas.width = imageSize;   // Atlas bitmap width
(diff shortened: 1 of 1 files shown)


raysan5:
@megagrump Thanks for the review! I'm merging this improvement but actually, afair, that value was decreased some time ago because using 1.4f generated very big texture atlas with empty space. Probably a best estimation mechanism should be designed.

megagrump:
Yeah, in my specific case it couldn't fit the last 5 glyphs in an estimated 512x512 atlas. After the change, a 1024x1024 atlas is used, with lots of empty space. 1024x512 would have been sufficient here.

A simple changle to improve packing efficiency would be sorting the glyphs by height beforehand.

I'll come back to this when I find the time and see if I can improve atlas generation.

PR #2367, opened 2022-02-28
Add repology badge

jubalh:
So we can see which distributions package raylib.
And in which version they are.


raysan5:
@jubalh Thanks for the improvement, merging.

PR #2374, opened 2022-03-07
Update raylib-config.cmake

marcogmaia:
Updated cmake to windows users.
So they can `find_package(raylib [...])` with exported target as raylib instead of ${raylib_LIBRARIES}, etc


raysan5:
@marcogmaia I'm not an expert on Cmake, I assume this small change does not break anything...

PR #2375, opened 2022-03-08
Update parser

IrishBruse:
So I was parsing the raylib_api.json file and I found that var args are not parsed correctly
```c
RLAPI void TraceLog(int logLevel, const char *text, ...);         // Show trace log messages (LOG_DEBUG, LOG_INFO, LOG_WARNING, LOG_ERROR...)
```
```json
    {
      "name": "TraceLog",
      "description": "Show trace log messages (LOG_DEBUG, LOG_INFO, LOG_WARNING, LOG_ERROR...)",
      "returnType": "void",
      "params": {
        "logLevel": "int",
        "text": "const char *",
        "": "" <- ... here was really confusing to figure out what this was
      }
    },
```

The fix ive added will give a type and name to variable len args
```json
    {
      "name": "TraceLog",
      "description": "Show trace log messages (LOG_DEBUG, LOG_INFO, LOG_WARNING, LOG_ERROR...)",
      "returnType": "void",
      "params": [
        {
          "type": "int",
          "name": "logLevel"
        },
        {
          "type": "const char *",
          "name": "text"
        },
        {
          "type": "...",
          "name": "args"
        }
      ]
    },
```

I also added a format type of ALL so that its easier to refresh the bindings with each new release :)
`parser.exe -f ALL`

```
Output file:      raylib_api.txt
Output file:      raylib_api.lua
Output file:      raylib_api.json
Output file:      raylib_api.xml
```
is printed when running the command now

### DIFF
--- a/parser/raylib_parser.c
+++ b/parser/raylib_parser.c
@@ -820,6 +820,12 @@ static void GetDataTypeAndName(const char *typeName, int typeNameLen, char *type
             MemoryCopy(name, typeName + k + 1, typeNameLen - k - 1);
             break;
         }
+        else if (typeName[k] == '.' && typeNameLen == 3) // Handle varargs ...);
+        {
+            MemoryCopy(type, "...", 3);
+            MemoryCopy(name, "args", 4);
+	        break;
+        }
     }
 }
 
--- a/parser/raylib_api.json
+++ b/parser/raylib_api.json
@@ -2349,9 +2349,14 @@
           "description": "Blend textures subtracting colors (alternative)"
         },
         {
-          "name": "BLEND_CUSTOM",
+          "name": "BLEND_ALPHA_PREMUL",
           "value": 5,
-          "description": "Belnd textures using custom src/dst factors (use rlSetBlendMode())"
+          "description": "Blend premultiplied textures considering alpha"
+        },
+        {
+          "name": "BLEND_CUSTOM",
+          "value": 6,
+          "description": "Blend textures using custom src/dst factors (use rlSetBlendMode())"
         }
       ]
     },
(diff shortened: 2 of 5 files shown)


Peter0x44:
quite pedantic, but if a function like the following:
```c
struct a {};
void something(a b);
```
were ever to be added to the raylib API, then the parser would think that is a varargs
I think it would be better to check if it contains a ., because that cannot be a valid C identifier

IrishBruse:
I think that should fix all the feedback you gave me in the discord thanks for it im not very experienced writing C code I mostly know c#. I have been writting a code generator here https://github.com/IrishBruse/Raylib-CsLo/tree/safe-codegen
and ran into an issue with the json file said above also having all the files in the git is handy as we submodule this repo and use the json to generate our bindings.

parser/raylib_parser.c
```diff
@@ -610,17 +610,47 @@ int main(int argc, char* argv[])
...
-    else if (outputFormat == LUA) printf("\nOutput format:    LUA\n\n");
 
-    ExportParsedData(outFileName, outputFormat);
+    if (outputFormat == ALL)
+    {
+        // Process input file to output
+        if (outFileName[0] == '\0') MemoryCopy(outFileName, "raylib_api\0", 11);
+
+        char outFileNameAll[512];
+        sprintf_s(outFileNameAll, 512, "txt/%s%s", outFileName, ".txt");
```
raysan5, line 623: I'm afraid raylib-parser avoids usage of string functions.

raysan5:
@IrishBruse I prefer to avoid adding all the generated files, just let the users generate the ones they desire.

parser/raylib_parser.c
raysan5: Accessing the individual chars and changing them as required.

parser/raylib_parser.c
```diff
@@ -124,7 +124,7 @@ typedef struct DefineInfo {
 } DefineInfo;
 
 // Output format for parsed data
-typedef enum { DEFAULT = 0, JSON, XML, LUA } OutputFormat;
+typedef enum { DEFAULT = 0, JSON, XML, LUA, ALL } OutputFormat;
```
raysan5, line 127: Why adding `ALL`? Usually users will only require one output format. Please, could you remove the `ALL` option?

parser/raylib_parser.c
```diff
@@ -654,7 +684,7 @@ static void ShowCommandLineInfo(void)
     printf("                                      Supported extensions: .txt, .json, .xml, .h\n");
     printf("                                      NOTE: If not specified, defaults to: raylib_api.txt\n\n");
     printf("    -f, --format <type>             : Define output format for parser data.\n");
-    printf("                                      Supported types: DEFAULT, JSON, XML, LUA\n\n");
+    printf("                                      Supported types: DEFAULT, JSON, XML, LUA, ALL\n\n");
```
raysan5, line 687: Please, could you remove the `ALL` option?

parser/raylib_parser.c
raysan5: Also note that `sprintf_s()` is a Microsoft specific function, not available on most C compilers.

IrishBruse:
Ok removed the all flag and cleaned up the commits

raysan5:
@IrishBruse Thanks for the review, merging!

PR #2376, opened 2022-03-09
Vector2/Vector3: Add squared distance.

AnilBK:
Needed to sort polygons with distance to camera. In such cases, where we just do comparisons, squared distance seems good.
Might be useful as a general API.

Not sure if this is considered bloat, but as we can see vector2/3 has length and squared length functions.
https://github.com/raysan5/raylib/blob/43dbe67431dac3e44c514cf1839f49cae69777b6/src/raymath.h#L249-L263

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -278,6 +278,14 @@ RMAPI float Vector2Distance(Vector2 v1, Vector2 v2)
     return result;
 }
 
+// Calculate square distance between two vectors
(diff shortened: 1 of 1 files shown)


raysan5:
@AnilBK I consider it a bit bloat but I'm adding it...

AnilBK:
Thanks. <3 
Will surely contribute to no bloat stuffs, if I can :D

Issue #2379, opened 2022-03-10
[core] `IsKeyPressed()` does not work in the `custom frame_control` example on `PLATFORM_WEB`


thismarvin:
### Issue description

The `IsKeyPressed` function does not work in the "custom frame_control" example on [raylib.com/examples](https://www.raylib.com/examples/core/loader.html?name=core_custom_frame_control). The circle's movement cannot be paused and the target fps cannot be modified. 

The example does work on Desktop; however, it appears that because of some quirk with emscripten, the example does not work on the Web. Adapting the code to use `emscripten_set_main_loop` also does not work.

### Environment

Web, Linux, 3.3.0 NVIDIA 510.54, NVIDIA GeForce GTX 1060

I would like to add that I have tested this on Firefox 98 and Microsoft Edge 99 beta

### Code Example

Make sure to build raylib with the `SUPPORT_CUSTOM_FRAME_CONTROL` flag enabled and to target `PLATFORM_WEB`.

``` c
#include "raylib.h"

int main(void)
{
    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "raylib [core] example - custom frame control");

    double previousTime = GetTime();
    double currentTime = 0.0;
    double updateDrawTime = 0.0;
    double waitTime = 0.0;

    int targetFPS = 60;

    while (!WindowShouldClose())
    {
        PollInputEvents();

        if (IsKeyPressed(KEY_SPACE)) TraceLog(LOG_INFO, "Here!");

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();

        SwapScreenBuffer();

        currentTime = GetTime();
        updateDrawTime = currentTime - previousTime;

        if (targetFPS > 0)
        {
            waitTime = (1.0f/(float)targetFPS) - updateDrawTime;
            if (waitTime > 0.0)
            {
                WaitTime((float)waitTime*1000.0f);
                currentTime = GetTime();
            }
        }

        previousTime = currentTime;
    }

    CloseWindow();

    return 0;
}
```

thismarvin:
I am not sure if this is actually the best solution, but if anyone every runs into this in the future then just remember to call `PollInputEvents` after `IsKeyPressed` or else `IsKeyPressed` will always be false! 

You can read more about this in the comments of the closed pull request above.

raysan5:
[...]

PR #2380, opened 2022-03-10
Fixed "custom frame_control" example for PLATFORM_WEB (#2380)

thismarvin:
This PR addresses Issue https://github.com/raysan5/raylib/issues/2379.

After a lot of head-scratching I realized that `PollInputEvents` must always be called as late as possible! When `SUPPORT_CUSTOM_FRAME_CONTROL` is disabled, `EndDrawing` already calls `PollInputEvents` last.

It seems that the following pull request https://github.com/raysan5/raylib/pull/1573 was on to something.

I do not fully understand why this is a problem for `PLATFORM_WEB` and not `PLATFORM_DESKTOP`. I did spend enough time looking at `rcore.h` to conclude that this issue is a quirk of using emscripten, and not something inherently wrong with `rcore.h`'s logic. 

EDIT: I was wrong; the above mentioned pull request does not apply to this issue.

### DIFF
--- a/examples/core/core_custom_frame_control.c
+++ b/examples/core/core_custom_frame_control.c
@@ -54,8 +54,6 @@ int main(void)
     {
         // Update
         //----------------------------------------------------------------------------------
-        PollInputEvents();              // Poll input events (SUPPORT_CUSTOM_FRAME_CONTROL)
-        
         if (IsKeyPressed(KEY_SPACE)) pause = !pause;
         
         if (IsKeyPressed(KEY_UP)) targetFPS += 20;
@@ -113,6 +111,8 @@ int main(void)
         else deltaTime = updateDrawTime;    // Framerate could be variable
 
         previousTime = currentTime;
+
+        PollInputEvents();          // Poll input events (SUPPORT_CUSTOM_FRAME_CONTROL)
         //----------------------------------------------------------------------------------
     }
 


raysan5:
@thismarvin I'm afraid I'm not merging this PR. 

Actually, this example is not intended for `PLATFORM_WEB`. `WindowShouldClose()` introduces a sleep of 16ms when compiled with `ASYNCIFY` for `PLATFORM_WEB` so `PollInputEvents()` shouldn't be moved before `WindowShouldCLose()`, it will introduce a delay of 16ms on inputs... that's the logic but then `ASYNCIFY` does it magic and could even defer the input reading (maybe that's the real issue).

In any case, moving `PollInputEvents()` before `WindowShouldClose()` is not the proper solution... neither using `SUPPORT_CUSTOM_FRAME_CONTROL` with `ASYNCIFY`, results could vary between browsers.

thismarvin:
> EDIT 1: This is probably the issue: https://github.com/emscripten-core/emscripten/blob/main/src/library_glfw.js#L20

The fact that `glfwPollEvents` literally does [nothing](https://github.com/emscripten-core/emscripten/blob/0834fb6bf296fda861b125ab853705d2e9c86d95/src/library_glfw.js#L1184) definitely throws a wrench into all of this...

>  I don't know why but moving the function to the end is not the solution, or at least not a solution originated from an understanding of the problem.

I definitely do not fully understand the problem at hand; however, the official glfw [documentation](https://www.glfw.org/docs/latest/input_guide.html#events) mentions the following:

> Event processing is normally done each frame after [buffer swapping](https://www.glfw.org/docs/latest/window_guide.html#buffer_swap).

The above quote seems to justify the changes I made; however, as you mentioned, also including `WindowShouldClose()` causes a delay for `PLATFORM_WEB`. With that said, should I adapt this particular example to use `emscripten_set_main_loop` in addition to my fix?

raysan5:
> Event processing is normally done each frame after [buffer swapping](https://www.glfw.org/docs/latest/window_guide.html#buffer_swap).

Yes, and actually, it's done after ` SwapScreenBuffer()` in current implementation because we are inside a loop. 

 > should I adapt this particular example to use `emscripten_set_main_loop` in addition to my fix?

No, I don't think so. Now that I understand better the problem I agree with your initial comment that this example should work in web as is, like all the other examples.

The real issue should be investigated and fixed.

PR #2383, opened 2022-03-11
Simplify build.zig to not require user to specify raylib path

Hejsil:
We can figure out the source file location based on the location of the
build.zig file. No need to require the library user to specify where
raylib is stored.

```zig
// Before
const rl = @import("raylib/src/build.zig").Pkg("raylib/src");
const exe = b.addExecutable(name, path);
exe.linkLibrary(rl.addRaylib(exe.builder, exe.target));

// After
const rl = @import("raylib/src/build.zig");
const exe = b.addExecutable(name, path);
exe.linkLibrary(rl.addRaylib(exe.builder, exe.target));
```

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,89 +1,81 @@
 const std = @import("std");
 
-pub fn Pkg(srcdir: []const u8) type {
-    return struct {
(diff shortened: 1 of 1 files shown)


raysan5:
@Hejsil Thanks for the improvement!

PR #2384, opened 2022-03-12
Fixed an issue in Makefile when using raygui and physac on unix systems

tixvage:
Hash "#" is single-line comment character in bash so echo ignores all of line if #include and #define written and make gives an error because it cannot find raygui.c and physac.c

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -664,11 +664,11 @@ raygui.o : raygui.c
 	$(CC) -c $< $(CFLAGS) $(INCLUDE_PATHS)
 raygui.c:
 ifeq ($(PLATFORM_SHELL), cmd)
-		@echo #define RAYGUI_IMPLEMENTATION > raygui.c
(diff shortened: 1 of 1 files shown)


src/Makefile
```diff
@@ -663,17 +663,26 @@ raudio.o : raudio.c raylib.h
 raygui.o : raygui.c
 	$(CC) -c $< $(CFLAGS) $(INCLUDE_PATHS)
 raygui.c:
-	echo #define RAYGUI_IMPLEMENTATION > raygui.c
-	echo #include "$(RAYLIB_MODULE_RAYGUI_PATH)/raygui.h" >> raygui.c
+ifeq ($(PLATFORM_SHELL), cmd)
+		@echo #define RAYGUI_IMPLEMENTATION > raygui.c
```
raysan5, line 667: Please, could you remove the double TAB? I think one TAB should be enough.

raysan5:
@tixvage Thanks for the review! :D

PR #2385, opened 2022-03-12
Updated Zig bindings to reflect updates to that library

guidoism:


raysan5:
@guidoism Thanks for the update!

PR #2386, opened 2022-03-13
Add naylib

planetis-m:
This commit adds naylib https://github.com/planetis-m/naylib


raysan5:
@planetis-m thanks for the addition! 😄

PR #2391, opened 2022-03-14
Improve joystick visualisation in gamepad example again

kristianlm:
Sorry for the spam, you were too fast! I wanted to force-push this into my previous PR, but you beat me to it.

As prior commit, but complete fix for XBox controller and trigger
buttons too.

> The joystick range is a float from -1 through +1. Casting this to int
> yields only three possible values: -1, 0, and 1. This gives a
> misleading joystick placement in the demo.
>
> By casting to int after the multiplication to get pixel values, the
> demo reveals the analog feel of the joystick.


raysan5:
@kristianlm hahaha... no worries! Merged! Thanks again!

PR #2396, opened 2022-03-17
OpenURL() for Android Platform

futureapricot:
This PR implements the OpenURL() function for Android Platform.
It will call an Intent to open the given url with the default browser app.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -242,6 +242,7 @@
     //#include <android/sensor.h>           // Required for: Android sensors functions (accelerometer, gyroscope, light...)
     #include <android/window.h>             // Required for: AWINDOW_FLAG_FULLSCREEN definition and others
     #include <android_native_app_glue.h>    // Required for: android_app struct and activity management
+    #include <jni.h>                        // Required for: JNIEnv and JavaVM [Used in OpenURL()]
 
     #include <EGL/egl.h>                    // Native platform windowing system interface
     //#include <GLES2/gl2.h>                // OpenGL ES 2.0 library (not required in this module, only in rlgl)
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -3457,6 +3457,29 @@ void OpenURL(const char *url)
 #endif
 #if defined(PLATFORM_WEB)
         emscripten_run_script(TextFormat("window.open('%s', '_blank')", url));
+#endif
+#if defined(PLATFORM_ANDROID)
+        JNIEnv *env;
```
raysan5, line 3462: Please, could you init it to NULL?

src/rcore.c
```diff
@@ -3457,6 +3457,29 @@ void OpenURL(const char *url)
...
+        JavaVM *vm = CORE.Android.app->activity->vm;
```
raysan5, line 3463: As per [the documentation](https://developer.android.com/training/articles/perf-jni#javavm-and-jnienv), `JNIEnv` and `JavaVM` require `jni.h`, shouldn't the include be added?

src/rcore.c
```diff
@@ -3457,6 +3457,29 @@ void OpenURL(const char *url)
...
+        (*vm)->AttachCurrentThread(vm, &env, NULL);
+
+        jstring url_string = (*env)->NewStringUTF(env, url);
```
raysan5, line 3466: Please, could you follow [raylib naming conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md) when naming variables on raylib code? It should be `urlString`. Same for the following code.

src/rcore.c
futureapricot: You're right! somehow I missed it!

futureapricot:
I have made the mentioned corrections

raysan5:
@futureapricot Thanks for the improvement and the review

PR #2401, opened 2022-03-19
[Textures] Make DrawTextureNPatch check the batch limits

JeffM2501:
This PR adds a check to the render batch to verify there is enough room for the geometry that may be created. Without this the batch can overflow.


raysan5:
@JeffM2501 Thanks for the improvement!

PR #2405, opened 2022-03-23
[CORE] Add a matrix to GetRayCollisionModel

JeffM2501:
This PR adds a matrix to GetCollisionModel so that you don't have to modify the model every time you want collisions. This allows the same unmodified model for drawing and collision.

### DIFF
--- a/examples/models/models_mesh_picking.c
+++ b/examples/models/models_mesh_picking.c
@@ -123,7 +123,7 @@ int main(void)
 
             // Check ray collision against model
             // NOTE: It considers model.transform matrix!
-            RayCollision meshHitInfo = GetRayCollisionModel(ray, tower);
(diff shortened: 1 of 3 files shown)


raysan5:
@JeffM2501 I can't see the benefits of providing an external transform matrix, `Model` already defines and internal `model.transform` Matrix just for that. I also find think it can be confusing for the user, in case model already had an internal transform.

raysan5:
After long thinking about this function I decided the better option is just removing it and let the manage model collisions directly with `GetRayCollisionMesh()`, it will also provide full control over the colliding mesh, `GetRayCollisionModel()` does not provide that control and I feel adding and extra parameter makes the function more confusing.

`DrawModel*()` functions receive simple transform parameters (translate, rotate, scale) as inputs, they are decomposed for beginners to avoid the Matrix transform. That matrix is also provided inside the `Model` for advance users that know what they do.

PR #2422, opened 2022-03-31
[CORE] Allow for setting a array of icons instead of only one

SplatTab:
Allow for setting a array of icons (most likely different resolutions) instead of just one icon


raysan5:
@SplatTab Sorry, I'm not changing that function, others users could be using it and probably most users don't need to set more than one icon. If you need that advance functionality I recommend calling GLFW directly.

SplatTab:
Yea but you could still set only one icon and it would be annoying to have to add GLFW for others

SplatTab:
I guess that's fair I'll just patch it so that I can set my second icon

raysan5:
@SplatTab Why do you need to set your icons in runtime? If they are static they are usually set as an icon resource at compilation time.

PR #2426, opened 2022-04-04
Fix string defines generating invalid XML

eutro:
The parser was previously generating invalid XML.


raysan5:
@eutro Tahnks for the review!

PR #2427, opened 2022-04-05
Fix OpenGL 4.3 graphics option in CMake

GoldenThumbs:
Simple fix to a CMake script that will allow it to set OpenGL version to 4.3.


raysan5:
@GoldenThumbs thanks for the fix!

PR #2428, opened 2022-04-05
Augment raymath.h with useful functions

saccharineboi:
I use these functions heavily during gamedev so I thought they could be useful for others too:

- FloatEquals:  Check whether two given floats are almost equal
- Vector{n}Invert: Invert the given vector
- Vector{n}Clamp: Clamp the components of the vector between min and max values specified by the given vectors
- Vector{n}ClampMagnitude: Clamp the magnitude of the vector between two given min and max values
- Vector{n}Equals:  Check whether two given vectors are almost equal

Additionally Vector3{Up,Down,Left,Right,Forward,Backward} are added for convenience.

Vector3Refract refracts a vector given a normal and a refractive index.

QuaternionEquals checks whether two given quaternions are almost equal.

Furthermore, a macro named "EPSILON" with a value 1e-6 is added (used in floating-point comparison functions), and stdbool.h is included for functions that return boolean values.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -449,7 +449,7 @@ RMAPI Vector2 Vector2Clamp(Vector2 v, Vector2 min, Vector2 max)
 
 // Clamp the magnitude of the vector between two
 // given min and max values
-RMAPI Vector2 Vector2ClampMagnitude(Vector2 v, float minMag, float maxMag)
+RMAPI Vector2 Vector2ClampValue(Vector2 v, float minMag, float maxMag)
 {
     Vector2 result = { 0 };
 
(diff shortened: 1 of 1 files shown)


CrezyDud:
rename Vector2ClampMagnitude and Vector3ClampMagnitude to Vector2ClampValue and Vector3ClampValue
for consistency

raysan5:
@saccharineboi I think some of those functions are unneded and just add a level of undirection not required. i.e users cand define required Vector3 up/down/back...  directly, no need to provide functions for everything.

raysan5:
@saccharineboi thanks for the review.

planetis-m:
I don't know if it's intentional or an oversight but missing from this PR is MatrixEquals.

raysan5:
@planetis-m Could you list a use case for that function?

saccharineboi:
> I don't know if it's intentional or an oversight but missing from this PR is MatrixEquals.


Intentional. In contrast to Vector{n}Equals which is useful for checking degenerate vectors, MatrixEquals is almost never used. Though it could be added for completeness' sake.


@raysan5 what do you mean it broke raylib? Was it because of stdbool.h? Instead of removing those functions, you could change return types to int

raysan5:
@saccharineboi I already change return types to int.

PR #2429, opened 2022-04-06
Optimize Image Draw Functions.

AnilBK:
Previously, for every pixels, `ImageDrawPixel()` was called which checks for image format and performs calculations on the given pixel coordinate.

https://github.com/raysan5/raylib/blob/07240b96ad7c689119e70714f3fb39852b3657b7/src/rtextures.c#L2419-L2432

The functions optimized here just copies same data over and over, so the calculation doesn't need to be performed more than once.

This optimization is already used somewhere else.
https://github.com/raysan5/raylib/blob/07240b96ad7c689119e70714f3fb39852b3657b7/src/rtextures.c#L2797

I was thinking of a general function sth like this.
`void _repeat_pixels_in_memory(unsigned char *src, int count, int bytesPerPixel)
{
    for(int i = 1; i < count; i++)
    {
        memcpy(src + i * bytesPerPixel, src, bytesPerPixel);
    }
}
`
But not many use cases, so didn't add that.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -1889,10 +1889,10 @@ void ImageColorTint(Image *image, Color color)
             unsigned char b = (unsigned char)(((float)pixels[index].b/255*cB)*255.0f);
             unsigned char a = (unsigned char)(((float)pixels[index].a/255*cA)*255.0f);
 
-            pixels[y*image->width + x].r = r;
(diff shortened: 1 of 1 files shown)


raysan5:
@AnilBK thanks! this is a nice improvement! :)

Issue #2371, opened 2022-03-04
[core] Scroll wheel direction incorrect on macOS Monterey 


joebentley:
### Issue description

`GetMouseWheelMove()` seems to detect only horizontal scrolling of the mac trackpad (two finger swipe left and right), rather than the vertical scrolling. Related to https://github.com/raysan5/raylib/commit/7c1889c28205b83afaa39da73f1831a72bd9c502?

### Environment

MacBook Pro 13-inch M1 2020

macOS Monterey

### Code Example

```c
#include "raylib.h"

int main(void)
{
    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "MouseWheelIssue");

    SetTargetFPS(60);
    
    float y = 0;
    
    while (!WindowShouldClose())
    {
        y += GetMouseWheelMove();
        
        BeginDrawing();
            ClearBackground(RAYWHITE);
            DrawCircle(screenWidth/2, y + screenHeight/2, 20, GREEN);
        EndDrawing();
    }
    
    CloseWindow();
    return 0;
}
```

raysan5:
@joebentley This issue could be tricky, it should detect both but only return the current one. I would need to know if it works for other macOS users.

PR #2437, opened 2022-04-24
Fix issue #2371

jcgamestoy:
In MouseScrollCallback priority was given to the horizontal movement, I have changed it to return the largest movement (in absolute value) between the X and Y axis.

Perhaps it would be more interesting to have different functions for the X and Y axis.


raysan5:
@jcgamestoy hey! nice solution! thanks for the fix! 👍😄

Issue #2441, opened 2022-04-26
[rmodels] small memory leak while loading GLTF models


leomonta:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Raylib internal function LoadGLTF leaks 28 bytes (7 unsigned int) while loading the model
The issue is at line n. 4646 of rmodels.c in the function LoadGLTF(const char* filename)

<pre>
// Load our model data: meshes and materials
model.meshCount = primitivesCount;
for (int i = 0; i < model.meshCount; i++) model.meshes[i].vboId = (unsigned int*)RL_CALLOC(MAX_MESH_VERTEX_BUFFERS, sizeof(unsigned int));

</pre>

This allocates preemptively the vboId array (7 uint, 28 Bytes)

This array is also allocated in LoadModel(), by UploadMesh(), right after LoadGLTF() gets called
UploadMesh should be the only function that allocate the vboId thus i propose to remove line 4646 in rmodels.c since LoadGLTF() doesn't actually uses the vboId

raysan5:
Thanks for the fix!

PR #2442, opened 2022-04-26
Fix Issue #2441

leomonta:
Remove line that caused a memory leak


raysan5:
@leomonta Oh! Good catch! Thanks for the fix!

PR #2446, opened 2022-04-27
use GetWindowScaleDPI to calculate size for rlReadScreenPixels in sceenshot/recording

gulrak:
On macOS when running on a display with retina resolution or on an external 4k screen, the `TakeScreenshot` function triggered by F12 and the gif recording triggered by Ctrl-F12 ignore the scaling so they only grab the lower left quarter of the window.

This change uses the factors from `GetWindowScaleDPI()` to scale width and height to allow the full window content to be grabbed even if scaling is active.

Not sure if there would be need for a check of `FLAG_WINDOW_HIGHDPI`, from my understanding, the scaling would be needed there to, for other cases if should be a scaling by 1, and the overhead of `GetWindowScaleDPI()` seems small compared to the gif or png encoding that happens afterwards.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2031,8 +2031,9 @@ void EndDrawing(void)
         {
             // Get image data for the current frame (from backbuffer)
             // NOTE: This process is quite slow... :(
-            unsigned char *screenData = rlReadScreenPixels(CORE.Window.screen.width, CORE.Window.screen.height);
-            msf_gif_frame(&gifState, screenData, 10, 16, CORE.Window.screen.width*4);
+            Vector2 scale = GetWindowScaleDPI();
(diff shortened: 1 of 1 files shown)


raysan5:
@gulrak I'm afraid there have been many attempts to implement this same solution and everytime it broke some platform in some specific use case. I prefer to let the user manage it manually as required, that's the reason to expose `GetWindowScaleDPI()` for the user instead of applying it directly by raylib.

Still, I'm not closing the PR, I want to take a closer look to it.

gulrak:
Yeah, I understand the problem, the easy way to reduce that risk would be to tie it to the Plattform, still I'm planning to test it on Windows with and without high dpi flag and scaling and the same on Linux, just need to get some time for it, so no hurry.

raysan5:
@gulrak After cheking this improvement more carefully I decided to merge it.

D-T-666:
yay!

PR #2451, opened 2022-04-28
Update raylib haxe binding name

foreignsasquatch:
As title says


raysan5:
@foreignsasquatch Nice! Thanks for the update! 👍😄

PR #2455, opened 2022-05-01
Added thickness argument to DrawLine function.

ScriptLineStudios:
A thickness can be specified for the regular DrawLine function.


raysan5:
@ScriptLineStudios I'm afraid I'm not adding that extra parameter. Two reasons:

 1. I try to keep raylib simple. Most raylib users won't benefit of that extra parameter.
 2. It could break many games that use that function.

PR #2462, opened 2022-05-04
Reorder parser

lazaray:
This PR reorders the parsers output and code to be similar to raylib.h, with the following order:
 * Defines
 * Structs
 * Aliases
 * Enums
 * Callbacks
 * Functions

It also orders everything to be internally consistent.

### DIFF
--- a/parser/raylib_api.json
+++ b/parser/raylib_api.json
@@ -1,4 +1,318 @@
 {
+  "defines": [
+    {
+      "name": "RAYLIB_H",
+      "type": "GUARD",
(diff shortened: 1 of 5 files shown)


raysan5:
@lazaray Wow, that's a big PR... I'm just crossing-fingers and merging! ;)

lazaray:
Yeah, sorry :P I tried making smaller commits but moving stuff around makes for huge diffs.
Future PRs will be byte sized with pretty diffs :)

PR #2463, opened 2022-05-04
Add support for calculated defines to parser

lazaray:
This small PR adds support for calculated defines to the parser. The new define types with the `_MATH` suffix mean that a define is a mathematical expression that would have the preceding type when calculated.

raylib.h examples:
```json
    {
      "name": "DEG2RAD",
      "type": "FLOAT_MATH",
      "value": "(PI/180.0f)",
      "description": ""
    },
    {
      "name": "RAD2DEG",
      "type": "FLOAT_MATH",
      "value": "(180.0f/PI)",
      "description": ""
    },
```

### DIFF
--- a/parser/raylib_api.json
+++ b/parser/raylib_api.json
@@ -26,13 +26,13 @@
     },
     {
       "name": "DEG2RAD",
-      "type": "UNKNOWN",
+      "type": "FLOAT_MATH",
       "value": "(PI/180.0f)",
       "description": ""
     },
     {
       "name": "RAD2DEG",
-      "type": "UNKNOWN",
+      "type": "FLOAT_MATH",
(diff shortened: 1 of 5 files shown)


raysan5:
@lazaray thanks for the improvement... I'm merging it but I wonder how useful it is for users of the parser, I can't see the application of it at this moment and it adds about 140 lines of code complexity to the parser.

At this point you are very familiarized with the parser but keep in mind that if it requires some review in the future others users could find it difficult to maintain...

lazaray:
My thinking was that it would be useful to know the type of constants for bindings for strongly typed languages. But there aren't any calculated constants that would be useful to bind at the moment so this wasn't really necessary. Sorry about that.

PR #2464, opened 2022-05-05
Add support for truncating parser input

lazaray:
This PR adds support for truncating parser input lines with a command line arg. This can be used to not parse implementations in headers that contain them.

I noticed the implementations being parsed because two function implementations in rlgl.h had the RLAPI specifier which caused them to be duplicated in the parser output.


raysan5:
@lazaray Please note that `raylib-parser` is intended for raylib parsing only, if it works with other files, it's ok but that's not the pourpose of the tool. 

I'm seeing the parser is growing in size and complexity lately and despite most of the additions are great, I'm a bit worried about its maintenance. Maintenance requires being familiar with the code and it takes time...

lazaray:
Not parsing the implementation is useful for implementations that contain internal structs, enums and macros. raylib.h doesn't do this but the extras and especially rlgl.h do and using this improves the parser output to only contain the API.

I'll do my best to make maintenance easier and not add any unnecessary complexity.

PR #2465, opened 2022-05-05
Flush stdout after trace messages

nagy:
This immediately shows log messages when stdout is not connected to a
tty.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -148,6 +148,7 @@ void TraceLog(int logType, const char *text, ...)
     strcat(buffer, text);
     strcat(buffer, "\n");
     vprintf(buffer, args);
+    fflush(stdout);
 #endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@nagy Usually I try to minimize the required system functions calls, `fflush()` introduces a probably unnecessary performance hit... Still, I see the point of this commit and I will merge it, considering that `TraceLog()` calls are usually very low.

If a see problems in any of the raylib supported platforms I will remove it.

nagy:
Thank you for the merge. I just wanted to leave a note for the future, that the same effect can be achieved externally on a per-invocation basis with `stdbuf` like so:

```
$ stdbuf -oL -eL ./my-raylib-prog
```

https://linux.die.net/man/1/stdbuf

PR #2467, opened 2022-05-08
added: IsKeyRepeating(key) for PLATFORM_DESKTOP

ProPuke:
Currently `GetCharPressed()` responds to os key-repeat events, making it useful for natural typing, however non-visible keys like backspace, or cursor keys etc. do not, creating an awkward inconsistency.

This adds a `IsKeyRepeating(key)` function, allowing users to query if a key is being pressed and is currently repeating (on this frame).

I've only implemented this for glfw and I've probably overlooked some things (like the `GetKeyPressed()` queue).
If we wanted the `GetKeyPressed` queue to also include key repeats, then maybe there should be a way of marking _which_ keys should repeat in this queue. Then users could do specifically mark the text control keys as repeating, when performing text editing, and turn this feature off when done. But that sounds like more of a complicated consideration.

If this is to be a thing, we'd obvs need this for other platforms too. So I guess this PR is partly just a proof-of-concept suggestion.

### DIFF
--- a/projects/Notepad++/raylib_npp_parser/raylib_to_parse.h
+++ b/projects/Notepad++/raylib_npp_parser/raylib_to_parse.h
@@ -172,6 +172,7 @@ RLAPI void OpenURL(const char *url);                              // Open URL wi
 // Input-related functions: keyboard
 RLAPI bool IsKeyPressed(int key);                             // Check if a key has been pressed once
 RLAPI bool IsKeyDown(int key);                                // Check if a key is being pressed
+RLAPI bool IsKeyRepeating(int key);                           // Check if a key is being pressed and is currently repeating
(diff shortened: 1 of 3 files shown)


ProPuke:
So, as a simple, if you were inputting text and wanted to support the backspace key you would do
```c
if (IsKeyPressed(KEY_BACKSPACE) || IsKeyRepeating(KEY_BACKSPACE)) {
    // delete a char
    ...
}
```
`IsKeyPressed` would return true on just the initial key down, as it always has, and `IsKeyRepeating` would return true on each repeat interval after, for as long as it was held.

raysan5:
@ProPuke Thanks for the PR but I'm afraid I'm not adding this functionality. It seems intended for a very specific use case and it has not been tested/supported by all raylib platforms.

PR #2470, opened 2022-05-12
Moving guiIconName enum out of #ifdef RAYGUI_IMPLEMENTATION to fix compilation errors

Neolisk:
The guiIconName enum gets compiled out if RAYGUI_IMPLEMENTATION isn't defined, which makes code referencing/using the enum fail to compile. Pulling it out fixed raygui for me.


raysan5:
@Neolisk Actually this change is already implemented in raygui repo.

PR #2472, opened 2022-05-12
Fix crash with delay demo.

ptarabbia:
The audio effects system works great !
The issue with the delay effect was simply caused by an uninitialized variable.


raysan5:
@ptarabbia Nice! Thanks for the review! 👍😄

PR #2481, opened 2022-05-16
generate compile_commands.json to be used by language server

anggape:


argc1:
Im not expert on cmake, but some language server need json compilation database to make it works, so this is required if you not use ide like visual studio or clion

anggape:
hai @raysan5  like @argc1 mentions, both clangd and ccls languages server need this, like they say in the documentations here:
https://clangd.llvm.org/installation.html#compile_commandsjson

raysan5:
@anggape Thank you very much for the explanation, merged! :)

PR #2486, opened 2022-05-19
Fixed CMake's include of Android's native app glue

patm1987:
Changed the `set` to a `list(APPEND` operation

You can test by creating an Android Studio C++ project and adding raylib to the CMake file
```
set(PLATFORM Android CACHE INTERNAL "Set Android platform")
add_subdirectory(external/raylib)
```

...

```
target_link_libraries( # Specifies the target library.
        demogame
        ...
        raylib

        # Links the target library to the log library
        # included in the NDK.
        ${log-lib})
```


raysan5:
@patm1987 Thanks for the review!

Issue #1125, opened 2020-03-08
[core] SetExitKey() bug (with fix)


TakUnderhand:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch **(I checked master and it has the same bug)** [core.c line 4641](https://github.com/raysan5/raylib/blob/master/src/core.c)
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

SetExitKey() didn't work for me: I wanted to disable KEY_ESCAPE.
So I fixed it:
In core.c, function ProcessKeyboard(), the value 256 is used for checking for exit in one place, where it should use CORE.Input.Keyboard.exitKey.

Current code:
`if (bufferByteCount == 1) CORE.Input.Keyboard.currentKeyState[256] = 1;> //raylib key: KEY_ESCAPE`
Fixed code:
`if (bufferByteCount == 1) CORE.Input.Keyboard.currentKeyState[CORE.Input.Keyboard.exitKey] = 1; // default raylib key: KEY_ESCAPE`

### Environment

I'm cross developing on MacOS/Linux/Window (MSYS).  Although I only tested this bug fix on MacOS so far.

### Issue Screenshot

Not applicable

### Code Example

See above

raysan5:
@TakUnderhand thanks! fixed!

TakUnderhand:
You're welcome!
Thanks for the amazing API of raylib :)

PR #2498, opened 2022-05-30
Disable Exit Key Config Option

Gota7:
Make a config flag for the window (as to be used by SetConfigFlags) that allows the program to continue running even if the exit key is pressed.

Note: https://github.com/raysan5/raylib/issues/1125 has been implemented, but is setting it to something "unpressable" the cleanest way of disabling an exit key?


raysan5:
@Gota7 I'm afraid this change is not needed, just `SetExitKey(0)`

Issue #2434, opened 2022-04-15
[core] Video stuttering with `SUPPORT_PARTIALBUSY_WAIT_LOOP` on Linux


flashback-fx:
### Issue description

There seems to be a regression in the rcore.c WaitTime function that causes video stuttering when compiling with SUPPORT_PARTIALBUSY_WAIT_LOOP for PLATFORM_DESKTOP running under Linux (I am not sure if other platforms are affected). I have been testing different raylib version tags and this stuttering does not happen in 3.7.0, and happens in 4.0.0. The cause seems to be changes made to the WaitTime function of rcore.c (previously Wait in core.c).

This is the output of doing
`git diff 3.7.0 master -- core.c rcore.c`

```
-static void Wait(float ms)
+void WaitTime(float ms)
 {
-#if defined(PLATFORM_UWP)
-    UWPGetSleepFunc()(ms/1000);
-    return;
-#endif
-
 #if defined(SUPPORT_BUSY_WAIT_LOOP)
-    double prevTime = GetTime();
-    double nextTime = 0.0;
+    double previousTime = GetTime();
+    double currentTime = 0.0;

     // Busy wait loop
-    while ((nextTime - prevTime) < ms/1000.0f) nextTime = GetTime();
+    while ((currentTime - previousTime) < ms/1000.0f) currentTime = GetTime();
 #else
-    #if defined(SUPPORT_HALFBUSY_WAIT_LOOP)
-        #define MAX_HALFBUSY_WAIT_TIME  4
-        double destTime = GetTime() + ms/1000;
-        if (ms > MAX_HALFBUSY_WAIT_TIME) ms -= MAX_HALFBUSY_WAIT_TIME;
+    #if defined(SUPPORT_PARTIALBUSY_WAIT_LOOP)
+        double busyWait = ms*0.05;     // NOTE: We are using a busy wait of 5% of the time
+        ms -= (float)busyWait;
     #endif

+    // System halt functions
     #if defined(_WIN32)
         Sleep((unsigned int)ms);
     #endif
@@ -4571,14 +4833,80 @@ static void Wait(float ms)
         usleep(ms*1000.0f);
     #endif

-    #if defined(SUPPORT_HALFBUSY_WAIT_LOOP)
-        while (GetTime() < destTime) { }
+    #if defined(SUPPORT_PARTIALBUSY_WAIT_LOOP)
+        double previousTime = GetTime();
+        double currentTime = 0.0;
+
[...]

PR #2503, opened 2022-06-04
Fix regression that was causing video stuttering

flashback-fx:
Hello,

This PR fixes the regression described in this issue (https://github.com/raysan5/raylib/issues/2434), in which video stuttering or high amount of tearing was found when running Raylib programs compiled with SUPPORT_PARTIALBUSY_WAIT_LOOP.

The regression was introduced in commit 4e9afac2a55cd3d7b347d9e3706e0dd370386ff7 (https://github.com/raysan5/raylib/commit/4e9afac2a55cd3d7b347d9e3706e0dd370386ff7), and is caused by the way the the busy wait time limit is checked before returning. In the previous version, a target time was calculated by adding the desired time to wait to the current time, and, after sleeping, the remaining busy waiting was done by continously checking the current time until it passed the target time. The stated commit changed this behavior, making it calculate the desired duration of the busy wait, sleeping, and then making the busy wait continously check the difference between the current time and the time the busy wait started, until that difference was higher than the desired duration of the busy wait. I suppose that this second way of checking was less accurate because it does not account for possible imprecision of the actual time spent sleeping.

This PR restores the old partial busy waiting end checking behavior while keeping the busy wait time to 5% of the requested wait time. I've tested it using the method described in the issue linked above and got results similar to the ones I had for version 3.7, while keeping the CPU utilization similar to the one of the current 4.1-dev version.

Please let me know if you're OK with this small change or if you'd like me to change something.

I've also made some optimizations to WaitTime to avoid unnecessary type conversions and divisions but did not include them on this PR to keep the scope on resolving the regression. If you merge this PR I'd like to open a new one with these optimizations.

Thanks a lot, and as always thanks for Raylib.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4834,6 +4834,7 @@ void WaitTime(float ms)
     while ((currentTime - previousTime) < ms/1000.0f) currentTime = GetTime();
 #else
     #if defined(SUPPORT_PARTIALBUSY_WAIT_LOOP)
+        double destTime = GetTime() + ms/1000.0f;
         double busyWait = ms*0.05;     // NOTE: We are using a busy wait of 5% of the time
         ms -= (float)busyWait;
     #endif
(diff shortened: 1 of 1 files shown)


raysan5:
@flashback-fx thank you very much for reviewing this issue and the fix! :D 

Feel free to send a second PR with the optimizations for review and discussion.

flashback-fx:
Here is the second PR: https://github.com/raysan5/raylib/pull/2504

PR #2504, opened 2022-06-05
Avoid unnecesary casts and divisions in WaitTime

flashback-fx:
The changes in this PR avoid unnecessary casts and divisions in WaitTime.

I have observed that, while many of Raylib's time functions work with seconds, WaitTime works with milliseconds. When WaitTime is called from EndDrawing, the time data is converted from seconds to milliseconds. Inside WaitTime, the time data is converted back to seconds (by dividing it by 1000) in some places, one being the continuous check the busy wait loop does. The only place where the milliseconds value is used straight away seems to be when calling the Win32 Sleep function. Since Win32 Sleep, MacOS usleep and Linux nanosleep seem to use different time magnitudes, I think it makes more sense for WaitTime to work with the standard Raylib time magnitude and only scale it when calling each platform sleeping mechanism (instead of scaling it to fit Win32 and then scaling it back at many points inside the function).

This PR does that. It makes WaitTime work with seconds. It also makes WaitTime receive a double argument instead of a float one. This matches the return type of GetTime and the types of the CORE.Time fields, and avoids some unnecessary casting. Finally, this PR alters the body WaitTime sightly to avoid mutating its parameter through its code. I believe that this makes the code easier to understand. Calls to WaitTime have been adapted to fit these changes.

Everything should be pretty equivalent to what the current code in master does, but unfortunately I cannot test neither on Windows nor on Mac. It seems to work well under Linux.

Please let me know if you're OK with this change and if you'd like me to alter something from the PR.

I've seen more potential improvements but, as with my previous PR, I prefer to go step by step.

Thanks again!

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -958,7 +958,7 @@ RLAPI void DisableEventWaiting(void);                             // Disable wai
 // To avoid that behaviour and control frame processes manually, enable in config.h: SUPPORT_CUSTOM_FRAME_CONTROL
 RLAPI void SwapScreenBuffer(void);                                // Swap back buffer with front buffer (screen drawing)
 RLAPI void PollInputEvents(void);                                 // Register all input events
-RLAPI void WaitTime(float ms);                                    // Wait for some milliseconds (halt program execution)
+RLAPI void WaitTime(double secondsToWait);                        // Wait for some time (halt program execution)
 
 // Cursor-related functions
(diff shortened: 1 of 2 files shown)


flashback-fx:
I've just realized that WaitTime can not be so happily changed because it is a public function available in raylib.h since the custom frame control option was implemented. I will edit this message later to discuss alternatives.

Issue #2505, opened 2022-06-05
[core] Change WaitTime signature to work with seconds and have its parameter be of type double


flashback-fx:
Most of Raylib's time functions work with time represented as seconds in variables of type double. Examples of this are the return value of GetTime and the fields of CORE.Time. The WaitTime function is different: it takes a milliseconds value in a variable of type float. This makes WaitTime not consistent with the rest of the framework and forces some unnecessary conversions.

I think the reason for WaitTime (originally called Wait) being different is because it was originally designed around the Win32 Sleep function (even though in its first implementation it already had Linux support, see commit b681e8c2775bbb467c0f0607afd9840fda25c563).

This is a breaking change, since WaitTime became a public function when the custom game loop control feature was added. In any case, this feature is relatively new (it's still not listed in the cheatsheet) and I think the sooner this change is done the less people's code will break.

I've stumbled upon this when working in this PR (https://github.com/raysan5/raylib/pull/2504).

I've opened an issue because I would like to know the opinion of the community regarding this proposed change.

Thank you.

raysan5:
@flashback-fx Did you measure in some way performance improvement of this change? I see the consistency with the other time-related functions but, in terms of benefits for the users, do you think it will be more intuitive to use? Afaik, most `wait` functions out there use milliseconds, don't they?

raysan5:
@flashback-fx Thanks for the investigation and the report. Ok, let's unify it to seconds. Feel free to send a PR.

flashback-fx:
Thank you. Here is the PR https://github.com/raysan5/raylib/pull/2506 .

orcmid:
In the Standard C library with header `<time.h>`, the preprocessor variable `CLOCKS_PER_SECOND` is your friend as is the defined type `clock_t`.  I would hope these percolate up into the raylib provisions around sub-second time-keeping.

PR #2506, opened 2022-06-06
Change WaitTime argument from milliseconds to seconds

flashback-fx:
Pull request corresponding to issue https://github.com/raysan5/raylib/issues/2505 .


raysan5:
@flashback-fx thanks for the improvement!

PR #2508, opened 2022-06-06
Unify busy waiting behavior across conditional compilation branches

flashback-fx:
I have more changes to propose regarding WaitTime.

Currently, WaitTime does the busy waiting in a sightly different way depending on wether Raylib is compiled SUPPORT_BUSY_WAIT_LOOP or SUPPORT_PARTIALBUSY_WAIT_LOOP defined.

When compiling Raylib with SUPPORT_BUSY_WAIT_LOOP, the busy waiting is done like this:
```
    double previousTime = GetTime();
    double currentTime = 0.0;

    // Busy wait loop
    while ((currentTime - previousTime) < seconds) currentTime = GetTime();
```
When compiling Raylib with SUPPORT_PARTIALBUSY_WAIT_LOOP, the busy waiting is done like this:
```
    while (GetTime() < destinationTime) { }
```

This PR extracts the busy waiting code to a static function that is only compiled if SUPPORT_BUSY_WAIT_LOOP or SUPPORT_PARTIALBUSY_WAIT_LOOP are defomed, and calls it in the appropiate places. The waiting code used in this PR is the one that is currently being used in master for SUPPORT_PARTIALBUSY_WAIT_LOOP.

This could be also achieved without an extra static function but that would result in duplicated code so I preferred to add it. I can change it if you don't like the extra function.

If possible, I would also like to extract lines 4843 to 4859 to another static function, that could be called `SleepFor` or `SleepTime`. I think that would make the body of WaitTime more readable, especially regarding the conditional compilation branches. I think that, if we do that, we can also get rid of the `sleepSeconds` variable and pass the value straight to this functions. These lines correspond to the sleep code:
```
    // System halt functions
    #if defined(_WIN32)
        Sleep((unsigned int)sleepSeconds*1000.0);
    #endif
    #if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__EMSCRIPTEN__)
        struct timespec req = { 0 };
        time_t sec = sleepSeconds;
        long nsec = (sleepSeconds - sec)*1000000000L;
        req.tv_sec = sec;
        req.tv_nsec = nsec;

        // NOTE: Use nanosleep() on Unix platforms... usleep() it's deprecated.
        while (nanosleep(&req, &req) == -1) continue;
    #endif
    #if defined(__APPLE__)
        usleep(sleepSeconds*1000000.0);
    #endif
```
What do you think?

I'd also like to take the opportunity to ask if it's OK to do so many PRs. I don't want to overwhelm you with so many, I am doing it with the intention of proposing very tightly scoped changes and document them, but maybe you prefer less PRs with more changes? (The change I am proposing here could have also been in https://github.com/raysan5/raylib/pull/2503 , but I really like how that PR is focused on solving the regression and no more)

In any case, thanks for your time.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -686,6 +686,10 @@ static void RecordAutomationEvent(unsigned int frame);      // Record frame even
 static void PlayAutomationEvent(unsigned int frame);        // Play frame events (from internal events array)
 #endif
 
+#if defined(SUPPORT_BUSY_WAIT_LOOP) || defined(SUPPORT_PARTIALBUSY_WAIT_LOOP)
+static void BusyWaitUntil(double destinationTime);
+#endif
+
 #if defined(_WIN32)
 // NOTE: We declare Sleep() function symbol to avoid including windows.h (kernel32.lib linkage required)
 void __stdcall Sleep(unsigned long msTimeout);              // Required for: WaitTime()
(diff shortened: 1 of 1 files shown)


raysan5:
@flashback-fx The improvement is ok but I prefer to avoid additional functions, actually, I have no problem with duplicate code. 

About the PRs, I prefer multiple small self-contained PRs than big ones.

flashback-fx:
OK, pushed a commit that removes the static function.

raysan5:
@flashback-fx Nice! Thanks for the review! Merging!

flashback-fx:
The pleasure is mine, I hope to keep contributing in the future.

PR #2514, opened 2022-06-09
GetMonitorWidth()/GetMonitorHeight(): current video resolution instead max available

gulrak:
This change makes `GetMonitorWidth()` and `GetMonitorHeight()` return the dimensions of the actual video mode used, which is the one the user chose to run his desktop on and allows to make informed code to center or position the window in respect to the desktop resolution. It is based on the code that is used for `InitGraphicDevice` to detect the current mode.

The previous code returned the maximum possible resolution and does not help in window positioning/resizing as it does not relate to the current desktop resolution at all.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1723,7 +1723,7 @@ int GetCurrentMonitor(void)
 #endif
 }
 
-// Get selected monitor width
+// Get selected monitor position
 Vector2 GetMonitorPosition(int monitor)
 {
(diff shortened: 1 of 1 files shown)


gulrak:
The failed cmake linux build looks like a ci pickup, as it didn't even come to any project specific work. I did'n want to push some whitespace change to just retriever the build.

raysan5:
@gulrak Thanks for the review! :)

PR #2515, opened 2022-06-09
Minor README edit. Starting text is under raylib image

gaenrique:
The initial text on the README file looks awkward when aligned to the right side of the raylib image. The text is too close to the image and makes it seems like some of the text is cut off. This PR makes it so the text starts below the image.

image


gaenrique:
If you believe the image might look better center-aligned with the text under it I am happy to change that too.

raysan5:
@gaenrique Sorry but I like the text aligned to the right of the logo.

PR #2517, opened 2022-06-12
Add function to read both X and Y mouse scrolling from a trackpad

### LINKED ISSUE #2480: [core] Provide a function to get wheel movement on x and y, for trackpad-based input devices
### Issue description

On my Mac, the trackpad scroll provides a blend of both the X and Y scroll. This was "fixed" recently via #2437, but in a way that still only provides one value. So writing, for instance, a GUI which has both horizontal and vertical scrollbars is impossible.

However, looking [at the code](https://github.com/raysan5/raylib/blob/020e90319388fb6e0975d2bb502d4824d369aece/src/rcore.c#L5403-L5407), openGL is providing both x and y scrolling. Raylib also has many functions that are repeated for both X/Y and Vector2 overloads. So it would make logical sense to provide a `GetMouseWheelMoveV` function that returns a `Vector2` with both x and y components, instead of picking the largest amplitude. I'll also note that for whatever reason the x scroll scale is much bigger on the trackpad (that is, the x value seems to be larger than the y value based on similar movements on my trackpad when scrolling).

I actually have a working version of this function, and can submit a PR, but I have not explored all the ramifications, and don't want to move further unless it's agreed such a function can be added.

### Environment

MacOS Monterey 12.3.1, Apple M1 Pro ARM, magic trackpad.

### Code Example

Simple D program that writes a scroll position with movement of the text on the screen:

```d
import raylib;

void main()
{
    InitWindow(500, 500, "raylibtest");
    SetTargetFPS(60);

    float pos = 0;
    while(!WindowShouldClose())
    {
        float mwm = GetMouseWheelMove();
        pos = pos + mwm;
        auto drawpos = pos;
        if(drawpos < 0) drawpos = 0;
        BeginDrawing();
        ClearBackground(Colors.WHITE);
        DrawText(TextFormat("%f", pos), 10, cast(int)drawpos, 20, Colors.BLACK);
        EndDrawing();
    }
}
```

schveiguy:
This substitutes the internal representation of the single scroll value with a `Vector2`. When the original `GetMouseScrollWheel` function is called, it performs the same task that was previously done for the x/y callback function (take the component with the largest magnitude). Callbacks that just provide one value are put into the Y component.

~~I also modified the scroll wheel example as appropriate, but I can modify it to get the original functionality back if needed.~~ EDIT: reverted back to the original.

BIG NOTE: this is only tested on MacOS by me (and only with a modified scroll wheel example), I don't have experience building any of the other platforms (Android, web, etc).

I tried to update what I think is some playback system, but I'm not sure if I did that right, please pay close attention! I did not test that functionality.

I'm also unfamiliar with the requirements for coding style, etc. Please let me know what I might need to change!

Should close #2480.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -456,8 +456,8 @@ typedef struct CoreData {
 
             char currentButtonState[MAX_MOUSE_BUTTONS];     // Registers current mouse button state
             char previousButtonState[MAX_MOUSE_BUTTONS];    // Registers previous mouse button state
-            float currentWheelMove;         // Registers current mouse wheel variation
-            float previousWheelMove;        // Registers previous mouse wheel variation
+            Vector2 currentWheelMove;       // Registers current mouse wheel variation
+            Vector2 previousWheelMove;      // Registers previous mouse wheel variation
 #if defined(PLATFORM_RPI) || defined(PLATFORM_DRM)
             // NOTE: currentButtonState[] can't be written directly due to multithreading, app could miss the update
             char currentButtonStateEvdev[MAX_MOUSE_BUTTONS]; // Holds the new mouse state for the next polling event to grab
@@ -531,7 +531,7 @@ typedef enum AutomationEventType {
     INPUT_MOUSE_BUTTON_UP,          // param[0]: button
     INPUT_MOUSE_BUTTON_DOWN,        // param[0]: button
     INPUT_MOUSE_POSITION,           // param[0]: x, param[1]: y
-    INPUT_MOUSE_WHEEL_MOTION,       // param[0]: delta
+    INPUT_MOUSE_WHEEL_MOTION,       // param[0]: x delta, param[1]: y delta
     INPUT_GAMEPAD_CONNECT,          // param[0]: gamepad
     INPUT_GAMEPAD_DISCONNECT,       // param[0]: gamepad
     INPUT_GAMEPAD_BUTTON_UP,        // param[0]: button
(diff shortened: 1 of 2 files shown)


examples/core/core_input_mouse_wheel.c
```diff
@@ -31,7 +33,7 @@ int main(void)
     {
         // Update
         //----------------------------------------------------------------------------------
-        boxPositionY -= (GetMouseWheelMove()*scrollSpeed);
+        boxPosition = Vector2Subtract(boxPosition, Vector2Scale(GetMouseWheelMoveV(), scrollSpeed));
```
raysan5, line 36: Please, could we keep this file as it was, for most novice users `GetMouseWheelMove()` is enough, no need to increase in one level of complexity this basic example.

src/rcore.c
```diff
@@ -3829,8 +3829,26 @@ float GetMouseWheelMove(void)
...
+#endif
+
+    return result;
+}
+
+// Get mouse wheel movement X/Y as a vector
+Vector2 GetMouseWheelMoveV(void)
+{
+#if defined(PLATFORM_ANDROID)
+    return Vector2Zero();
```
raysan5, line 3848: Please, I prefer to minimize `raymath.h` usage. For those simple functions a return `(Vector2){ 0.0f, 0.0f }` or a manual scaling of values should work.

src/rcore.c
```diff
@@ -4926,7 +4944,7 @@ void PollInputEvents(void)
 
     // Register previous mouse states
     CORE.Input.Mouse.previousWheelMove = CORE.Input.Mouse.currentWheelMove;
-    CORE.Input.Mouse.currentWheelMove = 0.0f;
+    CORE.Input.Mouse.currentWheelMove = (Vector2){0.0f, 0.0f};
```
raysan5, line 4947: Please, could you add space after the parentesis, it's a raylib convention I use: `(Vector2){ 0.0f, 0.0f }`

schveiguy:
Comments addressed

raysan5:
@schveiguy Thanks for the addition!

PR #2521, opened 2022-06-14
add wrap function

Tekkitslime:
Implementation of a wrap function, mainly to avoid overflow.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -158,7 +158,7 @@ typedef struct float16 {
     float v[16];
 } float16;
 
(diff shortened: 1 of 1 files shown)


Peter0x44:
To clarify, godot also provides such a function
https://docs.godotengine.org/en/stable/classes/class_@gdscript.html#class-gdscript-method-wrapf
so it's not entirely without precedent
still not sure this is something commonly used or worth it, but I can't see a reason not to include it either

raysan5:
@Tekkitslime Thanks for the addition!

quantumedbox:
It's quite quite handy when you need to operate on wrapping subrange. For example, clock might be wrapped between -PI and PI to get rotation in closed range. Time in general follows such repeating systems often.

Issue #2523, opened 2022-06-16
[rtexture] ColorAlphaBlend round off error


royqh1979:
Please, before submitting a new issue verify and check:

 - [yes] I tested it on latest raylib version from master branch
 - [yes] I checked there is no similar issue already reported
 - [maybe] My code has no errors or misuse of raylib

### Issue description

In  ColorAlphaBlend, source corlor is tinted before blend as the following:
```
Color ColorAlphaBlend(Color dst, Color src, Color tint)
{
    Color out = WHITE;

    // Apply color tint to source color
    src.r = (unsigned char)(((unsigned int)src.r*(unsigned int)tint.r) >> 8);
    src.g = (unsigned char)(((unsigned int)src.g*(unsigned int)tint.g) >> 8);
    src.b = (unsigned char)(((unsigned int)src.b*(unsigned int)tint.b) >> 8);
    src.a = (unsigned char)(((unsigned int)src.a*(unsigned int)tint.a) >> 8);
```
There's a round off error in these calculations. Since  we usally define color's rgba components in range 0-255, src.r*tint.r >> 8 will always generate a new number that is less than the original number.

So even if  we set the tint color as WHITE,  the result is still not the original source color. If we alphablend a color many times, finally it will become BLACK.

### Code Example

I'm writing a program that display the sin wave animatedly.  I'm trying to copy part of the last frame in each iteration, to make the wave line. It's done by the ImageDraw funtion, and it calls ColorAlphaBlend. The frame becomes blacked as it is copied many times.

```
#include <raylib.h>
#include <math.h>



int main() {
	InitWindow(1200,480,"rdrawing");
	SetTraceLogLevel(LOG_WARNING);
	SetTargetFPS(60);
	
	Image imgUnitCircle=GenImageColor(400,480,BLANK);
	
	Image imgSin=GenImageColor(760,480,BLANK);
	
	float r = 0;
	float lastY=240;
	float x,y;
	while(!WindowShouldClose()) {
		//update datas
		r += 0.02;
		if (r > PI * 2) r -= PI * 2;
		x=200+150*cos(r);
		y=240-150*sin(r);
		//update image (in CPU)
		ImageClearBackground(&imgUnitCircle,BLANK);
		ImageDrawCircle(&imgUnitCircle,200,240,150,RED);
		ImageDrawLine(&imgUnitCircle,200,240,x,y,RED);
		ImageDrawCircle(&imgUnitCircle,x,y,7,DARKBROWN);
[...]

PR #2524, opened 2022-06-16
fix: round off error in ColorAlphaBlend

royqh1979:
This PR fixes the problem described in https://github.com/raysan5/raylib/issues/2523


raysan5:
@royqh1979 thanks for the fix 🙂

PR #2525, opened 2022-06-16
Changed ExportFontAsCode to use given font's padding.

TheTophatDemon:
When using ExportFontAsCode(...), the exported .h file sets the font padding to the value of the constant FONT_TTF_DEFAULT_CHARS_PADDING. This results in incorrect rendering of certain AngelCode fonts, so this pull request instead uses the glyphPadding of the Font given as an argument.


raysan5:
@TheTophatDemon thanks for the fix

PR #2532, opened 2022-06-19
Share PCM streaming buffer to reduce dynamic allocations

dbechrd:
Assuming `UpdateMusicStream()` is not meant to be called from multiple threads at the same time, this drastically reduces the number of dynamic allocations that Raylib's Audio API does at runtime when playing music.

However, if thread safety is required, you might consider doing a similar optimization by maintaining a PCM sample buffer per thread. I tried to use `thread_local`, but it didn't compile in the default MSVC C compiler as it requires C11. When I tried compiling Raylib using the C11 switch, I got other, unrelated errors about `stdalign.h`.

I did compile this and my music played properly, but that's about all the testing I've done. Just figured I'd share in case you find the general idea useful. You may consider any code in this contribution public domain or Zlib licensed, at your preference.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -351,6 +351,8 @@ typedef struct AudioData {
         ma_device device;           // miniaudio device
         ma_mutex lock;              // miniaudio mutex lock
         bool isReady;               // Check if audio device is ready
+        size_t pcmCapacity;
+        void *pcm;
     } System;
     struct {
         AudioBuffer *first;         // Pointer to first AudioBuffer in the list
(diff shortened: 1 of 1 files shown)


src/raudio.c
```diff
@@ -1726,7 +1729,12 @@ void UpdateMusicStream(Music music)
     unsigned int subBufferSizeInFrames = music.stream.buffer->sizeInFrames/2;
 
     // NOTE: Using dynamic allocation because it could require more than 16KB
-    void *pcm = RL_CALLOC(subBufferSizeInFrames*music.stream.channels*music.stream.sampleSize/8, 1);
+    size_t pcmSize = subBufferSizeInFrames * music.stream.channels * music.stream.sampleSize / 8;
+    if (AUDIO.System.pcmCapacity < pcmSize) {
+        RL_FREE(AUDIO.System.pcm);
+        AUDIO.System.pcm = RL_CALLOC(1, pcmSize);
```
dbechrd, line 1735: You should obviously just use REALLOC here instead, but the compiler was whining about possible memory leaks due to not checking return value of realloc before overwriting the pointer. -_-

PR #2535, opened 2022-06-23
examples: fix passing long* to localtime when it expects long long*

quantumedbox:


raysan5:
`GetFileModTime()` is expected to return a long by design.

PR #2536, opened 2022-06-23
fix: small things in CONVENTIONS.md

quantumedbox:


raysan5:
@quantumedbox thanks for the review

PR #2538, opened 2022-06-23
file loading: ditch SEEK_END strategy for telling stream len

quantumedbox:
This PR changes how files are loaded in `LoadFileData`. Instead of using `fseek`/`ftell`, that might not be meaningfully supported for some streams, it uses strategy of progressively reallocated buffer and one-pass reading.

This allows, for example, using filesystem bound sockets for loading resources in raylib.

There's probably a negative impact to performance by doing so, but in my opinion, as resource loading is usually only ever done once, - not so significant slowdowns are acceptable.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -2438,9 +2438,9 @@ static unsigned char *LoadFileData(const char *fileName, unsigned int *bytesRead
 
                 if (ferror(file))
                 {
-                    TRACELOG(LOG_WARNING, "FILEIO: [%s] Failed to read file", fileName);
                     RL_FREE(data);
                     data = NULL;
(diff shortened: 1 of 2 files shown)


quantumedbox:
Not sure whether `GetFileLength` should be reimplemented in the same manner though. It probably might, as currently it implies that anything passed there should support seeking. So, reading till the EOF to get the size, and then setting the cursor to beginning might be valid.

raysan5:
@quantumedbox I'm afraid I'm neither merging this change, it adds a unneded level of code complexity and cost-penalty for a very unprobable specific situation. I never had any issue `fseek()` but if it happens at some point, I will review this PR.

PR #2539, opened 2022-06-23
Fix signedness in rlBindImageTexture

hanaxar:
rlGetGlTextureFormats was expecting unsigned int, corrected variables according to that.


raysan5:
@hanaxar thanks for the review! 😄

PR #2541, opened 2022-06-23
fix: precision loss, discard unused

quantumedbox:


raysan5:
@quantumedbox thanks for the review!

PR #2542, opened 2022-06-23
WIP rcamera redesign

Crydsch:
## rcamera re-design

This PR addresses the `rcamera.h` redesign as discussed in #2507.

The core functionality is finished and working as expected.
There are 4 camera modes: `CAMERA_FREE`, `CAMERA_FIRST_PERSON`, `CAMERA_THIRD_PERSON` and `CAMERA_ORBITAL`.
Alongside `rcamera.h` I also extended the `core_3d_camera_first_person` example for testing and as a viable demo.
Despite the filename it currently includes all camera modes and additional information for testing.

## Key-Design notes

All static variables were removed and the camera struct contains the entire camera state.
It can be directly manipulated at any time and the camera will function accordingly.

The camera struct has changed. For example, it now contains two variable `target_position` and `target_distance`.
These take the obvious roles in the modes `CAMERA_THIRD_PERSON` and `CAMERA_ORBITAL`.
However, while prototyping I found that the very same calculations also produce a `CAMERA_FIRST_PERSON` and `CAMERA_FREE`,
by setting the `target_distance` to 0!

In these modes `target_position` becomes the camera eye position.
For convenience I also provide a function `GetCameraEyePosition()` which returns the correct eye position regardless of the mode. This design enables smooth transitions between different camera modes.
Sidenote: A negative `target_distance` is valid and creates a zoom like behavior.

All functions are available in all modes, there are no restrictions.
It may make no sense to use them though - this is at the users discretion.
I did this deliberately to allow for a maximum of freedom for the user.
For example, while at first glance a `CameraRoll(..)` makes no sense in the `CAMERA_FIRST_PERSON` mode,
it may be used to implement a "looking around the corner" feature.

The functions `CameraYaw(..)`, `CameraPitch(..)` and `CameraRoll(..)` take an angle parameter (in radians) which allow precise manipulation.
A `CameraYaw(camera, 90 * DEG2RAD);` call results precisely in a 90 degree turn.

## Behavioural differences to the old rcamera.h
- CAMERA_FREE is no longer a panning camera, but a real 3D free float camera.
- CAMERA_THIRD_PERSON is not restricted to "look up".
   Many third person games allow similar camera freedom, so I left it for now.
   Is this ok?

## API and usage

<details>
<summary>API</summary>

```C
typedef struct Camera3D {
    int mode;
    Vector3 target_position; // Camera target it looks at
    float target_distance;   // Camera eye distance to target_position (set to 0 in free/fps mode)
    Quaternion orientation;  // Camera orientation (its rotation in 3D)
    float fovy;              // Camera field-of-view apperture in Y (degrees) in perspective, used as near plane width in orthographic
    float aspect;            // Camera aspect ratio (typically window_width / window_height)
    int projection;          // Camera projection: CAMERA_PERSPECTIVE or CAMERA_ORTHOGRAPHIC
} Camera3D;

// Initializes a camera with default values (Can be used to reset a camera)
void InitializeCamera(Camera3D* camera, float aspect);
// Select camera mode (multiple camera modes available)
void SetCameraMode(Camera3D* camera, int mode);
// Returns the cameras current front vector
Vector3 GetCameraFront(Camera3D* camera);
// Returns the cameras current up vector
Vector3 GetCameraUp(Camera3D* camera);
// Returns the cameras current right vector
Vector3 GetCameraRight(Camera3D* camera);
// Returns the cameras current eye position
// Note: This is equivalent to target_position if target_distance == 0
//       (i.e. in modes CAMERA_FREE and CAMERA_FIRST_PERSON)
Vector3 GetCameraEyePosition(Camera3D* camera);
// Moves the camera in its current front vector direction
void CameraMoveForward(Camera3D *camera, float distance);
// Moves the camera in its current up vector direction
void CameraMoveUp(Camera3D *camera, float distance);
// Moves the camera target in its current right vector direction
void CameraMoveRight(Camera3D *camera, float distance);
// Rotates the camera around its current up vector
// Yaw is "looking left and right"
// Note: angle must be provided in radians
void CameraYaw(Camera3D *camera, float angle);
// Rotates the camera around its current right vector
// Pitch is "looking up and down"
// Note: angle must be provided in radians
void CameraPitch(Camera3D *camera, float angle);
// Rotates the camera around its current front vector
// Roll is "turning your head sideways to the left or right"
// Note: angle must be provided in radians
void CameraRoll(Camera3D *camera, float angle);
// Returns the current camera view matrix
Matrix GetCameraViewMatrix(Camera3D* camera);
// Returns the current camera projection matrix
Matrix GetCameraProjectionMatrix(Camera3D* camera);
```

</details>

<br>
  
Usage (still using raylibs input system)
```C
Camera3D camera = {0};
InitializeCamera(&camera, (double)screenWidth / (double)screenHeight);
camera.target_position = (Vector3){ 0.0f, 2.0f, 0.0f }; // This is the eye position in FREE (and FIRST_PERSON) mode
//SetCameraMode(camera, CAMERA_THIRD_PERSON); // Set a different camera mode

// Camera movement
if (IsKeyDown(KEY_W)) CameraMoveForward(camera, CAMERA_MOVE_SPEED);
if (IsKeyDown(KEY_S)) CameraMoveForward(camera, -CAMERA_MOVE_SPEED);
if (IsKeyDown(KEY_D)) CameraMoveRight(camera, CAMERA_MOVE_SPEED);
if (IsKeyDown(KEY_A)) CameraMoveRight(camera, -CAMERA_MOVE_SPEED);

// Camera rotation
Vector2 mousePositionDelta = GetMouseDelta();
CameraYaw(camera, mousePositionDelta.x * CAMERA_MOUSE_MOVE_SENSITIVITY);
CameraPitch(camera, mousePositionDelta.y * CAMERA_MOUSE_MOVE_SENSITIVITY);
```
  
## TODO
- [x] Core functionality (math)
- [ ] Adapt all examples
- [ ] View bobbing
- [ ] CAMERA_ORBITAL autorotation
- [ ] New input system
- [ ] Review compatibility with old rcamera.h usage
- [ ] Review stand alone mode

## Feedback

I'd much appreciate feedback!

If you want to have a look at the code or the demo, you can checkout my feature branch [rcamera_redesign](https://github.com/Crydsch/raylib/tree/rcamera_redesign) or this PR..

One question regarding the API:
Would you prefer  
`void InitializeCamera(Camera3D* camera, float aspect);`  
in line with all other functions or  
`Camera3D InitializeCamera(float aspect);` ?

### DIFF
--- a/examples/core/core_3d_camera_first_person.c
+++ b/examples/core/core_3d_camera_first_person.c
@@ -10,6 +10,7 @@
 ********************************************************************************************/
 
 #include "raylib.h"
+#include "raymath.h"
 
 #define MAX_COLUMNS 20
 
@@ -25,13 +26,14 @@ int main(void)
 
     InitWindow(screenWidth, screenHeight, "raylib [core] example - 3d camera first person");
 
-    // Define the camera to look into our 3d world (position, target, up vector)
-    Camera camera = { 0 };
-    camera.position = (Vector3){ 4.0f, 2.0f, 4.0f };
-    camera.target = (Vector3){ 0.0f, 1.8f, 0.0f };
-    camera.up = (Vector3){ 0.0f, 1.0f, 0.0f };
-    camera.fovy = 60.0f;
-    camera.projection = CAMERA_PERSPECTIVE;
+    // Define the camera to look into our 3d world
+    Camera3D camera = {0};
+    InitializeCamera(&camera, (double)screenWidth / (double)screenHeight);
+    camera.target_position = (Vector3){ 0.0f, 2.0f, 0.0f }; // This is the eye position in FREE (and FIRST_PERSON) mode
+    //SetCameraMode(camera, CAMERA_THIRD_PERSON); // Set a different camera mode
+
+    // Catch curser
+    DisableCursor();
 
     // Generates some random columns
     float heights[MAX_COLUMNS] = { 0 };
@@ -41,11 +43,9 @@ int main(void)
     for (int i = 0; i < MAX_COLUMNS; i++)
     {
         heights[i] = (float)GetRandomValue(1, 12);
-        positions[i] = (Vector3){ (float)GetRandomValue(-15, 15), heights[i]/2.0f, (float)GetRandomValue(-15, 15) };
+        positions[i] = (Vector3){ (float)GetRandomValue(-15, 15), heights[i] / 2.0f, (float)GetRandomValue(-15, 15) };
         colors[i] = (Color){ GetRandomValue(20, 255), GetRandomValue(10, 55), 30, 255 };
     }
-
-    SetCameraMode(camera, CAMERA_FIRST_PERSON); // Set a first person camera mode
 
     SetTargetFPS(60);                           // Set our game to run at 60 frames-per-second
     //--------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Crydsch:
Please note that only the `core_3d_camera_first_person` example has been adapted yet.
This is why any CI check building all examples is failing.

quantumedbox:
As i understand, Raylib API tends to use value semantics for returning the result, so, `Camera3D InitializeCamera(float aspect);` makes it more in line with everything else.

As a small side-effect it will allow languages like Zig, that could wrap C directly, to infer type of variables.
```zig
var camera = ray.InitializeCamera(...);
```

Also, other API functions related to initialization tend to have `Init` prefix, would be better to follow that too.

Iniquitatis:
Wouldn't it be more correct to rename `GetCameraFront` to `GetCameraForward`? I believe it's called "forward vector", not "front vector".

examples/core/core_3d_camera_first_person.c
```diff
@@ -25,13 +26,14 @@ int main(void)
...
-    // Define the camera to look into our 3d world (position, target, up vector)
-    Camera camera = { 0 };
-    camera.position = (Vector3){ 4.0f, 2.0f, 4.0f };
-    camera.target = (Vector3){ 0.0f, 1.8f, 0.0f };
-    camera.up = (Vector3){ 0.0f, 1.0f, 0.0f };
-    camera.fovy = 60.0f;
-    camera.projection = CAMERA_PERSPECTIVE;
+    // Define the camera to look into our 3d world
+    Camera3D camera = {0};
+    InitializeCamera(&camera, (double)screenWidth / (double)screenHeight);
```
raysan5, line 31: I prefer current approach, exposing the standard camera parameters to the user directly instead of hiding them inside an opaque `InitCamera()`.

src/raylib.h
```diff
@@ -297,11 +297,13 @@ typedef struct Font {
 
 // Camera, defines position/orientation in 3d space
 typedef struct Camera3D {
-    Vector3 position;       // Camera position
-    Vector3 target;         // Camera target it looks-at
-    Vector3 up;             // Camera up vector (rotation over its axis)
-    float fovy;             // Camera field-of-view apperture in Y (degrees) in perspective, used as near plane width in orthographic
-    int projection;         // Camera projection: CAMERA_PERSPECTIVE or CAMERA_ORTHOGRAPHIC
+    int mode;
```
raysan5, line 300: I'm afraid I have no plans to change current camera structure. Provided parameters are clear and standard for users. It will also break many codebases.

src/rcore.c
```diff
@@ -2558,25 +2536,9 @@ Ray GetMouseRay(Vector2 mouse, Camera camera)
...
-    {
-        float aspect = (float)CORE.Window.screen.width/(float)CORE.Window.screen.height;
-        double top = camera.fovy/2.0;
-        double right = top*aspect;
-
-        // Calculate projection matrix from orthographic
-        matProj = MatrixOrtho(-right, right, -top, top, 0.01, 1000.0);
-    }
+    // Get camera matrices
+    Matrix matView = GetCameraViewMatrix(&camera);
```
raysan5, line 2540: This is a great improvement but camera should be aware at all time of width/height. How is it managed if window is resized?

examples/core/core_3d_camera_first_person.c
Crydsch: Sure. I agree it is nicer for the user to see everything directly.
I'll remove the function and adjust this in the example.

src/raylib.h
raysan5: @Crydsch I'm afraid changing the current `Camera` structure is a breaking change and it could break many codebases and implementations. I also find the new implementation more confusing for users than current one, raylib prioritize simplicity for less-experienced users and current Camera provides that. I think a `Quaternion orientation` approach is confusing for many developers starting with raylib, probably not for more advance users.

In any case, I think current camera parameters could be kept as they are and transformations could be done inside the update code.

src/rcore.c
raysan5: I just noticed from that approach that `rcamera` module is coupled to `rcore` module, current camera module is decouple from other modules, it can be removed at any time without problem (letting the users manage their own cameras).

 Also, moving aspect or width/height to camera side will require a continuous update in case windows was resized or increasing the coupling between modules (doing the camera update in rcore screen resizing functions).

src/raylib.h
TokyoSU: Well a camera example like the other one already available could just fix the problem, i agree that the current implementation is just really simple, i would prefer a more professional way than a limited camera, its only my answer through, you can create as many example to ease the breaking change.

Crydsch:
I just pushed another approach (#2563) for this redesign, keeping the camera struct and vectors based design.
If there are no other comments this PR can be closed in favor for #2563.

raysan5:
@Crydsch Excuse me for the late response and thank you very much for all the work you are putting on this redesign, it's a big change for raylib and I'd like to break as few things as possible. Keeping the original struct and the module decoupled helps on that side.

We can keep this PR open for the moment, at least until a final solution is merged.

raysan5:
@Crydsch I'm closing this PR now, the new one is getting big and we can focus the follow-up on it.

Issue #2546, opened 2022-06-23
Examples fail to link on Raspberry Pi Bullseye X11


electronstudio:
```
gcc -o textures/textures_sprite_button textures/textures_sprite_button.c -Wall -std=c99 -D_DEFAULT_SOURCE -Wno-missing-braces -Wunused-result -s -O1 -D_DEFAULT_SOURCE -I. -I/home/pi/raylib/src -I/home/pi/raylib/src/external -I/usr/local/include -L. -L/home/pi/raylib/src -L/home/pi/raylib/src -L/usr/local/lib -lraylib -lGL -lm -lpthread -ldl -lrt -lX11 -DPLATFORM_DESKTOP
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_resource_manager_data_stream_set_absolute_cursor':
raudio.c:(.text+0x8d8c): undefined reference to `__atomic_exchange_8'
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_job_queue_post':
raudio.c:(.text+0x12738): undefined reference to `__atomic_load_8'
/usr/bin/ld: raudio.c:(.text+0x12760): undefined reference to `__atomic_load_8'
/usr/bin/ld: raudio.c:(.text+0x12774): undefined reference to `__atomic_load_8'
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_job_queue_next':
raudio.c:(.text+0x128a4): undefined reference to `__atomic_load_8'
/usr/bin/ld: raudio.c:(.text+0x128c0): undefined reference to `__atomic_load_8'
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o):raudio.c:(.text+0x128e0): more undefined references to `__atomic_load_8' follow
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_node_set_state_time':
raudio.c:(.text+0x3e104): undefined reference to `__atomic_exchange_8'
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_node_get_state_time':
raudio.c:(.text+0x3e154): undefined reference to `__atomic_load_8'
/usr/bin/ld: /usr/local/lib/libraylib.a(raudio.o): in function `ma_node_read_pcm_frames':
raudio.c:(.text+0x3e8bc): undefined reference to `__atomic_fetch_add_8'
[...]

PR #2548, opened 2022-06-24
Fix CMake build on Raspberry Pi OS Bullseye

electronstudio:
Fixes https://github.com/raysan5/raylib/issues/2546 for CMake builds at least


raysan5:
@electronstudio Thanks for the fix!

Issue #2550, opened 2022-06-24
[rtext] `LoadFontEx()` fails to package some specified codepoints 


iikare:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

I am using the SMuFL specification and a SMuFL-compatible font to render music symbols using raylib through codepoints in Unicode range `0xE000`-`0xED5F`. However, only the first 8 codepoints I specify are rendered by raylib (ref: first screenshot). I attempted to use a different SMuFL font (i.e. Bravura), and had the same issue. 

Changing the order of the vector defining the codepoints allows different glyphs to be rendered, but the limit of 8 total
characters remains (ref: second screenshot).

Interestingly, the debug output shows the following:

```
INFO: FILEIO: [leland.otf] File loaded successfully
WARNING: FONT: Failed to package character (8)
WARNING: FONT: Failed to package character (9)
WARNING: FONT: Failed to package character (10)
WARNING: FONT: Failed to package character (11)
WARNING: FONT: Failed to package character (12)
WARNING: FONT: Failed to package character (13)
INFO: TEXTURE: [ID 3] Texture loaded successfully (256x256 | GRAY_ALPHA | 1 mipmaps)
INFO: FONT: Data loaded successfully (200 pixel size | 14 glyphs)
```

### Environment

Issue reproduced on both raylib 4.0 and 4.2-dev. 

Desktop/Arch Linux, OpenGL 4.6/3.3, Comet Lake GT2/GTX1070

Compiled via:
`clang++ test.cc -lraylib -o test && ./test`

### Issue Screenshot

The remainder of the codepoints should render next to the 8th codepoint "4". The red text is my commentary.

![image](https://user-images.githubusercontent.com/13582030/175478961-3d27340a-1a31-4ab6-bf93-f23d1254ae22.png)

![image](https://user-images.githubusercontent.com/13582030/175480855-def17832-5d81-4d70-bdc9-ffe6d6e4c7fb.png)

### Code Example

The font provided is from this repository: https://github.com/MuseScoreFonts/Leland

I reproduced this issue with the below minimal working example.

```C++
#include "raylib.h"
//#include "../../src/agh/leland.h"

#include <vector>

using std::vector;

vector<int> codepointSet = {
[...]

PR #2556, opened 2022-06-27
dynamically resize font atlas when atlas is undersized

iikare:
This addresses #2550, #2356. 

When the rectangle packing algorithm estimates the wrong atlas size, the process is repeated with an atlas size with
twice the x/y dimension. This PR does not change the behavior when the original estimated atlas size allows for all glyphs to fit in the atlas. This should address all cases where the estimated atlas size requirement is not sufficient. 

If non-power-of-two textures are acceptable the atlas can be expanded by doubling only one dimension at a time instead. This potentially also allows for the guesstimate of the atlas size to be smaller (i.e. `guessSize`).


raysan5:
@iikare Excuse the late response, I'm still considering adding this feature; it increases code complexity (I don't like `while()` loops, raylib hardly ever uses one of those) and I wonder if atlas size estimation should actually be inside `GenImageFontAtlas()` or not.

iikare:
The `for()` loop can be used here instead quite easily, I've refactored the code to do so.

Also, I don't believe this PR deals with the size estimation, but rather with repacking the atlas when the size estimation fails to be accurate. The size estimation is a tricky problem, I haven't been able to think of a clean solution to the estimation without the current guess and check approach, which only has any performance penalty when the size estimation fails (i.e. rarely).

raysan5:
I'm not merging this PR, I consider it makes the function way more complex and I prefer to avoid it. I reviewed the size estimation formula and now it should work way better (area is considerably increased).

In any case, all estimation code is packed in just 4 lines that can be replaced or maybe moved out of this function in the future. I don't know if size estimation inside the atlas packaging funtion is the best approach. It should probably moved out and provided by the user, and consequently, not fitting characters could just be not packed.

I take a note for future investigation.

PR #2561, opened 2022-07-03
Meson support for Raylib

troglobyte-coder:
Working on providing Meson build support for the RayLib library


src/meson.build
```diff
@@ -0,0 +1,17 @@
+raylib_src = [
+  'raudio.c',
+  'rcore.c',
+  'rglfw.c',
+  'rmodels.c',
+  'rshapes.c',
+  'rtext.c',
+  'rtextures.c',
+  'utiles.c'
```
eli-schwartz, line 9: typo

src/meson.build
```diff
@@ -0,0 +1,17 @@
+raylib_src = [
+  'raudio.c',
+  'rcore.c',
+  'rglfw.c',
```
eli-schwartz, line 4: rglfw seems to be an amalgamation of the glfw source code, it is redundant both:
- when using an external dependency on glfw
- when using a vendored copy (subproject wrap, or src/external/glfw/ checked into git)

src/meson.build
```diff
@@ -0,0 +1,17 @@
+raylib_src = [
+  'raudio.c',
```
eli-schwartz, line 2: cmake has an "audio" option that determines whether to use this.

src/meson.build
```diff
@@ -0,0 +1,17 @@
...
+  'rcore.c',
+  'rglfw.c',
+  'rmodels.c',
+  'rshapes.c',
+  'rtext.c',
+  'rtextures.c',
+  'utiles.c'
+]
+
+ray_lib = static_library('raylib,
```
eli-schwartz, line 12: This should just be a regular `library()`, and respect Meson's "default_library" option for determining whether to build a static or shared library.

Cmake has an option to control which one to build (it defaults to off, in Meson this would be setting `default_options: ['default_library=static']`).

eli-schwartz, line 12: This should be marked for installation (`install: true`). It also should specify the

```meson
version: '4.0.0',
soversion: '400',
```

eli-schwartz (changes requested):

src/meson.build
troglobyte-coder: Yea that makes sense, gonna add that plus some of the helpful suggestions to the first draft.

src/meson.build
troglobyte-coder: That change was gonna be added after ensuring all including the audio RayLib components compiled.

raysan5:
@troglobyte-coder I'm afraid I'm not adding this new build system, simple reason is because I can't maintain it. I recommend implementing it on a separate repo for meson users. I link can be added later on on raylib Wiki.

PR #2563, opened 2022-07-04
WIP rcamera redesign vector

Crydsch:
## rcamera re-design

This PR addresses the `rcamera.h` redesign as discussed in #2507.

This PR is very similar to #2542.
There we evaluated a quaternion-based design, but noticed a breaking change in
the camera struct. The new design was also more confusing to new users.
As raylib prioritizes simplicity for less-experienced users this was problematic.

This PR aims to keep the camera struct as close as possible to the old one
and not introduce breaking changes.
It therefore also keeps the more intuitive vector based implementation (`position, target, up`).

The discussed improvements can still be applied though!
Such as the low-level functions for easy custom camera control.

Note: Some changes may still be necessary to improve the camera module and implement the set requirements.

## Key-Design notes

All static variables will be removed and the camera struct contains the entire camera state.
It can be directly manipulated at any time and the camera will function accordingly.
This gives users full control over the camera and allows customized modes.

All functions are available in all modes, there are no restrictions.
It may make no sense to use them though - this is at the users discretion.
I did this deliberately to allow for a maximum of freedom for the user.
For example, while at first glance a `CameraRoll(..)` makes no sense in the `CAMERA_FIRST_PERSON` mode,
it may be used to implement a "looking around the corner" feature.

The functions `CameraYaw(..)`, `CameraPitch(..)` and `CameraRoll(..)` take an angle parameter (in radians) which allow precise manipulation.
A `CameraYaw(camera, 90 * DEG2RAD);` call results precisely in a 90 degree turn.

## Behavioural differences to the old rcamera.h
- CAMERA_FREE is no longer a panning camera, but a real 3D free float camera.

## API and usage

<details>
<summary>API</summary>

```C
typedef struct Camera3D {
    Vector3 position;      // Camera position
    Vector3 target;        // Camera target it looks-at
    Vector3 up;            // Camera up vector (rotation around the view axis)
    float fovy;            // Camera field-of-view apperture in Y (degrees) in perspective, used as near plane width in orthographic
    int projection;        // Camera projection: CAMERA_PERSPECTIVE or CAMERA_ORTHOGRAPHIC
    int swingCounter;      // Camera view-bobbing. Set to 0 to deactivate
} Camera3D;

API exposed through raylib.h:
// Update camera position for selected mode
// Camera mode: CAMERA_FREE, CAMERA_FIRST_PERSON, CAMERA_THIRD_PERSON, CAMERA_ORBITAL or CUSTOM
void UpdateCamera(Camera3D *camera, int mode);

API exposed through rcamera.h:
// Returns the cameras forward vector (normalized)
Vector3 GetCameraForward(Camera3D* camera);
// Returns the cameras up vector (normalized)
Vector3 GetCameraUp(Camera3D* camera);
// Returns the cameras right vector (normalized)
Vector3 GetCameraRight(Camera3D* camera);
// Moves the camera in its forward direction
void CameraMoveForward(Camera3D *camera, float distance);
// Moves the camera in its up direction
void CameraMoveUp(Camera3D *camera, float distance);
// Moves the camera target in its right direction
void CameraMoveRight(Camera3D *camera, float distance);
// Moves the camera position closer/farther to/from the camera target
void CameraZoom(Camera3D *camera, float delta);
// Rotates the camera around its up vector
// Yaw is "looking left and right"
// Note: angle must be provided in radians
void CameraYaw(Camera3D *camera, float angle);
// Rotates the camera around its right vector
// Pitch is "looking up and down"
// Note: angle must be provided in radians
void CameraPitch(Camera3D *camera, float angle);
// Rotates the camera around its forward vector
// Roll is "turning your head sideways to the left or right"
// Note: angle must be provided in radians
void CameraRoll(Camera3D *camera, float angle);
// Moves camera slightly to simulate a walking motion
// Note: Only active if camera->swingCounter > 0
void CameraViewBobbing(Camera3D* camera);
// Returns the camera view matrix
Matrix GetCameraViewMatrix(Camera3D* camera);
// Returns the camera projection matrix
Matrix GetCameraProjectionMatrix(Camera3D *camera, float aspect);
```

</details>

<br>

Usage
```C
Camera camera = { 0 };
camera.position = (Vector3){ 0.0f, 2.0f, 4.0f };
camera.target = (Vector3){ 0.0f, 2.0f, 0.0f };
camera.up = (Vector3){ 0.0f, 1.0f, 0.0f };
camera.fovy = 60.0f;
camera.projection = CAMERA_PERSPECTIVE;

Using raylib.h (raylib internally):
// This function will handle input and move/turn accordingly.
UpdateCamera(&camera, CAMERA_FIRST_PERSON);

Using rcamera.h (ex. in a different engine):
// Camera movement
if (IsKeyDown(KEY_W)) CameraMoveForward(camera, CAMERA_MOVE_SPEED, true);
if (IsKeyDown(KEY_S)) CameraMoveForward(camera, -CAMERA_MOVE_SPEED, true);
if (IsKeyDown(KEY_D)) CameraMoveRight(camera, CAMERA_MOVE_SPEED, true);
if (IsKeyDown(KEY_A)) CameraMoveRight(camera, -CAMERA_MOVE_SPEED, true);

// Camera rotation
Vector2 mousePositionDelta = GetMouseDelta();
CameraYaw(camera, mousePositionDelta.x * CAMERA_MOUSE_MOVE_SENSITIVITY, false);
CameraPitch(camera, mousePositionDelta.y * CAMERA_MOUSE_MOVE_SENSITIVITY, true, false, false);
```
  
## TODO
- [x] Core functionality + CAMERA_FREE
- [x] Camera modes CAMERA_FIRST_PERSON, CAMERA_THIRD_PERSON, CAMERA_ORBITAL
- [x] Adapt all examples
- [x] View bobbing
- [ ] CAMERA_ORBITAL autorotation
- [ ] Frame time compensation
- [x] New input system ?
- [x] Review compatibility with old rcamera.h usage
- [ ] Review stand alone mode
- [ ] Check all examples for correct functionality/info boxes

## Feedback

I'd much appreciate feedback!

If you want to have a look at the code or the demo, you can checkout my feature branch [rcamera_redesign_vector](https://github.com/Crydsch/raylib/tree/rcamera_redesign_vector) or this PR..

The `core_3d_camera_first_person` example is functional and can be used for testing.

Note: The camera struct was extended by the camera `mode` and `swingCounter` (for view bobbing).
This should not be a breaking change, as the struct is initialized by `Camera camera = { 0 };` 
and the camera mode is set with `SetCameraMode(..);`, both which always produce a valid state.

Edit: Updated API and usage to reflect current iteration (for quick overview)


raysan5:
@Crydsch Thanks for the redesign, I like this approach more than previous one, less breaking changes and a code simplification. Adding the camera.mode was also considered in the past. Removing the globals in camera is fantastic!

I had problems to switch between camera modes without keeping some of those globals but I think it should be possible to do so without them... I just was not able to do it in the past.

I've been also doing some changes to `rcamera.h` lately because I needed them for a project, I basically implemented `GetFrameTime()`-based movements and simplified camera waiving and swinging (aka "bobbing"). For the moment I'm not pushing those changes, they still require some review... actually I find camera rotations implementation quite complex to follow.

As always, I prioritize code simplicity, in case someone in the future needs to review this API.

raysan5:
@Crydsch Amazing work! I need some time to review it properly and give you a better answer.

About the functions `SetCamera*Controls()` they can be removed. I don't know if anyone ever used them.

Crydsch:
@raysan5 Sorry for pushing so quickly. Please take your time for the review, I just wanted to keep working on other parts in the meantime.
Also I had to force-push because I re-based to the upstream master.

### View Bobbing

While trying to design the new input system I noticed that the view bobbing
needs to be investigated first. Especially since it needs to store some state to
produce the frame by frame animation.

Previously this was done with the `int swingCounter` static variable.
I kept this approach, but added it to the camera struct as we want to remove any static data.
This is no breaking change for the same reasons as with the addition of the camera mode.
The camera struct is normally initialized as `Camera3D camera = {0};`, which sets a valid state.
And typically followed by `SetCameraMode(&camera, CAMERA_*);`, which also enables or disables the view bobbing.

The view bobbing is disabled if `swingCounter == 0` and can be enable by setting it to 1.
If enabled the counter is increased every frame and produces the animation.
It is automatically enabled with a call to `SetCameraMode(&camera, CAMERA_FIRST_PERSON);` and disabled for all other camera modes.
This provides ease of use and sane defaults, while still giving the user full control.

I kept the core behaviour as I am not to familiar with view bobbing.
Also I may not be the best to optimize the behaviour, as I personally get motion sick quickly and typically disable it in games I play.
Therefore the function `void CameraViewBobbing(Camera3D* camera)` may require some review by someone more familiar with it.


### Adjusting all examples

The previous change from `SetCameraMode(camera, CAMERA_*);` to `SetCameraMode(&camera, CAMERA_*);` needed to be adjusted in all other examples to resolve the compilation issue.
The CI should now build everything correctly again :)
I started and checked every one and made sure they work.

Currently I adjusted the camera_first_person example to showcase all camera modes.
We could keep such a combined camera example (an remove the others) or revert to one example per mode.
Which do you prefer?

Some minor fixes were/are necessary in:
- `models_mesh_picking` allow toggling camera controls.
- `core_3d_picking` allow toggling camera controls.
- `core_3d_camera_free` info box text explaining controls. (not done yet)

### New Input System ?

This comment got quite long, but I felt like explaining my thought process.
Therefore I moved it to the discussion #2507 [here](https://github.com/raysan5/raylib/discussions/2507#discussioncomment-3170531).

GoodNike:
Please consider using delta time between frames for camera position calculation.

For now camera speed depends on target FPS, which may be a problem if FPS is adjustable or not stable.

Crydsch:
@GoodNike 
Thanks for mentioning it.
Yes, frame time compensation is planed. I added it to the todo list.

@raysan5 
It's been a while since my last pushes, did you have time to review it?
Also I would like to hear what you think about the necessity of an additional input system
(on top of the low-level API).

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+    
+    return MatrixIdentity();
+}
 
-// Set camera move controls (1st person and 3rd person cameras)
-void SetCameraMoveControls(int keyFront, int keyBack, int keyRight, int keyLeft, int keyUp, int keyDown)
+
+#ifndef CAMERA_STANDALONE
+
+static int init_frames = 3; // TODO review and remove
```
raysan5, line 473: This is a global variable.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
-    CAMERA.moveControl[MOVE_LEFT] = keyLeft;
-    CAMERA.moveControl[MOVE_UP] = keyUp;
-    CAMERA.moveControl[MOVE_DOWN] = keyDown;
+    // Avoid inital mouse "jump"
+    if (init_frames > 0) {
+        SetMousePosition(0, 0);
+        init_frames--;
+        return;
+    }
+    Vector2 mousePositionDelta = GetMouseDelta();
```
raysan5, line 482: Input states can be read and stored at the beginning of the function to avoid spreading the raylib-dependand functions all around the code.

raysan5:
@Crydsch Just pushed the latest camera changes I was working on:

 - Reviewed camera swinging (up-down movement)
 - Reviewed camera tilting (left-right movement)
 - Make movement independent of frame-rate
 - Removed unneeded variables

It's been about a month but I didn't push them because there is still some issue on the camera rotation for the first-person view, it moves too fast when fixed to 60 fps, the framerate independant movement is not compretely right... in any case I pushed those changes in case they can be helpful for you as a reference.

About all your changes:

## New signature for `SetCameraMode(&camera, mode)`

This is a breaking change and I'm still thinking about the best way to approach it... I understand that from a user point of view it's more intuitive to store the camera mode inside the camera structure but from a pure data point of view, mode only affects how the camera values are internally calculated in the next update. Think about it this way: if we change camera.mode but we don't call `CameraUpdate()`, nothing changes in the camera, actually, we can manually setup a custom camera independently of the mode stored. Mode only defines `CameraUpdate()` behaviour.

Still, I think a `SetCameraMode(&camera, mode)` is more intuitive than `SetCameraMode(mode)`, that should probably be the right approach... Maybe it can just be replaced by a `SetCameraUpdateMode(mode)` or moved to `UpdateCamera(&camera, mode)`? Actually, I like that last function, it just passes to `UpdateCamera()` the required mode for calculation and camera remains agnostic of mode.

Did you try changing between modes at runtime? Does it work as expected? I remember having some issues with it...

## New Input System

Personally I prefer option 1 for now, wrap raylib specific input handling in an `#if !defined(CAMERA_STANDALONE)`. 

About the keys to use, I would set some default ones as defines and let the user configure them recompiling the module. I would remove all external individual keys configuration and try to find a better solution for keys configuration... 

Functions to remove/replace:
```c
void SetCameraPanControl(int keyPan);                   // Set camera pan key to combine with mouse movement (free camera)
void SetCameraAltControl(int keyAlt);                   // Set camera alt key to combine with mouse movement (free camera)
void SetCameraSmoothZoomControl(int szoomKey);          // Set camera smooth zoom key to combine with mouse (free camera)
void SetCameraMoveControls(int keyFront, int keyBack,
                           int keyRight, int keyLeft,
                           int keyUp, int keyDown);     // Set camera move controls (1st person and 3rd person cameras)
```

In any case, I would keep a global structure for the input values that could be updated by the user if required, it can be just an `int inputs[]` with all the positions `#defined` to be updated like `SetCameraControl(CAMERA_FIRST_PERSON_MOVE_FORDWARD, KEY_W)`.

It's not easy to find a good solution... there could be many types of controls involved and supporting all the options (keyboard, mouse, gamepad, touch...) it's almost impossible. Only the most common inputs that users usually wants to configure should be chosen, most of them could be probably just kept inside as `#defines`.

## Camera swinging and tilting

I just pushed a commit with some improvements and clearer comments (and names) about how they work. About exposing those values to be configured by the user, maybe a similar approach to keys input can be used: `SetCameraParams(CAMERA_FIRST_PERSON_MOVE_TILTING, 80)`. In any case, I would keep them internal to camera module because they are not part of the camera data, only define how next camera values are calculated (like the camera.mode).

I think that's it for now. Thank you very much again for all your hard work in this big improvement of the library.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+    CameraYaw(camera, mousePositionDelta.x * -CAMERA_MOUSE_MOVE_SENSITIVITY);
+    CameraPitch(camera, mousePositionDelta.y * -CAMERA_MOUSE_MOVE_SENSITIVITY);
+
+    // Zoom target distance
+    CameraZoom(camera, -GetMouseWheelMove());
+    if (IsKeyPressed(KEY_KP_SUBTRACT)) CameraZoom(camera, 2.0f);
+    if (IsKeyPressed(KEY_KP_ADD)) CameraZoom(camera, -2.0f);
+
+    // Switch camera mode
+    if (IsKeyPressed(KEY_ONE)) SetCameraMode(camera, CAMERA_FREE);
```
raysan5, line 509: This is up to the user, it  shouldn't be considered inside camera.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+    if (IsKeyPressed(KEY_TWO)) SetCameraMode(camera, CAMERA_FIRST_PERSON);
+    if (IsKeyPressed(KEY_THREE)) SetCameraMode(camera, CAMERA_THIRD_PERSON);
+    if (IsKeyPressed(KEY_FOUR)) SetCameraMode(camera, CAMERA_ORBITAL);
+
+    // Switch camera projection
+    if (IsKeyPressed(KEY_P)) {
```
raysan5, line 515: This is up to the user, it shouldn't be considered inside camera.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+            camera->position = (Vector3){ 0.0f, 2.0f, 10.0f };
+            camera->target = (Vector3){ 0.0f, 2.0f, 0.0f };
+            camera->up = (Vector3){ 0.0f, 1.0f, 0.0f };
+            camera->projection = CAMERA_PERSPECTIVE;
+            camera->fovy = 60.0f;
+        }
+    }
+
+    // Apply view bobbing when moving around (per default only active in CAMERA_FIRST_PERSON)
+    if (IsKeyDown(KEY_W) || IsKeyDown(KEY_A) || IsKeyDown(KEY_S) || IsKeyDown(KEY_D)) CameraViewBobbing(camera);
```
raysan5, line 542: Tilting and swinging is mode dependant (first person camera) and setting the tilting and swinging to 0 should be enough, it shouldn't depend on the inputs, I mean, it's dependant on the movement.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+        camera->target.y = camera->target.y - 0.25f * sinf((camera->swingCounter - 1) / CAMERA_FIRST_PERSON_STEP_TRIGONOMETRIC_DIVIDER) / CAMERA_FIRST_PERSON_STEP_DIVIDER;
+
+        // Update counter for next frame
+        camera->swingCounter %= 2147483647 /* INT_MAX */; // Counter must be positive
+        camera->swingCounter++;
+    }
+}
+
+// Returns the camera view matrix
+Matrix GetCameraViewMatrix(Camera3D *camera)
```
raysan5, line 448: Probably no need for this function. It adds a dependency to `raymath` and  users can already get the matrix with the camera exposed parameters.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
+{
+    return MatrixLookAt(camera->position, camera->target, camera->up);
+}
+
+// Returns the camera projection matrix
+Matrix GetCameraProjectionMatrix(Camera3D *camera, float aspect)
```
raysan5, line 454: Probably no need for this function. It adds a dependency to `raymath` and  users can already get the matrix with the camera exposed parameters.

src/rcamera.h
```diff
@@ -44,15 +45,26 @@
 #ifndef RCAMERA_H
 #define RCAMERA_H
 
+// The only dependency // TODO review standalone mode
```
raysan5, line 48: I'm seeing that the dependency on `raymath` is really tight now... that could be a problem and something to review carefully...

src/raylib.h
```diff
@@ -306,6 +306,8 @@ typedef struct Camera3D {
     Vector3 up;             // Camera up vector (rotation over its axis)
     float fovy;             // Camera field-of-view apperture in Y (degrees) in perspective, used as near plane width in orthographic
     int projection;         // Camera projection: CAMERA_PERSPECTIVE or CAMERA_ORTHOGRAPHIC
+    int mode;               // Camera mode: CAMERA_FREE, CAMERA_FIRST_PERSON, CAMERA_THIRD_PERSON, CAMERA_ORBITAL or CUSTOM
```
raysan5, line 309: As commented, `mode` and `swingCounter` can be avoided from the struct.

src/raylib.h
```diff
@@ -1150,7 +1152,7 @@ RLAPI float GetGesturePinchAngle(void);                 // Get gesture pinch ang
 //------------------------------------------------------------------------------------
 // Camera System Functions (Module: rcamera)
 //------------------------------------------------------------------------------------
-RLAPI void SetCameraMode(Camera camera, int mode);      // Set camera mode (multiple camera modes available)
+RLAPI void SetCameraMode(Camera* camera, int mode);      // Set camera mode (multiple camera modes available)
```
raysan5, line 1155: As commented, `SetCameraMode()` can be just removed and its functionality moved to `UpdateCamera(&camera, mode)`.

raysan5:
@Crydsch I'm checking the code more carefully and I see no quaternions are used for camera rotations, it keeps using the same previous Euler maths but now though the raymath functions...

src/rcamera.h
Crydsch: It will be removed as per comment. I am not sure what causes this, but when starting an example the intial call(s) to GetMouseDelta() report very high values, which caused the camera to jerk around in the first frames.
This may be a bug.
This `init_frames' ist just my quick fix, to allow proper testing - it will be removed.

src/rcamera.h
Crydsch: Makes sense to me - I'll change this!

src/rcamera.h
Crydsch: Yes, this will be changed. The `UpdateCamera(..)` function is currently in a testing version. It is like this to allow testing all functionality of the camera module in the example. (Switching modes, moving around, changing projection mode, ...).
The final version will consider the current camera mode and frame time compensation as well. 

I was waiting for a general approval of the API and all, before continuing.

src/rcamera.h
Crydsch: Yes, this will be changed.

src/rcamera.h
Crydsch: The `CameraViewBobbing(..)` function does nothing if the `swingCounter` is set 0.
It is only activated (set to 1) in the first person mode.

If we remove the mode variable anyway though, and just pass it as a parameter to `UpdateCamera(..)`, then this problem is solved too.

src/rcamera.h
Crydsch: This function was meant for convenience reasons and for the stand alone mode.
Anyway it can be removed and users just have to call `raymath` themselves.

src/rcamera.h
Crydsch: This function was meant for convenience reasons and for the stand alone mode.
Anyway it can be removed and users just have to call raymath themselves.

src/rcamera.h
Crydsch: Right, we agreed to start out depending on `raymath` and I think the code is much more readable this way.
Further, without the code duplication any bug fix in `raymath` also benefits the camera module.

Alternatively, we can duplicate the relevant functions and become independent again - at the cost of duplication though.

src/raylib.h
Crydsch: The `mode` will be removed and added as a parameter to `UpdateCamera(..)`.

I am not sure how to remove `swingCounter` though?
For the view bobbing we need to store some state, right? Are there other algorithms to achieve the same effect without state?

src/raylib.h
Crydsch: Will be done. :)

Crydsch:
@raysan5 The quaternion approach is fundamentally different and would have required a different camera struct (as per the first PR).
Calculating the current quat every frame, just to implement quat-based rotations and then reverting back and updating the vectors just does not make any sense. The code would be unnecessary complex and we would gain nothing from it.

But we can still improve the camera module with the low-level API, general API improvements, removing the static data, enabling the stand alone mode, ...

raysan5:
> Calculating the current quat every frame, just to implement quat-based rotations and then reverting back and updating the vectors just does not make any sense. The code would be unnecessary complex and we would gain nothing from it.

Agree.

Crydsch:
@raysan5 

I had a look at the change regarding `camera->mode` and found some problems.

1. SetCameraMode(..) includes other logic than just setting `camera->mode`.
a) For one it calls `DisableCursor();` to retain compatibility with the old camera module functionality. Of course we can shift this responsibility to the user. This makes especially sense, because such a function to catch the cursor is window-system/engine specific.
b) It resets the up vector (`roll`) to help switching between camera modes.
As with 1.a we can shift this responsibility to the user. Changing between camera modes is probably seldom used anyway. 
I guess most users set one mode at the beginning and never change it. If so, they would have to reset `roll` themselves.

2. `camera->mode` is not only used by `UpdateCamera(..)` but also by several low-level functions.
We could solve this by passing it to those functions as well, it does clutter the API a bit though.


Should we continue with the change anyway?

raysan5:
@Crydsch I see the issue with the `camera.mode`, some modes require some specific initialization, it's the same issue we have with current implementation. Probably that's unavoidable...

Which low-level functions depend on the `camera.mode`? I think they shouldn't depend on it, that's why they are low-level functions...

Crydsch:
@raysan5 

The low-level functions requiring`camera.mode` are:

- `CameraMoveForward(..)` and `CameraMoveRight(..)`
  In the modes `CAMERA_FIRST_PERSON` and `CAMERA_THIRD_PERSON` we need to project the forward vector onto the world plane.
  We do not want to move downwards when the player is looking down.
 
- `CameraYaw(..)`
  In the modes `CAMERA_FREE` and `CAMERA_FIRST_PERSON` we need to rotate the target around the position. 
  Whereas in `CAMERA_THIRD_PERSON ` and `CAMERA_ORBITAL` we need to rotate the position around the target.

- `CameraPitch(..)`
  When pitching we need to clamp the pitch angle in the modes `CAMERA_FIRST_PERSON`, `CAMERA_THIRD_PERSON` and `CAMERA_ORBITAL` to lock the view vector when locking straight "up" or "down".
  Further, we need to apply the rotation as in `CameraYaw(..)`.

As you can see, there are some dependencies on `camera.mode`, because the underlying math differs based on the current mode. 
I don't think it can be avoided.

raysan5:
@Crydsch I think those functions shouldn't depend on `camera.mode`, in any case, they could include some additional config parameter to accomodate on calling, depending on the mode (checked by caller).

For example:
```c
void CameraMoveForward(Camera3D *camera, float distance, Vector3 axisLock);
void CameraPitch(Camera3D *camera, float angle, bool lockView);
```

Crydsch:
@raysan5 That might be a possible fix. I'll have a look into those functions. With that change we can then remove `camera.mode` from the struct and pass it as an argument to `UpdateCamera(..)`.

That would also mean that `SetCameraMode(..)` becomes obsolete, if we assume it's the users responsibility to reset `camera.up` when changing camera modes and calling `DisableCursor()` initially.
I think both those things are reasonable. Especially since `DisableCursor()` is engine/window system specific.

Crydsch:
Hey @raysan5 

Sorry it has been a while since I worked on this. The semester ended and I was on vacation.
But I found some time and worked on the changes we talked about.


The low-level functions now have additional parameters and no longer rely on `camera.mode` and it has been remove from the struct altogether.
I am a bit unsure about the variable naming. If you would like something different, please just let me know.
The mode is now passed to `UpdateCamera(..)`. With this, `SetCameraMode(..)` was also removed.

While adjusting the examples to use the new `UpdateCamera(..)` function I also changed a few camera modes to something different than CAMERA_FREE where I felt like it gives a better experience, since CAMERA_FREE is now a real free float and just running an example should be easy (whereas the handling of CAMERA_FREE can be a bit confusing).

Looking at your review again, you mentioned removing the `Get*Matrix` functions, since in raylib the user can always use the same fucntion in raymath.
Still I think it makes sense to have them in the camera module logically, since I would search there first.
Also this may be very usefull in the stand-alone mode. (Even though raymath would be present then too...).
Should we remove or keep them?


Also I rebased onto master again.

PS: I found a new example missing from the VS2022 project ;)
    `models/models_loading_m3d`

raysan5:
@Crydsch Excuse me for this super-late response, it's a big change and I it requires time to review but I can't find the moment to jump into it... 

Taking a quick look, it seems and amazing improvement! It not only simplifies the API a lot but also opens a door for future improvements. I'll try to review it more carefully as soon as possible.

Thank you very much for all the work put into it!

examples/core/core_3d_camera_first_person.c
```diff
@@ -29,11 +30,14 @@ int main(void)
...
-    camera.position = (Vector3){ 4.0f, 2.0f, 4.0f };
-    camera.target = (Vector3){ 0.0f, 1.8f, 0.0f };
+    camera.position = (Vector3){ 0.0f, 2.0f, 4.0f };
+    camera.target = (Vector3){ 0.0f, 2.0f, 0.0f };
     camera.up = (Vector3){ 0.0f, 1.0f, 0.0f };
     camera.fovy = 60.0f;
     camera.projection = CAMERA_PERSPECTIVE;
+    camera.swingCounter = 1; // Enable view bobbing
+    
+    int cameraMode = CAMERA_FIRST_PERSON;
```
JeffM2501, line 40: Recommend that you use the enum `CameraMode `and not an int. This helps people understand that it is an enum, and makes porting to C++ easier.

src/raylib.h
```diff
@@ -304,9 +304,10 @@ typedef struct Font {
 typedef struct Camera3D {
     Vector3 position;       // Camera position
     Vector3 target;         // Camera target it looks-at
-    Vector3 up;             // Camera up vector (rotation over its axis)
+    Vector3 up;             // Camera up vector (rotation around the view axis)
     float fovy;             // Camera field-of-view apperture in Y (degrees) in perspective, used as near plane width in orthographic
     int projection;         // Camera projection: CAMERA_PERSPECTIVE or CAMERA_ORTHOGRAPHIC
+    int swingCounter;       // Camera view-bobbing. Set to 0 to deactivate.
```
JeffM2501, line 310: while I understand moving this here stores the data relative to the camera and not in a global, I feel that people are not going to understand setting a count to 0 as disable (even with the comment). It may be worth having a viewBob enabled Boolean just to ease confusion.

src/raylib.h
```diff
@@ -1151,13 +1152,7 @@ RLAPI float GetGesturePinchAngle(void);                 // Get gesture pinch ang
...
 // Camera System Functions (Module: rcamera)
 //------------------------------------------------------------------------------------
-RLAPI void SetCameraMode(Camera camera, int mode);      // Set camera mode (multiple camera modes available)
-RLAPI void UpdateCamera(Camera *camera);                // Update camera position for selected mode
-
-RLAPI void SetCameraPanControl(int keyPan);             // Set camera pan key to combine with mouse movement (free camera)
-RLAPI void SetCameraAltControl(int keyAlt);             // Set camera alt key to combine with mouse movement (free camera)
-RLAPI void SetCameraSmoothZoomControl(int keySmoothZoom); // Set camera smooth zoom key to combine with mouse (free camera)
-RLAPI void SetCameraMoveControls(int keyFront, int keyBack, int keyRight, int keyLeft, int keyUp, int keyDown); // Set camera move controls (1st person and 3rd person cameras)
+RLAPI void UpdateCamera(Camera3D *camera, int mode);              // Update camera position for selected mode
```
JeffM2501, line 1155: Recommend use the enum here and not an int.

src/rcamera.h
```diff
@@ -107,17 +115,19 @@
...
-void SetCameraPanControl(int keyPan);                       // Set camera pan key to combine with mouse movement (free camera)
-void SetCameraAltControl(int keyAlt);                       // Set camera alt key to combine with mouse movement (free camera)
-void SetCameraSmoothZoomControl(int szoomKey);              // Set camera smooth zoom key to combine with mouse (free camera)
-void SetCameraMoveControls(int keyFront, int keyBack,
-                           int keyRight, int keyLeft,
-                           int keyUp, int keyDown);         // Set camera move controls (1st person and 3rd person cameras)
-#endif
+Vector3 GetCameraForward(Camera3D* camera);
+Vector3 GetCameraUp(Camera3D* camera);
+Vector3 GetCameraRight(Camera3D* camera);
```
JeffM2501, line 120: Recommend that these be GetCameraForwardVector, GetCameraUpVector, and GetCameraRightVector for clarity.
Also remove the whitespace so all 3 vector commands are together

src/rcamera.h
```diff
@@ -107,17 +115,19 @@
...
+void CameraMoveForward(Camera3D* camera, float distance, bool moveInWorldPlane);
+void CameraMoveUp(Camera3D* camera, float distance);
+void CameraMoveRight(Camera3D* camera, float distance, bool moveInWorldPlane);
+void CameraZoom(Camera3D* camera, float delta);
```
JeffM2501, line 124: I'm not a fan of this name, I understand why it's named this because it does move closer to the target, but often a Zoom is considered a change in FOV and this function does not change the FOV. This is a move command so it should have the word in it. Maybe `CameraMoveToTarget`, or something.

src/rcamera.h
```diff
@@ -107,17 +115,19 @@
...
+void CameraYaw(Camera3D* camera, float angle, bool rotateAroundTarget);
+void CameraPitch(Camera3D* camera, float angle, bool lockView, bool rotateAroundTarget, bool rotateUp);
+void CameraRoll(Camera3D* camera, float angle);
+void CameraViewBobbing(Camera3D* camera);
```
JeffM2501, line 128: If the user has to call this on it's own to get bobbing and not part of update, I don't think the bob state should be part of the camera itself. Perhaps a separate structure for camera state that needs to be passed in.
Really the '0' to disable on the sway count is not correct, what disables bobbing is not calling this function.
Because of this I would not couple the bob state to the camera, since it's only used by this function. It's extra data that other camera systems don't need. Bobbing is such a minor feature I think that people who want it should have to track it's state separately.

tedajax:
Hello, just reading through and finding this PR interesting because I've found the default camera behavior to be wanting for many of the reasons this PR exists.

I would just like to throw in my 2 cents that the entire view bobbing feature should just be removed or at least explicitly be an opt-in option for the camera. I would never expect in any engine or toolkit to switch to a default first person camera and get a view bobbing as part of that, especially a view bob that isn't readily configurable or quick to disable.

nilsbecker:
Just to leave a note here: In the previous predefined cameras, the baked-in mouse buttons assigned to the different motions made it impossible e.g. to pan on macbooks with a touchpad. the new implementation should make sure to 1. choose compatible defaults 2. make buttons and modifier keys configurable in the predefined cameras.

src/rcamera.h
```diff
@@ -134,434 +148,401 @@ void SetCameraMoveControls(int keyFront, int keyBack,
...
-void SetCameraAltControl(int keyAlt) { CAMERA.altControl = keyAlt; }
+    // Rotate up direction around forward axis
+    camera->up = Vector3RotateAxis(camera->up, forward, angle);
+}
 
-// Set camera smooth zoom key to combine with mouse (free camera)
-void SetCameraSmoothZoomControl(int szoomKey) { CAMERA.smoothZoomControl = szoomKey; }
+// Moves camera slightly to simulate a walking motion
+// Note: Only active if camera->swingCounter > 0
+void CameraViewBobbing(Camera3D* camera)
```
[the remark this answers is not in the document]
raysan5: It will be removed after merge and review

src/rcamera.h
raysan5: I like `CameraMoveToTarget()`

raysan5:
Merged new camera redesign, reviewing it carefully in following commits...

PR #2567, opened 2022-07-07
Fix fat universal (arm64 + x86_64) macos Github Actions build

electronstudio:
It now builds universal macos binaries.  However I am not able to test if they work!

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -53,16 +53,36 @@ jobs:
       run: |
         cd src
         clang --version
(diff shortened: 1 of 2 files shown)


raysan5:
@electronstudio Thank you very much for the improvement! Does it also work on arm64?

benbeshara:
@raysan5 @electronstudio Works fine on my M1 Mac!

raysan5:
@electronstudio thanks for the improvement! 👍😄

PR #2569, opened 2022-07-09
Allow DLL creation using TCC

audinue:
This fixes the following errors during DLL creation:

```
raylib.h:918: error: declaration for parameter 'InitWindow' but no such parameter
```

TCC uses `__attribute__` instead of `__declspec`

```
#if defined(__TINYC__)
    #include <_mingw.h>
#endif
#define RLAPI __declspec(dllexport)
```

`#include <_mingw.h>` is inspired by `windows.h`

```
tcc: error: unsupported linker option '--out-implib,../src/libraylibdll.a'

tcc: error: invalid option -- '-static-libgcc'

tcc: error: undefined symbol '_DragQueryFileW@16'
tcc: error: undefined symbol '_DragQueryPoint@8'
tcc: error: undefined symbol '_DragFinish@4'
tcc: error: undefined symbol '_DragAcceptFiles@8'
```

Fixed by adding conditional `ifeq ($(CC), tcc)` and adjusting `LDLIBS` and `LDFLAGS` appropriately in the `Makefile`.

I tested the generated DLL with `raylib-master\examples\models>tcc -L..\..\src -lraylib -I..\..\src -run models_animation.c` and it seems working fine.

Please note that I changed `CC` and `AR` to TCC on my machine and reverted it back to GCC before commiting. It might be useful if we add a flag or an override to choose which compiler to be used. I need you to decide that.

Thank you.


src/raylib.h
```diff
@@ -86,6 +86,9 @@
 // NOTE: Microsoft specifiers to tell compiler that symbols are imported/exported from a .dll
 #if defined(_WIN32)
     #if defined(BUILD_LIBTYPE_SHARED)
+        #if defined(__TINYC__)
+            #include <_mingw.h>
```
raysan5, line 90: Sorry, in `raylib.h` I try to avoid any non-standard include.

audinue: Should we inline it like this?
```c
#if defined(__TINYC__)
    #define RLAPI __attribute__(dllexport)
#else
    #define RLAPI __declspec(dllexport)
#endif
```

raysan5: yes, way better

audinue: Done :smile:

audinue: Wait! There is some errors

audinue: `RLAPI` is redefined in `rlgl.h`

Thus I changed the declaration in `raylib.h` to be

```c
#if defined(__TINYC__)
    #define __declspec(x) __attribute__((x))
#endif
#define RLAPI __declspec(dllexport)
```

Tested with `make clean && make RAYLIB_LIBTYPE=SHARED` and running some examples and everything works just fine now.

Thank you.

raysan5:
@audinue I'm merging this improvement for now but not sure how many raylib users really require TINYCC, it seems a very specific fix that pollutes the header and Makefile.

PR #2574, opened 2022-07-11
Add front/back culling toggling to `rlgl.h`

GithubPrankster:
This enables a few techniques such as rendering inverted hull outlines with more simplicity.


raysan5:
@GithubPrankster I'm afraid I'm not adding those functions, anyone requiring them can use OpenGL calls directly.

PR #2575, opened 2022-07-11
Ability to assign a new key for screen capture

hanaxar:
Instead of hardcoding the screen capture key, it's assigned to a variable, and also added 2 simple get/set functions.


raysan5:
@hanaxar There is no need to add new functions. You can just disable screenshots in config and use already provided `TakeScreenhsot()` function.

Issue #2228, opened 2021-12-18
[audio] Music looping based on FPS issue


Kabcorp:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Music looping is based on FPS, which is odd in 99% of time.
With this approach, if the music reach the end of its own length, you have to wait the next frame to restart looping.

It was tried of course using WAV pcm file with basic code:

### Environment

Windows 10, mingw_w64, GTX1050 3gB
*Provide your Platform, Operating System, OpenGL version, GPU details where you experienced the issue.*

### Issue Screenshot

![image](https://user-images.githubusercontent.com/33150965/146653181-c5abb297-8ccf-4dda-bda1-61fbf8f5e5a0.png)
![image](https://user-images.githubusercontent.com/33150965/146653145-3ce3d77e-6bad-4511-a9c8-19e89a978453.png)

### Code Example

(sfx.wav is downloadable here: [http://kabcorp.fr/sendfiles/ray/sfx.wav](http://kabcorp.fr/sendfiles/ray/sfx.wav))
sfx.wav is a pure sine wave with correct repeat sample.

```C
#include <raylib.h>

int main(void)
{
    int Fps=30;
    InitWindow(700, 200, "Music Loop Issue");
    SetTargetFPS(Fps);

    InitAudioDevice();

    Music Sfx=LoadMusicStream("sfx.wav");
    PlayMusicStream(Sfx);

    while (!WindowShouldClose())
    {
            UpdateMusicStream(Sfx);

            Fps+=(IsKeyPressed(KEY_RIGHT)-IsKeyPressed(KEY_LEFT))*10;
            Fps=Fps < 10 ? 10 : Fps;
            SetTargetFPS(Fps);

            BeginDrawing();
            ClearBackground(GRAY);

            DrawText(TextFormat("(Press L/R to change FPS)\n%i/%i fps", GetFPS(), Fps), 8, 8, 20, LIGHTGRAY);

            EndDrawing();
    }

    CloseWindow();
    UnloadMusicStream(Sfx);
    return 0;
}
```

raysan5:
@Kabcorp Yes, I'm aware of this issue, `raudio` module needs some improvements on that regards. There are some open PRs on that line but not completely reviewed yet.

massung:
[...]

PR #2579, opened 2022-07-16
Audio looping fix

veins1:
An incomplete (non)fix for #2228. Handling only WAVs. Logic is the same for other formats but I didn't want to copy-paste, there's probably a better solution. First PR in my life!


veins1:
Decided to fix other formats by copy-pasting. Maybe it's not too bad. I believe all formats work correctly now. Can't really test short XM and MOD files ( I don't know where to get them), but full songs loops work.

raysan5:
@veins1 thanks for the review. I tested it with the provided VS2022 project (MSVC compiler) and compilation failed with multiple errors, related to the `void *` casting while reading the audio data from file, maybe you tested it compiling as C++? what compiler did you use?

I reviewed the lines and when playing there is no stuttering but I can hear some cracks, did you experienced the same issue?

Also note the PR is not following the [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md), please, could you review it?

raysan5:
@veins1 I'm merging this fix as a temporal solution but the full audio system should actually be redesigned, as noted in the related issue.

PR #2587, opened 2022-07-22
Update README.md

andsiu:
Simple change of
- "specially" -> "especially" when appropriate
- "fileformats" -> "file formats"

No reference issue because problem is too small to justify making a reference for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **raylib is a simple and easy-to-use library to enjoy videogames programming.**
 
-raylib is highly inspired by Borland BGI graphics lib and by XNA framework and it's esspecially well suited for prototyping, tooling, graphical applications, embedded systems and education.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 **raylib is a simple and easy-to-use library to enjoy videogames programming.**
 
-raylib is highly inspired by Borland BGI graphics lib and by XNA framework and it's specially well suited for prototyping, tooling, graphical applications, embedded systems and education.
+raylib is highly inspired by Borland BGI graphics lib and by XNA framework and it's esspecially well suited for prototyping, tooling, graphical applications, embedded systems and education.
```
raysan5, line 5: I prefer to keep specially in this case:

> the word 'specially' refers to a special or unique purpose, while 'especially' implies an exceptional quality or characteristic. The latter is also slightly more formal than the former.

raysan5:
@andsiu thanks for the review!

orcmid:
reviewing the word specially altogether is better, since well-suited says it all.  At the same time, having this be in Ramon's own voice is nice.

PR #2590, opened 2022-07-24
[raymath] add Vector3RotateByAxisAngle

Crydsch:
While working on the rcamera redesign I noticed that there was no function in `raymath.h` to rotate a vector around an axis in 3D.

This PR adds the function `Vector3RotateByAxisAngle(Vector3 v, Vector3 axis, float angle)` to do just this.
Maybe this can be useful in other applications as well.

I tried to follow the `raymath.h` conventions and general code style.


raysan5:
@Crydsch This is a nice addition, thanks! Actually I though that function (or similar) was already in there...

PR #2592, opened 2022-07-25
add: cmake config include guard

MyUncle:
raylib does not currently guard against multiple inclusion in cmake projects.

This causes issues when you want to first check for existence of a raylib install and actually add it later.

For example, in my case, I first check for a local install of all my dependencies using find_package('dependency here' QUIET). If not found, I then use FetchContent to automatically grab the missing dependency. Works fine when I DON'T have a local install of raylib. But this causes issues when a local install exists and I attempt the find_package(raylib REQUIRED) call because raylib's config will attempt to add an already existing library ("raylib") when calling find_package on it multiple times.

This change will only run the cmake configuration logic if the raylib target does not already exist.

### DIFF
--- a/cmake/raylib-config.cmake
+++ b/cmake/raylib-config.cmake
@@ -11,67 +11,69 @@
 #   raylib_LDFLAGS - The linker flags needed with raylib
 #   raylib_DEFINITIONS - Compiler switches required for using raylib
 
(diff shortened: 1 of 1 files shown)


raysan5:
@MyUncle Thanks for the improvement! Merged!

Issue #2577, opened 2022-07-11
[textures] `ImageResize()` crashes under certain parameters


zzxoto:
The commit didn't fix the issue. This is reproducable. Try below. Also I have attached the image I have used during testing.

```
#include "stdio.h"
#include "raylib.h"
#include "stdlib.h"

int screenW = 800;
int screenH = 800;

void draw(Image &src, Image &dst)
{
  Rectangle srcRec = { 89.99992, 89.999969, 320.0003, 320.0003 };
  Rectangle dstRec = { 0, 0, 600, 600 };

  ImageDraw(&dst, src, srcRec, dstRec, WHITE);
}

int main(void)
{
  InitWindow(screenW, screenH, "Pleasant paint program");

  Image src = LoadImage("image2_png.png");
  Image dst = ImageCopy(src);
  ImageResize(&dst, 600, 600);

  SetTargetFPS(60);

  while (!WindowShouldClose())
  {
    BeginDrawing();

    ClearBackground(ORANGE);

    draw(src, dst);

    EndDrawing();
  }

  CloseWindow();

  return 0;
}
```

Image: 
https://user-images.githubusercontent.com/24535070/178372387-5e88ea0d-e9de-4fe6-862e-70e1aeaf29e0.png

Library installation info
```
CRTLinkage: dynamic
LibraryLinkage: dynamic
Version: 3ebfee5dbc1d0de5555105f4989954cd4547bafd
```

I am getting error at `void ImageResize(Image *image, int newWidth, int newHeight)
` at line `unsigned char *output = (unsigned char *)RL_MALLOC(newWidth*newHeight*bytesPerPixel);
` and line `RL_FREE(image->data);` following the ` ImageDraw(&dst, src, srcRec, dstRec, WHITE);` call.

Error
```
Critical error detected c0000374

HEAP[main.exe]: Heap block at 000001733BC65180 modified at 000001733BCC91C4 past requested size of 64034
```

raysan5:
@zzxoto A fix was merget, please could you verify it works now?

raysan5:
I'm closing this issue. Feel free to reopen if problem persists.

PR #2594, opened 2022-07-26
rtextures: Fix ImageFromImage crash

wiertek:
Height of the rectangle can be float, which may lead to doing extra iteration of loop and writing out of bounds.

This only fixes problem reported in https://github.com/raysan5/raylib/issues/2577, but there may be another usages of using rect dimensions without casting in other pieces of code.


raysan5:
@wiertek Thanks for the fast review! :)

PR #2596, opened 2022-07-27
rtextures: Improve numerical stability of float multiplication

wiertek:
Dimensions of Rectangle should be casted to int before multiplication,
otherwise there is a risk for underallocation/overallocation of memory.

In case of overallocation its just a waste of memory. In case of underacllocation its a crash in best case.

Example: https://godbolt.org/z/zjhKh7xr8

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -875,7 +875,7 @@ Image ImageFromImage(Image image, Rectangle rec)
 
     result.width = (int)rec.width;
     result.height = (int)rec.height;
-    result.data = RL_CALLOC((int)(rec.width*rec.height)*bytesPerPixel, 1);
+    result.data = RL_CALLOC((int)rec.width*(int)rec.height*bytesPerPixel, 1);
     result.format = image.format;
     result.mipmaps = 1;
 


raysan5:
@wiertek thanks for the review! At some point raylib `Rectangle` values were converted from `int` to `float` and I still regret that design decision some times...

wiertek:
@raysan5  Huh, no worries. 
I dont know much about computer graphics, but I can see that it may be valuable for user to operate on floats for better precision. 
However, from the perspective of library, I dont think it makes sense to "think" about fractional parts of pixel, and the smallest unit will always be pixel. Correct me if I'm wrong :smile: 
If there is a need to keep floats in Rectangle in public API but library always operates on integer Rectangle, then maybe introducing some private struct like RectangleInternal, which will use integers. And then at the beginning of every library function that takes Rectangle, convert to RectangleInternal and operate on that. 
Otherwise every single line which refers to Rectangle in library has some risk if casts to int are forgotten or misaligned.

raysan5:
I prefer to avoid overengineering, the changes of Rectangle crashing code are actually very low and can be avoided by users good practices.

Issue #2595, opened 2022-07-27
[raymath] Result of MatrixRotate() and MatrixRotateX/Y/Z differs


GoodNike:
### Issue description

Resulting matrix of MatrixRotate() and MatrixRotateX/Y/Z differs, watch example code below.

### Code Example

```
Matrix mat1 = MatrixRotate({1, 0 , 0}, PI/4);    /* Rotate around X. */
Matrix mat2 = MatrixRotateX(PI/4);
```

The mat1 matrix is correct. The mat2 matrix fields m6 and m9 have a wrong sign. Same goes to MatrixRotateY() (m2 and m8), MatrixRotateZ() (m1 and m4).

raysan5:
@GoodNike Thanks for reporting this issue! Feel free to send a PR with the fix!

chriscamacho:
isn't -m6 m9 vs m6 -m9 equivalent once converted back to eular rotations ?

there are multiple matrix (and quats) that can result from the same eular angles, depending on method of calculation, comparing matrices is likely to not work very well under different circumstances and probably not great for checking if an orientation is within a specific range.

GoodNike:
@chriscamacho thanks for reply! MatrixRotate rotates counter-clockwise, MatrixRotateX/Y/Z rotates clockwise

We could:
1. do nothing, consider it correct
2. fix it to get same result
3. live a comment explaining things

If those functions intend to simplify MatrixRotate we should do 2.

@raysan5 what do you think?

raysan5:
@GoodNike Agree with **2** and also **3**. I think the rotation should be **clockwise**, it's the most intuitive for raylib users.

The issue now is not only updating `MatrixRotate()` but also check **all the other rotation functions** to make sure they stick to the same convention!

orcmid:
I think "natural" sense of rotation depends on how the coordinate system is anchored.  If the origin is considered upper-left corner, then yes.  If lower-left, (positive ordinates to the right and also up) than anti-clockwise is traditional, at least when I was learning trigonometry :).  

On the other hand, when sailing and looking at a compass, the degrees grow clockwise.  (90 is East, 180 is South, etc.)

I suspect how transformations work should be decisive.  (I am being lazy.)

raysan5:
Afaik, the following functions are the only ones implementing some kind of rotation:
```c
Vector2 Vector2Rotate(Vector2 v, float angle)
Vector3 Vector3RotateByQuaternion(Vector3 v, Quaternion q)
[...]

PR #2599, opened 2022-07-29
[raymath] Rotation functions returns clockwise rotation matrix, #2595

GoodNike:
This commit fixes matrix rotation functions.


GoodNike:
I didn't analyze vector rotations.

I didn't do any additional comments because as i assume its not needed now.

raysan5:
@GoodNike thanks you very much for the review! :)

PR #2602, opened 2022-07-30
removing typo in line 41 line, presssed -> pressed

sDos280:


raysan5:
@sDos280 thanks for the fix!

PR #2603, opened 2022-07-31
removing typo, Colours->Colors

sDos280:


raysan5:
@sDos280 I'm merging this change but please note that `colour` is actually correct, it's used in most English-speaking contries other than US.

sDos280:
I really sorry, I never heard this word before

PR #2605, opened 2022-08-01
Music doesn't stop looping fix

veins1:
This PR fixes a bug when music.looping is true, then music.looping is set to false but the music continues looping.
Also using proper functions to seek to start of an audio file.
Removed old comment.


raysan5:

PR #2607, opened 2022-08-01
GetKeyRepeat and IsKeyRepeated

BlueStaggo:
GetCharPressed is useful for text boxes because it provides a repeat rate. This is when you hold down a key and it repeats itself while held (example: aaaaaaaaaaaaaaaaaa). However, GetCharPressed does not support special control keys like backspace and enter. GetKeyPressed exists, but that does not have repeat rate. This is why I created GetKeyRepeat and IsKeyRepeated: they are functions that are a mix of GetCharPressed and GetKeyPressed/IsKeyPressed where it listens to key presses with a repeat rate. The text_input_box example has been updated to use this new functionality and a demonstration video is shown below:

https://user-images.githubusercontent.com/71090710/182228431-6adbdbbd-6ca1-4ebf-8e0a-c0094b3d237b.mp4

### DIFF
--- a/examples/text/text_input_box.c
+++ b/examples/text/text_input_box.c
@@ -2,7 +2,7 @@
 *
 *   raylib [text] example - Input Box
 *
-*   Example originally created with raylib 1.7, last time updated with raylib 3.5
+*   Example originally created with raylib 1.7, last time updated with raylib 4.0
 *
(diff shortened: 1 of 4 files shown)


raysan5:
@BlueStaggo Sorry, I'm afraid I'm not adding this functionality, most of the users do not need it and it seems confusing to me.

PR #2608, opened 2022-08-02
Fix rendering issue in audio_raw_stream.c example

TheManTheMythTheGameDev:
Fix issue where old sine waves are still rendered if you decrease the wavelength on the audio_raw_stream example.


raysan5:
@TheManTheMythTheGameDev thanks for the review!

PR #2614, opened 2022-08-05
GetMusicTimePlayed bug fix

veins1:
GetMusicTimePlayed() was calculated using buffered frames, not played frames.
For example: calling LoadMusicStream->UpdateMusicStream->GetMusicTimePlayed would return non-zero value, even though no music was playing.


raysan5:

PR #2615, opened 2022-08-05
Small simplifications

veins1:


raysan5:
@veins1 thanks for the review

PR #2616, opened 2022-08-06
Fix 4coder project template

anggape:


raysan5:
@anggape thanks for the review

PR #2618, opened 2022-08-06
Can't alt-tab out of fullscreen window fix

veins1:
This prevented alt-tabbing out of fullscreen window with fullscreen resolution. I moved the code to the else clause because of the comment "If we are windowed fullscreen". It also says "ensures that window does not minimize when focus is lost" but it doesn't happen even without this code ( at least on Win10 ). Maybe it's needed on other OS, not sure.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4070,14 +4070,6 @@ static bool InitGraphicsDevice(int width, int height)
                 }
             }
         }
-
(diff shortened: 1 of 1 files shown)


raysan5:
@veins1 Thanks for the review. Fullscreen mode has been problematic in the past, maybe it should be reviewed again in the future...

PR #2620, opened 2022-08-08
Updated examples to work with zig master branch

DaJobat:
Specifically, std.fs.OpenDirOptions struct no longer has the iterable flag, now it's a standalone function call std.fs.openIterableDir.
This was added to zig with https://github.com/ziglang/zig/pull/12060


raysan5:
@DaJobat thanks for the review! :D

PR #2629, opened 2022-08-11
update raylib-python-cffi bindings to 4.2

electronstudio:


raysan5:
@electronstudio Wow! that was fast! thanks for the update!

PR #2630, opened 2022-08-11
updating jaylib binding to 4.2

electronstudio:


raysan5:
@electronstudio thanks for the update!

PR #2636, opened 2022-08-13
Fix Codepoint position truncation

dbechrd:
This truncation causes text that spans the zero coord boundary to round differently left of zero vs. right of zero, in turn causing letters to appear squished together. If you actually need the position to be an integer, you should instead `floorf()` the float, rather than doing an integer truncation, but I don't see any reason to convert it to an integer in the first place. Everything else in the equation is a float.

Here's an example of the visual "squishing" artifact that occurs before this fix:
https://i.imgur.com/mxpq0rS.mp4


raysan5:
@dbechrd Thanks for the review! I can't remember the reason for that cast... 🤔

PR #2637, opened 2022-08-13
Fix string lacking null termination in IsFileExtension

HarriP:
When file extension is longer or equal length compared to buffer holding lowercased string, strncpy does not null terminate the string.
Increased buffer size by 1 to ensure it will always be null-terminated, so that following strcmp does not read out of buffer bounds.


raysan5:
@HarriP thanks for the review!

PR #2640, opened 2022-08-15
Tiny documentation fix in reasings.h

trumoose:
The header comments for the function `EaseSineInOut` were mislabeled as `EaseSineOut`. This has been fixed. Not much else to say... weather's nice, huh?


raysan5:
@trumoose Thanks for the fix! 👍😄

PR #2641, opened 2022-08-15
adding a crop to the image

sDos280:
When exporting the image there is a little white area in the top that the bar is hiding. the crop remove this little white area.


sDos280:
I am not sure if that is Intentional or not, but I think this is not Intentional

raysan5:
@sDos280 I'm afraid a crop shouldn't be there. Defined franebuffer for drawing is intentional.

PR #2643, opened 2022-08-17
Win32: resolve some symbols re-definition of windows.h in glfw3native.h

daipom:
This reflects GLFW's fix: https://github.com/glfw/glfw/issues/1348

Currently, we can't build raylib with an external GLFW in Win32
since some symbol re-definition errors occur.

The internal GLFW in raylib seems to avoid this problem by customizing `glfw3native.h`: https://github.com/raysan5/raylib/commit/2feea87b616292b5bce4454a42c2d048f1cce7d8

This fix enables building with an external GLFW in Win32 without any customization,
if it contains the fix of https://github.com/glfw/glfw/issues/1348.

This fix is compatible with the current internal GLFW too.

If we update the internal GLFW to the latest before we merge this branch,
we don't need the commit: 6609c689248bf7931d30a468dd4631edc1a75a0a.
In that case, I will remove that commit from this branch.

### DIFF
--- a/src/external/glfw/include/GLFW/glfw3native.h
+++ b/src/external/glfw/include/GLFW/glfw3native.h
@@ -92,9 +92,6 @@ extern "C" {
 // @raysan5: Actually, only HWND handler needs to be defined
 // Including windows.h could suppose symbols re-definition issues (i.e Rectangle type)
 //#include <windows.h>
- typedef void *PVOID;
- typedef PVOID HANDLE;
- typedef HANDLE HWND;
 #elif defined(GLFW_EXPOSE_NATIVE_COCOA) || defined(GLFW_EXPOSE_NATIVE_NSGL)
  #if defined(__OBJC__)
   #import <Cocoa/Cocoa.h>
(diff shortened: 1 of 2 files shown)


raysan5:
@daipom Oh! That's great! I lost track of that issue long ago! Nice to see a review has been implemented! It will also make it easy to update GLFW!

raysan5:
@daipom Thanks for the update! Merged!

daipom:
Thanks for merging!

In fact, we are developing IME support in GLFW to allow Japanese input, etc.

* https://github.com/glfw/glfw/pull/2130

We would like to use this feature in raylib too once we have finished developing it in GLFW.
Perhaps we need to add some corresponding APIs.

Please let me discuss this with you again on issues and PR!

raysan5:
> In fact, we are developing IME support in GLFW to allow Japanese input, etc.

That's amazing! I've been asked several times to support it on raylib! We can add that feature to raylib when ready!

Please note that raylib GLFW is not updated at the moment, I was waiting for the 3.4 official release of the library but it seems it's taking longer than expected...

daipom:
Thanks so much!

We are trying to get that feature into GLFW v3.4!
I would appreciate the update when GLFW v3.4 is released!

PR #2645, opened 2022-08-17
Changed the button logic

Dinastia-Harzu:
I only wanted to make a better logic button for this particular example.
Before, you could hold mouse button left outside the button, then drag the mouse unto the button and, upon release, the sfx would play. It would also change the sprite of the button to the one where appears to be pressed.
This time, you actually have to press the button in order to change the sprite and play the sfx. I also made it so that if you hold mouse button left on the button and then you drag the mouse outside, the button will release and play the sfx.
Maybe the way it was was the intended way, so if it's not welcomed, I'll understand.

### DIFF
--- a/examples/textures/textures_sprite_button.c
+++ b/examples/textures/textures_sprite_button.c
@@ -36,7 +36,7 @@ int main(void)
 
     int btnState = 0;               // Button state: 0-NORMAL, 1-MOUSE_HOVER, 2-PRESSED
     bool btnAction = false;         // Button action should be activated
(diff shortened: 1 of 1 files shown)


raysan5:
@Dinastia-Harzu I'm afraid something is wrong with this PR, it is replacing all file. I'm not merging it.

Dinastia-Harzu:
Understandable. I'll try again later once I get more familiar with both Git and GitHub, and I hope next time will work properly.

PR #2647, opened 2022-08-19
Fix PATH for Web target

futureapricot:
The current Makefile ignores the system PATH variable when appending the emsdk paths. This PR corrects this for both Windows and Linux


raysan5:
@futureapricot Thanks for the review!

PR #2652, opened 2022-08-21
BINDINGS: raylib-cpp has support for raylib 4.2

RobLoach:
https://github.com/robloach/raylib-cpp


raysan5:
@RobLoach Thanks for the update! :)

PR #2658, opened 2022-08-22
Fix GetCurrentMonitor

evgenydrainov:
GetCurrentMonitor now finds the monitor that has the largest area of intersection with the game window

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1692,7 +1692,7 @@ int GetMonitorCount(void)
 #endif
 }
 
-// Get number of monitors
+// Get current connected monitor
 int GetCurrentMonitor(void)
 {
 #if defined(PLATFORM_DESKTOP)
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -1718,6 +1718,9 @@ int GetCurrentMonitor(void)
         int x = 0;
         int y = 0;
 
+        int best_overlap = 0;
```
raysan5, line 1721: Please, could you review those variables naming to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md).

raysan5:
@evgenydrainov I have my doubts on this improvement, it increases code complexity to take care of a very specific use case.

In what real-use-case situation a user would need that functionality? I mean, user usually open the window in one monitor or another. Even in the case the window was wrongly opened in the middle of two monitors, why does the application need to know that? I can think about scaling but the application can just chust the first one, it would be up to the user to move the window to another monitor...

raysan5:
I'm not merging this improvement for now, it seems to solve a very specific use case situation in exchange of a code complexity increment.

PR #2659, opened 2022-08-22
Update npes_saved_w64devkit.txt

ohboymcgee:
I noticed raylib_resources_compile was calling a different openGL version by default than the ones that were linked in raylib_compile_excecute.  I believe this discrepancy is responsible for the "exited with the code: -1073741819" error that some people were getting where it would compile but not launch a window.


raysan5:
@ohboymcgee The resources must be compiled for 32 bit or 64 bit, current script compiles to 32 bit because it's the default version distributed with raylib installer. It's up to the user to change that.

In any case, resource ompilation has no relation to OpenGL.

PR #2684, opened 2022-09-03
GetRandomFloat

acejacek:
Adding helper function GetRandomFloat, which returns float value within boundaries passed as parameters.
This function can substitute GetRandomValue whenever float number precision is needed.


acejacek:
Adding function for collision detection between point and polygon.

raysan5:
@acejacek Sorry, I'm not adding `GetRandomFloat()`, it can be accomplished with `GetRandomValue()` and divind as required.

Please, can you send a separate PRfor the other function?

raysan5:
@acejacek Please, also review the raylib coding conventions. Thanks.

PR #2685, opened 2022-09-03
Check collision point polygon

acejacek:
Adding a function to _Basic shapes collision detection functions_. Detects if point is inside of polygon described as series of vertices.


raysan5:
@acejacek Thnaks for the addition! Please note that I'm reviewing the function name to `CheckCollisionPointPoly()` for consistency with other functions.

PR #2687, opened 2022-09-04
Fixed a bug in the 2d camera platformer example

skylar779:
`canJump` used to alternate between `true` and `false` when on the ground.
When using `IsKeyPressed` instead of `IsKeyDown` for jumping, the player might not jump due to that behavior. This behavior got fixed by adjusting the condition for checking if an obstacle is being hit. With this change, `canJump` doesn't alternate between `true` and `false` when the player is on the ground and instead stays on `true`.


raysan5:
@skylar779 Good catch! Thanks for the review!

PR #2694, opened 2022-09-11
Enable DXT compression on __APPLE__ targets

Not-Nik:
Uses a macro defined in `<OpenGL/gl3ext.h>`.
Also some general syntax cleanup, removing a few conditionals for unoptimized builds. Should save a few CPU cycles ;)

Tested on macOS Monterey 12.5.1 only, but since Apple's OGL hasn't been updated in ages this should work on older systems too.


raysan5:
@Not-Nik Thanks for the review!

PR #2696, opened 2022-09-15
Fix typo

murilluhenrique:


raysan5:
@murilluhenrique thanks for the fix! 👍🙂

PR #2700, opened 2022-09-17
CMake: Project template fix to easily target raylib version

RobLoach:
This updates the cmake project template to make it easier to target different raylib versions.


raysan5:
@RobLoach thanks for the review!

PR #2702, opened 2022-09-17
Fix touchscreen input related functions on Android

deniska:
Fix display -> screen coordinate conversion for android platform and move it to the platform event handling code, simplifying GetTouchPosition() function implementation.

Fixes previous behavior on Android which is reporting touchscreen input coordinates as a number in range from 0 to 1 to reporting values in screen coordinates.

Tested with native resolution and upscaling. Downscaling currently seems to have another bug somewhere deeper in rendering setup.


raysan5:
@deniska Thanks for the review!

PR #2703, opened 2022-09-18
Fix viewport scaling bug on android after context rebind

deniska:
On android after rebinding context (which happens when you minimize and navigate back to an app, or when you turn a screen off and back on) there's a bug that viewport has a wrong scale and part of it is off screen.

The change fixes it on devices I tried, but the solution feels hacky to me. However when I attempted instead to call SetupViewport() again which feels like a more proper solution, it didn't fix it.


raysan5:
@deniska Thanks for reviewing this issue. `SetupViewport()` had also some issues on the past, I think every platform has its details to take into consideration...

Issue #2695, opened 2022-09-11
[rlgl] rlLoadExtensions makes possibly wrong assumptions about existing extensions


Not-Nik:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

`rlLoadExtensions` always marks certain extensions as supported even on systems that might not have them. If `GRAPHICS_API_OPENGL_33` is defined, it assumes all of its core features exist, but defining `GRAPHICS_API_OPENGL_21` in turn also defines `GRAPHICS_API_OPENGL_33`. Extensions like `GL_ARB_vertex_array_object`, might not be supported on OpenGL 2.1, since it only has been in core OGL since version 3.0.

The best way to handle this, would be to use the corresponding `GLAD` macros to check if the extensions are supported, but I currently don't have a dev environment that uses `GLAD`, so I can't fix it myself.

### Environment

Any environments using OpenGL 2.1 would be affected.

raysan5:
@Not-Nik Good catch! Feel free to send a PR with the GLAD checks for the specific cases.

PR #2706, opened 2022-09-18
[rlgl] Check for extensions before enabling them

Not-Nik:
Updates `external/glad.h` with the new required extension macros.
Draft because I couldn't get it to build on my machine.
Addresses #2695


Not-Nik:
Oh yea, I'll need to do extra work on macOS as well...

Not-Nik:
Unsure whats with the Windows build, I'll look into that later

Not-Nik:
Note to self: ASTC compression might not be present in all OGL 3.3 implementations

src/external/glad.h
```diff
@@ -8304,7 +8336,7 @@ static int glad_gl_get_extensions( int version, const char **out_exts, unsigned
             const char *gl_str_tmp = (const char*) glad_glGetStringi(GL_EXTENSIONS, index);
             size_t len = strlen(gl_str_tmp) + 1;
 
-            char *local_str = (char*) GLAD_MALLOC(len * sizeof(char));
```
raysan5, line 8307: It should use `GLAD_MALLOC` and `GLAD_FREE`

src/external/glad.h
```diff
@@ -12,16 +12,16 @@
  *  - ALIAS = False
  *  - DEBUG = False
  *  - HEADER_ONLY = True
- *  - LOADER = True
+ *  - LOADER = False
```
raysan5, line 15: I think that statement is not correct (?)

src/rlgl.h
```diff
@@ -2018,28 +2011,32 @@ void rlLoadExtensions(void *loader)
     for (int i = 0; i < numExt; i++) TRACELOG(RL_LOG_INFO, "    %s", glGetStringi(GL_EXTENSIONS, i));
 #endif
 
+#if defined(GRAPHICS_API_OPENGL_21)
```
raysan5, line 2014: This code structure looks a bit complex. I prefer to just use a `#if defined(GRAPHICS_API_OPENGL_21)` check for the required extensions and duplicate code, it's more visual.

src/external/glad.h
Not-Nik: Yes it is. raylib doesn't use glad's loader function, but instead `glfwGetProcAddress`. Using glad's loader requires inclusion of `windows.h` which introduces name conflicts for some functions.

src/rlgl.h
```diff
@@ -2028,18 +2033,18 @@ void rlLoadExtensions(void *loader)
...
-    #if defined(__APPLE__)
-    // Apple provides its own extension macros
-    RLGL.ExtSupported.texCompDXT = GL_EXT_texture_compression_s3tc;
-    #else
-    // NOTE: With GLAD, we can check if an extension is supported using the GLAD_GL_xxx booleans
-    RLGL.ExtSupported.texCompDXT = GLAD_GL_EXT_texture_compression_s3tc;  // Texture compression: DXT
-    RLGL.ExtSupported.texCompETC2 = GLAD_GL_ARB_ES3_compatibility;        // Texture compression: ETC2/EAC
-    #endif
+
+#undef CORE_OPENGL_33
```
raysan5, line 2047: No need for this undef

raysan5:
@Not-Nik Looks great! Thanks for the improvement and all the reviews!

PR #2709, opened 2022-09-18
Basic gamepad support for Android

deniska:
Currently assumes a single gamepad, has no code specific to gamepad detection (gamepad is "detected" when an event related to gamepad arrives). Also assumes that all gamepads look roughly like an xbox/ps controller. Both assumptions are not strictly true, but an implementation like that probably covers 85% of usecases.

Also it doesn't update previousButtonState so functions IsGamepadButtonPressed() and IsGamepadButtonReleased() don't work, but they didn't work previously on Android anyway, and they are flaky on desktop as they are now, so the mechanism for these two functions probably should be reworked anyway.

It's certainly an improvement compared to the previous android gamepad handling code, which put gamepad events into touch related structs.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5580,6 +5580,32 @@ static void AndroidCommandCallback(struct android_app *app, int32_t cmd)
     }
 }
 
+static GamepadButton AndroidTranslateGamepadButton(int button)
+{
+    switch (button)
+    {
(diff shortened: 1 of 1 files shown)


raysan5:
@deniska Thanks for the improvement! I'm still trying to remember why touch inputs were checked in gamepad events... I'm sure there was some reason for that but anyway...

PR #2711, opened 2022-09-18
Remove touch points on touch up events on Android

### LINKED ISSUE #2683: [core] `GetTouchPosition()` never indicates no touch

### Issue description

GetTouchPosition seems to retain the previous position that was touched. It does not return to {0,0}

### Environment

Android  7

### Code Example

Set up a simple loop similar to

    loop {
      BeginDrawing()
      ClearBackground(WHITE)
      DrawText(GetTouchPosition(0).x)
      EndDrawing()
    }

Notice the touch position shown never zeroes out after the first touch, suggesting there is *always* a touch present even if the screen is not touched.

deniska:
Fixes #2683

Remove elements from touch point related arrays when touch up and similar events are processed. This makes GetTouchPointCount() always report the actual count of touch points, and all positions returned by GetTouchPosition() correspond to positions of currently happening touches.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5755,9 +5755,6 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
     int32_t action = AMotionEvent_getAction(event);
     unsigned int flags = action & AMOTION_EVENT_ACTION_MASK;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@deniska thanks for the review! 👍🙂

PR #2717, opened 2022-09-24
Switch to glBlendFuncSeparate() to fix alpha blending issue with FBO

SAOMDVN:
When drawing to FBO of type RGBA, if using glBlendFunc the RGB component is blended normally, but the alpha component is overwritten instead of blended, this resulting in incorrect result when the FBO texture is drawn back to the main buffer or exported to file. Changing to glBlendFuncSeparate will fix this

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -594,7 +594,7 @@ RLAPI void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsig
 RLAPI void rlClearScreenBuffers(void);                  // Clear used screen buffers (color and depth)
 RLAPI void rlCheckErrors(void);                         // Check and log OpenGL error codes
 RLAPI void rlSetBlendMode(int mode);                    // Set blending mode
-RLAPI void rlSetBlendFactors(int glSrcFactor, int glDstFactor, int glEquation); // Set blending mode factor and equation (using OpenGL factors)
+RLAPI void rlSetBlendFactors(int glSrcRGB, int glDstRGB, int glSrcAlpha, int glDstAlpha, int glEquation); // Set blending mode factor and equation (using OpenGL factors)
 
 //------------------------------------------------------------------------------------
 // Functions Declaration - rlgl functionality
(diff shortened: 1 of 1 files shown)


RobLoach:
Makes sense to me. Is there an example you have where this is observable?

SAOMDVN:
Consider these two transparent textures: 
R
B
This is how it normally look like when drawn on an opaque white background: 
```c
int main() {
	InitWindow(500, 500, "Test");
	Texture2D red = LoadTexture("R.png");
	Texture2D blue = LoadTexture("B.png");
	while(!WindowShouldClose()) {
		BeginDrawing();
			ClearBackground(WHITE);
			DrawTexture(red, 0, 0, WHITE);
			DrawTexture(blue, 0, 0, WHITE);
		EndDrawing();
	}
	return 0;
}
```
F1


When I try to do the same thing but inside an FBO, here's what happened:
```c
int main() {
	InitWindow(500, 500, "Test");
	Texture2D red = LoadTexture("R.png");
	Texture2D blue = LoadTexture("B.png");
	RenderTexture2D target = LoadRenderTexture(500, 500);
	while(!WindowShouldClose()) {
		BeginTextureMode(target);
			ClearBackground(WHITE);
			DrawTexture(red, 0, 0, WHITE);
			DrawTexture(blue, 0, 0, WHITE);
		EndTextureMode();
		BeginDrawing();
			ClearBackground(GREEN);
			DrawTextureRec(target.texture, (Rectangle){ 0, 0, (float)target.texture.width, (float)-target.texture.height }, (Vector2){ 0, 0 }, WHITE);
		EndDrawing();
	}
	return 0;
}
```
F2
Even though the `ClearBackground(WHITE)` call should make the render texture opaque, it is somehow transparent and the GREEN leaked through

raysan5:
@SAOMDVN Thanks for the improvement! It seems this change breaks some examples, they need to be reviewed.

SAOMDVN:
Hi! How should I proceed with this issue? Do I need to change the examples to use the new rlSetBlendFactors or should I create a separate function rlSetBlendFactorsSeparate and keep the old one the same to avoid breaking changes?

raysan5:
> Hi! How should I proceed with this issue? Do I need to change the examples to use the new rlSetBlendFactors or should I create a separate function rlSetBlendFactorsSeparate and keep the old one the same to avoid breaking changes?

@SAOMDVN This is a good question, let me think about it...

raysan5:
@SAOMDVN I decided to merge PR https://github.com/raysan5/raylib/pull/2741 instead of this one for the reasons stated in the comments. Thank you very much to you both for working on this improvement! :)

PR #2720, opened 2022-09-26
update build.zig

ryupold:
zig `master` now enforces to use addIncludePath instead of addIncludeDir


raysan5:
@ryupold Nice! Thanks for the review! :)

PR #2721, opened 2022-09-26
Add circles to beziers

infdivzero:
Fills in the gaps between line segments made by the draw bezier functions when thick > 1. Also makes the ends round

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -177,6 +177,9 @@ void DrawLineBezier(Vector2 startPos, Vector2 endPos, float thick, Color color)
 
         DrawLineEx(previous, current, thick, color);
 
+        if (thick > 1)
+            DrawCircleV(current, thick / 2, color);
+
         previous = current;
     }
 }
@@ -203,6 +206,9 @@ void DrawLineBezierQuad(Vector2 startPos, Vector2 endPos, Vector2 controlPos, fl
 
         DrawLineEx(previous, current, thick, color);
 
+        if (thick > 1)
+            DrawCircleV(current, thick / 2, color);
+
         previous = current;
     }
 }
@@ -228,6 +234,9 @@ void DrawLineBezierCubic(Vector2 startPos, Vector2 endPos, Vector2 startControlP
         current.x = a*startPos.x + b*startControlPos.x + c*endControlPos.x + d*endPos.x;
 
         DrawLineEx(previous, current, thick, color);
+
+        if (thick > 1)
+            DrawCircleV(current, thick / 2, color);
 
         previous = current;
     }


RobLoach:
Could you provide an example to test this?

infdivzero:
Here's a demo program:
```C
{
        InitWindow(800, 600, "Bezier");

                BeginDrawing();
                ClearBackground(WHITE);

                Vector2 p0, p1, ctl0, ctl1;

                // top bezier
                p0 = (Vector2){800.0f / 8, 600.0f / 8};
                p1 = (Vector2){800.0f / 8 * 7, 600.0f / 8 * 2};
                DrawLineBezier(p0, p1, 32, RED);

                // middle bezier
                p0.y = 600.0f / 8 * 2; p1.y = 600.0f / 8 * 4;
                ctl0 = (Vector2){400.0f, 300.0f};
                DrawLineBezierQuad(p0, p1, ctl0, 32, GREEN);

                // bottom bezier
                p0.y = 600.0f / 8 * 4; p1.y = 600.0f / 8 * 6;
                ctl0 = (Vector2){800.0f / 8, 500.0f};
                ctl1 = (Vector2){800.0f / 8 * 6, 500.0f};
                DrawLineBezierCubic(p0, p1, ctl0, ctl1, 32, BLUE);

                // 1 width bezier
                p0 = (Vector2){10, 10};
                p1 = (Vector2){790.0f, 590.0f};
                ctl0 = (Vector2){-50, 650.0f};
                DrawLineBezierQuad(p0, p1, ctl0, 1, GREEN);

                EndDrawing();
        }

        CloseWindow();

        return 0;
}
```

Without circles:
bezier_nocircle
The gaps between the line segments are more apparent the thicker the bezier gets. Even though it's kinda ugly, the green bezier doesn't have those gaps. Thicknesses of 2 or 3 might also not have it depending on the tightness of the curves. As the beziers get thicker, the gaps will get bigger beginning as single pixels

With circles:
bezier_circle
Here, the gaps are filled in by circles. I put in the thickness > 1 check because it doesn't seem very useful for beziers with thickness of 1. The circles do also round off the ends of the beziers, so I'm not sure if that'll be a desired feature/it should be checked for... The bezier drawn by DrawLineBezier() is kinda weird with both its flat and rounded ends. Either needs an extra circle draw at the start or a check to stop circles from being drawn at the end

Should I add checks for circles at the ends and add some kind of option to enable/disable them?

raysan5:
@infdivzero @RobLoach this change implies a big performance cost for the function (2 triangles vs 36 triangles per point) and also a lot of overdraw, why draw first using rectangles and then overdraw the same line with circles? A clear issue could be drawing bezier lines with transparency.

The visual improvement is really good but, actually, it's not the right solution for the problem.

`DrawLineBezier()` calls `DrawLineEx()` that calls `DrawTriangleStrip()`. That last function should actually manage the gaps properly because it's a triangle strip. 

`DrawLineBezier()` shouldn't use `DrawLineEx()` to draw the bezier in small pieces but use `DrawTriangleStrip()` to define all required points, or event better, use its own implementation based on `DrawTriangleStrip()` one, to avoid the requirement of pre-storing all calculated bezier lines points.

Issue #2729, opened 2022-09-28
[build] `rcore` module does not compile under linux with `-std=c99` *without* `-D_GNU_SOURCE`


Peter0x44:
Please, before submitting a new issue verify and check:

 - [ ] I tested it on latest raylib version from master branch
 - [ ] I checked there is no similar issue already reported

### Issue description
Under Linux, rcore.c cannot compile as c99, without defining _GNU_SOURCE

`gcc -c rcore.c -DGRAPHICS_API_OPENGL_33 -DPLATFORM_DESKTOP -std=c99`
```
In file included from rcore.c:128:
rgestures.h: In function ‘rgGetCurrentTime’:
rgestures.h:540:21: error: storage size of ‘now’ isn’t known
  540 |     struct timespec now;
      |                     ^~~
rgestures.h:541:5: warning: implicit declaration of function ‘clock_gettime’ [-Wimplicit-function-declaration]
  541 |     clock_gettime(CLOCK_MONOTONIC, &now);
      |     ^~~~~~~~~~~~~
rgestures.h:541:19: error: ‘CLOCK_MONOTONIC’ undeclared (first use in this function)
  541 |     clock_gettime(CLOCK_MONOTONIC, &now);
      |                   ^~~~~~~~~~~~~~~
rgestures.h:541:19: note: each undeclared identifier is reported only once for each function it appears in
rcore.c: In function ‘GetApplicationDirectory’:
rcore.c:3074:19: warning: implicit declaration of function ‘readlink’ [-Wimplicit-function-declaration]
 3074 |     ssize_t len = readlink("/proc/self/exe", appDir, size);
      |                   ^~~~~~~~
rcore.c: In function ‘WaitTime’:
rcore.c:4782:16: error: variable ‘req’ has initializer but incomplete type
 4782 |         struct timespec req = { 0 };
      |                ^~~~~~~~
rcore.c:4782:33: warning: excess elements in struct initializer
 4782 |         struct timespec req = { 0 };
      |                                 ^
rcore.c:4782:33: note: (near initialization for ‘req’)
rcore.c:4782:25: error: storage size of ‘req’ isn’t known
 4782 |         struct timespec req = { 0 };
      |                         ^~~
rcore.c:4789:16: warning: implicit declaration of function ‘nanosleep’ [-Wimplicit-function-declaration]
 4789 |         while (nanosleep(&req, &req) == -1) continue;
      |                ^~~~~~~~~
```
[...]

PR #2733, opened 2022-09-30
Update Gestures to use GetTime() if it's available

RobLoach:
Related to https://github.com/raysan5/raylib/issues/2729 , makes gestures.h use `GetTime()` if it's available.


raysan5:
@RobLoach Nice improvement... actually this module requires a full review at this moment...

PR #2738, opened 2022-10-02
Added rudimentary SVG support.

bXi:
I've added NanoSVG to add the loading of SVG files into Raylib.

Created 2 functions to load .svg files.

```cpp 
ImageLoadSvg(filename); //Loads a file using the configured size inside the .svg file
ImageLoadSvgWithSize(filename, width, height); //Loads a file and scales it up to fit inside the provided width/height
```


examples/textures/textures_svg_loading.c
```diff
@@ -0,0 +1,69 @@
...
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    const int screenWidth = 800;
+    const int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [textures] example - image loading");
+
+    // NOTE: Textures MUST be loaded after Window initialization (OpenGL context is required)
+
+    Image image = LoadImage("resources/raylib_logo.png");     // Loaded in CPU memory (RAM)
```
raysan5, line 32: It seems this example is no the expected one, no svg loading happening in here...

src/rtextures.c
```diff
@@ -187,6 +187,13 @@
     #include "external/stb_image_resize.h"  // Required for: stbir_resize_uint8() [ImageResize()]
 #endif
 
+#define NANOSVG_IMPLEMENTATION	// Expands implementation
```
raysan5, line 190: It requires a flag in `config.h` for `SUPPORT_FILEFORMAT_SVG` and the corresponding checks to support this functionality.

src/raylib.h
```diff
@@ -1226,6 +1226,8 @@ RLAPI Rectangle GetCollisionRec(Rectangle rec1, Rectangle rec2);
 // NOTE: This functions do not require GPU access
 RLAPI Image LoadImage(const char *fileName);                                                             // Load image from file into CPU memory (RAM)
 RLAPI Image LoadImageRaw(const char *fileName, int width, int height, int format, int headerSize);       // Load image from RAW file data
+RLAPI Image LoadImageSvg(const char* fileName);                                                          // Load image from SVG file data with default size from SVG
```
raysan5, line 1229: I prefer to keep base SVG loading inside `LoadImage()` as an additional supported file format.

src/raylib.h
```diff
@@ -1226,6 +1226,8 @@ RLAPI Rectangle GetCollisionRec(Rectangle rec1, Rectangle rec2);
...
+RLAPI Image LoadImageSvgWithSize(const char* fileName, int width, int height);                                   // Load image from SVG file data with specified size
```
raysan5, line 1230: This function can be just renamed to `LoadImageSvg()` for now but not sure what I'll do with it because disabling SVG support would break it, raylib does not expose any file-format-specific loading function, tries to keep that data internal always... but I can review it later.

src/rtextures.c
```diff
@@ -275,6 +282,88 @@ Image LoadImageRaw(const char *fileName, int width, int height, int format, int
...
+    Image image = { 0 };
+
+    unsigned int dataSize = 0;
+    unsigned char* fileData = LoadFileData(fileName, &dataSize);
+
+    if (fileData != NULL)
+    {
+        struct NSVGimage* svgImage = nsvgParse(fileData, "px", 96.0f);
+
+
```
raysan5, line 321: Too many spaces.

src/rtextures.c
```diff
@@ -275,6 +282,88 @@ Image LoadImageRaw(const char *fileName, int width, int height, int format, int
...
+    return image;
+}
+
+// Load an image from SVG file data with a custom size
+Image LoadImageSvgWithSize(const char* fileName, int width, int height)
+{
+    Image image = { 0 };
+
+    unsigned int dataSize = 0;
+    unsigned char* fileData = LoadFileData(fileName, &dataSize);
```
raysan5, line 315: Be careful with pointer `*`, raylib conventions place it on the right.

src/rtextures.c
```diff
@@ -275,6 +282,88 @@ Image LoadImageRaw(const char *fileName, int width, int height, int format, int
...
+
+        // Set the largest of the 2 scales to be the scale to use
+    	const float scale = (scaleHeight > scaleWidth) ? scaleWidth : scaleHeight;
+
+        int offsetX = 0;
+        int offsetY = 0;
+
+        if (scaleHeight > scaleWidth)
+        {
+            offsetY = (height - svgImage->height * scale) / 2;
```
raysan5, line 338: Please, review raylib coding conventions: https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md

No spaces between `*` and `/`

src/rtextures.c
```diff
@@ -275,6 +282,88 @@ Image LoadImageRaw(const char *fileName, int width, int height, int format, int
...
+        struct NSVGrasterizer* rast = nsvgCreateRasterizer();
+        nsvgRasterize(rast, svgImage, (int)offsetX, (int)offsetY, scale, img, width, height, width * 4);
+
+        // Populate image struct with all data
+        image.data = img;
+        image.width = width;
+        image.height = height;
+        image.mipmaps = 1;
+        image.format = PIXELFORMAT_UNCOMPRESSED_R8G8B8A8;
+
```
raysan5, line 356: One space is fine.

raysan5:
@bXi Adding some review notes. What is the license of the provided SVG image? Is `test.svg` its original name?

examples/textures/textures_svg_loading.c
bXi: I did accidentally reverted my local changes before uploading this. Fixed this one.

src/rtextures.c
bXi: I've added the flag in `config.h` and put it around the NanoSVG headers.

src/raylib.h
bXi: I've moved loading partially to inside LoadImage(). This will load the .svg file now at the defined size inside the .svg file.

src/raylib.h
bXi: I've renamed LoadImageSvgWithSize() to LoadImageSvg() and added LoadImageSvgFromString as a helper function to work from both LoadImageSvg() and LoadImage(). This has the added side bonus of allowing one to construct their own svg strings and have these loaded as images.

src/rtextures.c
bXi: Reviewed. And I'm wondering why there are no spaces around `*` and `/` but there are spaces around `+` and `-`.

src/rtextures.c
bXi: Reviewed.

src/rtextures.c
bXi: Reviewed.

src/rtextures.c
bXi: Reviewed.

src/raylib.h
raysan5: I think this function is overengineering, probably not many users (if any) will need this function. I prefer to avoid it.

src/rtextures.c
raysan5: It's a coding convention I choose for raylib, probably just a personal preference, it seems easier for my to see distinction between additions and multiplications.

orcmid: Yes, for me it emphasizes the relative precedence of the operations as well.

raysan5:
@bXi Why did you close this PR?

bXi:
@raysan5 I've decided to do this PR over from scratch using a hopefully less overengineered method for both functions.

raysan5:
@bXi Ok, perfect. Thank you very much!

Dima-aka-dima:
@bXi sorry for the ping. Is there any progress on this feature?

bXi:
@Dima-aka-dima sadly not. I've been to busy to work on this properly. You can use the commits to implement this for yourself. All the code is available up there.

grantcarthew:
Hi @raysan5, I was looking for SVG support. Is the code here ok to use? I'm no expert at raylib yet.

raysan5:
@grantcarthew yes, the code here should work.

@bXi Is it possible to load the SVG files at a default size, for example defined to 96ppi? That way SVG could be just processed by `LoadImage()`.

Also, is there some SVG identifier on a SVG string? That way `LoadImageSvg()` could validate if input is a file or a string to load it at a provided size/ppi.

bXi:
@raysan5 the LoadImageSvg() function reads the data inside the .svg file to load it at the size specified in the .svg.

If you have a look at the first tag inside the .svg file: `<svg xmlns="http://www.w3.org/2000/svg" width="2500" height="2484" viewBox="0 0 192.756 191.488">` it would be trivial to detect svg files based on the existence of the "<svg" string inside the file.

The only problem I can see from a quick glance would be that whitespace is allowed before the `<svg>` tag so that would have to be trimmed out before checking.

I will add that the power of .svg comes from the fact that it is a scalable format. That is the reason I added the function to allow svg loading at a specific size so you can always have the most crisp output for your needs. (inspired by the font loading functions basically)

After checking the commits (it has been a while) I do see that I changed LoadImage() to work with .svg files as described.

grantcarthew:
This is great guys. I was planning on building it myself as a method of learning zig through the [raylib-zig](https://github.com/Not-Nik/raylib-zig) library. I'd rather save the time if I can.

Is there some way I can help? I am green when it comes to C, raylib, and zig. I've done some C++ before but not for a couple of years.

grantcarthew:
This is fantastic @raysan5 and @bXi  I'll have a play as soon as I get time. Thanks.
If there is a feature missing for SVG, I'll open a PR.

raysan5:
Merged for further review...

PR #2741, opened 2022-10-05
Add rlSetBlendFactorsSeparate and checks for custom blend mode modifications

pure01fx:
> Coincidentally it seems like #2717 found some similar problems mentioned in this PR... I have made these changed a long time ago, sorry for that.

There are several problems.

1. In some cases rlSetBlendMode is not enough for all situations (like some 3rd party libraries like Live2D, actually I have written a raylib renderer for Live2D for fun, with changes in this PR)
2. If custom blend factors or equations is changed, some functions, like glBlendFunc, should be called. No function call is made now if RLGL.State.currentBlendMode is not changed

For problem 1, I used a more modest approach. I think glBlendFuncSeparate will not be used in most cases, so I add one more blend mode for this case. I don't know if it is appropriate to force developers to use glBlendFuncSeparate...

And I add custom blend mode modification checks in rlSetBlendFactors (and rlSetBlendFactorsSeparate) so that we can know if the custom blend factors and equations changed in rlSetBlendMode and avoid useless blend mode modification calls.

If #2717 is accepted, I can open a new PR to solve problem 2 😊.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -413,7 +413,8 @@ typedef enum {
     RL_BLEND_ADD_COLORS,                // Blend textures adding colors (alternative)
     RL_BLEND_SUBTRACT_COLORS,           // Blend textures subtracting colors (alternative)
     RL_BLEND_ALPHA_PREMULTIPLY,         // Blend premultiplied textures considering alpha
-    RL_BLEND_CUSTOM                     // Blend textures using custom src/dst factors (use rlSetBlendFactors())
+    RL_BLEND_CUSTOM,                    // Blend textures using custom src/dst factors (use rlSetBlendFactors())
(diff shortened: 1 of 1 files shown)


SAOMDVN:
Hi @pure01fx do you have plan to publish your Live2D renderer?

SAOMDVN:
Hi @raysan5 I think there's a missing BLEND_CUSTOM_SEPARATE enum in raylib.h. It's small so I don't think I need to make a PR

raysan5:
@SAOMDVN Thanks! Added! :)

beetaa:
@pure01fx @raysan5 thanks a lot for this pr, this dramatically increase the font rendering, especially in webassembly compiling. btw, I use this function for example:

```c
BeginBlendMode(BLEND_CUSTOM_SEPARATE);
rlSetBlendFactorsSeparate(0x0302, 0x0303, 1, 0x0303, 0x8006, 0x8006);

...balalalala

EndBlendMode();
```

PR #2742, opened 2022-10-06
Set initial window position for display-sized fullscreen

daipom:
Environment

* Ubuntu 20.04 LTS
* X11 window system

We can create a display-sized fullscreen as follows:

```c
SetConfigFlags(FLAG_FULLSCREEN_MODE);
InitWindow(0, 0, "title");
```

However, at this time, `window.position.x` and `window.position.y` are set to `0`.
There is no way to set that value for `ToggleFullscreen`.

This causes problems on X11 when `ToggleFullscreen`.

When toggling a display-sized fullscreen to windowed screen by `ToggleFullscreen`, the window is at (0, 0) as follows:

```c
ToggleFullscreen();
SetWindowSize(800, 450);
```

Screenshot from 2022-10-06 16-39-58

However, sometimes (about 50%) it goes wrong as follows:

Screenshot from 2022-10-06 16-40-20

* The side-bar and top-bar disappear.
* The header of the window doesn't appear.

The initial windowed position should be near the center, not just on this issue.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4051,8 +4051,18 @@ static bool InitGraphicsDevice(int width, int height)
     if (CORE.Window.fullscreen)
     {
         // remember center for switchinging from fullscreen to window
-        CORE.Window.position.x = CORE.Window.display.width/2 - CORE.Window.screen.width/2;
-        CORE.Window.position.y = CORE.Window.display.height/2 - CORE.Window.screen.height/2;
+        if ((CORE.Window.screen.height == CORE.Window.display.height) && (CORE.Window.screen.width == CORE.Window.display.width))
+        {
+            // If screen width/height equal to the dislpay, we can't calclulate the window pos for toggling fullscreened/windowed.
+            // Toggling fullscreened/windowed with pos(0, 0) can cause problems in some platforms, such as X11.
+            CORE.Window.position.x = CORE.Window.display.width/4;
+            CORE.Window.position.y = CORE.Window.display.height/4;
+        }
+        else
+        {
+            CORE.Window.position.x = CORE.Window.display.width/2 - CORE.Window.screen.width/2;
+            CORE.Window.position.y = CORE.Window.display.height/2 - CORE.Window.screen.height/2;
+        }
 
         if (CORE.Window.position.x < 0) CORE.Window.position.x = 0;
         if (CORE.Window.position.y < 0) CORE.Window.position.y = 0;


raysan5:
@daipom Actually, the current code should be enough. `Display size` shouldn't be the same as `screen size` and, in that case, the logic result is allowing all the game area to be visible. I think this is more related to the windowing system than raylib itself.

In any case, this issue can be managed on user side with `SetWindowPosition()`.

daipom:
@raysan5 Thanks for your review!

>  the current code should be correct. In case display size is the same as screen size, it makes sense that window part is not visible and is placed outside of screen.

This is understandable, but when toggling from initial fullscreened state to windowed state, does the window need to be at (0, 0)?

> In any case, this issue can be managed on user side with SetWindowPosition().

I think we can't use `SetWindowPosition()` to solve this issue, since it doesn't change `window.position`.

`window.position` seems to be the previous windowed position, and it is updated only at 
https://github.com/clear-code/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L1190

So it is not possible to set `window.position` in advance and then call `ToggleFullscreen` afterward.

On the other hand, if we call `ToggleFullscreen` first and then call `SetWindowPosition`, we have the same problem:

```c
ToggleFullscreen();
SetWindowSize(800, 400);
SetWindowPosition(300, 300);
```

Screenshot from 2022-10-11 00-12-08

If there is currently no workaround for this problem, and if this fix is still not a good idea, I will create the issue and try to find a better solution!
I'd really appreciate it if you could give me your opinion!

gen2brain:
Is this maybe related to https://github.com/raysan5/raylib/pull/2732#issuecomment-1264586896 ?

raysan5:
@gen2brain Yeah, I also thought about it! Actually, that change already afffected me, windows not always initialize centered on my system: Windows 10.

daipom:
@gen2brain @raysan5 

Hmm, I think this is not related to #2732 
The codes that disappeared with that fix certainly seem unnecessary.

When I revive those codes, I can reproduce this issue.

On Windows, a situation similar to this issue occurs, but it is simply that the header of the window extends beyond the display, and does not behave strangely as on X11.
On Windows, It is simply inconvenient since the header cannot be dragged back.

win32_hide_header

daipom:
I'm sorry. Perhaps my explanation was insufficient.
Let me explain this issue again.

We have 3 cases of toggling fullscreened/windowed.

# A. Init window as windowed state

1. Init a window (windowed state)
1. `ToggleFullscreen()` to fullscreened
1. `ToggleFullscreen()` to windowed again

In this case, `CORE.Window.position` is stored at `2.` as following:

* https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L1190

Then at `3.`, it is applied as following:

* https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L1221

So the window is displayed at the previous position. This is the behavior we expect.

# B. Init window as fullscreened state with different size from the display

This is a case of changing the resolution to fullscreen without changing the frame buffer size.

1. Init a fullscreened window with different size from the display
    ```c
    SetConfigFlags(FLAG_FULLSCREEN_MODE);
    InitWindow(800, 600, "title");
    ```
1. `ToggleFullscreen()` to windowed state

In this case, `CORE.Window.position` is calculated at `1.` as following:

* https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L4063-L4064

This logic exists before this PR.

So the window is displayed in the center at `2.`. This is the behavior we expect.

# C. Init window as fullscreened state with the dislay size

This is the case at issue here, not changing the resolution.

1. Init a fullscreened window with the display size
    ```c
    SetConfigFlags(FLAG_FULLSCREEN_MODE);
    InitWindow(0, 0, "title");
    ```
1. `ToggleFullscreen()` to windowed state
    ```c
    ToggleFullscreen();
    SetWindowSize(800, 600);
    ```

In this case, the existing calculation logic:

* https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L4063-L4064

can not calculate `CORE.Window.position` because the screen size completely equals to the display size.
So `CORE.Window.position` will be `(0, 0)`.

So I added calculation logic for this case:

* https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L4054-L4060

Without this fix, we can move the window after `ToggleFullscreen()` by `SetWindowPosition()`.

However, on X11, the window sometimes goes wrong (Seemingly transformed into a window, but the window header doesn't appear and it behaves exclusively like a fullscreen. In addition, the desktop sidebar and topbar doesn't displayed.)
(The screenshot is in https://github.com/raysan5/raylib/pull/2742#issuecomment-1273496188)

This problem seems to be caused by the toggling fullscreened/windowed at `(0, 0)`.

There is no way to manage `CORE.Window.position` by the application side, but it makes sense since it just stores the previous windowed position for `ToggleFullscreen()`.

This fix is mainly for this problem, but it would generally be preferable to have the window appear closer to the center as well for the case `C`.

daipom:
https://github.com/glfw/glfw/issues/2141

It seems that a similar problem is recognized on the GLFW side too.
When the window's top-left is at `(0, 0)`  and the size equals the current video mode,  the window state can go wrong on X11.

In the case `C`  in https://github.com/raysan5/raylib/pull/2742#issuecomment-1274727050 , the following code always create this situation.

https://github.com/raysan5/raylib/blob/46f41722771ea3bb18cd1784a1d1a079aba25d23/src/rcore.c#L1221

This problem can be avoided by making the window appear closer to the center.

raysan5:
@daipom Excuse the delay reviewing this issue and thank you very much for the very detailed explanation. I see the problem is partially related to GLFW and X11... I merged the provided solution for not but the offset added is arbitrary and it could confuse some users... probably it will need some further review in the future.

daipom:
Thanks for reviewing and merging!

> the offset added is arbitrary and it could confuse some users... probably it will need some further review in the future.

OK!
According to https://github.com/glfw/glfw/issues/2141, it seems difficult to fundamentally resolve this problem in GLFW.
However, it may be possible to find a better workaround or to create a workaround for X11 on the GLFW side.

Issue #2415, opened 2022-03-27
[core] ToggleFullscreen() does not work on linux


moonburnt:
### Issue description

Long story short - ToggleFullscreen() does not seem to work correctly (or at least on my setup). Attempting to run game in windowed mode and then switch to fullscreen cause screen (not game's, but whole monitor's) resolution to get messy for a second, then it goes back to normal and window loses focus. Any attempts to bring window back cause the bug to repeat.
At first I thought its my own fault, coz I did not change window's resolution to screen's resolution before calling ToggleFullscreen(). So I did that and then... it didn't work again. I went to google and found this example: https://gist.github.com/JeffM2501/6e4630a0e34c0c7dddf066f7192e342d which seems legit. But attempts to toggle fullscreen in third party code cause said bug to appear too.

### Environment

Manjaro linux, Cinnamon desktop environment, Intel 4700HQ CPU.
GPU/OpenGL info from raylib's log:
```
Vendor:   Intel Open Source Technology Center
Renderer: Mesa DRI Intel(R) HD Graphics 4600 (HSW GT2)
Version:  4.5 (Core Profile) Mesa 21.3.7
GLSL:     4.50
```

### Issue Screenshot

I don't think its possible to screenshot this issue, but I've tried to record a video. But its buggy due to resolution switch shenanigans and thus may not work properly
https://user-images.githubusercontent.com/52989889/160263006-72ddc997-9dd5-40a3-92c2-4c5bc8e976f8.mp4

### Code Example
```cpp
#include "raylib.h"

int main() {
    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "test");

    SetTargetFPS(60);

    while (!WindowShouldClose()) {
        if (IsKeyPressed(KEY_ENTER) && (IsKeyDown(KEY_LEFT_ALT) || IsKeyDown(KEY_RIGHT_ALT))) {
            int display = GetCurrentMonitor();
 
            if (IsWindowFullscreen()) {
                SetWindowSize(screenWidth, screenHeight);
            }
            else {
                SetWindowSize(GetMonitorWidth(display), GetMonitorHeight(display));
            }

            ToggleFullscreen();
        }

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText("Press Alt + Enter to Toggle full screen!", 190, 200, 20, LIGHTGRAY);
        EndDrawing();
    }

    CloseWindow();
    return 0;
}
```

[...]

Issue #2459, opened 2022-05-03
[core] `SetMousePosition()` delayed by one frame on X11 systems


hkva:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

*Copied from my Discord message - HK#2413*

I'm having an issue where calls to `SetMousePosition` are effectively delayed by 1 frame and I have no idea why. I'm in the process of writing my own first-person mouse look function but every time I try to reset the cursor to the middle of the screen, `GetMousePosition` reports the old position for an extra frame.

### Environment

Arch Linux (X11 + dwm), OpenGL 4.6, AMD Radeon RX 580 **Bug reproduced consistently**

Windows 10 virtual machine, OpenGL 4.6, Nvidia GTX 960 **Could not reproduce bug**

### Code Example

```c
#include "raylib.h"

int main() {
    InitWindow(800, 600, "Mouse issue demo");
    SetTargetFPS(60);

    while (!WindowShouldClose()) {
        // If the mouse x position is > 100, reset it back to 5
        Vector2 mouse = GetMousePosition();
        TraceLog(LOG_INFO, "Mouse x: %f", mouse.x);
        if (mouse.x > 100) {
            TraceLog(LOG_INFO, "Mouse is > 100, resetting");
            SetMousePosition(5, mouse.y);
        }

        BeginDrawing();
            ClearBackground(BLACK);
            DrawLine(mouse.x, 0, mouse.x, 600, GREEN);
            DrawLine(100, 0, 100, 600, RED);
        EndDrawing();
    }

    CloseWindow();
}
```

If you move the mouse right at a medium speed, you can see the issue in the debug messages

```
INFO: Mouse x: 92.000000
INFO: Mouse x: 99.000000
INFO: Mouse x: 106.000000
INFO: Mouse is > 100, resetting                <- SetMousePosition called here
INFO: Mouse x: 115.000000                      <- Next frame, mouse hasn't moved
INFO: Mouse is > 100, resetting
INFO: Mouse x: 13.000000
INFO: Mouse x: 12.000000
INFO: Mouse x: 19.000000
```

hkva:
I looked into the issue some more and I think I've found the problem. It seems to be caused by how Xorg buffers window events. In the above code example, here's what's going on:

1. The user moves the mouse, generating Xorg mouse move events that are buffered by GLFW
[...]

PR #2743, opened 2022-10-06
SetWindowSize: make sure to apply new window-size in one frame

daipom:
On X11, "WindowSizeCallback" is called on the next frame. So "CORE.Window.screen.width/height" can differ from the actual size in the frame where we call "SetWindowSize".

This can be a problem, for example, with the codes like follows (#2415):

```c
const int monitorIndex = GetCurrentMonitor();
SetWindowSize(GetMonitorWidth(monitorIndex), GetMonitorHeight(monitorIndex));
ToggleFullscreen();
```

Since "CORE.Window.screen.width/height" is not updated until the next frame, `ToggleFullscreen` is processed with the previous screen size, and the resolution is unintentionally changed.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -630,6 +630,7 @@ static void ScanDirectoryFilesRecursively(const char *basePath, FilePathList *li
 static void ErrorCallback(int error, const char *description);                             // GLFW3 Error Callback, runs on GLFW3 error
 // Window callbacks events
 static void WindowSizeCallback(GLFWwindow *window, int width, int height);                 // GLFW3 WindowSize Callback, runs when window is resized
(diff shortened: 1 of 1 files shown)


raysan5:
@daipom I'm afraid this is a X11 specific issue and I prefer to avoid this kind of paltform-specific patches. Actually there is was already a related issue: https://github.com/raysan5/raylib/issues/2459

I'm keeping this PR open for the moment while thinking what to do with it.

raysan5:
I'm finally closing this PR, it's a platform specific issue that should probably be addressed at GLFW level.

daipom:
> it's a platform specific issue that should probably be addressed at GLFW level.

I agree! Thanks for your review!

PR #2749, opened 2022-10-11
build raylib_api without the 'vectex' typos

sDos280:
i rebuild the raylib_api(s)


raysan5:
@sDos280 It's perfect! Thanks!

Issue #2744, opened 2022-10-06
[rshapes] Inconsistent and wrong rendering of lines


hatkidchan:
### Issue description

Line drawing functions are not consistent between normal and texture mode, as well as just wrong in general (i.e. they're off by a pixel, not long enough and so on). When line is exactly vertical or horizontal, they're offset to one or other side, and sometimes end point of line is not reaching specified coordinates.

### Environment

Platform: Desktop
OS: Arch Linux
OpenGL: `4.6 (Core Profile) Mesa 22.1.7`
GPU: `Intel(R) HD Graphics 4600 (HSW GT2)`

### Issue Screenshot

![lines](https://user-images.githubusercontent.com/35804000/194289136-397fd4be-02d0-4e97-a281-9a247f92b508.gif) ![lines-004](https://user-images.githubusercontent.com/35804000/194289459-ae651f4a-bf73-4a3d-bd87-eebb2a4e433f.png)
![lines-000](https://user-images.githubusercontent.com/35804000/194289346-32d1fd1b-8ac0-4d05-b94d-0921d38a93fe.png) ![lines-180](https://user-images.githubusercontent.com/35804000/194289340-a32676e3-a175-4b4b-afae-e203bd664683.png)
![lines-270](https://user-images.githubusercontent.com/35804000/194290663-f97dda30-38e5-41fd-b08f-90855ff9a2a3.png) ![lines-090](https://user-images.githubusercontent.com/35804000/194290670-6fcee753-c1c9-4653-a30f-e02683d8fbab.png)

### Code Example

```c
#include <math.h>
#include <raylib.h>
#include <raymath.h>

const Color PURE_RED = { 255, 0, 0, 255 },
      PURE_GREEN = { 0, 255, 0, 255 },
      PURE_BLUE = { 0, 0, 255, 255 };

int main(void) {
  InitWindow(256, 256, "lines");
  SetTargetFPS(10);

  RenderTexture2D retex = LoadRenderTexture(256, 256);
  for (int angle = 0; !WindowShouldClose(); angle = (angle + 1) % 360) {
    Vector2 dir = Vector2Rotate((Vector2){ 64, 0 }, angle * M_PI / 180.);
    Vector2 start = { 128, 128 }, end = Vector2Add(start, dir);

    BeginDrawing();
    {
      BeginTextureMode(retex);
      {
        ClearBackground(BLACK);
        DrawLineV(start, end, PURE_RED);
      }
      EndTextureMode();

[...]

PR #2750, opened 2022-10-11
Use RL_QUADS/RL_TRIANGLES for single-pixel drawing

hatkidchan:
Addresses problem mentioned in https://github.com/raysan5/raylib/issues/2744#issuecomment-1273568263 (in short: when drawing pixels using `DrawPixel`/`DrawPixelV` in camera mode, upscaled pixel becomes a line instead of bigger pixel).

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -104,21 +104,50 @@ void SetShapesTexture(Texture2D texture, Rectangle source)
 // Draw a pixel
 void DrawPixel(int posX, int posY, Color color)
 {
-    rlBegin(RL_LINES);
-        rlColor4ub(color.r, color.g, color.b, color.a);
-        rlVertex2f(posX, posY);
-        rlVertex2f(posX + 1, posY + 1);
-    rlEnd();
+  DrawPixelV((Vector2){ posX, posY }, color);
 }
 
 // Draw a pixel (Vector version)
 void DrawPixelV(Vector2 position, Color color)
 {
-    rlBegin(RL_LINES);
+#if defined(SUPPORT_QUADS_DRAW_MODE)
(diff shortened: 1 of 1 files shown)


src/rshapes.c
```diff
@@ -104,21 +104,50 @@ void SetShapesTexture(Texture2D texture, Rectangle source)
...
-        rlVertex2f(posX + 1, posY + 1);
-    rlEnd();
+  DrawPixelV((Vector2){ posX, posY }, color);
 }
 
 // Draw a pixel (Vector version)
 void DrawPixelV(Vector2 position, Color color)
 {
-    rlBegin(RL_LINES);
+#if defined(SUPPORT_QUADS_DRAW_MODE)
```
hatkidchan, line 113: It's basically the `DrawRectanglePro` without rotation stuff and bounding calculations. We *could* call it instead, but I think just pasting it here is more efficient

src/rshapes.c
```diff
@@ -104,21 +104,50 @@ void SetShapesTexture(Texture2D texture, Rectangle source)
 // Draw a pixel
 void DrawPixel(int posX, int posY, Color color)
 {
-    rlBegin(RL_LINES);
-        rlColor4ub(color.r, color.g, color.b, color.a);
-        rlVertex2f(posX, posY);
-        rlVertex2f(posX + 1, posY + 1);
-    rlEnd();
+  DrawPixelV((Vector2){ posX, posY }, color);
```
hatkidchan, line 107: Probably could be `inline`d now to save some cycles on jumps, but not sure if that's a big deal

raysan5:
@hatkidchan Thank you very much for the review and the quick PR! :)

PR #2752, opened 2022-10-12
Fix & Simplify .vox signature check

CrezyDud:
and make version check be only 150 not over 150
because everytime a version changes, the way to load the file gets different, we would first have to add support, then allow

the old version check was big and didnt work(it just checked for the first letter, and if the first letter was not "V" and "O" and "x" AND " " THEN it would fail, not good, so i made this!


raysan5:
@CrezyDud Nice! Thanks for the review!

CrezyDud:
Thanks for merging!

Issue #2755, opened 2022-10-13
[rtextures] ImageDrawTextEx is not scaling down resulting image


hatkidchan:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Glyphs are not scaled down when rendering text using `ImageDrawTextEx`

### Environment

OS: Arch Linux
OpenGL: 4.6 (Core Profile) Mesa 22.1.7
GPU: Intel(R) HD Graphics 4600

### Issue Screenshot
![image](https://user-images.githubusercontent.com/35804000/195663453-349ac74b-bb77-4007-934d-bec59e0650f6.png)
Gray area has the same size as returned by `MeasureTextEx` on same font with same size and spacing

### Code Example

```c
#include <raylib.h>

int main(void) {
  InitWindow(320, 240, "LoadFont");
  SetTargetFPS(15);

  // Changing it to LoadFontEx("./unscii-16-full.ttf", 16, 0, 256) fixes the issue for that font size only.
  Font font = LoadFont("./unscii-16-full.ttf");
  Image img = GenImageColor(320, 240, BLACK);
  Vector2 box = MeasureTextEx(font, "Hello, world!", 16, 0);
  ImageDrawRectangle(&img, 4, 4, box.x, box.y, GRAY);
  ImageDrawTextEx(&img, font, "Hello, world!", (Vector2){ 4, 4 }, 16, 0, WHITE);
  Texture2D tex = LoadTextureFromImage(img);
  UnloadImage(img);

  while (!WindowShouldClose()) {
    BeginDrawing();
    DrawTexture(tex, 0, 0, WHITE);
    EndDrawing();
  }

  UnloadFont(font);
}
```

Font used: [unscii-16-full.ttf](http://viznut.fi/unscii/unscii-16-full.ttf) ([homepage](http://viznut.fi/unscii/))

### Details

Problem lies in the `ImageTextEx` function (it uses base font size instead of the one we specified):
https://github.com/raysan5/raylib/blob/11fd883ee443d3ea478e750d0116611d72d5708d/src/rtextures.c#L1231

Also, we're scaling font only up, but not down:
https://github.com/raysan5/raylib/blob/11fd883ee443d3ea478e750d0116611d72d5708d/src/rtextures.c#L1270-L1279

~~I think it'd be better to change condition on L1270 to `fontSize != imSize.y`~~

[...]

PR #2756, opened 2022-10-13
Fix ImageTextEx and ImageDrawTextEx scaling

hatkidchan:
Already explained in #2755, should close it.

Also it would be nice to specify a scaling method 'cause not all fonts look great when using cubic interpolation, but that can be easily done manually by drawing text with "native" font size first and then using `ImageResizeNN` if desired


hatkidchan:
oh wow, I probably should've created another fork for that or just didn't sync it with upstream

that looks scary
image

hopefully it's not big of a deal

raysan5:
@hatkidchan Hey! Thanks for the detailed issue review and the fix provided! :D

hatkidchan:
no problem!

sometimes it feels like raylib lacks some deep testing of obscure features like that though-
glad to be able to help with at least some of them

raysan5:
@hatkidchan You are right, raylib lacks some deep testing of features but I'm developing alone and I can't do everything in the way I'd like to, sometimes new features are not properly tested or just tested on Windows platform... Hopefully community can help with that...

PR #2761, opened 2022-10-16
Add DrawCapsule(Wires)

IanBand:
These changes add the DrawCapsule() and DrawCapsuleWires() functions. They are heavily based off of the DrawCylinder() and DrawSphere() functions. I came across the need to render capsules based off of two points and a radius when dealing with [capsule collision](https://wickedengine.net/2020/04/26/capsule-collision-detection/) and decided to implement these functions. 

I'm not sure if these should actually be the DrawCapsuleEx() functions, and that the DrawCapsule() functions should always draw capsules vertically and instead of taking a second point, just take a "length" parameter, but having them defined by two points and a radius seems simple enough and its what I need for my use case. 

I also updated the models_geometric_shapes example and its image.


raysan5:
@IanBand Thanks for the addition! :)

PR #2763, opened 2022-10-17
build.zig: let user decide how to set build mode + fix linker warning

InKryption:
This should delegate the responsibility of calling `standardReleaseOptions` and setting the build mode of the `*LibExeObjStep` step to the caller, especially since this might not be the process by which one wants to determine the build mode.

Also changes hides `getSrcDir` to enforce usage of `srcdir`, and asserts that the file is in fact inside a directory.

Also fixes a linker error, by passing `null` to the `root_src` param of `addStaticLibrary` instead of the header file.


raysan5:
@InKryption I'm not an expert on Zig build system but I'm merging those changes.

PR #2765, opened 2022-10-20
Review parser Makefile

Peter0x44:
Currently, the `make all` target will segfault if it is run without all the necessary files present, now make instead puts an error. Also, the path to the old "easings.h" has been updated to "reasings.h" 

The parser itself has been reviewed to print a more helpful error message instead of segfaulting, if a nonexistent file is passed on the command line.


raysan5:
@Peter0x44 thanks for the review! 👍🙂

Issue #2735, opened 2022-10-02
[shapes] `DrawLineBezier*()` functions redesign


raysan5:
As stated in this PR (https://github.com/raysan5/raylib/pull/2721), `DrawLineBezier*()` functions do not work as expected, leaving gaps between the lines. Solution requires a redesign of those functions using triangle-strip.

`DrawLineBezier()` calls `DrawLineEx()` that calls `DrawTriangleStrip()`. That last function should actually manage the gaps properly because it's a triangle strip. 

`DrawLineBezier()` shouldn't use `DrawLineEx()` to draw the bezier in small pieces but use a custom `DrawTriangleStrip()` implementation to define all required points while drawing the triangles connection between them.

kgpjay:
hi @raysan5, please assign me this issue, I would can work on this.

raysan5:
@kgpjay You got the issue assigned, feel free to ask if you have any question. Remember to check the required [coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md).

raysan5:
@kgpjay Are you working on this feature? Any progress?

nobytesgiven:
Working on this, will open a pull request soon.

PR #2767, opened 2022-10-22
Fix bezier line breaking #2735

nobytesgiven:
This fixes the issue #2735.
I used the testing code from the issue above.
image

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -207,11 +207,12 @@ void DrawLineBezier(Vector2 startPos, Vector2 endPos, float thick, Color color)
         current.y = EaseCubicInOut((float)i, startPos.y, endPos.y - startPos.y, (float)BEZIER_LINE_DIVISIONS);
         current.x = previous.x + (endPos.x - startPos.x)/ (float)BEZIER_LINE_DIVISIONS;
 
-        double dy = current.y-previous.y;
-        double dx = current.x-previous.x;
-        double size = 0.5*thick/sqrt(dx*dx+dy*dy);
+        float dy = current.y-previous.y;
+        float dx = current.x-previous.x;
+        float size = 0.5*thick/sqrt(dx*dx+dy*dy);
 
-        if (i==1) {
+        if (i==1)
+        {
             pts[0].x = previous.x+dy*size;
             pts[0].y = previous.y-dx*size;
             pts[1].x = previous.x-dy*size;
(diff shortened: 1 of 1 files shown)


src/rshapes.c
```diff
@@ -197,19 +198,36 @@ void DrawLineBezier(Vector2 startPos, Vector2 endPos, float thick, Color color)
...
         // NOTE: Easing is calculated only for y position value
         current.y = EaseCubicInOut((float)i, startPos.y, endPos.y - startPos.y, (float)BEZIER_LINE_DIVISIONS);
         current.x = previous.x + (endPos.x - startPos.x)/ (float)BEZIER_LINE_DIVISIONS;
 
-        // TODO: Avoid drawing the line by pieces, it generates gaps for big thicks,
-        // Custom "triangle-strip" implementation should be used, check DrawTriangleStrip() for reference
-        DrawLineEx(previous, current, thick, color);
+        double dy = current.y-previous.y;
+        double dx = current.x-previous.x;
+        double size = 0.5*thick/sqrt(dx*dx+dy*dy);
```
JosephSBoyle, line 212: Should `dy`, `dx` and `size` be floats instead of doubles? 

`EaseCubicInOut` returns a float, and the values contained in `Vector2` are also floats.

nobytesgiven: You are right, thanks for the review!

orcmid: > Should `dy`, `dx` and `size` be floats instead of doubles?
> 
> `EaseCubicInOut` returns a float, and the values contained in `Vector2` are also floats.

I suspect it is because the prototype of `sqrt` is

```C
double sqrt(double x)
```
and you don't want to lose precision going into or out of that function, especially because `dy` and `dx` are differences and those can be fragile if the differences between current and previous are relatively small.

Rather than simply line up types, it might be useful to do some test computations to see if there are noticeable differences in the actual precision achieved.

nobytesgiven: I don't think it matters, 7 digits of precision are enough. I guess `size` can remain double but it's really nitpicky in my opinion.

raysan5: please, could you avoid doubles? I tried to avoid them in all raylib.

orcmid: It is valuable to keep in mind that differences of same-signed floating-point values are notorious for loss of _accuracy_, in some cases, regardless of the capacity (what we tend to call precision) of the floating-point values.  Then multiplying those differences by something (or squaring them) makes it worse.  (There's good reason that `sqrt` produces a `double` result.)

So, rather than believing in _x_-digits precision, one needs to be able to show how at least enough _accuracy_ is achieved.  In this case, enough to ensure smoothness in graphical determination.  Or simply hedge and use higher-precision for intermediate operations and their stored results.  The compiler and the floating-point instructions are already doing some of that without requiring any attention, so long as the code does not get in the way.

orcmid: > please, could you avoid doubles? I tried to avoid them in all raylib.

@raysan5 Do you use a compiler option to force that in intermediate calculations, or is your concern just about stored values and parameter types?  (I'm not certain the intermediate cases are avoidable.)

src/rshapes.c
```diff
@@ -197,19 +198,37 @@ void DrawLineBezier(Vector2 startPos, Vector2 endPos, float thick, Color color)
     Vector2 previous = startPos;
     Vector2 current = { 0 };
 
+    Vector2* pts = RL_MALLOC((2*BEZIER_LINE_DIVISIONS+2)*sizeof(Vector2));
```
raysan5, line 201: I prefer to avoid memory allocations. We can just use a stack allocated array:
```c
Vector2 points[2*BEZIER_LINE_DIVISIONS + 2] = { 0 };
```

nobytesgiven: Fixed.

raysan5:
@nobytesgiven thank you very much for the improvement!

PR #2768, opened 2022-10-23
Refactor out unnecessary stack allocations in collision detection procedures

JosephSBoyle:
There are some unnecessary variables being created in the 'CheckCollision' functions which this PR refactors out. Additionally, removing these unnecessary vars has benefits for readability and terseness.

---------------------------------------
Notes:
I checked the following toy example with [godbolt](https://godbolt.org/):
```C
// Existing pattern
int checkStackAllocation(int foo, int bar) {
    int baz = 0;
    if (foo > bar) baz = 1;
    return baz;
}
// Refactored version
int checkStackAllocation2(int foo, int bar) {
    return (foo > bar);
}
```
With optimization enabled (even `-O1` in gcc 12.2) the compiler avoids allocating 'baz' in the first example.

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1557,18 +1557,12 @@ bool CheckCollisionPointRec(Vector2 point, Rectangle rec)
 // Check if point is inside circle
 bool CheckCollisionPointCircle(Vector2 point, Vector2 center, float radius)
 {
-    bool collision = false;
-
-    collision = CheckCollisionCircles(point, 0, center, radius);
-
-    return collision;
+    return CheckCollisionCircles(point, 0, center, radius);
 }
 
 // Check if point is inside a triangle defined by three points (p1, p2, p3)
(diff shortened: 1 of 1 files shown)


raysan5:
@JosephSBoyle Sorry, I'm not merging this PR. Those variables are deliverately created to be consistent with all collision detection functions and follow same pattern.

630Studios:
Asinine reason to keep unneeded and wasteful stack allocations.

orcmid:
> ... unneeded and wasteful stack allocations.

Those are the simplest and most efficient allocations and compilers these days can take advantage of them to remove them from code, using registers as needed.  @JosephSBoyle  confirmed that can be the case.  

I also favor the simplification of the @JosephSBoyle example and I would do things like that in *my* code.  @raysan5 gets to set the style for raylib.  It's his ball and bat.

This is a favorite of mine, but not without explanation for beginners.

```C
#include <stdio.h>

int main(void)
{   return fputs("Hello World \n", stdout);
    }
```

I don't ask you to like it.  I expect you can figure out the reason for it.

PR #2770, opened 2022-10-23
Added Box and Gaussian blurring

nobytesgiven:
I implemented Box and Gaussian blurring. Efficient blurring code is hard to write when prototyping and is used in many places (pause-screens, shadows).

### Example code
```c
#include "raylib.h"

{
        InitWindow(800, 600, "Blur");
		Image im =  LoadImage("redstar.png");
		ImageBlurGaussian(&im, 5);
		//ImageBlurBox(&im, 5);
		Texture2D  tx = LoadTextureFromImage(im);
                BeginDrawing();
                ClearBackground(WHITE);
				
				 DrawTexture(tx, 0, 0, WHITE);

                EndDrawing();
        }

		UnloadImage(im);
		UnloadTexture(tx);
        CloseWindow();

        return 0;
}
```
image

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -194,6 +194,10 @@
     #define PIXELFORMAT_UNCOMPRESSED_R5G5B5A1_ALPHA_THRESHOLD  50    // Threshold over 255 to set alpha as 0
 #endif
 
+#ifndef GAUSSIAN_BLUR_ITERATIONS
+    #define GAUSSIAN_BLUR_ITERATIONS  4    // Number of box blur iterations to approximate gaussian blur
+#endif
+
 //----------------------------------------------------------------------------------
 // Types and Structures Definition
 //----------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


raysan5:
@nobytesgiven Could we convert it into a single function?
```c
ImageBlur(Image *image, int blurSize, int iterations);
```

src/rtextures.c
```diff
@@ -1494,6 +1498,147 @@ void ImageAlphaPremultiply(Image *image)
     ImageFormat(image, format);
 }
 
+// Apply box blur
+void ImageBlurBox(Image *image, int blurSize) {
```
JosephSBoyle, line 1502: I'm not familiar with this algorithm so this function is a little hard for me to understand. Is there a description you could add that names it, or even better a link to the algorithm's e.g wikipedia page?

nobytesgiven: https://en.m.wikipedia.org/wiki/Box_blur

raysan5:
> I guess we can but ImageBlur is very generic, there are multiple types of blur (motion blur, zoom blur, out of focus blur, etc).
> Also, I don't think a user will be able to understand why they need more iterations in their blurring code.

I agree, then we can just keep the `ImageBlurGaussian()`, with the iterations define in case someone needs another kind of blur. Personally I always use gaussian blur, never box blur, quality is higher and, considering it is a function not to be called on a loop, just providing the best quality (and most common) version is a good trade-off.

nobytesgiven:
@raysan5 Everything should be fixed now.

raysan5:
@nobytesgiven thank you very much for the addition! 🙂

PR #2774, opened 2022-10-24
Updated bindings support for crystal

sol-vin:
All fresh and current.


raysan5:
@sol-vin thanks for the update

PR #2775, opened 2022-10-25
Fixed blur issue on opaque pictures & added example

nobytesgiven:
Sorry for the second pull request on this, there was a problem on opaque pictures.
Also added blur on an example.


raysan5:
@nobytesgiven No problem! Thanks for the fix!

PR #2776, opened 2022-10-26
Add ChezScheme support

Yunoinsky:
I've just write a binding generator for raylib in Chez Scheme 9.5.8, and rewrite some of the examples in scheme✅. This idea is inspired by  Petelliott's [raylib-guile](https://github.com/petelliott/raylib-guile).


raysan5:
@Yunoinsky Nice! Great addition!

PR #2786, opened 2022-11-02
Fix examples/build.zig for the latest Zig version

RomanAkberov:
#2720 fixed the build.zig in src/, but not the one in examples/.


raysan5:
@RomanAkberov thanks for the review!

PR #2787, opened 2022-11-04
Fix ExportDataAsCode() data types

RGDTAB:
When I tried using negative numbers, or numbers over 128, the hexadecimal value returned would have 8 digits instead of only having 2. 
ExportDataAsCode() uses the 'x' format specifier to convert decimal to hexadecimal, which is only usable with unsigned data.


raysan5:
@RGDTAB thanks for the review

PR #2790, opened 2022-11-08
removing typo

sDos280:


raysan5:
@sDos280 thanks for the fix!

PR #2793, opened 2022-11-09
Add frameworks needed on macos

SpexGuy:
I just built the project on mac os, and needed to add several frameworks to resolve missing symbol linker errors.


raysan5:
@SpexGuy thanks for the review!

PR #2797, opened 2022-11-12
Added rlCullFace

jtainer:
Currently, rlgl does not expose glCullFace. Changing the face culling mode is useful for shadow mapping. rlCullFace sets the face culling mode to RL_FRONT or RL_BACK, corresponding to GL_FRONT and GL_BACK respectively.


raysan5:
@jtainer Thanks for the addition.

PR #2798, opened 2022-11-12
Raylib-py updated to 4.2, plus parallel project

overdev:
Now [Raylib-py](https://github.com/overdev/raylib-py) is a releases-only project.
For now on, code maintenance will happen in a parallel project created to automate the binding generation: [RaylibpyCtbg](https://github.com/overdev/raylibpyctbg)


raysan5:
@overdev Nice! thanks for the update! 👍🙂

PR #2800, opened 2022-11-14
Warning on GetRandomValue range limit

Pere001:
Added a comment explaining the range limitations of GetRandomValue. Added a run-time warning TRACELOG when GetRandomValue is called with an invalid range.


raysan5:
@Pere001 Thanks for the addition!

PR #2804, opened 2022-11-21
Add raylib-vapi

lxmcf:
Adding Vala bindings for raylib core


raysan5:
@lxmcf Nice! Thanks for the addition! :D

PR #2805, opened 2022-11-22
Rename lighting_instanced shader (glsl100) to lighting_instancing

gtrxAC:
This is related to fixing the shaders_mesh_instancing example on web.


raysan5:
@gtrxAC thanks for the review! :)

PR #2807, opened 2022-11-23
Use const for pointer float array

planetis-m:


raysan5:
@planetis-m thanks for the improvement!

PR #2808, opened 2022-11-23
Correct types for rlBindImageTexture

planetis-m:
format is passed to rlGetGlTextureFormats which has a signed int parameter.


raysan5:

Issue #1945, opened 2021-08-23
[core] IME Support


mia-bentzen:
In order to be able to type text in japanese (and i assume chinese too, though i'm not familiar with it) you pretty much have to use to OS's IME (Input MEthod) to get input working right.

To give a small example of how it works: when you type が you don't actually type が directly, rather you type ga and it then automatically replaces that with が. It also comes up with suggestions based on a dictionary, and pressing space changes kana (basically letters) into kanji (basically complicated letters) so for instance you could have おんな, press space, and then it would change it into 女.

In order for the OS to know where and when to draw the IME suggestions box, the application has to tell it. The way SDL2 accomplishes this is with three methods: SDL_StartTextInput, SDL_StopTextInput, and SDL_SetTextInputRect.

The reason this would be nice to have is that without it japanese users simply can't type japanese text in any raylib application, and i'm not aware of any alternative or workaround.

orcmid:
That's very interesting.  I was thinking that we need a way to do that and there are also considerations with respect to accessibility.  This is also valuable for additional languages/keyboards.

The TLDR: This software is extensive and maybe larger than raylib :).  Figuring out a mapping and then being able to deal with SDL2 updates is going to be challenging.  I recommend that this issue be moved to Discussion since adapting SDL for inter-working with raylib is going to require extensive effort and substantial analysis and testing.  

=== Observations

The [SDL2 Tutorial](https://www.matsson.com/prog/sdl2-mingw-w64-tutorial.php) is tied to mingw-w64 and it involves libraries as well as a DLL that applications are expected to bundle.  It is useful that the code downloads are signed.  There is no public development server (GitHub or elsewhere).

It may take some work to adapt that to the raylib way of doing things, whether as an external or extra.  I assume that raylib should be moved to handle Unicode (probably UTF-8) in these cases.

I gave the source-code download a quick look.  The code is under a BSD-flavor license.  Considering that this is about a Windows-centric facility (I thought) the plethora of Unix-isms is startling (*.sh, *.m4, etc.) in areas involving use of Visual C..   Apparently there are comparable integrations for other platforms.   There are Visual Studio solutions and projects as part of the code.  Some may be dated (one is for VS2012).  The dates on files in the Zip appear to be from late July although it is not clear what that means.  There is no `.git` and change-control history.

mia-bentzen:
First off, I might be misunderstanding what you're saying, so absolutely correct me if i'm wrong!

[...]

PR #2809, opened 2022-11-25
[rcore] Draft: IME support

daipom:
sample

This is a fix for issue #1945.
We are mainly assuming Japanese input, but this feature can be used for other languages using IME as well.

We are implementing IME support features in GLFW now:

* https://github.com/glfw/glfw/pull/2130

This fix adds the corresponding APIs to raylib.

It may take a while to be merged in GLFW, so we have created this PR as a draft ahead of time.

# APIs to be added

* `void SetPreeditCallback(PreeditCallback callback)`
  * PreeditCallback: `void (*PreeditCallback)(int preeditLength, int *preeditString, int blockCount, int *blockSizes, int focusedBlock, int caret)`
  * Preedit version of `CharCallback`.
  * An app gets preedit info (preedit text, blocks, and caret position) by this callback and draws preedit.
  * The reason why we have an app register the callback directly, which is a different from `CharCallback`, is that this is not like `CharCallback`, where the input is resolved for each character, but rather requires managing the entire current preedit info at all times. Therefore, it is inefficient for an app to ask for preedit info every frame, so we use a callback to notify the change of preedit info.
* `void SetPreeditCursorRectangle(int x, int y, int w, int h)`
  * Set the area of the preedit cursor. This is used to decide the candidate window position that displays a list of conversion candidates. (For example, converting hiragana into kanji in Japanese).
  * In general, the position of this window should be managed by an app side by using this API.
* `void GetPreeditCursorRectangle(int *x, int *y, int *w, int *h)`
  * The getter version of `SetPreeditCursorRectangle`.
* `void ResetPreedit(void)`
  * Clear the current preedit. 
* `bool IsImeOn(void)`
  * Get if the IME is ON.
* `void SetImeStatus(bool on)`
  * Set IME ON/OFF.

**Feature to take candidate list (Win32 ONLY)**

For a detailed explanation, please see below.

* `void SetPreeditCandidateCallback(PreeditCandidateCallback callback)`
  * PreeditCandidateCallback: `void (*PreeditCandidateCallback)(int candidatesCount, int selectedIndex, int pageStart, int pageSize)`
  * The app takes the current candidate status by this callback.
  * To get the candidate text for each index, use `GetPreeditCandidate`.
* `int *GetPreeditCandidate(int index, int *textCount)`
  * Get the current candiadate text of the specified index.

# Specifications for each platform

Details are in the GLFW PR https://github.com/glfw/glfw/pull/2130.

**Win32**

* All APIs work fine.

**macOS**

* All APIs work fine.

**X11**

* Use over-the-spot style, so an app side doesn't need to draw preedit.
  * There are 2 main styles: on-the-spot and over-the-spot.
  * In on-the-spot style, an app side draws preedit, and this allows suitable drawing for each app. The other platforms use this style.
  * In over-the-spot style, the IME side draws preedit. This reduces what an app has to do, but it does not allow flexible display.
  * On X11, the behavior of on-the-spot style is not stable, so we use over-the-spot style.
* Only `SetPreeditWindowPosition` and `GetPreeditWindowPosition` work, and other APIs don't work.
* This is OK for this style because the IME side manages the drawing of preedit, so all an app side needs to do is managing the candidate window position.

**Wayland**

* Currently, we support the minimum required APIs: `SetPreeditCallback`, `SetPreeditWindowPosition`, and `GetPreeditWindowPosition`.

# Sample Application

We are implementing the sample application `RaylibIMEInputSampleApp`.

* https://github.com/clear-code/RaylibIMEInputSampleApp

We can check all newly added APIs in this application.

# Known issues

* X11: Some keys (arrow, Enter, BackSpace, ...) are passed to an app during preediting.
  * This will be fixed by https://github.com/glfw/glfw/pull/1972

# Additional planned fixes

Since the fix of GLFW is large, we will submit separate PRs to GLFW for the following features.
For this reason, we have also separated the fixes for raylib.

* Support soft fullscreen mode for IME features.
  * The candidate window doesn't be displayed correctly with fullscreen mode.
    * The default fullscreen is exclusive, and this causes the problem for IME displaying the candidate window.
  * We already have this fix for GLFW.
    * https://github.com/clear-code/glfw/tree/im-support-fullscreen 
  * The corresponding fix of raylib is this.
    * https://github.com/clear-code/raylib/tree/better-ime-support-fullscreen
    * This will add `FLAG_SOFT_FULLSCREEN` config flag to make the fullscreen not exclusive.

# Feature to take candidate list (Win32 ONLY)

Usually, the IME displays the candidate window, so the app side doesn't need to get canidate list.

However, sometimes, we need to display it on the app side.
For example, it could be for displaying it correctly on exclusive fullscreen, or for using a custom design.

This is especially common in games for Windows.

This feature is OFF by default.
We can use this by setting the `FLAG_MANAGE_PREEDIT_CANDIDATE` config flag.

Currently, this feature is available only on Win32, since this feature is especially common on Windows.
We'd like to make it available on every platform if possible, but it may be difficult because of platform specific issues.

This feature is also involved in the issue with fullscreen and IME.
If fullscreen is strongly exclusive, then the candidate window of IME can't be displayed correctly.
This problem is especially serious in Windows, so this feature is one workaround for fullscreen on Windows.

We also have another fix that reduces fullscreen exclusivity for the IME, but the fix of GLFW for Windows is tricky...
(Please see `Additional planned fixes`)

We can check this behavior in the sample application with `MANAGE_PREEDIT_CANDIDATE` cmake option.

* https://github.com/clear-code/RaylibIMEInputSampleApp/wiki/CMake-Build-Options#manage_preedit_candidate

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -1417,6 +1417,11 @@
           "description": "Set to run program in fullscreen"
         },
         {
+          "name": "FLAG_MANAGE_PREEDIT_CANDIDATE",
+          "value": 131072,
+          "description": "Set to manage the drawing of preedit candidates by the application side"
+        },
+        {
           "name": "FLAG_WINDOW_RESIZABLE",
           "value": 4,
           "description": "Set to allow resizable window"
@@ -3096,6 +3101,60 @@
         {
           "type": "char *",
           "name": "text"
+        }
+      ]
+    },
+    {
+      "name": "PreeditCallback",
+      "description": "",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int",
+          "name": "preeditLength"
+        },
+        {
+          "type": "int *",
+          "name": "preeditString"
+        },
+        {
+          "type": "int",
+          "name": "blockCount"
+        },
+        {
+          "type": "int *",
+          "name": "blockSizes"
+        },
+        {
+          "type": "int",
+          "name": "focusedBlock"
+        },
+        {
+          "type": "int",
+          "name": "caret"
+        }
+      ]
+    },
+    {
+      "name": "PreeditCandidateCallback",
+      "description": "",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int",
+          "name": "candidatesCount"
+        },
+        {
+          "type": "int",
+          "name": "selectedIndex"
+        },
+        {
+          "type": "int",
+          "name": "pageStart"
+        },
+        {
+          "type": "int",
+          "name": "pageSize"
         }
       ]
     },
@@ -4808,6 +4867,110 @@
         {
           "type": "int",
           "name": "key"
+        }
+      ]
+    },
+    {
+      "name": "SetPreeditCallback",
+      "description": "Set a callback for preedit",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "PreeditCallback",
+          "name": "callback"
+        }
+      ]
+    },
+    {
+      "name": "SetPreeditCursorRectangle",
+      "description": "Set the preedit cursor area that is used to decide the position of the candidate window",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int",
+          "name": "x"
+        },
+        {
+          "type": "int",
+          "name": "y"
+        },
+        {
+          "type": "int",
+          "name": "w"
+        },
+        {
+          "type": "int",
+          "name": "h"
+        }
+      ]
+    },
+    {
+      "name": "GetPreeditCursorRectangle",
+      "description": "Get the preedit cursor area",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int *",
+          "name": "x"
+        },
+        {
+          "type": "int *",
+          "name": "y"
+        },
+        {
+          "type": "int *",
+          "name": "w"
+        },
+        {
+          "type": "int *",
+          "name": "h"
+        }
+      ]
+    },
+    {
+      "name": "IsImeOn",
+      "description": "Check if IME is ON",
+      "returnType": "bool"
+    },
+    {
+      "name": "SetImeStatus",
+      "description": "Set IME status",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "bool",
+          "name": "on"
+        }
+      ]
+    },
+    {
+      "name": "ResetPreedit",
+      "description": "Reset preedit text",
+      "returnType": "void"
+    },
+    {
+      "name": "SetPreeditCandidateCallback",
+      "description": "Set a callback for preedit candidates",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "PreeditCandidateCallback",
+          "name": "callback"
+        }
+      ]
+    },
+    {
+      "name": "GetPreeditCandidate",
+      "description": "Get the text of the preedie candidate. This can be used only when `FLAG_MANAGE_PREEDIT_CANDIDATE` ConfigFlag is enabled on Win32int *",
+      "returnType": "int *",
+      "params": [
+        {
+          "type": "int",
+          "name": "index"
+        },
+        {
+          "type": "int *",
+          "name": "textCount"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


daipom:
It has not yet been merged into GLFW, so the fails of the tests are as expected.

Please see the wiki of `RaylibIMEInputSampleApp` about how to check this feature.

* https://github.com/clear-code/RaylibIMEInputSampleApp/wiki

raysan5:
@daipom Thank you very much for all the hard work on this new feature! I couldn't review it carefully yet, probably some function could be renamed by the example looks fantastic! Thanks! 😄

daipom:
Thanks!
If you have any questions about this feature or see something to be fixed, please let me know!

daipom:
I have just rebased this branch to the latest master.

daipom:
@raysan5

In connection with #2814, many characters can be entered at one time when using an IME.
So the current `MAX_CHAR_PRESSED_QUEUE` value is small for input using IME.

https://github.com/raysan5/raylib/blob/50a716c0d91f90e73f1fe2b910b3f7426a0d1360/src/rcore.c#L335-L337

It would be great if this could be made larger, such as `64`, or if it could be changed on the application side.
I would like to hear your opinion on this.

daipom:
I have fixed some APIs: `SetPreeditCursorRectangle`, `GetPreeditCursorRectangle`.

raysan5:
@daipom Hi! It's been quite long since this amazing improvement was proposed, what is the current state? Is there some workaround to allow this feature to be added in raylib?

Peter0x44:
This will definitely require some rebasing and modifications now the platform split is merged, to move the relevant code to rcore_glfw.c

DreepyYunky:
I think It'll be better to change `SetImeStatus()` to `TurnIMEOn` or `InitIME()`.

raysan5:
I just saw there was some update on this PR. I'm keeping it open for a bit longer, it seems lately GLFW is receiving several updates...

ashie:
Hi, I'm a collaborator of this work.
I've rebased this branch on top of current master (1fad8277a37b366841eff1497ca07a7042d4a014).
The previous one is backed up as https://github.com/clear-code/raylib/tree/4.2-2022-12-14%2Bbetter-ime-support
Since bundled GLFW isn't support IME yet, CI will fail as well as before.

> I just saw there was some update on this PR. I'm keeping it open for a bit longer, it seems lately GLFW is receiving several updates...

IME support of GLFW isn't merged yet, it's planned as a part of v3.5: https://github.com/glfw/glfw/milestone/25
So we need to wait next minor update of GLFW (current is v3.4 released 2 weeks ago).

ashie:
A commit is added automatically by CI:
https://github.com/raysan5/raylib/pull/2809/commits/ccf2ddf1f88807d534a1833a3b548a7d73ea4b66

I'm not sure we can leave it as is, or we should do something more or not.

raysan5:
@daipom @ashie Thanks for the update! Glad to see that this PR is still active! Very nice! 😄

raysan5:
@daipom @ashie Please, could you provide some update to this PR? Maybe @ColleagueRiley could help on RGFW side to make it possible?

seavhong0001 (changes requested):
pipi

seavhong0001 (changes requested):
Change

ColleagueRiley:
I agree with Raysan that we should investigate smoother ways to integrate these features into the Raylib API. I would like to implement these features into RGFW at some point. 

I'm also curious if SDL has a way of handling IME support.

hanaxars:
@ColleagueRiley SDL2 has a text input API which handles IME. I'm pretty sure SDL3 also has that. 
[https://wiki.libsdl.org/SDL2/Tutorials-TextInput#ime](https://wiki.libsdl.org/SDL2/Tutorials-TextInput#ime)

raysan5:
I'm afraid I'm closing this PR. It has been open for +3 years and GLFW does not seem to still support it easely, there is neither an easy implementation for raylib and not supported in too many target platforms.... It will be a reference for future implementations...

PR #2814, opened 2022-11-28
Fix array out of range

daipom:
Simply, the last indexes of the arrays are wrong.

Generating 16 characters or more of input at one time will destroy the other values of the structure.


raysan5:
@daipom Good catch! Thanks for the review!

daipom:
Thanks for the merging!

PR #2815, opened 2022-11-28
Fix wrong compile definition

daipom:
`MAX_KEY_PRESSED_QUEUE` is duplicated, and there is no `MAX_CHAR_PRESSED_QUEUE`.
Maybe this should be `MAX_CHAR_PRESSED_QUEUE`.


raysan5:
@daipom Yeah, that's it! Thanks!

daipom:
Thanks for the merging!

PR #2816, opened 2022-11-28
Updated rcore.c, renamed 'time' to 'nanoSeconds'

jtainer:
When PLATFORM_ANDROID, PLATFORM_RPI or PLATFORM_DRM are defined, there is a compilation error due to redefinition of the variable 'time' in the function GetTime. So the second instance of 'time' was changed to 'time_nsec'. This both resolves the name collision and more accurately describes what that variable represents.

Verified working on Raspberry Pi 3B+ with Raspbian Buster.


raysan5:
@jtainer Thanks for reporting! Just modified that function today. Please, could you rename variable to `nanoSeconds` to follow raylib coding conventions (snake_case is not used by raylib).

raysan5:
@jtainer thanks for the review!

PR #2820, opened 2022-11-30
Fix an issue when compiling for web

object71:
It would try to use the glfw on the system but we're cross-compiling for web where the implementation is provided by emscripten's team


raysan5:
@object71 Thanks for the review!

PR #2821, opened 2022-12-03
[example] Extruding a texture along a curve

JeffM2501:
This is an example that shows how to segment a curve and extrude a texture along it using a road as an example.

road_better_drag

textures_textured_curve


raysan5:
@JeffM2501 thanks for this great addition! 👍🙂

PR #2830, opened 2022-12-13
Prevent redefinition errors when the includes are not ordered

planetis-m:
In the Nim language bindings we have no control over the order the includes are generated. Code didn't compile in some cases, now works. Followup on https://github.com/raysan5/raylib/commit/b4fddf146b1f3d7c0548263594ad7283e6a9ac71

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -140,13 +140,13 @@
 // Other modules (raymath, rlgl) also require some of those types, so,
 // to be able to use those other modules as standalone (not depending on raylib)
 // this defines are very useful for internal check and avoid type (re)definitions
-#define RL_COLOR_TYPE
-#define RL_RECTANGLE_TYPE
-#define RL_VECTOR2_TYPE
-#define RL_VECTOR3_TYPE
-#define RL_VECTOR4_TYPE
-#define RL_QUATERNION_TYPE
-#define RL_MATRIX_TYPE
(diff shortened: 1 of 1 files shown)


planetis-m:
RL_COLOR_TYPE, RL_RECTANGLE_TYPE commented out as they seem unused?

raysan5:
@planetis-m I think this change affects readability of code, sorry, I prefer to avoid this approach for the moment.

planetis-m:
@raysan5 Would you prefer to move types in a separate header file for better readability?

raysan5:
@planetis-m Not really, I prefer to avoid multiple headers. Maybe I didn't understand exactly the problem with current approach, raylib is C and afaik this implementation follow C rules, is an specific issue with Nim language?

planetis-m:
The C file generated may have #include rlgl or raymath before raylib. That affects all the available bindings. Tldr: I can hack around it, but then it will break the corner case where one imports a type only without andly procedure.

raysan5:
But with that fix, if someone defines a custom `RL_VECTOR2_TYPE` could break raylib or even generate a security problem.
Actually, those defines were added for convenience (to allow `raymath` and `rlgl` to share those data types) but it's up to the user to include the files in the right order.

planetis-m:
Ok, I understand I will do the hacky way. Thanks for the explanation!

PR #2831, opened 2022-12-14
Vector2Angle fix for non-normalized vectors

veins1:


src/raymath.h
```diff
@@ -309,7 +309,9 @@ RMAPI float Vector2DistanceSqr(Vector2 v1, Vector2 v2)
 // Calculate angle from two vectors
 RMAPI float Vector2Angle(Vector2 v1, Vector2 v2)
 {
-    float result = -acos(v1.x*v2.x + v1.y*v2.y);
+    float v1Length = sqrtf((v1.x*v1.x) + (v1.y*v1.y));
+    float v2Length = sqrtf((v2.x*v2.x) + (v2.y*v2.y));
+    float result = -acos((v1.x*v2.x + v1.y*v2.y)/(v1Length*v2Length));
```
raysan5, line 314: Considering we operate with floats, this function should be `acosf()` (C99)

src/raymath.h
```diff
@@ -309,7 +309,9 @@ RMAPI float Vector2DistanceSqr(Vector2 v1, Vector2 v2)
 // Calculate angle from two vectors
 RMAPI float Vector2Angle(Vector2 v1, Vector2 v2)
 {
-    float result = -acos(v1.x*v2.x + v1.y*v2.y);
+    float v1Length = sqrtf((v1.x*v1.x) + (v1.y*v1.y));
```
raysan5, line 312: Just a note that this is probably the easier approach but also increases the cost of the function considerably...

raysan5:
@veins1 This improvement collides with #2831, I'm merging the last one because it avoids `sqrtf()`, that is a costly function and it's align with GLM. Still, I'm manually copying the code just in case, for future reference.

PR #2829, opened 2022-12-13
Fix Vector2Angle()

AlxHnr:
This fix ensures that the angle can be computed for all normalized vectors. But we still return negative values, which is not correct, but is required to maintain backwards compatibility.

Fixes #2828

PR #2832, opened 2022-12-15
Fix vector2angle

planetis-m:
Same as [glm::angle](https://github.com/g-truc/glm/blob/fc8f4bb442b9540969f2f3f351c4960d91bca17a/glm/gtx/vector_angle.inl#L20) Alternative to https://github.com/raysan5/raylib/pull/2829 https://github.com/raysan5/raylib/pull/2831


raysan5:
@planetis-m Thanks for the review.

PR #2836, opened 2022-12-19
[example] Writing into the depth buffer

BugraAlptekinSari:
An example to demonstrate how to write into the depth buffer so people don't have to search like I have in the future.

This example replaces the depth values by `1 - depth` to essentially invert the depth and make the farthest pixel appear on top.
box

Also I only tested the build on PLATFORM_WEB and PLATFORM_DESKTOP on a Linux machine. I don't know if it works on windows but I don't see why it shouldn't.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -520,7 +520,8 @@ SHADERS = \
     shaders/shaders_spotlight \
     shaders/shaders_hot_reloading \
     shaders/shaders_mesh_instancing \
-    shaders/shaders_multi_sample2d
+    shaders/shaders_multi_sample2d \
+    shaders/write_depth
 
 AUDIO = \
     audio/audio_module_playing \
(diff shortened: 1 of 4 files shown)


examples/Makefile
```diff
@@ -520,7 +520,8 @@ SHADERS = \
     shaders/shaders_spotlight \
     shaders/shaders_hot_reloading \
     shaders/shaders_mesh_instancing \
-    shaders/shaders_multi_sample2d
+    shaders/shaders_multi_sample2d \
+    shaders/write_depth
```
BugraAlptekinSari, line 524: I wrote write_depth instead of shaders_write_depth.
Time to figure out how to fix this

BugraAlptekinSari (commented):
Found a typo in the makefile

raysan5:
@BugraAlptekinSari Thanks for the example!

PR #2838, opened 2022-12-25
Add new Delphi/Lazarus bindings

turborium:
TurboRaylib - dynamic bindings of ray lib for Delphi/Lazarus


raysan5:
@turborium thanks for the addition!

PR #2839, opened 2022-12-28
Add a None value for GamepadButton enum

planetis-m:
GetGamepadButtonPressed may return -1 but the enum does not contained this value, for properly type checked languages this may create issues.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -684,7 +684,8 @@ typedef enum {
 
 // Gamepad buttons
 typedef enum {
-    GAMEPAD_BUTTON_UNKNOWN = 0,         // Unknown button, just for error checking
+    GAMEPAD_BUTTON_NONE = -1,           // No button pressed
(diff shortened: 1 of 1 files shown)


raysan5:
Gamepad button values are used as a button index in `IsGamepadButton*()` functions to check state, specifically in `CORE.Input.Gamepad.currentButtonState[gamepad][button]`. It should never be `-1` or assigned `-1`.

In any case, `GetGamepadButtonPressed()` shouldn't return `-1`. I changed it to return `0` (`GAMEPAD_BUTTON_UNKNOWN`) but depending on the situation it could require a redesign.

It seems the potentially affected function is `GamepadThread()` where `gamepadEvent.number` could be `0`.

planetis-m:
I guess GAMEPAD_BUTTON_NONE could be defined as an alias for UNKNOWN.

PR #2840, opened 2022-12-29
OpenGLES 2.0 support on PLATFORM_DESKTOP

wtnbgo:
In the current version of raylib, if you specify OPENGL_VERSION="ES 2.0" when compiling the DESKTOP version, the GL Context at runtime is set correctly by glfw, but the OpenGL ES2 header file is missing, so it cannot be compiled 

This is a diff to make it possible to compile and work using glad's gles2.h

```
# build on Windows/Linux/OSX
cmake -DOPENGL_VERSION="ES 2.0" -B build
cmake --build build
cd build/examples
# windows
./Debug/raylib_opengl_interop.exe
# linux/mac
./raylib_opengl_interop
```

On environments that do not support the OpenGL ES2 context, copying and deploying libEGL.dll and libGLESv2.dll (libEGL.dylib, libGLESv2.dylib on Mac OS) from ANGLE (or Chrome, etc.) and it will be loaded by egl and work correctly


src/rlgl.h
```diff
@@ -789,10 +789,16 @@ RLAPI void rlLoadDrawQuad(void);     // Load and draw a quad
 #endif
 
 #if defined(GRAPHICS_API_OPENGL_ES2)
+
+    #if defined(PLATFORM_DESKTOP)
+    #define GLAD_GLES2_IMPLEMENTATION
+    #include "external/gles2.h"
```
raysan5, line 795: This file should probably be named `glad_gles2.h` instead for clarity and to avoid confusion with gl specific headers.

src/rlgl.h
```diff
@@ -2144,6 +2150,12 @@ void rlLoadExtensions(void *loader)
 #endif  // GRAPHICS_API_OPENGL_33
 
 #if defined(GRAPHICS_API_OPENGL_ES2)
+
+    #if defined(PLATFORM_DESKTOP)
+    if (gladLoadGLES2((GLADloadfunc)loader) == 0) TRACELOG(RL_LOG_WARNING, "GLAD: Cannot load OpenGL ES2.0 functions");
```
raysan5, line 2155: Why is it required to load OpenGL ES2 extensions? Afaik the official `gl2.h`/`gl2ext.h` headers already provide all the required function pointers for raylib.

examples/others/raylib_opengl_interop.c
```diff
@@ -26,21 +26,29 @@
...
-#else
-    #include "glad.h"       // Required for: OpenGL functionality 
-#endif
-#include "raymath.h"        // Required for: MatrixMultiply(), MatrixToFloat()
-
 #if defined(PLATFORM_DESKTOP)
-    #define GLSL_VERSION            330
+    #if defined(GRAPHICS_API_OPENGL_ES2)
+        #include "gles2.h"       // Required for: OpenGL functionality 
+        #define glGenVertexArrays glGenVertexArraysOES
```
raysan5, line 32: Why those functions pointers need to be defined here? rlgl should probably do that mapping internally.

src/rlgl.h
wtnbgo: I renamed it.

src/rlgl.h
wtnbgo: In PLATFORM_DESKTOP, gl2.h / gl2ext.h usually does not exist.
Instead, glad_gles2.h is used, in which case the defined all functions are 
only available when dynamically loaded at runtime. this part is for loading functions

Should we add a separate definition for environments that have gl2.h / gl2ext.h?
・Add "ES2.0 GLAD" to the OPENGL_VERSION specification.
・In that case, define USE_GLADES2 and GRAPHICS_API_OPENGL_ES2.
・Adjust ifdef in the related section as follows
#if defined(PLATFORM_DESKTOP) && defined(USE_GLADES2)

examples/others/raylib_opengl_interop.c
wtnbgo: I will consider moving it to rlgl.h

raysan5: ok, nice

raysan5:
@wtnbgo Thanks for the improvement! I'm merging and reviewing the missing bits.

PR #2845, opened 2022-12-31
core_loading_thread example join thread on completion

planetis-m:
otherwise it leaks the thread


raysan5:

PR #2849, opened 2023-01-02
Use explicit atomics

planetis-m:
makes dataProgress atomic as well.

### DIFF
--- a/examples/core/core_loading_thread.c
+++ b/examples/core/core_loading_thread.c
@@ -69,7 +69,7 @@ int main(void)
             case STATE_LOADING:
             {
                 framesCounter++;
-                if (atomic_load_explicit(&dataLoaded, memory_order_acquire))
+                if (atomic_load_explicit(&dataLoaded, memory_order_relaxed))
                 {
                     framesCounter = 0;
                     int error = pthread_join(threadId, NULL);
@@ -149,7 +149,7 @@ static void *LoadDataThread(void *arg)
     }
 
     // When data has finished loading, we set global variable
-    atomic_store_explicit(&dataLoaded, true, memory_order_release);
+    atomic_store_explicit(&dataLoaded, true, memory_order_relaxed);
 
     return NULL;
 }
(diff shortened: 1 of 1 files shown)


examples/core/core_loading_thread.c
```diff
@@ -84,8 +84,8 @@ int main(void)
                 if (IsKeyPressed(KEY_ENTER))
                 {
                     // Reset everything to launch again
-                    atomic_store(&dataLoaded, false);
-                    dataProgress = 0;
+                    atomic_store_explicit(&dataLoaded, false, memory_order_relaxed);
+                    atomic_store_explicit(&dataProgress, 0, memory_order_relaxed);
```
planetis-m, line 88: There are no more threads running at this point.

examples/core/core_loading_thread.c
```diff
@@ -23,10 +23,10 @@
 
 // Using C11 atomics for synchronization
 // NOTE: A plain bool (or any plain data type for that matter) can't be used for inter-thread synchronization
-static atomic_bool dataLoaded = ATOMIC_VAR_INIT(false); // Data Loaded completion indicator
+static atomic_bool dataLoaded = false; // Data Loaded completion indicator
 static void *LoadDataThread(void *arg);     // Loading data thread function declaration
 
-static int dataProgress = 0;                // Data progress accumulator
+static atomic_int dataProgress = 0;                // Data progress accumulator
```
planetis-m, line 29: dataProgress is just a counter so every load/store is relaxed

examples/core/core_loading_thread.c
```diff
@@ -69,7 +69,7 @@ int main(void)
             case STATE_LOADING:
             {
                 framesCounter++;
-                if (atomic_load(&dataLoaded))
+                if (atomic_load_explicit(&dataLoaded, memory_order_acquire))
                 {
                     framesCounter = 0;
                     int error = pthread_join(threadId, NULL);
```
planetis-m, line 75: Here only pthread_join matters and I guess it can't be reordered anyway?

planetis-m:
ready.

raysan5:

orcmid:
**Side Comment.**  The atomics arrived with C11, so that must be accounted for in compilation dependencies.  (rayLib already has such dependencies with regard to structure initializations.)  Apparently there must also be `#include <stdatomic.h>` according to ISO C11 and C17.

I just confirmed that `stdatomic.h` is available to VS 2022 Build Tools.  I have not checked older Visual Studio versions.

PR #2850, opened 2023-01-02
Add the ability to change screenshot key

BlueStaggo:
Currently in Raylib, F12 is always set as the screenshot key, so there is no way that a developer can use the F12 key for a different action without taking a screenshot. This pull request introduces a `SetScreenshotKey` function, very much like `SetExitKey`, that allows a developer to change the key used to take a screenshot or even disable screenshotting. It is useful for when a developer wants to use a different method of screenshotting, take screenshots of a RenderTexture or put screenshots into a different directory. For developers using bindings, this is a good way to disable screenshotting without access to the `SUPPORT_SCREEN_CAPTURE` macro.


raysan5:
@BlueStaggo I have my doubts about this addition. Screenshot feature was added just for convenience for my students, it's not expected to be enabled in production code. Actually, `TakeScreenshot()` function is provided to allow users to call it when required.

Adding `SetScreenshotKey()` could be confusing for many users in the case the internal screenshot mechanism was disabled, it will suppose a useless function and also an extra variable unused in the code.

I'm afraid I'm not merging this PR.

PR #2856, opened 2023-01-13
Add RAYLIB_VERSION numbers to raylib.h

RobLoach:
Ran into an issue in raylib-cpp where a user was using raylib 4.5-dev, even though the library currently only targets 4.2. With having RAYLIB_VERSION_MAJOR and RAYLIB_VERSION_MINOR, we will be able to target different versions of raylib in different ways with the same code, using C preprocessor conditionals.

For example:
``` c
#if RAYLIB_VERSION_MAJOR >= 4 && RAYLIB_VERSION_MINOR >= 5
newColor = ColorTint(BLUE, RED);
#else
TraceLog(LOG_INFO, "The color should be tinted, but this isn't supported in raylib <= 4.2");
#endif
```

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -81,6 +81,9 @@
 
 #include <stdarg.h>     // Required for: va_list - Only used by TraceLogCallback
 
+#define RAYLIB_VERSION_MAJOR 4
+#define RAYLIB_VERSION_MINOR 5
+#define RAYLIB_VERSION_PATCH 0
 #define RAYLIB_VERSION  "4.5-dev"
 
 // Function specifiers in case library is build/used as a shared library (Windows)


orcmid:
This is an interesting situation, with dependence on up-level features.  It could also be related to breaking changes up-level.

The problem I have in the example is that a compile-time problem is handled by a runtime TraceLog, sort of like an implanted assert failure.

A narrower, more-specific solution would be a way to tell if ColorTint has been declared, and then take a preprocessor action.  That's rather fine-grained, so could become cumbersome.  It also depends on a developer paying close attention to version differences (but so does checking version numbers).  I'm also not a big fan of complex #if cases in headers and code.

I assume if RAYLIB_VERSION_MAJOR and/or _MINOR are not defined, the test will fail without consequence.

While this modification is benign, I think a more useful one is with a single sequence numeral, such as RAYLIB_VERSION >= 0405.  I'm not certain the 4.5 format would work properly in pre-processors though.  

Probably the only thing I would use this for is in creating the text of an About message for the application, or maybe other logging cases, just to support external troubleshooting and support.  

And that's good enough reason to have it, in my book.

raysan5:
@RobLoach I'm merging this improvement but I'm not a big fan of that kind of defines... to cover all possible use cases usually more defines are required and I got the feeling it could end-up with unexpected versioning issues... anyway, let's see how it works.

PR #2857, opened 2023-01-14
Fixed bug : touches become sticky

ImazighenGhost:
Touches became sticky and didn't disappear after using more than 2 fingers, fixed by getting the touch count of how many fingers are on the screen, and only looping through the available/pressed down touch points instead of looping through the maximum touch points. Tested with more than 10 touch points, and with different MAX points value, working perfectly.


raysan5:
@ImazighenGhost Thanks for the review!

PR #2860, opened 2023-01-15
Updated Touch events

ImazighenGhost:
Touch events updated in both rcore.c and raylib.h
now it is possible to check if a Touch point was pressed or released, or either pressed once as in TAP or released after a press
CAUTION : these functions wont work on desktop as it will add another layer of complexity to Raylib, some mouse functions are mapped to touches in Raylib, but it should be all up to the user if they want to map Mouse input to Touches. This is irrelevant with Gesture as they are two different structures.
PS : this was tested on 3 mobile devices and proved to work normally.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -3907,6 +3907,40 @@ int GetTouchPointCount(void)
     return CORE.Input.Touch.pointCount;
 }
 
+// Check if a Touch point is pressed once (if the finger touch got in contact with the the screen once, a.k.a TAP, also found in gestures)
+bool IsTouchPointPressed(int index)
+{
+    // Compare the previous and current Touch point state
+    if ((CORE.Input.Touch.currentTouchState[index] == 1) && (CORE.Input.Touch.previousTouchState[index] == 0)) return true;
+
+    return false;
+}
+
+// Check if a Touch point has been released once (if the finger touch was in contact with the screen and now is not anymore)
+bool IsTouchPointReleased(int index)
+{
+    // Compare the previous and current Touch point state
+    if ((CORE.Input.Touch.currentTouchState[index] == 0) && (CORE.Input.Touch.previousTouchState[index] == 1)) return true;
+
+    return false;
+}
+
+// Check if a Touch Point is being pressed (if the finger touch is in contact with the screen)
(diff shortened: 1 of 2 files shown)


src/rcore.c
```diff
@@ -3907,6 +3907,40 @@ int GetTouchPointCount(void)
...
+}
+
+// Check if a Touch Point is being pressed (if the finger touch is in contact with the screen)
+bool IsTouchPointDown(int index)
+{
+    // Get the state of the Touch point
+    if (CORE.Input.Touch.currentTouchState[index] == 1) return true;
+
+    return false;
+}
```
JeffM2501, line 3935: why the if and not just return the test result?

ImazighenGhost: I was following the code syntax in the Mouse events, something to make code easier to read for future users, should I just return `CORE.Input.Touch.currentTouchState[index]` ? I mean the state can be something different in the future or something.

raysan5: Yeah, I usually kept the `result` as a separate variable for readibility...

raysan5:
I'm not sure if this implementation is actually valid, the touch system requires a full review. The index just stores the touch order but not the touch identifier, I think they can differ and so those checks won't be correct.

ImazighenGhost:
I will delve deeper into the implementation, thanks for the input people!

src/rcore.c
orcmid: So

```C
if ((CORE.Input.Touch.currentTouchState[index] == 0) && (CORE.Input.Touch.previousTouchState[index] == 1)) return true;

return false;
```

can simply be

``` C
return (CORE.Input.Touch.currentTouchState[index] == 0) && 
           (CORE.Input.Touch.previousTouchState[index] == 1);
```
which is the minimalist solution, since the expression produces the correct truth (**bool**) value.  Alternatively, you can do

```
result = (CORE.Input.Touch.currentTouchState[index] == 0) && 
             (CORE.Input.Touch.previousTouchState[index] == 1);
return result;
```

having declared *result* to be the same type specified for the **function** result.

I prefer the minimalisms and you'll find them in my own code.  Here it is more important to follow the standard pattern preferred by Ramon.

PR #2862, opened 2023-01-16
[rlgl] Add `rlTextureParameters` alternative for cubemaps

GithubPrankster:
I was using a skybox for my renderer and noticed the cubemap was bilinear-filtered, which was not desired. However, I found that the available `rlgl` function for changing texture parameters only works for normal textures. I made a `rlCubemapParameters` function to take care of this.

misery_ftmVogFFum
misery_0QXf3HDozP


raysan5:
@GithubPrankster Thanks for the improvement! Actually, I was thinking about some other alternative to avoid a duplicate function for Cubemap textures but for now it will be ok...

PR #2867, opened 2023-01-19
[MODELS] Fix crash with some GLTF models

JeffM2501:
When a buffer is read using cgltf_load_buffer_base64, the data is a raw void*, so we can't call `cgltf_free`. `cgltf_free `assumes the data is a `cgltf_data `structure and will try to free the various buffers in that structure. 
This causes a crash with some files (attached)
[house.zip](https://github.com/raysan5/raylib/files/10453346/house.zip)

This PR calls a normal free for the data, so it's properly deallocated so it won't crash.
I could not find a pure free method in cgltf, so I just called raylib's free since it looks like that is what is installed for cgltf's allocators anyway.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4645,7 +4645,7 @@ static Image LoadImageFromCgltfImage(cgltf_image *cgltfImage, const char *texPat
                 if (result == cgltf_result_success)
                 {
                     image = LoadImageFromMemory(".png", (unsigned char *)data, outSize);
-                    cgltf_free((cgltf_data*)data);
+                    MemFree((cgltf_data*)data);
                 }
             }
         }


raysan5:
@JeffM2501 This issue could be a bit tricky... 

`data` should be freed by the same allocation library used to allocate it. In this case, `cgltf` calls `cgltf_default_alloc()` that calls `CGLTF_MALLOC` that `rmodels` map to `RL_CALLOC`, so, calling `MemFree()` that calls `RL_FREE` end-up using the right allocators pair... BUT there could be some situation, like user setting up custom allocation/free pairs on raylib, that could let to undesired effects.

Clearly `cgltf_free()` can not be used to unload `data`. Ideally, `cgltf_load_buffer_base64()` should provide a unload function equivalent... In the meantime, I'm merging this fix. 

Related to https://github.com/jkuhlmann/cgltf/issues/200

PR #2871, opened 2023-01-20
[CORE] Fix warnings in MSVC

JeffM2501:
Just another fixup of warnings that show in visual studio.

1) par shapes, jar xm, and qoi have warnings that can't be fixed, so this PR suppresses them on the MSVC compiler
2) casting as normal
3) some of the new gltf code was using unsigned ints to iterate signed sizes.


raysan5:
@JeffM2501 thanks for the review! probably the pragma-ignored warnings should be fordwarded to the related libraries... in any case I plan to update all libraries for a new release in the following days.

Issue #2869, opened 2023-01-19
[models] `LoadMaterials` not returning any materials


Anut-py:
- [x] I tested it on latest raylib version from master branch
 - [X] I checked there is no similar issue already reported
 - [X] My code has no errors or misuse of raylib

### Issue description

In the source code for `LoadMaterials`, `*materials` is set to `NULL`. There is also a TODO comment saying "TODO: Process materials to return". `LoadModel` properly loads materials from `.mtl` files, so I'm fairly sure that `LoadMaterials` was just overlooked. 

### Environment

Windows 11, OpenGL 3.3, AMD Radeon Graphics

### Code Example

```c
int num = 0;
Material* materials = LoadMaterials("./assets/Model.mtl", &num); // Model.mtl and Model.obj load properly with `LoadModel`

printf("Number of materials: %d\n", num); // Output: 3
printf("Materials pointer: %d\n", materials); // Output: 0
printf("First material param: %f\n", materials[0].params[0]); // Fails
```

JeffM2501:
I have made a PR that fixes this (PR#2872), but I can also see reason to remove the function since it hasn't worked, thus nobody would be using it.

raysan5:
@JeffM2501 I left that function there for the future in case some specific material file-format got popular, unfortunately, beside the old .mtl it seems no material file-formats are popular... not sure how useful is this function but thank you very much for the implementation.

JeffM2501:
Yeah I had a feeling it was not super useful, we may want to just remove the function then.

raysan5:
@JeffM2501 I think we can keep it for 4.5.

PR #2872, opened 2023-01-20
[Models] Fix LoadMaterials

JeffM2501:
The function `LoadMaterials `did not do anything with the materials parsed from the MTL file. This PR moves the code that the OBJ reader uses for materials into a function that can be used by both `LoadObj `and `LoadMaterials`.

I'm not sure if you want to keep `LoadMaterials `as a function because nobody has used it, but if you do, this PR makes it work. 
I can see justification for removing `LoadMaterials `too.

This should resolve issue #2869

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1851,6 +1851,42 @@ bool ExportMesh(Mesh mesh, const char *fileName)
     return success;
 }
 
+#if defined(SUPPORT_FILEFORMAT_MTL)
+// Process obj materials
+
+static void ProcessOBJMaterials(Material* rayMaterials, tinyobj_material_t* materials, int materialCount)
+{
+	// Init model materials
+	for (unsigned int m = 0; m < materialCount; m++)
+	{
+		// Init material to default
+		// NOTE: Uses default shader, which only supports MATERIAL_MAP_DIFFUSE
+		rayMaterials[m] = LoadMaterialDefault();
+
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 thanks for the implementation, we can keep it for now but I also have my doubts on the usefulness of this function.

hanaxar:
The function ```ProcessOBJMaterials``` is in ```#if defined(SUPPORT_FILEFORMAT_MTL)``` block now, and If I comment out ```SUPPORT_FILEFORMAT_MTL``` in ```config.h``` I get couple of compiler errors and warnings because of several ```ProcessOBJMaterials``` function calls in ```rmodels.c```.

raysan5:
@hanaxar @JeffM2501 I'm reviewing it right now

JeffM2501:
I think that the code in the obj reader should be #defed out to not try and load materials if MTL files are not supported. If you support OBJ but not MTL, then you don't support materials in OBJ files at all.

Other options are
1) remove MTL as a separate file that is 'supported' and assume that OBJ support also means MTL support since it's a sidecar file.
2) make the #def around `ProcessOBJMaterials  `be OBJ or MTL support

raysan5:
@JeffM2501 Just reviewed it on commit https://github.com/raysan5/raylib/commit/5b3c5e1a16c2d0f97e407542f1a0837b9d5b7c71, it should be ok now

PR #2874, opened 2023-01-20
Fix Android sound support

planetis-m:
My solution is based on this SO question https://stackoverflow.com/questions/15328751/android-macro-suddenly-not-defined


raysan5:
@planetis-m this solution should be provided to `miniaudio` main repo...

PR #2876, opened 2023-01-20
[rlgl] Stub `rlCubemapParameters` if under GL 1.1

GithubPrankster:
`ARB_texture_cube_map` only became core in OpenGL 1.3. A friend experimenting with porting my image viewer to GL 1.1 found that the function I recently added threw a error. Now it should only be usable if GL1.1 isn't the version being used.


raysan5:
@GithubPrankster thanks for the review!

PR #2878, opened 2023-01-21
[models] Fix M3D vertex color import.

### LINKED ISSUE #2855: [models] M3D import does not load vertex colors correctly

### Issue description

Loading any `.m3d` correctly loads all data except vertex colors, which are left blank white instead. The model thus cannot match what was exported by the [official m3d blender exporter](https://gitlab.com/bztsrc/model3d/-/tree/master/blender).

### Environment

Desktop, Windows 10, OpenGL 3.3
```
INFO:     > Renderer: NVIDIA GeForce GTX 1060 3GB/PCIe/SSE2
INFO:     > Version:  3.3.0 NVIDIA 527.56
INFO:     > GLSL:     3.30 NVIDIA via Cg compiler
```

### Issue Screenshot

image
*Note the warped texturing is a stylistic choice rather than wrong behavior.*

image
*RenderDoc screenshot showing that all vertex colors are white.*

And original (in Blender 3.4.1):
image

```
Color Map Chunk
  Magic:  CMAP
  Length: 32
  Data:   #ff0000c7 #ff00c7c7 #ffffffff #ffc700c7 #ffcccccc #ff7f7f7f
```
Output from the m3d [web validator](https://bztsrc.gitlab.io/model3d/validator/) confirming the colors were indeed stored.

### Code Example

```c
// Note the shader must be using `fragColor` for it to be visible.
Shader shader = LoadShader(0, "shader.fs");
...
Model model = LoadModel("path/to/cube.m3d");
model.materials[1].shader.id = shader.id;
...
DrawModel(model, position, 1.0f, WHITE);
```
Here's the cube itself.
[cube.zip](https://github.com/raysan5/raylib/files/10407749/cube.zip)

GithubPrankster:
Fixes #2855. There's been no reply from bztsrc so I'm going ahead with the PR.


raysan5:
@GithubPrankster Many 3d models do not require vertex colors but this solution forces it, allocating an amount of memory that many times is not required. I think it should only be loaded if vextex color data is provided by M3D model data.

GithubPrankster:
I'll look into how I can do this.

GithubPrankster:
I tested this on a model with both, then either and neither vertex colors nor materials and it worked fine. The previous solution would just load white in the absence of a material and not much else. I still hope for some return from bztsrc but for now it does the job.

src/rmodels.c
```diff
@@ -5663,16 +5661,13 @@ static Model LoadM3D(const char *fileName)
...
-            {
-                // without vertex color (full transparency), we use the default color
-                if (m3d->vertex[m3d->face[i].vertex[0]].color & 0xFF000000)
-                    memcpy(&model.meshes[k].colors[l * 12 + 0], &m3d->vertex[m3d->face[i].vertex[0]].color, 4);
-                if (m3d->vertex[m3d->face[i].vertex[1]].color & 0xFF000000)
-                    memcpy(&model.meshes[k].colors[l * 12 + 4], &m3d->vertex[m3d->face[i].vertex[1]].color, 4);
-                if (m3d->vertex[m3d->face[i].vertex[2]].color & 0xFF000000)
-                    memcpy(&model.meshes[k].colors[l * 12 + 8], &m3d->vertex[m3d->face[i].vertex[2]].color, 4);
-            }
+            // without vertex color (full transparency), we use the default color
```
raysan5, line 5664: I think this code should only execute if color buffer has been allocated...

raysan5:
@GithubPrankster I'm checking M3D specs and I'm confused with color map, it seem to provide a color index array for every vertex... but I'm confused about it...

Refs: 
 - https://gitlab.com/bztsrc/model3d/blob/master/docs/m3d_format.md#color-map
 - https://gitlab.com/bztsrc/model3d/blob/master/docs/m3d_format.md#vertex-chunk

src/rmodels.c
GithubPrankster: Onto it!

bztsrc:
Hi,

Sorry about the delay, I'm pretty busy with other projects.

So...

> I'm checking M3D specs and I'm confused with color map, it seem to provide a color index array for every vertex... but I'm confused about it...

Color map is optional. Usually you won't use it, you would simply store RGBA pixels and that's it. Color map is for compressing the model even further, using a predefined set of colors only, and referencing that with an index.

Depending on the bits in the header, the vertex chunk is either storing no vertex color, just index, or the RGBA pixels itself.

About this part:
```c
if (mi == M3D_UNDEF)
```
Vertex colors should only be used if there's no material associated (that's because with a material one should use the diffuse color specified in that material or the texture).

I'll try to take time to assess this issue properly, but for now I'd say it is not necessarily a bug if you can see vertex colors in Blender, but you've also set up a material, and therefore no colors in raylib.

Cheers,
bzt

bztsrc:
About the spec, I've thought I explained it, but please help to add more explanation, what would make your confusion disappear?

The spec now reads (in the header section):
```
For all the other index, size and count types the bits are:
BitVal	Type	Description
00	uint8_t	from 0 to 255
01	uint16_t	from 0 to 65535
10	uint32_t	from 0 to 2^32
11	-	(not defined)
Here "not defined" does not mean that the value is not defined by the specification, but that the corresponding chunk
in the file does not exists so the index is not defined in the model.
```
And in the [color map](https://gitlab.com/bztsrc/model3d/blob/master/docs/m3d_format.md#color-map) section:
```
This is a unique chunk, can occur only once in the model. It only exists if ci_t is uint8_t or uint16_t.
```

A simple example code to decode the vertex chunk's color looks like this (assuming cmap points to the cmap chunk's data or NULL):
```c
switch(ci_t) {                                          /* get RGBA color if exists */
    case 1:  vertex[i].rgba = cmap[ptr[0]]; ptr++; break;
    case 2:  vertex[i].rgba = cmap[*((uint16_t*)ptr)]; ptr += 2; break;
    case 4:  vertex[i].rgba = *((uint32_t*)ptr); ptr += 4; break;
    default: vertex[i].rgba = 0xFF808080;  /* no ptr increment and default color (should have a material with diffuse color/texture) */
}
```
Does this explanation help?

Cheers,
bzt

raysan5:
Hi @bztsrc! Thank you very much for the detailed explanation, now I understand it.

 > Vertex colors should only be used if there's no material associated (that's because with a material one should use the diffuse color specified in that material or the texture).

Agree.

 > Depending on the bits in the header, the vertex chunk is either storing no vertex color, just index, or the RGBA pixels itself.

This line made it more clear for me.

bztsrc:
Hi @raysan5!

You're welcome! I've checked the patch, looks good, but there's one little bug in it (line [5636](https://github.com/GithubPrankster/raylib/blob/e85e623e014147629ecbf8cb61720cbc0c45dc3e/src/rmodels.c#L5636)):
```c
                if(m3d->cmap != NULL || mi != M3D_UNDEF)
                {
                    model.meshes[k].colors = RL_CALLOC(model.meshes[k].vertexCount*4, sizeof(unsigned char));
                }
```
This won't allocate model.meshes[k].colors even if vertex color exists, but not cmap encoded. Should be simply
```c
                if(mi != M3D_UNDEF)
                {
                    model.meshes[k].colors = RL_CALLOC(model.meshes[k].vertexCount*4, sizeof(unsigned char));
                }
```
Now this will work perfectly (you don't have to care about the color encoding, the M3D SDK will fill up `m3d->vertex[x].color` with RGBA pixels for you, no matter how that color is actually stored in the file).

If the intention is to allow vertex colors even with materials set (as the original issue suggests; that would be against the M3D spec, but if this helps raylib users I'm not against it), then just remove the `if` entirely and always allocate the color array unconditionally.

Cheers,
bzt

raysan5:
@bztsrc Thank you very much for the fix, implemented in commit https://github.com/raysan5/raylib/commit/21961a786df76a86302942f3da8f7c08b7f31863

 > Now this will work perfectly (you don't have to care about the color encoding, the M3D SDK will fill up m3d->vertex[x].color with RGBA pixels for you, no matter how that color is actually stored in the file).

At this moment we are doing:

```c
// Without vertex color (full transparency), we use the default color
if (model.meshes[k].colors != NULL)
{
    if (m3d->vertex[m3d->face[i].vertex[0]].color & 0xFF000000)
        memcpy(&model.meshes[k].colors[l*12 + 0], &m3d->vertex[m3d->face[i].vertex[0]].color, 4);
    if (m3d->vertex[m3d->face[i].vertex[1]].color & 0xFF000000)
        memcpy(&model.meshes[k].colors[l*12 + 4], &m3d->vertex[m3d->face[i].vertex[1]].color, 4);
    if (m3d->vertex[m3d->face[i].vertex[2]].color & 0xFF000000)
        memcpy(&model.meshes[k].colors[l*12 + 8], &m3d->vertex[m3d->face[i].vertex[2]].color, 4);
}
```

Is that correct?

 > If the intention is to allow vertex colors even with materials set (as the original issue suggests; that would be against the M3D spec, but if this helps raylib users I'm not against it), then just remove the if entirely and always allocate the color array unconditionally.

I prefer to adjust to M3D specs.

bztsrc:
> Is that correct?

Perfect!

> I prefer to adjust to M3D specs.

Ok!

Cheers,
bzt

ps: sorry about the delay, I'm extremely busy these days.

PR #2881, opened 2023-01-21
Fix to use TRACELOG() instead of TraceLog() for internal modules

RobLoach:
There were a few raylib modules that continued to use `TraceLog()` instead of the `TRACELOG()` macro. This change ensures that all the internal raylib modules use the `TRACELOG()` pattern consistently.


raysan5:
@RobLoach Ouch! Good catch! Thanks for the review! 👍😄

PR #2882, opened 2023-01-22
[models] Load bone names from IQM file if available

### LINKED ISSUE #2863: [rmodel] IQM Animations do not load joint names
### Issue description

I'm trying to search for bone transforms using bone name, but `LoadModelAnimationsIQM` discards joint names for some reason and just writes `"ANIMJOINTNAME"` to `BoneInfo.name`

### Code Example
https://github.com/raysan5/raylib/blob/master/src/rmodels.c#L4504

PencilAmazing:
Simply checks if joints were loaded in with animations file, and copies string data to name.
One issue is that maybe we should be using `strcpy_s` instead of raw memcpy?
Fixes #2863

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4498,8 +4498,8 @@ static ModelAnimation *LoadModelAnimationsIQM(const char *fileName, unsigned int
     memcpy(framedata, fileDataPtr + iqmHeader->ofs_frames, iqmHeader->num_frames*iqmHeader->num_framechannels*sizeof(unsigned short));
 
     // joints
-    IQMJoint* joints = RL_MALLOC(iqmHeader->num_joints * sizeof(IQMJoint));
-    memcpy(joints, fileDataPtr + iqmHeader->ofs_joints, iqmHeader->num_joints * sizeof(IQMJoint));
+    IQMJoint *joints = RL_MALLOC(iqmHeader->num_joints*sizeof(IQMJoint));
+    memcpy(joints, fileDataPtr + iqmHeader->ofs_joints, iqmHeader->num_joints*sizeof(IQMJoint));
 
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -4501,7 +4511,9 @@ static ModelAnimation *LoadModelAnimationsIQM(const char *fileName, unsigned int
 
         for (unsigned int j = 0; j < iqmHeader->num_poses; j++)
         {
-            strcpy(animations[a].bones[j].name, "ANIMJOINTNAME");
```
raysan5, line 4504: In case no bone name was provided, probably `ANIMJOINTNAME` should be used...

src/rmodels.c
```diff
@@ -4491,6 +4497,10 @@ static ModelAnimation *LoadModelAnimationsIQM(const char *fileName, unsigned int
     //fread(framedata, iqmHeader->num_frames*iqmHeader->num_framechannels*sizeof(unsigned short), 1, iqmFile);
     memcpy(framedata, fileDataPtr + iqmHeader->ofs_frames, iqmHeader->num_frames*iqmHeader->num_framechannels*sizeof(unsigned short));
 
+    // joints
+    IQMJoint* joints = RL_MALLOC(iqmHeader->num_joints * sizeof(IQMJoint));
```
raysan5, line 4501: Please, could you review raylib coding style, `*` pointer should be placed in the variable and multiplications have no spaces.

raysan5:
@PencilAmazing Thanks for the improvement, just added some reviews.

raysan5:
@PencilAmazing Thank you very much for the review! :)

PencilAmazing:
@raysan5 Thanks for the review!

PR #2883, opened 2023-01-23
Add wayland support

planetis-m:
I have limited experience with make, please review carefully.


raysan5:
@planetis-m Wow! Is that really required to run raylib under Wayland protocol??? 

I think I never used that kind of system calls on a Makefile, did you test it? does it work? It looks quite cumbersome...

planetis-m:
It compiles but haven't tested yet, except with my own bindings which work fine. No glfw doesn't provide the required files they need to be generated. I could replace the calls to pkg-config to use fixed paths but cmake does the exact same thing.

raysan5:
@planetis-m Ok, just let me know when you test it on a Wayland-based distro for confirmation and I'll merge it.

planetis-m:
Yep it works!

raysan5:
@planetis-m thanks for the improvement! :)

PR #2888, opened 2023-01-25
[CORE] Add include guard to config.h

JeffM2501:
The file config.h doesn't have any kind of include guards. This means that if some external library (like raylib extras) needs to check options in in raylib, it will get warning/errors if it tries to include config.h.

This PR simply adds a guard to the file to prevent that.


raysan5:
@JeffM2501 oh! good catch! thanks!

PR #2892, opened 2023-01-26
Add IsReady() methods to ensure assets loaded successfully

RobLoach:
I found I kept on re-checking what data within the structs to check in order to make sure assets were loaded successfully. Is it `.data != NULL`? `.buffer`? Having consistent helper methods will make it easier to check if the assets are ready for use. These match the `IsWindowReady()` pattern...

```
IsAudioStreamReady()
IsShaderReady()
IsImageReady()
IsFontReady()
IsWaveReady()
IsSoundReady()
IsMusicReady()
```

These arn't really needed, but they do clean up the code, and make it easier to read. I was unsure how to verify `IsModelReady()` as on failure, it loads default data. Should we just check if it's the default model?

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -836,6 +836,12 @@ Wave LoadWaveFromMemory(const char *fileType, const unsigned char *fileData, int
     return wave;
 }
 
+// Checks if wave data is ready
+bool IsWaveReady(Wave wave)
+{
+    return wave.data != NULL;
+}
+
 // Load sound from file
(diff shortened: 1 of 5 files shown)


GithubPrankster:
I tend to come into this pattern a lot. I think these functions are useful.

raysan5:
@RobLoach This is a nice addition! Actually, I think we can take advantage of those functions to do a more careful validation of the loaded/not-loaded assets. For example, not only checking `image.data` but also `image.width > 0`, valid `image.format`, etc.

PR #2893, opened 2023-01-27
Fixed GCC Warnings

CraftingInC:
I fixed a lot of the warnings that GCC / MinGWx64 has with raylib. I kept original code where needed so as to work with other compilers by adding an if define.

### DIFF
--- a/src/external/rl_gputex.h
+++ b/src/external/rl_gputex.h
@@ -9,7 +9,7 @@
 *
 *     Note that some file formats (DDS, PVR, KTX) also support uncompressed data storage.
 *     In those cases data is loaded uncompressed and format is returned.
-* 
+*
 *   TODO:
 *     - Implement raylib function: rlGetGlTextureFormats(), required by rl_save_ktx_to_memory()
 *     - Review rl_load_ktx_from_memory() to support KTX v2.2 specs
@@ -771,7 +771,11 @@ void *rl_load_astc_from_memory(const unsigned char *file_data, unsigned int file
 // Module Internal Functions Definition
 //----------------------------------------------------------------------------------
 // Get pixel data size in bytes for certain pixel format
+#ifdef __GNUC__
+static inline int get_pixel_data_size(int width, int height, int format)
+#else
 static int get_pixel_data_size(int width, int height, int format)
+#endif // __GNUC__
 {
     int data_size = 0;       // Size in bytes
     int bpp = 0;            // Bits per pixel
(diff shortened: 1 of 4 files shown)


src/external/rl_gputex.h
```diff
@@ -771,7 +771,11 @@ void *rl_load_astc_from_memory(const unsigned char *file_data, unsigned int file
 // Module Internal Functions Definition
 //----------------------------------------------------------------------------------
 // Get pixel data size in bytes for certain pixel format
+#ifdef __GNUC__
+static inline int get_pixel_data_size(int width, int height, int format)
```
raysan5, line 775: Why this function needs to be `inlined`?

src/external/stb_truetype.h
```diff
@@ -500,7 +500,11 @@ int main(int arg, char **argv)
 #define __STB_INCLUDE_STB_TRUETYPE_H__
 
 #ifdef STBTT_STATIC
+#ifdef __GNUC__
+#define STBTT_DEF static inline
```
raysan5, line 504: Why?

In any case, external libraries shouldn't be modified in raylib but in the original repositories.

raysan5:
@CraftingInC Thanks for the review. I added some comments but I noticed most of the changes are trailing-spaces removal, formatting changes should be moved into a separate PR.

src/external/stb_truetype.h
CraftingInC: GCC / MinGWx64 would give a warning without the inline being there. Here is more info about INLINE on the official GCC website.

https://gcc.gnu.org/onlinedocs/gcc/Inline.html

CraftingInC:
> 

I saw no trailing space issues on my end. So that must be from something else. Odd that you see that, and I did not.

src/external/rl_gputex.h
CraftingInC: Check my below answer. To remove the GCC / MinGWx64 WARNING, I posted a link below.

src/external/stb_truetype.h
raysan5: I'm afraid that link does not answer my question, I know what is an inline function.

raysan5:
@CraftingInC Sorry, I'm not merging this PR, I think it is not needed.

src/external/stb_truetype.h
CraftingInC: In GCC / MinGWx64 you will get these WARNINGS : defined but not used [-Wunused-function]

To remove those warnings without supressing it, you use the keyword INLINE.

PR #2894, opened 2023-01-27
[BUILD] Fix MSVC warnings and bad project settings for 4.5 release

JeffM2501:
This PR fixes up warnings in the main raylib project when used in Visual Studio 2022.

Changes are
* Typical float/int/unsigned casting issues
* Had to use time_t for the time type in the shader hotload example, a long won't work, the platform may be using a 64 bit time structures
* Disabled CRT warnings in some new examples
* Fixed compute shader project (it was not building anything at all)

### DIFF
--- a/examples/audio/audio_raw_stream.c
+++ b/examples/audio/audio_raw_stream.c
@@ -43,7 +43,7 @@ void AudioInputCallback(void *buffer, unsigned int frames)
     float incr = audioFrequency/44100.0f;
     short *d = (short *)buffer;
 
-    for (int i = 0; i < frames; i++)
+    for (unsigned int i = 0; i < frames; i++)
     {
         d[i] = (short)(32000.0f*sinf(2*PI*sineIdx));
         sineIdx += incr;
--- a/examples/core/core_custom_frame_control.c
+++ b/examples/core/core_custom_frame_control.c
@@ -86,8 +86,8 @@ int main(void)
             
             DrawCircle((int)position, GetScreenHeight()/2 - 25, 50, RED);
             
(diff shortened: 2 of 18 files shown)


orcmid:
While these are useful to do, it is also useful to change the VC/C++ warning level to reduce these, something that is easy to do in Visual Studio as well as in direct command-line compiles.  This is also good for overcoming some slack-byte warnings in x64 compiles.

The business about in-lining can also be avoided for VC/C++ by using link-time code generation options.  The optimizer can detect and employ in-lining and unneeded function opportunities, including in static libraries.

raysan5:

@orcmid Thank you very much for the detailed information.

JeffM2501:
> While these are useful to do, it is also useful to change the VC/C++ warning level to reduce these, something that is easy to do in Visual Studio as well as in direct command-line compiles. This is also good for overcoming some slack-byte warnings in x64 compiles.

My concern with just doing it all at the project level is that some people copy the project file and use it as a template, and I would not want to enforce some specific warning level on them. They may want/need to be warned about such issues in there own code.  I'd much rather have the code be correct. There are also many build situations where the project/build system is generated by a third party app (premake, cmake, etc...) and forcing them to know the magic warning level is not great.

My reasoning is that fixing the code is a one place fix for each issue, changing the project settings can spread out.

raysan5:
@planetis-m It fixes warning when the loop control variable is unsigned.

JeffM2501:
> Why unsigned for the for loop variables? Does it fix any specific issue?

because they are being compared to unsigned values, and fixes a warning

planetis-m:
Good to know thanks

PR #2895, opened 2023-01-27
Add IsReady() for Model, Material, Texture, RenderTexture

RobLoach:
Following #2892, this adds the following functions to check if the objects are ready for use...

``` c
IsModelReady()
IsMaterialReady()
IsTextureReady()
IsRenderTextureReady()
```


raysan5:
@RobLoach thanks for the addition! I will review all those functions for a more detailed validation of all the parameters and possibly add output logs in case of errors/warning.

RobLoach:
That's a great idea 👍

PR #2896, opened 2023-01-27
Raylib.h has exceeded 512 functions!

planetis-m:
Without that increase I get:

Number of text lines in buffer: 1589
malloc(): corrupted top size
core dumped


planetis-m:
Its 518 now.

raysan5:
@planetis-m Wow! I never imagined this day would come! Thanks for the review! 😄

Issue #2884, opened 2023-01-23
[build] Building raylib for WASM using Zig


Not-Nik:
Currently the `build.zig` file only provides support for Windows, Linux, macOS and *BSD, but not WASM. Zig does have a `wasm32/64-emscripten` target, but I don't know if we have to link anything else to get it up and running.

Akselmo:
Just wanted to mention that I found this repository today that has WASM support for Zig https://github.com/ryupold/raylib.zig

Maybe @ryupold could help, so we can have things in same place? :)

(I would like to help as well, but I actually do not know anything about Zig, but I want to use Raylib to learn it :) )

Not-Nik:
So, to get a step further than `@panic("Unsupported OS")`, we can add `emscripten` to the `os.tag` switch in `build.zig`, but since Zig doesn't come with a built-in libc for emscripten it fails anytime raylib includes it.

```diff
diff --git a/src/build.zig b/src/build.zig
index e44271e8..8bcce797 100644
--- a/src/build.zig
+++ b/src/build.zig
@@ -3,7 +3,6 @@ const std = @import("std");
 pub fn addRaylib(b: *std.build.Builder, target: std.zig.CrossTarget) *std.build.LibExeObjStep {
     const raylib_flags = &[_][]const u8{
         "-std=gnu99",
-        "-DPLATFORM_DESKTOP",
         "-D_GNU_SOURCE",
         "-DGL_SILENCE_DEPRECATION=199309L",
         "-fno-sanitize=undefined", // https://github.com/raysan5/raylib/issues/1891
@@ -32,6 +31,8 @@ pub fn addRaylib(b: *std.build.Builder, target: std.zig.CrossTarget) *std.build.
             raylib.linkSystemLibrary("gdi32");
             raylib.linkSystemLibrary("opengl32");
             raylib.addIncludePath("external/glfw/deps/mingw");
+
+            raylib.defineCMacro("PLATFORM_DESKTOP", null);
         },
         .linux => {
             raylib.addCSourceFiles(&.{srcdir ++ "/rglfw.c"}, raylib_flags);
@@ -40,6 +41,8 @@ pub fn addRaylib(b: *std.build.Builder, target: std.zig.CrossTarget) *std.build.
             raylib.linkSystemLibrary("dl");
             raylib.linkSystemLibrary("m");
             raylib.linkSystemLibrary("X11");
+
+            raylib.defineCMacro("PLATFORM_DESKTOP", null);
         },
[...]

PR #2901, opened 2023-01-30
Add WASM support for Zig build

Not-Nik:
Addresses #2884

This does create a `libraylib.a` using `zig build -Dtarget=wasm32-emscripten --sysroot /usr/local/opt/emscripten/libexec/` in `src` on macOS 13.2, but I couldn't yet get any programs to work with it.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -67,8 +67,8 @@ ifeq ($(PLATFORM),PLATFORM_RPI)
     endif
 endif
 
-# Determine PLATFORM_OS in case PLATFORM_DESKTOP selected
-ifeq ($(PLATFORM),PLATFORM_DESKTOP)
+# Determine PLATFORM_OS in case PLATFORM_DESKTOP or PLATFORM_WEB selected
+ifeq ($(PLATFORM),$(filter $(PLATFORM),PLATFORM_DESKTOP PLATFORM_WEB))
     # No uname.exe on MinGW!, but OS=Windows_NT on Windows!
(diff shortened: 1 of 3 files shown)


Not-Nik:
All the example build with the `libraylib.a` that Zig generates using the Makefile now (excluding `others/raylib_opengl_interop` which doesn't look like it should work on web). I haven't checked all, but they seem to work. Next step is making the `build.zig` file work for examples

Not-Nik:
Apparently emscripten obfuscates function names and I don't think there is any way we can guess those, ore generate our own js/html files, so doing that will be disabled.

examples/Makefile
```diff
@@ -126,17 +126,7 @@ ifeq ($(PLATFORM),PLATFORM_DRM)
 endif
 
 # Define raylib release directory for compiled library
-RAYLIB_RELEASE_PATH 	?= $(RAYLIB_PATH)/src
-
-ifeq ($(PLATFORM),PLATFORM_WEB)
```
raysan5, line 131: Afaik, those lines are required to build examples with emscripten on Windows

Not-Nik: I think users should handle their PATH themselves, assuming the position of emsdk seems dangerous and bad practice.

raysan5: sorry but I prefer to keep those lines, they are useful for me and probably also to other users

raysan5:
@Not-Nik Thanks for the review! Feel free to send another PR if some further improvement is required!

Issue #2661, opened 2022-08-22
[core] Mouse orientation not working on FPS camera examples in `PLATFORM_RPI` or `PLATFORM_DRM`


Trithilon:
-+Please, before submitting a new issue verify and check:

 - [Y] I tested it on latest raylib version from master branch
 - [Y] I checked there is no similar issue already reported
 - [Y] My code has no errors or misuse of raylib

### Issue description

Any example with mouse look (for instance, core_3d_camera_first_person ) will instantly start spinning on RPI or DRM environments. This does not happen in X11/Desktop mode. I am not sure if its just the example that is broken OR if the way raylib is handling input is not correct.

### Environment

Raspbian Buster Lite running on RPI02W with OR without KMS drivers and DRM.

### Issue Video

https://user-images.githubusercontent.com/7444189/186026675-cabceddb-4a6b-4ac5-88e8-00a5d1146bf5.mov

### Code Example

Just run the example core_3d_camera_first_person in a non X11 environment to reproduce this.

raysan5:
This issue is probably related to the missing `DisableCursor()` implementation for `PLATFORM_RPI`/`PLATFORM_DRM`. That function locks cursor inside the window returning only relative movements.

Issue #2660, opened 2022-08-22
[core] Mouse wheel not working in `PLATFORM_RPI` or `PLATFORM_DRM`


Trithilon:
Please, before submitting a new issue verify and check:

 - [Y] I tested it on latest raylib version from master branch
 - [Y] I checked there is no similar issue already reported
 - [Y] My code has no errors or misuse of raylib

### Issue description

Any example which uses mouse-wheel input doesn't work (for instance - core_input_mouse_wheel). So any example which involves zoom-in using mouse-wheel won't work.

### Environment

Running on RPI02W on Raspbian Lite Buster (32 bit).

### Code Example

Just run make PLATFORM_RPI or DRM against raylib 4.2 src and examples on a Raspbian Buster Lite or Bullseye Lite installation. You will notice that in /core_input_mouse examples, you will be able to move the mouse, detect left, right and middle click - however you won't be able to detect mouse wheel movement in the core_input_mouse_wheel example.

raysan5:
@Trithilon Afaik, native mouse device access does not support mouse wheel events, it requires a windowing system. I could be wrong, please correct me if you know any native headless mouse driver supporting mouse wheel.

clbr:
The native /dev/mouse does support wheels, and for more complex mice, /dev/input/eventX nodes can be used too (no X/windowing system required to use evdev input).

```Mousedev will generate either PS/2, ImPS/2 (Microsoft IntelliMouse) or
ExplorerPS/2 (IntelliMouse Explorer) protocols, depending on what the
program reading the data wishes. ... You'll need ImPS/2 if you want to make use of a wheel on a USB
mouse```

Trithilon:
> The native /dev/mouse does support wheels, and for more complex mice, /dev/input/eventX nodes can be used too (no X/windowing system required to use evdev input).
> 
> ```
> ExplorerPS/2 (IntelliMouse Explorer) protocols, depending on what the
> program reading the data wishes. ... You'll need ImPS/2 if you want to make use of a wheel on a USB
> mouse```
> ```
So something like this?
https://forums.raspberrypi.com/viewtopic.php?t=30762#p267522

clbr:
Yes.

raysan5:
Reviewed this issue but it seems mouse wheel is actually read: https://github.com/raysan5/raylib/blob/master/src/rcore.c#L6469

Please, could someone review this issue?

chrislavoy:
[...]

PR #2905, opened 2023-02-03
[core]  rcamera bugfix - messed up orientation

the-argus:
I believe this addresses #2660 and maybe #2661 though I'm not sure. Maybe I was just misusing the library.
Moving the assignment of the camera's y value to ``playerEyesPosition`` before the projection matrix was applied a.) made more sense to me and b.) fixed a weird issue I was having with v4.2.0 where mouse movement would move the player in a very small range *if* the playerEyesPosition ever changed from what it was initialized at.

I am on linux and have only tested the build on linux. Works well, though, and resolved my issues.


raysan5:
@the-argus Thanks for the review! I'm testing it on Windows...

PR #2907, opened 2023-02-04
Avoid warnings about [lossy] implicit type conversions

cross:
I'm building another project that pulls and compiles raylib.  When I build my project with Xcode on the Mac, I get warnings from raylib compilation.  This avoids those.  Mostly, it extends some smaller data types from int to long.

### DIFF
--- a/src/external/jar_mod.h
+++ b/src/external/jar_mod.h
@@ -217,7 +217,7 @@ typedef struct tracker_state_
     int cur_pattern;
     int cur_pattern_pos;
     int cur_pattern_table_pos;
-    unsigned int buf_index;
+    unsigned long buf_index;
     track_state tracks[32];
 }tracker_state;
 
@@ -1114,7 +1114,7 @@ bool jar_mod_setcfg(jar_mod_context_t * modctx, int samplerate, int bits, int st
 }
 
 // make certain that mod_data stays in memory while playing
-static bool jar_mod_load( jar_mod_context_t * modctx, void * mod_data, int mod_data_size )
+static bool jar_mod_load( jar_mod_context_t * modctx, void * mod_data, mulong mod_data_size )
 {
     muint i, max;
     unsigned short t;
--- a/src/external/jar_xm.h
+++ b/src/external/jar_xm.h
@@ -2196,7 +2196,7 @@ uint64_t jar_xm_get_remaining_samples(jar_xm_context_t* ctx) {
 
 int jar_xm_create_context_from_file(jar_xm_context_t** ctx, uint32_t rate, const char* filename) {
     FILE* xmf;
-    int size;
+    long size;
     int ret;
 
     xmf = fopen(filename, "rb");
--- a/src/external/qoi.h
+++ b/src/external/qoi.h
@@ -286,7 +286,7 @@ is filled with the description from the file header.
 
 The returned pixel data should be free()d after use. */
 
-void *qoi_decode(const void *data, int size, qoi_desc *desc, int channels);
+void *qoi_decode(const void *data, long size, qoi_desc *desc, int channels);
 
 
 #ifdef __cplusplus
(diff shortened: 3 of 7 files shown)


raysan5:
@cross I'm afraid external libraries shouldn't be modified in raylib but in its origin repos. Also seems several warnings are related to `int`-`long` types conversions and afaik, `long` size is platform and compiler dependant and changing current types can cause further issues. Personally I prefer to avoid long data types if possible.

orcmid:
I solved this problem by changing the warning level (when using the Microsoft VC/C++ compiler) to suppress warnings (option "/w") when compiling the *released* raylib components.  For user/under-development code, I use warning level "/W3".  I expect there are equivalents with GCC.

cross:
> @cross I'm afraid external libraries shouldn't be modified in raylib but in its origin repos. Also seems several warnings are related to `int`-`long` types conversions and afaik, `long` size is platform and compiler dependant and changing current types can cause further issues. Personally I prefer to avoid long data types if possible.

I agree that managing data types across architectures is hard.  But, for that reason, I use long more than int.  \*shrug\*.  I'd started switching things to standard types like `size_t`, but that made for a more complex change so stuck with simpler changes.  Let me know if you'd be more amenable to using the POSIX standard types for things, and I can do that too.

And if you can point me to the upstream for the things in `external/` I'm happy to go there.  Sorry I didn't think about that.  

If you want my updates to the three files _not_ in external, let me know.  I should've used `size_t` in `src/rcore.c` as well, but other than that those changes seem clean.

raysan5:
@cross Actually in raylib I tried to avoid the use of `long`, it's hardly used on the source code (I think only used 2-3 very specific times), same for `size_t`. I try to stay to the most common C provided types.

About the external libraries, you can check the original libraries here: https://github.com/raysan5/raylib/wiki/raylib-dependencies.

 > If you want my updates to the three files not in external, let me know. I should've used size_t in src/rcore.c as well, but other than that those changes seem clean.

I think all changes involve replacing `int` by `long` or `size_t`, I prefer to avoid those changes.

orcmid:
I'm inclined to disagree about `size_t`.  It is a well-established (since C99) technique for ensuring a non-negative (i.e., unsigned) integer capable of holding a `sizeof` result.  It tends to be capable of holding a pointer (to a `void` as it were) in most cases.  It is interesting that ISO Standard C (99) has `size_t` defined via any of `<stddef.h>`, `<stdio.h>`, and `<wchar.h>` for convenience.  This `size_t` was introduced in C89.

My inclination is not a strong disagreement, although we are left to deal somehow with functions in libraries (standard and externals) that return or employ parameters of type 'size_t'.

PS: Because it will come up, '<stdatomic.h>` arrived with C11 (where apparently it was an optional facility) and not C99.  That's the case for `<threads.h>` too.

cross:
C99 introduced standard fixed-width integer types.  Both int and long can vary in size.  I was inclined to move things defined int in the code, and receiving a long, because ints can be shorter than long.  A long is always as big as or bigger than an int, as far as I know.
But I can agree to disagree if you'd like to use int most often.  My recommendation is like Dennis that we confirm to the documented argument and return types, so when those are size_t use that.  I'd be happy if all code everywhere used the fixed-width types from C99, but, I've been doing this for too long to do that myself, so.  ;-)

raysan5:
@orcmid @cross I agree that using `size_t` is probably the best approach but at this moment it will require a complete review of the library to adapt to it. I prefer to avoid it.

About the reason of avoiding `size_t`: raylib was originally created for education, specifically to teach programming to art students with no programming background and with a hard time-constrain. They should be able to code a game from scratch in only 40h of class, including the basic programming learning. I tried to limit the library to the most basic data types for them (`int`, `char`, `float`).

orcmid:
I sympathize with the use of raylib (releases) as part of a course on basic programming.  That's different than working from the dev branch and working-on/maintaining the raylib code.  

In my own case, I would certainly make use of `size_t` and `sizeof` early because of how the first shows up in the standard libraries and how `sizeof` is important in writing clean/safe C/C++ languages.  And I will always employ that in my example code and releases.

 I would not want to start with suspect practices concerning passing around structures and especially functions such as `memcpy` and `malloc`.  Likewise with functions that deliver strings/structs/arrays.  

I don't want to encourage running naked through the bits.  I also encourage the use of git (and GitHub) from the beginning.  It is great that the most popular tools all support smooth git integration and, now, many online programming courses emphasize/depend-on that.

PR #2910, opened 2023-02-05
Update zig build to latest master

star-tek-mb:
Version: approximately usable at `zig 0.11.0-dev.1571+c1f71963a`.

Also, adds package manager support (early MVP).

We need to add root build zig file to support Zig Package Manager. Once package manager gets support for overidding build root directory - we can move this logic in `src/build.zig` and remove root `build.zig`.

Also, now every build artifacts must accept target and optimize arguments, so it is also adjusted.

For now, there is not much info about how to use package manager in Zig, so I will leave it here:

Create `build.zig.zon`:
```zig
.{
    .name = "raylib_usage",
    .version = "0.0.0",

    .dependencies = .{
        .raylib = .{
            .url = "https://github.com/raysan5/raylib/archive/543333cae47c397f9bce93fe654aeba7cbe7e33d.tar.gz",
            .hash = "1220a2bf5e5a53060256428b107b7a86077581861e9cb7b42c5083f443575646d59f",
        },
    },
}
```

Hashes and commits may need adjusting (once merged).

Then in `build.zig` of your project:

```zig
...
const raylib_dep = b.dependency("raylib", .{
    .target = target,
    .optimize = optimize,
});
...
exe.linkLibrary(raylib_dep.artifact("raylib"));
...
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,0 +1,18 @@
+const std = @import("std");
+const raylib = @import("src/build.zig");
+
+pub fn build(b: *std.Build) void {
+    // Standard target options allows the person running `zig build` to choose
+    // what target to build for. Here we do not override the defaults, which
+    // means any target is allowed, and the default is native. Other options
+    // for restricting supported target set are available.
+    const target = b.standardTargetOptions(.{});
+    // Standard optimization options allow the person running `zig build` to select
+    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall. Here we do not
+    // set a preferred release mode, allowing the user to decide how to optimize.
+    const optimize = b.standardOptimizeOption(.{});
+
+    const lib = raylib.addRaylib(b, target, optimize);
+    lib.installHeader("src/raylib.h", "raylib.h");
+    lib.install();
+}
--- a/examples/build.zig
+++ b/examples/build.zig
@@ -1,14 +1,10 @@
 const std = @import("std");
 const builtin = @import("builtin");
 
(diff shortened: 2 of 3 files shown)


raysan5:
@star-tek-mb thank you very much for the review! Maybe thos information could be added somewhere in the Wiki for future users?

star-tek-mb:
Yeah, it should be added to Wiki, but after proper tooling for zig package manager (cli). Like adding dependencies (like npm install).

Not-Nik:
I feel like this PR was premature given that Zig 0.11 hasn't been released to this day. The average user (me :)) now can't use raylib's `build.zig` because it uses `std.Build` which isn't available in 0.10.1 (the latest release). In the future I would recommend waiting for official releases before updating `build.zig` given the possibility for sudden unannounced API changes breaking the script.

#2913 related

raysan5:
@Not-Nik Thanks for noticing it, I expected Zig 0.11 to be released by next raylib release but it's not...

star-tek-mb:
> I feel like this PR was premature given that Zig 0.11 hasn't been released to this day. The average user (me :)) now can't use raylib's `build.zig` because it uses `std.Build` which isn't available in 0.10.1 (the latest release). In the future I would recommend waiting for official releases before updating `build.zig` given the possibility for sudden unannounced API changes breaking the script.
> 
> #2913 related

You are right to some extent. But zig 0.11 have many C ABI fixes on windows and linux. (https://github.com/ziglang/zig/pull/14333). So, i recommend to upgrade anyways, but if you want to use raylib for 0.10.1 you can use older raylib version I think.

PR #2913, opened 2023-02-06
zig build: do not use deprecated functions

star-tek-mb:
this is follow up of #2910.

i renamed deprecated functions/structs std.build.Builder and related stuff.


raysan5:
@star-tek-mb thanks for the review!

PR #2914, opened 2023-02-06
Fixed some grammar mistakes and typos.

stickM4N:
As asked, I made fewer commits in this pull request. I will keep reviewing the project... And adding what I found in time. Let me know about your revision.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -474,7 +474,7 @@ void InitAudioDevice(void)
         return;
     }
 
-    // Mixing happens on a seperate thread which means we need to synchronize. I'm using a mutex here to make things simple, but may
+    // Mixing happens on a separate thread which means we need to synchronize. I'm using a mutex here to make things simple, but may
     // want to look at something a bit smarter later on to keep everything real-time, if that's necessary.
     if (ma_mutex_init(&AUDIO.System.lock) != MA_SUCCESS)
     {
@@ -2107,8 +2107,8 @@ void UnloadAudioStream(AudioStream stream)
 }
 
 // Update audio stream buffers with data
-// NOTE 1: Only updates one buffer of the stream source: unqueue -> update -> queue
-// NOTE 2: To unqueue a buffer it needs to be processed: IsAudioStreamProcessed()
+// NOTE 1: Only updates one buffer of the stream source: dequeue -> update -> queue
+// NOTE 2: To dequeue a buffer it needs to be processed: IsAudioStreamProcessed()
 void UpdateAudioStream(AudioStream stream, const void *data, int frameCount)
 {
     if (stream.buffer != NULL)
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -314,7 +314,7 @@ typedef enum {
 #endif
 
 // NOTE: Different logic is used when feeding data to the playback device
-// depending on whether or not data is streamed (Music vs Sound)
+// depending on whether data is streamed (Music vs Sound)
 typedef enum {
     AUDIO_BUFFER_USAGE_STATIC = 0,
     AUDIO_BUFFER_USAGE_STREAM
@@ -1089,7 +1089,7 @@ void PlaySoundMulti(Sound sound)
     unsigned int oldAge = 0;
     int oldIndex = -1;
 
-    // find the first non playing pool entry
+    // find the first non-playing pool entry
     for (int i = 0; i < MAX_AUDIO_BUFFER_POOL_CHANNELS; i++)
     {
         if (AUDIO.MultiChannel.channels[i] > oldAge)
@@ -1105,7 +1105,7 @@ void PlaySoundMulti(Sound sound)
         }
     }
 
-    // If no none playing pool members can be index choose the oldest
+    // If no none playing pool members can be indexed choose the oldest
     if (index == -1)
     {
         TRACELOG(LOG_WARNING, "SOUND: Buffer pool is already full, count: %i", AUDIO.MultiChannel.poolCounter);
@@ -1113,7 +1113,7 @@ void PlaySoundMulti(Sound sound)
         if (oldIndex == -1)
         {
             // Shouldn't be able to get here... but just in case something odd happens!
-            TRACELOG(LOG_WARNING, "SOUND: Buffer pool could not determine oldest buffer not playing sound");
+            TRACELOG(LOG_WARNING, "SOUND: Buffer pool could not determine the oldest buffer not playing sound");
             return;
         }
 
@@ -2305,7 +2305,7 @@ static ma_uint32 ReadAudioBufferFramesInInternalFormat(AudioBuffer *audioBuffer,
 
     if (currentSubBufferIndex > 1) return 0;
 
-    // Another thread can update the processed state of buffers so
+    // Another thread can update the processed state of buffers, so
     // we just take a copy here to try and avoid potential synchronization problems
     bool isSubBufferProcessed[2] = { 0 };
     isSubBufferProcessed[0] = audioBuffer->isSubBufferProcessed[0];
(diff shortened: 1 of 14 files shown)


src/raudio.c
```diff
@@ -2319,7 +2319,7 @@ static ma_uint32 ReadAudioBufferFramesInInternalFormat(AudioBuffer *audioBuffer,
     {
         // We break from this loop differently depending on the buffer's usage
         //  - For static buffers, we simply fill as much data as we can
-        //  - For streaming buffers we only fill the halves of the buffer that are processed
```
orcmid, line 2322: raudio.c line 2322: I am unclear on what change captures the intended sense here.  I think it is "**a half of the buffer that is processed**."  That might mean both haves in sequence when the buffer has been completely drained.

src/raylib.h
```diff
@@ -839,10 +839,10 @@ typedef enum {
 typedef enum {
     CUBEMAP_LAYOUT_AUTO_DETECT = 0,         // Automatically detect layout type
     CUBEMAP_LAYOUT_LINE_VERTICAL,           // Layout is defined by a vertical line with faces
-    CUBEMAP_LAYOUT_LINE_HORIZONTAL,         // Layout is defined by an horizontal line with faces
+    CUBEMAP_LAYOUT_LINE_HORIZONTAL,         // Layout is defined by a horizontal line with faces
     CUBEMAP_LAYOUT_CROSS_THREE_BY_FOUR,     // Layout is defined by a 3x4 cross with cubemap faces
     CUBEMAP_LAYOUT_CROSS_FOUR_BY_THREE,     // Layout is defined by a 4x3 cross with cubemap faces
-    CUBEMAP_LAYOUT_PANORAMA                 // Layout is defined by a panorama image (equirectangular map)
+    CUBEMAP_LAYOUT_PANORAMA                 // Layout is defined by a panorama image (equirrectangular map)
```
orcmid, line 845: raylib.h line 845: the original spelling is correct.

src/rcore.c
```diff
@@ -5456,7 +5456,7 @@ static void MouseButtonCallback(GLFWwindow *window, int button, int action, int
     gestureEvent.position[0].x /= (float)GetScreenWidth();
     gestureEvent.position[0].y /= (float)GetScreenHeight();
 
-    // Gesture data is sent to gestures system for processing
+    // Gesture data is sent to gestures-system for processing
```
orcmid, line 5459: rcore.c line 5459: The original seems correct.  It is not used as an adjective as gestures-system would be.

src/rmodels.c
```diff
@@ -1383,7 +1383,7 @@ void DrawMesh(Mesh mesh, Material material, Matrix transform)
     }
 
     // Get a copy of current matrices to work with,
-    // just in case stereo render is required and we need to modify them
+    // just in case stereo render is required, and we need to modify them
```
orcmid, line 1386: The "and" here may be stronger than the conjunction it is with the comma.  I will not die on this though.

src/rmodels.c
```diff
@@ -4671,7 +4671,7 @@ static Image LoadImageFromCgltfImage(cgltf_image *cgltfImage, const char *texPat
 {
     Image image = { 0 };
 
-    if (cgltfImage->uri != NULL)     // Check if image data is provided as a uri (base64 or path)
+    if (cgltfImage->uri != NULL)     // Check if image data is provided as an uri (base64 or path)
```
orcmid, line 4674: I guess the problem here is how we thing "uri" is pronounced.   The "u" is not always a vowel.  I am also not dying here.  However the W3C specifications for URIs (including URLs) use "a url" and "a uri."  I presume because "a universal resource locator" is correct. (I once worked on a Univac computer. :).

src/rmodels.c
```diff
@@ -5110,7 +5110,7 @@ static Model LoadGLTF(const char *fileName)
                 {
                     // The primitive actually keeps the pointer to the corresponding material,
                     // raylib instead assigns to the mesh the by its index, as loaded in model.materials array
-                    // To get the index, we check if material pointers match and we assign the corresponding index,
+                    // To get the index, we check if material pointers match, and we assign the corresponding index,
```
orcmid, line 5113: models.c line 5113, again, it seems better without the ",".  Again, it is about a combined act that has the result.  I don't claim authority though :).

src/rshapes.c
```diff
@@ -978,7 +978,7 @@ void DrawRectangleRounded(Rectangle rec, float roundness, int segments, Color co
     rlSetTexture(texShapes.id);
 
     rlBegin(RL_QUADS);
-        // Draw all of the 4 corners: [1] Upper Left Corner, [3] Upper Right Corner, [5] Lower Right Corner, [7] Lower Left Corner
+        // Draw all the 4 corners: [1] Upper Left Corner, [3] Upper Right Corner, [5] Lower Right Corner, [7] Lower Left Corner
```
orcmid, line 981: shapes.c line 981 or just day "Draw the 4 corners: ..."
Also lines 1210, 1345

src/rtext.c
```diff
@@ -647,7 +647,7 @@ GlyphInfo *LoadFontData(const unsigned char *fileData, int dataSize, int fontSiz
                     }
                 }
 
-                // Get bounding box for character (may be offset to account for chars that dip above or below the line)
+                // Get bounding box for character (maybe offset to account for chars that dip above or below the line)
```
orcmid, line 650: rtext.c line 650:  Is it safe to say "(will be offset ..." ?

src/rtext.c
```diff
@@ -1040,7 +1040,7 @@ void DrawTextEx(Font font, const char *text, Vector2 position, float fontSize, f
 
     int size = TextLength(text);    // Total size in bytes of the text, scanned by codepoints in loop
 
-    int textOffsetY = 0;            // Offset between lines (on line break '\n')
+    int textOffsetY = 0;            // Offset between lines (on linebreak '\n')
```
orcmid, line 1043: rtext.c line 1043: The term in the C Language specifications is "newline" for the '\n' escape.  "linebreak" refers to any of the ways that one line is separated from the next and newline might be transformed into/from another in the stdio.h operations, depending on the computer platform.  But in the C program, it is `\n` (for code 0x0A always) with the Standard C library.  I know, fussy.

orcmid:
Well-done.  This is more manageable.  I have fewer comments and some may be coin-flips.  Ramon's review will determine what's important enough to him.

src/raylib.h
stickM4N: You are right. My mistake

src/rcore.c
stickM4N: In that case it should not have s in gestures...

src/rmodels.c
stickM4N: Ok, let's use a then.

src/rtext.c
stickM4N: Better to use newline then

src/rtext.c
stickM4N: On seconds thoughts it should be used might... Wouldn't it fit better?

stickM4N:
Yes, hope this helps. I still hava a pull request regarding a buffer too small for that data stored. I'll check it soon and some other things I noticed so you can review it.

src/rcore.c
orcmid: That does sound better.  Is it accurate?

src/rtext.c
raysan5: both sound good to me

src/rshapes.c
raysan5: agree

src/rcore.c
raysan5: gesture-system is ok

src/raudio.c
raysan5: good

PR #2915, opened 2023-02-06
Enable GetWindowHandle() on macOS

Not-Nik:
This isn't the nicest fix, but it works, because `id` is a typedef for `struct objc_object *`.


raysan5:
@Not-Nik Thanks for the review... not much elegant but it's ok for now

PR #2919, opened 2023-02-10
[example] Hybrid Rendering

BugraAlptekinSari:
[You can see a demonstration in this video](https://www.youtube.com/watch?v=QoQEYKSWQ1A)

This example shows how you can render a single frame using both rasterization and ray marching.
It does this by manually calculating the depth of each pixel and writing it into the depth buffer.

I tested this on PLATFORM_WEB and PLATFORM_DESKTOP on a Linux machine like my previous example.

Additionally I fixed indentation on the makefile. `ifeq` isn't supposed to be indented with tabs.


raysan5:
@BugraAlptekinSari This is a great example! Thank you very much for the addition!

PR #2920, opened 2023-02-10
Fix ScissorMode for RenderTextures on macOS

Not-Nik:
ScissorMode would previously use the window height when repositioning the scissor origin to the top left.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2323,7 +2323,7 @@ void BeginScissorMode(int x, int y, int width, int height)
 
 #if defined(__APPLE__)
     Vector2 scale = GetWindowScaleDPI();
-    rlScissor((int)(x*scale.x), (int)(GetScreenHeight()*scale.y - (((y + height)*scale.y))), (int)(width*scale.x), (int)(height*scale.y));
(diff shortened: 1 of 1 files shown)


Not-Nik:
Actually, now it seems something is off with HighDPI scaling

Not-Nik:
I previously tested ignoring DPI settings for render textures completely, but that might not be what you want @raysan5 

```c
Vector2 scale;
if (CORE.Window.renderTexture) scale = (Vector2) {1, 1};
else scale = GetWindowScaleDPI();
rlScissor((int)(x*scale.x), (int)(CORE.Window.currentFbo.height*scale.y - (((y + height)*scale.y))), (int)(width*scale.x), (int)(height*scale.y));
```

raysan5:
@Not-Nik I'm afraid this issue has been reviewed many times and there is always some side case (macOS version, CPU platform, display type) that fails and the issue comes again. I don't know what is the best solution for it, I considered leaving all HighDPI management to user but some raylib internal parts are not directly accessible for that...

Not-Nik:
I feel like render textures shouldn't be affected by high DPI, because they are rarely ever used for 2d interfaces. Also creation is independent of the current DPI value, so having scissor mode be affected by it seems inconsistent

raysan5:
I'm closing this draft. HighDPI and macOS implies many considerations.

PR #2923, opened 2023-02-14
Move camera after rotation

veins1:
Moving camera after rotation is set, to avoid 1 frame delay in movement direction


raysan5:

PR #2925, opened 2023-02-15
[CAMERA] Add Input Key mapping function for camera controls

JeffM2501:
This PR adds a new function to the new 4.5-dev camera API
`SetCameraInputKey`
It allows the user to change the keys used for the input used by `UpdateCamera`, or disable the features completely.
This function works similar to `SetExitKey`
The enum `CameraInputKeys `contains all the available controls.
The function is available in standalone mode as well.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1173,6 +1173,7 @@ RLAPI float GetGesturePinchAngle(void);                 // Get gesture pinch ang
 // Camera System Functions (Module: rcamera)
 //------------------------------------------------------------------------------------
 RLAPI void UpdateCamera(Camera *camera, int mode);      // Update camera position for selected mode
+RLAPI void SetCameraInputKey(int input, int key);       // Set the camera input keys using the enum CameraInputKeys
 
 //------------------------------------------------------------------------------------
 // Basic Shapes Drawing Functions (Module: shapes)
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -895,6 +895,22 @@ typedef enum {
     CAMERA_ORTHOGRAPHIC             // Orthographic projection
 } CameraProjection;
 
+typedef enum {
+    CAMERA_INPUT_FORWARD = 0,
+    CAMERA_INPUT_BACK,
+    CAMERA_INPUT_LEFT,
+    CAMERA_INPUT_RIGHT,
+    CAMERA_INPUT_PITCH_UP,
+    CAMERA_INPUT_PITCH_DOWN,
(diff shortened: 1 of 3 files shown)


raysan5:
@JeffM2501 Thinking about the best way to implement an input system... I see this approach is similar to previous implementation but it still falls short in many situations:

 - No mouse input setup
 - No multiple keys support (i.e. WASD + CursorKeys)
 - No other input methods setup (i.e Gamepad, Touch)

Despite it can be useful for some users, probably many users sooner or later will concern about mentioned features... 

So, I decided to just keep the hardcoded implementation for now, as a simple-internal-use camera implementation sample, and force the user to move to an external camera when more advance inputs are required.

Definitely, camera module shouldn't belong to raylib...

JeffM2501:
Makes sense. I can see how it's a tough thing.

raysan5:
@JeffM2501 One option I consider maybe for a future is just moving the input states out of `rcamera.h` and let the users send them every frame (or when required). I've seen that approach in other engine-independant camera modules. For example:

```c
void UpdateCameraState(float moveForward, float moveRight, float moveUp, float moveDown, float moveToTarget, float yaw, float pitch, float roll);

// Usage example (not tested, just to illustrate the idea)
UpdateCameraState(
    (IsKeyDown(KEY_W) || IsKeyDown(KEY_UP))*CAMERA_MOVE_SPEED - 
    (IsKeyDown(KEY_A) || IsKeyDown(KEY_DOWN))*CAMERA_MOVE_SPEED,
    (IsKeyDown(KEY_S) || IsKeyDown(KEY_RIGHT))*CAMERA_MOVE_SPEED -
    (IsKeyDown(KEY_D) || IsKeyDown(KEY_LEFT))*CAMERA_MOVE_SPEED,
    GetMouseWheelMove()*2.0f,
    GetMouseDelta().x*CAMERA_ROTATION_SPEED,
    GetMouseDelta().y*CAMERA_ROTATION_SPEED,
    0.0f);
```

This way, the `CameraUpdate` control will be fully in hands of the user... but maybe arrived this point, just exposing the new `CameraMove*()` functions would be enough.

JeffM2501:
Yes that could be cleaner and allow for simpler integration of other input systems like the controller

raysan5:
> Yes that could be cleaner and allow for simpler integration of other input systems like the controller

Yeah, that would be one more step towards a completely raylib-decoupled camera and input-customizable.

Mmmh... maybe that function could be provided as `UpdateCameraEx()` or `UpdateCameraPro()` version for advance users requiring that extra level of control...

PR #2926, opened 2023-02-15
[CAMEARA] Fix Orbital Camera Mode

JeffM2501:
This PR makes the orbital camera mode work like it used to before the change to the new camera system.

I think there was some confusion that the author of the camera system had about what the orbital camera did and they just made it a copy of the third person camera. It is not that. It is a always turning camera that has no other input. You move the camera by moving the camera manually.

This PR restores that functionality using the new camera logic.

Also during the camera update the cursor was disabled in all examples that used an orbit camera, possibly because they thought it was a third person camera. This PR restores the cursor because the mouse is not used with orbital camera.

With this PR all the demos should be back to working like they used to, with the new camera logic.

### DIFF
--- a/examples/models/models_billboard.c
+++ b/examples/models/models_billboard.c
@@ -55,8 +55,6 @@ int main(void)
     float distanceRotating;
     float rotation = 0.0f;
 
-    DisableCursor();                    // Limit cursor to relative movement inside the window
-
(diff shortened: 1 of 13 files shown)


raysan5:
@JeffM2501 Thank you very much for the review!

Issue #2928, opened 2023-02-18
[raudio/feature request] Audio stream processor for mixed output


hatkidchan:
We already have 3 functions related to processing/generating sound programmatically:
```c
void SetAudioStreamCallback(AudioStream stream, AudioCallback callback);
void AttachAudioStreamProcessor(AudioStream stream, AudioCallback processor);
void DetachAudioStreamProcessor(AudioStream stream, AudioCallback processor);
```
but they work on separate audio streams, not the mixed audio that is being sent to underlying miniaudio library. It would be nice to have same functionality available for all sounds as a last step before sending samples to the sound hardware.

I see it as 2 functions that accept another function as a callback just before sending samples:
```c
void AttachAudioMixerProcessor(AudioCallback processor);
void DetachAudioMixerProcessor(AudioCallback processor);
```
that should work the same way as already existing functions, but instead of taking separate stream as their argument, they read/modify mixed output.

Use cases:
 * Global sound effects (such as lowering the volume when window is out of focus, low-pass filters, etc);
 * Visualization of sound being played (can be useful when creating applications/games that are heavily sound-bound)
 * Beat detection
 * Volume normalization
 * [ list goes on ]

---

Looking at the source of `OnSendAudioDataToDevice` it should be pretty trivial to implement, maybe right after `MixAudioFrames` call we could do the same as with each buffer?

I could try doing that, but I'm not sure if that addition would be welcome in the library

hatkidchan:
Ooookay, deeply nested loop/conditional mess confused me and it appears that we call `MixAudioFrames` after processing each buffer, so calling it after that wouldn't really help.

Though, we could call it after all of those loops maybe...

hatkidchan:
This seems to work, but far from ideal:

https://github.com/hatkidchan/raylib-patches/commit/085b59f8784b4a2ba8dc4f867081d4372b1130f3

Right now it appears to skip some frames, which causes sound to be choppy on writes.

But it seems work good enough for visualizations:
<details>
<summary>raudiotest.c</summary>

```c
// x-run: ~/scripts/runc.sh % libraylib.a -lm
#include "raylib.h"
#include <stddef.h>
#include <stdint.h>
#include <stdio.h>
#include <string.h>
#include <sys/types.h>

void audio_callback(float *_buf, unsigned int frames);

#define BUFFER_SIZE 2048

float audio_buffer[BUFFER_SIZE];
size_t buf_offset = 0;

[...]

PR #2929, opened 2023-02-19
Added mixed audio processor

hatkidchan:
Addresses #2928 and implements that, tested on Linux but also should work on Windows and other platforms, since it's not a significant change.

Adds two new functions:
```c
RLAPI void AttachAudioMixedProcessor(AudioCallback processor); // Attach audio stream processor to the entire audio pipeline
RLAPI void DetachAudioMixedProcessor(AudioCallback processor); // Detach audio stream processor from the entire audio pipeline
```

They should be used the same way as `AttachAudioStreamProcessor`/`DetachAudioStreamProcessor`, with exception of not requiring a `stream` parameter. 

Reasoning behind adding those functions can be tracked in the related issue.

PR *should* contain 2 commits, one of which just renames structure member to follow https://github.com/raysan5/raylib/wiki/raylib-coding-conventions, but Git being funky and for some reason it shows commits with my previous changes as well. Hope it won't cause any problems!

I may also add example for those functions into `examples/audio` folder, if it's okay


raysan5:
@hatkidchan Thank you very much for the addition, the detailed description and clean implementation.

PR #2944, opened 2023-02-24
Add raylib_misc to the wrapper list

RealDoigt:
This PR adds raylib_misc, an object oriented wrapper (with some extra utilities and extensions) of raylib for D, to the wrapper list in BINDINGS.md


raysan5:
@RealDoigt Please, could you rename the repo to reflect is D language realted, `raylib_misc` seems confusing.

RealDoigt:
@raysan5 it would break entire my package system. I'd need to change many repos and I could potentially break other people's code as well. I wanted to change the name for a while because it's a really bad name, but I think I'm stuck with it. If you don't want to merge for this reason then it's ok, just close the pull request.

raysan5:
@RealDoigt I'm afraid I'm closing this PR, I think it can be confusing for users checking the BINDINGS document.

RealDoigt:
@raysan5 The bindings document has the language right next to the binding name, you can't miss it unless you're severely blind. In a table with only two other entries, it's even more of a moot point. But it's ok, your repo, your rules. I was going to drop raylib as a backend in the far future anyway.

PR #2951, opened 2023-03-06
fix(build): Fix DLL build on Windows

Skaytacium:
Changes the DLL export condition to apply to platform WIN32 instead of compiler MSVC.

Tested with:
Ninja with Clang 15.0.7
Ninja with MSVC 19.35
Visual Studio solution with MSVC 19.35


raysan5:
@Skaytacium thanks for the review!

PR #2961, opened 2023-03-16
Fix: QuaternionFromEuler() and QuaternionToEuler()

krysperz2:
Fixes wrong conversion from quaternions to euler angles and vice versa.
It was a convention issue. Raylib uses xyzw convention for quaternions but the conversions were for wxyz convention

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -2070,10 +2070,10 @@ RMAPI Quaternion QuaternionFromEuler(float pitch, float yaw, float roll)
     float z0 = cosf(roll*0.5f);
     float z1 = sinf(roll*0.5f);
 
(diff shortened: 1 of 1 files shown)


raysan5:
@krysperz2 Thanks for the review! Could you provide some visual example that illustrates this functionality?

krysperz2:
@raysan5 Well if I am being honest I just copied the code from wikipedia. And I have never used these functions myself. I just saw it on the wish list and thought it was easy enough to fix so I did. But I will try to get an example working as soon as possible

raysan5:
@krysperz2 Sorry, it's not possible to change the convention for raylib. That's the reason I asked for a working example. I'm not merging this PR.

PR #2962, opened 2023-03-16
Added my BQN Library to the list

Brian-ED:
I'm the only maintainer and have been working hard recently to get my library to be as nice to use as possible. I found out about this language list, and getting a new language on it didn't seem so bad. A bonus for my efforts.


raysan5:
@Brian-ED thanks for the addition!

PR #2963, opened 2023-03-17
Calculate exact image size in GenImageFontAtlas

hanaxar:
Calculate exact image size with a method based on total glyph width and glyph row count.
Current method seemed a little bit overkill with square root, log and power functions and only approximates image size which can be wonky with some weird fonts like cursive fonts.
Proposed method calculates image size directly with a simpler (and probably faster) method and results exact image size needed.


raysan5:
@hanaxar Thanks for the review. Would it be possible to use a simple `while()` loop? I try to avoid the `do-while` approach in raylib.

hanaxar:
Converted ```do-while``` into ```while()``` loop.
And also added an extra step to calculate maximum glyph width and excluding it from the image width for extra safety.

raysan5:
@hanaxar Thank you very much for the improvement!

PR #2968, opened 2023-03-19
Fix missing include in rglfw.c on BSD platforms

Koromix:
Building raylib 4.5 (with rglfw.c) on BSD platforms fails with missing symbol `_glfwPollPOSIX`.

The root cause seems to be that rglfw.c does not include `src/posix_poll.c` on BSD systems.

This commit adds the missing include and allows Raylib to successfully build on OpenBSD and FreeBSD atr least (I haven't tested on NetBSD).


raysan5:
@Koromix thanks for the fix! I couldn't test those pkatforms...

PR #2970, opened 2023-03-19
Update h-raylib bindings version

Anut-py:


raysan5:
@Anut-py thanks for the update!

PR #2971, opened 2023-03-19
Update raylib.zig version to 4.5

ryupold:


raysan5:
@ryupold thanks for the update!

PR #2972, opened 2023-03-19
Add const qualifier to char * path argument in qoaplay_open()

Webfra:
Function `qoaplay_open(char * path)` doesn't change the path it is given.
Therefore, it should be safe to add the const qualifier.
Avoids a compiler warning in rauido.c.

Tested on Win10+mingw-gcc 12.2 and Ubuntu-20.04 (on WSL2).


raysan5:
@Webfra Thanks for the improvement!

Issue #2948, opened 2023-03-03
[core] Feature Request - Event Recording Enhancements


MatthewOwens:
## Feature Request Summary
I'm currently developing a game wherein Raylib's event recording/playing features would be very helpful for creating an arcade-style attract mode and saving replays of successful runs.

### Start/Stop Recording via function call
Currently it is only possible to it `ExportAutomationEvents`/`PlayAutomationEvents` with F11/F9, adding something like `BeginEventRecording(char *fileName)`, `StopEventRecording` and `PlayEventRecording(char *fileName)`. 

### Specifying I/O File
Passing a `fileName` parameter to `BeginEventRecording` and `PlayEventRecording` could be used as an alternative to the hard-coded eventsrec.rep

### Toggleable DrawText calls while recording/playing events
Currently when recording or playing an event file, `DrawText` is called. I think that if this exposed this should be toggleable with either a boolean passed to `BeginEventRecording`/`PlayEventRecording` or via a compile-time define.

### Concerns
I understand that this is currently an experimental feature in Raylib, and that there is a TODO to determine how `PlayAutomationEvents` should interact with `PollInputEvents`. If this feature is rejected on these grounds I understand and will maintain my own branch. Otherwise I'd be happy to implement the feature described above.

raysan5:
@MatthewOwens Thanks for your feedback! I had almost forgoten about the `Event Recording System`, it's still in experimental state and it has not been reviewed in several years. I will give it a review after raylib 4.5 release!

varshneydevansh:
I am looking to contribute to this project, is there any good place where I can start from? =)

raysan5:
@varshneydevansh For the event reocrding system there is already a PR proposal and I'm reviewing it myself.

raysan5:
Implemented in commit https://github.com/raysan5/raylib/commit/99dac5451cad1c45c0a6b1da5c6175a7575f3403

PR #2975, opened 2023-03-20
Event recording enhancements

MatthewOwens:
Implements the feature request described in #2948. Added the following functions:
```
RLAPI void BeginEventRecording(bool displayText);
RLAPI void StopEventRecording(const char *fileName);
RLAPI void PlayEventRecording(const char *fileName, bool displayText);
```
This change also modifies the behavior of when `PlayAutomationEvents` is called. It will now play out a recorded event file from when `PlayAutomationEvents` is called, rather than only playing the file if  `CORE.Time.frameCounter < eventCount`.


src/raylib.h
```diff
@@ -1581,6 +1581,15 @@ RLAPI void DetachAudioStreamProcessor(AudioStream stream, AudioCallback processo
 RLAPI void AttachAudioMixedProcessor(AudioCallback processor); // Attach audio stream processor to the entire audio pipeline
 RLAPI void DetachAudioMixedProcessor(AudioCallback processor); // Detach audio stream processor from the entire audio pipeline
 
+#if defined(SUPPORT_EVENTS_AUTOMATION)
+//------------------------------------------------------------------------------------
+// Event Recording & Replaying
+//------------------------------------------------------------------------------------
+RLAPI void BeginEventRecording(bool displayText);
```
raysan5, line 1588: I prefer to avoid this kind of parameters, text display can be configured with a compilation flag. Due to the advance nature of the API there shouldn't be a problem for adavance users.

src/raylib.h
```diff
@@ -1581,6 +1581,15 @@ RLAPI void DetachAudioStreamProcessor(AudioStream stream, AudioCallback processo
...
+RLAPI void StopEventRecording(const char *fileName);
```
raysan5, line 1589: Following raylib API conventions, `BeginEventRecording()` should align with `EndEventRecording()`. But not sure if that's the best name for those functions...
Also, file IO is usually used on `Load*()` and `Export*()` functions. I always try to separate IO file loading from data processing logic. Actually, the API already has:
```c
#if defined(SUPPORT_EVENTS_AUTOMATION)
static void LoadAutomationEvents(const char *fileName);     // Load automation events from file
static void ExportAutomationEvents(const char *fileName);   // Export recorded automation events into a file
static void RecordAutomationEvent(unsigned int frame);      // Record frame events (to internal events array)
static void PlayAutomationEvent(unsigned int frame);        // Play frame events (from internal events array)
#endif
```
I need to think better about this API...

MatthewOwens:
Cheers for the review, I agree that a compilation flag approach would probably be better. I wasn't sure if you'd be open to an additional flag, but I'll switch to that approach and update the function names to better align with raylib's style.

Any concerns around the functionality changes regarding playing events?

raysan5:
@MatthewOwens I know it's been a while since this PR but API has been finally redesigned on commit: https://github.com/raysan5/raylib/commit/99dac5451cad1c45c0a6b1da5c6175a7575f3403

PR #2979, opened 2023-03-21
Fix packing logic error in ```GenImageFontAtlas```

hanaxar:
Basic packing algorithm currently follows this order: 
Copy pixel data -> Move offsetX for current glyph -> Check remaining space for current glyph... 
Since X offset already moved according current glyph, remaining space should be checked for the next glyph. Because of this, current logic causes glyphs wrapping around texture. 
Proposed change fixes that by moving offsetX check to the beginning of the loop.


hanaxar:
An example for the font OpenSans-Regular.ttf with current logic:
(The letter m in the texture is wrapped around)

basic_packing

And after proposed fix:

basic_packing_fixed

raysan5:
@hanaxar Good catch for the issue! And thanks again for the improvement!

PR #2982, opened 2023-03-22
slightly optimize Vector3Normalize

RicoP:
I changed the code of Vector3Normalize to be more similar to Vector2Normalize.

This slight optimization generates slightly better code under clang and msvc: https://godbolt.org/z/7Gzd34YEh


raysan5:
@RicoP Thanks for the improvement! :)

PR #2992, opened 2023-03-29
Update BINDINGS.md (raylib-py -> 4.5)

overdev:
Includes also updates to raylibctbg, from which raylib-py release was generated.


raysan5:
@overdev Thanks for the update!

PR #3000, opened 2023-04-04
Fix GLYPH_NOTFOUND_CHAR_FALLBACK index value

hanaxar:
Currently if codepoint is not found; '_' character is returned. (Supposed to be '?')
It's fixed by using index value of fallback character instead of using the character value.


raysan5:
@hanaxar Actually, checking the code, this fix could neither work as expected. If the charset is unordered, the `?` codepoint must be fetched, it could not be in position - 32, because it's unordered.

So, the function is wrong and it needs a redesign, just pushing a review myself

PR #3003, opened 2023-04-08
CORE : Disable warning logs when loading shaders 

JeffM2501:
This PR sets the log level to only log errors when loading the default shader locations. This is to prevent log spam for these uniforms and attributes that most likely do not exist in custom shaders. This is to prevent warning fatigue to users.


raysan5:
@JeffM2501 I'm afraid I'm not merging this PR, it implies adding a new function just to avoid a tracelog. I just commented the tracelog in `rlGetLocationUniform()`. Users in need of the result could just check for `location != -1`.

PR #3004, opened 2023-04-08
Build : Fix warnings in raylib for MSVC

JeffM2501:
This is another warning fixup for the raylib library.

one size_t casting issue, and disable some warnings on the qoi headers


raysan5:

PR #3006, opened 2023-04-09
Minor fix in DrawLineBezier*

eternalStudent:
When `i` starts with `0`, `t` is also `0`, which results in `previous == startPos == current`, this segment is not only redundant, but it also causes division-by-zero since `sqrtf(dx*dx + dy*dy)` is zero.


raysan5:
@eternalStudent thanks for the review! :)

PR #3007, opened 2023-04-10
Add and optimize text rendering functions

Andersama:
New functions operate on two char*'s in the event strings are not null terminated
* MeasureTextViewEx
* MeasureTextView
* DrawTextViewEx
* DrawTextViewPro

Rewrote some conditional operations as math in DrawTexturePro to try to get compiler to generate branchless code even in debug mode.

Some style changes were made by auto-formatting, feel free to remove those diffs..

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1362,16 +1362,22 @@ RLAPI void UnloadFont(Font font);                                               
 RLAPI bool ExportFontAsCode(Font font, const char *fileName);                               // Export font as code file, returns true on success
 
 // Text drawing functions
+RLAPI void DrawTextViewEx(Font font, const char* text, const char* textEnd, Vector2 position, float fontSize, float spacing, Color tint);
 RLAPI void DrawFPS(int posX, int posY);                                                     // Draw current FPS
 RLAPI void DrawText(const char *text, int posX, int posY, int fontSize, Color color);       // Draw text (using default font)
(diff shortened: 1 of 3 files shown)


RobLoach:
Perhaps rather than having `text` and `textEnd`, we could consider combining them into a `TextView` struct, which would have the `const char*` and a length. But that may go beyond raylib's simple API design choices.

raysan5:
@Andersama Sorry, those are breaking changes that most users won't benefit of. I'm afraid I'm not merging them.

Andersama:
No worries, although looking at it a second time I'm not sure what I broke.

orcmid:
@Andersama The changes of `DrawTextEx` and `MeasureTextEx` to` DrawTextViewEx` and `MeasureTextViewEx` will break code that already depends on the current names.

Some of the improvements in your pull request might best be handled in smaller chunks that deal with individual cases.  That facilitates review and any adjustments that come out of review.

PS: I suspect using conditional expressions ( `b ? m : n` forms) might not fit Ramon's coding-style preferences.  (I do favor them myself.)

Andersama:
I didn't remove functions though, I don't think, unless I messed up my commits, I only added new function signatures.

orcmid:
@Andersama  `rtext.c` has an `#include "raylib.h"` so there should be no issue concerning exported functions.  All the signatures will have already been seen.  

I'm uncertain whether the order of definition matters.  I will look into that.  It will be useful to know.  I have never been tripped up by it though.

Andersama:
It shouldn't matter given the header, but I tend to try to write functions before they get used.

PR #3013, opened 2023-04-14
[raudio] Rewritten `ExportWaveAsCode()` file saving to be more like rtextures `ExportImageAsCode()`

RadsammyT:
The current `ExportWaveAsCode()` has some various bugs/inconsistencies:

The data array name will list the **_entire_** path to the file. I don't know if ':', '/', and '\\' are valid symbols in C var names but I made it so only the file name without its path and extension are listed. Also, note the missing file names before the wave data information preprocessor defines, and the unusual data array var name casing.
Before:
vwM5Kir2Xe

After: 
Xpqc0rfUQ3


raysan5:
@RadsammyT Thanks for the review! :)

PR #3019, opened 2023-04-21
Fix unbounded writing of sscanf

szsam:
'%s' may lead to buffer overflow if the width is not provided.


raysan5:
@szsam this issue is related to an external library, it should be done in original repo, not in raylib repo.

PR #3020, opened 2023-04-21
Fix null pointer dereference

szsam:
Add null check for return values of _glfwPlatformGetTls().


raysan5:
@szsam this issue is related to an external library, it should be done in original GLFW repo, not in raylib repo. or it would be reverted when updating to latest GLFW.

PR #3021, opened 2023-04-21
Fix offset used before range check

szsam:
This use of offset 'i' should follow the range check.


raysan5:
Thanks for the review, I already implemented it to original raygui repo.

akulpillai:
Hi! I’m a student researcher working with @szsam. We have identified this patch as one for a security vulnerability. 

Will a CVE be issued for this?

PR #3022, opened 2023-04-21
Fix off-by-one error in CheckCollisionPointRec

dbechrd:
Checking `<= x + w` causes off-by-one error where `CheckCollisionPointRec` will return true at the same time for two rectangles rendered right next to each, but which don't overlap (e.g. when making a 2D tile editor). This is clearly not what was intended.

The convention for collision checks of this nature should always be lower bound inclusive and upper bound exclusive.


raysan5:
@dbechrd Hey! Good catch! Thanks for the fix!

PR #3029, opened 2023-04-25
Update screen scale and mouse scale when window size is changed

DavidLyhedDanielsson:
I am unsure how the scaling works exactly, but the mouse scale and screen scale is not updated when the window is resized, leading to (at least) incorrect mouse positions. I just copied this code from the `InitGraphicsDevice` function.

The following videos show me starting a program and resizing the window. The mouse position is reported in the main loop with `std::cout << GetMousePosition().x << ", " << GetMousePosition().y << std::endl;` In withoutfix.mp4, the window size has to be very oblong before the reported mouse position is close to a square, in withfix.mp4 this is not the case.

https://user-images.githubusercontent.com/30801498/234382097-1c050f39-3265-4474-90ff-ac96d4ebff96.mp4

https://user-images.githubusercontent.com/30801498/234382103-ecab12e6-acde-4053-abeb-047dfdee77d2.mp4

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5376,6 +5376,13 @@ static void WindowSizeCallback(GLFWwindow *window, int width, int height)
 
         CORE.Window.screen.width = (unsigned int)(width/windowScaleDPI.x);
         CORE.Window.screen.height = (unsigned int)(height/windowScaleDPI.y);
+
+        int fbWidth = 0;
+        int fbHeight = 0;
+        glfwGetFramebufferSize(CORE.Window.handle, &fbWidth, &fbHeight);
+
+        CORE.Window.screenScale = MatrixScale((float)fbWidth/CORE.Window.screen.width, (float)fbHeight/CORE.Window.screen.height, 1.0f);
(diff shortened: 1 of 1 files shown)


raysan5:
@DavidLyhedDanielsson Screen and mouse shouldn't be scaled when window is  resized. It's up to user to manage a virtual screen if that's the objective: https://github.com/raysan5/raylib/blob/master/examples/core/core_window_letterbox.c

raysan5:
It seems this issue could be related to HighDPI screen only... is that the case?

DavidLyhedDanielsson:
> It seems this issue could be related to HighDPI screen only... is that the case?

I think so, yes. I feel like someone would have noticed something otherwise.

I just saw that the mouse position was off and that it was multiplied with the scaling factor. It seemed that the scaling factor was based on the framebuffer size. In that case, do you have any idea what the issue could be?

raysan5:
@DavidLyhedDanielsson I'm afraid HighDPI is still an issue with raylib, it has been reviewed multiple times but implementation is platform specific and very difficult to cover all possible situations.

Please, could you open a separate issue with this same PR description for further review?

At the moment, the best solution is just let the user manage it on user-code for every specific use case... but I now it's not ideal... HighDPI is a recurring issue...

PR #3031, opened 2023-04-26
Ensure distance is greater than 0 in CameraMoveToTarget

kolunmi:
Hello! This fixes the distance check in `CameraMoveToTarget` to avoid a 0 value. Thank you!


raysan5:
@kolunmi thanks for the review!

PR #3032, opened 2023-04-28
Correction of values ​​used only once in GenMeshCubicmap

Bigfoot71:
I had just noticed that the mapWidth and mapHeight values ​​were only used as a limit in the for loop while they could be used throughout the function.


raysan5:
@Bigfoot71 I think it would be better to just remove `mapWidth` and `mapHeight` and use `cubicmap.width` and `cubicmap.height` everywhere instead.

Bigfoot71:
Done and tested

raysan5:

PR #3033, opened 2023-04-29
WingBGI link in "HISTORY.md" now leads to the correct website!

Stopfield:
It's just a simple issue.
The old link led to an unregistered site (http://www.codecutter.net/tools/winbgim/). The actual domain has changed to (https://winbgim.codecutter.org/).  I checked the wayback machine, it's the same site, take a look: https://web.archive.org/web/20190421035959/http://www.codecutter.net/tools/winbgim/.


raysan5:
@Stopfield nice! thanks for the review! 👍🙂

PR #3037, opened 2023-05-01
[zig build] update to latest zig master

star-tek-mb:
update for `zig 0.11.0-dev.2868+1a455b2dd` and upper

main changes:

- `lib.install()` or `exe.install()` replaced with `b.installArtifact`
- `lib.setOutputDir()` and `exe.setOutputDir()` are removed
- root `build.zig` just calls `src/build.zig`

All examples are working. But output path is set to default `zig-out`. There are some workarounds, but it requires to set CLI paramater `--prefix`, so it's kinda complicated.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,17 +2,5 @@ const std = @import("std");
 const raylib = @import("src/build.zig");
 
 pub fn build(b: *std.Build) void {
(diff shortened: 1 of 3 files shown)


raysan5:
@star-tek-mb thanks for the update!

PR #3043, opened 2023-05-04
Lightmap example.

nullstare:
I used basic_light example by Chris Camacho (@codifies) as a base. Not sure if he should be mentioned in the contributors , though I'm not sure how much of his code there still is. Also copied some textures since there was no suitable ones in the shader resources.


raysan5:
@nullstare Thank you very much! It's a very nice example! I'm just reviewing the TABS to be 4 spaces but I can do it! :)

PR #3045, opened 2023-05-06
Fixed broken build.zig files.

Gamer-Kold:
I noticed the build.zig file wasn't working with the latest zig compiler. So I went ahead and fixed it. It went ahead and built fine on my linux machine, and I used the resulting library to compile the basic example, so I don't think there were any hidden errors.

(Small side note, is the zig build script community maintained? I'd want to add a notice to the build file itself if that was the case.)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,6 +1,6 @@
 const std = @import("std");
 const raylib = @import("src/build.zig");
 
-pub fn build(b: *std.Build) void {
+pub fn build(b: *std.build.Builder) void {
     raylib.build(b);
 }
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,6 +1,6 @@
 const std = @import("std");
 
-pub fn addRaylib(b: *std.Build, target: std.zig.CrossTarget, optimize: std.builtin.OptimizeMode) *std.Build.CompileStep {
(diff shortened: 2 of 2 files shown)


raysan5:
@Gamer-Kold Thanks for the review, actually it was reviewed last week I think...

Yes, the zig build script is community maintained, feel free to add a note if you think it will be helpful for other users.

star-tek-mb:
latest raylib should be built with latest zig. if you are using zig 0.10.0 use older raylib tagged version please.

@raysan5 can we revert this please?

Gamer-Kold:
Makes sense. It didn't work on my machine with master; but if it's working on most other people's machines then I have no reason to hold back the revert. Not when I can keep a custom build file anyway. 

As an aside, I thought it was using a deprecated version of the system; a note in the build script stating which version it has been tested with (as said previously) would really be nice.

star-tek-mb:
@Gamer-Kold yeah. i think we should leave a note with specific zig version `build.zig` file was tested with. can you do that? i would appreciate your work.

Gamer-Kold:
I'll try, I'll also revert the pull request since it seems thay hasn't happened yet.

PR #3048, opened 2023-05-07
Fix vs code project

Shoozza:
Removes unnecessary trailing whitespace from VSCode Project template
Fixes compile issues due to SetCameraMode call and missing argument to UpdateCamera call.


raysan5:
@Shoozza thanks for the review!

Issue #2041, opened 2021-10-09
Proper key repeat behaviour


tomc1998:
Hello, is there a way to get proper key repeat behaviour with non-char keys like backspace?

I'm implementing a console, and text input is really awkward because i can't hold down backspace to delete chars. GetKeyPressed doesn't detect key repeats, GetCharPressed *does* detect repeats but only works for characters (not non-char keys like backspace).

I can't implement manually with IsKeyDown because repeat rates differ across OSes.

If there was a GetRepeatRate() function, that would suffice (although still annoying).

On the other hand, adding repeats to GetKeyPressed would probably break a lot of input systems that rely on this being a single 'key down'.

Is there another function I'm missing, e.g. GetKeyPressedRepeat? Maybe copy the GLFW system with a callback that passes a `repeat` bool val? Alternatively can i just hook into GLFW without breaking stuff?

tomc1998:
hm, how can i measure the repeat rate with IsKeyDown? I'm talking about getting the OS-configured repeat rate & repeat delays - e.g. if you hold down the 'a' key, it triggers once, waits, then triggers repeatedly with a fixed delay

Surely IsKeyDown is just returning true every frame if the key is down, and false if the key is up?

Is it possible  to just hook directly into GLFW here?

jmorel33:
> hm, how can i measure the repeat rate with IsKeyDown? I'm talking about getting the OS-configured repeat rate & repeat delays - e.g. if you hold down the 'a' key, it triggers once, waits, then triggers repeatedly with a fixed delay
> 
> Surely IsKeyDown is just returning true every frame if the key is down, and false if the key is up?
> 
> Is it possible to just hook directly into GLFW here?

Hey Tom, my suggestion, as I am in the same world as you about making a console etc etc, is to look as how GLFW handles key repeat.  If you like it and want to create a hook to that, it is possible without too much customisation.

You don't have to go with the "I'm afraid that" speech you often read.  Just find the hook for what you need and implement it yourself .  Then you can always propose your PR see if it interests the author or not.

The other way around is just to not use the RayLib keyboard management (if you are making a Windows only app) and use the oem keyboard handling library which should be exposed in your mingw install (if that is what you are using).

TBH I find the Windows implementation of key bindings incredibly easy to work with.

good luck!

Jacques

raysan5:
[...]

PR #3056, opened 2023-05-12
Add option to process key repeats as presses

tomc1998:
A while ago I opened #2041 asking about key repeat behaviour in raylib

You mentioned this wouldn't be put into raylib & I should instead patch the source

Recently I settled on a very simple fix that was easier to keep rebased against upstream. It was so unobtrusive I thought I'd add an option to toggle it & submit a PR, on the off-chance it gets merged :pleading_face: 

It's *massively* useful when making tools that need text input, because otherwise you can't hold backspace to delete text, or arrow keys, etc. You can use IsKeyDown / IsKeyUp & a timer to simulate key repeats, but this won't match the user's keyboard settings. Basically the only way to get good text input is to patch raylib, which is a shame!

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1105,6 +1105,7 @@ RLAPI bool IsKeyDown(int key);                                // Check if a key 
 RLAPI bool IsKeyReleased(int key);                            // Check if a key has been released once
 RLAPI bool IsKeyUp(int key);                                  // Check if a key is NOT being pressed
 RLAPI void SetExitKey(int key);                               // Set a custom key to exit program (default is ESC)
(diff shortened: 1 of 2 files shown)


raysan5:
@tomc1998 I'm afraid I'm not merging this solution, I don't like the proposed approach. I'll try to review it in another way. For the moment I keep the issue open. In any case, if that solution works for your use case, that's ok.

BrianWill:
In case anyone comes to this conversation late, I found this solution solves this problem for me in the current version:

` // Evals true in the frame where the A key is initially pressed and then in later frames if key is held.
// (the delay after the initial press and the repeat rate are determined by the platform, I believe)
if  rl.IsKeyPressed(rl.KeyA) || rl.IsKeyPressedRepeat(rl.KeyA) {...}`

PR #3067, opened 2023-05-15
Fix Android app freeze after calling CloseWindow()

Bigfoot71:
Fixed that the Android application was not closed properly after calling `CloseWindow()` and continued to run.

**Edit**: I want to clarify the call to `exit(0)` in `android_main`. This will not be a problem in normal use of Raylib on Android, only if there are other processes in progress from Java/Kotlin this may in some cases not complete them, hence the precision in the comments, I hope it's pretty clear.

Otherwise, you have to handle closing the application through Java, which can complicate/messy things for the user. So I thought to myself that if a person would be able to tweak the Java code of his Raylib game, he would be able to do the right things to properly close his application.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -718,6 +718,23 @@ void android_main(struct android_app *app)
 
     // NOTE: Return codes != 0 are skipped
     (void)main(1, (char *[]) { arg0, NULL });
+
+    // Finish native activity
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Thanks! This is a nice improvement! :D

PR #3069, opened 2023-05-18
Update README.md

uwussimo:
image

**the animation logo was sticking to the text, I tried to make it nicer:**

image


raysan5:
@uwussimo Thanks for pointing the issue, I just reviewed it myself to add inline styling to the image instead of split changes into several lines.

PR #3073, opened 2023-05-21
Fix integer overflow when calculating memory allocation size

szsam:
Allocating memory with a size controlled by an external user can result in integer overflow.
Also, we need to check return values of malloc/calloc for null.

-----------
* There are more than 20 such patterns in the codebase. I will try to find and fix them all if this patch looks good to you.
* Is there a better way to indicate memory allocation error for function `GenImageColor()`, other than returning an `Image` with the `data` field set to `NULL`?


raysan5:
@szsam I'm afraid I'm not applying this change for now, it will imply a big redesign of many raylib parts to validate all user inputs and allocation results.

PR #3074, opened 2023-05-21
Replaced GenImageGradientH and GenImageGradientV with GenImageLinearGradient

danemadsen:
This allows the user to create linear gradients with a single function. In addition to horizontal and vertical gradients this function also allows the user to create diagonal gradients and gradients in any direction in 360 degrees.

Screenshot from 2023-05-21 18-58-09
Screenshot from 2023-05-21 18-58-16
Screenshot from 2023-05-21 18-58-21

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1239,7 +1239,7 @@ RLAPI bool ExportImageAsCode(Image image, const char *fileName);                
 
 // Image generation functions
 RLAPI Image GenImageColor(int width, int height, Color color);                                           // Generate image: plain color
-RLAPI Image GenImageLinearGradient(int width, int height, int direction, Color start, Color end);        // Generate image: linear gradient
+RLAPI Image GenImageGradientLinear(int width, int height, int direction, Color start, Color end);        // Generate image: linear gradient
 RLAPI Image GenImageGradientRadial(int width, int height, float density, Color inner, Color outer);      // Generate image: radial gradient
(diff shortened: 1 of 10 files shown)


src/raylib.h
```diff
@@ -1239,8 +1239,7 @@ RLAPI bool ExportImageAsCode(Image image, const char *fileName);
 
 // Image generation functions
 RLAPI Image GenImageColor(int width, int height, Color color);                                           // Generate image: plain color
-RLAPI Image GenImageGradientV(int width, int height, Color top, Color bottom);                           // Generate image: vertical gradient
-RLAPI Image GenImageGradientH(int width, int height, Color left, Color right);                           // Generate image: horizontal gradient
+RLAPI Image GenImageLinearGradient(int width, int height, int direction, Color start, Color end);        // Generate image: linear gradient
```
raysan5, line 1242: To be aligned with `GenImageGradientRadial()`, this function should be named `GenImageGradientLinear()`
Also, direction parameter should be detailed in function description.

raysan5:
@danemadsen Despite being a breaking change for the library, I don't think many users depend on those replaced functions so I can merge it.

Only concern is the direction parameter, it could be confusing for some users, it would require some small description on the function comment.

src/raylib.h
danemadsen: Added a comment and renamed the function

raysan5:
@danemadsen Thank you very much for this addition!

danemadsen:
@raysan5 My pleasure :)

PR #3077, opened 2023-05-22
Add GenImageGradientSquare

danemadsen:
This gradient is useful for users who want to implement a falloff map for procedural generation, or potentially for users who want to create a custom vignette for their screen.
Screenshot from 2023-05-22 14-26-08
Screenshot from 2023-05-22 14-26-16

### DIFF
--- a/examples/textures/textures_image_generation.c
+++ b/examples/textures/textures_image_generation.c
@@ -13,7 +13,7 @@
 
 #include "raylib.h"
 
-#define NUM_TEXTURES  7      // Currently we have 7 generation algorithms
+#define NUM_TEXTURES  10      // Currently we have 8 generation algorithms but some are have multiple purposes (Linear and Square Gradients)
 
 //------------------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


raysan5:
@danemadsen Please, would it be possible to reduce the number of parameters? In raylib I try to keep all functions under [**7 parameters**](https://github.com/raysan5/raylib/wiki/raylib-syntax-analysis#16-functions-parameters). No problem if functionality exposed is limited, users can implement their own versions looking at source code if they really require more fine-grained control.

danemadsen:
Ok gimme a sec I'll change it slightly

danemadsen:
@raysan5 Ok, resolved. The only thing to note now is that the gradients will now only follow the dimensions of the image, so the gradient will not be square in a rectangular image. 

That being said, I only need it for a falloff map so its a fair call to remove it.

raysan5:
@danemadsen Perfect! Thank you very much for the addition and the review! 🙂

PR #3078, opened 2023-05-23
Added ImageRotate

danemadsen:
This function can replace ImageRotateCW and ImageRotateCCW however I left them in as it would make this a breaking change. If you want me to rip them out though i can.

In addition to allowing clockwise and counter clockwise rotation this function also allows rotation to any degree in the range of -359 to 359.

Screenshot from 2023-05-23 21-44-35
Screenshot from 2023-05-23 21-59-11
Screenshot from 2023-05-23 21-59-17
textures_image_rotate


src/rtextures.c
```diff
@@ -2118,6 +2118,65 @@ void ImageFlipHorizontal(Image *image)
...
+{
+    // Security check to avoid program crash
+    if ((image->data == NULL) || (image->width == 0) || (image->height == 0)) return;
+
+    if (image->mipmaps > 1) TRACELOG(LOG_WARNING, "Image manipulation only applied to base mipmap level");
+    if (image->format >= PIXELFORMAT_COMPRESSED_DXT1_RGB) TRACELOG(LOG_WARNING, "Image manipulation not supported for compressed formats");
+    else
+    {
+        double rad = degrees * PI / 180.0;
+        double sin_rad = sin(rad);
```
raysan5, line 2132: Please could you use `float` instead of `double`? I tried to avoid `double` usage in raylib.

Please, could you also review code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

raysan5:
@danemadsen This is a nice addition! Just added some details for review.

I prefer to keep the other `ImageRotate*()` functions for now, I'm using them in some tools, but maybe I remove them in the future.

src/rtextures.c
danemadsen: Resolved as requested

PR #3079, opened 2023-05-24
Add new file formats to FAQ

Luramoth:
I noticed some file formats from the new release of raylib were missing so I decided to help out and update it!

added M3D model format and QOA file format to the list


raysan5:
@Luramoth Nice! Thank you very much for the review!

PR #3080, opened 2023-05-25
Add Chunk3D to help with procedural generation

danemadsen:
I understand if you think its out of scope and dont want to include it, I'm just using it in my own project.


raysan5:
@danemadsen Sorry, this is out of scope for the moment. I try to avoid adding new structures to raylib base library and just let the users create them as required on their own projects.

danemadsen:
Thats ok, thanks anyway. Ill package all my work into a header file anyway for later use

PR #3081, opened 2023-05-26
Use "vertices" instead of "points" in CheckCollisionPointPoly

archydragon:
Cosmetic code change. Having variables `point` and `points` as a single function parameters is a bit confusing, as for me. Especially when the comment refers to "array of vertices" :)


raysan5:
I agree the naming could be a bit confusing but it is aligned with other functions:
```c
DrawLineStrip(Vector2 *points, int pointCount, Color color);    
DrawTriangleFan(Vector2 *points, int pointCount, Color color);  
DrawTriangleStrip(Vector2 *points, int pointCount, Color color);
```
No function refers to 2d screen points as vertices. Also note that vertices could refer to multiple different data attributes: position, texcoords, normals, tangents...

I know it's a tricky one, no easy to find a proper naming...

archydragon:
Fair point, thanks for fast response!

orcmid:
This will not break any code, because it is about the naming of signature variables.

However, it is vertexCount, just as it was pointCount.  Replacing `points[]` with `vertices[]` is OK, apart from Ramon's consistency objection.  

This is always a place where C Language is a bit clumsy.  We tend to speak of `vertex[i]` and `point[i]` and don't have a clean concept for array names apart from member-pointing names.

In documentation I finesse this using `vertex[]` and `point[]`.  That's a big change to prototypes though.

PR #3082, opened 2023-05-27
Fix a link in the FAQ

jasonliang-dev:


raysan5:
@jasonliang-dev thanks for the review

PR #3084, opened 2023-05-27
Add function LoadTextureFromMemory();

Wytekol:
This function loads a texture from memory. Similar to LoadImageToMemory().


raysan5:
@Wytekol I'm afraid I'm not adding this function, users requiring it can just load the image from memory.

VaguinerG:
> @Wytekol I'm afraid I'm not adding this function, users requiring it can just load the image from memory.

Please reopen for discussion.

There is real use for the function, mainly to store textures in the binary.

PR #3085, opened 2023-05-28
Fixed compile on OpenBSD

rayit:
Had to remove dl and atomic for compiling on OpenBSD


raysan5:
@rayit Thanks for the review! Afaik, not many raylib users are on OpenBSD, glad to know it also works on that platform now.

PR #3086, opened 2023-05-28
Enhanced cmake part for OpenBSD

rayit:
I think this is nicer as I now use default part
Had to remove dl and atomic for OpenBSD otherwise it did not make


raysan5:
@rayit thanks for the review!

PR #3089, opened 2023-05-30
Catch key repeats for non-char keys

dbechrd:
This fixes a bug in Raylib's input handling where non-character keys have their repeat events completely discarded in a way that makes it impossible to detect such events (without calling IsKeyPressed on every key code every frame and re-inventing the key queue).

Specifically, this bugfix enables you to handle Backspace, Delete, arrow keys, etc. repeat events in a way that respects the user's OS key repeat delay and interval settings.

It seems like glfw triggers both a PRESS and REPEAT event on the first frame, so it's necessary to change the order of the code slightly such that you can check the IsKeyPressed state in order to de-duplicate adding both events for the same key to the queue (though if the queue also stored some extra metadata about the event, this wouldn't be necessary and would give the caller more discretion). Theoretically this could discard legitimate events if the user presses the same key twice during a single frame, but considering the general use-case of Raylib for real-time applications, this seems quite unlikely. The trade-off seems very worth it vs. the current behavior of completely discarding all repeat events.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5324,6 +5324,16 @@ static void KeyCallback(GLFWwindow *window, int key, int scancode, int action, i
 {
     if (key < 0) return;    // Security check, macOS fn key generates -1
 
+    // Check if there is space available in the key queue
+    if (CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE &&
+        (action == GLFW_PRESS || action == GLFW_REPEAT) &&
+        !IsKeyPressed(key))
+    {
+        // Add character to the queue
+        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
+        CORE.Input.Keyboard.keyPressedQueueCount++;
+    }
+
     // WARNING: GLFW could return GLFW_REPEAT, we need to consider it as 1
     // to work properly with our implementation (IsKeyDown/IsKeyUp checks)
     if (action == GLFW_RELEASE) CORE.Input.Keyboard.currentKeyState[key] = 0;
@@ -5334,14 +5344,6 @@ static void KeyCallback(GLFWwindow *window, int key, int scancode, int action, i
     if (((key == KEY_CAPS_LOCK) && ((mods & GLFW_MOD_CAPS_LOCK) > 0)) ||
         ((key == KEY_NUM_LOCK) && ((mods & GLFW_MOD_NUM_LOCK) > 0))) CORE.Input.Keyboard.currentKeyState[key] = 1;
 #endif
-
-    // Check if there is space available in the key queue
-    if ((CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE) && (action == GLFW_PRESS))
-    {
-        // Add character to the queue
-        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
-        CORE.Input.Keyboard.keyPressedQueueCount++;
-    }
 
     // Check the exit key to set close window
     if ((key == CORE.Input.Keyboard.exitKey) && (action == GLFW_PRESS)) glfwSetWindowShouldClose(CORE.Window.handle, GLFW_TRUE);


dbechrd:
This is WIP. There may still be a few bugs. I think it's double-detecting releases at the moment. Do not merge yet.

dbechrd:
I believe this is working correctly now. Someone else may want to sanity check it and/or add test coverage if that's somehow possible. But I did some basic manual testing on Windows 11 in my app using a custom textbox control and it seems to be working as one would expect.

If you're curious, this is how I'm using the key callbacks in my code:
```c
    int key = GetKeyPressed();
    while (key) {
        if (!isascii(key)) {
            stb_textedit_key(&str, &stbState, key);
        }
        key = GetKeyPressed();
    }
    int ch = GetCharPressed();
    while (ch) {
        if (isascii(key)) {
            stb_textedit_key(&str, &stbState, ch);
        }
        ch = GetCharPressed();
    }
```

nullstare:
Is the "!IsKeyPressed(key)" really necessary? I haven't noticed any problems without it.

dbechrd:
> Is the "!IsKeyPressed(key)" really necessary? I haven't noticed any problems without it.

Without it, keys will double-report on the first frame the key is pressed. I.e. glfw calls the callback twice, and it gets added to the queue twice. If you replace `stb_textedit_key` with a printf in my example code above, then tap a key as quickly as you can, you can see that it's impossible to ever get it to only report a single keypress without the check. In a textbox control this causes "ff" to appear when you press the "f" key once, quickly, for example.

raysan5:
@dbechrd Is this change somewhat related to #3096?

Are you building a text editor using `stb_textedit` and `raylib`?

I'm also building a text editor and I didn't need any change to raylib so, I need to review this change carefully becasue it could imply redesigning my implementation. I want to evaluate the real need for it (same as the linked PR).

dbechrd:
> @dbechrd Is this change somewhat related to #3096?
> 
> Are you building a text editor using `stb_textedit` and `raylib`?
> 
> I'm also building a text editor and I didn't need any change to raylib so, I need to review this change carefully becasue it could imply redesigning my implementation. I want to evaluate the real need for it (same as the linked PR).

Seems related. Yes, I'm writing a single-line textbox control and would like the arrow key to function properly, which is not possible unless `GetKeyPressed` supports repeat properly (without querying the Win32 api to get the key repeat rate and writing a bunch of timing code yourself).

I ended up writing a bunch of timing code anyways for other reasons, but I still think Raylib should just support this behavior better internally.

--------------

On a somewhat related note, detecting double-click properly also isn't possible in Raylib (or GLFW !?) at the moment, which I posted some info about here:
https://discord.com/channels/426912293134270465/426912531429457930/1113915958986285186
```
TIL glfw doesn't have a double-click event.

In SDL, it's as simple as checking if SDL_MouseButtonEvent.clicks is > 1. SDL also accounts for the OS-specific double click speed setting, as well as how far the mouse moved between clicks.

I might try to take a look at glfw's code later and see if I can create a PR for them to handle this better, but I wonder if there's some way for Raylib to handle this without glfw's help?

Specifically, on Windows, we would need to call Win32 API's GetDoubleClickTime() method, and fallback on 500ms for platforms where we don't know how to query the user's mouse settings (e.g. it's very tricky on X11). Is it viable to add a platform-specific function, e.g. IsMouseButtonDoubleClicked() to Raylib's core?
```
and here:
https://discord.com/channels/426912293134270465/426912531429457930/1113922425239900211
```
It actually seems like Windows has WM_*BUTTONDBLCLK events so adding support to glfw would be rather straightforward on Windows. I think I'll submit an issue on their github about it.
```

--------------------

Here's my current textbox code built on top of stb_textedit.
https://github.com/dbechrd/RayNet/blob/81b41cb4fc6c34456db730137fd5280550bc978d/src/common/ui/ui.cpp#L403

It supports some things that even Dear Imgui doesn't currently support like double-click + drag to select words at a time. (Triple-click + drag to select lines at a time could also be implemented using the same pattern, but my textbox needs are currently single-line only so I haven't implemented that yet). Arrow keys are working in my project with correct repeats since I patched Raylib on my local machine with the changes this PR is proposing.

I also noticed at least one off-by-one pixel error in MeasureText, but haven't dug deep enough to root cause them and submit them as Raylib bugs. Hence the few random "+ 1"s in my code (see `RN_stb_get_char_width`, sorry for the profane comments, I was clearly upset at the time 😅 ). They could also just be me doing something stupid.

Feel free to reach out to me on Discord if you have any questions about my use-case, or whatever else that might help you with your text editor or Raylib.

raysan5:
@dbechrd Recently `IsKeyPressedRepeat()` functionality was added: https://github.com/raysan5/raylib/pull/3245, could this PR be closed?

raysan5:
After reviewing, I'm closing this PR. It changes the default behaviour of the function and it could affect existing codebases using it. The recently added `IsKeyPressedRepeat()` should cover the proposed use case.

dbechrd:
@raysan5 Thanks for the ping. I haven't had time to work on my project recently, but if I find anything interesting about the new function when I get back to it I'll let ya know. Good luck w/ your text editor!

raysan5:
@dbechrd Hey! Thank you very much! I neither had time to work on my text editor since July, been busy with other project! 😄

dbechrd:
@raysan5 I just took the time to upgrade Raylib. PlaySoundMulti was broken so I had to refactor a bunch of audio stuff, but it was fairly straightforward and the new API is much better so no big deal.

I just got to looking at this specific issue and realized that I don't understand how `IsKeyPressedRepeat` is supposed to help me with a text editor? Isn't that for checking a specific key?

My code uses a combination of GetKeyPressed() and GetCharPressed() at the moment to read from the queue. I'm not sure how I could use IsKeyPressed* variants without writing explicit code for every possible key on the keyboard?

When I first read your comment I was expecting to find a GetKeyPressedRepeat() or GetKeyPressed(&repeat) function to be what was added.

dbechrd:
Okay, with a bit more testing, I remember the problem now. `GetCharPressed()` works fine for both the initial key event and repeat events, and respects OS repeat settings. The problem is that GetCharPressed() cannot be used for non-ascii keys, like the arrow keys, which is something that needs to work for a text box. So this bug still exists in the latest Raylib master branch.

dbechrd:
Alright so I tried refactoring the code to use `IsKeyPressedRepeat()` instead of GetKeyPressed() with my patch to Raylib's code. It's a bit messier (since I have to check every single possible key), but I made it fairly terse with a macro and it's is at least possible without modifying Raylib now.

First, I should mention that I have helper functions that manage capturing focus in UI with higher input priority, which are quite simple but relevant to the example below:
```cpp
bool KeyPressed(int key, bool include_repeats = false) {
    if (KeyboardCaptured()) {
        return false;
    }
    return IsKeyPressed(key) || (include_repeats && IsKeyPressedRepeat(key));
}
bool KeyDown(int key) {
    if (KeyboardCaptured()) {
        return false;
    }
    return IsKeyDown(key);
}
```

Original code for handling non-ascii keys alongside stb_textedit (with my Raylib patch this PR is proposing)
```cpp
int key = GetKeyPressed();
while (key) {
    if (io.KeyDown(KEY_LEFT_CONTROL) || io.KeyDown(KEY_RIGHT_CONTROL)) {
        key |= STB_TEXTEDIT_K_CTRL;
    }
    if (io.KeyDown(KEY_LEFT_SHIFT) || io.KeyDown(KEY_RIGHT_SHIFT)) {
        key |= STB_TEXTEDIT_K_SHIFT;
    }
    // Notice how this lets me generically forward *any* non-ASCII key to stb_textedit with very little effort.
    if (RN_stb_key_to_char(key) == -1) {
        stb_textedit_key(&str, &stbState, key);
    }
    if (key == (STB_TEXTEDIT_K_CTRL | KEY_A)) {
        stbState.cursor = text.size();
        stbState.select_start = 0;
        stbState.select_end = text.size();
    }
    if (key == (STB_TEXTEDIT_K_CTRL | KEY_X)) {
        int selectLeft = MIN(stbState.select_start, stbState.select_end);
        int selectRight = MAX(stbState.select_start, stbState.select_end);
        int selectLen = selectRight - selectLeft;
        if (selectLen) {
            std::string selectedText = str.data.substr(selectLeft, selectLen);
            SetClipboardText(selectedText.c_str());
            stb_textedit_cut(&str, &stbState);
        }
    }
    if (key == (STB_TEXTEDIT_K_CTRL | KEY_C)) {
        int selectLeft = MIN(stbState.select_start, stbState.select_end);
        int selectRight = MAX(stbState.select_start, stbState.select_end);
        int selectLen = selectRight - selectLeft;
        if (selectLen) {
            std::string selectedText = str.data.substr(selectLeft, selectLen);
            SetClipboardText(selectedText.c_str());
        }
    }
    if (key == (STB_TEXTEDIT_K_CTRL | KEY_V)) {
        const char *clipboard = GetClipboardText();
        if (clipboard) {
            size_t clipboardLen = strlen(clipboard);
            if (clipboardLen) {
                stb_textedit_paste(&str, &stbState, clipboard, clipboardLen);
            }
        }
    }
    if (key == KEY_TAB) {
        // tell next textbox that draws to active
        tabToNextEditor = true;
    } else if (key == (STB_TEXTEDIT_K_SHIFT | KEY_TAB)) {
        // tell previously drawn textbox to activate next frame
        tabToPrevEditor = lastDrawnEditor;
    }
    key = GetKeyPressed();
}
```

New code, which uses `IsKeyDownRepeat()` explicitly for each possible key:
```cpp
bool ctrl = io.KeyDown(KEY_LEFT_CONTROL) || io.KeyDown(KEY_RIGHT_CONTROL);
bool shift = io.KeyDown(KEY_LEFT_SHIFT) || io.KeyDown(KEY_RIGHT_SHIFT);

int mod = 0;
if (ctrl) mod |= STB_TEXTEDIT_K_CTRL;
if (shift) mod |= STB_TEXTEDIT_K_SHIFT;

// With the unmodified Raylib API, I have to explicitly check each individual key in order to detect repeat events:
#define STB_KEY(key) if (io.KeyPressed(key, true)) { stb_textedit_key(&str, &stbState, mod | key); }
STB_KEY(KEY_LEFT);
STB_KEY(KEY_RIGHT);
STB_KEY(KEY_UP);
STB_KEY(KEY_DOWN);
STB_KEY(KEY_DELETE);
STB_KEY(KEY_BACKSPACE);
STB_KEY(KEY_HOME);
STB_KEY(KEY_END);
STB_KEY(KEY_PAGE_UP);
STB_KEY(KEY_PAGE_DOWN);
STB_KEY(KEY_Z);
#undef STB_KEY

if (ctrl && io.KeyPressed(KEY_A)) {
    stbState.cursor = text.size();
    stbState.select_start = 0;
    stbState.select_end = text.size();
}
if (ctrl && io.KeyPressed(KEY_X, true)) {
    int selectLeft = MIN(stbState.select_start, stbState.select_end);
    int selectRight = MAX(stbState.select_start, stbState.select_end);
    int selectLen = selectRight - selectLeft;
    if (selectLen) {
        std::string selectedText = str.data.substr(selectLeft, selectLen);
        SetClipboardText(selectedText.c_str());
        stb_textedit_cut(&str, &stbState);
    }
}
if (ctrl && io.KeyPressed(KEY_C, false)) {
    int selectLeft = MIN(stbState.select_start, stbState.select_end);
    int selectRight = MAX(stbState.select_start, stbState.select_end);
    int selectLen = selectRight - selectLeft;
    if (selectLen) {
        std::string selectedText = str.data.substr(selectLeft, selectLen);
        SetClipboardText(selectedText.c_str());
    }
}
if (ctrl && io.KeyPressed(KEY_V, true)) {
    const char *clipboard = GetClipboardText();
    if (clipboard) {
        size_t clipboardLen = strlen(clipboard);
        if (clipboardLen) {
            stb_textedit_paste(&str, &stbState, clipboard, clipboardLen);
        }
    }
}
if (io.KeyPressed(KEY_TAB, true)) {
    if (shift) {
        // tell previously drawn textbox to activate next frame
        tabToPrevEditor = lastDrawnEditor;
    } else {
        // tell next textbox that draws to activate
        tabToNextEditor = true;
    }
}
```

I still think having `GetKeyPressed()` return repeat events (or adding a `GetKeyPressedRepeat()` queue) would be very handy for any use-case that generically needs to handle all non-ascii keys in a way that uses repeat information. I don't see why the `GetKeyPressed()` queue should handle repeats any differently than the `GetCharPressed()` queue. It's weird that one of them discards repeats while the other reports them, and I highly suspect anyone handling events via the queues (rather than polling explicit keys) would expect repeats to show up.

raysan5:
@dbechrd I'm reopening it for following discussion. One point to avoid key-repeat is consistency between platforms. GLFW allows automatic key repeat processing but raylib works in multiple platforms and it can suppose an inconsistency between platforms.

raysan5:
I'm closing this PR, no more changes on inputs planned for next release.

PR #3092, opened 2023-05-30
Fix full screen resolution issue

tanolino:
I was writing a image viewer in Raylib and high res images appeared  in the same low resolution as the small window. Investigating the issue showed that glfwSetWindowMonitor is called with the screen resolution, not the current monitor resolution.

It is possible I am misunderstanding things here and the current work flow is desired for pixelart games with not resizable windows ?


tanolino:
I found another solution without my modifications, which was already in this library. Some games call it "fullscreen window". Going from there also allow switching to my desired fullscreen behavior. Have a nice day.

PR #3095, opened 2023-06-02
Simple Analog Stick 8 Direction Fix

PixelPhobicGames:
I Added Analog Stick Support to the rcamera module a Week or Two Ago, However looking back this code only allowed for 4 Directions of Movement, This Change adds the full range of Movement to the Right Analog Stick.


raysan5:
@PixelPhobicGames Nice! Thanks for the review! :D

PR #3102, opened 2023-06-10
Added a wrapper for glfwSetWindowAspectRatio, called SetWindowAspectRatio

Khhs167:
Added the function `SetWindowAspectRatio(int numerator, int denominator)` that simply calls `glfwSetWindowAspectRatio`. I believe this could come in handy for some.

Made a minor mistake so there are 2 identically named commits, sorry about that.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -957,6 +957,7 @@ RLAPI void SetWindowPosition(int x, int y);                       // Set window 
 RLAPI void SetWindowMonitor(int monitor);                         // Set monitor for the current window (fullscreen mode)
 RLAPI void SetWindowMinSize(int width, int height);               // Set window minimum dimensions (for FLAG_WINDOW_RESIZABLE)
 RLAPI void SetWindowSize(int width, int height);                  // Set window dimensions
+RLAPI void SetWindowAspectRatio(int numerator, int denominator);  // Set window aspect ratio (for FLAG_WINDOW_RESIZABLE)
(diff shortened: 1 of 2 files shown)


raysan5:
@Khhs167 Are you using it? What is the use case? I try to avoid adding new functions just "in case it could be useful for someone".

Khhs167:
Personally, I'm using it in a Pac-Man clone. Sometimes you want to do something like render to a fixed resolution(let's say, you're doing a retro-like game, maybe you want to lock it to 256x240, or something in that fashion), or you've designed all UI for 16:9, and you want to allow the user to resize the window.

In these cases this little function would come incredibly in handy. You *can* also resize the window etc, but I've found that to be extremely buggy and behave weirdly, whilst `glfwSetWindowAspectRatio` has been working smoothly. You mot likely want your game that is rendered in 4:3 to *stay* in 4:3, and not change aspect ratios and have ugly stretching etc.

It's mostly useful for cases where you render to lower resolutions, or just don't want to have to deal with multiple aspect ratios.

Hope this all made sense :)

raysan5:
@Khhs167 As per my understanding, this function only forces the aspect ratio of the window when resized but it does not modify the framebuffer size. [`core_window_letterbox` example](https://github.com/raysan5/raylib/blob/master/examples/core/core_window_letterbox.c) already covers that use-case, even with an non-restrictive aspect-ratio. Sorry, I can't see a real need for this function on raylib.

PR #3103, opened 2023-06-10
GLTF: fix segfault in animNormals memcpy when mesh.normals == NULL

charles-l:
I ran across a GLTF file with animation data but no normals which caused raylib to segfault on a memcpy since normals were expected to be defined. This PR fixes it.


raysan5:
@charles-l thanks for the review!

Issue #3100, opened 2023-06-05
[core] `GetTouchPointCount()` keeps returning 1 after the last touch is released


chemaguerra:
### Issue description

In `PLATFORM_WEB` (Emscripten) raylib's `GetTouchPointCount()` insists in returning 1 after the last finger touch is released.

`GetTouchPointCount()` correctly returns 0 on start, and then correctly returns the number of fingers touching the screen. However, when the last finger is released, the function keeps returning 1 (instead of 0).

The problem seems to be that `CORE.Input.Touch.currentTouchState` is not used to zero out the touch count after the last `EmscriptenTouchCallback` is received.

As a matter of fact, `IsMouseButtonReleased()` (which makes use of `currentTouchState`) returns true when the last touch is released (which can be used as a cheap workaround).

### Environment

Reproducible on all the Android devices I have access to, using the most recent version of Chrome (=> `PLATFORM_WEB`).

Not happening if I compile/run my project as an .apk (=> Not happening on `PLATFORM_ANDROID`). I don't have any `PLATFORM_DESKTOP` devices with a touchscreen at my disposal to test, though.

### Issue Screenshot

![report](https://github.com/raysan5/raylib/assets/33331554/37fdcd32-6f63-47f4-86fc-169772eaa393)

Reproducible in raylib's multi-touch demo.

### Workaround

The workaround I opted for at the moment is to create a function to query `currentTouchState` and trim myself the list of touches I build with `GetTouchPointId/Position()`.

raysan5:
@chemaguerra Thanks for reporting, I though that issue was already fixed... In any case, multi-touch improvements are required, afair, at this moment the specific touch ids are not read processed properly, required for some gesture events.

raysan5:
@ubkp thanks for the detailed review of the issue and the fix provided! very professional!

chemaguerra:
Awesome. I concur in thanking you. :-)

PR #3107, opened 2023-06-11
Continuation of support for ES3/WebGL2

chemaguerra:
This is in response to:
- https://github.com/raysan5/raylib/discussions/2447
- https://github.com/raysan5/raylib/commit/e497603678823dc93f01dd9686819a92d780ad67

I continued the work that @raysan5 started for OpenGL ES 3.0 context creation:

1- Fixed `#define GRAPHICS_API_OPENGL_33`.
2- Adapted `rlActiveDrawBuffers()` which now can be used to successfully initialize MRTs in WebGL2.
3- Did a (dirty) update to `RLGL.ExtSupported.*` since some ES 2.0 feature extensions can be taken for granted in ES 3.0.
4- Adapted `rlGetGlTextureFormats()` to support floating-point textures in ES 3.0.

5- I also edited the Makefile so raylib builds for ES3 (maybe exclude this on PR merge...).

The changes I am PR'ing have been tested in a couple of experimental projects I am conducting at the moment, in a bunch of web browsers across multiple devices (Windows, MacOS, Android, iPad, iPhone).

(3) needs some work to correctly detect the capabilities of the device. i.e., whether the device supports linear filtering for floating-point textures (many devices don't).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -245,7 +245,8 @@ ifeq ($(PLATFORM),PLATFORM_DRM)
 endif
 ifeq ($(PLATFORM),PLATFORM_WEB)
     # On HTML5 OpenGL ES 2.0 is used, emscripten translates it to WebGL 1.0
-    GRAPHICS = GRAPHICS_API_OPENGL_ES2
+    #GRAPHICS = GRAPHICS_API_OPENGL_ES2
+    GRAPHICS = GRAPHICS_API_OPENGL_ES3      # Uncomment to use ES3/WebGL2 (preliminary support).
 endif
 ifeq ($(PLATFORM),PLATFORM_ANDROID)
     # By default use OpenGL ES 2.0 on Android
--- a/src/Makefile
+++ b/src/Makefile
@@ -245,8 +245,8 @@ ifeq ($(PLATFORM),PLATFORM_DRM)
 endif
 ifeq ($(PLATFORM),PLATFORM_WEB)
     # On HTML5 OpenGL ES 2.0 is used, emscripten translates it to WebGL 1.0
-    #GRAPHICS = GRAPHICS_API_OPENGL_ES2
-    GRAPHICS = GRAPHICS_API_OPENGL_ES3      # Uncomment to use ES3/WebGL2 (preliminary support).
+    GRAPHICS = GRAPHICS_API_OPENGL_ES2
+    #GRAPHICS = GRAPHICS_API_OPENGL_ES3      # Uncomment to use ES3/WebGL2 (preliminary support).
 endif
 ifeq ($(PLATFORM),PLATFORM_ANDROID)
     # By default use OpenGL ES 2.0 on Android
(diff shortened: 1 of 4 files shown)


src/Makefile
```diff
@@ -245,7 +245,8 @@ ifeq ($(PLATFORM),PLATFORM_DRM)
 endif
 ifeq ($(PLATFORM),PLATFORM_WEB)
     # On HTML5 OpenGL ES 2.0 is used, emscripten translates it to WebGL 1.0
-    GRAPHICS = GRAPHICS_API_OPENGL_ES2
+    #GRAPHICS = GRAPHICS_API_OPENGL_ES2
```
raysan5, line 248: I prefer to keep `GRAPHICS_API_OPENGL_ES2` as the default, instead of `GRAPHICS_API_OPENGL_ES3`

src/Makefile
```diff
@@ -344,7 +345,8 @@ endif
 
 ifeq ($(RAYLIB_BUILD_MODE),RELEASE)
     ifeq ($(PLATFORM),PLATFORM_WEB)
-        CFLAGS += -Os
+        #CFLAGS += -Os
```
raysan5, line 348: I prefer to keep the `-Os` on web instead of `-O3`, from [emscripten documentation](https://emscripten.org/docs/optimizing/Optimizing-Code.html#how-to-optimize-code):

> Building with -O3 or -Os can produce an ever better build than -O2, and are worth considering for release builds. -O3 builds are even more optimized than -O2, but at the cost of significantly longer compilation time and potentially larger code size. -Os is similar in increasing compile times, but focuses on reducing code size while doing additional optimization.

src/raylib.h
```diff
@@ -1149,6 +1149,7 @@ RLAPI int GetTouchX(void);                                    // Get touch posit
 RLAPI int GetTouchY(void);                                    // Get touch position Y for touch point 0 (relative to screen size)
 RLAPI Vector2 GetTouchPosition(int index);                    // Get touch position XY for a touch point index (relative to screen size)
 RLAPI int GetTouchPointId(int index);                         // Get touch point identifier for given index
+RLAPI bool GetTouchPointState(int index);                     // Patch for touch events in web browsers
```
raysan5, line 1152: This addition does not seem to belong to this PR.

raysan5:
@chemaguerra Thank you very much for the addition and the testing! Just added a couple of comments.

src/Makefile
chemaguerra: Regarding your 3 comments: my bad! They were not intended to be part of the PR. Those are changes I made afterwards that I require for my project but are not meant to be merged into raylib. Switching by default to ES3 would break existing projects indeed.

src/Makefile
chemaguerra: Same. In my project -O3 runs significantly faster than -Os and the extra build time/size is a price I am willing to pay. But this change wasn't meant to be part of the PR.

src/raylib.h
chemaguerra: Same. Changes I made afterwards not meant to be part of this PR. Actually, this is in relation to this issue: https://github.com/raysan5/raylib/issues/3100

chemaguerra:
I made some amends following your comments.

raysan5:
@chemaguerra thanks for the review!

PR #3108, opened 2023-06-13
Update raylib.pc.in to fix colors on macs running raylib.

Philo2010:
Raylib is using c++11 colors, by default macOS does not use c++11 colors, so to fix this we have to put a flag to specify.

For example, this will not work for now :
DrawRectangle(X, Y, size, size, YELLOW);

As "YELLOW" is using c++11 colors. So for this to work we have to fix the raylib.pc to include "-std=c++11".


raysan5:
@Philo2010 raylib is a C library, not C++, and the compound literals used for colors are already properly considered in `raylib.h`, afaik.

PR #3113, opened 2023-06-15
Log the magic bytes if an image fails to load

Peter0x44:
Often, on the raylib discord server, people try to load things like jpg renamed to png, which obviously doesn't work. Adding this log will allow easier identification of when this is happening.


JeffM2501:
Would it not be better to check if the extension is PNG and the magic bytes are not correct and tell the user the file is not a PNG? That would be a more direct error messages for people who are not well versed in the data format for images.

Peter0x44:
LoadImage handles more than just PNG, so I'd have to get a list of magic bytes for all file formats to compare, which is more effort

JeffM2501:
you can show what you show now for non PNG use cases. I think this PNG thing is common now, so putting a little effort into making a better message for that one case will provide a lot of help. It does not have to be perfect for every file type, but doing something for common cases would be good.

raysan5:
I'm afraid I'm not adding this trace, users should be owners of the external data provided to the library, also, logging it would imply logging all loaded files for consistency.

PR #3115, opened 2023-06-17
Add options to zig compile

dantecatalfamo:
Support for compiling with `raygui`, `raymath`, and `physac`.
Also outputs the required headers.

`addRaylib` gains an additional positional argument that takes an Options struct.

Raygui should be located `../raygui` relative to the repo root
Physac should be located `../physac` relative to the repo root

These locations match the options in the Makefile

One step generates `src/raygui.c`, `src/raymath.c`, or `src/physac.c` with the `[lib]_IMPLEMENTATION` macro defined depending on which are enabled

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,11 +1,5 @@
 const std = @import("std");
 
-const Options = struct {
-    raygui: bool = false,
-    raymath: bool = false,
-    physac: bool = false,
-};
-
 // This has been tested to work with zig master branch as of commit 87de821 or May 14 2023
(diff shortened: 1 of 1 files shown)


raysan5:
@dantecatalfamo `physac` shouldn't be included, the library is mostly abbandoned and has issues.

dantecatalfamo:
Okay, I've removed physac. 

> EDIT: Also note that `raymath` is already compiled into raylib, adding it could generate issues.
Oh I didn't see it was pulled in by `rcore`, I've removed the generated file :slightly_smiling_face: 

I set it to always install `rlgl.h` and `raymath.h` because the Makefile also does that.

raysan5:
@dantecatalfamo Thanks for the improvement!

PR #3117, opened 2023-06-17
Update CMakeLists.txt for ninja

kaissouDev:
# Ninja Build System Error

when I tried to compile my project with cmake 
and ninja I got this error : 

image

so I deleted the line of code, and now it works


kaissouDev:
I found the solution! in fact my files contained special characters that's why, sorry for the inconvenience ...

PR #3124, opened 2023-06-20
Allow building raylib directly into zig binaries

dantecatalfamo:
Instead of returning a compile step for a static library from `addRaylib`, we make it add the required C source files and their compilation flags onto an existing compile step.

This lets us add raylib directly to any zig project by just calling `addRaylib(b, exe)` on our zig compilations within other build files.

I've also defined the `-DPLATFORM_DESKTOP` flag only for the C compilation units directly related to raylib, so if we're compiling into another zig project that includes other C source files, they don't see any raylib-related flags.

For example, I'm working on a project that uses raylib to create a graph. Now all I have to do to compile my project with raylib is this.

```zig
const raylib = @import("deps/raylib/src/build.zig");
/// [...]
        .name = "zig-visual",
        .root_source_file = .{ .path = "src/main.zig" },
        .target = target,
        .optimize = optimize,
    });
raylib.addRaylib(b, exe, .{ .raygui = true });
/// [...]
```

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,7 +1,7 @@
 const std = @import("std");
 
 // This has been tested to work with zig master branch as of commit 87de821 or May 14 2023
-pub fn addRaylib(b: *std.Build, target: std.zig.CrossTarget, optimize: std.builtin.OptimizeMode, options: Options) *std.Build.CompileStep {
+pub fn addRaylib(b: *std.Build, compile: *std.Build.Step.Compile , options: Options) void {
     const raylib_flags = &[_][]const u8{
         "-std=gnu99",
         "-D_GNU_SOURCE",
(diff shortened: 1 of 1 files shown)


raysan5:
@dantecatalfamo This seems to be a big change on the raylib compilation. Does it fit all Zig users needs?

dantecatalfamo:
@raysan5 It should make it much easier to integrate raylib into any zig project. It works well for me but I would be happy to see other zig developers comment on this

dantecatalfamo:
I should also add the include path of the raylib src directory to the executable passed in. I'm not home right now but I'll add that later

Not-Nik:
I don't really see the point in this. Linking with raylib before was one line more than it is with this PR, so this doesn't really change much in that area and building a standalone version of raylib would be made unnecessarily complicated. Also setting options for raylib seperately is no longer possible, so you can't mix an optimized raylib build with a debuggable executable. I'd rather have more fine control over the raylib part of my build than have one less line in my build script.

DaJobat:
I'm with @Not-Nik, this actually is less useful than the current way of building raylib. It'd be fine as an option, but as a replacement it's a bit of a downgrade.

dantecatalfamo:
I didn't realize you could link compile steps together using `linkLibrary`, sorry for the confusion!

PR #3138, opened 2023-06-27
Update BINDINGS.md

Its-Kenta:
Change C3 to Zlib
Update Raylib-nelua's name as Raylib.nelua to reflect repo change. 
Update Raylib.nelua licence to Zlib


raysan5:
@Its-Kenta Thanks for the update! :D

PR #3144, opened 2023-06-29
Added constructors to Vector2, Vector3 and Vector4.

GameJarne:
This makes it easier to create a Vector variable, without having to separately set the values.
`Vector2 position = Vector2(53, 32);`
This also makes it easier to pass a Vector in a function.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -204,6 +204,13 @@
 typedef struct Vector2 {
     float x;                // Vector x component
     float y;                // Vector y component
+
+    Vector2() {}
+    Vector2(float x, float y)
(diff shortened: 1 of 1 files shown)


jamesl-github:
I don't know if they want C++ in raylib. It says its a C99 library.  
You can just do this             
Vector2 position = { .x = 53, .y = 32 };          
or even              
Vector2 position = { 53, 32 };

GameJarne:
I didn't realise that it wasn't possible in C99.
I'll close this pull request.
Thank you for letting me know a different way to do it.

PR #3150, opened 2023-07-02
fix vector angle example mode 0 circle segment drawing

Gisteron:
also expanded the makefile to add "others" as an option and to compile those examples by default. requires libglfw3, of course.


raysan5:
@Gisteron so fast! thanks for the improvement!

Issue #3141, opened 2023-06-29
[core] All swipe gestures are being interpreted as `GESTURE_SWIPE_LEFT` on web browsers with Touch Screen.


ubkp:
- [x] I tested it on latest raylib version from master branch.
 - [x] I checked and there was a similar issue [#2418](https://github.com/raysan5/raylib/issues/2418) already reported, but no solution was found at the time.
 - [x] My code has no errors or misuse of raylib.

### Issue description

All swipe gestures are being interpreted as `GESTURE_SWIPE_LEFT` on web browsers with Touch Screen:
- Firefox, on Android;
- Chrome, on Android;
- Firefox with `Responsive Design Mode` and `Touch Simulation` enabled, on Linux;
- Chrome with `Toogle Device Toolbar` enabled, on Linux;

After some investigation, it appears that the `event.position[0]` assigned to `GESTURES.Touch.downPositionA` at [rgestures.h#L286](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L286) is usually normalized. However, for some reason, on the Web (HTML5) build, when _Touch_ or _Touch Emulation_ are used, the value being passed to `GESTURES.Touch.downPositionA` is not normalized, therefore causing the `GESTURES.Drag.angle` calculation at [rgestures.h#L310](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L310) to get a wrong result.

The swipe gestures are correctly interpreted on Chrome or Firefox on Linux without _Touch Emulation_.

Probably not the best solution, but a workaround would be adding the following at [rgestures.h#L309](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L309):
```
#if defined( PLATFORM_WEB )
    if (GESTURES.Touch.downPositionA.x > 1.0f || GESTURES.Touch.downPositionA.y > 1.0f)
    {
        GESTURES.Touch.downPositionA.x = Normalize(GESTURES.Touch.downPositionA.x, 0.0f, GetScreenWidth());
        GESTURES.Touch.downPositionA.y = Normalize(GESTURES.Touch.downPositionA.y, 0.0f, GetScreenHeight());
    }
#endif
```
If there isn't any better solution, I can try to send a PR.

### Environment

Platform: _Linux (x86_64)_
Operating System: _Linux Mint 21.1_
OpenGL version: _3.1 Mesa 22.0.5_
GPU: _Intel HD Graphics 3000_

Platform: _Android_
Operating System: _Android 5.1.1_
[...]

PR #3151, opened 2023-07-03
Fix swipe gestures for web

ubkp:
Adds normalization to `gestureEvent.position[i]` that was missing for `PLATFORM_WEB`.
Fixes https://github.com/raysan5/raylib/issues/3141.


raysan5:
@ubkp thanks for the review!

Issue #1963, opened 2021-08-28
[gestures] Pinch In and Pinch Out are messy


SAOMDVN:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

[Input Gestures Example](https://github.com/raysan5/raylib/blob/master/examples/core/core_input_gestures.c) failed to recognise Pinch in and Pinch out, in some cases one appear instead of the expected other results, in other cases both are logged even when only one action was performed

### Environment

Happened on HTML5, Safari, iPhone X, iOS 14.4
OpenGL ES 2.0

raysan5:
I don't have the required hardware/environment to solve this issue. Feel free to send me the required hardware. Closing for now.

raysan5:
@ubkp thank you very much for taking care of this issue!

SAOMDVN:
@ubkp Thank you a lot!

PR #3153, opened 2023-07-04
Fix pinch gestures for web

ubkp:
Adds `previousPositionA` and `previousPositionB` to the `rgestures.h`'s `Touch` struct to keep track of the previous Pinch positions. Didn't want to try to reuse `downPositionA` and `downPositionB` for this case to not break the other gestures that use them and are working properly.

Assign the values of `previousPositionA` and `previousPositionB` during the `event.touchAction == TOUCH_ACTION_DOWN` instead, so they can exist before the `event.touchAction == TOUCH_ACTION_MOVE`.

Removes the now no longer used `GESTURES.Touch.downPositionA` and `GESTURES.Touch.downPositionB` assigns inside the `event.touchAction == TOUCH_ACTION_MOVE`.

Changes the `MINIMUM_PINCH` check to now use the `previousPositionA` and `previousPositionB`.

Changes the criteria to define if it's `GESTURE_PINCH_IN` or `GESTURE_PINCH_OUT` by checking the distance between `previousPosition` and `moveDownPosition`. If `previousPosition` distance is larger than `moveDownPosition` distance, then it's a `GESTURE_PINCH_IN`, else it's a `GESTURE_PINCH_OUT`.

Fixes https://github.com/raysan5/raylib/issues/1963.

### DIFF
--- a/src/rgestures.h
+++ b/src/rgestures.h
@@ -203,6 +203,8 @@ typedef struct {
         Vector2 downDragPosition;       // Touch drag position
         Vector2 moveDownPositionA;      // First touch down position on move
         Vector2 moveDownPositionB;      // Second touch down position on move
+        Vector2 previousPositionA;      // Previous position A to compare for pinch gestures
+        Vector2 previousPositionB;      // Previous position B to compare for pinch gestures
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp thanks for the review! great improvement! :)

PR #3156, opened 2023-07-05
Change examples_template.c to raylib 4.5

Wytekol:
Why nobody saw that?


raysan5:
@Wytekol Good catch! Updated!

PR #3157, opened 2023-07-06
update bindbd-raylib3 to raylib 4.5

o3o:


raysan5:
@o3o thanks for the update!

PR #3164, opened 2023-07-10
Add raylib-ffi to bindings list

ewpratten:
I have a super lightweight Rust binding for raylib. Forgot to add it to the list, so here it is.


raysan5:
@ewpratten thanks for the addition

PR #3166, opened 2023-07-11
Update latest h-raylib version in bindings

Anut-py:


raysan5:
@Anut-py Nice! Thanks for the update! :)

PR #3169, opened 2023-07-12
Add a new task the issue template about checking the wiki

ubkp:
Would like to suggest adding a new task to the issue template reminding users to check the documentation on the wiki. The information there is excellent and is possibly being overlooked.


raysan5:
@ubkp Very good addition! Thanks!

PR #3170, opened 2023-07-13
fix: cmake option "OPENGL_VERSION" doesn't work

royqh1979:
Cmake option "OPENGL_VERSION" doesn't work.
The suggested graphics (GRAPHICS_API_OPENGL_33 in windows) is always used to build raylib.
```
cmake .. -DOPENGL_VERSION=2.1 
```


raysan5:
@royqh1979 thanks for the review!

PR #3172, opened 2023-07-14
[example] Core Input Gestures for Web

ubkp:
Adds an example for core input gestures for Web.

Implemented the following features:

1. Gesture log in a more "log history" style (last gesture on the botton, previous gestures stacking up) with filters to toggle repeated and/or Hold gestures;
2. Geometric style display to track the last gesture performed (Swipe, Tap, Doubletap, Grab, Pinch and Touch count);
3. Protractor to display the last Swipe or Pinch degree angle;
4. Touch and Mouse Pointer points with linked Pinch gestures;
5. Gestures color matched between all features;
6. Screen width dynamically set for Web.

The example was optimized for Web/HTML5 on smartphones, but should run fine on other platforms. Touch specific gestures (e.g.: Pinch) still require Touch capable hardware or Touch Emulation, but the regular Mouse Pointer works for everything else. Tested native on Linux, Web/HTML5 on Linux and Web/HTML5 on Android.


raysan5:
@ubkp Very nice! Thanks for all the improvements! Definitely, returning seconds is more standard than ms.

Issue #3182, opened 2023-07-19
[core] `GESTURE_DRAG` and `GESTURE_SWIPE_*` issues (mostly) for web.


ubkp:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

After becoming aware that _Mouse Position_ was zero for mouse movement on Touch for web (https://github.com/raysan5/raylib/issues/3174), I suspected that it could be the cause of some problems on `GESTURE_DRAG` (and, to an extent, to `GESTURE_SWIPE_*`).

After a closer look, `GESTURE_DRAG` detection [(rgestures.h#L350-L355)](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L350-L355) on `ProcessGestureEvent()` [(rgestures.h#L265)](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L265) calculates the distance between `GESTURES.Touch.downPositionA` and `GESTURES.Touch.moveDownPositionA` to compare it with the `MINIMUM_DRAG` distance.

However, `GESTURES.Touch.moveDownPositionA` is always zero. So when it calculates the distance, it's calculating the distance from the `GESTURES.Touch.downPositionA` (which is wherever the touch was pressed down) to the `0,0` (top left) point on the screen.

This wasn't easily noticeable because `MINIMUM_DRAG` value was really small (`0.015f`) [(rgestures.h#L182)](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L182). But once that was changed to a higher value, it becomes very apparent. A `GESTURE_DRAG` that started near `0,0` would start on the correct distance from `MINIMUM_DRAG`, but a `GESTURE_DRAG` that started on the other corner (bottom right) would trigger instantly.

### Proposed solution

Since `GESTURES.Touch.downPositionA` is not available, it's not possible to reliably calculate the distance and an alternative check to detect `GESTURE_DRAG` is necessary. The best option I could find was using using time (subtracting the `GESTURES.Touch.eventTime` from `rgGetCurrentTime()` and comparing it to a new minimal `DRAG_TIMEOUT` define).

### Issue description

Albeit that solved the `GESTURE_DRAG`, it broke `GESTURE_SWIPE_*` for web.

[...]

PR #3183, opened 2023-07-19
Fix GESTURE_DRAG and GESTURE_SWIPE_* issues (mostly) for web

ubkp:
Adds a different check to detect `GESTURE_DRAG` for `PLATFORM_WEB`.

Adds a `DRAG_TIMEOUT` define for the previous check.

Adds `GESTURE_HOLD` to the check to update the `GESTURES.Touch.upPosition`.

Changes the default value of `FORCE_TO_SWIPE` to make it less twitchy for high sensitivity web browsers like Chrome for Android.

Fix https://github.com/raysan5/raylib/issues/3182.

### DIFF
--- a/src/rgestures.h
+++ b/src/rgestures.h
@@ -178,8 +178,9 @@ float GetGesturePinchAngle(void);                       // Get gesture pinch ang
 //----------------------------------------------------------------------------------
 // Defines and Macros
 //----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp Thank you very much for all the amazing work put into improving the gestures module! Actually the review was int the list of future improvements for the library and you mostly did it all by yourself! Thanks!

One detail, trying the example on Windows (with mouse), I notice that HOLD is always issued after just a TAP, shouldn't it take some time to be detected?

PR #3185, opened 2023-07-19
Optimization of ImageDrawRectangleRec()

smalltimewizard:
A significant performance increase can be had by copying the first row to all other rows.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -3135,26 +3135,28 @@ void ImageDrawRectangleRec(Image *dst, Rectangle rec, Color color)
     if (rec.height < 0) rec.height = 0;
 
     int sy = (int)rec.y;
-    int ey = sy + (int)rec.height;
-
     int sx = (int)rec.x;
 
     int bytesPerPixel = GetPixelDataSize(1, 1, dst->format);
 
-    for (int y = sy; y < ey; y++)
+    // Fill in the first pixel of the first row based on image format
+    ImageDrawPixel(dst, sx, sy, color);
+
+    int bytesOffset = ((sy*dst->width) + sx)*bytesPerPixel;
+    unsigned char *pSrcPixel = (unsigned char *)dst->data + bytesOffset;
+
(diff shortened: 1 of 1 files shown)


raysan5:
@smalltimewizard did you review the performance gain? Please, could you provide more info, and example or some metrics?

smalltimewizard:
The following code performs a barebones test.
`// gcc main.c -o bench.exe -I "./include" -L "./lib" -lraylib -lgdi32 -lwinmm

#include <stdio.h>
#include <time.h>

#include "raylib.h"

#define GRIDW 64
#define GRIDH 64
#define CELLW 24
#define CELLH 24
#define RUNS 10000

int main(int argc, char* argv[]) {
	SetTraceLogLevel(LOG_WARNING);
	
	Image benchimage = {0};
	benchimage = GenImageColor(GRIDW * CELLW, GRIDH * CELLH, (Color){0,0,0,0}); // Set transparent image of the proper size
	ImageFormat(&benchimage, PIXELFORMAT_UNCOMPRESSED_R8G8B8A8);
	
	double timerstart = (double) clock();
	
	int runs = 0;
	while (runs < RUNS) {
		for (int i = 0; i < GRIDW * GRIDH; i++) {
			ImageDrawRectangleRec(&benchimage, (Rectangle){i % GRIDW, i / GRIDW, CELLW, CELLH}, (Color){GetRandomValue(0,255), GetRandomValue(0,255), GetRandomValue(0,255), 255});
		}
		runs++;
	}
	
	printf("%d calls took %f seconds to run.", GRIDW * GRIDH * RUNS, (double) (clock() - timerstart)/CLOCKS_PER_SEC);
	
	return 0;
};`

Without the change (libraylib.a from raylib-4.5.0_win64_mingw-w64), this takes 92 seconds to run on my machine.
With the change (libraylib.a compiled with mingw32-make), it takes 14 seconds.

smalltimewizard:
The program where I was trying to use this before/after (55 FPS vs 128 FPS):
image
image

PR #3186, opened 2023-07-19
Document buffer format for audio processors.

AlbertoGP:
There is a comment in the example https://github.com/raysan5/raylib/blob/d6f16b76649944f65491633a75ad648a6ddfdacf/examples/audio/audio_mixed_processor.c#L26 about this, but it is more discoverable in the API docs.


AlbertoGP:
From the stream effects example https://github.com/raysan5/raylib/blob/d6f16b76649944f65491633a75ad648a6ddfdacf/examples/audio/audio_stream_effects.c#L154 it seems that this applies to `AttachAudioStreamProcessor()` too, should I add the note also there?

raysan5:
@AlbertoGP thanks for the review!

PR #3187, opened 2023-07-19
Update android ndk path

lesleyrs:
This goes along with my PR to the game template https://github.com/raysan5/raylib-game-template/pull/22

It's the recommended way by google to set up the sdk/ndk on the command line by using sdkmanager to install them in this place. https://developer.android.com/tools/sdkmanager


raysan5:
@lesleyrs This PR assumes an NDK version of `25.2.9519653`, I prefer to let the user provide the desired path externally isntead of force one specific version.

lesleyrs:
That's fine, i'll change it back in the other PR as well

PR #3188, opened 2023-07-20
Add note about sample format to AttachAudioStreamProcessor()

AlbertoGP:
Please excuse the noise, I should have included this in the previous PR https://github.com/raysan5/raylib/pull/3186 instead of asking there whether I should do it. I wasn’t sure whether the buffer passed to `AttachAudioStreamProcessor()` always used the same type as `AttachAudioMixedProcessor()` although it looked like that in the example.

I’ve been reading carefully the example for `AttachAutdioStreamProcessor()` and the source code, and it is now clear to me that this function too uses `float` for the samples in `buffer`.


raysan5:
@AlbertoGP thanks for the addition

PR #3195, opened 2023-07-24
Added TextToFloat()

Z0RIK:
I recently created a pull request for the raygui library (#309), which adds GuiFloatBox() that requires TextToFloat() to function properly


raysan5:
@Z0RIK I'm afraid this function is way more complex than it seems and that was the reason it is not implemented. Depending on the locale of the system, a float can be represented in many possible ways:
```
3.141592
3,141592
3'141592
...
```
It is not simple to consider all case and it also depends on the system locale.

So, I prefer to avoid adding this function.

orcmid:
A simplification would be 

```C
#include <stddef.h>
 ...
float TextToFloat(char* s)
        {     return (float) atof(s);
              }
```

There appears to be no locale consideration for `atof()`.

And yes, this could be left to the user although that is likely to be error-prone, inaccurate, or all of the above.

Z0RIK:
Thanks for helpfull comments. To be honest i didn't consider that this function can be used outside of GuiFloatBox() that i was trying to implement, so i just forced text value in that function to use '.' even if user tries to press ','.

Issue #3197, opened 2023-07-25
[rmodel] Strange behavior when rendering flipped billboards using `DrawBillboardPro`


bohonghuang:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

Hello! I found that when using `DrawBillboardPro` to render, if the height/width of the `source` is negative, the width/height of the `size` also needs to be changed to negative, otherwise the billboard will not be rendered. This may be detrimental to the API's usability and consistency. For example, when using `DrawBillboardPro` to render `RenderTexture` and flipping it upside down, both the height of `source` and the width of `size` need to be changed to negative, which is inconsistent with the behavior of `DrawTexturePro`. I believe the expected behavior should be able to use the `source` and `size` parameters independently to control the flipping of the billboard without interfering with each other. Any help you can provide would be greatly appreciated!

### Code Example

Simply change the `source` variable in the billboard example of `raylib` to:

```c
Rectangle source = { 0.0f, (float)bill.height, (float)bill.width, (float)-bill.height };
```

raysan5:
@bohonghuang It seems like a bug, feel free to send a PR to address it.

PR #3199, opened 2023-07-25
Fix `DrawBillboardPro` to allow `source` of negative size (#3197)

bohonghuang:


raysan5:
@bohonghuang It seems that PR does not allow negative size, if the expected behaviou is the same as `DrawTexturePro()`, this is not the correct solution.

bohonghuang:
Sorry, perhaps my solution is not the implementation you expected, but this solution has been tested and it does work well in my use case.

raysan5:
@bohonghuang As per the open issue I understood the expected behaviour was the same as `DrawTexturePro()`.

 >  I believe the expected behavior should be able to use the source and size parameters independently to control the flipping of the billboard without interfering with each other.

And actually, that makes sense to make the behaviour consistent, that's what users expect.

PR #3203, opened 2023-07-26
Re-fix `DrawBillboardPro` to allow `source` of negative size (#3197)

bohonghuang:
The root cause of issue #3197 seems to be that having only one dimension with a negative value in `source` or `size` causes the vertex order of the billboard to become clockwise, which cannot be rendered correctly when backface culling is enabled. This PR should allow `source` and `size` with any negative dimensions to be rendered correctly, and flipped as expected.


raysan5:
@bohonghuang Nice! Thank you very much for this improvement! :)

bohonghuang:
My pleasure :)

PR #3204, opened 2023-07-26
[textures] Fix LoadTextureCubemap for manual layouts

Not-Nik:
Passing any other layout than `CUBEMAP_LAYOUT_AUTO_DETECT` to `LoadTextureCubemap` loads an empty texture with 0 by 0 dimensions. This is because the size of each face is caculated like so
```c
int size = cubemap.width;
```
but `cubemap` is zero initialized and only set when auto detecting the layout.

This PR fixes this issue, by manually setting the width for each layout.


raysan5:
@Not-Nik thanks for the review! Actually I think that function had some other issues... and an example was required! I think it was not properly tested...

Not-Nik:
There is [an example](https://github.com/raysan5/raylib/blob/master/examples/models/models_skybox.c) for it, and from what I've seen it does work for the layouts it actually has implementations for.

It only isn't implemented for `CUBEMAP_LAYOUT_PANORAMA`, but the aforementioned example does have code that loads those.

raysan5:
@Not-Nik Wow! Really? I didn't remember about it! 😄

PR #3206, opened 2023-07-27
Add GetWindowSize() function

ghost:
Adds a `GetWindowSize()` function that returns the current window size to a `Vector2`.

This is helpful in case the window size was changed by the user (or the OS, WM, etc). And will be necessary for fixing the `SetWindowMonitor()`.


raysan5:
@ubkp Actually, it has been proposed several times, many users look for `GetWindow*()` and ask me why `GetScreen*()`. Despite it could be less intuitive, I tried to avoid `GetWindow{Width/Height}()` because some platforms supported by raylib do not use the concept of window, screen framebuffer is "directly" output to display, usually a screen fit on a window (when available) but even on then it could not be the case.

About why `GetScreen{Width/Height}()` instead of size, two reasons:

 1. Usually the pixel size returned is an integer value, `Vector2` only support floats and I preferred to avoid adding a `Point` structure using ints (despite it could be needed on some circunstances).
 2. Usage reason, I noticed most of the time it's more useful to have the width and height in separate functions to be queried and used separately. Same happens with some other functions so I tried to be consistent.

PR #3211, opened 2023-07-29
Fix android soname in src/Makefile

ndytts:
Since **API_VERSION** not exist   libraylib.$(**API_VERSION**).so  is equal to  libraylib..so


raysan5:
@ndytts Good catch! Thanks for the fix! :)

PR #3212, opened 2023-07-30
Update BINDINGS.md to include rayjs

mode777:
Hi everyone, I made a standalone Javascript interpreter based on [QuickJS](https://bellard.org/quickjs/) with bindings for [Raylib](https://www.raylib.com/) and would like to add it to the list.


raysan5:
@mode777 Nice! Great addition! 👍😄

PR #3217, opened 2023-08-02
Update BINDINGS.md

joseph-montanez:
Updated Raylib-PHP to support Raylib 4.5


RobLoach (approved):
You were so preoccupied with whether you could, you didn’t stop to think if you should :upside_down_face: 

Nice work!

raysan5:
@joseph-montanez Nice! Thanks for the update! 😄

PR #3218, opened 2023-08-02
integer to float warnings in rcore

JeffM2501:
This PR fixes two float to int casting warnings in rcore. All the arguments are ints, so the divide should be ints too.


raysan5:

PR #3219, opened 2023-08-02
[AUDIO] Add a function to make an alias of a sound and share it's sample data

JeffM2501:
This PR adds a function that makes an alias from an existing sound. The new sound has a pointer to the sound buffer used by the donor sound and does not own the sample buffer memory.

Since PlaySoundMulti was removed, this feature makes it easy to make shallow copies of sounds that can be played multiple times with out the need to load multiple copies of the sample data. Since each sound is still it's own audio buffer, the sound instances are controllable with the existing sound API.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -912,7 +912,6 @@ Sound LoadSoundFromWave(Wave wave)
         sound.stream.sampleSize = 32;
         sound.stream.channels = AUDIO_DEVICE_CHANNELS;
         sound.stream.buffer = audioBuffer;
-        sound.noFree = false;
     }
 
     return sound;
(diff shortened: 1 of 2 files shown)


src/raudio.c
```diff
@@ -420,7 +420,7 @@ static bool SaveFileText(const char *fileName, char *text);         // Save text
 // NOTE: Those functions are not exposed by raylib... for the moment
 //----------------------------------------------------------------------------------
 AudioBuffer *LoadAudioBuffer(ma_format format, ma_uint32 channels, ma_uint32 sampleRate, ma_uint32 sizeInFrames, int usage);
-void UnloadAudioBuffer(AudioBuffer *buffer);
+void UnloadAudioBuffer(AudioBuffer *buffer, bool freeSampleData);
```
raysan5, line 423: I'd prefer to avoid the extra parameter in the `UnloadAudioBuffer()`. As a custom `UnloadSoundAlias()` is provided, it could implement the custom unloading directly inside it.

JeffM2501: Understandable. PR updated.

raysan5:
@JeffM2501 thanks for addition and the review! 👍🙂

PR #3220, opened 2023-08-02
Support 16-Bit HDR textures

Not-Nik:
Makes creation of HDR textures with 16-Bit half floats possible. This needs some minor compatibility changes because in OpenGL ES half float textures are in a separate extension, but it won't change platform support in any major way (the two even have a [single entry in Khronos' extension registry](https://registry.khronos.org/OpenGL/extensions/OES/OES_texture_float.txt)).

The image manipulation functions are a bit more complicated, because there is no native half float support in C so I've used two new util functions `HalfToFloat` and `FloatToHalf` there.

This doesn't change the API apart from 3 (6 if you count rlgl) new enum variants.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -488,26 +488,3 @@ static int android_close(void *cookie)
     return 0;
 }
 #endif  // PLATFORM_ANDROID
-
-// From https://stackoverflow.com/questions/1659440/32-bit-to-16-bit-floating-point-conversion/60047308#60047308
-
-unsigned int AsUnsignedInt(const float x) {
-    return *(unsigned int*)&x;
-}
-float AsFloat(const unsigned int x) {
-    return *(float*)&x;
-}
-
-float HalfToFloat(unsigned short x) {
-    const unsigned int e = (x&0x7C00)>>10; // exponent
-    const unsigned int m = (x&0x03FF)<<13; // mantissa
-    const unsigned int v = AsUnsignedInt((float)m)>>23; // evil log2 bit hack to count leading zeros in denormalized format
-    return AsFloat((x&0x8000)<<16 | (e!=0)*((e+112)<<23|m) | ((e==0)&(m!=0))*((v-37)<<23|((m<<(150-v))&0x007FE000))); // sign : normalized : denormalized
-}
-
-unsigned short FloatToHalf(float x) {
-    const unsigned int b = AsUnsignedInt(x)+0x00001000; // round-to-nearest-even: add last bit after truncated mantissa
-    const unsigned int e = (b&0x7F800000)>>23; // exponent
-    const unsigned int m = b&0x007FFFFF; // mantissa; in line below: 0x007FF000 = 0x00800000-0x00001000 = decimal indicator flag - initial rounding
-    return (b&0x80000000)>>16 | (e>112)*((((e-112)<<10)&0x7C00)|m>>13) | ((e<113)&(e>101))*((((0x007FF000+m)>>(125-e))+1)>>1) | (e>143)*0x7FFF; // sign : normalized : denormalized : saturate
-}
--- a/src/utils.c
+++ b/src/utils.c
@@ -488,3 +488,26 @@ static int android_close(void *cookie)
     return 0;
 }
 #endif  // PLATFORM_ANDROID
+
+// From https://stackoverflow.com/questions/1659440/32-bit-to-16-bit-floating-point-conversion/60047308#60047308
+
+unsigned int AsUnsignedInt(const float x) {
+    return *(unsigned int*)&x;
+}
+float AsFloat(const unsigned int x) {
+    return *(float*)&x;
+}
+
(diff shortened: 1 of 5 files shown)


src/utils.c
```diff
@@ -488,3 +488,26 @@ static int android_close(void *cookie)
     return 0;
 }
 #endif  // PLATFORM_ANDROID
+
+// From https://stackoverflow.com/questions/1659440/32-bit-to-16-bit-floating-point-conversion/60047308#60047308
+
+unsigned int AsUnsignedInt(const float x) {
```
raysan5, line 494: Considering this function is only used 1-2 times and it's really minimal, I'd prefer to implement it directly in the required function.

src/utils.c
```diff
@@ -488,3 +488,26 @@ static int android_close(void *cookie)
...
+    return *(unsigned int*)&x;
+}
+float AsFloat(const unsigned int x) {
```
raysan5, line 497: Considering this function is only used 1-2 times and it's really minimal, I'd prefer to implement it directly in the required function.

src/utils.c
```diff
@@ -488,3 +488,26 @@ static int android_close(void *cookie)
...
+    return *(float*)&x;
+}
+
+float HalfToFloat(unsigned short x) {
```
raysan5, line 501: Considering this function is only used on `rtextures` module, I'd prefer to move it there as a internal `static` function, it will also remove the additional coupling to `utils` module.

src/utils.c
```diff
@@ -488,3 +488,26 @@ static int android_close(void *cookie)
...
+    const unsigned int e = (x&0x7C00)>>10; // exponent
+    const unsigned int m = (x&0x03FF)<<13; // mantissa
+    const unsigned int v = AsUnsignedInt((float)m)>>23; // evil log2 bit hack to count leading zeros in denormalized format
+    return AsFloat((x&0x8000)<<16 | (e!=0)*((e+112)<<23|m) | ((e==0)&(m!=0))*((v-37)<<23|((m<<(150-v))&0x007FE000))); // sign : normalized : denormalized
+}
+
+unsigned short FloatToHalf(float x) {
```
raysan5, line 508: Considering this function is only used on `rtextures` module, I'd prefer to move it there as a internal `static` function, it will also remove the additional coupling to `utils` module.

raysan5:
@Not-Nik Thanks for the improvement! Actually it was on the raylib future plans but it was discarded... I can reconsider it, it looks good. Just added a couple of review comments.

Not-Nik:
> Just added a couple of review comments.
Updated accordingly

raysan5:
@Not-Nik Thank you very much! This is a really nice addition! 😄

asdqwe:
@Not-Nik @raysan5 Sorry for raising the PR again, but while compiling raylib with `-02` (or `-03`) `HalfToFloat` and `FloatToHalf` will raise `dereferencing type-punned pointer will break strict-aliasing rules`, which, AFAIK, incurs in undefined behavior, so probably worth looking into. Not sure how to address these ones tho.

<details><summary>Compiling log:</summary>

``` bash
[...]
gcc -c rtextures.c -Wall -D_GNU_SOURCE -DPLATFORM_DESKTOP_GLFW -DGRAPHICS_API_OPENGL_33 -Wno-missing-braces -Werror=pointer-arith -fno-strict-aliasing -std=c99 -fPIC -O2 -Werror=implicit-function-declaration -D_GLFW_X11  -I.  -Iexternal/glfw/include
rtextures.c: In function ‘HalfToFloat’:
rtextures.c:5396:30: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 5396 |     const unsigned int v = (*(unsigned int *)&fm) >> 23; // Evil log2 bit hack to count leading zeros in denormalized format
      |                              ^~~~~~~~~~~~~~~~~~~
rtextures.c:5399:15: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 5399 |     result = *(float *)&r;
      |               ^~~~~~~~~~~
rtextures.c: In function ‘FloatToHalf’:
rtextures.c:5409:30: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
 5409 |     const unsigned int b = (*(unsigned int *) & x) + 0x00001000; // Round-to-nearest-even: add last bit after truncated mantissa
      |                              ^~~~~~~~~~~~~~~~~~~~
rtextures.c: In function ‘HalfToFloat’:
rtextures.c:5396:29: warning: ‘fm’ is used uninitialized [-Wuninitialized]
 5396 |     const unsigned int v = (*(unsigned int *)&fm) >> 23; // Evil log2 bit hack to count leading zeros in denormalized format
      |                            ~^~~~~~~~~~~~~~~~~~~~~
rtextures.c:5395:17: note: ‘fm’ declared here
 5395 |     const float fm = (float)m;
      |                 ^~
[...]
```
</details>

PR #3222, opened 2023-08-03
[CORE] Support for SetWindowTitle and InitWindow for web

VitusVeit:
### Changes:
`SetWindowTitle` and `InitWindow` will now change the current tab name of the browser, if compiled for the web platform.
They use the emscripten function `emscripten_set_window_title`.

### Why:
As stated in the header file, the `SetWindowTitle` function works only for desktop.
But this feature can be useful also on the web, for example by setting the tab title to the one of your game/app to make it more recognizable without having to change the title manually in the HTML file or interacting directly with emscripten.

### Status:
Tested on Windows 11 with the latest version of emscripten on chrome with raylib's basic window example with no errors nor code changes.


raysan5:
@VitusVeit Nice improvement! Thank you very much!

PR #3223, opened 2023-08-04
[AUDIO] Add an example of how to use LoadSoundAlias

JeffM2501:
This PR adds an example that shows how to play the same sound multiple times, replicating the functionality of the old PlaySoundMulti.


raysan5:
@JeffM2501 thank you very much for the addition!

Issue #3179, opened 2023-07-18
[rcore] Issue with texture with transparency when using open gl es2.0 (google angle) on macos


ypujante:
This is a follow up to my post on [discord](https://discord.com/channels/426912293134270465/706078465476001902/1130558894402445603)

The issue on macOS is that when using OpenGL ES 2.0 (implemented by Google Angle on macOS), then the texture is rendered with a "halo" where there is transparency:

<img width="96" alt="tape_opengl_es20" src="https://github.com/raysan5/raylib/assets/357116/624e6a48-b89c-4302-b7cc-97d947620325">

```text
INFO: GLAD: OpenGL ES2.0 loaded successfully
INFO: GL: Supported extensions count: 101
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Google Inc. (Apple)
INFO:     > Renderer: ANGLE (Apple, Apple M2 Max, OpenGL 4.1 Metal - 83.1)
INFO:     > Version:  OpenGL ES 3.0.0 (ANGLE 2.1.20982 git hash: f3e3810b917c)
INFO:     > GLSL:     OpenGL ES GLSL ES 3.00 (ANGLE 2.1.20982 git hash: f3e3810b917c)
```

When using the macOS (deprecated) built-in OpenGL there is no such issue:

<img width="85" alt="tape_opengl" src="https://github.com/raysan5/raylib/assets/357116/7c6a0fc4-035b-4055-9a7c-4bb4ef1d3c0f">

```text
INFO: GLAD: OpenGL extensions loaded successfully
INFO: GL: Supported extensions count: 43
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Apple
INFO:     > Renderer: Apple M2 Max
INFO:     > Version:  4.1 Metal - 83.1
INFO:     > GLSL:     4.10
```

The code is very simple and does not do anything fancy:

```cpp
#define IMAGE_PNG_PATH "/absolute/path/to/image.png"

int main()
{
  InitWindow(800, 450, "raylib [core] example - basic window");

  auto tape = LoadTexture(IMAGE_PNG_PATH);
  //  SetTargetFPS(60);

  constexpr Color myGrayColor{153, 153, 153, 255};

  auto bgColor = myGrayColor;

  while (!WindowShouldClose())
  {
    if(IsKeyPressed(KEY_B))
      bgColor = BLACK;
    if(IsKeyPressed(KEY_R))
      bgColor = RED;
[...]

PR #3224, opened 2023-08-06
Fix compositing mixing with framebuffer for web and ANGLE

ghost:
Adds a `#if defined(GRAPHICS_API_OPENGL_ES2)` to clear the alpha channel on the end of the frame at the end of `rlDrawRenderBatch()` ([R2894-R2900](https://github.com/raysan5/raylib/pull/3224/files#diff-30abe7ae1669d72f0c978b3e2b30985524f8a9b0d90a19c7bee81134c954c213R2894-R2900)). It does that by:
1. Setting the `clearColor` alpha to `1`;
2. Setting a `colorMask` to only consider the alpha from that `clearColor`;
3. Clearing it;
4. Restoring the `colorMask` to consider all values.

Removes `background-color: black;` from `minshell.html` ([L35](https://github.com/raysan5/raylib/pull/3224/files#diff-94e6fb974bc8fc76ab1f1e3e7c828b7f302bfa249cab0f6493c10d0539770633L35)) and `background: black;` from `shell.html` ([L78](https://github.com/raysan5/raylib/pull/3224/files#diff-3dfaf4a1998bb76cd59e6121ae71c10987c32fc478918aa92c63bdb1d61d1c06L78)), since with the previous changes they are no longer needed. Those can still be set, but `ClearBackground()` will take precedence in defining the background color.

Tested these changes successfully on Firefox (102.11.0esr 64-bit) and Chrome (115.0.5790.110) on Linux (Linux Mint 21.1 x86_64).

@ypujante Could you please test/confirm if these changes fix the issue on ANGLE?

@raysan5 I tried my best to place that where I think is the end of the frame. If these proposed changes are accepted, could you please check if that's the ideal/correct place for them?

Fixes https://github.com/raysan5/raylib/issues/3179.

**Edit:** added line marks.

### DIFF
--- a/src/minshell.html
+++ b/src/minshell.html
@@ -32,7 +32,7 @@
 
     <style>
         body { margin: 0px; }
-        canvas.emscripten { border: 0px none; background-color: black; }
+        canvas.emscripten { border: 0px none; }
     </style>
     <script type='text/javascript' src="https://cdn.jsdelivr.net/gh/eligrey/FileSaver.js/dist/FileSaver.min.js"> </script>
     <script type='text/javascript'>
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -2890,6 +2890,14 @@ void rlDrawRenderBatch(rlRenderBatch *batch)
     // Change to next buffer in the list (in case of multi-buffering)
     batch->currentBuffer++;
     if (batch->currentBuffer >= batch->bufferCount) batch->currentBuffer = 0;
+
+#if defined(GRAPHICS_API_OPENGL_ES2)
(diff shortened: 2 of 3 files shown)


raysan5:
@ubkp I'm not sure about this change, does it consider the case that some applications could require a transparent background (some of my tools require it). Also, it should be probably called at `EndDrawing()` but calling it partially removes framebuffer cleaning control from user and it can derive in other unexpected issues, actually framebuffer cleaning was originally inside `BeginDrawing()` and it was problematic, so `ClearBackground()` was added.

PR #3225, opened 2023-08-06
Tweak build.zig to work with cross-compiling

yujiri8:
I made a similar change a while ago (#3090), but the same failure mode has reappeared for me when cross-compiling. I suspect the recent Zig updates have changed requirements. In particular, I see:
```
zig build-lib raylib Debug x86_64-linux-musl: error: error: unable to find Dynamic system library 'GL' using strategy 'paths_first'. searched paths: none
error: unable to find Dynamic system library 'X11' using strategy 'paths_first'. searched paths: none
```

This patch fixes it for me. I suspect similar changes are necessary for the other targets, but I only added this one because I haven't tested building from other targets.


raysan5:
@yujiri8 thanks for the review!

PR #3229, opened 2023-08-06
Consolidate all build.zig files into the root.

Manuzor:
I was wondering why there were multiple `build.zig` files in this repo and was unable to find issues or commits that would explain this. Out of curiosity, I tried merging them all into the `build.zig` file in the root of the repo. This worked well so I went ahead and added some changes that I thought were worth doing.

If there was a specific reason to keep `src/build.zig` and `examples/build.zig` separate, please share. I'd be happy to re-do my changes to the split versions if desired. Although I would think keeping only a single `build.zig` file would be simpler and less surprising to people.

My changes include:

* Add an option `-Dexamples` that will build all examples along with raylib.
    * Off by default, so when you do `zig build` you only get raylib as a statically linked library.
* Examples now link directly to the `std.Build.Step` produced by `addRaylib`.
    * The old version picked out .lib/.a files from the `zig-out` folder manually, branched per platform. This is what originally drove me to try out and improve this.
* Change the `-Draygui` option from `bool` to `[]const u8` so that people can point to `raygui.h` instead of having to locate it in `../raygui/src/raygui.h`. The default location is still supported but commandline usage changes from `zig build -Draygui` to `zig build -Draygui=` (i.e. pass an empty string). So all of the following are valid now:
    * `zig build` -- No raygui support
    * `zig build -Draygui=` -- With raygui support. Looks for `../raygui/src/raygui.h`.
    * `zig build -Draygui=../../path/to/raygui/src/raygui.h` -- With raygui support and a specific source header location.
* The generated file `raygui.c` is now properly integrated into the zig build system by using `std.build.Step.WriteFile` the way it was intended.
* Remove the need for comptime `srcdir`.
    * All relative file paths will always be relative to the `build.zig` file, so it was a matter of replacing `srcdir ++ "/file.c"` with `"src/file.c"`. This hopefully makes it simpler and more idiomatic.
    * _NOTE: the old `examples.zig` file did not use a comptime known `srcdir` but used "../src/file.c" all along._

In general, I tried to keep things simple and add improvements by using helpers defined on `std.Build` instead of reaching for general purpose `std.*` stuff.

I tested a bunch of the examples compiled with `zig build -Dexamples` and they all seemed to work flawlessly. However, I only did this on Windows 11 Pro (22621) because I don't have easy access to other setups right now.

I have also tested using this version with the new zig Package Manager introduced with zig 0.11 and it worked very well. (I only tried `b.anonymousDependency(...)`, and only on Windows).

Unfortunately, due to the merge/move/edits, the diff shows the old files simply as removed, and the root file as gaining a large amount of added LOCs. I'm not sure if I could improve this diff in any way. I tried listing all the things I did modify above.

I hope this is helpful.


Not-Nik:
I like the changes overall, but I feel like keeping separate build scripts for the main library and the examples is more in-line with how raylib is structure otherwise. The CMake build, for example, does this similarly.

Not-Nik:
cc @ryupold

raysan5:
@Manuzor @Not-Nik I agree that having several separated build files is easier to manage than everything in just one. Is it possible to add some of the proposed improvements into the separate build files?

Manuzor:
> I agree that having several separated build files is easier to manage than everything in just one. Is it possible to add some of the proposed improvements into the separate build files?

Some improvements can be added, yea. The main thing is that `examples/build.zig` is not allowed to access `src/build.zig` because it would reach outside of its package path.

```
error: import of file outside package path: '../src/build.zig'
```

We would have to fall back to the [previous implementation](https://github.com/raysan5/raylib/blob/0959f6ebf69f44d5bb0225cc94fd187ebedf8be5/examples/build.zig#L29C1-L35) but the user experience with that implementation is not great because it kind of circumvents the build system.

Right now, it requires users to manually build raylib as a library *before* building the examples. Otherwise they're greeted with `raylib.lib FileNotFound`, depending on the platform. It also doesn't automatically rebuild raylib if you made changes to the library and want to test it with the examples. Here's the workflow if you wanted to do that:

```bash
> cd /path/to/raylib
> vim src/...   # make some edits
> zig build
> cd examples
> zig build
```

Additionally, It allows `examples/build.zig` to become out of sync with `src/build.zig`. For example, the current version of `examples/build.zig` in `master` doesn't compile with zig 0.11.0.

If you would still like to keep the split build files, I would close this PR and make a new one with fixes to the existing files in `master`.

raysan5:
@Manuzor Thanks for the additional information. Still, I don't think it's that problematic for the user to manage the manual building of raylib and examples separately, the current `Makefile` build system works the same way and for my tests on platforms requiring it I just do it manually or create a simple script to automatize it (for cases I really need to recompile raylib-example all the time).

I agree that both `build.zig` should be aligned to current zig 0.11.0.

raysan5:
As commented, I'm not merging this change, I prefer to keep the build systems separate, like we do with Makefile.

PR #3230, opened 2023-08-07
fix: gif recordering doesn't work if num/caps lock is active.

mohad12211:
- current behavior is to check if only the ctrl modifier is active, the result is that gif recording will not work if num or caps lock are active, which seems like a bug to me.
- note: this fix allows other modifiers to be set at the same time (ctrl+shift+f12 will also trigger gif recording).
I don't consider this as a problem, but if you want I can explicitly remove the caps/num lock bits and _after_ that check if `mods` is _equal_ to `GLFW_MOD_CONTROL`

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -5609,7 +5609,7 @@ static void KeyCallback(GLFWwindow *window, int key, int scancode, int action, i
     if ((key == GLFW_KEY_F12) && (action == GLFW_PRESS))
     {
 #if defined(SUPPORT_GIF_RECORDING)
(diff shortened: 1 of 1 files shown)


raysan5:
@mohad12211 Good catch! Actually I didn't considered that possibility... the proposed review looks ok to me. Keep in mind that this functionality was just added for convenience for my students, anyone willing a more solid GIF recording mechanism should implement it manually on user code.

mohad12211:
true! but it's a very good primitive solution for quick recording.

PR #3232, opened 2023-08-08
Remove PLATFORM_RPI

michaelfiber:
This removes PLATFORM_RPI from source files and Makefiles (not all the project files though because some of those are very outdated and it seems better to get them synced up with master rather than trying to apply these changes to them).

Almost all the edits have very little impact as most of the code for PLATFORM_RPI is shared with PLATFORM_DRM. 

I built the examples for PLATFORM_DESKTOP and ran through a bunch to see if there were any side effects but couldn't find any.


QosmosTheQ (approved):

raysan5:
@michaelfiber thank you very much for taking the time to review this, it's a very big change! probably it will require some time of testing and further reviews. Thanks!

PR #3234, opened 2023-08-08
Fix typos in rcore.c comments

RadsammyT:
corrects all instances of:
'acording' > according
'usualy' > usually
'aprtially' > partially
'dettach' > detach


raysan5:
@RadsammyT thank you very much for reviewing all those typos! there are probably some more around the code!

PR #3236, opened 2023-08-09
fix typos in rmodels, rshapes, rtext modules

RadsammyT:
Found around 6 comment typos in rmodels, rshapes, and rtext modules.
I accidentally undid these changes (and more I believe) at one point so there may be something I missed. If thats the case then let me know.

rmodels:

-  'dyamond' > diamond
- 'verted' > vertex
- 'attibute' > attribute
- (_not a typo but im lengthening it just in case, so it doesnt get confused with quad_) 'quat' > quaternion

rshapes:

 - 'missused' > misused

rtext:

 - 'fo' > to


raysan5:
@RadsammyT thank you very much for the fixes! 👍😄

PR #3241, opened 2023-08-11
Ignore unused return value of GetCodepointNext in GetCodepointCount

ashn-dot-dev:
Removes the last warning from non-external libraries when compiling with the default build configuration on x64 Linux.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1766,7 +1766,7 @@ int GetCodepointCount(const char *text)
     while (*ptr != '\0')
     {
         int next = 0;
-        int letter = GetCodepointNext(ptr, &next);
+        (void)GetCodepointNext(ptr, &next);
 
         ptr += next;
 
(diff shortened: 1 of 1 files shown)


src/rtext.c
```diff
@@ -1766,7 +1766,7 @@ int GetCodepointCount(const char *text)
     while (*ptr != '\0')
     {
         int next = 0;
-        int letter = GetCodepointNext(ptr, &next);
+        (void)GetCodepointNext(ptr, &next);
```
raysan5, line 1769: I usually avoid the `(void)` at beginning when not interested in return value.

ashn-dot-dev: Good point. It doesn't even suppress anything in this case. Removed the `void` cast in 814b23b367806ec9a9c2887fa0a1a8e33fa7ff5b.

raysan5:
@ashn-dot-dev thanks for the review!

Issue #2827, opened 2022-12-12
[core] Mod-tap key presses not registered properly (PRESS-RELEASE in single frame)


dmlary:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

Very fast key presses not registering in `IsKeyPressed()`, but do show up in `GetKeyPressed()`.

I've tracked this issue down to using a keyboard with QMK firmware, using mod-tap.  When tapping the spacebar, I get both the KeyDown & KeyUp event in the same frame.  This key pressed event isn't registered when I check `IsKeyPressed()`, but does show up when I call `GetKeyPressed()`.

It exhibits most clearly when trying to play the Pang game on the raylib website: https://www.raylib.com/games/classics/loader.html?name=classic_pang

### Environment

This happens both natively and for the raylib web-based demos.

* Platforms: Web & Desktop
* OS: OSX 11.7
* OpenGL: 
```txt
INFO: GL: OpenGL device information:
INFO:     > Vendor:   ATI Technologies Inc.
INFO:     > Renderer: AMD Radeon Pro 5500M OpenGL Engine
INFO:     > Version:  4.1 ATI-4.6.21
INFO:     > GLSL:     4.10
```

### Issue Screenshot

![Screen Shot 2022-12-11 at 7 12 00 PM](https://user-images.githubusercontent.com/857742/206941184-16800fa0-0e4f-4c43-b6b2-0e95d35dc902.png)

Screenshot of logging messages for code snippet below.

Top part of the log is me hitting space on laptop keyboard, bottom is me hitting space on external keyboard running QMK firmware and mod-tap on spacebar. 

### Code Example

minimal code logging discrepancy between `IsKeyPressed()` and `GetKeyPressed()`:
```c
                PollInputEvents();
                if (IsKeyPressed(KEY_SPACE)) {
                    log_debug("space pressed");
                }
                for (int k = GetKeyPressed(); k != 0; k = GetKeyPressed()) {
                    log_debug("pressed {}", k);
                }
```

raysan5:
@dmlary Yes, that's as expected. `IsKeyPressed()` consumes key polled and polling happens per frame while `GetKeyPressed()` accumulates key presses in an internal queue. At 60 fps it shouldn't be a big issue.

[...]

PR #3245, opened 2023-08-16
Add `IsKeyPressedRepeat` (desktop only)

actondev:
Hi Ray!

Currently `IsKeyPressed` ignores key repeats. This PR keeps the current behavior the same while adding the additional `IsKeyPressedRepeat`.
Another approach could be changing the existing `IsKeyPressed` functionality to take into account key repeats as well, and add `IsKeyPressedExt`, expecting an extra `bool repeat` flag for taking into account key repeats or not.

Personally I would be in favor of the second approach (changing existing behavior), but it's up for discussion.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1115,6 +1115,7 @@ RLAPI unsigned char *DecodeDataBase64(const unsigned char *data, int *outputSize
 
 // Input-related functions: keyboard
 RLAPI bool IsKeyPressed(int key);                             // Check if a key has been pressed once
+RLAPI bool IsKeyPressedRepeat(int key);                       // Check if a key has been pressed again (Only PLATFORM_DESKTOP)
 RLAPI bool IsKeyDown(int key);                                // Check if a key is being pressed
 RLAPI bool IsKeyReleased(int key);                            // Check if a key has been released once
 RLAPI bool IsKeyUp(int key);                                  // Check if a key is NOT being pressed
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -438,6 +438,8 @@ typedef struct CoreData {
             int exitKey;                    // Default exit key
             char currentKeyState[MAX_KEYBOARD_KEYS];        // Registers current frame key state
             char previousKeyState[MAX_KEYBOARD_KEYS];       // Registers previous frame key state
+            // NOTE: Since key press logic involves comparing prev vs cur key state, we need to handle key repeats specially
+            char keyRepeatInFrame[MAX_KEYBOARD_KEYS];       // Registers key repeats for current frame.
 
             int keyPressedQueue[MAX_KEY_PRESSED_QUEUE];     // Input keys queue
             int keyPressedQueueCount;       // Input keys queue count
@@ -3755,6 +3757,13 @@ bool IsKeyPressed(int key)
     return pressed;
 }
 
+// Check if a key has been pressed again (only PLATFORM_DESKTOP)
+bool IsKeyPressedRepeat(int key)
+{
+    if (CORE.Input.Keyboard.keyRepeatInFrame[key] == 1) return true;
+    else return true;
+}
+
 // Check if a key is being pressed (key held down)
 bool IsKeyDown(int key)
 {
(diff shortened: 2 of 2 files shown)


raysan5:
@actondev I understand this functionality returns true for one frame every certain time (i.e. every 10 frames, depending on the framerate), right?

I'm seeing several PRs requiring key-repeat detection functionality. Personally, I never needed it, it just managed it myself using exposed functions. First of all, that functionality does not seem to be available in all supported platforms, only PLATFORM_DESKTOP (correct me if I'm wrong), second, I got many programs depending on keys input and managing key-repeat myself, not sure how those changes could affect my code (or other users codebases in similar situation).

Why is that feature so important? What are the problems of just managing the key-repeats with already provided functionality?

actondev:
@raysan5 If I understood your question correctly, yes.

Regarding it's usefulness, in keyboard-driven application where on can navigate/perform actions on the GUI, I feel this feature being necessary. Sure one can implement key repeat themselves but a) it's more work, and b) it doesn't respect the system wide setting.
Another issue is if one enables "waiting for events" (so that the main loop isn't rerun without any user input). Is implementing key-repeat yourself also viable in that case?

As I mentioned in the PR description, existing functionality should stay as it is.

Also, if you agree with the general approach here, the same could be used for fixing the "pressed" logic when both the down & up events arrive at the same frame. (see https://github.com/raysan5/raylib/issues/2827#issuecomment-1353756146 ). So similarly as done in this PR, we'd have a `keyPressInFrame`. I also noted this issue in macbook touchpad: when quickly tapping (for clicking) I rarely get mouse presses, since in the same frame while processing the events, a mousedown and a mouseup.

actondev:
Btw I find it funny that 2 PRs for this issue got opened back to back. Personally I'm building a custom GUI & was adding some keyboard navigation, so not having key repeat was such a bummer heh

raysan5:
@actondev thanks for your answer, here my comments:

 > Another issue is if one enables "waiting for events" (so that the main loop isn't rerun without any user input). Is implementing key-repeat yourself also viable in that case?

Yes, it works because keeping a key down is an event, so loop keeps executing.

 > As I mentioned in the PR description, existing functionality should stay as it is.

Agree, that's my main concern with all the related PRs/Issues.

I have to review all the other related PRs more carefully but this one seems to me the most feasible option to support this feature.

rmn20:
It's not a big deal, but maybe IsKeyPressedRepeat could be renamed to IsKeyRepeated for convenience?

n77y:
> It's not a big deal, but maybe IsKeyPressedRepeat could be renamed to IsKeyRepeated for convenience?

@actondev, I also think ```IsKeyPressedRepeat``` should be renamed. Could you rename it to perhaps ```IsKeyRepeated``` or something else in case you can think of somthing better?

If you do rename it I can close my pr and reopen it to add my additional changes if @raysan5 merges your pr.

Also a side request I have is could you make your ```IsKey...``` repeat body consitent with the other ```IsKey...```'s?
```c
bool IsKey...(int key)
{
    if (CORE.Input.Keyboard.keyRepeatInFrame[key] == 1) return true;
    else return false;
}

raysan5:
@rmn20 @n77y About renaming, I also though about that but actually `IsKeyPressedRepeat()` is the most accurate naming for the provided functionality. It returns `IsKeyPressed()` events repeately. Probaly for an advance programmer `IsKeyRepeat()` is clear enough but for a person that starts on programming it can be confusing.

n77y:
@raysan5, I agree and change my mind. ```IsKeyPressedRepeat()``` Is the most accurate even if it’s more verbose.
It looks as if it’s an addition to the ```IsKeyPressed...``` functionality:
```c
if (IsKeyPressed(KEY_BACKSPACE) ||
    IsKeyPressedRepeat(KEY_BACKSPACE))

if (IsKeyPressed(KEY_BACKSPACE) || IsKeyPressedRepeat(KEY_BACKSPACE))

if (IsKeyPressed(KEY_BACKSPACE) || IsKeyRepeated(KEY_BACKSPACE))

actondev:
forced pushed formatting changes

raysan5:
@actondev Thank you very much for this new feature and the required reviews.

@n77y thank you for the review and related improvements.

Issue #3246, opened 2023-08-16
[rlgl] Undefined behavior in `rlLoadTexture` when computing applying a zero offset to a null pointer


ashn-dot-dev:
Please, before submitting a new issue verify and check:

- [x] I tested it on latest raylib version from master branch
- [x] I checked there is no similar issue already reported
  - Similar issue in scope (but separate in cause) to https://github.com/raysan5/raylib/issues/1891.
- [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
- [x] My code has no errors or misuse of raylib
  - Can be replicated using an in-tree example program.

### Issue description

The function `rlLoadTexture` contains the call `glTexImage2D(GL_TEXTURE_2D, i, glInternalFormat, mipWidth, mipHeight, 0, glFormat, glType, (unsigned char *)data + mipOffset)` [here](https://github.com/raysan5/raylib/blob/a86c93ebc0095f6c2ffc14656bfc9e1e37070f72/src/rlgl.h#L3014). When `data` is `NULL`, the expression `(unsigned char *)data + mipOffset` will resolve to `NULL + 0` in the first iteration of the mipmap load loop.

### Environment

Discovered on commit a86c93ebc0095f6c2ffc14656bfc9e1e37070f72 using a `PLATFORM_DESKTOP` build for Linux via Windows Subsystem for Linux with `CUSTOM_CFLAGS='-fsanitize=address,undefined'`.

OpenGL information:
```
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Microsoft Corporation
INFO:     > Renderer: D3D12 (NVIDIA GeForce RTX 3070 Ti)
INFO:     > Version:  4.2 (Core Profile) Mesa 23.0.4-0ubuntu1~22.04.1
INFO:     > GLSL:     4.20
```

### Issue Screenshot

N/A

### Code Example

The UBSan error will trigger on the in-tree example `examples/shapes/shapes_top_down_lights.c` on commit a86c93ebc0095f6c2ffc14656bfc9e1e37070f72 without modification.

```sh
$ (cd /path/to/raylib/src/ && make clean all CC=clang CUSTOM_CFLAGS='-fsanitize=address,undefined' && sudo make install)
$ (cd /path/to/raylib/examples/ && make clean shapes/shapes_top_down_lights)
$ /path/to/raylib/examples/shapes/shapes_top_down_lights 2>&1 | grep 'runtime error'
[...]

PR #3248, opened 2023-08-17
[Feature] Add IsKeyRepeated and key safety checks

n77y:
Hi, In order to make it possible to pick up repeated key events I have added the ```bool GetKeyRepeat(int key)``` function.
A good example of where this would be essential would be in the 'input box' example.
Replace:
```c
if (IsKeyPressed(KEY_BACKSPACE))
{
    letterCount--;
    if (letterCount < 0) letterCount = 0;
    name[letterCount] = '\0';
}
```
With:
```c
for (bool pressed = IsKeyPressed(KEY_BACKSPACE); pressed || GetKeyRepeat(KEY_BACKSPACE); pressed = false)
{
    letterCount--;
    if (letterCount < 0) letterCount = 0;
    name[letterCount] = '\0';
}
```
If you have any questions about renaming, changing, removing or anything else please let me know.
I do not expect the current changes in the pr to be set in stone.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4470,13 +4470,16 @@ static bool InitGraphicsDevice(int width, int height)
 #endif
 
     // Try to enable GPU V-Sync, so frames are limited to screen refresh rate (60Hz -> 60 FPS)
-    // NOTE: V-Sync can be enabled by graphic driver configuration
+    // NOTE: V-Sync can be enabled by graphic driver configuration, it doesn't need
+    // to be activated on web platforms since VSync is enforced there.
+#if !defined(PLATFORM_WEB)
     if (CORE.Window.flags & FLAG_VSYNC_HINT)
     {
         // WARNING: It seems to hit a critical render path in Intel HD Graphics
         glfwSwapInterval(1);
         TRACELOG(LOG_INFO, "DISPLAY: Trying to enable VSYNC");
     }
+#endif
 
     int fbWidth = CORE.Window.screen.width;
     int fbHeight = CORE.Window.screen.height;


raysan5:
@n77y I understand this functionality returns true for one frame every certain time (i.e. every 10 frames, depending on the framerate), right?

I'm seeing several PRs requiring key-repeat detection functionality. Personally, I never needed it, it just managed it myself using exposed functions. First of all, that functionality does not seem to be available in all supported platforms, only PLATFORM_DESKTOP (correct me if I'm wrong), second, I got many programs depending on keys input and managing key-repeat myself, not sure how those changes could affect my code (or other users codebases in similar situation).

Why is that feature so important? What are the problems of just managing the key-repeats with already provided functionality?

n77y:
You would know more than me about whether this would work only on a desktop or not. I could not find any GLFW docs that specified any platform support for GLFW_REPEAT. If that is an automatic game breaker then this pr can be closed, however wouldn’t GetCharPressed only work with desktop as well?

I understand that the GLFW_REPEAT and GLFW CharCallback are both capped at the framerate, however when the framerate is above the char-repeat rate they behave according to the system's settings and not any hardcoded values.

A great example would be if I managed the key-repeat to match my system's. Then my app is used by another user, and they have a different repeat-rate on their system. This would result in inconsistent repeat rates with various keys.

The best workaround I can see with the current state of raylib and its bindings would be to use GetKeyPressed instead of GetCharPressed so I can write my own repeat functionality that's consistent with all the keys and implement a repeat speed setting.

The changes are not designed to change the functionality of any current code bases, that could obviously break their functionality and that would not be ideal. Rather this is intended as a feature meant for newer codebases and older if chosen to add it.

n77y:
If GLFW_REPEAT is confermed to be desktop only then this pr can be closed and @actondev's pr #3245 would work in this case.

raysan5:
> A great example would be if I managed the key-repeat to match my system's. Then my app is used by another user, and they have a different repeat-rate on their system. This would result in inconsistent repeat rates with various keys.

Yes, that's it, depending on the OS and maybe keyboard driver the key-repeat events could not be consistent between systems.

 > The best workaround I can see with the current state of raylib and its bindings would be to use GetKeyPressed instead of GetCharPressed so I can write my own repeat functionality that's consistent with all the keys and implement a repeat speed setting.

That's similar to what I do on some of my tools, I just manage the key-repeats myself checking if `IsKeyDown()` and measuring the timming to allow key-repeat behaviour. Timming is usually consistent between platform and functionality does not rely on the OS or driver. But in that case the user experience could not be the best for the user and the implementation cost is a bit higher.

 > If GLFW_REPEAT is confermed to be desktop only then this pr can be closed and @actondev's pr https://github.com/raysan5/raylib/pull/3245 would work in this case.

I'm looking to the several related issues/PRs to see the best option for raylib.

n77y:
@raysan5, I have made a few changes. I renamed it to IsKeyRepeated because its only one repeat per frame only.
```c
if (IsKeyPressed(KEY_BACKSPACE) || IsKeyRepeated(KEY_BACKSPACE))
{
    letterCount--;
    if (letterCount < 0) letterCount = 0;
    name[letterCount] = '\0';
}
```
Internally there are a few changes as well.
If you prefer my previous commit let me know so I can revert it.

src/rcore.c
```diff
@@ -6049,12 +6054,13 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
...
         {
-            CORE.Input.Keyboard.currentKeyState[keycode] = 1;   // Key down
+            KEY_STATE(keycode) = KEY_STATE_PRESSED;
 
             CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = keycode;
             CORE.Input.Keyboard.keyPressedQueueCount++;
         }
-        else CORE.Input.Keyboard.currentKeyState[keycode] = 0;  // Key up
+        else if (AKeyEvent_getAction(event) == AKEY_EVENT_ACTION_UP) KEY_STATE(keycode) = KEY_STATE_RELEASED;
+        else if (AKeyEvent_getAction(event) == AKEY_EVENT_ACTION_MULTIPLE) KEY_STATE(keycode) = KEY_STATE_REPEATED;
```
n77y, line 6063: @raysan5, also it would appear that andriod supports repeated keys perhaps.

ubkp:
I'm sorry, but is this really necessary? It appears to be a big change from the standard logic of directly accessing `CORE.*` which is used everywhere (and is really handy).

IMHO, if some keypress repeated function is to be added, I'd rather something like https://github.com/raysan5/raylib/pull/3245 which is more minimal and isolated.

src/rcore.c
```diff
@@ -436,8 +436,7 @@ typedef struct CoreData {
 #endif
         struct {
             int exitKey;                    // Default exit key
-            char currentKeyState[MAX_KEYBOARD_KEYS];        // Registers current frame key state
-            char previousKeyState[MAX_KEYBOARD_KEYS];       // Registers previous frame key state
+            char keyState[MAX_KEYBOARD_KEYS];               // Registers frame key state
```
n77y, line 439: @ubkp, correct me if im wrong. The issue is this change? I was unware CORE was used so oftin if this is the case. I can make some changes to the pr if you can be more specific with what the issue is.

ubkp: @n77y That too, but I was mostly looking at `FLAGS_CHECK(KEY_STATE(key)...` and `KEY_STATE(...`. Right now I can tap into anything `CORE.*` and extend it directly/easily without breaking anything and keeping fully upstream compatibility. Your PR proposes a different logic, it changes too much.

IMHO, _keypress repeated_ should be handled by the user (it's not a one size fits all), but if that's to be added, I'd much rather an approach that keeps it mostly isolated/independent to not break what is already working really well. That's why I mentioned https://github.com/raysan5/raylib/pull/3245.

n77y: @ubkp, I have taken your requests and staged them.

ubkp: @n77y That's looking so much better. But why are you implementing a `KEY_SAFE(...`?

Personally, I don't think that making `IsKeyPressed()` require `IsKeyDown()` and `IsKeyReleased()` require `IsKeyUp()` would be a good idea, makes them dependent. IMHO, the current implementation is very resilient due to being independent of each other. Does those functions really need to change?

n77y: @ubkp, I believe ```KEY_SAFE(key)``` to be necessary due to the exposed nature of the ```IsKey...```'s. It only seems mandatory to ensure that we aren’t reading memory we aren’t supposed to and rather simply return ```false``` when the user inputs a invalid key.

I agree, removing dependencies on the ```IsKey...```'s is sensible. I removed them.

n77y: ```KEY_SAFE(key)``` macro removed. Replaced with normal safety checks

ubkp: @n77y Regarding the key safe check, the first sentence of this comment https://github.com/raysan5/raylib/issues/3246#issuecomment-1682256790 comes to mind. So I assumed it was up to the user to validate that. I'm not against it tho, I think that's good.

Just tested your PR on `PLATFORM_DESKTOP` (Linux Mint 21.1 64-bit) and `PLATFORM_WEB` (Chromium 115.0.5790.170 64-bit and Firefox 115.1.0esr 64-bit, both on Linux), and it's working great. Very good job.

---

Tested with:
```
#include "raylib.h"

int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        if (IsKeyRepeated(KEY_ONE)) TraceLog(LOG_INFO, "1 repeated");

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

ubkp:
@n77y Sorry to bump this again. But today I compared this PR with @actondev PR (https://github.com/raysan5/raylib/pull/3245) and (after all the changes) this PR ended up being almost the same as @actondev PR. So, since he presented that solution first, I think his PR should be the accepted one.

This PR, however, adds key safe checks, Raspberry Pi and Android support. So, these improvements could be added after @actondev PR.

Either way, I'd suggesting waiting to see what @raysan5 decides to do about _keypress repeated_ before moving further.

raysan5:
@ubkp I agre that @actondev solution could be merged first and any further improvement could be added as a separate PR.

Despite it isolated nature, not affecting current functionality (something that I love), this is a big change and I prefer to keep PRs as small and self-contained as possible.

src/rcore.c
```diff
@@ -3775,10 +3779,19 @@ bool IsKeyReleased(int key)
...
 {
+    if ((key < 0) || (key >= sizeof(CORE.Input.Keyboard.currentKeyState))) return false;
     if (CORE.Input.Keyboard.currentKeyState[key] == 0) return true;
     else return false;
 }
 
+// Check if a key has been repeated
+bool IsKeyRepeated(int key)
+{
+    if ((key < 0) || (key >= sizeof(CORE.Input.Keyboard.currentKeyState))) return false;
```
actondev, line 3790: Not important, but some comments:
Should be `sizeof(repeatKeyState)`. But then, use another array every time? Checking with `MAX_KEYBOARD_KEYS` instead could make things consistent.

That said, I think these should be asserts instead. If you call this some invalid value this is a silent failure.
```
bool static IsValidKey(int key) {
return key >= 0 && key < MAX_KEYBOARD_KEYS);
}
// and using it
assert(isValidKey(key));
```

I have no objections with merging this one instead, since more platforms seem to work as well!

raysan5:
@actondev Just merged your PR.

@n77y Required improvements from this PR could be added to already added PR.

src/rcore.c
raysan5: > Checking with MAX_KEYBOARD_KEYS instead could make things consistent.

Agree.

 > I think these should be asserts instead. If you call this some invalid value this is a silent failure.

I'm afraid raylib does not use `assert()` anywhere. It can just `TraceLog(LOG_WARNING, ...)` if really required.

actondev:
@raysan5  I fucked up apparently in the last moment force push, apologies!
 @n77y could you fix this bit? https://github.com/raysan5/raylib/commit/b3f82a148ac59ae0615597e553fb64b836616d63#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R3763-R3764 to `return false;` 🤦

n77y:
@actondev, @raysan5, @ubkp, #3256

Issue #3249, opened 2023-08-18
[Text] (Web-based showcase) Emoji example "text_unicode" freezes up after enough emojis are clicked


evanrinehart:
This issue is for the web-hosted emoji example on the raylib website. I could not trigger it on desktop builds of the example.

Click many different emojis and eventually the program stops responding. The number of tries and position of final triggering emoji seems to be random, which makes sense because the program is randomized.

Clicking the same set of working emojis over and over doesn't seem to trigger it. Every time it happens, it was a click on an emoji that was not clicked before. If nothing on the first set of emojis triggers the freeze, press space and try the second set.

I was able to reproduce it on Chrome on OSX, and on Firefox on Ubuntu. The two computers have very different specs so I would be surprised if it was a resource exhaustion issue.

In the screenshot of the frozen program I tried to click on the blue emoji.

![image](https://github.com/raysan5/raylib/assets/1022635/ddf334fe-cda8-41dd-a0c8-0400b3c98779)

ubkp:
Was able to reproduce the issue:

![img](https://github.com/raysan5/raylib/assets/118854183/827bede7-f64a-43d8-b26d-035c0bb6e8ce)

Apparently there's an unescaped line break somewhere on [examples/text/text_unicode.c](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c).

raysan5:
@evanrinehart @ubkp I reproduced it on web and tried to reproduce it on desktop but not possible. This is a compelx example provided by a contributor long time ago. I'm afraid the cost to review it to find the platfomr-dependant issue could be considerable and sincerely I think it does not work it, it's just an example to be used as a reference.

If anyone is ready for that task, I'll keep this issue open for a while but if not, I'm closing it.

ubkp:
@raysan5 I'm working on https://github.com/raysan5/raylib/issues/3231, but, if you don't mind keeping this open for a few days, I can take a look at it after that.

raysan5:
@ubkp Ok, thank you very much, just take it easy, it could be tricky and the work required for this platform-specific issue probably does not worth the time invested.

PR #3250, opened 2023-08-19
Fix text_unicode.c example crashing

ubkp:
The issue causing the `examples/text/text_unicode.c` to crash on `PLATFORM_WEB` was the `Line Feed (LF)` char `\x0A` on some entries of `messages[]` ([L71](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L71), [L77](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L77), [L78](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L78), [L79](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L79), [L80](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L80), [L81](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L81), [L119](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L119), [L120](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L120)) not being escaped when being passed to `SetClipboardText()` ([L195](https://github.com/raysan5/raylib/blob/master/examples/text/text_unicode.c#L195)). All chars on those messages resolve to a singular unicode char, however, the `Line Feed (LF)` resolves to `\n` (with the `\` plus `n`) which causes the unescaped error on web.

This can be solved by escaping the `\x0A` to `\\n` before passing it to `SetClipboardText()` if on `PLATFORM_WEB` ([R195-R202](https://github.com/raysan5/raylib/pull/3250/files#diff-e7b395c4916ca70c8cca98a16064c1b5fd19c53d6dea7c0d3cceee604f8f53a7R195-R202)).

@raysan5 Could you please check if that `MemFree()` call is correct? I saw the warning on the header about `TextReplace()` ([L1417](https://github.com/raysan5/raylib/blob/master/src/raylib.h#L1417)) but I'm not sure if I did that right.

Also (unrelated) small fix to `float textOffsetY` ([R352](https://github.com/raysan5/raylib/pull/3250/files#diff-e7b395c4916ca70c8cca98a16064c1b5fd19c53d6dea7c0d3cceee604f8f53a7R352)).

Tested these proposed changes on _Chromium_ (version 115.0.5790.170 64-bit) and _Firefox_ (version 115.1.0esr 64-bit) both running on _Linux_ (Linux Mint 21.1 64-bit).

Fixes https://github.com/raysan5/raylib/issues/3249.

**Edit 1:** added line marks.
**Edit 2:** forgot to add the test environment.

### DIFF
--- a/examples/text/text_unicode.c
+++ b/examples/text/text_unicode.c
@@ -192,7 +192,14 @@ int main(void)
         {
             selected = hovered;
             selectedPos = hoveredPos;
-            SetClipboardText(messages[emoji[selected].message].text);
+            #if defined(PLATFORM_WEB)
+                // Line Feed (LF) (New Line) must be escaped for SetClipboardText() for web
+                char *escapedString = TextReplace(messages[emoji[selected].message].text, "\x0A", "\\n");
+                SetClipboardText(escapedString);
+                MemFree(escapedString);
+            #else
+                SetClipboardText(messages[emoji[selected].message].text);
+            #endif
         }
 
         Vector2 mouse = GetMousePosition();
@@ -267,7 +274,7 @@ int main(void)
                     a = b;
                     b = tmp;
                 }
-                
+
                 if (msgRect.x + msgRect.width > screenWidth) msgRect.x -= (msgRect.x + msgRect.width) - screenWidth + 10;
 
                 // Draw chat bubble
@@ -287,11 +294,11 @@ int main(void)
                 DrawText(info, (int)pos.x, (int)pos.y, 10, RAYWHITE);
             }
             //------------------------------------------------------------------------------
-            
+
             // Draw the info text
             DrawText("These emojis have something to tell you, click each to find out!", (screenWidth - 650)/2, screenHeight - 40, 20, GRAY);
             DrawText("Each emoji is a unicode character from a font, not a texture... Press [SPACEBAR] to refresh", (screenWidth - 484)/2, screenHeight - 16, 10, GRAY);
-            
+
         EndDrawing();
         //----------------------------------------------------------------------------------
     }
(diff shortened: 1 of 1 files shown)


ubkp:
### Follow up issue

There appears to be some other problem on `SetClipboardText()` for web. Passing an unescaped char to it shouldn't cause the entire program/game to crash on web. Looking at its code ([L2150](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L2150)) I expected it just not to finish, but not crashing.

I assembled a minimal reproduction code to test the issue:

```
#include "raylib.h"

int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        // Works
        if (IsKeyPressed(KEY_ONE))   SetClipboardText("1\\n2");

        // Crashes
        if (IsKeyPressed(KEY_TWO))   SetClipboardText("1\n2");

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

**Note:** Regardless, the `text_unicode.c` fix on the PR was still necessary because `text_unicode.c` should still pass a valid string to `SetClipboardText()` so it can complete successfully.

raysan5:
@ubkp The issue is actually in `SetClipboardText()`, not properly supporting UTF-8 string data as it should, or at least not on `PLATFORM_WEB`. I think the easiest solution for this example is just removing that line. It's not needed and I don't know why it is called.

`SetClipboardText()` on web could be threated as another separate issue.

ubkp:
@raysan5 Adjusted the fix as instructed ([L190](https://github.com/raysan5/raylib/pull/3250/files#diff-e7b395c4916ca70c8cca98a16064c1b5fd19c53d6dea7c0d3cceee604f8f53a7L190), [L195](https://github.com/raysan5/raylib/pull/3250/files#diff-e7b395c4916ca70c8cca98a16064c1b5fd19c53d6dea7c0d3cceee604f8f53a7L195)).

I'll open a new issue about the `SetClipboard()`.

raysan5:
@ubkp thanks for the review!

Issue #3251, opened 2023-08-19
[core] `SetClipboardText()` UTF-8 issue for `PLATFORM_WEB`


ubkp:
- [x] I tested it on latest raylib version from master branch
 - [x] Issue forked from https://github.com/raysan5/raylib/pull/3250#issuecomment-1684737568
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

`SetClipboardText()` is not properly supporting `UTF-8` string data on `PLATFORM_WEB`. E.g.: passing an unescaped char (like `\n`, `\x0A`, `0x0A`) to it causes the program/game to crash on web.

Relevant parts of code: [L2142-L2153](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L2142-L2153) and [L2150-L2151](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L2150-L2151). I'll investigate further.

### Environment

Platform: _Web_
Browser: _Chromium_
Version: _115.0.5790.170 (64-bit)_
OS: _Linux Mint 21.1 (64-bit)_

Platform: _Web_
Browser: _Firefox_
Version: _115.1.0esr (64-bit)_
OS: _Linux Mint 21.1 (64-bit)_

### Code Example

Minimal reproduction code to test the issue:

```
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);
    while (!WindowShouldClose()) {

        // Works
        if (IsKeyPressed(KEY_ONE))   SetClipboardText("1\\n2");

        // Crashes
        if (IsKeyPressed(KEY_TWO))   SetClipboardText("3\n4");
        if (IsKeyPressed(KEY_THREE)) SetClipboardText("\x35\x0A\x36");

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

**Edit 1:** updated code example for clarity.
**Edit 2:** added UTF-8 literals test to code example.

PR #3253, opened 2023-08-19
Fix SetClipboardText for web

ubkp:
Changes the `single quotations` for `backticks` on the `string` of the javascript line that gets executed by `emscripten_run_script`. The `single quotations` were likely getting mixed at somepoint between emscripten and the browser.

Tested these proposed changes successfully on Chromium (version 115.0.5790.170 64-bit) and Firefox (version 115.1.0esr 64-bit) both running on Linux (Linux Mint 21.1 64-bit).

Fixes https://github.com/raysan5/raylib/issues/3251.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2148,7 +2148,7 @@ void SetClipboardText(const char *text)
 #if defined(PLATFORM_WEB)
     // Security check to (partially) avoid malicious code
     if (strchr(text, '\'') != NULL) TRACELOG(LOG_WARNING, "SYSTEM: Provided Clipboard could be potentially malicious, avoid [\'] character");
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp It seems `backticks` change the string data to _template literals_. It seems [they are something different}(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), could it imply potential security concerns (considering it accepts variables interpretation)? Alos, it seems they are not supported on Internet Explorer (minimal concern).

ubkp:
@raysan5 Uhn, I hadn't consider that, good catch. I'll investigate further.

PR #3254, opened 2023-08-19
add build.zig options for individual modules

actondev:
This adds the ability to select only some of the modules of raylib.
The motivation is that I had an issue with using miniaudio myself & I was missing all the decoding functionality (since `MA_NO_WAV`, `MA_NO_FLAC`) etc are defined in `raudio.c`


raysan5:
@actondev thanks for the review, actually this functionality is also available in provided `Makefile`, some modules are optional and not required in many cases (i.e. `raudio`, `rmodels`).

Just note that some modules are slightly dependant on others: 
 - `rtext` depends on `rtextures`
 - `rshapes` could depend on `rtextures`
 - `rcore` could depend on `rtext` (default font)

PR #3255, opened 2023-08-20
[CORE] Fix the emscripten_set_main_loop_timing warning when enabling VSync on web

VitusVeit:
### Changes:
When setting the flag `FLAG_VSYNC_HINT` with `SetConfigFlags` the warning

> emscripten_set_main_loop_timing: Cannot set timing mode for main loop since a main loop does not exist! Call emscripten_set_main_loop first to set one up.

caused by `glfwSwapInterval(1)` won't appear anymore.

### Why:
This fix was [already added](https://github.com/raysan5/raylib/commit/0bacc978c39a234da4a2b73ce3cb5b43e072a103) into the source code of raylib but it was since then removed.

I think it was accidental since adding this fix to my up-to-date local build of raylib fixed the problem and didn't seem to cause any problems on platforms that do "support" VSync like desktop.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4470,13 +4470,16 @@ static bool InitGraphicsDevice(int width, int height)
 #endif
 
     // Try to enable GPU V-Sync, so frames are limited to screen refresh rate (60Hz -> 60 FPS)
-    // NOTE: V-Sync can be enabled by graphic driver configuration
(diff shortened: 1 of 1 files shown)


raysan5:
@VitusVeit Good catch! Thanks for the review!

PR #3256, opened 2023-08-20
[Feature] IsKey... safety checks and more

n77y:
Let me know if there should be tracelogs or asserts or only return false in the safety checks


n77y:
not complete

src/rcore.c
```diff
@@ -5607,6 +5620,8 @@ static void KeyCallback(GLFWwindow *window, int key, int scancode, int action, i
     else if(action == GLFW_PRESS) CORE.Input.Keyboard.currentKeyState[key] = 1;
     else if(action == GLFW_REPEAT) CORE.Input.Keyboard.keyRepeatInFrame[key] = 1;
 
+    if (action == GLFW_REPEAT) CORE.Input.Keyboard.keyRepeatInFrame[key] = 1;
```
actondev: not needed

src/rcore.c
```diff
@@ -6075,6 +6090,8 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
         }
         else CORE.Input.Keyboard.currentKeyState[keycode] = 0;  // Key up
 
+        if (AKeyEvent_getAction(event) == AKEY_EVENT_ACTION_MULTIPLE) CORE.Input.Keyboard.keyRepeatInFrame[keycode] = 1;
```
actondev, line 6091: it seems that presently when we get `AKEY_EVENT_ACTION_MULTIPLE` we mark it as keyup.
```suggestion
const eventAction = AKeyEvent_getAction(event);
// handle each AKEY_EVENT_ACTION_DOWN, AKEY_EVENT_ACTION_UP, AKEY_EVENT_ACTION_MULTIPLE
```

actondev (commented):
Some comments. My opinion still stands that the key safety checks should be an assert

n77y:
> Some comments. My opinion still stands that the key safety checks should be an assert

I think in order to be consistent with raylib we should just leave it as just a ```return false;```
We could add a TraceLog as @raysan5 pointed out, however that would seem excessive in my opinion.

ubkp:
I really like this. Good job.

Retested successfully on `PLATFORM_DESKTOP` (_Linux Mint 21.1 64-bit_) and `PLATFORM_WEB` (_Chromium 115.0.5790.170 64-bit_ and _Firefox 115.1.0esr 64-bit_, both running on _Linux_) with:
```
#include "raylib.h"

int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        int square = 0;
        if (IsKeyPressedRepeat(KEY_ONE)) square = 1;

        BeginDrawing();
        ClearBackground(RAYWHITE);
        if (square == 1) DrawRectangle(20,  20, 50, 50, RED);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

n77y:
> Some comments. My opinion still stands that the key safety checks should be an assert

To be clear I'm not against using ```assert```, however I believe it’s best to adhere to the owners standards

n77y:
@ubkp, Thanks!

raysan5:
@n77y @actondev @ubkp Thank you very much for the improvements and review. Returning false is the behaviour I prefer.

Is issue #2827 being addressed with this PR and previous key-repeat feature addition?

n77y:
@raysan5, No, I believe #2827 is separate. @actondev mentioned it here https://github.com/raysan5/raylib/pull/3245#:~:text=(see%20%232827%20(comment)%20).%20So%20similarly%20as%20done%20in%20this%20PR%2C%20we%27d%20have%20a%20keyPressInFrame.

It looks like a separate ```IsKeyPressed...(int key)``` would be needed to account for a key pressed and released in a single frame. Interestingly a user could without too much difficulty get around this issue by ```GetKeyPressed()``` because when a key is released it does not remove it from the queue. I don’t think that would be ideal though.

Issue #3260, opened 2023-08-24
[rlgl.h] WebGL default shaders not linking on no less than 1 browser 


branc116:
Please, before submitting a new issue verify and check:

 - [+] I tested it on latest raylib version from master branch
 - [+] I checked there is no similar issue already reported
 - [+] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [+] My code has no errors or misuse of raylib

#### Init
When using default shader on web, there is a linking error and nothing is shown.

#### More testing

I noticed this when working on something that I work on, but now I see that the same problem is present in older versions of raylib.
On https://www.raylib.com/examples.html every example outputs the same linking error.

#### My guess

My guess is that this may have something to do with my browser. I can recall that the same examples worked maybe 2 months ago..

#### Testing my guess

I can confirm It's a bug in the browser...
Tested on a different browser:
```
* Working:        Brave Browser 116.1.57.47 
* Working:        Brave - Beta 1.58.83, Chromium 116.0.5845.62 ( Andriod )
* Not working: Brave Browser 116.1.58.84 beta
* Working:        Chromium 118.0.5969.0
```

#### I'm sad and a bit mad
What do I do? I guess this is brave problem.
It looks like everything works on every browser exept for anything on browser that I'm using...

### Environment
On desktop:
```
INFO: Initializing raylib 4.6-dev
INFO: DISPLAY: Device initialized successfully
INFO:     > Display size: 1366 x 768
INFO:     > Screen size:  1280 x 720
INFO:     > Render size:  1280 x 720
INFO:     > Viewport offsets: 0, 0
INFO: GLAD: OpenGL extensions loaded successfully
INFO: GL: Supported extensions count: 226
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Intel
INFO:     > Renderer: Mesa Intel(R) UHD Graphics 600 (GLK 2)
INFO:     > Version:  4.6 (Core Profile) Mesa 23.1.6
INFO:     > GLSL:     4.60
INFO: GL: VAO extension detected, VAO functions loaded successfully
INFO: GL: NPOT textures extension detected, full NPOT textures supported
INFO: GL: DXT compressed textures supported
INFO: GL: ETC2/EAC compressed textures supported
```
On web:
```
INFO: DISPLAY: Device initialized successfully
INFO:     > Display size: 800 x 450
INFO:     > Screen size:  800 x 450
INFO:     > Render size:  800 x 450
[...]

PR #3261, opened 2023-08-24
Fix bug where default shaders was not linking.

branc116:
I've tested this on every browser that I tested in issue #3260 and default example works on all of them.


branc116:
Idk, maybe it's not worth mergin if it only fixes things in my browser on my machine.
I lost many hours on finding a solution for this and don't want it to be for nothing. :)

raysan5:
@branc116 thanks for the improvement!

PR #3264, opened 2023-08-25
[Image] Validate that ImageDrawRectangleRec is drawing entirely inside the image

JeffM2501:
ImageDrawRectangleRec does not call ImageDrawPixel for every pixel it places and it has a bad assumption that the rect is fully contained by the image.
For cases where this is not true, it will write to the incorrect data, or overrun the image buffer.
This is bad.

This PR clamps the rectangle to the image bounds and verifies that the rectangle is inside the image before drawing it, preventing this kind of buffer overrun.

This fixes several drawing functions since many call ImageDrawRectangleRec to do fills.


raysan5:
@JeffM2501 Great! Thank you very much for this improvent! I remember fighting with those checks for long time while implementing it!

PR #3266, opened 2023-08-27
Fix CMake extraneous -lglfw

### LINKED ISSUE #3265: [build] Generated `raylib.pc` by CMake shouldn't define `Libs.private: -lglfw`
## Reproduction

Do the following in shell:

```
# be in raylib repo
cmake -B build -D USE_EXTERNAL_GLFW=ON
cd build
ninja raylib/package # or some other build command
cat raylib/raylib.pc
```

What `raylib/raylib.pc` looks like

```
prefix=/usr/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: raylib
Description: A simple and easy-to-use library to enjoy videogames programming
URL: https://github.com/raysan5/raylib
Version: 4.5.0
Libs: -L"${libdir}" -lraylib 
Libs.private: -lglfw
Requires.private: glfw3
Cflags: -I"${includedir}"
```

## What's the problem

The line `Libs.private: -lglfw` is not needed. It actually causes problem, because on my system, `-lglfw3` is correct.

Once installed, the output of `pkg-config --libs --static raylib` will contain `-lglfw` (which is incorrect).

If you reproduce this again but without `-D USE_EXTERNAL_GLFW=ON`, the same problem exists. `raylib.pc` should now look like

```
Libs.private: -lglfw
Requires.private: 
```

... which is also wrong.

## Workaround

After install, remove the line `Libs.private: -lglfw` in `/usr/local/lib/pkgconfig/raylib.pc`.

iacore:
Closes #3265.

The problem: LIBS_PRIVATE is a list of library names (used by pkg-config), but the shared library of the same name doesn't always exist. In this case, `pkg-config glfw` is valid, while `-lglfw` is not. Therefore, `library_path_to_linker_flags` can't be used, since `LIBS_PRIVATE` are not executable library names.

I'm not sure if this doesn't break other platforms (other than Linux). @raysan5, can you find someone good at CMake to look at this?

### DIFF
--- a/cmake/InstallConfigurations.cmake
+++ b/cmake/InstallConfigurations.cmake
@@ -9,8 +9,7 @@ install(
 # PKG_CONFIG_LIBS_PRIVATE is used in raylib.pc.in
 if (NOT BUILD_SHARED_LIBS)
     include(LibraryPathToLinkerFlags)
-    library_path_to_linker_flags(__PKG_CONFIG_LIBS_PRIVATE "${LIBS_PRIVATE}")
-    set(PKG_CONFIG_LIBS_PRIVATE ${__PKG_CONFIG_LIBS_PRIVATE} ${GLFW_PKG_LIBS})
+    set(PKG_CONFIG_LIBS_PRIVATE ${GLFW_PKG_LIBS})
     string(REPLACE ";" " " PKG_CONFIG_LIBS_PRIVATE "${PKG_CONFIG_LIBS_PRIVATE}")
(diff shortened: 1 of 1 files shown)


raysan5:
@iacore thanks for the fix! I will ask the raylib community for some CMake expert for review.

raysan5:
@iacore Just merged it, if there is any problem someone will complain sooner or later. Thanks for the review.

PR #3272, opened 2023-08-29
Fix example/audio/audio_module_player.c help instructions and small bug

asdqwe:
Hello,

The [example/audio/audio_module_player.c](https://github.com/raysan5/raylib/blob/master/examples/audio/audio_module_playing.c) is missing the help instructions for its implemented interactions:

| Bind    | Action             |
| ------- | ------------------ |
| [SPACE](https://github.com/raysan5/raylib/blob/master/examples/audio/audio_module_playing.c#L78)   | Restart music      |
| [P](https://github.com/raysan5/raylib/blob/master/examples/audio/audio_module_playing.c#L85)       | Pause/resume       |
| [UP/DOWN](https://github.com/raysan5/raylib/blob/master/examples/audio/audio_module_playing.c#L93-L94) | Change music speed |

It also has a small issue where if the user pauses the music/animation and then proceeds to restart the music, the animation will remain paused.

This pull request adds the missing help instructions and fixes resuming the animation if the music is restarted.

Best regards.


raysan5:
@asdqwe Thank you very much! This is a really nice addition! Please note that the screenshot for the example should also be updated. You can take a new screenshot automatically pressing F12 while running the example, it should be renamed to same name as example code file.

PR #3283, opened 2023-09-02
Fix uninitialized thread-locals in stbi #3282

### LINKED ISSUE #3282: [rtextures] Uninitialized thread-locals in stbi
### Issue description
- I had an issue where textures were loaded upside-down.
- As it turns out, stb_image has thread-local variables to make an image be flipped after loading [here](https://github.com/raysan5/raylib/blob/ec6d3bb688ae8c137152eca7cdef9885badd9bee/src/external/stb_image.h#L1119-L1133).
- With `-O3`, these variables are not initialized and contain garbage. With `-O0` they are zeroed apparently.
- As a result, they evaluate to `true` and the STBI image loading code took the branch (`if (stbi__vertically_flip_on_load)`) to vertically flip the image.

### Environment

`clang 16.0.6` (from `zig 0.12.0-dev.252+5dc2db805`), mingw64, `-O3` (`-Doptimize=ReleaseFast`)

### Code Example
`examples/textures/textures_image_loading.c`

### Suggested Fix
1. Either the thread-locals are kept and initialized with `stbi_set_flip_vertically_on_load_thread` before being used, or
2. I prefer disabling them with `#define STBI_NO_THREAD_LOCALS` in `rtextures.c` (1-line PR will follow)

jbarthelmes:
Fixes #3282

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -147,6 +147,8 @@
     #define STBI_FREE RL_FREE
     #define STBI_REALLOC RL_REALLOC
 
+    #define STBI_NO_THREAD_LOCALS
+
     #define STB_IMAGE_IMPLEMENTATION
     #include "external/stb_image.h"         // Required for: stbi_load_from_file()
                                             // NOTE: Used to read image data (multiple formats support)


raysan5:
@jbarthelmes thanks for the fix!

PR #3287, opened 2023-09-03
Set correct includes for ES3 support, fix GL_HALF_FLOAT with custom define

manuel5975p:
Maybe @Not-Nik needs to check it.


Not-Nik:
`GL_HALF_FLOAT` is [part of OpenGL ES3](https://registry.khronos.org/OpenGL-Refpages/es3.0/) and [of WebGL 2](https://registry.khronos.org/webgl/specs/2.0/#3.7). Does this fix a specific build error you're experiencing?

manuel5975p:
Yes, somehow it is not defined on my desktop es3 headers, but maybe they are done wrong. For web, the fix is not necessary. Then again, my desktop platform's glfw does not support es3 at all, so I don't know.

Not-Nik:
If you are on Desktop, why build for ES3 at all? Just use normal 3.3.

manuel5975p:
Oh I'll gladly toss that out, I thought it was something Ray wanted to maintain support for.

raysan5:
> I thought it was something Ray wanted to maintain support for.

@manuel5975p Not really, I did not add support for OpenGL ES 3.0, it was a user contribution and personally, I don't need it. It was accepted because some users could need it but I'm not maintaining it.

raysan5:
I'm not merging these changes. It mixes `PLATFORM_xxx` defines that belong to `rcore` with `GRAPHICS_xxx` definitions that belong to `rlgl` module. Also, it seems the benefits of ES3 are minimal for most users and it could require more testing.

manuel5975p:
I got rid of the platform check in rlgl (there already many others though, so it's already been mixed), but we have to check the Graphics macro to even include glfw on emscripten. How do you suggest to get around this?

raysan5:
> How do you suggest to get around this?

I don't know, it could be the case that it is not possible.

Do you want me to reopen this PR? Are you still working on this feature?

Issue #1891, opened 2021-07-25
[models] UBSAN flags 0 offset to NULL in DrawMesh


rcorre:
**WARNING: Please, read this note carefully before submitting a new issue:**

It is important to realise that **this is NOT A SUPPORT FORUM**, this is for reproducible BUGS with raylib ONLY.

There are lots of generous and helpful people ready to help you out on [raylib Discord forum](https://discord.gg/raylib) or [raylib reddit](https://www.reddit.com/r/raylib/).

Remember that asking for support questions here actively takes developer time away from improving raylib.

---

Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

*Briefly describe the issue you are experiencing (or the feature you want to see added to raylib). Tell us what you were trying to do and what happened instead. Remember, this is not the best place to ask questions. For questions, go to [raylib Discord server](https://discord.gg/raylib).*

UBSAN flags https://github.com/raysan5/raylib/blob/87b542085be9dc55dae35b788596877ba56be9d8/src/models.c#L1186 as undefined behavior (adding an offset to a NULL pointer, which happens at https://github.com/raysan5/raylib/blob/87b542085be9dc55dae35b788596877ba56be9d8/src/rlgl.h#L3124)

```
rlgl.h:3124:84: runtime error: applying zero offset to null pointer
```

From some reading around (e.g. https://patchwork.kernel.org/project/git/patch/20200125053834.GB744673@coredump.intra.peff.net/), it sounds like adding _any_ offset (even 0) to NULL is technically undefined behavior.

This was introduced sometime between raylib 3.5.0 and raylib 3.7.0, and still exists on `master`.

### Environment

Linux, desktop.

### Issue Screenshot

*If possible, provide a screenshot that illustrates the issue. Usually an image is better than a thousand words.*

### Code Example

*Provide minimal reproduction code to test the issue. Please, format the code properly and try to keep it as simple as possible, just focusing on the experienced issue.*

1. Compile raylib with the undefined behavior sanitizer
2. Run `shaders_basic_lighting` example

rcorre:
Actually I'm kind of confused by this line anyways:

```
            if (mesh.indices != NULL) rlDrawVertexArrayElements(0, mesh.triangleCount*3, 0);
```

Why check if `indices` is non-null, only to pass `0`? Is that supposed to be

```
[...]

PR #3292, opened 2023-09-05
Disable UBSAN in zig builds.

rcorre:
Zig debug builds automatically enable ubsan.
As the fix for #1891 had to be reverted, debug builds using zig will crash like so:

```
Illegal instruction at address 0x3237d2
raylib/src/rlgl.h:3690:91: 0x3237d2 in rlDrawVertexArrayElements (/home/rcorre/src/raylib-zig-template/raylib/src/rcore.c)
    glDrawElements(GL_TRIANGLES, count, GL_UNSIGNED_SHORT, (const unsigned short *)buffer + offset);
```

This disables UBSAN when using zig to build raylib.


raysan5:
@rcorre thanks for the review!

PR #3293, opened 2023-09-05
DrawText - DrawFPS improvements

uuxxaa:
DrawText - DrawFPS improvements
1. acquire FPS only once per DrawFPS
2. use const for non changing values as hint to the compiler.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1069,13 +1069,14 @@ bool ExportFontAsCode(Font font, const char *fileName)
 // NOTE: Uses default font
 void DrawFPS(int posX, int posY)
 {
-    Color color = LIME;                         // Good FPS
-    int fps = GetFPS();
+    const int fps = GetFPS();
 
-    if ((fps < 30) && (fps >= 15)) color = ORANGE;  // Warning FPS
-    else if (fps < 15) color = RED;             // Low FPS
+    Color color =
+        fps < 15 ? RED :    // Low FPS
+        fps < 30 ? ORANGE : // Warning FPS
+        LIME;               // Good FPS
 
-    DrawText(TextFormat("%2i FPS", GetFPS()), posX, posY, 20, color);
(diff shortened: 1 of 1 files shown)


src/rtext.c
```diff
@@ -1069,13 +1069,14 @@ bool ExportFontAsCode(Font font, const char *fileName)
 // NOTE: Uses default font
 void DrawFPS(int posX, int posY)
 {
-    Color color = LIME;                         // Good FPS
```
raysan5, line 1072: I prefer the variable initialized at init and only updated when required.

src/rtext.c
```diff
@@ -1069,13 +1069,14 @@ bool ExportFontAsCode(Font font, const char *fileName)
...
-    int fps = GetFPS();
+    const int fps = GetFPS();
```
raysan5, line 1072: What is the real benefit of `const` on this situation? I try to avoid `const` usage in raylib as much as possible.

src/rtext.c
uuxxaa: In that case, then the code will assign the value twice wasting cycles.

src/rtext.c
uuxxaa: It is a hint to the compiler, and to the reader that author does not expect the value to change. Compiler can make optimizations based on this.

src/rtext.c
raysan5: I think it's better for readability, I use that approach in most functions. I prefer to keep the consistency between functions despite the additional cost.

src/rtext.c
raysan5: Yes, just to keep the code simpler.

uuxxaa: Alright, next iter.

src/rtext.c
uuxxaa: Alright, next iter.

raysan5:
As the change only requires changing fps, I can do it, so no need to keep asking for reviews.

PR #3300, opened 2023-09-08
[CORE] expose a function to get the raw frame buffer size regardless of DPI mode.

JeffM2501:
This PR adds a single function to raylib core. It exposes the raw GLFW frame buffer size.

This is needed for integrations with third party libs such as DearImGui. In order to have these libraries know the correct sizes for the frame buffer (and scissors), and to work aground some problems on MacOS.

I have been trying for about a year to find a way to get this to work on all platforms without adding a function to raylib, but there is just no consistent way to get the data.
I've been directly calling GLFW and that works for platforms that link raylib static, but any binding that uses raylib dynamic (like C#) can't do that, so I need a function that exposes this that is part of the export library.

This PR will make these external libs (mainly rlImGui) cleaner and easier to maintain.

If there is any other way to get this data, please let me know.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -979,6 +979,7 @@ RLAPI int GetMonitorPhysicalHeight(int monitor);                  // Get specifi
 RLAPI int GetMonitorRefreshRate(int monitor);                     // Get specified monitor refresh rate
 RLAPI Vector2 GetWindowPosition(void);                            // Get window position XY on monitor
 RLAPI Vector2 GetWindowScaleDPI(void);                            // Get window scale DPI factor
+RLAPI Vector2 GetFrameBufferSize(void);                           // Get the actual frame buffer size regardless of DPI mode
 RLAPI const char *GetMonitorName(int monitor);                    // Get the human-readable, UTF-8 encoded name of the specified monitor
 RLAPI void SetClipboardText(const char *text);                    // Set clipboard text content
 RLAPI const char *GetClipboardText(void);                         // Get clipboard text content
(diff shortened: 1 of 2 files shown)


raysan5:
Hi @JeffM2501, raylib exposes the following functions to get "screen" size: 
```c
RLAPI int GetScreenWidth(void);                                   // Get current screen width
RLAPI int GetScreenHeight(void);                                  // Get current screen height
RLAPI int GetRenderWidth(void);                                   // Get current render width (it considers HiDPI)
RLAPI int GetRenderHeight(void);                                  // Get current render height (it considers HiDPI)
```
but internally it stores the following variables:
```c
Size display;                       // Display width and height (monitor, device-screen, LCD, ...)
Size screen;                        // Screen width and height (used render area)
Size previousScreen;                // Screen previous width and height (required on borderless windowed toggle)
Size currentFbo;                    // Current render width and height (depends on active fbo)
Size render;                        // Framebuffer width and height (render area, including black bars if required)
```

I understand `screen` could not represent the real framebuffer size when black bars get added but what about `render`?

Also, checking the code, I see it gets `CORE.Window.currentFbo` but later is overriden by `glfwGetFramebufferSize(glfwGetCurrentContext(), &w, &h);`. 

How do they differ? Why `CORE.Window.currentFbo` is not valid on `PLATFORM_DESKTOP`? Is it because the HighDPI consideration? Technically HighDPI should be considered because it modifies current framebuffer...

In any case, if it gets added I prefer to keep the consistency with the other functions:
```c
RLAPI int GetFramebufferWidth(void);
RLAPI int GetFramebufferHeight(void);
```

JeffM2501:
I'm going to do some more research on this. Thanks

PR #3301, opened 2023-09-09
Remove extra if directive

michaelfiber:
I noticed this superfluous #if in the code that I missed with the initial cleanup. It serves no purpose as it's wrapped in an identical if statement.


raysan5:
@michaelfiber thanks for the review!

PR #3311, opened 2023-09-13
Submodules

michaelfiber:
This is an experiment into the feasibility of moving rcore functions to platform specific submodules:
- rcore_desktop.c
- rcore_drm.c
- rcore_web.c
- rcore_android.c

This is accomplished by moving appropriate function definitions to a new rcore.h that rcore.c and all the submodules include. There are also changes to the Makefile to conditionally include only the appropriate submodule when compiling.

The goal is to figure out what decisions have to be made to actually divide rcore.c into submodules.  I'll start issues for discussion when this process exposes interesting choices that must be made to make this change.


raysan5:
@michaelfiber Wow! This is a BIG change of raylib and I see it still requires lot of work but it seems it's going in the right direction! Thank you very much for starting working on it!

Input polling should be defined per platform and maybe some other part...

michaelfiber:
> @michaelfiber Wow! This is a BIG change of raylib and I see it still requires lot of work but it seems it's going in the right direction! Thank you very much for starting working on it!
> 
> Input polling should be defined per platform and maybe some other part...

Yes there is a lot more work to do on it and draft PR showing up here was actually a mistake on my part, I meant to keep it in my fork until I had more work done!  Right now I am focusing on having a good way to check if I am breaking anything on a platform that I don't have access to at the moment and Github actions are great for that.

michaelfiber:
All of the functions from rcore that were listed as depeneding on glfw have moved to submodules.  A few other functions that were mostly platform specific code moved as well. There may still be more functions in rcore that it would make sense to move to submodules.

I think from where it is right now it may be possible for people to grab a specific submodule and start debugging further to get this closer to ready for release.

raysan5:
@michaelfiber Thanks!

PR #3314, opened 2023-09-14
Allow excluding color constants from raylib.h

Peter0x44:
Add a macro, RL_NO_COLOR_CONSTANTS, to allow excluding the color macros
from raylib.h

Certain libraries conflict with raylib's color macros, and can't be
included in the same file

One library affected by this is reactphysics3d. the RED macro mangles
this particular line:
https://github.com/DanielChappuis/reactphysics3d/blob/17dd22e677ed861b0d4ece0c00a7e3cb503cc2f0/include/reactphysics3d/utils/DebugRenderer.h#L62
So, it cannot be used in the same TU as raylib.h

Users may also choose to provide their own choice of color constants,
different to raylib's choice. This gives users the freedom to do so,
without modifying raylib.h

One practical example of this is raylib-namespace.h
https://github.com/raylib-extras/extras-cpp/blob/master/raylib_namespace.h
which prefers to define the color constants as constants instead of
macros

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -161,6 +161,7 @@
 
 // Some Basic Colors
 // NOTE: Custom raylib color palette for amazing visuals on WHITE background
+#if !defined(RL_NO_COLOR_CONSTANTS)
 #define LIGHTGRAY  CLITERAL(Color){ 200, 200, 200, 255 }   // Light Gray
 #define GRAY       CLITERAL(Color){ 130, 130, 130, 255 }   // Gray
 #define DARKGRAY   CLITERAL(Color){ 80, 80, 80, 255 }      // Dark Gray
@@ -188,6 +189,7 @@
 #define BLANK      CLITERAL(Color){ 0, 0, 0, 0 }           // Blank (Transparent)
 #define MAGENTA    CLITERAL(Color){ 255, 0, 255, 255 }     // Magenta
 #define RAYWHITE   CLITERAL(Color){ 245, 245, 245, 255 }   // My own White (raylib logo)
+#endif
 
 //----------------------------------------------------------------------------------
 // Structures Definition
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -143,7 +143,7 @@
 // Some compilers (mostly macos clang) default to C++98,
 // where aggregate initialization can't be used
 // So, give a more clear error stating how to fix this
-#if !defined(_MSC_VER) && (defined(__cplusplus) && __cplusplus < 201103L)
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 Sorry, but I try to avoid any config flag in the raylib header for the moment.

Peter0x44:
No problem - I realized I can just undefine the macros myself, anyway
There is no need to actually touch raylib.h to do that

cpowell:
> If there is any conflict with other libraries, it should be resolved in the other libraries.

@raysan5 I'm filled with respect for you and for raylib itself. But I think this last comment deserves pushback. Saying "I choose not to namespace my code, including basic everyday tokens like 'RED' or 'WHITE' - and this is totally the other libraries' problem" is pretty cold and starkly hostile. Name collisions are everybody's concern.

It's not that a fix isn't available - Peter0x44 mentioned one - but your "talk to the other guy" attitude is uncool.

raysan5:
@cpowell No intention to be rude or have an uncool attitude. It's just a design decision and I prefer to stick to it. Users facing this issue can customize raylib for their needs, I just prefer to avoid that kind of changes in main project.

orcmid:
I think there is a misunderstanding.  

In particular, see all of the things that become broken if the color constants are *not* defined in `raylib.h`.  Even the very first simple example relies on them.  And  "namespacing" them (i.e., putting prefixes on all of them) is going to be a gigantic breaking change.  There may be enough of that for 5.0 if the 

I think the @Peter0x44 notion of redefining them in ones own project, including undefining them, is an useful alternative for someone who wants to have that happen for their own projects.   Making the adjustment locally seems a better solution that wanting it solved globally.  It also does not require touching the raylib `src/` directory.

I see is a kind of scope-creep being requested that is a great distance from providing a straightforward modernization of BGI.  And in any case, if Ramon does not see that as his  mission in life, creating tasks for him is not going to get far.

Peter0x44:
@cpowell I'm happy to discuss this further elsewhere but I don't think breaking 10+ years of code with prefixes is the right idea either
I agree with Ray's reasoning
Your demand of this change in this manner, and attitude here is uncool too

Anyway, I simply realized there was a better way to do what I wanted, it didn't occur to me at the time

cpowell:
<sigh> @Peter0x44 I made no such demand. It's @raysan5 's library to maintain as he sees fit, and I accepted his and @orcmid 's explanation of the matter. I'm certainly not going to bicker with them.

My gripe, and one I think is fair, is that Ray said that it's incumbent upon other library writers to adjust for his choices. I don't agree with that reasoning at all. Still, Ray makes a fine product and I admire him, and it, very much, and I'll employ workarounds as necessary to keep using raylib.

Issue #3315, opened 2023-09-14
[rmodels] DrawMeshInstanced() is not working


Steelsmasher:
I can't get DrawMeshInstanced() to display anything, I just get a black screen. Same issue as #2707.
The Mesh Instancing web example `examples/shaders/shaders_mesh_instancing` doesn't work regardless of which browser I use.

ubkp:
Was able to confirm it's no longer working. Actually, by chance, I tested it on the start of the month and it was working. So some recent change must have broken it.

Steelsmasher:
Are shaders necessary to use DrawMeshInstanced? I was hoping to draw a large number of instances but I haven't got the handle of shaders yet.

raysan5:
I just tried the example on desktop and web (locally) and it works as expected with latest raylib at this same moment.

Attaching my local compilation: [shaders_mesh_instancing.zip](https://github.com/raysan5/raylib/files/12620822/shaders_mesh_instancing.zip)

I vaguely remember I already faced this issue in the past but it's really weird, it works ok locally but once uploaded to GitHub it just stops working, issue is related to shader line-breaks (`LF` vs `CR LF`) but I don't know why it only fails for the specific shaders used in this example.

raysan5:
> Are shaders necessary to use `DrawMeshInstanced()`? I was hoping to draw a large number of instances but I haven't got the handle of shaders yet.

@Steelsmasher yes, they are needed. Make sure they are properly loaded, that could be your issue on desktop.

ubkp:
@Steelsmasher When you run the `shaders_mesh_instancing` example, does the logs show any warnings or errors?

Steelsmasher:
When I run `shaders_mesh_instancing` from example folder with raylibs notepad++ everything runs fine. When I tried @raysan5's local compilation I got a blank screen as below:
![image](https://github.com/raysan5/raylib/assets/18646376/8682302e-d514-4f01-bfdf-2a6ee4aab0fe)

raysan5:
@Steelsmasher the log say that shaders can not be loaded, are you copying them in the right path?

ubkp:
I got that error too if I executed the example from some other dir other than inside `examples/shaders/`, e.g.:
```
~/test/raylib-master/examples$ ./shaders/shaders_mesh_instancing 
[...]
WARNING: FILEIO: [resources/shaders/glsl330/lighting_instancing.vs] Failed to open text file
[...]

PR #3316, opened 2023-09-15
Fix shaders path

ubkp:
Adds the `ApplicationDirectory` path to the shader files path.

Fixes https://github.com/raysan5/raylib/issues/3315.

Very likely also fixes _"REVIEW: Example: models_mesh_instancing. Not working properly on web, issues with the shader but ONLY when uploade to GitHub, it works as expected locally."_ from https://github.com/raysan5/raylib/discussions/2952.


raysan5:
@ubkp I'm afraid this is not the issue. If that was the issue, no shaders example would work, all follow the same loading pattern.

I just tried the example on desktop and web (locally) and it works as expected with latest raylib at this same moment.

Attaching my local compilation: [shaders_mesh_instancing.zip](https://github.com/raysan5/raylib/files/12620822/shaders_mesh_instancing.zip)

I vaguely remember I already faced this issue in the past but it's really weird, it works ok locally but once uploaded to GitHub it just stops working, issue is related to shader line-breaks (`LF` vs `CR LF`) but I don't know why it only fails for the specific shaders used in this example.

ubkp:
@raysan5 Yes, you're right. I've jumped into the first thing that appeared to be a solution. I should review these more carefully. My bad, sorry.

PR #3317, opened 2023-09-15
Update TurboRaylib bindings  => change version in bindings.md

turborium:
Update TurboRaylib bindings


raysan5:
@turborium thanks for the update!

PR #3321, opened 2023-09-17
Added UBSAN complaint fix to rLoadTexture #1891

Codom:
I managed to trigger the ubsan by simply calling `LoadRenderTexture(1280, 720);` from zig. Looking at recent PR's and issues it seems like this issue has already been mitigated elsewhere, but still exists in the codepath for LoadRenderTexture.

All I did was apply a similar fix (as seen in 30a9a24db9d2561335a2a42d423cf9cfeab6e769) for #1891 to the rLoadTextures function. This seems to fix the issue.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -3009,11 +3009,15 @@ unsigned int rlLoadTexture(const void *data, int width, int height, int format, 
 
         TRACELOGD("TEXTURE: Load mipmap level %i (%i x %i), size: %i, offset: %i", i, mipWidth, mipHeight, mipSize, mipOffset);
 
+        // NOTE: Added pointer math separately from function to avoid UBSAN complaining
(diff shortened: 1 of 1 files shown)


raysan5:
Please note that there is an error, `mipOffset += mipSize;` is called later in the loop, so `dataPtr += mipSize` should be added.

Codom:
Ah, missed that. I just recompute the offset from data on every iteration, `dataPtr = (unsigned char*)data + mipOffset` which is more in line with the previous version.

raysan5:
@Codom did you add the chnage? I can't see it.

Codom:
It should be here: f0e75112dc6072fcbca919618694f15038a1d203 
Apologies for any confusion.

raysan5:
@Codom I'm merging it but I'm reviewing this implementation, I'll take the variable definition out of the look and also the conditional check.

PR #3322, opened 2023-09-17
Update raylib-ocaml bindings to 4.5 version

tjammer:
Long overdue, but better late than never


raysan5:

PR #3325, opened 2023-09-18
Move mutex initialization before `ma_device_start()`

Bigfoot71:
This is the solution proposed by @mackron in response to the crash issue on Android after reopening the application, which was discussed on Discord and here: https://github.com/mackron/miniaudio/issues/733

The solution was to move the mutex initialization before `ma_device_start`.
I have tested it on my end, and it has successfully resolved the issue for me.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -474,22 +474,22 @@ void InitAudioDevice(void)
         return;
     }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Very nice! Thanks for reviewing this issue!

PR #3326, opened 2023-09-18
[core] debug `LoadVrStereoConfig()` so it accepts `RL_OPENGL_43` and `RL_OPENGL_21`

SuperUserNameMan:
I think there was a confusion about whether `rlGetVersion()` returns `RL_OPENGL_33` or `GRAPHICS_OPENGL_33` in the code of `LoadVrStereoConfig()` which was thus rejecting OpenGL4.3 and OpenGL2.1 with a `WARNING: RLGL: VR Simulator not supported on OpenGL 1.1` ... 

This patch solves it, and the `core_vr_simulator` example works with all version of OpenGL (other than 1.1 indeed)

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2602,7 +2602,7 @@ VrStereoConfig LoadVrStereoConfig(VrDeviceInfo device)
 {
     VrStereoConfig config = { 0 };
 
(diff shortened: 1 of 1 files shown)


raysan5:
@SuperUserNameMan Thanks for the review.

PR #3319, opened 2023-09-16
general syntax fixes

Brian-ED:
I don't know much C, but i found a spelling mistake and i tried making the code a bit prettier in raymath.c. I haven't tested everything, but it does compile and i don't think it needs testing since i didn't change any code, only spacing and comments.

PR #3330, opened 2023-09-20
[raymath] Added macros for EPSILON on each function it's used in

Brian-ED:
This is so the functions can be easily copied and used. This was discussed in a separate pull:
https://github.com/raysan5/raylib/pull/3319#issuecomment-1722461911

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -214,6 +214,10 @@ RMAPI float Wrap(float value, float min, float max)
 // Check whether two given floats are almost equal
 RMAPI int FloatEquals(float x, float y)
 {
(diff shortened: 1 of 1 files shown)


Brian-ED:
I have noticed a difference between my addition and the one recommended by ray, and that's that i used "ifndef" while he used "#if !defined(RAYMATH_EPSILON)".
I didn't notice it on pull, i'll fix it in an hour when i'm free

Brian-ED:
I am now done

raysan5:
@Brian-ED thanks for this review!

PR #3329, opened 2023-09-20
[split] Fix compilation for web (and desktop)

ubkp:
### PR changes
1. Fixes compilation for `PLATFORM_WEB` ([R12](https://github.com/raysan5/raylib/pull/3329/files#diff-7341b65fe444443a67b5c67d194f97dd0c919b63c579f15ce018de206cf84864R12)). `#include "GLFW/glfw3.h"` ([R15)](https://github.com/raysan5/raylib/pull/3329/files#diff-7341b65fe444443a67b5c67d194f97dd0c919b63c579f15ce018de206cf84864R15) is necessary for `CoreData`'s `GLFWwindow *handle;` ([R109](https://github.com/raysan5/raylib/pull/3329/files#diff-7341b65fe444443a67b5c67d194f97dd0c919b63c579f15ce018de206cf84864R109)) on `PLATFORM_WEB`.

2. Removes `EM_ASM_INT` from `core_input_gestures_web` example ([L311-L317](https://github.com/raysan5/raylib/pull/3329/files#diff-6bb5ee6018759da67168502ea865172f16f99f175c871a405024c1a219b85d87L311-L317)) as per https://github.com/raysan5/raylib/pull/3328#pullrequestreview-1635184774.

3. Fixes `undefined symbol` compilation errors from `raymath` for `PLATFORM_DESKTOP` and `PLATFORM_WEB`, by moving `#define RAYMATH_IMPLEMENTATION` from `rcore.c` ([L112](https://github.com/raysan5/raylib/pull/3329/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339L112)) to `rcore.h` ([R31](https://github.com/raysan5/raylib/pull/3329/files#diff-7341b65fe444443a67b5c67d194f97dd0c919b63c579f15ce018de206cf84864R31)).<br>
**Alternative solution for item 3:** https://github.com/raysan5/raylib/pull/3331/commits/b75e631084c8095ffdc03145a27221d746d37ea8 at https://github.com/raysan5/raylib/pull/3331

4. Fixes `TraceLog` on `PLATFORM_DESKTOP` and `PLATFORM_WEB` by adding `#define SUPPORT_TRACELOG` to `rcore.h` ([R10](https://github.com/raysan5/raylib/pull/3329/files#diff-7341b65fe444443a67b5c67d194f97dd0c919b63c579f15ce018de206cf84864R10)). Also fixes the `core_custom_logging` example.

[...]

PR #3331, opened 2023-09-20
[split] Including raymath.h in rcore.h is causing an issue

michaelfiber:
rcore.h is included before raymath.h in rcore.c and it doesn't specify `#define RAYMATH_IMPLEMENTATION`

One fix is to simply remove it from rcore.h


ubkp:
@michaelfiber Would this https://github.com/raysan5/raylib/pull/3329/commits/063c0af955f8657983512d236a5b858efaf8d676 be ok?

michaelfiber:
That would accomplish the same thing. And I just realized I am still committing to this branch so I'm going to close this request since it shouldn't be merged yet.  Whenever your fix is merged that works for me.

PR #3350, opened 2023-09-25
Factor's Raylib bindings mention being at v4.5 now.

WraithGlade:
I looked at the linked Factor binding page for Raylib and found that the file [summary.txt](https://github.com/factor/factor/blob/master/extra/raylib/summary.txt) has the following text in it:

```
Bindings for Raylib 4.5
```

The bindings have also been updated within the last week.

This implies that Factor's Raylib bindings are either already up to date with v4.5 or will be soon.

Thus, I have submitted this likely correction to Raylib's bindings table.

(PS: There are other Factor Raylib bindings on the internet that are less up-to-date, so don't confuse those with this.)


raysan5:
@WraithGlade Thanks for the update!

PR #3351, opened 2023-09-25
[rcore] reveiwed GetWorldToScreenEx

Brian-ED:
I finished a commented TODO, which was to use Vector3Transform instead of quaternions.

I also realized during review that it was using CORE.Window.screen.width in one place instead of the inputted width. The input width is definitely the intended value to be placed there since the convenience function GetWorldToScreen gives screen width as input.

I have tested this with a few different examples and the outputs of both old and new versions are very consistent.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2922,7 +2922,7 @@ Matrix GetCameraMatrix2D(Camera2D camera)
     //   2. Rotate and Scale
     //   3. Move by -target
     Matrix matOrigin = MatrixTranslate(-camera.target.x, -camera.target.y, 0.0f);
-    Matrix matRotation = MatrixRotateZ(-camera.rotation*DEG2RAD);
+    Matrix matRotation = MatrixRotate((Vector3){ 0.0f, 0.0f, 1.0f }, camera.rotation*DEG2RAD);
     Matrix matScale = MatrixScale(camera.zoom, camera.zoom, 1.0f);
     Matrix matTranslation = MatrixTranslate(camera.offset.x, camera.offset.y, 0.0f);
 
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2952,7 +2952,7 @@ Vector2 GetWorldToScreenEx(Vector3 position, Camera camera, int width, int heigh
     }
     else if (camera.projection == CAMERA_ORTHOGRAPHIC)
     {
-        float aspect = ((double)width/(double)height);
+        double aspect = ((double)width/(double)height);
         double top = camera.fovy/2.0;
         double right = top*aspect;
 
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2963,14 +2963,17 @@ Vector2 GetWorldToScreenEx(Vector3 position, Camera camera, int width, int heigh
     // Calculate view matrix from camera look at (and transpose it)
     Matrix matView = MatrixLookAt(camera.position, camera.target, camera.up);
 
+    // Convert world position vector to quaternion
+    Quaternion worldPos = { position.x, position.y, position.z, 1.0f };
+
     // Transform world position to view
-    Vector3 worldPos = Vector3Transform(position, matView);
+    worldPos = QuaternionTransform(worldPos, matView);
 
     // Transform result to projection (clip space position)
-    worldPos = Vector3Transform(worldPos, matProj);
+    worldPos = QuaternionTransform(worldPos, matProj);
 
     // Calculate normalized device coordinates (inverted y)
-    Vector3 ndcPos = { worldPos.x, -worldPos.y, worldPos.z };
+    Vector3 ndcPos = { worldPos.x/worldPos.w, -worldPos.y/worldPos.w, worldPos.z/worldPos.w };
 
     // Calculate 2d screen position vector
     Vector2 screenPosition = { (ndcPos.x + 1.0f)/2.0f*(float)width, (ndcPos.y + 1.0f)/2.0f*(float)height };
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2922,7 +2922,7 @@ Matrix GetCameraMatrix2D(Camera2D camera)
     //   2. Rotate and Scale
     //   3. Move by -target
     Matrix matOrigin = MatrixTranslate(-camera.target.x, -camera.target.y, 0.0f);
-    Matrix matRotation = MatrixRotate((Vector3){ 0.0f, 0.0f, 1.0f }, camera.rotation*DEG2RAD);
+    Matrix matRotation = MatrixRotateZ(-camera.rotation*DEG2RAD);
     Matrix matScale = MatrixScale(camera.zoom, camera.zoom, 1.0f);
     Matrix matTranslation = MatrixTranslate(camera.offset.x, camera.offset.y, 0.0f);
 
@@ -2952,7 +2952,7 @@ Vector2 GetWorldToScreenEx(Vector3 position, Camera camera, int width, int heigh
     }
     else if (camera.projection == CAMERA_ORTHOGRAPHIC)
     {
-        float aspect = (float)CORE.Window.screen.width/(float)CORE.Window.screen.height;
+        float aspect = ((double)width/(double)height);
         double top = camera.fovy/2.0;
         double right = top*aspect;
 
@@ -2963,19 +2963,14 @@ Vector2 GetWorldToScreenEx(Vector3 position, Camera camera, int width, int heigh
     // Calculate view matrix from camera look at (and transpose it)
     Matrix matView = MatrixLookAt(camera.position, camera.target, camera.up);
 
-    // TODO: Why not use Vector3Transform(Vector3 v, Matrix mat)?
-
-    // Convert world position vector to quaternion
-    Quaternion worldPos = { position.x, position.y, position.z, 1.0f };
-
     // Transform world position to view
-    worldPos = QuaternionTransform(worldPos, matView);
+    Vector3 worldPos = Vector3Transform(position, matView);
 
     // Transform result to projection (clip space position)
-    worldPos = QuaternionTransform(worldPos, matProj);
+    worldPos = Vector3Transform(worldPos, matProj);
 
     // Calculate normalized device coordinates (inverted y)
-    Vector3 ndcPos = { worldPos.x/worldPos.w, -worldPos.y/worldPos.w, worldPos.z/worldPos.w };
+    Vector3 ndcPos = { worldPos.x, -worldPos.y, worldPos.z };
 
     // Calculate 2d screen position vector
     Vector2 screenPosition = { (ndcPos.x + 1.0f)/2.0f*(float)width, (ndcPos.y + 1.0f)/2.0f*(float)height };


Brian-ED:
Found an inconsistency while doing some extra testing, give me a few more minutes

Brian-ED:
Yes, so i've tested with CAMERA_ORTHOGRAPHIC camera which was working flawlessly but now i've tested with CAMERA_PERSPECTIVE and the new version gives a different (presumably wrong answer).

Brian-ED:
Ok yeah i think this function used quaternion for good reasons. When the veiw is CAMERA_PERSPECTIVE the 4th component of the quaternion of worldpos is not 1 after second mutation, meaning that it's being used for good reason. 
but atleast the first quaternion transform can be turned into a vector3Transform. I'll see if i can make the code for that look nice

ubkp:
Could we please not do that? If anything we should be using more quaternions.
Reference: [Comparison_with_other_representations_of_rotations](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Comparison_with_other_representations_of_rotations), [Performance_comparisons](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Performance_comparisons).
IMHO, it's ok to use euler as function parameters, but internally we should be using quaternions.

Brian-ED:
in the current state of raylib, QuaternionTransform is basically equivalent to Vector3Transform interms of code and i assume performance since it's prettymuch the same code except Vector3Transform has less multiplications and additions to do. seems like a no-brainer. Imo you shouldn't use quaternions for the sake of quaternions if there's no benefit of them.

ubkp:
Performance is already a huge issue for something like `PLATFORM_WEB`. Anything we can improve (or keep, in this case) is very valuable. Here is a test:

```
#include <stdio.h>
#include <time.h>
#include "raymath.h"

#define MAXLOOP 1000000000

int main(void) {

    Vector3 vecResult;
    double vecTime = 0.0f;

    Quaternion quaResult;
    double quaTime = 0.0f;

    {
        float v[3] = { 0.0f };
        float m[16] = { 0.0f };
        clock_t timeStart = clock();
        for (int i=0; i<MAXLOOP; i++) {
            vecResult = Vector3Transform((Vector3){ v[0], v[1], v[2] }, (Matrix){ m[0], m[1], m[2], m[3], m[4], m[5], m[6], m[7], m[8], m[9], m[10], m[11], m[12], m[13], m[14], m[15] });
            int ii = i%3; v[ii]++;
            unsigned iii = i%12; m[iii]++;
        }
        clock_t timeEnd = clock();
        vecTime = (double)(timeEnd - timeStart) / CLOCKS_PER_SEC;
    }

    {
        clock_t timeStart = clock();
        float v[4] = { 0.0f };
        float m[16] = { 0.0f };
        for (int i=0; i<MAXLOOP; i++) {
            quaResult = QuaternionTransform((Quaternion){ v[0], v[1], v[2], v[3] }, (Matrix){ m[0], m[1], m[2], m[3], m[4], m[5], m[6], m[7], m[8], m[9], m[10], m[11], m[12], m[13], m[14], m[15] });
            int ii = i%4; v[ii]++;
            unsigned iii = i%12; m[iii]++;
        }
        clock_t timeEnd = clock();
        quaTime = (double)(timeEnd - timeStart) / CLOCKS_PER_SEC;
    }

    printf("Vec: %f sec (%f)\n", vecTime, vecResult.x);
    printf("Qua: %f sec (%f)\n", quaTime, quaResult.x);

    return 0;
}
```
These are my results for `PLATFORM_DESKTOP`:
```
~/test$ ./test 
Vec: 3.762668 sec (844424930131968.000000)
Qua: 3.054455 sec (1125899906842624.000000)
~/test$ ./test 
Vec: 3.766946 sec (844424930131968.000000)
Qua: 3.060343 sec (1125899906842624.000000)
~/test$ ./test 
Vec: 3.767291 sec (844424930131968.000000)
Qua: 3.058038 sec (1125899906842624.000000)
```

And these are my results for `PLATFORM_WEB`:
```
Vec: 13.185000 sec (844424930131968.000000)
Qua: 10.047000 sec (1125899906842624.000000)

Vec: 13.473000 sec (844424930131968.000000)
Qua: 10.223000 sec (1125899906842624.000000)

Vec: 13.441000 sec (844424930131968.000000)
Qua: 10.003000 sec (1125899906842624.000000)
```
So no, it's not the basically equivalent of Vector3Transform. And performance is not pretty much the same.

Brian-ED:
I'm confused. That result indicates that vec operation was waster. Did you miss the decimal point?

Brian-ED:
Running the same test, i can't replicate your result. Vec has been faster every run, on `PLATFORM_DESKTOP`

ubkp:
> I wonder how much compiler optimization and platform factor into these results.

@orcmid If you notice the code, you'll see that the possibility of substancial compiler optimization was removed from the equation by forcing a deterministic progression of the rotated variables and also calling the `r.x` result so the compiler could not skip/predict/omit any calls/symbols and also as a way to validate the results. Otherwise, the benchmark wouldn't be testing the transforms performance at all.

> I also don't see how `PLATFORM_DESKTOP` and `PLATFORM_WEB` are reflected in the running of that test file. Is `raymath.h` sensitive to that?

@orcmid The test reflects the difference between running both transforms on native code on native hardware with "direct" (through OS) access to hardware resources. Against running their wasm equivalent on the browser layer having "indirect" (through the browser) access to hardware resources. Thus why `PLATFORM_WEB` is so much slower than running anything on native, which is expected result. And also why performance is so much more important for `PLATFORM_WEB` than other platforms.

orcmid:
@ubkp I should have been more clear.  I was wondering how PLATFORM_DESKTOP and PLATFORM_WEB were specified, not why the differences are important.

However, I can tell you that even compiled as desktop applications, one can arrive at quite different results, and the preference can change depending on the compilation.

I used your code and compiled it twice using VC/C++ to compile an x64 .exe for running on windows.  My processor is an at least 12 years old  i7-3770 CPU @ 3.40GHz.

Here is the default compilation with no specification of anything about performance:
VecsOverQuads-default

I had near-identical results in a previous run to ensure I had it running properly.  It is painfully slow and Vec wins over Qua in about 4 seconds out of 69..

Here is the compilation with performance optimization.

VecsOverQuads-optimized

In this case the results are almost 28 TIMES faster, and Qua beats out Vec by about .5 seconds out of 2.5 seconds.

Both compilations were for x64.  In the optimized one, the parameters are

- `/O2` for maximum speed
- `/favor:blend` to optimize for a mix of x64 architectures
- `/fp:fast` for maximum floating-point speed, including use of special instructions, not fussing with intermediate rounding, etc.

I assume `raymath.h` defaulted to desktop.  I didn't check.  It certain ran on my desktop. 

What's interesting to me is the fact that the optimized .obj file is slightly smaller, as is the .exe, and the output results are identical.

There you have it.  As they like to say around her, YMMV.

ubkp:
@orcmid It's unfortunate that `MSVC` is not open source so we have no idea what it does, but, on `gcc` without `-O` or `-O0` these are the results for `PLATFORM_DESKTOP`:
```
~/test$ ./test 
Vec: 29.624024 sec (844424930131968.000000)
Qua: 26.406193 sec (1125899906842624.000000)
```
And with `-O2` (AFAIK, the most common one):
```
~/test$ ./test 
Vec: 3.771921 sec (844424930131968.000000)
Qua: 3.056916 sec (1125899906842624.000000)
```
This is fairly consistent and in line with quarternions well documented performance ([reference](https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation#Performance_comparisons)).

orcmid:
@ubkp So the gcc -O2 case is clearly beneficial and the approximately 20% improvement with Qua might not be negligible.

I assume you don't propose reading the gcc source code to know what it does. 

For me, [this description](https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=msvc-170) is helpful enough.

ubkp:
@orcmid Sure, we can **trust** them having properly implemented their compiler optimizations. After all, they have a pristine track record for QA (https://www.reddit.com/r/cpp/comments/ilzbqd/almost_30_year_old_optimization_bug_in_msvc/).

orcmid:
> @orcmid Sure, we can **trust** them having properly implemented their compiler optimizations. After all, they have a pristine track record for QA 

I'm not going down this rathole with you.  Let's stick with the situation at hand.

orcmid:
@ubkp 

I was curious what execution changes might be noticeable using recent-generation x64 chips.  I reran the code, default first, then recompiled optimization, using the same VS Build Tools version.  Here's how it works using an 11th Gen i7-11370H 3.30 GHz processor on a new laptop.

VecsOverQuads-optimizedLaptop

ubkp:
Those are very interesting results. Good to know the performance difference is even higher in modern CPUs with "O2". I'm glad this wasn't a complete waste of time, since also found MSVC without "optimizations" can't do basic math.

Brian-ED:
because I'm not knowledgeable enough with performance and c, I will revert the `quaternion→vector3transform` change.

Brian-ED:
in summery, this merge now has removed the TODO comment and fixed `(float)CORE.Window.screen.width` (and height) being used instead of the given input `(float)width` and height.

src/rcore.c
```diff
@@ -2952,7 +2952,7 @@ Vector2 GetWorldToScreenEx(Vector3 position, Camera camera, int width, int heigh
     }
     else if (camera.projection == CAMERA_ORTHOGRAPHIC)
     {
-        float aspect = (float)CORE.Window.screen.width/(float)CORE.Window.screen.height;
+        float aspect = ((double)width/(double)height);
```
ubkp, line 2955: Should be `double aspect = `.

Brian-ED: don't width and height need to be floats instead?

ubkp: If you're asking about the base `CORE.Window.screen` dimensions, they are `Size` ([L393](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L393)). `Size` is `unsigned int` ([L358](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L358)). AFAIK, this is correct. I'm unaware of window sizes with float point dimensions (e.g.: [reference](https://www.glfw.org/docs/3.3/window_guide.html#window_size)).

If you're asking about [R2955](https://github.com/raysan5/raylib/pull/3351/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R2955) specifically, ahead it will multiply `aspect` ([R2957](https://github.com/raysan5/raylib/pull/3351/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R2957)) by a `double` ([R2956](https://github.com/raysan5/raylib/pull/3351/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R2956)). So either keep as:
```
double aspect = ((double)width/(double)height);
```
Or keep as floats (and let [R2957](https://github.com/raysan5/raylib/pull/3351/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R2957) promote the result to `double`):
```
float aspect = ((float)width/(float)height);
```

Peter0x44:
@orcmid the "equivalent" optimizations you're talking about (the /fp:fast) is not done by gcc -O2, the gcc version is "-ffast-math" (also enabled by -Ofast)
These are disabled by default because they can break correct programs that depend on certain behaviors
like... reading errno (which some math.h functions do modify) 

https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Optimize-Options.html#index-ffast-math

/fp:fast is not a "free" performance win in all situations, hence why it is not on by default

orcmid:
@Peter0x44
> @orcmid the "equivalent" optimizations you're talking about (the /fp:fast) is not done by gcc -O2, 

I don't know where I said anything about "equivalent" optimizations so I don't understand this aspect of your reply.

I understand about the cases of `/fp:fast` sometimes being undesirable.  I was not promoting any of that.  I simply demonstrated the maximum performance I found in a specific case where the results of the @ubkp tests were identical but achieved much more economically.  And in the particular case, the performance of quaternions over vector3 flipped, demonstrating that there is more involved than blanket assumption of absolute preference.  

I was also very specific that I was using a specific release of VC/C++ and not making any statements whatsoever about gcc.  I have nothing to offer with respect to gcc and I have no idea what the compiler option are and what they signify.  The description of the tradeoffs that applied in my *specific* VC/C++ case is clear enough in the
[applicable documentation](https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=msvc-170).

My only point is to demonstrate that generalizations about performance based on source code or literal interpretation of mathematical expressions are not always (maybe rarely ever) sound.  There is a more profound demonstration of that on a different issue with regard to string-handling primitives and the value of using the standard library in contrast with hand-rolled operations.

raysan5:
Merging current tweaks.

ubkp:
@raysan5 Note: this change was on `rcore.c`, so it will likely generate a small conflict when `rcore_platform_split` gets merged.

raysan5:
@ubkp Aaah! You are right! Sorry, just read the comments and thought it could be closed.

PR #3353, opened 2023-09-26
Texture Tiling Example - luis605

luis605:
This is an example of texture tiling with 3d models!


raysan5:
@luis605 Thanks for the example, I'm merging it but it requires quite some additional work to integrate with raylib build systems. I'll try to do it whenever possible.

PR #3355, opened 2023-09-27
[rcamera] DLL Exports

JeffM2501:
The rcamera functions were not exported to the dll, and thus are not available to users of the DLL or bindings that must link dynamically.

This PR exports them the same as other raylib modules.


raysan5:

PR #3358, opened 2023-09-27
fix BINDINGS.md

thechampagne:


raysan5:
@thechampagne thanks for the fix!

PR #3361, opened 2023-09-29
Update BINDINGS.md with raylib-pkpy-bindings

blueloveTH:
I am going to use raylib to make my games with [pocketpy](https://github.com/blueloveth/pocketpy).
So I wrote a binding generator for it. See https://github.com/blueloveTH/pkpy-bindings.
It uses raylib's JSON format for describing the API.

The output files are based on `v4.6-dev` that located in:

+ [output/raylibw.cpp](https://github.com/blueloveTH/pkpy-bindings/blob/main/output/raylibw.cpp)
+ [output/typings/raylib.pyi](https://github.com/blueloveTH/pkpy-bindings/blob/main/output/typings/raylib.pyi)

There is also an [online version](https://pkpy-bindings.streamlit.app/).


raysan5:
@blueloveTH thanks for the addition.

PR #3363, opened 2023-10-02
Optimize m3d mesh creation

DaveH355:
Hello, I've changed the m3d loader to sort faces by material. The old code assumes the faces are already grouped together by material, which may not be the case for every model. 

This will reduce the number of meshes being created and uploaded to the gpu for those cases

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5623,7 +5623,7 @@ static Model LoadM3D(const char *fileName)
         // We always need a default material, so we add +1
         model.materialCount++;
 
-        // Sort faces by material.
+        // failsafe, model should already have faces grouped by material
         qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
 
         model.meshes = (Mesh *)RL_CALLOC(model.meshCount, sizeof(Mesh));
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5567,6 +5567,15 @@ static Model LoadVOX(const char *fileName)
 unsigned char *m3d_loaderhook(char *fn, unsigned int *len) { return LoadFileData((const char *)fn, (int *)len); }
 void m3d_freehook(void *data) { UnloadFileData((unsigned char *)data); }
 
+// Comparison function for qsort
+static int m3d_compare_faces(const void *a, const void *b)
+{
+  m3df_t *fa = (m3df_t *)a;
+  m3df_t *fb = (m3df_t *)b;
+
+  return (fa->materialid - fb->materialid);
+}
+
 // Load M3D mesh data
 static Model LoadM3D(const char *fileName)
 {
@@ -5614,19 +5623,8 @@ static Model LoadM3D(const char *fileName)
         // We always need a default material, so we add +1
         model.materialCount++;
 
-        // Sort faces by material (insertion)
-        for (i = 1; i < m3d->numface; i++)
-        {
-          m3df_t key = m3d->face[i];
-          j = i - 1;
-    
-          while (j >= 0 && m3d->face[j].materialid > key.materialid)
-          {
-            m3d->face[j+1] = m3d->face[j];
-            j = j - 1;
-          }
-          m3d->face[j+1] = key;
-        }
+        // Sort faces by material.
+        qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
 
         model.meshes = (Mesh *)RL_CALLOC(model.meshCount, sizeof(Mesh));
         model.meshMaterial = (int *)RL_CALLOC(model.meshCount, sizeof(int));
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5567,15 +5567,6 @@ static Model LoadVOX(const char *fileName)
 unsigned char *m3d_loaderhook(char *fn, unsigned int *len) { return LoadFileData((const char *)fn, (int *)len); }
 void m3d_freehook(void *data) { UnloadFileData((unsigned char *)data); }
 
-// Comparison function for qsort
-static int m3d_compare_faces(const void *a, const void *b)
-{
-  m3df_t *fa = (m3df_t *)a;
-  m3df_t *fb = (m3df_t *)b;
-
-  return (fa->materialid - fb->materialid);
-}
-
 // Load M3D mesh data
 static Model LoadM3D(const char *fileName)
 {
@@ -5623,8 +5614,19 @@ static Model LoadM3D(const char *fileName)
         // We always need a default material, so we add +1
         model.materialCount++;
 
-        // Sort faces by material.
-        qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
+        // Sort faces by material (insertion)
+        for (i = 1; i < m3d->numface; i++)
+        {
+          m3df_t key = m3d->face[i];
+          j = i - 1;
+    
+          while (j >= 0 && m3d->face[j].materialid > key.materialid)
+          {
+            m3d->face[j+1] = m3d->face[j];
+            j = j - 1;
+          }
+          m3d->face[j+1] = key;
+        }
 
         model.meshes = (Mesh *)RL_CALLOC(model.meshCount, sizeof(Mesh));
         model.meshMaterial = (int *)RL_CALLOC(model.meshCount, sizeof(int));
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5567,6 +5567,15 @@ static Model LoadVOX(const char *fileName)
 unsigned char *m3d_loaderhook(char *fn, unsigned int *len) { return LoadFileData((const char *)fn, (int *)len); }
 void m3d_freehook(void *data) { UnloadFileData((unsigned char *)data); }
 
+// Comparison function for qsort
+static int m3d_compare_faces(const void *a, const void *b)
+{
+  m3df_t *fa = (m3df_t *)a;
+  m3df_t *fb = (m3df_t *)b;
+
+  return (fa->materialid - fb->materialid);
+}
+
 // Load M3D mesh data
 static Model LoadM3D(const char *fileName)
 {
@@ -5613,6 +5622,9 @@ static Model LoadM3D(const char *fileName)
 
         // We always need a default material, so we add +1
         model.materialCount++;
+
+        // Sort faces by material.
+        qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
 
         model.meshes = (Mesh *)RL_CALLOC(model.meshCount, sizeof(Mesh));
         model.meshMaterial = (int *)RL_CALLOC(model.meshCount, sizeof(int));


src/rmodels.c
```diff
@@ -5567,6 +5567,15 @@ static Model LoadVOX(const char *fileName)
 unsigned char *m3d_loaderhook(char *fn, unsigned int *len) { return LoadFileData((const char *)fn, (int *)len); }
 void m3d_freehook(void *data) { UnloadFileData((unsigned char *)data); }
 
+// Comparison function for qsort
+static int m3d_compare_faces(const void *a, const void *b)
```
raysan5, line 5571: I try to make functions as much self-contained as possible, avoiding dependencies to other functions, could it be implemented directly inside the function?

src/rmodels.c
```diff
@@ -5614,6 +5623,9 @@ static Model LoadM3D(const char *fileName)
         // We always need a default material, so we add +1
         model.materialCount++;
 
+        // Sort faces by material.
+        qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
```
raysan5, line 5627: I'd prefer to avoid using `qsort()` function, could the logic be implemented directly here?

DaveH355: Sure, but I think it is already the most simple way

raysan5: It could be but it adds a function-dependency that I prefer to avoid.

DaveH355: Ok, I've changed the sorting method to insertion

orcmid: @raysan5 I don't understand the objection to a `stlib.h` dependency.

Is it possible for an user to create a properly-sorted materials array and then deliver that to raylib in a straightforward manner, if that is what it will take?

orcmid:
I reviewed the insertion sort and confirmed that it is correct.

It has the unfortunate characteristics of having a worst-case on the order of `numface` squared compare-and-move operations and that might be messy if `numface` can be large.  The best case is for already-sorted, the worst case is for initially in reverse order.

Although this might work well under likely conditions, I have a small concern about the added burden on understandability and maintenance risks, in comparison with using a standard-library function.

src/rmodels.c
raysan5: @orcmid There was a point the raylib development that I just tried to [minimize libc dependency](https://github.com/raysan5/raylib/wiki/raylib-libc-dependency) for raylib, it was a personal decision. I agree it probably has not much sense.

bztsrc:
> The old code assumes the faces are already grouped together by material, which may not be the case for every model.

Yes, because according to the [M3D specification](https://gitlab.com/bztsrc/model3d/-/blob/master/docs/m3d_format.md#polygon-data-records) they must be stored like that in the file. The overall M3D policy here is, use as much of the time consuming operations as possible in m3d_save, and keep m3d_load fast.

(FYI: the SDK explicitly [calls qsort](https://gitlab.com/bztsrc/model3d/-/blob/master/m3d.h#L4935) the same way like your patch, and Blender's internal representation stores only one material per face, so in the [Blender plugin](https://gitlab.com/bztsrc/model3d/-/blob/master/blender/io_scene_m3d.py#L783) all polygons are already grouped together by material. In both cases the model file is saved with a mesh grouped by material.)

Considering that M3D files could be saved by other (non-SDK and non-Blender plugin) applications as well and if the mesh is already sorted (as it should be) then this adds just a very-very small overhead to the model load time, so all in all I think it's safe to do this in raylib. Making sure that the mesh is sorted won't hurt, but might be useful in some special cases with buggy model files. 

It is not necessary for model files that comply with the M3D specification though.

> I try to make functions as much self-contained as possible, avoiding dependencies to other functions, could it be implemented directly inside the function?

I'm afraid that's not possible, `qsort` explicitly requires a comparator function as an argument.

Cheers,
bzt

bztsrc:
> Although this might work well under likely conditions

No, this insertion sort patch has a terrible performance and ***must not be used*** (O(numface factorial) in best case).

It adds a significant overhead even when the mesh is already sorted, this is the worst possible solution! Using qsort is much-much better and uncomparably more efficient (in both best case and worst case), so having a comparator function might not be nice, but definitely worth it.

But as I've said, sorting the mesh isn't required with models that comply with the M3D specification in the first place. If this patch is going to be added, I'd like to ask to add a comment as well that states this is just a workaround, not required for valid M3D model files.

```diff
-  // Sort faces by material.
+ // just a workaround for buggy models, NOT required as valid M3D model files must store faces sorted
+ qsort(m3d->face, m3d->numface, sizeof(m3df_t), m3d_compare_faces);
```

Cheers,
bzt

DaveH355:
@bztsrc 
> Blender's internal representation stores only one material per face, so in the [Blender plugin](https://gitlab.com/bztsrc/model3d/-/blob/master/blender/io_scene_m3d.py#L783) all polygons are already grouped together by material.

It is true Blender only allows 1 material per face. But there is no guarantee Blender stores the faces by order of material. The order can be changed by creating new faces, subdividing, triangulating etc.

The [Blender plugin](https://gitlab.com/bztsrc/model3d/-/blob/master/blender/io_scene_m3d.py#L783) doesn't sort the faces either. It just checks if the material index of a face is greater than number of materials. 
Perhaps it is a bug in the M3D Blender addon?

I've also tested this to make sure. Here is a plane with 2 materials assigned to some faces. 
screenshot

Exporting this using the [addon](https://gitlab.com/bztsrc/model3d/-/blob/master/blender/io_scene_m3d.py?ref_type=heads) and loading it using Raylib from master branch gives
`Model has 16 meshes and 3 materials`
If Raylib sorts the faces by material. Either insertion or quicksort 
`Model has 2 meshes and 3 materials`

The results are the same for Blender 3.1 and 3.6
Here is the blend file used
[test_plane.zip](https://github.com/raysan5/raylib/files/12789526/test_plane.zip)

bztsrc:
> Perhaps it is a bug in the M3D Blender addon?

Indeed! Thanks for the heads up! Right now I'm very busy with another project, but I'll fix the Blender plugin as soon as I can.

Cheers,
bzt

DaveH355:
I've reverted back to qsort()
I think it is a good failsafe with little overhead

orcmid:
@bztsrc @DaveH355
> > Although this might work well under likely conditions
> 
> No, this insertion sort patch has a terrible performance and _**must not be used**_ (O(numface factorial) in best case).

Don't be so hasty Dave.  The insertion sort you wrote works very quickly when the data is already sorted.  It is essentially a linear check in that case. If there are a few items out of order, it also won't be terrible.

The worst case for the insertion comes with the items in reverse sequence.  But the cost is not n!, it is more like n(n+1)/2 (i.e., O(n^2)).  This is determined by the worst number of shifts of items in the array to make the hole for the next one to be moved earlier in what is already sorted.  (@bztsrc This is an addition of cases, not a product).

bztsrc:
Let's forget about the materialid for a moment (because that varies in every model) assuming that part of the expression is always false, and just focus on the loops to determine the worst case. (I wrote "best case" above, but that was wrong, I meant "worst case", sorry about that.)
```c
        for (i = 1; i < m3d->numface; i++)
        {
          m3df_t key = m3d->face[i];
          j = i - 1;

          while (j >= 0 && m3d->face[j].materialid > key.materialid)
          {
            m3d->face[j+1] = m3d->face[j];
            j = j - 1;
          }
          m3d->face[j+1] = key;
        }
```
The outter `for` runs `n` times. The inner `while` on the first iteration runs 1 time, on the second iteration runs 2 times, on third iteration 3 times, etc. That gives a total of `n!` factorial runs on the inner loop's block.

> This is an addition of cases, not a product

Not so. Two nested loops gives a product, you also admitted that by writing O(n^2) (which is O(n*n) and *not* O(n+n), so it is a product, not an addition).

Anyway, this doesn't matter anymore, because I've added the sort to the Blender plugin too (see [latest commit](https://gitlab.com/bztsrc/model3d/-/commit/fda8eb2b02dfb180bb9f3123490ed4a25b246469)). Needs some more testing to be sure, but it's already there and it is to stay.

@raysan5: you can safely remove the sort from raylib now (or if you decide to keep it to be on the safe side, then please add a comment `// just a workaround, sorting not needed for valid M3D model files`, thanks!).

Cheers,
bzt

orcmid:
For those playing along at home, there are two things to observer here.  One is algorithmic performance analysis.  The second is actual performance in practice. I will address the second first.

2. We know in the already-sorted case that the insertion-sort method on _n_ items will make _n_-1 comparisons and conclude without altering anything.  In this case we have no idea how much work qsort will do with that case.  We also don't know what it will take in the worst-case for the insertion sort.

1. The worst case for the insertion case is when the original _n_ values are in descending order.    Let's see what happens.  Let's keep it simple.  Let's use {4, 3, 2, 1, 0}, for 5 values.
a.  The idea in insertion sort is to walk through the list from first to last, with the idea that the list to the left of a current position is already sorted and we then take the next item and insert it where it belongs (shifting some out of the way to do it).  We only walk through the list once, each time moving an item into the sorting group only if it is otherwise out of order.
b.   We start with the second item (list[1]) because the single item at list[0] is already sorted.  If the item of interest is not less than the adjacent item from the already-sorted list, then there is nothing to do.  (The algorithm might pick it up and put it back - no harm, no foul.)
c. Otherwise, we keep comparing backwards through the already-sorted until we bump into an item that is not larger than the one "in hand" -- or we reach the end.
d. To put the item to insert in the sorted group, we need to move the items that are larger than the one we have further down in the list (this will occupy where we got the current item),
making a hole in front of them where we can drop in the item.

e. Here is the procedure with the example.

i. {4, ...} is fine without looking farther.

ii. With {4, 3, ...} it's seen that 3 is out of order.  Pick it up and slide the larger of the already-sorted larger ones down the list to make a hole, and drop the 3 in it.  So we are now at {3,4,...} and can continue.  For this to happen there was one compare and one shift of a value.  We'll call that 1 step.

iii. With {3, 4, 2, ...} the same thing happens but the 2 is compared with 2 values and 2 values are shifted to arrive at {2, 3, 4, ...}.  That's 2 compares and movement of 2 items.  So 2 steps.

iv. With {2, 3, 4, 1, ...}  The rules get us to {1, 2, 3, 4, ...} and there were 3 steps.

v. Finally, with {1, 2, 3, 4, 0} we take 4 steps and arrive at {0, 1, 2, 3, 4}.

What's the pattern?

The *worst-case* number of steps taken for _n_ items is the sum

  1 + 2 +  ..., + (_n_-1).

And that sum is the value _n_*(_n_-1).  And we speak of that as a complexity of order _n_*_n_ or O(_n_^2), because that's what dominates the growth as _n_ is increases.

That's the algorithmic analysis complexity.  We haven't addressed the comparative cost of different approaches.  The key thing to notice for the insertion sort is that if the list is already sorted, there are just n-1 simple comparisons and no movement.

orcmid:
@raysan5 @bztsrc @DaveH355 

**BREAKING NEWS:** The worst-case time for Quicksort is when the items are already sorted.  

That statement is made by Donald Knuth in remarks following Algorithm 4.2.2Q (Quicksort) in "The Art of Computer Programming," Volume 3, Sorting and Searching.  

Knuth's observation is from algorithmic analysis.  It does not tell us what the comparative performance between an insertion sort and a Quicksort in actual code might be, and when (as the number of items increases) does the better performance change from one to the other.

However, since we know that the input to the m3d loader is almost always already sorted, with perhaps a few out-of-order, a defensive approach would be to use an insertion sort that exits very quickly if the requirement for an already-sorted list is satisfied.  Quicksort is not the answer for that case.

In the spirit of defensive programming, I have developed an improvement to the previous insertion sort that is blazingly fast when the items are already sorted.  It doesn't monkey with the list at all except when it finds an out-of-order item.

Adopting the @ubkp playbook, I have created a stress test, `TrySorts.c` (using simple arrays of integers) to identify where the performance of qsort surpasses that of my sped-up insertion sort.  The code is at [orcmid/myCleanC](https://github.com/orcmid/myCleanC/tree/main/inserter).

I have run TrySorts (built using VC/C++) on two systems.  Both of them show the crossover where qsort becomes faster to be around 250 items.

Here's the run on my 12-year-old desktop.

TrySorts-2023-10-06-2145-XPS8500

Here's the run on my new laptop.

TrySorts-2023-10-06-2151-Laptop

ubkp:
@orcmid That's pretty cool. These benchmarks are awesome.

Since you're on Windows, would like to suggest also testing with mingw (that's packed with raylib [distribution](https://raysan5.itch.io/raylib)), since I guess that's what a lot of users use.

orcmid:
@ubkp
> Since you're on Windows, would like to suggest also testing with mingw (that's packed with raylib [distribution](https://raysan5.itch.io/raylib)), since I guess that's what a lot of users use.

That's probably the case for many raylib users.  I'd welcome seeing the results that anyone obtains with different compilers/platforms.  I'm not prepared to work with the mingw setup myself.

ubkp:
@orcmid Luckly that's super easy for Windows users: https://www.youtube.com/watch?v=R1iN3j7mAos

raysan5:
I'm merging this change BUT commenting it. The specs already specify data MUST be sorted.

PR #3370, opened 2023-10-03
Adds DrawModelPro() and DrawModelWiresPro(), accepting Matrixes as input

dbrats:
This allows us to easily draw the same model with multiple transformations, in particular rotation around multiple axises without having to change the attributes of the model itself (which the Yaw, Pitch, Roll example does).

This is particularly useful when loading one model and drawing it multiple times, such as in a game where you fight multiple enemies who all use the same model, but move and behave independently. Correct me if I'm wrong, but setting the model.transform will affect every instance of this model being drawn. Though this can be solved by specifically set that every time before drawing the model, i.e. per enemy instance, but I find a DrawModelPro to be both neater and more elegant, and in line with the other DrawXPro functions.

I didn't add the example to the ./examples/README.md file, but I'm open to doing that.

Considering I'm actually drawing just a cube, but using the DrawModelPro function, I figured it belongs in ./examples/models/.

I'm grateful for any and all feedback, especially since C isn't where I usually reside.


raysan5:
@dbrats I'm afraid I'm not merging this change, it implies adding two extra functions and same functionality can be easely accomplished at user side with current functions, just updating the `model.transform` as required and restoring it.

PR #3379, opened 2023-10-05
[parser] Update `raylib_api.*`

blueloveTH:
Meta files in `parser/output` are outdated. I have regenerated them based on the latest commit.


raysan5:
@blueloveTH thanks for the update!

PR #3380, opened 2023-10-06
Add more color functions

anstropleuton:
A function to invert a color, and a function to make a color grayscale would be a nice edition
I noticed that ImageColorInvert and ImageColorGrayscale exists, and I thought of making a color version of it wouldn't hurt


raysan5:
I prefer to avoid adding new functions if I clear use case is not provided.

Issue #3378, opened 2023-10-05
[rcore] Fullscreen without mode change


M374LX:
### Issue description

Problems with an incorrect resolution when toggling fullscreen mode have been reported in #3099. However, according to the GLFW documentation at https://www.glfw.org/docs/latest/window_guide.html#window_windowed_full_screen, "If the closest match for the desired video mode is the current one, the video mode will not be changed, making window creation faster and application switching much smoother".

The same effect can be achieved in SDL2 by calling ``SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN_DESKTOP);`` (with ``SDL_WINDOW_FULLSCREEN_DESKTOP`` instead of ``SDL_WINDOW_FULLSCREEN``).

With a borderless window, as implemented in #3216, there is still a problem: the window is movable (noticeably on a window manager that allows moving a window while holding a key, like dwm from https://suckless.org).

Fullscreen toggling without a video mode change can be implemented in GLFW like so:

```
const GLFWvidmode* mode = glfwGetVideoMode(monitor);

glfwSetWindowMonitor(window, monitor, 0, 0, mode->width, mode->height, mode->refreshRate);
```

As we can see, we need to pass the refresh rate along with the width and height of the monitor to ``glfwSetWindowMonitor()``.

In https://github.com/raysan5/raylib/blob/bc15c19518968878b68bbfe8eac3fe4297f11770/src/rcore.c#L1222C37-L1222C37, the refresh rate is not passed, but I don't think this should be changed, as it is possible that a programmer using raylib actually wants a video mode change and the monitor does not support the same refresh rate on a different video mode.

Some of the possible approaches to implement fullscreen toggling without a video mode change include:
1. Use FLAG_FULLSCREEN_MODE and ToggleFullscreen(), but pass the refresh rate to glfwSetWindowMonitor() when the screen size corresponds to monitor but GLFW_DONT_CARE otherwise.
2. Create a new flag, like FLAG_FULLSCREEN_DESKTOP_MODE, and a new function, like ToggleDesktopFullscreen(). This one could automatically use the monitor size and free the programmer from the need to call ``SetWindowSize()`` before toggling fullscreen mode.

### Environment

Linux

M374LX:
@ubkp Not yet tested extensively, but it worked well on a quick test with a single monitor in which the refresh rate was passed to ``glfwSetWindowMonitor()`` instead of ``GLFW_DONT_CARE``.

[...]

Issue #3198, opened 2023-07-25
[core] TOPMOST flag does not work with multiple monitors


Vessov:
### Issue description

I use Python API for Raylib - [Raylib 4.5.0.0](https://pypi.org/project/raylib/). When setting up a window, if different than default monitor is selected via `set_window_monitor()`, the `FLAG_WINDOW_TOPMOST` is not working as intended -  even when flag is set, window will minimize when clicked anywhere on any other monitor - regardless of if the flag was set before setting the window monitor, or after. This occurs always when using the `set_window_monitor()`, even when the default monitor is selected.

### Environment

 + Windows 10 Pro 22H2 64 bit
 + Intel(R) HD Graphics 520
 + OpenGL 4.6
 + Python 3.11.3
 + Python Raylib 4.5.0.0

### Code Example

```python
import pyray as pr

pr.set_config_flags(pr.ConfigFlags.FLAG_WINDOW_TOPMOST)
pr.set_config_flags(pr.ConfigFlags.FLAG_WINDOW_UNDECORATED)
pr.init_window(500, 500, "FooBar")
pr.set_window_position(0, 0)
pr.set_target_fps(60)

monitor = 0
pr.set_window_size(pr.get_monitor_width(monitor), pr.get_monitor_height(monitor))
pr.set_window_monitor(monitor)

while not pr.window_should_close():
    pr.begin_drawing()
    pr.clear_background(pr.RAYWHITE)
    pr.end_drawing()

pr.close_window()
```

raysan5:
@Vessov Please, try using `SetWindowState()` after `InitWindow()`, some flags are only functional after the window has been created. Here it is an usage example: https://github.com/raysan5/raylib/blob/master/examples/core/core_window_flags.c

Vessov:
Still, error persists - even after adding `SetWindowState()` after `InitWindow()`, or even inside the event loop

Vessov:
@ubkp @raysan5  I see. I wonder if there may be any workaround - since when window monitor isn't set via `SetWindowMonitor` (so, if I understand correctly, it defaults to `NULL`?), and window starts on the defaulft monitor, the issue doesn't occur - no matter if in fullscreen mode or otherwise, the window stays topmost.

[...]

PR #3382, opened 2023-10-06
Add FLAG_FULLSCREEN_DESKTOP

M374LX:
This PR implements the third approach suggested in #3378.

**Update:** not ready to be merged because it has the same problems as https://github.com/raysan5/raylib/pull/3391#issuecomment-1752211850.

### Code example ###
```C
#include <raylib.h>

int main()
{
	SetConfigFlags(FLAG_FULLSCREEN_DESKTOP);
	InitWindow(800, 600, "Test: fullscreen without video mode change");

		if (IsKeyPressed(KEY_F)) {
			ToggleFullscreen();
		}

		BeginDrawing();
		ClearBackground(GRAY);
		EndDrawing();
	}

	CloseWindow();

	return 0;
}

```

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -528,7 +528,8 @@ typedef enum {
     FLAG_WINDOW_MOUSE_PASSTHROUGH = 0x00004000, // Set to support mouse passthrough, only supported when FLAG_WINDOW_UNDECORATED
     FLAG_BORDERLESS_WINDOWED_MODE = 0x00008000, // Set to run program in borderless windowed mode
     FLAG_MSAA_4X_HINT       = 0x00000020,   // Set to try enabling MSAA 4X
-    FLAG_INTERLACED_HINT    = 0x00010000    // Set to try enabling interlaced video format (for V3D)
+    FLAG_INTERLACED_HINT    = 0x00010000,   // Set to try enabling interlaced video format (for V3D)
+    FLAG_FULLSCREEN_DESKTOP = 0x00020000    // Set to prevent a video mode change when toggling fullscreen mode
 } ConfigFlags;
 
 // Trace log level
--- a/src/rcore_desktop.c
+++ b/src/rcore_desktop.c
@@ -358,10 +358,27 @@ void ToggleFullscreen(void)
         }
         else
         {
+            int width = CORE.Window.screen.width;
+            int height = CORE.Window.screen.height;
+            int refreshRate = GLFW_DONT_CARE;
+
+            CORE.Window.previousScreen.width = width;
+            CORE.Window.previousScreen.height = height;
+
+            if (IsWindowState(FLAG_FULLSCREEN_DESKTOP))
+            {
+                width = GetMonitorWidth(monitorIndex);
+                height = GetMonitorHeight(monitorIndex);
+                refreshRate = GetMonitorRefreshRate(monitorIndex);
+
+                CORE.Window.screen.width = width;
+                CORE.Window.screen.height = height;
+            }
+
             CORE.Window.fullscreen = true;
             CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
 
-            glfwSetWindowMonitor(platform.handle, monitor, 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
+            glfwSetWindowMonitor(platform.handle, monitor, 0, 0, width, height, refreshRate);
         }
 
     }
@@ -370,7 +387,7 @@ void ToggleFullscreen(void)
         CORE.Window.fullscreen = false;
         CORE.Window.flags &= ~FLAG_FULLSCREEN_MODE;
 
-        glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
+        glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y, CORE.Window.previousScreen.width, CORE.Window.previousScreen.height, GLFW_DONT_CARE);
     }
 
     // Try to enable GPU V-Sync, so frames are limited to screen refresh rate (60Hz -> 60 FPS)
@@ -495,6 +512,11 @@ void SetWindowState(unsigned int flags)
         CORE.Window.flags |= FLAG_VSYNC_HINT;
     }
 
+    if (((CORE.Window.flags & FLAG_FULLSCREEN_DESKTOP) != (flags & FLAG_FULLSCREEN_DESKTOP)) && ((flags & FLAG_FULLSCREEN_DESKTOP) > 0))
+    {
+        CORE.Window.flags |= FLAG_FULLSCREEN_DESKTOP;
+    }
+
     // State change: FLAG_BORDERLESS_WINDOWED_MODE
     // NOTE: This must be handled before FLAG_FULLSCREEN_MODE because ToggleBorderlessWindowed() needs to get some fullscreen values if fullscreen is running
     if (((CORE.Window.flags & FLAG_BORDERLESS_WINDOWED_MODE) != (flags & FLAG_BORDERLESS_WINDOWED_MODE)) && ((flags & FLAG_BORDERLESS_WINDOWED_MODE) > 0))
@@ -608,6 +630,11 @@ void ClearWindowState(unsigned int flags)
     {
         glfwSwapInterval(0);
         CORE.Window.flags &= ~FLAG_VSYNC_HINT;
+    }
+
+    if (((CORE.Window.flags & FLAG_FULLSCREEN_DESKTOP) > 0) && ((flags & FLAG_FULLSCREEN_DESKTOP) > 0))
+    {
+        CORE.Window.flags &= ~FLAG_FULLSCREEN_DESKTOP;
     }
 
     // State change: FLAG_BORDERLESS_WINDOWED_MODE


blueloveTH:
I think this feature is useful.

raysan5:
@ubkp You were working on this feature, please could you take a look?

ubkp:
@raysan5 Sure thing. Tested (62d3066), some considerations:

1. When alt-tabbing, the window gets minimized, as the regular fullscreen.

2. Breaks toggling out of regular fullscreen using `ToggleFullscreen` if `FLAG_FULLSCREEN_DESKTOP` is not being used.

3. Should probably reuse `Size previousScreen;` ([L391](https://github.com/raysan5/raylib/pull/3382/commits/62d3066ad7e098da05ec79f281cfa80e27a0fd55#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339L391)) instead of `Size restoreSize;` ([R397](https://github.com/raysan5/raylib/pull/3382/commits/62d3066ad7e098da05ec79f281cfa80e27a0fd55#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R397))

4. IMHO, the same could be achieved with:
```
SetWindowSize(GetMonitorWidth(0), GetMonitorHeight(0));
ToggleFullscreen();
```

5. I had hopes this could replace `BORDERLESS_WINDOWED_MODE`, but, unfortunately, IMHO, it doesn't.

**Environment:** tested for `PLATFORM_DESKTOP` on `Gnome 44.0 (Wayland)`, `KDE Plasma 5.27.8 (Wayland and X11)` and `MATE 1.26.2 (X11)` on Linux `Fedora 38 64-bit`.

raysan5:
@ubkp thank you very much for taking a look and the fast response!

1. Yes, that's the expected behaviour.
2. Mmmh... `ToggleFullscreen()` in this situation shouldn't probably do anything...
3. Yes, agree.
4. Ok, having another way to accomplish it shouldn't be a problem...
5. Mmmh... I see, I though it was similar to `BORDERLESS_WINDOWED_MODE`?

ubkp:
@raysan5 No problem. :+1:

1 - Yes, but is expected for an actual fullscreen, not for a "desktop fullscreen". This was one of the main problems that `BORDERLESS_WINDOWED_MODE` aimed to solve (reference: [#3198](https://github.com/raysan5/raylib/issues/3198)).

2 - I mean, if I compile 62d3066 and use it on an example that just doesn't use `FLAG_FULLSCREEN_DESKTOP` at all, calling `ToggleFullscreen` while on fullscreen won't leave fullscreen, and it should (and it will) on current master branch.

4 - **Personal opinion:** IMHO that's a lot of change for something that can be done with 2 calls on the user side.

M374LX:
e47bb70 replaces ``restoreSize`` with ``previousScreen`` and also fixes the breakage of leaving regular fullscreen.

ubkp:
Sorry, but:

4. This adds something that can already be easily achieved with 2 calls;
5. It changes stable working fullscreen code;
6. It does not implement a real "desktop fullscreen" like `SDL`;
7. I don't  understand what's the issue with `GLFW_DONT_CARE` ([doc](https://www.glfw.org/docs/latest/window_guide.html#window_windowed_full_screen)) since:
_"GLFW_REFRESH_RATE specifies the desired refresh rate for full screen windows. A value of GLFW_DONT_CARE means the highest available refresh rate will be used. This hint is ignored for windowed mode windows."_
8. In summary, I have voiced my concerns, I want nothing to do with this change.

M374LX:
Actually, these two calls do not work as expected:
```C
SetWindowSize(GetMonitorWidth(0), GetMonitorHeight(0));
ToggleFullscreen();
```

The reason is because ``SetWindowSize()`` simply calls ``glfwSetWindowSize()`` without updating ``CORE.Window.screen.width`` and ``CORE.Window.screen.height``. This causes ``ToggleFulscreen()`` to actually use the previous window size.

The wrong video mode problem when entering fullscreen mode can be solved by doing:
```C
if (IsKeyPressed(KEY_F)) {
	SetWindowSize(GetMonitorWidth(0), GetMonitorHeight(0));
	fullscreen_toggled = true;
}

BeginDrawing();
EndDrawing();

if (fullscreen_toggled) {
	ToggleFullscreen();
	fullscreen_toggled = false;
}
```

However, this still causes a video mode change, which involves the annoying screen blackout for about a second.

ubkp:
Then `SetWindowSize` should set `CORE.Window.screen.*` not the other way around.
Also how `GLFW_DONT_CARE` won't resolve to the same frequency already being used?

M374LX:
For some reason, the mode change still happens in my system when using ``GLFW_DONT_CARE``, but not when the current refresh rate is passed.

If everyone agrees, we can implement the first approach suggested in #3378, in which the screen size is checked and, if it is the same as the monitor, the refresh rate is passed to``glfwSetWindowMonitor()``instead of ``GLFW_DONT_CARE``. In this case, there would be no need to add ``FLAG_FULLSCREEN_DESKTOP``.

M374LX:
Also note that the code example 2 in https://github.com/raysan5/raylib/pull/3382#issuecomment-1751846459 does not automatically restore the window size when leaving fullscreen mode. In this case, the responsibility of storing and then restoring the window size is left to the programmer, not done automatically by raylib.

ubkp:
The only thing that should be fixed at this point is `SetWindowSize` to update `CORE`.

ubkp:
Any fullscreen call will cause the window to be minimized if alt-tabbing. `GLFW` is not `SDL`.

M374LX:
> The only thing that should be fixed at this point is `SetWindowSize` to update `CORE`.

This and also the undesirable video mode change when the screen (or window) size is set to the same as the monitor.

M374LX:
I have done some tests with the first approach suggested in #3378, but found a considerable increase in the complexity in the program's code due to it needing to manage the window restore size (which would not be managed by raylib) and also that a mode changed happened when toggling between fullscreen and windowed mode multiple times too quickly.

Here is an example of the increased code complexity:

```C
#include <raylib.h>
#include <stdio.h>
#include <stdbool.h>

int main()
{
	int restore_width = 0;
	int restore_height = 0;
	bool fullscreen_toggled = false;
	bool leave_fullscreen = false;
	char str[32];

	InitWindow(800, 600, "Test: fullscreen without video mode change");

		if (IsKeyPressed(KEY_F)) {
			fullscreen_toggled = true;

			if (!IsWindowFullscreen()) {
				//Change to fullscreen
				restore_width = GetScreenWidth();
				restore_height = GetScreenHeight();
				SetWindowSize(GetMonitorWidth(GetCurrentMonitor()), GetMonitorHeight(GetCurrentMonitor()));
			} else {
				leave_fullscreen = true;
			}
		}

		sprintf(str, "Screen size: %dx%d %c", GetScreenWidth(), GetScreenHeight(), IsWindowFullscreen() ? 'F' : 'W');

		BeginDrawing();
		ClearBackground(GRAY);
		EndDrawing();

		if (fullscreen_toggled) {
			fullscreen_toggled = false;

			ToggleFullscreen();

			if (leave_fullscreen) {
				SetWindowSize(restore_width, restore_height);
				leave_fullscreen = false;
			}
		}
	}

	CloseWindow();

	return 0;
}
```

This would be simpler, but did not work as expected (probably related to ``SetWindowSize()`` not updating ``CORE``):
```C
#include <raylib.h>
#include <stdio.h>
#include <stdbool.h>

int main()
{
	int restore_width = 0;
	int restore_height = 0;
	bool fullscreen_toggled = false;
	char str[32];

	InitWindow(800, 600, "Test: fullscreen without video mode change");

		if (IsKeyPressed(KEY_F)) {
			fullscreen_toggled = true;

			if (!IsWindowFullscreen()) {
				//Change to fullscreen
				restore_width = GetScreenWidth();
				restore_height = GetScreenHeight();
				SetWindowSize(GetMonitorWidth(GetCurrentMonitor()), GetMonitorHeight(GetCurrentMonitor()));
			} else {
				SetWindowSize(restore_width, restore_height);
			}
		}

		sprintf(str, "Screen size: %dx%d %c", GetScreenWidth(), GetScreenHeight(), IsWindowFullscreen() ? 'F' : 'W');

		BeginDrawing();
		ClearBackground(GRAY);
		EndDrawing();

		if (fullscreen_toggled) {
			fullscreen_toggled = false;
			ToggleFullscreen();
		}
	}

	CloseWindow();

	return 0;
}
```

M374LX:
Marking it as ready for review, as it seems to work fine on two monitors.

M374LX:
This still does not work as expected:
```C
SetConfigFlags(FLAG_FULLSCREEN_DESKTOP | FLAG_FULLSCREEN_MODE);
```

Should raylib handle this case or can we just instruct the programmer to create the window initially without ``FLAG_FULLSCREEN_MODE`` and change to fullscreen later?

ubkp:
Repeating myself: _the only thing that should be fixed is `SetWindowSize` to update `CORE`._

@raysan5 I'm sorry, but I'm no longer wasting time reviewing or testing this.

M374LX:
#3391 proposes a fix to ``SetWindowSize()`` not updating ``CORE``.

M374LX:
> For some reason, the mode change still happens in my system when using GLFW_DONT_CARE, but not when the current refresh rate is passed.

Turns out that the refresh rate on my system was not set to the highest available. By changing to the highest available refresh rate, a mode change did not happen with ``GLFW_DONT_CARE``.

raysan5:
@M374LX I'm not implementing this change `FLAG_FULLSCREEN_DESKTOP`, as @ubkp commented current implementation looks solid enough for most situations and this PR looks to me as an overengineering effort to solve some very side cases, adding extra complexity to raylib.

PR #3385, opened 2023-10-07
Ensure m3d faces in non-decreasing materialid sequence

orcmid:
This modification replaces the expensive qsort protection with an insertion sort that is near-instantaneous in the expected ordered case.

The changes desk-check and they compile.  It is not confirmed in a running application and I am guessing that I have satisfied raylib coding style.

PS: This is modeled on Donald Knuth's Algorithm 5.2.1S (Straight insertion sort) and is a refinement of that and the code @DaveH355 provided earlier in order to handle mostly-ordered faces as quickly as possible.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5614,19 +5614,26 @@ static Model LoadM3D(const char *fileName)
         // We always need a default material, so we add +1
         model.materialCount++;
 
-        // WARNING: Sorting is not needed, valid M3D model files should already be sorted
-        // faces should already by grouped by material
-        // Just keeping the sorting function for reference (Check PR #3363)
-        /*
-        static int m3d_compare_faces(const void *a, const void *b)
+        // Faces must be in non-decreasing materialid order
+        // Verify that quickly, sorting them otherwise. 
+        for (i = 1; i < m3d->numface; i++)
         {
-            m3df_t *fa = (m3df_t *)a;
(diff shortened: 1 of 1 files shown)


orcmid:
I should add that this defensive approach is consistent with the steps taken a few lines below where this modification is inserted:

```C
 // there should be only one material switch per material kind, but be bulletproof for non-optimal model files
```

raysan5:
@orcmid Thank you very much for the review but as stated in the M3D specs, this sorting is not needed any more. I prefer to avoid extra code for non-allowed side cases. In any case I can replace the qsort() statement with this code but it should remain commented, only for reference.

orcmid:
I think most folks are very reluctant to maintain a patched raylib, and only a small subset will even read this, so commented-out defenses don't seem particularly helpful to me.

I'm not going to argue.  It's your project.

raysan5:
As commented, I'm merging it for reference but keeping it commented. It is not required for as per the official M3D specs.

raysan5:
Side note, raylib does not use the `do { } while` approach anywhere in the library.

orcmid:
@raysan5 
> Side note, raylib does not use the `do { } while` approach anywhere in the library.

In this particular case it works perfectly with rapid checking for already-sorted elements.   I use this so rarely I had to look it up to ensure I was using it correctly :).

PR #3391, opened 2023-10-08
Window size and fullscreen fixes

### LINKED ISSUE #3390: [core] Fullscreen toggling and resolution
### Issue description

A program using raylib might run in both windowed and fullscreen mode. For example, the window size might be 800x600 in windowed mode but the program might want use the current monitor resolution when in fullscreen.

This does not work, however:

```C
if (IsKeyPressed(KEY_F)) {
	int monitor = GetCurrentMonitor();
	SetWindowSize(GetMonitorWidth(monitor), GetMonitorHeight(monitor));
	ToggleFullscreen();
}
```

The reason is because ``SetWindowSize()`` simply calls ``glfwSetWindowSize()`` without updating ``CORE.Window.screen.width`` and ``CORE.Window.screen.height``. This causes ``ToggleFulscreen()`` to actually use the previous window size.

This is a possible workaround:

```C
if (IsKeyPressed(KEY_F)) {
	int monitor = GetCurrentMonitor();
	SetWindowSize(GetMonitorWidth(monitor), GetMonitorHeight(monitor));
	fullscreen_toggled = true;
}


BeginDrawing();
EndDrawing();

if (fullscreen_toggled) {
	ToggleFullscreen();
	fullscreen_toggled = false;
}
```

The examples above show how to toggle from windowed mode to fullscreen. When doing the opposite (changing to windowed mode), this also does not work:

```C
ToggleFullscreen();
SetWindowSize(800, 600);
```

### Environment

Linux

### Code Example

The same code examples given in https://github.com/raysan5/raylib/pull/3382#issuecomment-1751846459 apply to this issue.

M374LX:
Fixes #3390.

It also implements the first approach suggested in #3378.

### Code example ###

Note that, when changing to fullscreen, ``SetWindowSize()`` needs to be called before ``ToggleFullscreen()``, while the order needs to be reversed when changing to windowed mode.

Additionally, is does not automatically restore the previous window position when changing to windowed mode.

```C
#include <raylib.h>
#include <stdio.h>
#include <stdbool.h>

int main()
{
	char str[32];

	InitWindow(800, 600, "Test");

		if (IsKeyPressed(KEY_F)) {
			if (!IsWindowFullscreen()) {
				int monitor = GetCurrentMonitor();

				SetWindowSize(GetMonitorWidth(monitor), GetMonitorHeight(monitor));
				ToggleFullscreen();
			} else {
				ToggleFullscreen();
				SetWindowSize(800, 600);
			}
		}

		sprintf(str, "Screen: %dx%d %c (monitor: %d)",
				GetScreenWidth(), GetScreenHeight(),
				IsWindowFullscreen() ? 'F' : 'W',
				GetCurrentMonitor());

		BeginDrawing();
		ClearBackground(GRAY);
		EndDrawing();
	}

	CloseWindow();

	return 0;
}

```

### DIFF
--- a/src/rcore_desktop.c
+++ b/src/rcore_desktop.c
@@ -854,6 +854,7 @@ void SetWindowMaxSize(int width, int height)
 void SetWindowSize(int width, int height)
 {
     glfwSetWindowSize(platform.handle, width, height);
+    WindowSizeCallback(platform.handle, width, height);
 }
 
 // Set window opacity, value opacity is between 0.0 and 1.0
--- a/src/rcore_desktop.c
+++ b/src/rcore_desktop.c
@@ -362,6 +362,7 @@ void ToggleFullscreen(void)
             int monitorHeight = GetMonitorHeight(monitorIndex);
             int refreshRate = GLFW_DONT_CARE;
 
+            //Prevent GLFW from changing the video mode if the screen is the same size as the monitor
             if (CORE.Window.screen.width == monitorWidth && CORE.Window.screen.height == monitorHeight)
             {
                 refreshRate = GetMonitorRefreshRate(refreshRate);
--- a/src/rcore_desktop.c
+++ b/src/rcore_desktop.c
@@ -358,10 +358,19 @@ void ToggleFullscreen(void)
         }
         else
         {
+            int monitorWidth = GetMonitorWidth(monitorIndex);
+            int monitorHeight = GetMonitorHeight(monitorIndex);
+            int refreshRate = GLFW_DONT_CARE;
+
+            if (CORE.Window.screen.width == monitorWidth && CORE.Window.screen.height == monitorHeight)
+            {
+                refreshRate = GetMonitorRefreshRate(refreshRate);
+            }
+
             CORE.Window.fullscreen = true;
             CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
 
-            glfwSetWindowMonitor(platform.handle, monitor, 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, GLFW_DONT_CARE);
+            glfwSetWindowMonitor(platform.handle, monitor, 0, 0, CORE.Window.screen.width, CORE.Window.screen.height, refreshRate);
         }
 
     }


M374LX:
Problem found: when using two monitors of different sizes, placing the window partly in one monitor and partly in the other and then changing to fullscreen, if the window is placed just slightly in the left monitor and mostly in the right one, it will become fullscreen in the right monitor but detect the left monitor as the current one. Consequently, the size of the wrong monitor will be the screen size.

ubkp:
> Consequently, the size of the wrong monitor will be the screen size.

IMHO, it could be much worst than just that. If it uses the wrong monitor frequency, that could cause some serious problems/issues/damage. That's why `GLFW_DONT_CARE` appears to be a far safer option.

M374LX:
The following code example seems to fix the problem of the window becoming fullscreen in the wrong monitor.

It turns out that trying to move the window to a different monitor with ``glfwSetWindowPos()`` (and consequently with ``SetWindowPosition()``) does not work and the window will always end up in the monitor in which most of it already is (at least on dwm, the window manager on which I am testing it). I cannot test it on Openbox at the moment because the PC on which I have it installed is experiencing a hardware fault (turning off randomly). This can be tested by pressing P.

Once again, we need to toggle fullscreen after ``EndDrawing()``.

We try to position the window at the top-left corner of the current monitor and, after ``EndDrawing()``, we check if the current monitor changed (the monitor detected as "current" by ``GetCurrentMonitor()`` is the one in which the leftmost part of the window is).

With it apparently working, should we still use ``GLFW_DONT_CARE`` instead of current refresh rate when the window size is the same as the monitor?

```C
#include <raylib.h>
#include <stdio.h>
#include <stdbool.h>

int main()
{
	char str[32];

	InitWindow(800, 600, "Test");

		bool fullscreen_toggled = false;

		if (IsKeyPressed(KEY_F)) {
			int monitor = GetCurrentMonitor();
			Vector2 pos = GetMonitorPosition(monitor);

			SetWindowPosition((int)pos.x, (int)pos.y);

			fullscreen_toggled = true;
		}

		if (IsKeyPressed(KEY_P)) {
			//Try to position the window in the leftmost monitor
			SetWindowPosition(0, 0);
		}

		BeginDrawing();
		ClearBackground(GRAY);

		sprintf(str, "Screen: %dx%d %c (monitor: %d)",
				GetScreenWidth(), GetScreenHeight(),
				IsWindowFullscreen() ? 'F' : 'W',
				GetCurrentMonitor());
		DrawText(str, 32, 64, 32, BLACK);

		Vector2 winpos = GetWindowPosition();
		sprintf(str, "Window pos: %dx%d", (int)winpos.x, (int)winpos.y);
		DrawText(str, 32, 96, 32, BLACK);

		EndDrawing();

		if (fullscreen_toggled) {
			if (!IsWindowFullscreen()) {
				int monitor = GetCurrentMonitor();

				SetWindowSize(GetMonitorWidth(monitor), GetMonitorHeight(monitor));
				ToggleFullscreen();
			} else {
				ToggleFullscreen();
				SetWindowSize(800, 600);
			}
		}
	}

	CloseWindow();

	return 0;
}
```

M374LX:
It looks like an issue with ``glfwSetWindowMonitor()``, but raylib could still mitigate it.

Another weird fact (on dwm, at least) is that then the window is partly in the left monitor but mostly in the right one, with ``GetCurrentMonitor()`` detecting the left one as current, a mode change takes place on the left monitor but the window is placed in the right one.

M374LX:
> Problem found: when using two monitors of different sizes, placing the window partly in one monitor and partly in the other and then changing to fullscreen, if the window is placed just slightly in the left monitor and mostly in the right one, it will become fullscreen in the right monitor but detect the left monitor as the current one. Consequently, the size of the wrong monitor will be the screen size.

In fact, borderless windows are also affected by this problem.

raysan5:
I'm closing this draft, I can't see the usefulness of it.

PR #3392, opened 2023-10-09
Call nsvgDeleteRasterizer() on created rasterizer by nsvgCreateRasterizer().

masoudd:
the `NSVGrasterizer *rast` needs to be passed to nsvgDeleteRasterizer() when we are done with it.


raysan5:
@masoudd Good catch! Thanks for the improvement!

PR #3394, opened 2023-10-09
Hotfix for Vector2LineAngle(), should probably be reviewed along with…

Murlocohol:
… the rest of raylib angle functions to determine what coordinate system we want.


raysan5:
@Murlocohol thanks for the review!

PR #3393, opened 2023-10-09
Update zig build system to zig version 0.11.0

purple4pur:
Discard `build.zig` and use `src/build.zig` only, bc by default these two output in different directories.

build command: cd to `src` then `zig build`.

Close #3386

Issue #3386, opened 2023-10-08
[build] Cannot build examples with zig 0.11.0, build.zig needs to be updated


purple4pur:
### Issue description

I followed instructions in `examples` folder and built with command `zig build core`, but so many errors occurred.

### Environment

Windows 10 LTSC x64, raylib `4.5.0`, zig `0.11.0`

![image](https://github.com/raysan5/raylib/assets/49893724/6ec83906-b781-44a3-9465-bc6d398cbe33)

### Issue Screenshot

I tried some fixing and was stuck here...

![image](https://github.com/raysan5/raylib/assets/49893724/0388c0a5-ef08-4a9e-a8f5-cfefd4441dc9)
![image](https://github.com/raysan5/raylib/assets/49893724/595b4ca1-5e57-43a8-8ff7-1716a0750da8)
![image](https://github.com/raysan5/raylib/assets/49893724/189d2193-7be8-49f9-ba26-67699e2c43e8)
![image](https://github.com/raysan5/raylib/assets/49893724/7c0a7a6a-c5ea-4d9a-86c2-587867f58dfe)

raysan5:
@purple4pur I'm afraid I don't maintain Zig build system, fee free to send a PR with required fixes.

star-tek-mb:
Please use either zig master with raylib master, either find commit which will compile with zig 0.11.0.

raylib 4.5.0 compatible with zig 0.10.1 i think. But zig 0.10.1 has a lot of issues with C ABI compatibility.

PR #3395, opened 2023-10-10
fix zig syntax errors (0.11.0) in examples, and make it install executables correctly

purple4pur:
Related: #3386 #3393

Much appreciated @Not-Nik @star-tek-mb !

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,7 +1,7 @@
 const std = @import("std");
 const raylib = @import("src/build.zig");
 
-// This has been tested to work with zig master branch as of commit 87de821 or May 14 2023
+// This has been tested to work with zig 0.11.0 (67709b6, Aug 4 2023)
 pub fn build(b: *std.Build) void {
     raylib.build(b);
 }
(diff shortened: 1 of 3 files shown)


raysan5:
@Not-Nik @star-tek-mb please, could you review this PR?

star-tek-mb:
I don't know honestly.. I think we should either revert all recent changes. Either make a complex change with one top-level build.zig file. Because building from different directories - breaks everything.

I tried to build from top level and from src and from examples directory. It's hacking. I'm only using top level build.zig for package manager. And I'm not using examples and object files from src. So, we need better build design for using raylib with Zig, and using Zig toolchain to build .c files with raylib.

I can look at it, but on weekends.

Not-Nik:
This PR only updates the example build file to Zig 0.11. We can discuss further design changes elsewhere.

Not-Nik (approved):

raysan5:
@purple4pur @Not-Nik thanks for the review!

PR #3397, opened 2023-10-10
Split drm update input

michaelfiber:
- Remove platform specific code from GetGamepadAxisCount to facilitate moving that function back to rcore.c

### DIFF
--- a/src/rcore_drm.c
+++ b/src/rcore_drm.c
@@ -1,6 +1,6 @@
 /**********************************************************************************************
 *
-*   rcore_drm - Functions to manage window, graphics device and inputs 
+*   rcore_drm - Functions to manage window, graphics device and inputs
 *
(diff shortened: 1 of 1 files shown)


raysan5:
@michaelfiber Current implementation looks great to me. I doubt many users change gamepads between sessions. Let me know when it's ready for me. I'll take care of moving those functions to `rcore.c`.

michaelfiber:
> @michaelfiber Current implementation looks great to me. I doubt many users change gamepads between sessions. Let me know when it's ready for me. I'll take care of moving those functions to `rcore.c`.

Will do, I was able to make these small changes real quick but won't be able to test for a bit because it turns out something keeps PLATFORM_DRM from working on my main workstation. Once I can test these changes I'm going to take a look at that.

PR #3403, opened 2023-10-13
Ported to stb_image_resize2.h

BabakSamimi:
A simple port to the new stb_image_resize. I tried out the `textures_drawing_image.c` example on Windows 11 and it seems to work. 

I'm not sure if removing the "deprecated" `stb_image_resize.h` is the correct move, as some developers might be using it through Raylib, thoughts?


raysan5:
@BabakSamimi Nice! Thanks for the update! 👍😄

PR #3404, opened 2023-10-13
Fix `GetMouseDelta()` issue for Android

Bigfoot71:
As mentioned in the [discussion](https://github.com/raysan5/raylib/issues/3313#issuecomment-1760386046) regarding the split of rcore.c by platform, here is the fix for the issue with `Vector2 GetMouseDelta(void)` on Android.

This now allows obtaining the delta for the first touch on Android. Tests were conducted with multitouch, and there were no problems or interference reported.


raysan5:
@Bigfoot71 Nice! Great improvement!

PR #3405, opened 2023-10-13
Fix `UpdateSound` parameter name

KislyjKisel:
The parameter `sampleCount` is multiplied by `ma_get_bytes_per_frame` meaning it is actually the number of frames, not samples. `ma_get_bytes_per_frame(format, channels)` is `ma_get_bytes_per_sample(format) * channels`.


raysan5:
@KislyjKisel thanks for the review!

PR #3408, opened 2023-10-13
Added function SetRandomSeedWithTime

amnesiac0x7C5:
This function seeds the random number generator using the time since epoch, allowing for a unique seed from within the library.


raysan5:
@amnesiac0x7C5 I'm afraid there is no need for that function, you can send time directly to the parameter.

amnesiac0x7C5:
I see now, Thank you.

PR #3410, opened 2023-10-13
Fix GetFPS so it doesn't hold on last time from previous window

maksut:
Defect: GetFPS holds on to the last time value from previous windows.

In this example, second window displays the last FPS value from the first window. Until it lives longer than the first:
```
#include "raylib.h"

int main(void)
{
    InitWindow(800, 450, "first window");

    // Keep this window open N seconds then close it
    while (!WindowShouldClose())
    {
      BeginDrawing();
      ClearBackground(RAYWHITE);
      DrawFPS(10, 10);
      EndDrawing();
    }

    CloseWindow();

    InitWindow(800, 450, "second window");

    // Notice that DrawFPS is stuck on the last fps value from the first window
    // only after N seconds it starts drawing the correct fps
    while(!WindowShouldClose())
    {
        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawFPS(10, 10);
        EndDrawing();
    }

    CloseWindow();
    return 0;
}
```

The proposed fix relies on the fact that GetTime must return incremental (or same) values.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1235,9 +1235,14 @@ int GetFPS(void)
 
     if (fpsFrame == 0) return 0;
 
-    if ((GetTime() - last) > FPS_STEP)
+    float newTime = (float)GetTime();
+
+    if (last > newTime)
+        last = newTime; // Last value is from a previous window so reset it
+
+    if ((newTime - last) > FPS_STEP)
     {
-        last = (float)GetTime();
+        last = newTime;
         index = (index + 1)%FPS_CAPTURE_FRAMES_COUNT;
         average -= history[index];
         history[index] = fpsFrame/FPS_CAPTURE_FRAMES_COUNT;


raysan5:
@maksut This seems to be related to a quite unexpected use case of raylib (two windows creation, one after another) but in any case, shouldn't the time measure be reseted on every `InitWindow()` instead of hacking the `GetFPS()` function?

maksut:
I was trying to avoid introducing a global variable just for this "defect". Because it is a silly issue. So feel free to reject/close the PR.

FYI; I am playing with raylib & clojure where I try to livecode as much as possible. So far it has been fun. But whenever I close the window and reopen it, drawFPS gets stuck to the old value for some time. But it is not a big deal. I can just write my own getFPS, drawFPS etc.

raylib is fun! Thanks for your work.

maksut:
While I was implementing my own GetFPS, I thought maybe it doesn't have to call GetTime at all.
Do you think last commit is reasonable? It solves my little problem and it _might_ be more efficient for everybody else.

JeffM2501:
I'm not 100% sure this is the correct fix. The core problem is that static data is used and it doesn't get restored to the proper state when the window is reset.
The history should be cleared back to the same state as first run.
I've added PR #3445 that fully clears the static state on the first frame.

maksut:
@JeffM2501 thanks for sparing some time for this.

raysan5:
@maksut thanks for pointing the issue and the proposed review, finally @JeffM2501 approach was merged.

Issue #3337, opened 2023-09-21
[models] Adding color to drawMeshInstanced


khornel:
### Issue description

GPU instancing doesn't seem super useful if you can't have some variation in each instance drawn, as you rarely just want a bunch of **exactly** the same thing on screen. I propose adding an optional color array parameter to the drawMeshInstanced function. The user can then decide whether they want to use it for color information or something else, like texture animation or texture atlas offsets.

### Code Example

I tried just hacking it in rmodels.c and it worked perfectly.

```C++
instanceColors = (float4*)RL_MALLOC(instances * sizeof(float4));

for (int i = 0; i < instances; i++) instanceColors[i] = ColorToFloatV(colors[i]);

rlEnableVertexArray(mesh.vaoId);

instancesVboId = rlLoadVertexBuffer(instanceColors, instances * sizeof(float4), false);

rlEnableVertexAttribute(material.shader.locs[SHADER_LOC_COLOR_DIFFUSE]);
rlSetVertexAttribute(material.shader.locs[SHADER_LOC_COLOR_DIFFUSE], 4, RL_FLOAT, 0, sizeof(Vector4), 0);
rlSetVertexAttributeDivisor(material.shader.locs[SHADER_LOC_COLOR_DIFFUSE], 1);

rlDisableVertexBuffer();
rlDisableVertexArray();
```

raysan5:
@khornel This is a breaking change that I'd prefer to avoid, could you provide some use cases? also, beside instance colors, could be additional attributes required?

raysan5:
> I ended up changing it so you supply an array of floats and a stride, to avoid having floats clamped and discretized when converting to color data. This also gives the user way more flexibility in what type of data to supply. Maybe it could be a DrawMeshInstancedEx call to avoid breaking compatibility and sticking to the existing convention.

I think this is a very good alternative and it sounds quite generic and flexible. Feel free to send a PR.

raysan5:
It seems this new feature was not a priority but just for a very specific use case. Just closing this issue.

PR #3411, opened 2023-10-13
Allow instanced rendering without a custom shader

TibboddiT:
This PR enables the use of `DrawMeshInstanced` without necessitating a custom shader.

It seems to me that this is inline with 

> raylib is SIMPLE and EASY-TO-USE

A secondary default shader has been introduced. This shader includes a new attribute that stores instance transformation and is loaded and managed exactly like the initial shader. It is then used in `DrawMeshInstanced`.

To avoid any breaking change, if the material's linked shader is not the defaut shader, it will be used in place of the default instanced rendering shader, preserving previous behavior.

My use case: I need to draw a few million cubes for a simulation project, and wanted to try instanced rendering capabilities of raylib. I was disappointed when I realised that a custom shader was needed, but impressed enough by the project to eventually send a PR :)

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -116,6 +116,7 @@
 #define RL_DEFAULT_SHADER_ATTRIB_NAME_COLOR        "vertexColor"       // Bound by default to shader location: 3
 #define RL_DEFAULT_SHADER_ATTRIB_NAME_TANGENT      "vertexTangent"     // Bound by default to shader location: 4
 #define RL_DEFAULT_SHADER_ATTRIB_NAME_TEXCOORD2    "vertexTexCoord2"   // Bound by default to shader location: 5
+#define RL_DEFAULT_SHADER_ATTRIB_NAME_INSTANCE     "instanceTransform" // Bound by default to shader location: 6
 
 #define RL_DEFAULT_SHADER_UNIFORM_NAME_MVP         "mvp"               // model-view-projection matrix
 #define RL_DEFAULT_SHADER_UNIFORM_NAME_VIEW        "matView"           // view matrix
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -371,6 +371,7 @@ typedef struct MaterialMap {
 // Material, includes shader and maps
 typedef struct Material {
     Shader shader;          // Material shader
+    Shader shaderInstanced; // Material shader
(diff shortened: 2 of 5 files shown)


ubkp:
Out of curiosity, what's the issue of using a custom shader to warrant this change?

TibboddiT:
I just needed to draw the same mesh in a lot of different positions. I admit that writing a shader is not very complex, but not having to write one makes raylib even easier to use! Also, I noticed a lot of raylib examples on the web avoid using custom shaders, which leads me to the conclusion that "beginners" prefer not to have to think about it.

ubkp:
1. The problem is that it makes it easier by having the library to manage that for the user. On that direction, the library, trying to make everything easier, ends up having to manage everything, becoming an engine, instead of what it is, a library of a set of tools.

2. If the users are having difficulty with shaders, then I don't think this would help solving that. IMHO, it would only make them more dependent on the library implementing something for them.

3. Since there's already support for custom shaders, IMHO, I don't think this should be added.

TibboddiT:
OK, I understand.

`DrawMesh` worked immediately, and is using a default shader. After having read [raylib-default-shader](https://github.com/raysan5/raylib/wiki/raylib-default-shader) and [raylib-generic-uber-shader-and-custom-shaders](https://github.com/raysan5/raylib/wiki/raylib-generic-uber-shader-and-custom-shaders), I thought making `DrawMeshInstanced` as easy to use as `DrawMesh` was a good idea.

Thanks for your time, and for the library. Feel free to close.

raysan5:
@TibboddiT Thanks for the review, I'm afraid I agree with @ubkp points, actually I prefer to keep the instancing shader out of raylib, I think it's a feature most users do not need to be provided by raylib and advance users won't have a problem implementing it on user code.

Also note that users requiring instancing, could also require more data than only the `transforms` (as stated on https://github.com/raysan5/raylib/issues/3337), this feature could grow a lot and never fit exactly what an advance user needs, so I think it's better to keep it at user side (as much as possible).

Murtagy:
I agree with the author. For some unknown reasons DrawMesh works fine, but DrawMeshInstanced does not. It is absolutely annoying to see nothing at all on the screen and seeing meshes with default shaders and knowning that a problem is likely in a shader would be way better

PR #3415, opened 2023-10-14
Fix screen size check in `InitPlatform()`

Bigfoot71:
Fixes a comparison error between two undefined variables (width, height) and also the definition of `CORE.Window.currentFbo.height`


raysan5:
@Bigfoot71 thanks for testing and the fix!

PR #3414, opened 2023-10-14
Implement SetMouseCursor for PLATFORM_WEB

BeardedBread:
Use EM_ASM to set the cursor element in CSS.  
CSS cursor reference: [link](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor)

Fix #3368

Build tested on Linux Mint 21.2, kernel 5.15.0-86-generic x86_64.  *May need help test build on Windows and MacOS.*

Test Program (compiles for both DESKTOP and WEB):
```c
#include "raylib.h"
#include <stdio.h>

#if defined(PLATFORM_WEB)
#include <emscripten/emscripten.h>
#endif

// Window size
static const int width = 800;
static const int height = 600;

// Grid sizes
const float grid_width = 200.0f;
const float grid_height = 200.0f;
const int grid_w = 4;

// Display name for cursor
static const char* CURSOR_NAMES[11] = 
{
    "default", "arrow", "text", "crosshair",
    "pointer", "ew-resize", "ns-resize", "nwse-resize",
    "nesw-resize", "move", "not-allowed"
};

void update_step(void)
{
    // Set the cursor based on mouse position in the grid
    Vector2 mouse_pos = GetMousePosition();
    int mouse_idx = (int)(mouse_pos.x / grid_width) + (int)(mouse_pos.y / grid_height) * grid_w;
    mouse_idx = (mouse_idx < 0)  ? 0: mouse_idx;
    mouse_idx = (mouse_idx > 10) ? 10: mouse_idx;
    MouseCursor cursor = (MouseCursor)mouse_idx;
    SetMouseCursor(cursor);

    static char buffer[64];
    snprintf(buffer, 64, "x: %.2f, y: %.2f, idx: %d", mouse_pos.x, mouse_pos.y, mouse_idx);
    buffer[63] = '\0';

    BeginDrawing();
        ClearBackground(LIGHTGRAY);
        for (int i = 0; i < grid_width; ++i)
        {
            DrawLine(grid_width*i, 0, grid_width*i, height, BLACK);
        }
        for (int i = 0; i < grid_height; ++i)
        {
            DrawLine(0, grid_height*i, width, grid_height*i, BLACK);
        }
        for (int i = 0; i < 11; i++)
        {
            int x = (i % grid_w) * grid_width + grid_width / 2;
            int y = (i / grid_w) * grid_height + grid_height / 2;
            DrawText(CURSOR_NAMES[i], x, y, 12, BLACK);
        }
[...]

Issue #3366, opened 2023-10-02
[text] `TextFormat()` silently fails when the formatted text is longer than `MAX_TEXT_BUFFER_LENGTH`


danielchasehooper:
### Issue description

`TextFormat()` silently truncates when the formatted text is longer than MAX_TEXT_BUFFER_LENGTH. I spent a long time thinking there was an issue elsewhere but TextFormat was just truncating the result and not aborting or logging a message.

### Environment

macOS

### Code Example

```
char *someReallyLongString; // must be longer than MAX_TEXT_BUFFER_LENGTH
char *result = TextFormat("File contents:\n%s", someReallyLongString);
```

ubkp:
@danielchasehooper I could be wrong, but I believe data validation is up to the user on raylib (reference https://github.com/raysan5/raylib/issues/3246#issuecomment-1682256790).

danielchasehooper:
It's not a data validation issue. the parameters sending to the function were valid. The function doesn't do what it says it will do for certain inputs, and doesn't warn you when thats happening.

ubkp:
I don't disagree, but, IMHO, you still have to control what you're passing to the functions. If you look at other the library functions, you'll see very few of them have complete/full parameter validation.

raysan5:
> I could be wrong, but I believe data validation is up to the user on raylib

@ubkp Yes, you are right, raylib does not consider all the possible (wrong?) use case scenarios, it would imply lots of functions review and increase in code complexity. 

That was the original approach of the library that actually fit very well on the original education context to warn the students about writing valid code.

Still, in the past years many checks and validations have been added to the library, specially for memory-critic scenarios. I think in this specific case it could log some warning... but for this specific function the warning will be logged 60 times per second, that's something that I also avoid in raylib.

An easy solution I can think of is appending a `"TRUNCATED"` (or similar) to the input string, so the users will be aware that something happened in there.

ubkp:
> An easy solution I can think of is appending a `"TRUNCATED"` (or similar) to the input string, so the users will be aware that something happened in there.

@raysan5 This would be an excellent solution.

danielchasehooper:
I like that idea

Murlocohol:
So I looked into this and implementing it is rather trivial because vsnprintf() is used. vsnprintf() returns the number of chars that would have been required to make the string, so all we need to do is just compare that to MAX_TEXT_BUFFER_LENGTH and then we will know if we have overflow or not.

[...]

PR #3416, opened 2023-10-14
Fix SetMouseCursor implementation for PLATFORM_WEB

BeardedBread:
Correction to the merge #3414  
Restrict function to only set the cursor inside the canvas.

Done by getting the canvas element via its id and setting its cursor.

Build tested on Linux Mint 21.2 x64.  
Tested on Firefox 118.0.2 x64 and Brave Browser Version 1.59.117 Chromium: 118.0.5993.70 x64.  
The cursor return back to the default cursor outside of the canvas as expected.

Test program:
```c
#include "raylib.h"
#include <stdio.h>


//-------------------- PROTOTYPE CURSOR FUNCTION ------------------
#if defined(PLATFORM_WEB)
#endif
//-------------------------------------------------------------------

// Window size
static const int width = 800;
static const int height = 600;

// Grid sizes
const float grid_width = 200.0f;
const float grid_height = 200.0f;
const int grid_w = 4;

// Display name for cursor
static const char* CURSOR_NAMES[11] = 
{
    "default", "arrow", "text", "crosshair",
    "pointer", "ew-resize", "ns-resize", "nwse-resize",
    "nesw-resize", "move", "not-allowed"
};

void update_step(void)
{
    // Set the cursor based on mouse position in the grid
    Vector2 mouse_pos = GetMousePosition();
    int mouse_idx = (int)(mouse_pos.x / grid_width) + (int)(mouse_pos.y / grid_height) * grid_w;
    // Allow out of bound idx, function should handle this
    SetMouseCursor(mouse_idx);

    static char buffer[64];
    snprintf(buffer, 64, "x: %.2f, y: %.2f, idx: %d", mouse_pos.x, mouse_pos.y, mouse_idx);
    buffer[63] = '\0';

    BeginDrawing();
        ClearBackground(LIGHTGRAY);
        for (int i = 0; i < grid_width; ++i)
        {
            DrawLine(grid_width*i, 0, grid_width*i, height, BLACK);
        }
        for (int i = 0; i < grid_height; ++i)
        {
            DrawLine(0, grid_height*i, width, grid_height*i, BLACK);
        }
        for (int i = 0; i < 11; i++)
        {
            int x = (i % grid_w) * grid_width + grid_width / 2;
            int y = (i / grid_w) * grid_height + grid_height / 2;
            DrawText(CURSOR_NAMES[i], x, y, 12, BLACK);
        }
        DrawText(buffer, 20, 20, 12, BLACK);
    EndDrawing();
}

int main(void)
{
    InitWindow(width, height, "raylib");
    SetTargetFPS(60);

    #if defined(PLATFORM_WEB)
        puts("Setting emscripten main loop");
        emscripten_set_main_loop(update_step, 0, 1);
    #else
        puts("Regular main loop");
        while(true)
        {
            update_step();
            if (WindowShouldClose()) break;
        }
    #endif
    CloseWindow();

}
```

It is similar to the last test, but it allows out of bound int to test the default switch case.  
On the last grid, it prints out the warning as expected.

Edit: Forgot to set the CORE input mouse value. Updated

### DIFF
--- a/src/rcore_web.c
+++ b/src/rcore_web.c
@@ -685,8 +685,9 @@ void SetMouseCursor(int cursor)
         } break;
     }
 
-    // Set the cursor element on the CSS
-    EM_ASM({document.body.style.cursor = UTF8ToString($0);}, cursorName);
+    // Set the cursor element on the canvas CSS
+    // The canvas is coded to the Id "canvas" on init
+    EM_ASM({document.getElementById("canvas").style.cursor = UTF8ToString($0);}, cursorName);
 }
 
 // Register all input events
--- a/src/rcore_web.c
+++ b/src/rcore_web.c
@@ -659,6 +659,7 @@ void SetMousePosition(int x, int y)
 // Set mouse cursor
 void SetMouseCursor(int cursor)
 {
+    CORE.Input.Mouse.cursor = cursor;
     const char *cursorName;
     switch (cursor)
     {
@@ -682,6 +683,7 @@ void SetMouseCursor(int cursor)
         {
             TRACELOG(LOG_WARNING, "Cursor value out of bound (%d). Setting to default", cursor);
             cursorName = "default";
+            CORE.Input.Mouse.cursor = MOUSE_CURSOR_DEFAULT;
         } break;
     }
 


ubkp:
Tested successfully with Firefox 118.0.1 64-bit and Chromium 117.0.5938.149 64-bit on Linux Mint 21.1 64-bit.
Also, good catch on the missing `CORE.Input.Mouse.cursor`. Good job!

src/rcore_web.c
```diff
@@ -682,11 +683,13 @@ void SetMouseCursor(int cursor)
         {
             TRACELOG(LOG_WARNING, "Cursor value out of bound (%d). Setting to default", cursor);
```
ubkp, line 684: @raysan5 From your experience, do you know any scenario where `SetMouseCursor()` would normally be called during the drawing step? If yes, then I guess this TraceLog should be omitted (similar case: https://github.com/raysan5/raylib/issues/3366#issuecomment-1743836669).

raysan5: @ubkp Yes, it can be called in the update/drawing step, for example, when mouse is over an screen element to select it and cursor needs to be changed. In those cases, the function should also avoid changing cursor if requested one is already the current one.

raysan5:
@BeardedBread thanks for the update! I'm removing the Tracelog per @ubkp comment (and my response).

PR #3418, opened 2023-10-14
Fix some omissions

Bigfoot71:
Small correction following the mention in the issue: [[core] Splitting rcore by platform into submodules](https://github.com/raysan5/raylib/issues/3313#issuecomment-1762960531)
Changes the return type of `InitGraphicsDevice()` from `bool` to `int`.
Adds a `return 0` at the end of `InitPlatform()`.


raysan5:

PR #3420, opened 2023-10-14
Remove the `rcore.h` include from `drm`, `web`, `template`

ubkp:
1. While reviewing the current changes, noticed `rcore_desktop.c` no longer includes `rcore.h`.

2. Since the library and examples on `PLATFORM_DESKTOP` are compiling and running fine, I think that include is no longer needed on the modules after the previous updates.

3. So also removed it from `rcore_drm.c` ([L50](https://github.com/raysan5/raylib/pull/3420/files#diff-ae04a2ff3974ba8bcfeb7ddfde3038688d4c2aac31d22c745f52680dcdaf1ddaL50)), `rcore_web.c` ([L48](https://github.com/raysan5/raylib/pull/3420/files#diff-5204360a69ae27d12753764f60a1cd59e3a11ef10b86fce7691c74de0becc8cdL48)) and `rcore_template.c` ([L49](https://github.com/raysan5/raylib/pull/3420/files#diff-8b082901dd641b3d10f6444023946081d7a3d3d84e8ed1c89b119499a82f33d1L49)). Compiled and tested here and everything appears to be working fine without it.

4. **Note:** The include is still present on `rcore_android.c` ([L49](https://github.com/raysan5/raylib/blob/master/src/rcore_android.c#L49)), but I didn't want to change anything there while @Bigfoot71 is working on it.

**Edit:** added line marks.

### DIFF
--- a/src/rcore_drm.c
+++ b/src/rcore_drm.c
@@ -47,8 +47,6 @@
 *
 **********************************************************************************************/
 
-#include "rcore.h"
-
 #include <fcntl.h>   // POSIX file control definitions - open(), creat(), fcntl()
 #include <unistd.h>  // POSIX standard function definitions - read(), close(), STDIN_FILENO
 #include <termios.h> // POSIX terminal control definitions - tcgetattr(), tcsetattr()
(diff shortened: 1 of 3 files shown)


raysan5:
@ubkp thanks, actually I had my doubts about that include. Using the expected pre-processor logic, the `rcore_*.c` platform modules effectively belong to `rcore.c` that already includes `rcore.h`... but I thought some build systems could be more strict and consider the `rcore_*.c` modules as separate modules...

In any case, I like this change, if some problem arise we can just add the include again.

hbiblia:
This is somewhat silly, but wouldn't it be better to add the `void OpenURL(const char *url)` from `rcore_desktop.c` to `rcore_template.c`? The template has code that seems to be for Android.

image

PR #3421, opened 2023-10-14
[core] Change axisCount to be an array

michaelfiber:
Currently there is 1 axisCount variable for all gamepads and it is rewritten each time a gamepad is initialized (or polled for WASM). 

This change switches axisCount to an array, makes `GetGamepadAxisCount` use the passed in `gamepad` index, and updates the logic that sets the gamepad axisCount in each platform to reference an array element.  

There is also some updates to the core_input_gamepad_info example included with this to display more useful information.


michaelfiber:
And there are tons of commits from me merging in upstream changes, etc.  The actual changes this PR represents are relatively small and simple.

raysan5:
@michaelfiber Nice! Thanks for the update! :D

PR #3422, opened 2023-10-14
Make sure rcore.o gets compiled in more situations

Peter0x44:
Currently doing the following:
```
make
touch rcore_desktop.c
make
```

Will not result in rcore.o getting compiled again, despite that rcore_desktop.c was changed.


Peter0x44:
Oops - did not test properly, this is totally broken. My bad.

PR #3423, opened 2023-10-14
 Make sure rcore.o gets compiled in more situations

Peter0x44:
Currently doing the following:
```bash
make
touch rcore_desktop.c
make
```

Will not result in rcore.o getting compiled again, despite that rcore_desktop.c has changed

This commit resolves that

The one downside this has is that for example, when building for desktop, if rcore_web.c is modified, rcore.c will get compiled again, despite that its changes are not used in this situation - I think this is unimportant, though, as make clean would be required to switch between platforms anyway

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -632,6 +632,9 @@ endif
 
 # Compile all modules with their prerequisites
 
+# Prerequisites of core module
+rcore.o : rcore_android.c rcore_desktop.c rcore_drm.c rcore_template.c rcore_web.c
+
 # Compile core module
 rcore.o : rcore.c raylib.h rlgl.h utils.h raymath.h rcamera.h rgestures.h
 	$(CC) -c $< $(CFLAGS) $(INCLUDE_PATHS)


ubkp:
Two legit question:
1. Shouldn't those get `ifeq`'d by `PLATFORM_*`?
2. Shouldn't the prereqs be on a single line?

Peter0x44:
1) they could be, but it's not necessary. I'm only adding them as prerequisites so that the makefile considers rcore.o outdated of the modification time of any of those files is newer than it.
2) it doesn't matter, the for example, the following are equivalent in make
```make
game: graphics.o physics.o input.o
graphics.o: graphics.c
physics.o: physics.c
input.o: input.c
```
And:
```
game: graphics.o
game: physics.o
game: input.o
graphics.o: graphics.c
physics.o: physics.c
input.o: input.c
```
New prerequisites are *appended* to the previous prerequisites

Does that clear everything up?
@ubkp

Peter0x44:
Just to make it clear, the "downside" I talked about at the end of the PR comment is the consequence of not putting all the C files in the ifeq (PLATFORM_WHATEVER) ...
I don't deem it important enough to hinder the readability of the makefile, but perhaps ray thinks otherwise

ubkp:
@Peter0x44 Yup, thank you very much for the answers. :+1:

raysan5:
@Peter0x44 thanks for the review! Also a bit concerned about the different platforms, not sure if it could have any downside...

Peter0x44:
@raysan5 I don't believe this has any downsides
The only one which (might) be an issue is that the makefile will give an error if any of the prerequisites don't have a file that exists
Example:
```make
example.o : nonexistent.c
```
`make: *** No rule to make target 'nonexistent.c', needed by 'example.o'.  Stop.`
However, this can be solved quite easily by simply adding a build rule that does nothing
```make
example.o : nonexistent.c

nonexistent.c :
```
`make: Nothing to be done for 'example.o'.`

I'm happy to submit this change as a follow up if it's a desired behavior

PR #3425, opened 2023-10-15
Fix GenMeshPlane when resX != resZ

neyrox:


raysan5:
@neyrox Please, could you eleaborate what was exactly the issue? Some screenshot illustrating it would be welcome...

neyrox:
> @neyrox Please, could you eleaborate what was exactly the issue? Some screenshot illustrating it would be welcome...

```
Mesh backMesh = GenMeshPlane(40.0f, 40.0f, 4, 3);
```

Caused error like
```
[.WebGL-0x1000032db00] GL_INVALID_OPERATION: Vertex buffer is not big enough for the draw call
```
on rendering

raysan5:
@neyrox thanks for the details and the review!

PR #3428, opened 2023-10-16
[core] Fix some mouse issues on `SDL`

ubkp:
### Fixes
1. Fixes the mouse wheel getting stucked scrolling up or down by adding the missing reset ([R534-R536](https://github.com/raysan5/raylib/pull/3428/files#diff-b7c2628cbe74c1722110bdc42bc4239493b1be7ff2836f8892e3e942a42e9aefR534-R536)).

2. Fixes the mouse sensitivity being really high on 3D by adding the missing previous mouse position registry ([R538-R539](https://github.com/raysan5/raylib/pull/3428/files#diff-b7c2628cbe74c1722110bdc42bc4239493b1be7ff2836f8892e3e942a42e9aefR538-R539)).

3. Fixes the mouse button presses not working properly by adding the missing previous mouse button registry ([R561-R562](https://github.com/raysan5/raylib/pull/3428/files#diff-b7c2628cbe74c1722110bdc42bc4239493b1be7ff2836f8892e3e942a42e9aefR561-R562)) and by adding handling for `SDL_MOUSEBUTTONUP` ([R619-R622](https://github.com/raysan5/raylib/pull/3428/files#diff-b7c2628cbe74c1722110bdc42bc4239493b1be7ff2836f8892e3e942a42e9aefR619-R622)).
.
**VERY IMPORTANT NOTE:** @raysan5 
The `SDL2` mouse button order is `LEFT, MIDDLE, RIGHT, ...` ([doc](https://wiki.libsdl.org/SDL2/SDL_MouseButtonEvent)), but we currently use `LEFT, RIGHT, MIDDLE, ...` ([L674-L676](https://github.com/raysan5/raylib/blob/master/src/raylib.h#L674-L676)), so, when the migration to `SDL2` is complete, that `MouseButton enum` on `raylib.h` needs to be adjusted. Until then, the middle and right mouse binds will be inverted.

### Reference
Items 1, 2, 6 of https://github.com/raysan5/raylib/issues/3313#issuecomment-1763623854.

### Environment
Tested this changes successfully on Linux (Ubuntu 22.04 64-bit, gcc 11.2.0) with SDL2 (2.28.4).

### Edits
**1:** added line marks.
**2:** added another fix, environment, formatting.
**3:** added another fix.

### DIFF
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -535,6 +535,9 @@ void PollInputEvents(void)
     CORE.Input.Mouse.currentWheelMove.x = 0;
     CORE.Input.Mouse.currentWheelMove.y = 0;
 
+    // Register previous mouse position
+    CORE.Input.Mouse.previousPosition = CORE.Input.Mouse.currentPosition;
+
     // Reset last gamepad button/axis registered state
     CORE.Input.Gamepad.lastButtonPressed = GAMEPAD_BUTTON_UNKNOWN;
     for (int i = 0; i < MAX_GAMEPADS; i++) CORE.Input.Gamepad.axisCount[i] = 0;
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -60,7 +60,7 @@
 typedef struct {
     SDL_Window *window;
     SDL_GLContext glContext;
-    
+
     SDL_Joystick *gamepad;
 } PlatformData;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp thank you very much for the review! Good catch the mouse button order! Yes, it should be reviewed, not sure if it follows glfw order...

ubkp:
@raysan5 Checked it here and `GLFW` does indeed use `LEFT, RIGHT, MIDDLE, ...` ([doc](https://www.glfw.org/docs/latest/group__buttons.html)). I'd suggest not flipping them yet on `MouseButton enum` to not cause issues for anyone using current master branch. I can keep track here to remind this when the times comes, if you want.

raysan5:
@ubkp I see, probably the remap should be done on input polling then. Feel free to add a TODO in the code to remember it!

PR #3433, opened 2023-10-17
fix build.zig

ryupold:
I don't know which zig version is intended to be used here,
but if it is master then this PR adjusts to the new API of Compile.addCSourceFiles()


raysan5:
@ryupold Thanks for the review!

PR #3434, opened 2023-10-17
[raudio] Implement GetMasterVolume()

rexim:
It feels a little unfinished when you can SetMasterVolume but can't really Get it. So to finish the symmetry here is the GetMasterVolume implementation.


raysan5:
@rexim Nice! Thanks for the addition!

PR #3435, opened 2023-10-17
Fix SDL keyboard issue

Bigfoot71:
As originally mentioned by @ubkp in the issue [\[core\] Splitting rcore by platform into submodules](https://github.com/raysan5/raylib/issues/3313#issuecomment-1763623854), here is the fix for the keyboard handling issue with SDL.

This modification includes a mapping table between SDL scancodes and raylib keys, along with a function to safely retrieve them.

I've also revised the state change of `CORE.Window.shouldClose` by using `CORE.Input.Keyboard.exitKey` instead of `SDLK_ESCAPE` directly, but as indicated in the comment, I believe this part would be better placed outside of the switch statement.

### DIFF
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -72,10 +72,118 @@ extern CoreData CORE;                   // Global CORE state context
 static PlatformData platform = { 0 };   // Platform specific data
 
 //----------------------------------------------------------------------------------
+// Local Variables Definition
+//----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ubkp:
Tested successfully on Linux (Ubuntu 22.04 64-bit, gcc 11.2.0) with SDL2 (2.28.4). Great job!

raysan5:
@Bigfoot71 Nice! Thanks for the improvement!

PR #3438, opened 2023-10-18
Fix `SetGamepadMappings` returning

Bigfoot71:
I fixed the return of `SetGamepadMappings` which returned nothing. The function now returns `SDL_GameControllerAddMapping` which returns this information according to the documentation:
> returns 1 if a new mapping is added, 0 if an existing mapping is updated,
> -1 on error; call SDL_GetError() for more information.


ubkp:
@Bigfoot71 Good catch!

Bigfoot71:
I move this into  https://github.com/raysan5/raylib/pull/3439

PR #3439, opened 2023-10-18
[core] Add more missing implementations to SDL

Bigfoot71:
As requested in https://github.com/raysan5/raylib/issues/3313 here is the implementation of the `SetWindowState`, `ClearWindowState`, and `SetWindowIcon` functions for `rcore_desktop_sdl.c`.

The functions `SetWindowState` and `ClearWindowState` don't handle all flags, but it seems that some things can't be managed directly through SDL. Please take a look at the notes in the comments.

Otherwise, `SetWindowIcon` seems to work fine on my end on Cinnamon (Linux Mint). I don't have access to Windows to test it right now.

I'm sharing this work now so that if you have any feedback, I can make corrections, or you can merge it directly if everything looks good.

### DIFF
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -221,7 +221,16 @@ bool WindowShouldClose(void)
 // Toggle fullscreen mode
 void ToggleFullscreen(void)
 {
-    //SDL_SetWindowFullscreen
+    if (!IsWindowState(FLAG_FULLSCREEN_MODE))
+    {
+        SDL_SetWindowFullscreen(platform.window, SDL_WINDOW_FULLSCREEN);
+        CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
+    }
+    else
+    {
+        SDL_SetWindowFullscreen(platform.window, 0);
+        CORE.Window.flags &= ~FLAG_FULLSCREEN_MODE; 
+    }
 }
 
 // Toggle borderless windowed mode
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -647,8 +647,20 @@ int GetCurrentMonitor(void)
 // Get selected monitor position
 Vector2 GetMonitorPosition(int monitor)
 {
-    TRACELOG(LOG_WARNING, "GetMonitorPosition() not implemented on target platform");
-    return (Vector2){ 0, 0 };
+    if (monitor < 0 || monitor >= SDL_GetNumVideoDisplays())
+    {
+        TRACELOG(LOG_ERROR, "Invalid monitor index");
+        return (Vector2) { 0, 0 };
+    }
+
+    SDL_Rect displayBounds;
+    if (SDL_GetDisplayBounds(monitor, &displayBounds) != 0)
+    {
+        TRACELOG(LOG_ERROR, "Failed to get display bounds");
+        return (Vector2) { 0, 0 };
+    }
+
+    return (Vector2) { displayBounds.x, displayBounds.y };
 }
 
 // Get selected monitor width (currently used by monitor)
--- a/src/rcore_desktop_sdl.c
+++ b/src/rcore_desktop_sdl.c
@@ -255,17 +255,25 @@ void SetWindowState(unsigned int flags)
 {
     CORE.Window.flags |= flags;
 
+    if (flags & FLAG_VSYNC_HINT)
+    {
+        SDL_GL_SetSwapInterval(1);
+    }
     if (flags & FLAG_FULLSCREEN_MODE)
     {
         SDL_SetWindowFullscreen(platform.window, SDL_WINDOW_FULLSCREEN);
+    }
+    if (flags & FLAG_WINDOW_RESIZABLE)
+    {
+        SDL_SetWindowResizable(platform.window, SDL_TRUE);
     }
     if (flags & FLAG_WINDOW_UNDECORATED)
     {
         SDL_SetWindowBordered(platform.window, SDL_FALSE);
     }
-    if (flags & FLAG_WINDOW_RESIZABLE)
+    if (flags & FLAG_WINDOW_HIDDEN)
     {
-        SDL_SetWindowResizable(platform.window, SDL_TRUE);
+        SDL_HideWindow(platform.window);
     }
     if (flags & FLAG_WINDOW_MINIMIZED)
     {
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Uh wait, I'm still going to put tracelogs for the unmanaged flags

src/rcore_desktop_sdl.c
```diff
@@ -253,19 +253,182 @@ void RestoreWindow(void)
...
+    if (flags & FLAG_WINDOW_UNFOCUSED)
+    {
+        // NOTE: To be able to implement this part it seems that we should
+        // do it ourselves, via `Windows.h`, `X11/Xlib.h` or even `Cocoa.h`
+    }
+    if (flags & FLAG_WINDOW_TOPMOST)
+    {
+        // NOTE: To be able to implement this part it seems that we should
+        // do it ourselves, via `Windows.h`, `X11/Xlib.h` or even `Cocoa.h`
+    }
```
ubkp, line 287: @Bigfoot71 Could you check if `SDL_SetWindowAlwaysOnTop` ([doc](https://wiki.libsdl.org/SDL2/SDL_SetWindowAlwaysOnTop)) could replace it?

Bigfoot71: Thanks! I'll take care of that!

Bigfoot71:
It's good, the `SetWindowState` and `ClearWindowState` functions have been completed (I added the handling of certain flags I had forgotten), along with the correction proposed by @ubkp and some TraceLog statements for the flags that are not yet supported.

I've also added VSync support.

src/rcore_desktop_sdl.c
```diff
@@ -253,19 +253,247 @@ void RestoreWindow(void)
...
+    if (flags & FLAG_WINDOW_MOUSE_PASSTHROUGH)
+    {
+        //SDL_SetWindowGrab(platform.window, SDL_FALSE);
+        TraceLog(LOG_WARNING, "SetWindowState() - FLAG_WINDOW_MOUSE_PASSTHROUGH is not supported on PLATFORM_DESKTOP_SDL");
+    }
+    if (flags & FLAG_BORDERLESS_WINDOWED_MODE)
+    {
+        // NOTE: Same as FLAG_WINDOW_UNDECORATED with SDL ?
+        SDL_SetWindowBordered(platform.window, SDL_FALSE);
```
ubkp, line 317: @Bigfoot71 Could you replace it with `ToggleBorderlessWindowed()`? This state has a bit more going on ([L200-L272](https://github.com/raysan5/raylib/blob/master/src/rcore_desktop.c#L200-L272)) behind the scene.

ubkp: > Or should we do exactly the same thing,

@Bigfoot71 I think we should.

Unfortunately `SDL_SetWindowBordered` just removes the window decoration.

`FLAG_BORDERLESS_WINDOWED_MODE` through `ToggleBorderlessWindowed` does quite a lot of handling ([L200-L272](https://github.com/raysan5/raylib/blob/master/src/rcore_desktop.c#L200-L272)). E.g.: leaves real fullscreen if present, saves window pos/sizes (for toggling), removes the decoration, sets it top most, resize it to match the fullscreen and reposition it on the correct monitor. Which is necessary to simulate a fullscreen desktop (which although "present" on `SDL`, is still a different thing because SDL's fullscreen desktop will still cause the window to get minimized when alt tabbing).

ubkp: @Bigfoot71 If you want (and have the time), feel free to take `ToggleBorderlessWindowed`. Note: it probably could be done at the same time with `ToggleFullscreen`, `SetWindowMonitor` and `GetMonitorPosition` since they share a lot of the same handling. Just let me know. :+1:

Bigfoot71: @ubkp no problem, I'll take care of it.

However, I'm getting this kind of error from X11 when I call `SDL_SetWindowPosition`:
```console
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  40 (X_TranslateCoords)
  Resource id in failed request:  0x2a0d45e
  Serial number of failed request:  355
  Current serial number in output stream:  355
```
And this happens no matter what position I set. Here's the code snippet (it's for `ToggleBorderlessWindowed`):

```c
SDL_Rect displayBounds;
if (SDL_GetDisplayBounds(GetCurrentMonitor(), &displayBounds) == 0)
{
    SDL_SetWindowPosition(platform.window, displayBounds.x, displayBounds.y);
    SDL_SetWindowSize(platform.window, displayBounds.w, displayBounds.h);
}
```

Bigfoot71: No, it's good, I reversed the call order compared to the GLFW version and it no longer crashes, but it still doesn't work hehehe, don't panic, it will be fixed

ubkp: @Bigfoot71 I think you need to do it like this:
```
int currentMonitor = SDL_GetWindowDisplayIndex(platform.window);
SDL_Rect displayBounds;
int res = SDL_GetDisplayBounds(currentMonitor, &displayBounds);
if (res == 0)
{
    SDL_SetWindowPosition(platform.window, displayBounds.x, displayBounds.y);
    SDL_SetWindowSize(platform.window, displayBounds.w, displayBounds.h);
}
```

Bigfoot71: @ubkp If you, or anyone else, can try it on their system, to see if you have the same problem as with the taskbar which remains in front of the window with `ToggleBordelessWindowed` that would be cool. https://github.com/Bigfoot71/raylib/commit/232a0615ea61c404b37a976c1f62b03fcef1e81d

ubkp: @Bigfoot71 Tested it right now on `Fluxbox (X11)` and `MATE (X11)` on `Linux Mint 21.1 64-bit` and it's working great! Very good job! :+1:

I'll spin a `Fedora 38 64-bit` here to test it on `Gnome` and `KDE` with `Wayland` (may take me an hour to report back tho).

Are you on `Cinnamon (X11)`?

Bigfoot71: @ubkp Yes, I'm on Cinnamon, so it's probably related to that

Bigfoot71: @ubkp I have finished implementing the functions you mentioned to me. However, I only have one screen, so I believe (by deduction) that the implementation of 'SetWindowMonitor' should be correct, but I have no idea if it works, and furthermore, it is incomplete.

So, if you or someone else could try and correct this function, it would be great! (https://github.com/Bigfoot71/raylib/commit/db0674cfc4e96bc425b635d57d9f14811d5afa18)

ubkp: @Bigfoot71 `SetWindowMonitor` has a bit more handling ([L615-L643](https://github.com/raysan5/raylib/blob/master/src/rcore_desktop.c#L615-L643)). But don't worry about it, I'll finish it up. Thanks so much for all the work! :+1:

raysan5:
@Bigfoot71 @ubkp Wow! Amazing work! Thank you very much! 👍🙂

PR #3445, opened 2023-10-20
[CORE] Reset FPS averages when a window is inited

JeffM2501:
This PR forces the FPS counters to reset when the current frame is 0. Because the FPS system uses static data, we need to reset it whenever InitWindow is called. A user may close and reopen the window (to enable features like vsync). This PR uses the frame counter for automation to detect this case and reset the static data so each InitWindow always starts the same.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -484,9 +484,9 @@ void InitWindow(int width, int height, const char *title)
     }
 #endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 thanks for the review! I like this new approach!

PR #3453, opened 2023-10-22
Review prerequisites of rcore.c

Peter0x44:
I noticed rcore_desktop_sdl.c was missing from the prerequisites list
Now that the platforms are split to their own directory, it is simple to just use a "wildcard" to match them all in the makefile


raysan5:

PR #3455, opened 2023-10-23
Update Makefile

colleen05:
Mentioned .lib for Windows static libraries.


colleen05:
This was just done for consistency :slightly_smiling_face:

raysan5:
@colleen05 I think that comment is correct, it refers to the possible extensions of static libraries.

PR #3458, opened 2023-10-23
Fixed LoadSoundAlias() memory leak

gk646:
Fixes a memory leak as a result of creating an AudioBuffer* with the old source.frameCount. This internally allocates memory to the structs data pointer which is then later overridden by the correct sound data of the source sound.

Additionally added a volume assignment from old to new as currently there is no way to get the volume of a sound and the AudioBuffer struct is not reachable from user code due to opaque definition. Also it makes sense as an Alias is a new name for the same thing.


raysan5:
@gk646 thanks for the review!

Issue #3463, opened 2023-10-25
IsGestureDetected inconsistent parameters


jestarray:
```c

// src/rgestures.h line 127
bool IsGestureDetected(int gesture);                    // Check if a gesture have been detected

// src/raylib.h line 1170
RLAPI bool IsGestureDetected(unsigned int gesture);     // Check if a gesture have been detected
```
how come one is `int` and the other is `unsigned int`? Which one should it be? Originally it was an int(in rust bindings)

ubkp:
@jestarray `gesture` is `enum` ([L79-L91](https://github.com/raysan5/raylib/blob/master/src/rgestures.h#L79-L91)). Quoting [this answer](https://stackoverflow.com/questions/2579230/signedness-of-enum-in-c-c99-c-cx-gnu-c-gnu-c99): _"An enum is guaranteed to be represented by an integer, but the actual type (and its signedness) is implementation-dependent."_

In the case of `IsGestureDetected()` I don't think it matters, since underflowing or overflowing that parameter will be out of bounds of that `enum` valid gestures anyway. But, IMHO, could probably be changed to `int gesture` on `raylib.h` just for the sake of consistency.

PR #3464, opened 2023-10-25
Fix IsGestureDetected parameter inconsistency in raylib.h with rgextures.h

jestarray:
closes https://github.com/raysan5/raylib/issues/3463


raysan5:
@jestarray thanks for the review

raysan5:
@jestarray I'm afraid I reverted this PR, it broke the build because it was incomplete but when reviewing implementation I noticed that gesture should be `unsigned int` for the flag check agains `unsigned int` flag:
```c
// Check if a gesture have been detected
bool IsGestureDetected(unsigned int gesture)
{
    if ((GESTURES.enabledFlags & GESTURES.current) == gesture) return true;
    else return false;
}
```

Issue #3465, opened 2023-10-25
[example] The zoom in the Julia set example is not correct


joshcol9232:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

The Julia set example does not zoom correctly when using the mouse buttons. It does not scale with the current zoom level, and eventually even turns back on itself.
Ever since I added the example (about 4 years ago!) I very occasionally think about this bug I made in the original. I'm finally back to fix it!

PR #3466, opened 2023-10-25
Fix the zooming in the Julia set shader example.

joshcol9232:
Hello! Ever since I added the Julia set shader example (5 years ago!) it had always bugged me that the zoom function was never correct. This PR fixes this, and is also a general tidy-up.

### In summary:

- Fix zoom issue.
- Add a `COLOR_CYCLES` variable to the shader to allow for cycling of the color palette. This allows for finer details of the fractal to be shown when using a high max iteration number.
- Fix some pretty terrible if statement mess.
- Update copyright.

### Issue

Fixes: #3465

### Testing

I have tested the code on my Linux machine. I have not tested on glsl100 platforms, or Windows.


joshcol9232:
I was convinced I had fixed it, but further testing says no! I'll be back with a new PR. I think I understand the issue better after this attempt.

Issue #3456, opened 2023-10-23
[core] GetCurrentMonitor() inconsistency between GLFW and SDL


M374LX:
### Issue description

When building raylib and running a program in windowed mode, ``GetCurrentMonitor()`` returns the monitor containing the window's top-left corner when using GLFW (PLATFORM_DESKTOP). When using SDL (PLATFORM_DESKTOP_SDL), however, it returns the monitor containing the center of the window.

Since raylib's SDL implementation of ``GetCurrentMonitor()`` just calls ``SDL_GetWindowDisplayIndex()``, this is expected, as this function from SDL really returns the monitor containing the center of the window, according to the [documentation](https://wiki.libsdl.org/SDL2/SDL_GetWindowDisplayIndex).

Does anyone agree with the idea of changing the behavior on the GLFW implementation so it returns the monitor containing the center of the window for consistency? I think this would also make problems like https://github.com/raysan5/raylib/pull/3391#issuecomment-1752211850 and https://github.com/raysan5/raylib/pull/3391#issuecomment-1762364369 less common.

### Environment

Linux (Xorg as display server and dwm as window manager).

### Code Example

```C
#include <raylib.h>
#include <stdio.h>

int main()
{
	char str[32];

	InitWindow(800, 600, "Test");

	while (!WindowShouldClose()) {
		sprintf(str, "Monitor: %d", GetCurrentMonitor());

		BeginDrawing();
		ClearBackground(GRAY);
		DrawText(str, 0, 64, 32, BLACK);
		EndDrawing();
	}

	CloseWindow();

	return 0;
}
```

orcmid:
The GLFW delivery of (desktop) scaling information determines the current monitor (and its scaling) by determining which one the cursor is in.  There might need to be some harmonization with that.

ubkp:
> Does anyone agree with the idea of changing the behavior on the GLFW implementation so it returns the monitor containing the center of the window for consistency?

IMHO, that's a good idea. Just need to add:
```
x+=(int)CORE.Window.screen.width/2;
y+=(int)CORE.Window.screen.height/2;
```
To [L758](https://github.com/raysan5/raylib/blob/b3028e4891083c077236df437011184b16b6d293/src/platforms/rcore_desktop.c#L758) on `rcore_desktop.c`. Please send a PR.

raysan5:
> Does anyone agree with the idea of changing the behavior on the GLFW implementation so it returns the monitor containing the center of the window for consistency?

[...]

PR #3468, opened 2023-10-25
GetCurrentMonitor() - check window center instead of top-left corner

M374LX:
Partial fix for #3456.

At present, in case the center of the window is out of any monitor, it detects the first one as current on the GLFW implementation, while ``SDL_GetWindowDisplayIndex()`` apparently detects the monitor that is the closest to the center of the window (https://github.com/libsdl-org/SDL/blob/b29128994e87d82eae4751d65ff38b935ce7d6a2/src/video/SDL_video.c#L1114).

Maybe we can merge this PR as is and leave the issue of the center of the window being out of any monitor to another one.

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -755,6 +755,8 @@ int GetCurrentMonitor(void)
             int y = 0;
 
             glfwGetWindowPos(platform.handle, &x, &y);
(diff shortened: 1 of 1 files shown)


raysan5:

 > Maybe we can merge this PR as is and leave the issue of the center of the window being out of any monitor to another one.

Agree.

PR #3469, opened 2023-10-26
Fix missing `PLATFORM_DESKTOP_SDL` checks.

2Bear:
There are two missing `PLATFORM_DESKTOP_SDL` checks in `rlgl.h`, so that OpenGL ES support in `PLATFORM_DESKTOP_SDL` is broken.

Add missing PLATFORM* checks can fix that, but a bit dirty.

BTW, consider that more platforms will be supported here, so maybe `rlgl.h` should be more independent.


raysan5:
@2Bear thanks for the review!

 > BTW, consider that more platforms will be supported here, so maybe rlgl.h should be more independent.

raylib is intended for OpenGL so `rlgl` should already cover all platforms supporting the technology properly. OpenGL ES support on desktop platforms is not a common use case, they should target OpenGL desktop instead.

PR #3472, opened 2023-10-26
GetCurrentMonitor() - use closest monitor

### LINKED ISSUE #3456: [core] GetCurrentMonitor() inconsistency between GLFW and SDL

M374LX:
This PR complements #3468 by using the closest monitor when the center of the window is not within any of them (instead of the first one).

It is essentially adapted from https://github.com/libsdl-org/SDL/blob/b29128994e87d82eae4751d65ff38b935ce7d6a2/src/video/SDL_video.c#L1114.

Closes #3456.

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -751,15 +751,19 @@ int GetCurrentMonitor(void)
         }
         else
         {
-            int x = 0;
-            int y = 0;
+            int closestDist = 0x7FFFFFFF;
 
-            glfwGetWindowPos(platform.handle, &x, &y);
-            x += (int)CORE.Window.screen.width / 2;
-            y += (int)CORE.Window.screen.height / 2;
+            // Window center position
+            int wcx = 0;
+            int wcy = 0;
+
+            glfwGetWindowPos(platform.handle, &wcx, &wcy);
+            wcx += (int)CORE.Window.screen.width / 2;
+            wcy += (int)CORE.Window.screen.height / 2;
 
             for (int i = 0; i < monitorCount; i++)
             {
+                // Monitor top-left position
                 int mx = 0;
                 int my = 0;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@M374LX Merging the improvement but please note that it adds an extra level of code complexity to care for a very specific situation that probably a very low percentatge of users will face. In raylib I try to give priority to code simplicity over those very specific side cases...

ubkp:
IMHO this should not have been merged. `GLFW` will never completely match `SDL`. Since `SDL` is implemented now, if the desired funcionality is present there, then the users should use `rcore_desktop_sdl` instead. Also, this makes simple code a nightmare to maintain.

raysan5:
@ubkp I agree this is probably an overengineered solution to a very side case and trying to match another different library, at the cost of an increase in code complexity. Now it's merged, but if any issue arises, this will be reverted.

I'll try to avoid those kind of approaches in the future. raylib is intended to be simple.

PR #3429, opened 2023-10-16
[core] Remove `rcore.h` include from `Android`

ubkp:
Very short fix to remove the `rcore.h` include from `PLATFORM_ANDROID` ([L49](https://github.com/raysan5/raylib/pull/3429/files#diff-df7f2529f920b22e7f3b28d8c53b9042fe9ba3e0c4be0a421ac8d9e0c3bde72dL49)) so it matches the other submodules.

References: #3420, https://github.com/raysan5/raylib/issues/3313#issuecomment-1764276021.

**Edit:** added line marks.

PR #3475, opened 2023-10-28
[core] Remove `rcore.h` include from `SDL`

ubkp:
### Changes
1. Very short fix to remove the `rcore.h` include from `rcore_desktop_sdl.c` ([L52](https://github.com/raysan5/raylib/pull/3475/files#diff-43fa9fe6c2302f00d9c0b8bcbe58d8ca92ca52e157d5b19a612f90a81c698c12L52)) so it matches the other submodules.
**References:** [#3420](https://github.com/raysan5/raylib/pull/3420), [#3429](https://github.com/raysan5/raylib/pull/3429).

### Environment
Changes successfully tested on Linux (Ubuntu 22.04 64-bit, gcc 11.2.0) with SDL2 (2.28.4).

### Edits
1: added line marks.

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -49,8 +49,6 @@
 *
 **********************************************************************************************/
 
-#include "rcore.h"
-
 #include "SDL.h"            // SDL base library (window/rendered, input, timming... functionality)
 #include "SDL_opengl.h"     // SDL OpenGL functionality (if required, instead of internal renderer)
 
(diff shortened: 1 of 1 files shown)


ubkp:
@raysan5 After this change, the only other file using `rcore.h` is `core.c`.

Since `rcore.h` is just some defines and the `CoreData` struct, perhaps it would make sense to move it inside `rcore.c`?

raysan5:
> Since rcore.h is just some defines and the CoreData struct, perhaps it would make sense to move it inside rcore.c?

@ubkp thanks for the review! And, yeah, that's right, `rcore.h` could probably be moved back to `rcore.c`.

PR #3483, opened 2023-10-29
Fix compressed DDS texture loading issues

JaanDev:
This code fixes the black/crashing compressed DDS textures loading issue but instead forces only 1 mipmap to be loaded. This should still be reviewed though ig as this PR is not a final solution tothe problem.


raysan5:
@JaanDev thanks for the temporal fix!

JaanDev:
You're welcome! =)

Issue #3487, opened 2023-10-30
[examples] Review some examples not working properly on several platforms


raysan5:
### Issues on multiple platforms

- [x] **REVIEW: `core_3d_camera_free`**. Displayed controls do not match and camera behaves strangely.

### Issues on `PLATFORM_DESKTOP`,  **macOS** (and maybe other platforms/OSs)

 - [x] **REVIEW: `custom_frame_control`**  --> Requires raylib built with `SUPPORT_CUSTOM_FRAME_CONTROL`.
 - [ ] **REVIEW: `models_loading`**. It seems to be buggy at selecting the model, needs some review
 - [x] **REVIEW: `skybox_loading`**. Doesn't fill the entire window, only the bottom-left quadrant --> Probably related to HighDPI.
 - [x] **REVIEW: `shader_hot_reloading`, `shader_spotlight`**. Scaling issue between the mouse and circle positions --> Probably related to HighDPI.
 - [ ] **REVIEW: `top_down_lights`**. Fully illuminated (no shadows are drawn) until light is moved

### Issues on `PLATFORM_WEB`

- [x] **REVIEW: `models_skybox`**. WebGL: INVALID_VALUE: disableVertexAttribArray: index out of range _glDisableVertexAttribArray.
- [x] **REVIEW: `core_custom_frame_control`** --> Requires raylib built with `SUPPORT_CUSTOM_FRAME_CONTROL`.
- [x] **REVIEW: `core_loading_thread`** --> Threading support depends on the browser and security reasons.
- [x] **REVIEW: `core_window_flags`** --> Most window flags are intended for a window, not a canvas.
- [x] **REVIEW: `raylib_opengl_interop`** --> Example intended for OpenGL desktop, not OpenGL ES.
- [x] **REVIEW: `rlgl_standalone`** --> `PLATFORM_DESKTOP` exclusively
- [x] **REVIEW: `models_mesh_instancing`**. Not working properly on web, issues with the shader but ONLY when uploade to GitHub, it works as expected locally.

### Issues on `PLATFORM_DRM`

- [x] **REVIEW: Examples with 3D cameras (e.g.: `models_first_person_maze`, `core_3d_camera_free`)**. Camera moving uncontrollably.
- [ ] **REVIEW: Examples with visible cursors (e.g.: `shapes_draw_circle_sector`, `shapes_draw_ring`)**. Cursor not showing.
[...]

PR #3240, opened 2023-08-11
Fix rcamera.h issues

ubkp:
This PR address part of _"REVIEW: Example: core_3d_camera_free. Displayed controls do not match and camera behaves strangely."_ from https://github.com/raysan5/raylib/discussions/2952.

1. Fixes `CAMERA_FREE` behaving strangely (unintended rolling while yawing and pitching) by setting `rotateUp` to `false` ([R439](https://github.com/raysan5/raylib/pull/3240/files#diff-cb37d550d2b398139f0ed2cebf04598776b16fa58107b4b4b7b7099280a67db5R439)). This change does not impact the other cameras, because `rotateUp` is already `false` for them.

2. Fixes `CAMERA_FREE` mouse wheel zoom by adding it to the zoom mode check ([R503](https://github.com/raysan5/raylib/pull/3240/files#diff-cb37d550d2b398139f0ed2cebf04598776b16fa58107b4b4b7b7099280a67db5R503)).

3. Adds pan to `CAMERA_FREE` by adding handling for it before mouse and keyboard handling ([R462-R472](https://github.com/raysan5/raylib/pull/3240/files#diff-cb37d550d2b398139f0ed2cebf04598776b16fa58107b4b4b7b7099280a67db5R462-R472)) and a define for `CAMERA_PAN_SPEED` ([R187](https://github.com/raysan5/raylib/pull/3240/files#diff-cb37d550d2b398139f0ed2cebf04598776b16fa58107b4b4b7b7099280a67db5R187)).

4. Readds up (`Space`) and down (`Left Ctrl`) movement to `CAMERA_FREE` ([R496-R500](https://github.com/raysan5/raylib/pull/3240/files#diff-cb37d550d2b398139f0ed2cebf04598776b16fa58107b4b4b7b7099280a67db5R496-R500)). This change does not impact the other cameras because it was already disabled for them.

Tested these proposed changes successfully on `PLATFORM_DESKTOP` on Linux (Linux Mint 21.1 x86_64) and `PLATFORM_WEB`.

**Edit:** added line marks.

PR #3488, opened 2023-10-30
[examples] Fix `core_3d_camera_free`

ubkp:
### Changes
The free camera main issues were actually fixed on #3240, but I forgot to follow-up and finish/clean up the example ([R68-R69](https://github.com/raysan5/raylib/pull/3488/files#diff-1541d30f4fe89bcf2c94e5cb16e050cceb0c962b4cd3c4e7f7bd2486e2d066b8R68-R69), [L74-L76](https://github.com/raysan5/raylib/pull/3488/files#diff-1541d30f4fe89bcf2c94e5cb16e050cceb0c962b4cd3c4e7f7bd2486e2d066b8L74-L76)).

### Notes
@raysan5 IMHO, I don't think we should add/fix the `Alt + Mouse Wheel Pressed to Rotate` ([L74](https://github.com/raysan5/raylib/pull/3488/files#diff-1541d30f4fe89bcf2c94e5cb16e050cceb0c962b4cd3c4e7f7bd2486e2d066b8L74)), neither the (commented) `Alt + Ctrl + Mouse Wheel Pressed for Smooth Zoom` ([L75](https://github.com/raysan5/raylib/pull/3488/files#diff-1541d30f4fe89bcf2c94e5cb16e050cceb0c962b4cd3c4e7f7bd2486e2d066b8L75)) because, with the current camera setup, that would require hardcoding them on `rcamera.h`. IMHO, that would make the handling on `rcamera.h` needlessly complex just to accomodate this example. Please let me know if that's not the case, and they should be added.

### Reference
#3487

### Environment
Tested on Linux (Mint 21.1 64-bit).

### Edits
**1:** added line marks; editing.


raysan5:
> IMHO, that would make the handling on rcamera.h needlessly complex just to accomodate this example. Please let me know if that's not the case, and they should be added.

@ubkp Agree. With the `camera` module redesign, some of the examples were not properly adapted and some previous features were removed in exchange of other features and better code organization.

In any case, it would be nice if that functionality could be moved to user side in some way, the example tried to illustrate a 3d free navigation system similar to the ones available on DCC packages (3dMax, Maya, Blender...)

ubkp:
> In any case, it would be nice if that functionality could be moved to user side in some way, the example tried to illustrate a 3d free navigation system similar to the ones available on DCC packages (3dMax, Maya, Blender...)

@raysan5 Oh, I thought it was "free" as in "noclip/spectate mode".

If it's ok, I'd like to give a try at reviewing the camera system (specially moving the camera logic to the user side and the examples teaching/suggesting how a camera could be implemented with them), but I don't think I can get this ready for `5.0`.

raysan5:
> If it's ok, I'd like to give a try at reviewing the camera system (specially moving the camera logic to the user side and the examples teaching/suggesting how a camera could be implemented with them), but I don't think I can get this ready for 5.0.

Sure, feel free. Actually with current `camera` module should be possible to implement the camera rotation on user side, using the available `Camera*()` functions.

There is also a [`UpdateCameraPro()`](https://github.com/raysan5/raylib/blob/master/src/rcamera.h#L528) available as a prove-of-concept of a camera control independent of inputs, allowing users to provide their own inputs for custom camera movement designs. [`core_3d_camera_first_person`](https://github.com/raysan5/raylib/blob/master/examples/core/core_3d_camera_first_person.c#L122) contains an usage example but probably a better example should be provided.

In any case, `camera` module is just provided for user convenience but updating `Camera` structure at user side should be enough for custom camera systems implementation (specially targeted to advance users).

PR #3490, opened 2023-10-30
adding rlEnablePointMode to mirror rlEnableWireMode

JettMonstersGoBoom:
rlDisableWireMode and rlDisablePointMode would be identical, so I didn't submit that.


raysan5:
@JettMonstersGoBoom thanks for the addition

PR #3494, opened 2023-10-31
Fix QOA seeking

### LINKED ISSUE #3451: [audio] QOA seeking not working

### Issue description
Attempting to use `SeekMusicStream` with a QOA file does not do any proper seeking at all, specially bad if attempting to seek a track to a certain point before it begins to properly play, as instead it'll forcibly rewind.

Doing this with a OGG one works just fine. 

### Environment

Windows 10, OpenGL 3.3

### Issue Screenshot

(this is an audio issue...)

### Code Example
```
PlayMusicStream(track);
SeekMusicStream(track, 15.73f);
```
With the attached QOA version of a track, it will not skip its beginning. This will however work with the OGG version.
[music_for_testing.zip](https://github.com/raysan5/raylib/files/13061907/music_for_testing.zip)

veins1:
Fixes #3451

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1797,7 +1797,9 @@ void SeekMusicStream(Music music, float position)
         case MUSIC_AUDIO_MP3: drmp3_seek_to_pcm_frame((drmp3 *)music.ctxData, positionInFrames); break;
 #endif
 #if defined(SUPPORT_FILEFORMAT_QOA)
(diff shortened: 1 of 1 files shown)


raysan5:
@veins1 thanks for the fix! I'm reviewing it and maybe moving to `qoaplay` module.

PR #3497, opened 2023-10-31
[EXAMPLES] Remove unused structures from lighting fragment shaders

JeffM2501:
The lighting shaders had a structure that was not used, this PR removes it.


raysan5:
@JeffM2501 thanks for the review! I think it was there for some future improvement... or maybe from a previous example...

PR #3498, opened 2023-11-01
[core] Moves `keymapUS[]` and fixes `GetCharPressed` for `PLATFORM_DRM`

ubkp:
### Changes
1. Moves the `keymapUS[]` LUT out of `PollKeyboardEvents` ([L1457-L1476](https://github.com/raysan5/raylib/pull/3498/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922L1457-L1476)) to `Local Variables Definition` ([R143-R162](https://github.com/raysan5/raylib/pull/3498/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922R143-R162)) so it doesn't have to be redefined every time `PollKeyboardEvents()` runs on `PLATFORM_DRM`.

2. Adds handling for `CORE.Input.Keyboard.charPressedQueue` and `CORE.Input.Keyboard.charPressedQueueCount` ([R164-R171](https://github.com/raysan5/raylib/pull/3498/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922R164-R171), [R1533-R1543](https://github.com/raysan5/raylib/pull/3498/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922R1533-R1543)) that was missing on `PLATFORM_DRM`. That was the reason why couldn't type on `examples with type inputs` (e.g.: `text_draw_3d`, `text_input_box`), since they used `GetCharPressed()`. Therefore, fixes those examples.

### Note
- The `EvkeyToUnicodeLUT` ([R164-R171](https://github.com/raysan5/raylib/pull/3498/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922R164-R171)) is incomplete, I just mapped the most essential keys so we could get this working. If this doesn't get replaced on the `DRM` input rework, I'll comeback and finish the LUT.

### Reference
- https://github.com/raysan5/raylib/issues/3487

### Enviroment
- Tested on Raspberry Pi 3 Model B 1.2 with Linux (Raspberry Pi OS 11 Bullseye 32-bit armhf).

### Edits
- **1:** added line marks, formatting.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -138,6 +138,39 @@ extern CoreData CORE;                   // Global CORE state context
 static PlatformData platform = { 0 };   // Platform specific data
 
 //----------------------------------------------------------------------------------
+// Local Variables Definition
+//----------------------------------------------------------------------------------
+// Scancode to keycode mapping for US keyboards
+// TODO: Replace this with a keymap from the X11 to get the correct regional map for the keyboard:
+// Currently non US keyboards will have the wrong mapping for some keys
+static const int keymapUS[] = {
+    0, 256, 49, 50, 51, 52, 53, 54, 55, 56, 57, 48, 45, 61, 259, 258, 81, 87, 69, 82, 84,
+    89, 85, 73, 79, 80, 91, 93, 257, 341, 65, 83, 68, 70, 71, 72, 74, 75, 76, 59, 39, 96,
+    340, 92, 90, 88, 67, 86, 66, 78, 77, 44, 46, 47, 344, 332, 342, 32, 280, 290, 291,
+    292, 293, 294, 295, 296, 297, 298, 299, 282, 281, 327, 328, 329, 333, 324, 325,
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp thank you very much for the review!

 > so it doesn't have to be redefined every time `PollKeyboardEvents()` runs on `PLATFORM_DRM`.

Please not that, afaik, a `static` variable defined inside a function is equivalent to a global variable but with reduced scope. It is only initialized once and can be changed on later calls to the function containing it. In any case, having it exposed globally is probably better for organization and future updates.

ubkp:
> Please not that, afaik, a `static` variable defined inside a function is equivalent to a global variable but with reduced scope. It is only initialized once and can be changed on later calls to the function containing it.

@raysan5 Oh, I didn't know that. Thanks for the explanation. :+1:

Issue #3335, opened 2023-09-21
[gestures] Tap gesture not properly recognized on Raspberry Pi


AndrewCapon:
### Issue description

Using 'core_input_gestures' I see an issue where only HOLD, DRAG, PINCH IN and PINCH OUT are recognised.

I have had a look at the code, starting with why TAP is not recognised as I'm guessing if this started working the rest would!

So with the help of some debug logging:

 [0xf78cf040] is the main thread, [0xf0155440] is the event handling thread :

'UpdateGestures()' is called via 'EndDrawing()' -> 'PollInputEvents()' every frame.

I am logging the current gesture at the start of the main loop.

```
DEBUG: [0xf78cf040] Start Main Loop Gesture = 0
DEBUG: [0xf78cf040] UpdateGestures()

DEBUG: [0xf78cf040] Start Main Loop Gesture = 0
DEBUG: [0xf78cf040] UpdateGestures()

DEBUG: [0xf78cf040] Start Main Loop Gesture = 0
DEBUG: [0xf78cf040] UpdateGestures()

DEBUG: [0xf78cf040] Start Main Loop Gesture = 0
DEBUG: [0xf0155440] processGestureEvent() current = 1
DEBUG: [0xf0155440] processGestureEvent() current = 1
DEBUG: [0xf0155440] processGestureEvent() current = 1
DEBUG: [0xf0155440] processGestureEvent() current = 1
DEBUG: [0xf78cf040] UpdateGestures()

DEBUG: [0xf78cf040] Start Main Loop Gesture = 4
DEBUG: [0xf78cf040] UpdateGestures()
```

If we look at the code for 'UpdateGestures()' we can see the first thing it does is change the TAP (1) to HOLD (4)

```
void UpdateGestures(void)
{
    // NOTE: Gestures are processed through system callbacks on touch events
    TraceLog(LOG_DEBUG, "[%p] UpdateGestures()", pthread_self());
    // Detect GESTURE_HOLD
    if (((GESTURES.current == GESTURE_TAP) || (GESTURES.current == GESTURE_DOUBLETAP)) && (GESTURES.Touch.pointCount < 2))
    {
        GESTURES.current = GESTURE_HOLD;
        GESTURES.Hold.timeDuration = rgGetCurrentTime();
    }

    // Detect GESTURE_NONE
[...]

PR #3502, opened 2023-11-02
[core] Partial fix gestures for `PLATFORM_DRM`

ubkp:
### Changes
1. The issue with the gesture system on `PLATFORM_DRM` is that `ProcessGestureEvent(gestureEvent)` is called by `EventThread()` (LINE) that runs on its own thread. By the time `PollInputEvents()` calls `UpdateGestures()` (LINE) the `GESTURE.current` has been overwritten. That causes the meaningful gestures to not be detected when calling `GetGestureDetected()` (LINE) or `IsGestureDetected()` (LINE).

2.  This partial fix attempts to address that by moving the `ProcessGestureEvent(gestureEvent)` call from `EventThread()` (LINE) to `PollInputEvents()` (LINE), where it runs consistently and in a predictable manner.

3. It does that by making `GestureEvent gestureEvent` a global variable (LINE) that will hold the data from the gesture generated during `EventThread()`. That `gestureEvent` will then be used inside `PollInputEvents()` when calling `ProcessGestureEvent(gestureEvent)` (LINE).

4. To only run `ProcessGestureEvent(gestureEvent)` when there's a new gesture, a new control var `newGesture` was added (LINE) and is reseted after the `ProcessGestureEvent(gestureEvent)` call (LINE).

5. This fixes the gestures detection for `GESTURE_TAP`, `GESTURE_DOUBLETAP` and `GESTURE_HOLD`.

6. The `GESTURE_DRAG` and `GESTURE_SWIPE_*` work, but are not reliable. There's some other issue with the touch position or reset that is causing them to not be precise as on other platforms. I believe that's related to the way they are pulled from evdev inside `EventThread()` (LINE), but I was not able to finish debugging it yet.

### Note
- @raysan5 Unfortunately compiling on the Raspberry Pi is slow and trying to debug with the framebuffer hiding the log/prints makes progress even slower. I may need a lot more time to debug what's causing `GESTURE_DRAG` and `GESTURE_SWIPE_*` to malfunction, so I don't think I'll be able to get this fully fixed fast enough for the `5.0` release. This PR, however, should at least restore basic gesture operation for `PLATFORM_DRM` in the meantime.

### Reference
- https://github.com/raysan5/raylib/issues/3335

### Environment
- Tested on Raspberry Pi 3 Model B 1.2 with Linux (Raspberry Pi OS 11 Bullseye 32-bit armhf).

### Edits
- **1:** added line marks.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -170,6 +170,11 @@ static const int EvkeyToUnicodeLUT[] = {
     // LUT currently incomplete, just mapped the most essential keys
 };
 
+#if defined(SUPPORT_GESTURES_SYSTEM)
+GestureEvent gestureEvent = { 0 };      // Gesture event to hold data between EventThread() and PollInputEvents()
+bool newGesture = false;                // Var to trigger ProcessGestureEvent(gestureEvent) on PollInputEvents()
+#endif
+
 //----------------------------------------------------------------------------------
 // Module Internal Functions Declaration
 //----------------------------------------------------------------------------------
@@ -592,6 +597,18 @@ void PollInputEvents(void)
     // Reset touch positions
     //for (int i = 0; i < MAX_TOUCH_POINTS; i++) CORE.Input.Touch.position[i] = (Vector2){ 0, 0 };
 
+    // Map touch position to mouse position for convenience
+    CORE.Input.Touch.position[0] = CORE.Input.Mouse.currentPosition;
+
+#if defined(SUPPORT_GESTURES_SYSTEM)
+    // Call the ProcessGestureEvent here instead of on EventThread() to workaround the threads not matching
+    if (newGesture)
+    {
+        ProcessGestureEvent(gestureEvent);
+        newGesture = false;
+    }
+#endif
+
 #if defined(SUPPORT_SSH_KEYBOARD_RPI)
     // NOTE: Keyboard reading could be done using input_event(s) or just read from stdin, both methods are used here.
     // stdin reading is still used for legacy purposes, it allows keyboard input trough SSH console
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp thank you very much for the review and the so detailed explanation!

Reading inputs in separate threads has that kind of downsizes, there should be some synchronization at a fixed rate to get the current state and process it per-frame, similar to other platforms... or just simplify the input and let some inputs run in the main thread.

In any case, that solution looks good to me, but maybe @michaelfiber can take a look before merging, he has more experience than me with `PLATFORM_DRM` and touch-displays where gestures could apply.

ubkp:
> or just simplify the input and let some inputs run in the main thread.

IMHO, if possible, this would be the best scenario. By the way, now that `SDL` is an option, couldn't `DRM` also use `SDL2` too? That would make development/maintenance for it a lot easier.

michaelfiber:
> @ubkp thank you very much for the review and the so detailed explanation!
> 
> Reading inputs in separate threads has that kind of downsizes, there should be some synchronization at a fixed rate to get the current state and process it per-frame, similar to other platforms... or just simplify the input and let some inputs run in the main thread.
> 
> In any case, that solution looks good to me, but maybe @michaelfiber can take a look before merging, he has more experience than me with `PLATFORM_DRM` and touch-displays where gestures could apply.

Sorry for my quietness recently. I'm following these but haven't had a chance to chime in.

This fix is very much in line with what I've been working on. I've been trying to find a down side to moving more input processing into the main thread but so far performance has been fantastic.

ubkp:
@raysan5 Got it! That's very good reasoning. Thank you very much for the explanation! :+1:

PR #3506, opened 2023-11-04
Allow raylib to be built with zig 0.11.0 and zig 0.12.0dev (master)

freakmangd:
Tested on linux, shouldn't cause any problems for other operating systems


raysan5:
@freakmangd thanks for the review, please note that I'm not validating this PR, hopefully some Zig users could confirm it works as expected and nothing breaks.

PR #3508, opened 2023-11-05
Update info about Jaylib

glowiak:
It got updated to 4.5 overnight, let's include latest information.


raysan5:
@glowiak thanks for the update!

PR #3509, opened 2023-11-05
Update Makefile: Build shared lib when platform is SDL

JupiterRider:
When using `PLATFORM=PLATFORM_DESKTOP_SDL` and `RAYLIB_LIBTYPE=SHARED`, it didn't build the library (.so file).


raysan5:
@JupiterRider Thanks for the review!

PR #3510, opened 2023-11-05
BeginScissorMode checks for render texture to avoid using GetWindowScaleDPI

gulrak:
When on a window with active `FLAG_WINDOW_HIGHDPI` window flag or on macOS the implementation of `BeginScissorMode` uses `GetWindowScaleDPI` to get scaling and uses it on the scissor rect.

If using `BeginScissorMode` inside of an open `BeginTextureMode`/`EndTextureMode` this scaling is
wrong as a `RenderTexture` is created with a defined resolution and doesn't share any implicit window dpi scalings.

I fixed this by checking if currently a frame buffer is bound to detect the texture mode. If there is a binding the else path of the original implementation will be used to avoid the wrong scaling.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1085,20 +1085,25 @@ void BeginScissorMode(int x, int y, int width, int height)
 
     rlEnableScissorTest();
 
+    GLint id = 0;
+    glGetIntegerv(GL_FRAMEBUFFER_BINDING, &id); // Test for render texture
 #if defined(__APPLE__)
-    Vector2 scale = GetWindowScaleDPI();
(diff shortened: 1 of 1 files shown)


raysan5:
@gulrak Thanks for the review! Just added `CORE.Window.usingFbo` flag to register when an FBO is being used, we can check against it. I'm merging and reviewing it myself!

gulrak:
Sorry, I was so happy for it to be merged that I missed you flipped the logic when changing the way the texture mode is detected, now it always uses GetWindowScaleDPI if we are in texture mode, but my solution (and the one needed) was the opposite.

PR #3511, opened 2023-11-06
[build]: add uninstall cmake target

icy-comet:
The handwritten `Makefile` in `src` [has](https://github.com/raysan5/raylib/blob/a470cdce4c3d77e543d90a5fc2aab05e8d41b51b/src/Makefile#L764) a `uninstall` target. Auto-generated `Makefile` doesn't get one as [mentioned](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake) in the CMake's FAQs. So adding for completeness' sake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,7 @@ CMakeFiles
 CMakeScripts
 Testing
 cmake_install.cmake
+cmake_uninstall.cmake
 install_manifest.txt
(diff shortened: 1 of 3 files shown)


raysan5:
@icy-comet thanks for the improvement! 🙂

PR #3514, opened 2023-11-07
Add ES 3.0 to CMakeOptions

awfulcooking:
The ES 3.0 support seems viable:

Compiled with Emscripten -s FULL_ES3

But not offered in the `OPENGL_VERSION` option

It seems like a good single target for mobile, desktop and web.

This is an invitation to expose it :-)


raysan5:
@awfulcooking thanks for the improvement!

PR #3518, opened 2023-11-09
WindowShouldClose(), reset shouldClose flag when window is inited

maksut:
Similar issue with: https://github.com/raysan5/raylib/pull/3410

Window shouldClose flag does not reset after closing and opening another window. So for the subsequent windows `WindowShouldClose` always return true.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -658,6 +658,7 @@ void InitWindow(int width, int height, const char *title)
 #endif
 
     CORE.Time.frameCounter = 0;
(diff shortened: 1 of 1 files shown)


maksut:
This was a regression with platform split work. In version 4.5.0, desktop platform even reset the flag just after the WindowShouldClose. So it was possible to ignore the flag and continue next frame. But that was not consistent with other platforms. So I hope this little change is fine.

raysan5:
@maksut thanks for the review!

PR #3519, opened 2023-11-10
[Examples - Core] Automation events - browser shortcut/keybind clash

mmilenkov:
Most browsers have shortcuts assigned to the F keys, which causes problems when running the example.
Moved the keys used in the example to:
- S for start/stop 
- A for automate


raysan5:
@mmilenkov thank you very much for the review! looks great to me!

Issue #3372, opened 2023-10-03
[core] Very low rendering quality on 4K displays on `PLATFORM_WEB`


ypujante:
I have been trying to port my application [RE Edit](https://github.com/pongasoft/re-edit) which is a raylib/ImGui application to the web platform to offer a demo (since there is no filesystem access available I cannot really run it in a browser until it is available in wasm...)

I have a prototype running and it looks quite bad:

* On macOS (showing web on the left, native on the right):
<img width="348" alt="macOS" src="https://github.com/raysan5/raylib/assets/357116/012cb898-fbda-4254-843b-b3dc1a889e2b">

* On Windows (showing that the resolution is 4K on the left, web on the right):
<img width="643" alt="Windows" src="https://github.com/raysan5/raylib/assets/357116/d1cc011e-d91b-4e77-b92f-bc50ca7c5c47">

I have created a small reproducible test case to show the issue which uses raylib only (top is web, bottom is native):

<img width="487" alt="test_case" src="https://github.com/raysan5/raylib/assets/357116/62c107f9-de21-4f7c-aae2-d3089e536bf5">

The source code for the test case is this:

```cpp
#include "raylib.h"

int main()
{
  InitWindow(800, 450, "raylib [core] example - basic window");

  SetTargetFPS(60);

  while (!WindowShouldClose())
  {
    BeginDrawing();
    ClearBackground(RAYWHITE);
    DrawText("The quick brown fox jumps over the lazy dog", 10, 10, 20, BLACK);
    DrawText("Congrats! You created your first window!", 10, 40, 20, LIGHTGRAY);
    DrawFPS(10, 70);

    EndDrawing();
  }

  CloseWindow();

  return 0;
}
```

Is this a bug in raylib? A bug in glfw3? Or a misconfiguration? I have followed the steps described [here](https://github.com/raysan5/raylib/wiki/Working-for-Web-(HTML5)) but there isn't any mention of the quality of the rendering nor how to make it look better on 4K displays.

ubkp:
@ypujante Could you please test it with `minshell.html` instead of `shell.html`. `shell.html` has a `width:100%;` CSS which is likely affecting the scaling.

ypujante:
@ubkp I would gladly test it but I don't know what you mean by using `minshell.html`. The compilation produces this set of files:

```
-rw-r--r--@  1 ypujante  staff     1678 Oct  3 09:58 raylib_web_cmake.data
[...]

PR #3520, opened 2023-11-10
Changes in preparation for Issue 3372

ypujante:
Following our discussion on #3372, I am opening this PR.

The idea is that raylib can safely use `glfwGetFramebufferSize` to retrieve the correct frame buffer size. Before my emscripten [PR is merged](https://github.com/emscripten-core/emscripten/pull/20584), it simply returns `width` and `height` (see [code](https://github.com/emscripten-core/emscripten/blob/93b21352e4ea92dcae2c679919ecad7fa8c214e4/src/library_glfw.js#L1368)) so the screenScale will be 1.

The change in `WindowSizeCallback` is to follow what happens during `InitPlatform`
The change in `EmscriptenResizeCallback` is to simply call `SetWindowSize` as it will trigger the callback. Setting the canvas size explicitly (which is what is done prior to this PR) is in my opinion not the right thing to do anyway because it bypasses glfw.

Once the emscripten changes have been pushed, 4k is enabled by doing something like this:

```cpp
EM_ASM(
Module.setHiDPIAware(true);
);
```

which can be done directly in the client code. I can imagine that there could be a convenient API provided by raylib so that client code does not have to write something like this. It could also be tied to the flag `FLAG_WINDOW_HIGHDPI` if that was the path you would like to follow.

Although I did test my changes on my (own) local example, I would be more than happy to run deeper tests if you let me know how you proceed when you make this kind of changes so as to make sure I did not break anything.

Thank you


ypujante:
I also want to add that if you would prefer to wait until the changes are pushed to emscripten, it is obviously no issue with me. At this stage I have no clue if/when it will be merged as it has been in "waiting for somebody to approve" for several days now.

ubkp:
@ypujante Noticed that the previous/current `WindowSizeCallback()` and `EmscriptenResizeCallback()` had breaks if fullscreen was active ([L976](https://github.com/raysan5/raylib/pull/3520/files#diff-14c6a8a6962cf65222635a523e877bbaa1f07b791028685f92bf6b44d87530d7L976), [L1210](https://github.com/raysan5/raylib/pull/3520/files#diff-14c6a8a6962cf65222635a523e877bbaa1f07b791028685f92bf6b44d87530d7L1210)). Have you tested/considered this scenario?

ypujante:
@ubkp in my local test it seems to work, but I would definitely feel a lot more confident if I could test it with a piece of code I did not write to make sure that it works. Could you point me to a program/test that I can experiment with fullscreen mode?

ypujante:
@ubkp thank you for your feedback. I will double check that fullscreen works as you point out. Note that [emscripten](https://github.com/emscripten-core/emscripten/blob/93b21352e4ea92dcae2c679919ecad7fa8c214e4/src/shell.html#L18) provides a `Module.requestFullscreen` api so it would be better to invoke this rather than `document.getElementById('canvas').requestFullscreen()` because that is bypassing what emscripten does for fullscreen

ypujante:
I have spent some more time looking at the (raylib) code and some are my preliminary findings (note that all my findings are based on the **main** branch latest as of this writing a92c34d, **not** my changes)

* First of all, the fact that now each platform is its own separate file make it about a million times easier to reason about the code so thank you for this effort!
* Digging into the low level details I realized one thing that I guess I had missed/misunderstood: raylib web essentially uses glfw to create the window and that is pretty much it. The proof is that running the example core_window_letterbox with a slight addition generates this:
    Screenshot 2023-11-11 at 09 47 02
* No matter how much I resize the window, make it fullscreen or whatnot, the size, as reported by glfw never changes (`glfwGetWindowSize`)
* Clicking on the "Fullscreen" button also does not invoke any glfw callback because `false` is provided for the `resize` parameter. Note that this would change if my emscripten changes are ever merged in...
* As I pointed out in a previous comment, the `EmscriptenResizeCallback` bypasses glfw by calling this `emscripten_set_canvas_element_size("#canvas", width, height)` instead of `SetWindowSize(width, height)`. As a result glfw and raylib becomes disconnected in what is the size of the window.
* As another result, the callback `WindowSizeCallback` is never called (when executing `core_window_letterbox`) despite changing the window size
* The call `SetWindowSize` is implemented as `glfwSetWindowSize(platform.handle, width, height)` so that would be the only case when glfw is involved and of course after this call then `glfwGetWindowSize` returns the proper size.

I am not saying that bypassing glfw is a bad thing, but I feel like it should be consistent: if the `EmscriptenResizeCallback` bypasses glfw, then `SetWindowSize` should bypass it as well and essentially implement a logic similar to `EmscriptenResizeCallback` for consistency. Of course, the catch is that `Module.requestFullscreen` which is called when the `Fullscreen` button is clicked ends up going through the glfw layers...

I believe my changes are going the opposite direction: using glfw as the source of truth when it comes to size.

I will follow back with more findings...

ypujante:
As you can see I have done a [follow up PR](https://github.com/raysan5/raylib/pull/3533)

raysan5:
@ypujante thanks for the detailed analysis, afair, `EmscriptenResizeCallback` was implemented to overcome the limitations of GLFW implementation for Web (`library_glfw.js`). If this implementation can be more aligned to GLFW desktop one, that's great but also note that a browser canvas and a desktop windows work in a very different way, I can't see how that functionality can be unified.

Also, I personally would love to completely replace GLFW dependency on Web and just use the functionality already exposed by Emscripten, I think going through that middle-layer will generate more trouble than benefit in the future.

ypujante:
@raysan5 I would be happy to take a look at "_replace GLFW dependency on Web and just use the functionality already exposed by Emscripten_" if you want. That being said I don't want to spend a huge amount of time on this and then be turned down because the "_PLATFORM_WEB is currently stable and that represent too many changes_"

Just want to be clear about where I am standing...

I am closing this PR as well.

orcmid:
Before making extensive changes, it is valuable to always check first, say with an issue the allows the changes to be anticipated and considered before you do work that ends up not accepted.  And if the changes can be done in understandable smaller stages, even better.

ypujante:
@orcmid For the record, this PR as well as the other one, were not "out of the blue" but based on issue #3772 that was open well over a month ago including 53 comments and I have been working on it almost this entire time as I was getting feedback along the way. Also for the record, I was not expecting either PR, in their state, to be accepted. I understand that this is not my project and that the owner has the burden of maintaining it after the fact. I also believe that waiting for my emscripten changes to be merged will result in less changes to raylib which is obviously preferable for this project. 

Like I said, if @raysan5 wants me to take a look at "_replace GLFW dependency on Web and just use the functionality already exposed by Emscripten_" given the amount of time I have already spent looking at emscripten, I think it might be easier for me and I would be more than happy to do it given his blessing.

raysan5:
@ypujante At this moment I'm not aware of the cost and implications of replacing the GLFW usage by a direct Emscripten usage, if it simplifies the current implementation and reduces the amount of code and complexity (replacing the GLFW intermediate callbacks), I would go for it. 

If you want to take a look to it and evaluate how it can be accomplished, you are welcome.

PR #3523, opened 2023-11-11
[core] Fix automation recording

### LINKED ISSUE #3521: [example] `core_automation_events` not working properly on Web
This is a difficult to track issue, I've been investigating it for several hours.

It seems the recording works as expected but the replay does not.

First approach was to the timming system, considering the asynchronous execution on Web, I tried refactoring code to avoid ASYNCIFY ([core_automation_events_web.c.md](https://github.com/raysan5/raylib/files/13322524/core_automation_events_web.c.md)) but it does not seem to be the issue. Also considering that events recording and playing is frames-based. The jump physics on the example confused me but trying just a simple right-left movement also fails.

Investigating a bit further, I noticed there was not call in `rcore_web` for `glfwPollEvents()`, that call on `rcore_desktop` is the one in charge of launching the callbacks that read the inputs. In web it seems is not needed because probably the input reading is also asynchronous, managed by the browser.

As there are no callbacks involved in the automation events process, the key states do not seem to be properly reseted (key_up event) and they are just fixed and played forever, that's my guess.

Thinking about possible approaches to this issue, maybe a separate keys reseting mechanism on automation events playing for web?

Help with this issue is really welcome!

ubkp:
### Changes
- Moves `RecordAutomationEvent()` from [L931-L933](https://github.com/raysan5/raylib/pull/3523/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339L931-L933) to [R868-R870](https://github.com/raysan5/raylib/pull/3523/files#diff-39fea0e3daa23741e81daf967717f35c7842426aab26a0bc3f525d6a2f03e339R868-R870).
- Logic: [3521#issuecomment-1806829131](https://github.com/raysan5/raylib/issues/3521#issuecomment-1806829131)
- Fixes #3521

### Environment
- Tested on `PLATFORM_WEB` and `PLATFORM_DESKTOP`.

### Edits
- **1:** added line marks.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -865,6 +865,10 @@ void EndDrawing(void)
     }
 #endif
 
+#if defined(SUPPORT_AUTOMATION_EVENTS)
+    if (automationEventRecording) RecordAutomationEvent();    // Event recording
+#endif
+
 #if !defined(SUPPORT_CUSTOM_FRAME_CONTROL)
     SwapScreenBuffer();                  // Copy back buffer to front buffer (screen)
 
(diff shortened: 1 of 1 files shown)


raysan5:
@ubkp thanks for the fix!

PR #3524, opened 2023-11-11
Add comments for `LoadRandomSequence`

SergeyMC9730:


raysan5:
@SergeyMC9730 Sorry, I'm afraid those comments are not really necesary...

orcmid:
@SergeyMC9730 is there a reasonable explanation for producing a sequence of count random values that have no duplicates?  That rather removes the randomness.

The only randomization case I can think of is to have count be the first count values of a random shuffle of a very large array of sequential numbers in the specified range.  

What's the intended application for this, discarding fine details about the nature of such randomness?

PR #3529, opened 2023-11-11
[rlgl] Fixed return type of `rlGetCompressedFormatName`

Bigfoot71:
The return type of `rlGetCompressedFormatName` has been corrected from `char *` to `const char *`


raysan5:

PR #3531, opened 2023-11-12
Update BINDINGS.md for raylib-go

JupiterRider:
raylib-go uses raylib 5.0: https://github.com/gen2brain/raylib-go/blob/master/raylib/raylib.h


raysan5:
@JupiterRider Wow! That was fast! `raylib 5.0` is not published yet! 😄

PR #3532, opened 2023-11-12
Implement SDL Audio Backend in Platform_Desktop_SDL

PretzelBrosStudios:
**Problem:**

The graphics backend "Platform_Desktop_SDL" now supports several platforms that were previously unsupported, such as consoles. However, the audio backend "raudio" is currently incompatible with some of these new platforms.

**Solution:**

Miniaudio allows the implementation of custom audio backends. To enable audio playback on all platforms supported by SDL, including consoles, SDL has been integrated as a custom backend in "raudio". This implementation should facilitate audio playback on all SDL-supported platforms.

**Notes:**

A new file named "backend_sdl.h" has been added to the "/external" directory. This backend is primarily based on the SDL backend sample code provided by Miniaudio. The file "raudio.c" has undergone modifications to enable the use of the SDL backend. Currently, the SDL audio backend is activated when "PLATFORM_DESKTOP_SDL" is defined. This can be easily changed to something like "PLATFORM_AUDIO_SDL" if it is preferred to make the audio backend independent of the graphical backend.

**Problems:**

I encountered some issues related to threading. When using SDL as a custom backend, the device must be started before creating the mutex. Previously, "raudio" also started the device before creating the mutex. This order was altered in the most recent version of "raudio", where the mutex is now created before the device is started. Is a specific reason for this change?

**Testing:**

I have successfully tested the new SDL backend implementation on both Windows 11 and a Nintendo Switch Developer Console. Additional testing is needed.

And thank you for developing raylib, a truly fantastic library!

Best regards,
Alex

### DIFF
--- a/src/external/backend_sdl.h
+++ b/src/external/backend_sdl.h
@@ -1,0 +1,563 @@
+/**********************************************************************************************
+*
+*	SDL Backend for raudio
+*	based on the miniaudio example "custom_backend.c" by David Reid 
+*	David Reid - mackron@gmail.com
+*	GitHub: https://github.com/mackron
+*
+*	Modified by Alex Schaeferling
+*	Alex Schaeferling - alex.schaeferling@gmail.com
+*	
+*	This software is available as a choice of the following licenses. Choose
+*	whichever you prefer.
+*
+*	===============================================================================
+*	ALTERNATIVE 1 - Public Domain (www.unlicense.org)
+*	===============================================================================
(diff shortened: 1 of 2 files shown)


raysan5:
@PretzelBrosStudios This is a really big change that implies a considerable maintenance work that I can not assume. I'm afraid I'm not merging it. Anyone with the need of this backend could find this PR and implement it themselfs. Sorry for the inconvenience.

PR #3540, opened 2023-11-14
Fix tiny spelling mistake.

2Bear:


raysan5:
@2Bear thanks for the fix!

PR #3542, opened 2023-11-15
Fix for running OpenGL ES2 with SDL on desktop

realtradam:
System: Artix Linux, Desktop PC

---

When trying to compile Raylib configured with SDL and OpenGL ES2 it will fail like the following:

By editing `rcore_desktop_sdl.c` to not include `SDL_opengl.h` when compiling for OpenGL ES2 it compiles and works. See here:

However when looking through the headerfiles of SDL I see there is a `SDL_opengles2.h` which I suspect should be used in place of `SDL_opengl.h` but it fails to compile when I do that. I am not entire sure why, I tried to do some digging but havent come up with anything yet, if anyone else has any insight it would be much appreciated however I think the current fix is sufficient.

I have not tested this yet on Windows or Mac. I can test it on Windows and Mac most likely sometime later this week or next week.

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -49,7 +49,14 @@
 **********************************************************************************************/
 
 #include "SDL.h"            // SDL base library (window/rendered, input, timming... functionality)
+#ifdef GRAPHICS_API_OPENGL_ES2
+// I suspect that the gles2 version of the SDL header should be used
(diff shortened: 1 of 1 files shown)


raysan5:
@realtradam thanks for the review! Afaik, OpenGL ES 2.0 could be supported in different ways: 

 - On desktop, using the specific OpenGL extension for ES
 - On ES-specific platforms, directly using OpenGL ES

Probably this implementation could require some more work in the future...

PR #3543, opened 2023-11-16
Added glsl 100 and 120 shaders to lightmap example.

nullstare:
GLSL 100 and 120 shaders for lightmap example.


raysan5:
@nullstare thanks for the review!

nullstare:
You're welcome!

PR #3544, opened 2023-11-17
Disable unused miniaudio features

M374LX:
This PR just disables some high level miniaudio features that are not used by raylib.


raysan5:
@M374LX Very nice improvement! Actually it was on my TODO list to review that but I just forgot about it!

PR #3550, opened 2023-11-18
Preparing for 5.1-dev

orcmid:
Adjusting raylib.h to reflect post-5.0 changes now.  This is important for testing and verification that I am providing.


raysan5:
@orcmid thanks for the review! Sincerely, not sure if there will be another release after 5.0 but this change is good to avoid confusions.

PR #3551, opened 2023-11-18
BINDINGS: Update raylib-cpp support to raylib 5.0

RobLoach:
[raylib-cpp](https://github.com/RobLoach/raylib-cpp) supports raylib 5.0. Congrats on the release! :partying_face:


raysan5:
@RobLoach Wow! That was fast! Thanks for the update!

PR #3556, opened 2023-11-20
Added png preview for textures_image_kernel example

Zapunidi:
textures_image_kernel was missing preview.png
I decided to fix that and made 800x450 px (like the others) preview png with GIMP. It is bigger than any other preview file (0.77MB) because it is filled with photo data and png format is more suitable for charts and logos.


raysan5:
@Zapunidi Please note that screenshot should be taken using F12. If you run the program and press F12, a screenshot01.png is automatically saved in the same directory as the executable.

Zapunidi:
I remade the screenshot according to the requirements. It is about 1 MB in size now.

raysan5:
@Zapunidi I just reviewed some formating issues in the example and added the screenshot in the process, so no need for this PR. In anyc ase, thank you very much!

PR #3558, opened 2023-11-21
Fix version info in rlgl.h

schveiguy:
Doing the binding for 5.0, I noticed this seems to not have been changed.

Most of the .h files don't follow the release versions, so I'm not 100% sure if this is correct...


raysan5:
@schveiguy Actually `rlgl` versioning was not originally intended to be aligned with raylib... but considering the coupling with raylib, it's probably a good idea to keep them aligned.

PR #3561, opened 2023-11-21
Update raylib-d binding

schveiguy:
Unsure why some are bold and some are not...


raysan5:
@schveiguy thanks for the update! Bold is meant for the latest version...

PR #3564, opened 2023-11-22
Update `raylib_api.*`

blueloveTH:
Update api metadata for raylib 5.0


raysan5:
@blueloveTH Please note that you generated the files from latest master branch that is not `raylib 5.0` but `raylib 5.1-dev`, those version numbers should be reviewed.

PR #3567, opened 2023-11-22
add ColorFromInt for convenance

danilwhale:
because there's `ColorFromNormalized` -> `ColorToNormalized`, `ColorFromHSV` -> `ColorToHSV` 

**but** `GetColor` -> `ColorToInt`, not `ColorFromInt` -> `ColorToInt`


Peter0x44:
This is inconsistent.
GetColor returns int, while ColorToInt takes an unsigned int.
I think these functions are a little broken anyway, since the exact representation of (unsigned) int depends on endianness.
probably they deserve a review.

raysan5:
@danilwhale I prefer to avoid that function duplicate, despite the consistency it could add. Same behaviour can be accomplished with a `#define GetColor ColorFromInt` but probably as @Peter0x44 commented all those functions mentioned should be reviewed, I don't like the reference to a specific language data type in the name of the functions, that's counter-productive for other language bindings. Naming for those functions should be reviewed.

PR #3572, opened 2023-11-23
Expose glad functions when building raylib as a shared lib

Peter0x44:
Currently, if you build raylib as a DLL, and then try to build the examples, examples/other/raylib_opengl_interop.c will fail to link, because none of the opengl functions are exposed in raylib.dll
These can also help some binding users, like raylib-cs users who would like to call some opengl functions directly, bypassing rlgl.
Currently these users have to call wglGetProcAddress to get a function pointer manually.

examples/other/rlgl_standalone.c also has this issue, and it can be fixed by exposing GLFW. I don't know whether this is a good idea, though, since GLFW is not used by all platforms, while OpenGL is.
I'm happy to open a PR exposing those symbols too, if it's desired, though.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -773,6 +773,11 @@ RLAPI void rlLoadDrawQuad(void);     // Load and draw a quad
 
 #if defined(RLGL_IMPLEMENTATION)
 
+// Expose OpenGL functions from glad in raylib
+#if defined(BUILD_SHARED_LIBS)
+    #define GLAD_API_CALL_EXPORT_BUILD
+#endif
+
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 the change seems minimal and at this moment I can't think of any drawback, so merging it.

Peter0x44:
@raysan5 should I do the same for GLFW? the particular example not building with the shared lib is not really important, but it could still be useful to some users. Maybe they would just be better served by building raylib with an external GLFW though.

I'm still conflicted, leaning a little more towards "no"

raysan5:
@Peter0x44 I prefer to avoid those platforn-specific symbols.

PR #3573, opened 2023-11-24
Hide unneeded internal symbols when building raylib as an so or dylib

Peter0x44:
Currently, raylib makes no attempt to hide any of its symbols when building as an so or dylib.
This means the symbols of glfw, miniaudio, stb_image, etc are all exposed to call.
This results in a bit of inconsistency compared to the windows dll, which exposes far less (just raylib+glad+raymath+rlgl).

This PR handles this by passing -fvisibility=hidden to all of raylib's source files when building as an SO, and using \_\_attribute__((visibility("default"))) to expose each symbol, much like __declspec(dllexport).

In theory, this improves the load times and size of the so, as well as providing better "api hygine"
https://gcc.gnu.org/wiki/Visibility

I updated both CMake and the Makefile to do this. I checked zig, but it does not seem to handle building shared libs at all, so I don't think it needs any changes.

NOTE: I did not test BSD or OSX dylibs, but as far as I can tell, the changes here should do the same thing on both platforms.
The risk of breaking anything is fairly low, though testing is welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.0)
 project(raylib)
 
 # Avoid excessive expansion of variables in conditionals. In particular, if
-# "PLATFORM" is "DRM" than:
+# "PLATFORM" is "DRM" then:
 #
 # if (${PLATFORM} MATCHES "DRM")
 #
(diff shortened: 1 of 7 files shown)


raysan5:
@Peter0x44 Interesting change, thanks for the review.

Trying to think if it could suppose any future issue...

PR #3574, opened 2023-11-24
Corrected the path of android ndk toolchains for OSX platforms

ManuMario0:
The android ndk toolchain path for OSX is toolchains/llvm/prebuilt/darwin-x86_64 and not toolchains/llvm/prebuilt/linux-x86_64


raysan5:
@ManuMario0 thanks for the review!

Issue #3569, opened 2023-11-23
[rcore] `IsKeyPressed()` not working for some keys on RPI (`PLATFORM DRM`)


larry104:
### Discussed in https://github.com/raysan5/raylib/discussions/3565

<div type='discussions-op-text'>

<sup>Originally posted by **larry104** November 22, 2023</sup>
I notices that the block of keys (insert, delete home, end, pup, pdown) are not working. IsKeyPressed is not getting any event for these keys (I tried different keyboards and started the application from ssh terminal vs native terminal no difference). All other keys are working fine. Does anyone have an idea what is going wrong ?</div>

larry104:
Just to summarize: keys insert delete, home, end, page up, page down are not implemented in on the RPi (PLATFORM DRM in rcore_drm.c: 

https://github.com/raysan5/raylib/blob/a3a73b9332d0ca0d7ec81044edce31dd13d7c916/src/platforms/rcore_drm.c#L1333-L1375

only 3 byte and 5 byte scan codes for F keys and cursor keys are implemented:  here are the missing 4 byte codes:

insert
^[[2~    27 0033 0x1b
         91 0133 0x5b
         50 0062 0x32
        126 0176 0x7e

delete
^[[3~    27 0033 0x1b
         91 0133 0x5b
         51 0063 0x33
        126 0176 0x7e

home
^[[1~    27 0033 0x1b
         91 0133 0x5b
         49 0061 0x31
        126 0176 0x7e

end
^[[4~    27 0033 0x1b
         91 0133 0x5b
         52 0064 0x34
        126 0176 0x7e

page up
^[[5~    27 0033 0x1b
         91 0133 0x5b
         53 0065 0x35
        126 0176 0x7e

page down
^[[6~    27 0033 0x1b
         91 0133 0x5b
         54 0066 0x36
        126 0176 0x7e

larry104:
It doesn't work with raylib 4.5 ... is that something which might have been fixed in 5.x ?. 

I tired from ssh shell as well as native shell. (bullseye, raspberry pi zero

`./raylibtest1
INFO: Initializing raylib 4.5
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
[...]

PR #3583, opened 2023-11-30
[core] Fix `IsKeyPressedRepeat()` for `PLATFORM_DRM` direct input

ubkp:
### Changes
- Fixes `IsKeyPressedRepeat()` for `PLATFORM_DRM` direct input by adding the missing `CORE.Input.Keyboard.keyRepeatInFrame` to `PollKeyboardEvents()` ([R1712](https://github.com/raysan5/raylib/pull/3583/files#diff-757f4c9c5f0be7473e385fa72b35d509c5af4ee2dc8204d2dbe2418bb9b98922R1712)).

### Reference
- Issue spotted by @larry104 at [#3569#issuecomment-1832811076](https://github.com/raysan5/raylib/issues/3569#issuecomment-1832811076).

### Code example
```
#include "raylib.h"

int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);
        BeginDrawing();
        ClearBackground(RAYWHITE);

        if (IsKeyPressedRepeat(KEY_ONE)) {
            DrawText("1 is repeat", 20, 20, 20, BLACK);
            TraceLog(LOG_INFO, "1 is repeat");
        }
        if (IsKeyPressedRepeat(KEY_TWO)) {
            DrawText("2 is repeat", 20, 40, 20, BLACK);
            TraceLog(LOG_INFO, "2 is repeat");
        }

        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### Environment
- Tested on Raspberry Pi 3 Model B 1.2 (2015) with Linux (Raspberry Pi OS 12 Bookworm 32-bit armhf).

### Edits
- 1: added line marks.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -1709,6 +1709,7 @@ static void PollKeyboardEvents(void)
                     // Event interface: 'value' is the value the event carries. Either a relative change for EV_REL,
                     // absolute new value for EV_ABS (joysticks ...), or 0 for EV_KEY for release, 1 for keypress and 2 for autorepeat
                     CORE.Input.Keyboard.currentKeyState[keycode] = (event.value >= 1)? 1 : 0;
+                    CORE.Input.Keyboard.keyRepeatInFrame[keycode] = (event.value == 2)? 1 : 0;
                     if (event.value >= 1)
                     {
                         CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = keycode;     // Register last key pressed


raysan5:
@ubkp thanks for the review!

larry104:
I just compiled against the latest master branch but it seems the fix is not working properly. If I for instance press the down cursor sometimes it does something and sometimes not. Also the repeat interval seem randomly sometimes fast sometimes slow. As I mentioned the repeat works via ssh keyboard even when just using "IsKeyPressed" and it's super reliable.

larry104:
Maybe it has to do with my app ... it seemed smoother with an ssh keyboard.
Is the IsKeyPressedRepeat supposed to detect also a single press? Or do I need something like this?
`if (IsKeyPressedRepeat(KEY_UP) || IsKeyPressed(KEY_UP)) {..}`

PR #3585, opened 2023-11-30
Implement miters in DrawSplineLinear

Toctave:
Hi,

I implemented segment miters for linear splines, meaning that the lines go from this : 

Screenshot from 2023-11-30 11-09-39

to this :

Screenshot from 2023-11-30 11-07-37

This breaks when the angle between two consecutive line segments is close to 0° (works fine when it's 180° though), so I'm not sure it should stay as a replacement to DrawSplineLinear, but a DrawSplineLinearMitered could maybe be created instead ?

I feel like the usual use case is to have reasonably small angles, in order to draw arbitrary, nice-looking lines, but this is probably not worth breaking existing code. 

Let me know what you think!

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1546,16 +1546,20 @@ void DrawPolyLinesEx(Vector2 center, int sides, float radius, float rotation, fl
 // Draw spline: linear, minimum 2 points
 void DrawSplineLinear(Vector2 *points, int pointCount, float thick, Color color)
 {
-    if (pointCount < 2) {
+    if (pointCount < 2)
+    {
         return;
     }
     
     Vector2 prevNormal = (Vector2){-(points[1].y - points[0].y), (points[1].x - points[0].x)};
     float prevLength = sqrtf(prevNormal.x*prevNormal.x + prevNormal.y*prevNormal.y);
-    if (prevLength > 0.0f) {
+    if (prevLength > 0.0f)
+    {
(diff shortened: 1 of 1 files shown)


src/rshapes.c
```diff
@@ -1546,26 +1546,71 @@ void DrawPolyLinesEx(Vector2 center, int sides, float radius, float rotation, fl
 // Draw spline: linear, minimum 2 points
 void DrawSplineLinear(Vector2 *points, int pointCount, float thick, Color color)
 {
-    Vector2 delta = { 0 };
-    float length = 0.0f;
-    float scale = 0.0f;
+    if (pointCount < 2) {
```
raysan5, line 1549: Please, follow raylib code conventions, brackeds should be aligned.

raysan5:
@Toctave This is actually a very nice improvement! Same issue happens with `DrawSplineBezier*()` that also needs to be addressed (avoiding the use of `DrawSplineSegmentBezier*()`).

Also, somewhat related is the support of begin-end caps for the splines, a semi-circle could be added to make then more beautiful.

Toctave:
Thank you! I'm going to add it to Bézier lines as well then. Would you prefer me to change the existing functions (as I did so far), or to e.g. add Draw*Mitered functions instead ?

raysan5:
Just merging this improvement and reviewing it myself.

PR #3587, opened 2023-11-30
Fix mistake in pr #3572

Peter0x44:


raysan5:

PR #3588, opened 2023-11-30
Expose OpenGL take 2

Peter0x44:
For some reason, there are actually two macros needed to control this.
Yes, I tried with only one, both are needed.

Sorry for the iterations...


raysan5:
@Peter0x44 thanks for the review

PR #3595, opened 2023-12-02
[raymath.h] Explicitly zero-initialize structs in raymath functions

Valeron93:
When compiling using Apple-Clang 15.0.0 with -Werror -Wall -Wextra, there are a lot of warnings like that:
```
raymath.h:1062:26: error: missing field 'y' initializer [-Werror,-Wmissing-field-initializers]
Vector3 result = { 0 };
                     ^
```

Since raymath.h contains inline definitions of functions, I think it is a good idea to make it possible to compile raymath.h using C++ compiler with all warnings enabled. This pull request contains such fixes.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -379,7 +379,7 @@ RMAPI Vector2 Vector2Divide(Vector2 v1, Vector2 v2)
 // Normalize provided vector
 RMAPI Vector2 Vector2Normalize(Vector2 v)
 {
-    Vector2 result = { 0 };
+    Vector2 result = { 0.0f, 0.0f };
     float length = sqrtf((v.x*v.x) + (v.y*v.y));
 
     if (length > 0)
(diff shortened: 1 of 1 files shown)


orcmid (commented):
That's impressive.

Except for `float16`, would an `= {  0.0.f  }` have been good enough for the other cases? 

In the C++ compilation case I notice there is a complication in the **`RAYMATH_IMPLEMENTATION`** case.  No `extern "C" { ... }` adjustment is introduced.  This might matter for the `RMAPI  inline` case also.  I wonder if the file should simply be conditionally wrapped in the usual way.  Now my head is hurting :).

raysan5:
@Valeron93 I'm afraid I'm not merging this change, raylib uses everywhere `{ 0 }` to initialize to 0 all the members of any structure, and that's actually a valid initializer for the C99 standard, initializing every struct members is highly prompt to errors.

If `Apple-Clang 15.0.0` shows warnings on those cases, you are probably not compiling as C99 or the compiler is not C99 compliant.

Peter0x44:
@raysan5 Yes, this warning only shows up when raymath.h is included in C++, and the specific warning `-Wmissing-field-initializers` is enabled
I believe it's part of `-Wextra`

PR #3596, opened 2023-12-02
Introduce a way to add a 100% external custom platform / CMake

ypujante:
Let me start with a few things (in order to avoid misunderstanding):

* this is based on this [discussion/comment](https://github.com/raysan5/raylib/issues/3566#issuecomment-1830623956) 
* I closed another [PR](https://github.com/raysan5/raylib/pull/3594) because I was told it was not the right way to go about it
* if you want to close and forget all about it I am perfectly fine with it

The purpose of this PR is to implement a way for raylib to support platforms that are **NOT** embedded in raylib and as a result not maintained by @raysan5.

Goals (that I believe I have achieved with this PR)
* developing a new (external) platform should require 0 changes to raylib (the raylib repo) 
* it should be fairly simple for a platform developer to write an external platform (the actual code in the platform itself might be complicated, but that is besides the point...)
* it should be fairly simple for a user to use a platform that is not embedded in raylib
* changing raylib to support external/custom platforms should not affect the build/runtime/compile time performance of the embedded platforms
* there are no changes to current platforms

After looking at the source code, I believe there are 2 areas where there needs to be "hooks" for a platform to reside outside raylib:

* the source code: actual implementation of the platform + various defines to tailor the build
* the build files: which libraries to include, link options, compile options, etc... Note that for this PR I concentrated only on CMake because I do not know about Makefiles/zig. These other build frameworks could be added later (via other PR). 

The overall approach that I took is the following: use conventions as much as possible so that it is easy both from the user of the platform and the developer of the platform to integrate with raylib.

The "only" required pieces of information that raylib needs are:
* the fact that is a custom platform (added a new enum options "Custom")
*   where the platform is located on the file system. In CMake this translates into providing a `PLATFORM_CUSTOM_ROOT_DIR` variable.

The custom platform looks like this (by convention):
```text
root/cmake
       raylibPlatformCustomLibraryConfig.cmake [optional]
       raylibPlatformCustomTargetConfig.cmake  [optional]
       raylibPlatformCustomProjectConfig.cmake [optional]
root/src
       raylib_platform_custom.c [required]
       raylib_platform_custom.h [required]
```

* `cmake/raylibPlatformCustomProjectConfig.cmake`: optional and contain cmake definitions that needs to happen **before** `project()` is invoked
* `cmake/raylibPlatformCustomLibraryConfig.cmake`: optional and contain cmake definitions of libraries and generic option (like setting `CMAKE_STATIC_LIBRARY_SUFFIX`)
* `cmake/raylibPlatformCustomTargetConfig.cmake`: optional and contain cmake definitions that applies on the `raylib` target (ex: `target_link_options(raylib PRIVATE "-sUSE_GLFW=3")`)
* `src/raylib_platform_custom.c` is the implementation of the platform that gets included in `rcore.c` (similar to `platforms/rcore_desktop.c`)
* `src/raylib_platform_custom.h` is a new header file containing several definitions that are used in the code

Let me give an example of the last one: in rlgl.h there is this code:

```cpp
    #if defined(PLATFORM_DESKTOP) || defined(PLATFORM_DESKTOP_SDL)
```

so I changed it to

```cpp
    #if defined(PLATFORM_DESKTOP) || defined(PLATFORM_DESKTOP_SDL) || defined(PLATFORM_CUSTOM_ENABLE_OPENGL_ES2)
```

and `PLATFORM_CUSTOM_ENABLE_OPENGL_ES2` is a define that needs to be defined (or not) in the `src/raylib_platform_custom.h`

The name `PLATFORM_CUSTOM_ENABLE_OPENGL_ES2` as well as the other entries in this file have been inferred by what the `#if` is supposed to mean but not knowing the code, that was my best bet.

I understand this is a lot (to explain, but in the end **not** a lot of changes!)

I have another [branch](https://github.com/ypujante/raylib/tree/custom-platform) that you can checkout/clone which demonstrates 2 custom platforms and 2 examples of how to use them under the  [external](https://github.com/ypujante/raylib/tree/custom-platform/external) folder. 

As you will see the 2 custom platforms are simply clones of 2 built-in ones, but they get used as a custom/external platform, NOT a built-in one.

The PR in its current state, is fully functional. But obviously I am open to feedback and proposed changes, especially when it comes to naming and conventions. As stated previously, it works with CMake only.


ColleagueRiley:
I believe the example files should be `main.c` and not `main.cpp`, I'm pretty sure they're already written in C anyway.

PR #3598, opened 2023-12-03
Fix cmake-built libraylib.a to properly include GLFW's object files

Peter0x44:
I broke this in PR #3573 by accidentally removing too much

The examples still compiled fine so I didn't notice - my guess is that cmake was still adding a separate link to glfw manually.


raysan5:
@Peter0x44 ok, thanks for the review.

PR #3600, opened 2023-12-03
Fix Windows Hardcoding

ghost:
Compiles on Linux & co. now

And also consider getting rid of the partial busy loop as a default.
It makes 0 visual difference and just eats up CPU.
I think it has LESS input lag when it's off, but maybe that's just me, and also shows how little difference it makes that i don't even really know.
I know that's not part of this PR, but something to consider!

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -48,13 +48,13 @@
 *
 **********************************************************************************************/
 
-#include "SDL.h"                // SDL base library (window/rendered, input, timing... functionality)
+#include "SDL2/SDL.h"                // SDL base library (window/rendered, input, timing... functionality)
 
 #if defined(GRAPHICS_API_OPENGL_ES2)
     // It seems it does not need to be included to work
     //#include "SDL_opengles2.h"
 #else
-    #include "SDL_opengl.h"     // SDL OpenGL functionality (if required, instead of internal renderer)
+    #include "SDL2/SDL_opengl.h"     // SDL OpenGL functionality (if required, instead of internal renderer)
 #endif
 
 //----------------------------------------------------------------------------------


raysan5:
@mr-sihc thanks for the review. Note that the busy wait loop could be disabled with a flag is desired.

Peter0x44:
Actually, `#include "SDL.h"` is the official SDL-recommended way to do it.
I think you should review how you build raylib yourself, this change is not a good idea.

mr-sihc:
Its how its installed on Arch and everywhere else

Peter0x44:
That's not true
`pkg-config --cflags sdl2` returns `-I /usr/include/SDL2`

mr-sihc:
then the Makefile does it wrong, which i used, and the cmake file has outdated doesnt even list SDL as an option

raysan5:
@mr-sihc @Peter0x44 Just reverted this cahnge, checked the official SDL examples and they use `#include "SDL.h"`. Following the same pattern as the creators of SDL.

Peter0x44:
@mr-sihc Sorry, you're just wrong.
There's no other way to put it.
SDL2/SDL.h only works by accident, because your compiler automatically searches in /usr/include.
Both pkg-config and sdl2-config indicate that #include "SDL.h" is the intended and endorsed way to include the header.
It has nothing to do with "subprojects" or not, in fact it actually accommodates "subprojects" better to do it this way.

Peter0x44:
>  i like this, are you going to include this failsafe in the makefile?
I'm not sure. I do think it's a good idea, but I don't know if Ray would approve. It could work on Windows too, but it needs a posix-sh type shell, not cmd.exe.
A better idea might be to put it behind an:
```make
ifeq ($(PLATFORM_OS), $(filter $(PLATFORM_OS), LINUX BSD OSX))
# ...
endif
```
I'm unlikely to have enough time to formalize a PR for this, but I'm happy to review any such PR if you choose to submit it.

mr-sihc:
> > i like this, are you going to include this failsafe in the makefile?
> 
> I'm not sure. I do think it's a good idea, but I don't know if Ray would approve. It could work on Windows too, but it needs a posix-sh type shell, not cmd.exe. A better idea might be to put it behind an:
> 
> ```makefile
> ifeq ($(PLATFORM_OS), $(filter $(PLATFORM_OS), LINUX BSD OSX))
> # ...
> endif
> ```
> 
> I'm unlikely to have enough time to formalize a PR for this, but I'm happy to review any such PR if you choose to submit it.

Hmm, i'll see tomorrow if i could make it work.
But no promises i dont know much about makefiles xD!

If Ray even approves... Would this be a nice thing in the makefile?

PR #3602, opened 2023-12-04
Use System SDL2 As Failsafe

mr-sihc:
Attempt #2, but correctly solving it this time! (Until SDL3, where you don't have to worry about this, because they decided to then do it like every other library ever)

The code works, I tested (except the SDL_LIBRARY_PATH), I haven't on windows so if someone could, that still needs to be done.

Other than that, if there are issues with the code/looks bad, tell me.


mr-sihc:
> @mr-sihc Sorry, I can't see the need for those additions. It adds a considerable level of complexity to the Makefile to support a specific use case.

it's not actually very specific, SDL is not built standard and anyone who wants to, has to clone it, or change the dir

> What's the purpose of the shell if statement?

to check if the SDL_... path is not there, so to activate the failsafe

Peter0x44:
@mr-sihc I wouldn't bother checking

PR #3609, opened 2023-12-06
[core] Fix IsMouseButtonUp for PLATFORM_DESKTOP

### LINKED ISSUE #3606: [core] Problem with `IsMouseButtonUp` which always returns `true`
### Issue description

I noticed that the `IsMouseButtonUp` function always returns true with `PLATFORM_DESKTOP` on the master branch (https://github.com/raysan5/raylib/commit/984e83c2d0c6f75053a1ab533e12aece80e24b32).

### Environment

Linux Mint 21.2, 64-bit, Cinnamon

### Code Example

```c
#include "raylib.h"

int main(void)
{
    InitWindow(800, 450, "Button up test");

    while (!WindowShouldClose())
    {
        BeginDrawing();
        ClearBackground(WHITE);

        if (IsMouseButtonUp(MOUSE_LEFT_BUTTON))
        {
            DrawText("Button is UP", 10, 10, 20, BLACK);
        }
        else
        {
            DrawText("Button is DOWN", 10, 10, 20, BLACK);
        }

        EndDrawing();
    }

    CloseWindow();

    return 0;
}
```

**Edit**: Note that doing `if (!IsMouseButtonDown(MOUSE_BUTTON_LEFT))` works for me

Minmoose:
Fixes #3606

CORE.Input.Touch.currentTouchState[button] was never being set, causing a the `up` value in `IsMouseButtonUp()` to be overridden as `true` constantly.

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -1739,7 +1739,8 @@ static void MouseButtonCallback(GLFWwindow *window, int button, int action, int 
     // WARNING: GLFW could only return GLFW_PRESS (1) or GLFW_RELEASE (0) for now,
     // but future releases may add more actions (i.e. GLFW_REPEAT)
     CORE.Input.Mouse.currentButtonState[button] = action;
-
+    CORE.Input.Touch.currentTouchState[button] = action;
+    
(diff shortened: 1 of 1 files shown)


raysan5:
@Minmoose Oh! Good catch! Thank you very much for the review!

PR #3615, opened 2023-12-08
Updating CMake version to 5.0 so libraries are correctly versioned

davidthings:
CMakeLists.txt still shows the RayLib version to be 4.5.0.

This PR updates it to 5.0.0


raysan5:
@davidthings thanks for the review!

PR #3618, opened 2023-12-09
Update audio_stream_effects.c

lipx1508:
This may slightly improve performance and be more welcoming for new users despite being an more advanced feature. `void *` usually throws an error in most compilers and it would be better to just avoid it.

Also added `<stdbool.h>` because booleans are, sometimes, not defined by `<stddef.h>`.

### DIFF
--- a/examples/audio/audio_stream_effects.c
+++ b/examples/audio/audio_stream_effects.c
@@ -2,7 +2,7 @@
 *
 *   raylib [audio] example - Music stream processing effects
 *
-*   Example originally created with raylib 4.2, last time updated with raylib 4.2
+*   Example originally created with raylib 4.2, last time updated with raylib 5.0
 *
 *   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
 *   BSD-like license that allows static linking with closed source software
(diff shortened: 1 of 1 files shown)


examples/audio/audio_stream_effects.c
```diff
@@ -14,6 +14,7 @@
 #include "raylib.h"
 
 #include <stdlib.h>         // Required for: NULL
+#include <stdbool.h>        // Required for: bool
```
raysan5, line 17: It is already included by "raylib.h" when required, no need to include it again.

examples/audio/audio_stream_effects.c
```diff
@@ -149,13 +150,17 @@ static void AudioProcessEffectLPF(void *buffer, unsigned int frames)
     static const float cutoff = 70.0f / 44100.0f; // 70 Hz lowpass filter
     const float k = cutoff / (cutoff + 0.1591549431f); // RC filter formula
 
+    // Converts the buffer data before using it
+    float * bufferData = (float *)buffer;
```
raysan5, line 154: Please, it should follow the raylib coding conventions, it should be: `float *bufferData`.

lipx1508: Sorry! I generally use this convention and I forgot to change it before committing.
I'm getting used to contributing to open source, worth noting :)

lipx1508:
@raysan5 Fixed! (sorry for drafting, I'm getting used to GitHub lol)

raysan5:
@lipx1508 thanks for the review!

PR #3619, opened 2023-12-09
Fix typo (?) on core_window_should_close.c

lipx1508:
minor spelling mistake detected


Peter0x44:
"exit" and "program" are both not misspelled.
"exit" seems to be fairly standard language for this sort of thing. The exact prompt minecraft gives is "Do you want to exit Minecraft?"

orcmid:
I agree that there are no typos.  However, I do prefer "quit."  I think because it is used so often and is simpler.

raysan5:
@lipx1508 As @Peter0x44 noted, there is no typo or error, it's just a preference in wording. I think this change is not needed.

PR #3623, opened 2023-12-10
Update projects/CMake/CMakeLists.txt to raylib 5.0

Peter0x44:


raysan5:
@Peter0x44 thanks for the update!

PR #3624, opened 2023-12-10
Update audio_raw_stream.c

riadbettole:
AudioInputCallBack has this
    audioFrequency += 1.0f;
    audioFrequency -= 1.0f;

cancels out each others


raysan5:
@riadbettole oh! Good catch! thanks for the review!

PR #3627, opened 2023-12-11
fixed coding style in function TextToFloat

benjibst:


raysan5:
@benjibst thanks for the addition and review!

src/raylib.h
```diff
@@ -1481,6 +1481,7 @@ RLAPI const char *TextToUpper(const char *text);                      // Get upp
 RLAPI const char *TextToLower(const char *text);                      // Get lower case version of provided string
 RLAPI const char *TextToPascal(const char *text);                     // Get Pascal case notation version of provided string
 RLAPI int TextToInteger(const char *text);                            // Get integer value from text (negative values not supported)
+RLAPI float TextToFloat(const char *text);                            // Get float value from text (negative values not supported)
```
Iniquitatis, line 1484: Just to make sure: are those "negative values not supported" true as of now? Because the code seems to handle the minus sign just fine both in integer and floating point cases.

benjibst: Ah damn i mustve copied the line without modifying the comments... ill fix that too. The function works for negative numbers too

PR #3628, opened 2023-12-12
Ignore compiled `DLL` binaries.

2Bear:
I build Raylib as a share lib on Windows. Compiled `DLL` binaries should be Ignored.


raysan5:
@2Bear thanks for the review!

PR #3631, opened 2023-12-12
Fix box blur symmetry & simplify code

rmn20:
Previous version of box blur kernel was asymmetrical, because the leftmost and topmost pixels were subtracted from the average values too early, causing the image to offset by GAUSSIAN_BLUR_ITERATIONS pixels after blurring. You can see this at the rightmost and bottommost borders of the image. This commit fixes this bug.

|Original image|Old box blur|Fixed box blur|
|-|-|-|
|image|image|image|

Image is also a bit more blurry now since the blur kernel is now 1 pixel wider.
PS. Perhaps GAUSSIAN_BLUR_ITERATIONS could be a function argument rather than a predefined value?

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -1959,29 +1959,24 @@ void ImageBlurGaussian(Image *image, int blurSize) {
             float avgG = 0.0f;
             float avgB = 0.0f;
             float avgAlpha = 0.0f;
-            int convolutionSize = blurSize+1;
+            int convolutionSize = blurSize;
 
-            for (int i = 0; i < blurSize+1; i++)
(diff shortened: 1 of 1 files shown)


raysan5:
@rmn20 Nice! Great improvement! Actually I didn't notice the issue with previous implementation, but not it looks way better! Thanks!

rmn20:
@raysan5 You can see the offset issue with previous implementation much more obviously, if you apply box blur several times.
For example, here I've applied box blur with radius 1 three times:

|Original image|Old box blur|Fixed box blur|
|-|-|-|
|image|image|image|

PR #3633, opened 2023-12-13
[rcore] Flip VR screens

orosmatthew:
The left and right screens when calling `LoadVrStereoConfig` is flipped. This pull request simply flips them back to the correct configuration.


raysan5:
@orosmatthew thanks for the review!

PR #3635, opened 2023-12-14
Make centering clearer in 'Texture source and destination rectangles' example

Mapt04:
When trying to center a sprite using DrawTexturePro(), it can be done by setting the origin vector to be half the width and height of the destination rectangle. In the example, it is simply set to the original frame width and height. This works, but only because destRec is twice the size of the original. In my opinion it would be clearer for new users if origin was defined as half the size of destRec, as that works regardless of how much you scale destRec.


raysan5:
@Mapt04 Sorry but I think this is more confusing because origin refers to source rectangle, never destination rectangle. Referencing it through destination could confuse users. Imagine if destination is scaled like:

```c
Rectangle destRec = { screenWidth/2.0f, screenHeight/2.0f, frameWidth*4.0f, frameHeight };
```
Then origin would be quite confusing.

PR #3638, opened 2023-12-14
[rmodels] Add index buffer element type

denysmaistruk:
Title: Add **RL_INDEX_BUFFER_ELEMENT_TYPE** for Flexible Index Buffer Sizes

Description:

This pull request introduces a new define, `RL_INDEX_BUFFER_ELEMENT_TYPE`, to the library. The purpose of this define is to allow for more flexibility in the size of the index buffer.

Depending on whether OpenGL ES2 is used or not, `RL_INDEX_BUFFER_ELEMENT_TYPE` can be an unsigned short or an unsigned int. This is significant because it directly impacts the maximum number of elements that can be held in the index buffer.

With an unsigned short, the maximum number of elements is 65,536. However, there are scenarios where this limit is not sufficient. By allowing `RL_INDEX_BUFFER_ELEMENT_TYPE` to be an unsigned int, we can increase the maximum number of elements to 4,294,967,296.

This change provides greater flexibility and capacity, particularly beneficial in scenarios that require large index buffers.


raysan5:
@denysmaistruk Thank you very much for the review but I'm afraid I'm not merging it. It adds and ugly patch to `raylib.h` and additional code complexity. The indices limit makes the raylib games possible to be compiled for all the supported platforms, adding this review will discretize the paltform support.

Also, nowadays, afaik, indices are mostlty avoided and plain "triangle-soups" are preferred.

PR #3640, opened 2023-12-15
Add new raylib-beef binding

Starpelly:
New beef bindings, auto generated and updated to work with v5.0.


raysan5:
@Starpelly nice! thanks for your contribution!

PR #3643, opened 2023-12-16
Fix IsCursorHidden on web

lesleyrs:
Pressing escape while you called `DisableCursor()` causes it to not register that the cursor isn't hidden anymore. You also can't avoid escape presses in javascript, so this is needed to solve that. Also the emscripten pointerlock calls already hide/show cursor and it's no longer checking from rcore so it won't be needed.


raysan5:
@lesleyrs thanks for detecting this issue and the report and fix proposed, I reviewed @ubkp alternative and I prefer it. It only implies one file modification and I try to avoid platform-specific checks on `rcore.c` as much as possible.

PR #3646, opened 2023-12-17
Update GLFW to 3.3.9

M374LX:
A review of rglfw.c due to possible new GLFW source files is pending.

Building on Linux under X11, but needs tests on other platforms.


src/external/glfw/include/GLFW/glfw3.h
```diff
@@ -1,5 +1,5 @@
 /*************************************************************************
- * GLFW 3.4 - www.glfw.org
```
M374LX, line 2: This is not a downgrade, as GLFW currently uses the not yet released 3.4 version number on its master branch, which is in turn currently used by raylib.

raysan5:
@M374LX This is a really big change. Why not wait until the 3.4 official release? What are the improvements for raylib of this change?

M374LX:
Indeed, waiting for 3.4 might be a better idea.

PR #3648, opened 2023-12-17
Robust input example

benev0:
This is intended to be a frame rate independent demonstration of deletion of text. 

Resolving the issue I found with the text box input example [here](https://www.raylib.com/examples/text/loader.html?name=text_input_box) [github](https://github.com/raysan5/raylib/blob/master/examples/text/text_input_box.c)
The issue being that only one deletion call can be made per game loop.

### DIFF
--- a/examples/text/text_backspace_input.c
+++ b/examples/text/text_backspace_input.c
@@ -53,7 +53,7 @@ int main(void)
         {
 
             // ingest (key) backspace
-            // if posable delete one character from the string
+            // if possible delete one character from the string
             if (key == KEY_BACKSPACE && usedSize > 0)
             {
                 textField[usedSize - 1] = 0;
@@ -68,7 +68,7 @@ int main(void)
             }
 
             // ingest (letter) and (key)
-            // if possible wright ascii character to the string
+            // if possible write ascii character to the string
             else if ((letter > 0) && (letter >= 32) && (letter <= 125) && (usedSize < MAX_INPUT_CHARS))
             {
                 textField[usedSize] = (char) letter;
--- a/examples/text/text_backspace_input.c
+++ b/examples/text/text_backspace_input.c
@@ -2,6 +2,8 @@
 *
 *   raylib [core] example - Backspace input
 *
+*   Example demonstrates deletion of characters in a string independent of frame rate
+*
 *   Example originally created with raylib 5.0, last time updated with raylib 5.0
 *
 *   Example contributed by benevo (@benev0) and reviewed by Ramon Santamaria (@raysan5)
--- a/examples/text/text_backspace_input.c
+++ b/examples/text/text_backspace_input.c
@@ -1,0 +1,177 @@
+/*
+    WELCOME raylib EXAMPLES CONTRIBUTOR!
+
+    This is a basic template to anyone ready to contribute with some code example for the library,
+    here there are some guidelines on how to create an example to be included in raylib
+
+    1. File naming: <module>_<description> - Lower case filename, words separated by underscore,
+       no more than 3-4 words in total to describe the example. <module> referes to the primary
+       raylib module the example is more related with (code, shapes, textures, models, shaders, raudio).
+       i.e: core_input_multitouch, shapes_lines_bezier, shaders_palette_switch
+
+    2. Follow below template structure, example info should list the module, the short description
+       and the author of the example, twitter or github info could be also provided for the author.
+       Short description should also be used on the title of the window.
+
+    3. Code should be organized by sections:[Initialization]- [Update] - [Draw] - [De-Initialization]
+       Place your code between the dotted lines for every section, please don't mix update logic with drawing
+       and remember to unload all loaded resources.
+
+    4. Code should follow raylib conventions: https://github.com/raysan5/raylib/wiki/raylib-coding-conventions
+       Try to be very organized, using line-breaks appropiately.
+
+    5. Add comments to the specific parts of code the example is focus on.
+       Don't abuse with comments, try to be clear and impersonal on the comments.
+
+    6. Try to keep the example simple, under 300 code lines if possible. Try to avoid external dependencies.
+       Try to avoid defining functions outside the main(). Example should be as self-contained as possible.
+
+    7. About external resources, they should be placed in a [resources] folder and those resources
+       should be open and free for use and distribution. Avoid propietary content.
+
+    8. Try to keep the example simple but with a creative touch.
+       Simple but beautiful examples are more appealing to users!
+
+    9. In case of additional information is required, just come to raylib Discord channel: example-contributions
+
+    10. Have fun!
+
+    The following files should be updated when adding a new example, it's planned to create some
+    script to automatize this process but not available yet.
+
+     - raylib/examples/<category>/<category>_example_name.c
+     - raylib/examples/<category>/<category>_example_name.png
+     - raylib/examples/<category>/resources/*.*
+     - raylib/examples/Makefile
+     - raylib/examples/Makefile.Web
+     - raylib/examples/README.md
+     - raylib/projects/VS2022/examples/<category>_example_name.vcxproj
+     - raylib/projects/VS2022/raylib.sln
+     - raylib.com/common/examples.js
+     - raylib.com/examples/<category>/<category>_example_name.html
+     - raylib.com/examples/<category>/<category>_example_name.data
+     - raylib.com/examples/<category>/<category>_example_name.wasm
+     - raylib.com/examples/<category>/<category>_example_name.js
+*/
+
+/*******************************************************************************************
+*
+*   raylib [core] example - Backspace input
+*
+*   Example originally created with raylib 5.0, last time updated with raylib 5.0
+*
+*   Example contributed by benevo (@benev0) and reviewed by Ramon Santamaria (@raysan5)
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
+*   Copyright (c) 2023 benevo (@benev0)
+*
+********************************************************************************************/
+
+#include "raylib.h"
+
+#define MAX_INPUT_CHARS 40
+
+
+bool isKeyChar(int);
+
+//------------------------------------------------------------------------------------
+// Program main entry point
+//------------------------------------------------------------------------------------
+int main(void)
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    const int screenWidth = 800;
+    const int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [text] example - backspace input");
+
+    char textField[MAX_INPUT_CHARS + 1] = "\0";
+    int usedSize = 0;
+
+    SetTargetFPS(1);
+    //--------------------------------------------------------------------------------------
+
+    // Main game loop
+    while (!WindowShouldClose())    // Detect window close button or ESC key
+    {
+        // Update
+        //----------------------------------------------------------------------------------
+
+        int letter = GetCharPressed();
+        int key = GetKeyPressed();
+
+        if(key)
+            printf("%d, %d\n", letter, key);
+
+        while (key > 0 || letter > 0)
+        {
+
+            if (key == KEY_BACKSPACE && usedSize > 0)
+            {
+                textField[usedSize - 1] = 0;
+                usedSize--;
+                key = GetKeyPressed();
+            }
+            else if (!isKeyChar(key))
+            {
+                key = GetKeyPressed();
+            }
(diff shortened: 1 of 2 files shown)


benev0:
feedback appreciated. 
In particular some issues that I have before I mark as ready for review are:

- nothing except user text is rendered to the screen.
- comments do not fully explain the workings of the code and intent behind variables.
- against simplicity guideline there is a function present.

Do not feel the need to limit yourself to responding to these issues if you can pose a new one please do.

examples/text/text_backspace_input.c
```diff
@@ -0,0 +1,129 @@
...
+    return 0;
+}
+
+bool isKeyChar(int key)
+{
+    switch (key)
+    {
+    case KEY_SPACE:
+    case KEY_APOSTROPHE:
+    case KEY_COMMA...KEY_NINE:
```
Peter0x44, line 119: This ... feature seems non-standard, it may not work on all compilers (likely visual studio)
I don't have it available to try but it's probaby worth removing either way
```
bruh.c:119:5: warning: range expressions in switch statements are non-standard [-Wpedantic]
  119 |     case KEY_COMMA...KEY_NINE:
      |     ^~~~
```

examples/text/text_backspace_input.c
```diff
@@ -0,0 +1,129 @@
...
+        //----------------------------------------------------------------------------------
+
+        int letter = GetCharPressed();
+        int key = GetKeyPressed();
+
+        while (key > 0 || letter > 0)
+        {
+
+            // ingest (key) backspace
+            // if posable delete one character from the string
```
Peter0x44, line 56:
```suggestion
            // if possible delete one character from the string
```

examples/text/text_backspace_input.c
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            // ingest (key) non character
+            else if (!isKeyChar(key))
+            {
+                key = GetKeyPressed();
+            }
+
+            // ingest (letter) and (key)
+            // if possible wright ascii character to the string
```
Peter0x44, line 71:
```suggestion
            // if possible write ascii character to the string
```

examples/text/text_backspace_input.c
orcmid: @Peter0x44 @benev0 I don't understand how `IsKeyChar` is an improvement on using `ctypes.h` and `isprint(c)` or some other suitable combination of standard character classification functions.

Peter0x44: @orcmid Your advice is terrible. Never give it again, you are inviting bugs.
ctypes.h itself is a horrific footgun and any code you see including it is probably defective.
I ban it outright.

1) it's affected by locale (hopefully self-evident why this is bad)
2) it's designed for use with fgetc, with a domain of unsigned char + EOF. any negative inputs aside from EOF are UB,
So:
```c
char* x = ...;
if (isprint(x[0])) ...
```
is straight up incorrect.

I could rant even longer about it but I hope you get the message. Ban any use of ctype.h from your code immediately.

benev0: I have updated the function to use bitwise operations.

@orcmid 
As for `isprint(int)` I believe the functionality is different. `isprint(int)` takes in a character code while `isKeyChar(int)` takes in a key code, so although keycodes mostly directly correspond, `isPrint(int)` does not appear to take into account numpad input. Do note that I have recently added numpad keys to fix a crash, so I believe you where correct that these where interchangeable functions.

benev0:
> @benev0 Really liked your idea of showing how to get multiple key inputs in the same frame. I think this would be a really good addition. Great job!
> 
> But I'd like to suggest simplifing it. IMHO, examples are there to show the fundamentals in practice, the simpler the better.
> 
> I believe just this would be enough:
> 
> ```
> int letter = GetCharPressed();
> int key = GetKeyPressed();
> 
> while (letter || key)
> {
>     if (key == KEY_BACKSPACE)
>     {
>         if (usedSize > 0)
>         {
>             textField[usedSize -1] = 0;
>             usedSize--;
>         }
>     }
> 
>     if (letter)
>     {
>         if (usedSize < MAX_INPUT_CHARS)
>         {
>             textField[usedSize] = (char)letter;
>             usedSize++;
>         }
>     }
> 
>     letter = GetCharPressed();
>     key = GetKeyPressed();
> }
> ```
> 
> Also would suggest renaming it to `text_framerate_independent.c`, looks more accurate.
> 
> **Edits:** 1. typo; 2. simplify code.

I think this is an exelent 99% case; however, I think that there may be as issue with deleting multiple times with a successive add (in one frame). I will followup with further reasoning or a retraction, but regardless this works with nearly all typing that the threw at it and I assume that any case I generate will highly contrived. Good work :)

I like the file name suggestion I will change it in the next commit.

benev0:
The issue I have found with the code provided by @ubkp is that is does not guarantee that the queues are synchronized, in the case where each key press occurs on a different frame the queue is cleared and so the synchronization is guaranteed. In a high frame rate environment I believe the code is highly safe given that with two key presses per frame the code is safe.

note that backspaces are processed first.
> ```
> int letter = GetCharPressed();
> int key = GetKeyPressed();
> 
> while (letter || key)
> {
>     if (key == KEY_BACKSPACE)
>     {
>         if (usedSize > 0)
>         {
>             textField[usedSize -1] = 0;
>             usedSize--;
>         }
>     }
> 
>     if (letter)
>     {
>         if (usedSize < MAX_INPUT_CHARS)
>         {
>             textField[usedSize] = (char)letter;
>             usedSize++;
>         }
>     }
> 
>     letter = GetCharPressed();
>     key = GetKeyPressed();
> }
>```
By exhaustion it can shown that all combinations of delete and character key presses result in the expected output

$Q_c$ Queue of letters 
$Q_k$ Queue of key presses

|  | first | second | $\frac{expected}{result}$ |
|--|--|--|--|
| $Q_{c0}$ | a | a | "aa" |
| $Q_{k0}$ | a | a | "aa" |
| | | |
| $Q_{c1}$ | a | | "" |
| $Q_{k1}$ | a | $\leftarrow$ | ""|
||||
| $Q_{c2}$ | a | |\" $\leftarrow$ a\" |
| $Q_{k2}$ | $\leftarrow$  | a |   \" $\leftarrow$ a\" |
||||
| $Q_{c3}$ | |  | \" $\leftarrow \leftarrow$ \" 
| $Q_{k3}$ | $\leftarrow$  | $\leftarrow$ | \" $\leftarrow \leftarrow$ \"  |

To address the possible use of other 0 character keys such as shift in a sequence of two key presses:

- If the character key is replaced then no character adds occur only deletes are a factor. 
- If the delete key is replaced no deletes occur and only the character key is a factor. 

with two key presses it is impossible that the queues to de-sync and cause and issue. 

As for three key presses here is a sequence that de-syncs code.

|  | first | second | third | $\frac{expected}{result}$ |
|--|--|--|--|--|
| $Q_{c}$ | a |  |  |  " $\leftarrow \leftarrow$ a" |
| $Q_{k}$ | $\leftarrow$ | $\leftarrow$ | a | " $\leftarrow$ " |

As $Q_c$ does not store backspace characters the character from the third key press is processed in the same iteration as the first backspace. The second iteration results in the second deletion character being processed deleting the first character and third key press information. The third iteration does no processing and terminates the loop.

A de-sync will also occur when the first $Q_k$ item is another non character such as shift. 

Is this sufficient issue to warrant the complexity required to mitigate it?

raysan5:
@ubkp Agree 100% with your comments.

orcmid:
@benev0 @ubkp @raysan5 
I concur about the difficulties.  I also think that it is too easy to over-simplify the situation, especially with the importance of keystate and the fact that there are keypresses that should not count as characters or lead to characters in that queue.

In the example code, there might need to be some caveats about the assumptions that allow the example to be simple.  

Generally, there's also the matter of the conditions under which certain text input functions are used and whether the text is being shown to the user who may have some (limited) ability to edit it before signaling completion (`KEY_ENTER` or some other).

A simple example is arrow keys (including from the keypad depending on NumLock).  They might signify movement (in the manner of WASD) or they might signify movement of a **text** cursor and change an insertion point.

PS: Earlier, I confused "key" with "char" and I didn't catch myself until I explored the `KeyboardKey` **`enum`** in `raylib.h`.  With regard to my remark about `ctypes.h`, the implementations of those is-functions are very efficient, using nothing so wasteful (and difficult to check) as elaborate **`switch`** statements.  Instead bit vectors are used and it takes only one access into the array indexed by the char value to answer questions about a single character by using a bit mask.  This also allows certain accommodations of language and locale. In particular, one can change locale just by switching to a different version of the table.  That concept could be applied with respect to keys as well.  And, by the way, not all `ctypes.h` is-functions are locale dependent.  Function `isprint(c)` is not locale-dependent in ISO Standard C, for example.  (I confirmed that looking at the C11 Standard and then checking Windows SDK implementation, which uses a cute trick that works whether a char is signed or unsigned.)  I suspect that the keys situation is more complicated, even more than the situation with controllers.  There's a nice reference example of  using bit vectors in Plauger's *The Standard C Library* book.

Peter0x44:
@orcmid The issue of passing negative values (aside from EOF) as UB is true and in real practice, there are implementations that crash on them. As I mentioned, these functions are not for "characters" in strings. Passing unsigned chars makes most of them work "by accident", but passing signed chars is undefined for arbitrary input, which is really bad. There's no good reason to have the functions fall over on that.

And yeah, it does seem to be true that not all of them are affected by locale. But, enough of them are, that it's a footgun I'd rather leave out from my code.

ctype.h is an aggressively misleading API that gets misused in virtually every instance.
see the `Character classification and mapping` of this post for more details in a more neutral tone if you don't agree with me.
https://nullprogram.com/blog/2023/02/11/

As for optimization: compilers are more than capable of converting the checks to bit tests on their own.
https://gcc.godbolt.org/z/dxrzWjeex
All of the ctype.h functions compile to function calls, as far as I can tell, for gcc anyway.
I doubt there is any perf advantage to ctype.h, if anything, there would be a disadvantage because the compiler cannot inline them...

If you're interested, some of the optimizations are also described here
https://queue.acm.org/detail.cfm?id=3372264
See `Chained conditionals`

benev0:
@Peter0x44 as this argument with @orcmid is no longer referring directly to any code relevant to this PR, I ask that you stop discussion of it here.

benev0:
Thanks to the modifications suggested by @ubkp I believe that core functionality is ready to be reviewed, but before finalizing the submission I would like to consider the following:

- [ ] Dress up with draw calls
- [ ] Proper crediting for contributors 

I also have the concern that this example has become similar to the `text_input_box.c` example, so it may be more appropriate to update example instead. @raysan5, your input would be greatly valued on this matter.

orcmid:
>  bit vectors are used and it takes only one access into the array indexed by the char value to answer questions about a single character using choice of a bit mask. ... That concept could be applied with respect to keys as well. 

I would like that suggestion to not  be lost in the fog of war here.  I will say no more.

examples/text/text_framerate_independent.c
```diff
@@ -0,0 +1,91 @@
...
+//------------------------------------------------------------------------------------
+int main(void)
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    const int screenWidth = 800;
+    const int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [text] example - backspace input");
+    SetTargetFPS(1);
```
raysan5, line 33: This is not a framerate independent implementation if the example fixes the framerate to 1. Actually, that's a really bad and not-real practize, I'm afraid it could confuse many users.

examples/text/text_framerate_independent.c
```diff
@@ -0,0 +1,91 @@
...
+            {
+                textField[textPos -1] = 0;
+                textPos--;
+            }
+
+            // if there is a letter and space in the string
+            // then read letter in from character queue
+            if (letter && textPos < MAX_INPUT_CHARS)
+            {
+                textField[textPos] = (char) letter;
```
raysan5, line 63: This piece of code is not correct, it casts a codepoint to a char, Unicode is not supported, only english text. `letter` should be converted to UTF-8 to be a proper example.

raysan5:
@benev0 I'm afraid I'm not merging this example for the comments I reviewed.

examples/text/text_framerate_independent.c
orcmid: Wait, no-no, UTF8 are multi-byte characters.  Is `textField[]` a UTF-8 (`unsigned char`) string or a codepoint (`int`, UTF-32) string?  If `textField[]` is UTF-8, backspacing and determining whether or not a codepoint's UTF-8 fits does not overflow `textField[]` requires great care.  This is a situation where codepoints are more reliable.

Maybe I am misunderstanding the context here?

PR #3649, opened 2023-12-17
Remove dependencies from PBR example

TheManTheMythTheGameDev:
Remove dependencies from the PBR example. Switch to raylib structures and remove rpbr.h.
I had to move some things from rpbr.h into shaders_basic_pbr.c; for instance, I needed a light struct, and the functions in rpbr.h worked well.


raysan5:
@TheManTheMythTheGameDev Great work! Thank you very much for the review!

PR #3652, opened 2023-12-18
[rmodels] Set cgltf callbacks to use `LoadFileData()` and `UnloadFileData()`

kolunmi:
Hello! This ensures cgltf uses the callbacks configured by the user to load and unload file data, enhancing the consistency and predictability of raylib.


raysan5:
@kolunmi Thank you very much for the improvement!

PR #3655, opened 2023-12-19
fix for deferred rendering example.

JettMonstersGoBoom:
previous example didn't correctly render with depth outside of the deferred pass. the DrawSphereWires didn't work.


raysan5:
@JettMonstersGoBoom thanks for the review, I'm merging it to fix the example but I try to avoid exposing OpenGL-specific values from rlgl directly, those functions require some more work in the future...

PR #3657, opened 2023-12-20
2 important fixes for borderless fullscreen on rcore_desktop

lesleyrs:
Fix 1: In windowfocuscallback, toggle topmost flag if borderless fullscreen is active so that you can alt tab while retaining borderless fullscreen. Currently it will let you lose focus but it won't go to the background.

Fix 2: When in an opengl window, setting the window size to monitor size causes some flickering which can be really annoying. Other games get around this by adding or removing 1 pixel to either size or position. This is not opinionated you can see games like supertuxkart doing it for yourself. I decided to only add a parameter for extra width because imo it's the least noticable but you can still set it to 0 to have it work the same as now.

You can test 1. by going fullscreen and alt tabbing, and 2. by setting extrawidth to -1 or +1.

Also changed comments for some documentation and updating the functions to remove the desktop only as I think that was forgotten to be changed.

1 problem with fix 2 is that SetWindowState only sets extraWidth to 0 now, ideally you'd have some value you can set and read because you don't really change it anyway. But not sure how you want to add that and I just want this merged to have a nice cross platform way to get fullscreen (I used custom fullscreen before this implementing the above).

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -197,7 +197,7 @@ void ToggleFullscreen(void)
 }
 
 // Toggle borderless windowed mode
-void ToggleBorderlessWindowed(void)
+void ToggleBorderlessWindowed(int extraWidth) // increase or decrease window size to stop flickering
 {
     // Leave fullscreen before attempting to set borderless windowed mode and get screen position from it
     bool wasOnFullscreen = false;
@@ -240,7 +240,7 @@ void ToggleBorderlessWindowed(void)
 
                 // Set screen position and size
                 glfwSetWindowPos(platform.handle, monitorPosX, monitorPosY);
-                glfwSetWindowSize(platform.handle, monitorWidth, monitorHeight);
+                glfwSetWindowSize(platform.handle, monitorWidth + extraWidth, monitorHeight);
 
                 // Refocus window
                 glfwFocusWindow(platform.handle);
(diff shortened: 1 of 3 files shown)


lesleyrs:
This is the first time I tagged you for the side note because I saw you added fullscreen for web which i'm thankful for. So sorry about that I don't want to give you extra work and it's good as it is.

Are you implying my borderless fullscreen changes breaks anything? Because borderless fullscreen on Windows (desktop) in it's current state is not worth using for me and these changes are a must. If there is a problem I'd like to know.

raysan5:
@lesleyrs I'm afraid I'm not merging this PR, check the review comment for details.

@ubkp I understand your concerns and I'm sorry you decide to stop contributing to the project, your contributions has been indeed unvaluable. Thanks again for all the work you put into raylib and every great PR. It's been also a pleasure having you as a contibutor. Thanks and all the best with your future projects, I'm sure they will be amazing!

lesleyrs:
Hmm I feel like neither of you tried this PR before shutting it down, I realize that fullscreen functions were reworked to be compatible but this change doesn't really interact with the others. Even just the topmost change just lets you alt tab and does nothing else which has no downsides.

I don't like the way this PR got dismissed so quickly just because the other guy is upset, this along with the other PR regarding pointerlock were both good changes.

But yes, let's keep borderless fullscreen in it's current state and I'll just go away.

raysan5:
@lesleyrs I explained the reasons for the PR to be rejected, it's breaking raylib API and adding uneeded complexity to a simple function:

 > This change is breaking the API and adding an unneeded extra level of complexity with an unintuitive additional parameter to cover a very-specific use case that will only affect a very small percentadge of user. raylib priority is simplicity.

lesleyrs:
Alright, that message didn't appear for me.

PR #3661, opened 2023-12-21
Add new COBOL binding

glowiak:
I made a COBOL "binding" of some sort.

RL can be used in COBOL without it, but it is barely usable, and my "binding" greatly improves the usability.

Typed in public domain as I am not a lawyer and all my code is not licensed.


raysan5:
@glowiak Great addition! Thanks!

orcmid:
You can't put something in the public domain.  You hold copyright as an author, it is an automatic thing (unless you are a worker for hire and your employer has the rights).  That copyright has to expire for something to enter the public domain.  You can make a "quit claim" however. I think there are examples of that somewhere.  (My knowledge of this is specific to the USA, but the international conventions are consistent.)

PR #3666, opened 2023-12-23
VSCode sample project Mac fixup

tromero:
Change vscode config and makefile to add a missing IOKit framework reference and use whichever MacOSX sdk symlink is available


raysan5:
@tromero thanks for the review!

PR #3672, opened 2023-12-25
Update BINDINGS.md

iacore:
Update zig libs version.

Some version numbers are bold. Some are not. What's the rule here?


raysan5:
@iacore thanks for the review. The bold versions are supposed to be the latest but the list is not updated.

PR #3673, opened 2023-12-25
Add const modifier to drawing API

iacore:
there may be others than can be changed so


raysan5:
@iacore Sorry, no plans to implement this change for now.

iacore:
`const` is used in the header file.

for example, here

```
RLAPI Image GenImageFontAtlas(const GlyphInfo *glyphs, Rectangle **glyphRecs, int glyphCount, int fontSize, int padding, int packMethod); // Generate image font atlas using chars info
```

iacore:
@raysan5 please reconsider.

PR #3677, opened 2023-12-26
[rshapes] Expose shapes drawing texture and rectangle

JeffM2501:
In order to allow external code to do the same thing as the rshapes functions, we need to export the shape texture and texture rect so that external shape functions can use them. 

This allows people to copy code from rshapes and build extended or new shape functions in user code and still have all the same features and optimizations that internal shape functions have.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1220,6 +1220,8 @@ RLAPI void UpdateCameraPro(Camera *camera, Vector3 movement, Vector3 rotation, f
 // NOTE: It can be useful when using basic shapes and one single font,
 // defining a font char white rectangle would allow drawing everything in a single draw call
 RLAPI void SetShapesTexture(Texture2D texture, Rectangle source);       // Set texture and rectangle to be used on shapes drawing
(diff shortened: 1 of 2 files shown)


src/raylib.h
```diff
@@ -1220,6 +1220,8 @@ RLAPI void UpdateCameraPro(Camera *camera, Vector3 movement, Vector3 rotation, f
 // NOTE: It can be useful when using basic shapes and one single font,
 // defining a font char white rectangle would allow drawing everything in a single draw call
 RLAPI void SetShapesTexture(Texture2D texture, Rectangle source);       // Set texture and rectangle to be used on shapes drawing
+RLAPI Texture2D GetShapesTexture();                                     // Get texture that is used for shapes drawing
```
raysan5, line 1223: C language requires specifying `void` for input parameter, no parameter is understood as any number of parameters.

JeffM2501: Fixed

orcmid (changes requested):
Oops.  In `rshapes.c` line 122 is replaced twice.

 ... and I was just looking where to give this a thumbs-up :).

raysan5:

PR #3678, opened 2023-12-27
TextReplace const correctness

maverikou:
Fix for a minor annoyance.


raysan5:
@maverikou thanks for the review!

PR #3681, opened 2023-12-27
[rgraphics] Add LoadImageAnimFromMemory.

IoIxD:
# Context

I would like my Raylib game to be a singular executable, and not have to load any files at compile time.

#Changes 

- Adds a function that's basically LoadImageAnim but it assumes fileData is provided at compile time.


raysan5:
@IoIxD Thanks for the addition.

PR #3685, opened 2023-12-28
Add new chars to font ex

stevemolloy:
This PR proposes the addition of a new function `AddNewCharsToFontEx`.

# Problem

I found myself in need of adding a single codepoint to the default character set that is loaded when passing `NULL` to `LoadFontEx`.  (See [here](https://github.com/stevemolloy/fonts_in_raylib))

Lacking a suitable function I would have to load the font by passing the string I intended to supply.  This has the weakness that future unknown input (perhaps supplied by the user) would contain characters that are not part of this set.

What I really wanted was a way to say that I want the default character set plus the single extra codepoint.

# Idea

Provide a function that would allow reloading of a `Font` from file with the required extra codepoints.

# Potential weaknesses

One major potential weakness is the linear search I perform to search for requested codepoints that are already in the character set.  This scales with the length of the already-loaded set *and* the length of the requested set, and so could be somewhat long if these happen to be large.


raysan5:
@stevemolloy I'm afraid I'm not adding this function. I think it can be implemented at user side for specific use cases equiring it, `LoadFontEx()` should be enough to load a custom charset when required.

PR #3690, opened 2023-12-30
Example gizmo

alexeykarnachev:
Self-contained Gizmo gadget implementation.
It allows to rotate and translate objects in 3d space.

More formally, it is not coupled with any runtime object at all though. The gadget just yields some transformation matrix and it is up to user apply this matrix to any object on a scene.
gizmo


JeffM2501:
I'm not sure this should be a raylib example as it's more than a single file. this seems more like a reusable component. It would probably be better as part of raylib-extras https://github.com/raylib-extras

raysan5:
@alexeykarnachev Thanks for the example, it really looks amazing! But I agree with @JeffM2501 that it could be a better fit for [`raylib-extras`](https://github.com/raylib-extras/extras-c), considering the addition of an extra portable module.

alexeykarnachev:
Got it, thanks :) will take a look

PR #3691, opened 2023-12-31
rtext.c: Let user choose if they want an automatic FPS color or not

m-Torio:
I made an overloaded function of DrawFPS(int posX, int posY) which takes an extra parameter of type Color. In the overloaded DrawFPS, the user can select a constant color, they want their FPS counter to be, instead of the default variable FPS color.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1089,6 +1089,11 @@ void DrawFPS(int posX, int posY)
 
     DrawText(TextFormat("%2i FPS", fps), posX, posY, 20, color);
 }
+void DrawFPS(int posX, int posY, Color color)
+{                         
+    int fps = GetFPS();
+    DrawText(TextFormat("%2i FPS", fps), posX, posY, 20, color); // Draw FPS based on single color
+}
 
 // Draw text (using default font)
 // NOTE: fontSize work like in any drawing program but if fontSize is lower than font-base-size, then font-base-size is used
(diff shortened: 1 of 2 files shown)


src/rtext.c
```diff
@@ -1089,6 +1089,11 @@ void DrawFPS(int posX, int posY)
 
     DrawText(TextFormat("%2i FPS", fps), posX, posY, 20, color);
 }
+void DrawFPS(int posX, int posY, Color color)
```
m-Torio, line 1092: // Overloaded DrawFPS where you can choose color

RadsammyT:
The C99 standard does not support function overloads. 
When ignoring that oversight, with all due respect, I fail to see how your implementation could be useful to include in raylib, when comparing it to the original `DrawFPS(int, int)` (convenience factor?) or the possibility of the proposed feature's implementation being used in a standalone raylib project, rather than raylib itself. I'm certain @raysan5 would feel the same way.

m-Torio:
> The C99 standard does not support function overloads. When ignoring that oversight, with all due respect, I fail to see how your implementation could be useful to include in raylib, when comparing it to the original `DrawFPS(int, int)` (convenience factor?) or the possibility of the proposed feature's implementation being used in a standalone raylib project, rather than raylib itself. I'm certain @raysan5 would feel the same way.

The reason the overloading I thought worked because I forgot to change build from c++17 to c99, will edit pull request tomorrow. The reason for the request in the first place was because I was making a game and I had a limited color pallete to black and white, so I wanted to drawfps in white, I know I could've used drawtext but I thought this approach would be more clean and readable. This was my first pull request on github so I wanted to contribute to raylib in a small way. If you think this pull request is unnecessary, I might take it down

raysan5:
@m-Torio Thanks for the proposal but I'm afraid users requiring more control on the text format can just use `DrawTextEx()`, there is no need for additional functions.

PR #3692, opened 2023-12-31
Update raylib_api.* by CI

seiren-games:
Ensure `raylib_api.*` is always updated.

`parser/output/raylib_api.*` may be outdated and generating bindings from them will result in an error.
For example, `raylib_api.*` in tag 5.0 remains unupdated.
The purpose of this PR is to prevent such things.


raysan5:
@seiren-games Really nice addition! Thank you very much!

PR #3700, opened 2024-01-04
Introduce c-style context pointer on AudioCallback

LeandroSQ:
## Intro
Hey folks, 
I hope to be able to contribute to this amazing project, so let's get started with the PR description...

## Summary:
The existing [`AudioCallback`](https://github.com/raysan5/raylib/blob/e46b6147fc30e69dd300bbda90ff0a62bf8a4df0/src/raylib.h#L1588C16-L1588C29) definition, `void *buffer, unsigned int frames`, presents limitations by preventing the use of the same callback for multiple streams. In dynamic scenarios, accommodating this requirement becomes unfeasible.

This change introduces the capability to pass a void pointer as a parameter to the callback function, a common c pattern, enabling the function to have user-provided context for achieving desired results. 

## Changes Made:
Modified AudioCallback to accept a void pointer parameter, enhancing its flexibility both for AudioStreams and AudioProcessors in general.

## Example Usage:
Consider the following example to illustrate the enhanced flexibility:

```cpp
struct Info {
    AudioStream stream;
    float frequency;
    float time;
};

void callback(void *buffer, unsigned int frames, void *context) {
    Info *info = (Info *)context;
    TraceLog(LOG_DEBUG, "main: %.2f", info->frequency);
    short *sampleBuffer = (short *)buffer;

    float incr = info->frequency / 44100.0f;

    for (int i = 0; i < frames; i++) {
        sampleBuffer[i] = (short)(32000.0f * sinf(2.0f * PI * info->time));
        info->time += incr;
        if (info->time > 1.0f) info->time -= 1.0f;
    }
}

int main(void) {
    InitWindow(WIDTH, HEIGHT, "Raylib example!");
    InitAudioDevice();
    SetAudioStreamBufferSizeDefault(4096);
    SetMasterVolume(0.25f);

    Info info1 = {
        .stream = LoadAudioStream(44100, 16, 1),
        .frequency = 440.0f,
        .time = 0.0f
    };
    Info info2 = {
        .stream = LoadAudioStream(44100, 16, 1),
        .frequency = 880.0f,
        .time = 0.0f
    };

    SetAudioStreamCallback(info1.stream, callback, &info1);
    SetAudioStreamCallback(info2.stream, callback, &info2);

    PlayAudioStream(info1.stream);
    PlayAudioStream(info2.stream);

    SetTargetFPS(60);

        BeginDrawing();

        ClearBackground(BLACK);

        info1.frequency += GetFrameTime() * 40.0f;
        info2.frequency += GetFrameTime() * 40.0f;

        DrawText(TextFormat("%i FPS", GetFPS()), 10, 10, 20, WHITE);
        DrawText(TextFormat("Frequency 1: %.2f", info1.frequency), 10, 30, 20, WHITE);
        DrawText(TextFormat("Frequency 2: %.2f", info2.frequency), 10, 50, 20, WHITE);

        EndDrawing();
    }

    UnloadAudioStream(info1.stream);
    UnloadAudioStream(info2.stream);

    CloseAudioDevice();
    CloseWindow();

    return 0;
}
```

This modification ensures a more versatile and user-friendly AudioCallback interface, allowing developers to seamlessly manage multiple streams within dynamic scenarios.

### DIFF
--- a/examples/audio/audio_mixed_processor.c
+++ b/examples/audio/audio_mixed_processor.c
@@ -21,7 +21,7 @@ static float averageVolume[400] = { 0.0f };   // Average volume history
 //------------------------------------------------------------------------------------
 // Audio processing function
 //------------------------------------------------------------------------------------
-void ProcessAudio(void *buffer, unsigned int frames)
+void ProcessAudio(void *buffer, unsigned int frames, void *context)
 {
     float *samples = (float *)buffer;   // Samples internally stored as <float>s
     float average = 0.0f;               // Temporary average volume
@@ -57,7 +57,7 @@ int main(void)
 
     InitAudioDevice();              // Initialize audio device
 
-    AttachAudioMixedProcessor(ProcessAudio);
+    AttachAudioMixedProcessor(ProcessAudio, 0);
 
     Music music = LoadMusicStream("resources/country.mp3");
     Sound sound = LoadSound("resources/coin.wav");
--- a/examples/audio/audio_raw_stream.c
+++ b/examples/audio/audio_raw_stream.c
@@ -35,7 +35,7 @@ float oldFrequency = 1.0f;
 float sineIdx = 0.0f;
 
 // Audio input processing callback
-void AudioInputCallback(void *buffer, unsigned int frames)
+void AudioInputCallback(void *buffer, unsigned int frames, void *context)
 {
     audioFrequency = frequency + (audioFrequency - frequency)*0.95f;
 
@@ -69,7 +69,7 @@ int main(void)
     // Init raw audio stream (sample rate: 44100, sample size: 16bit-short, channels: 1-mono)
     AudioStream stream = LoadAudioStream(44100, 16, 1);
 
-    SetAudioStreamCallback(stream, AudioInputCallback);
+    SetAudioStreamCallback(stream, AudioInputCallback, 0);
 
     // Buffer for the single cycle waveform we are synthesizing
     short *data = (short *)malloc(sizeof(short)*MAX_SAMPLES);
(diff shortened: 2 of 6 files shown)


LeandroSQ:
It is a breaking change, since it modifies public API.
Not sure if the `AudioCallback` function pointer type can be backwards compatible, but the methods can.
By introducing new 'dummy' overloads with np such as:
```cpp
inline void SetAudioStreamCallback(AudioStream stream, AudioCallback callback) 
{
    SetAudioStreamCallbackWithContext(stream, callback, 0);
}

void SetAudioStreamCallbackWithContext(AudioStream stream, AudioCallback callback, void* context) { ... }
```

raysan5:
@LeandroSQ Thanks for the improvement but I'm afraid this is a breaking change that adds compelxity and that probably most users do not need. Users with this specific requirement can customize the library for their specific needs. I prefer to keep current functions are they are.

orcmid:
@LeandroSQ @raysan5  I've not worked that close to the metal in raylib.  Now I'm startled to learn that not all callback hooks of this sort carry and pass  back a context pointer.   That's not exactly recent practice, although perhaps more-recent than classic BGI.

I agree about the problem of breaking changes on 5.1-dev.  I'm unclear how users of raylib can work around this without customizing raylib itself.  I am less inclined to encourage forks of raylib that have to be maintained by users separately when this seems to be a pretty foundational technique.  

I think you could do this with a `SetAudioStreamCallbackEx(stream, callback, context)`  with `SetAudioStreamCallback(stream, callback)` retaining some flag so it is known which kind of call to make when firing the callback.  Having only a dummy value (e.g., `NULL` or `ERROR`) in place of unspecified context has bad edge cases.

In this regard, I am surprised that the set-callback functions don't return `int` success/fail values.  `SetAudioSttreamCallbackEx` should do so.  I presume the implementations protect themselves from bad inputs either way, although it could result in the callback not being registered.

LeandroSQ:
@raysan5 @orcmid I have pushed a commit for introducing compatibility, 
I think it has not updated the PR files since it was closed, so not really sure on how to proceed...
But the changes are backwards compatible now, using `Ex` suffix for methods that pass a context pointer.

PR #3701, opened 2024-01-04
High DPI bug fixes

karl-zylinski:
Simplified `GetWindowScaleDPI()` so it does not fetch the wrong DPI scale some times (Windows or GLFW looks at center of window while the old raylib code looked on upper left corner of window, now it just uses the `glfwGetWindowContentScale` function to fetch the window's current scaling). The old behavior caused problems when having a system with multiple monitors where one monitor had different scaling than the other. When you moved the window back and forth between the monitors, then `WindowSizeCallback` happened since the window size changes due to the scaling. But since the upper left corner was still on the other monitor (when pulling the window to a new monitor that is to the right of the old one), then `GetWindowScaleDPI()` reported the wrong scale and used that to update the screen.height and screen.width, giving them incorrect values.

Also introduced a callback to update the `CORE.Window.screenScaling` when the content scaling updates, previously it was never updated, so scaling did not work correctly on systems with multiple monitors that have different DPI scaling.

Note: I have only tested this on Windows. I did the same changes to the web platform, but I have not tested those.

I wrote about this on the Discord server, for reference here's my initial message: https://discord.com/channels/426912293134270465/426912293956222978/1192401955025334313 and my final message: https://discord.com/channels/426912293134270465/426912293956222978/1192503638267003011


raysan5:
@karl-zylinski Thank you very much for the improvement! Just need a bit more time for the review...

karl-zylinski:
@raysan5 no worries!

raysan5:
@karl-zylinski Thank you very much for the improvement!

Issue #3695, opened 2024-01-03
[rcore] Support nested Scissor calls


dbechrd:
Would be great if Raylib's `BeginScissorMode()`/`EndScissorMode()` supported nested calls. This is very common when trying to make UIs. It would also be nice if it took a Rectangle as a parameter, but that's not nearly as important.

I removed the DPI code from my version because I don't have access to `CORE`, but feel free to steal this code and put it in Raylib and re-add the DPI aware stuff:

```c
// NOTE: You'll probably want to use some Raylib/C data structure here.
std::stack<Rectangle> scissorStack{};

void PushScissorRect(const Rectangle &rect)
{
    rlDrawRenderBatchActive();

    if (scissorStack.empty()) {
        rlEnableScissorTest();
    }

    rlScissor(rect.x, GetRenderHeight() - (rect.y + rect.height), rect.width, rect.height);
    scissorStack.push(rect);
}

void PopScissorRect(void)
{
    rlDrawRenderBatchActive();

    scissorStack.pop();

    // Restore previous scissor rect
    if (!scissorStack.empty()) {
        Rectangle &rect = scissorStack.top();
        rlScissor(rect.x, GetRenderHeight() - (rect.y + rect.height), rect.width, rect.height);
    } else {
        rlDisableScissorTest();
    }
}
```

raysan5:
@dbechrd I'm afraid this is not trivial in C and it requires some additional complexity to support this feature. Out of scope for now.

dbechrd:
@raysan5 Could you be more specific? Both for my own implementation, but also in case someone wants to contribute a PR trying to solve this in the future. That would be helpful, thanks!

raysan5:
```c
// NOTE: You'll probably want to use some Raylib/C data structure here.
std::stack<Rectangle> scissorStack{};
```

There is no raylib structure for this, just an array with all the additional involved logic or an additional custom structure. It implies additional level of complexity overall, to cover a specific use case that users requiring it can implement by themselfs for their specific needs.

dbechrd:
@raysan5 I submitted a PR for this issue. Feel free to leave feedback there if there are any concerns. Thanks!

raysan5:
@dbechrd Thanks for the effort but as I said it adds a level of complexity to code that I prefer to avoid. I can keep the PR open as a reference for users really requiring it but I'm not merging it.

dbechrd:
[...]

PR #3702, opened 2024-01-04
[rcore] Add scissor stack and allow user to query current scissor region

dbechrd:
Solves #3695. API is fully backward compatible. DPI aware and should work with FBOs.


raysan5:
Closing this feature.

PR #3709, opened 2024-01-05
Added viewport independent raycast

luis605:
`GetMouseRay` did not allow the user to cast a ray from the camera to the world view while in a `viewport` or `renderTexture`. To improve raylib, I added `GetMouseRayEx`, which has new arguments (width and height), making it easier to cast a ray to a specific viewport.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1050,7 +1050,7 @@ RLAPI void UnloadShader(Shader shader);                                    // Un
 
 // Screen-space-related functions
 RLAPI Ray GetMouseRay(Vector2 mousePosition, Camera camera);      // Get a ray trace from mouse position
-RLAPI Ray GetMouseRayEx(Vector2 mousePosition, Camera camera, float width, float height);      // Get a ray trace from mouse position
+RLAPI Ray GetViewRay(Vector2 mousePosition, Camera camera, float width, float height);      // Get a ray trace from mouse position in a viewport
 RLAPI Matrix GetCameraMatrix(Camera camera);                      // Get camera transform matrix (view matrix)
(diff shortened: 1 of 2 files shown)


Elkantor:
> `GetMouseRay` did not allow the user to cast a ray from the camera to the world view while in a `viewport` or `renderTexture`. To improve raylib, I added `GetMouseRayEx`, which has new arguments (width and height), making it easier to cast a ray to a specific viewport.

Hey @luis605, great work, but you mentionned "GetMouseRayEx" in the PR, and it's named GetViewRay in the code. As you're using the parameter named mousePosition, I think GetMouseRayEx is better ;)

luis605:
> > `GetMouseRay` did not allow the user to cast a ray from the camera to the world view while in a `viewport` or `renderTexture`. To improve raylib, I added `GetMouseRayEx`, which has new arguments (width and height), making it easier to cast a ray to a specific viewport.
> 
> Hey @luis605, nice addition, but you mentionned "GetMouseRayEx" in the PR description, and it's named GetViewRay in the code. As you're using the parameter named mousePosition, I think GetMouseRayEx is better ;) 

Hi @Elkantor . I was talking to @raysan5 and @JeffM2501 and we thought GetViewRay was better named than Ex.

raysan5:
@luis605 @Elkantor Merging but function name still under consideration... `GetViewProjectionRay()` was also proposed as a more correct description of the functionality... Still `GetMouseRayEx()` seems more aligned with current API `GetMouseRay()`... what is the naming that could be more intuitive for new users?

I checked Unity API (that is usually VERY good at naming) and they have: `Camera.ViewportPointToRay()`.
It seems `Stride` engine uses a more splicit naming (not sure if same behaviour): `ScreenPositionToWorldPositionRaycast()`

Do you know what naming is used in other similar libraries/frameworks/engines?

PR #3710, opened 2024-01-06
Fix a bug in examples/shapes/shapes_following_eyes

casavaca:
Screenshot_20240105_234738
I found this bug and a trivial fix in the [shapes_following_eyes](https://www.raylib.com/examples/shapes/loader.html?name=shapes_following_eyes) example.


raysan5:
@casavaca Nice! Thanks for the review!

PR #3712, opened 2024-01-06
Simplified for loop for some image manipulation functions

Minnowo:
The image manipulation functions `ImageColorInvert`, `ImageColorBrightness`, `ImageColorReplace`, and `ImageColorTint` all used a nested for loop with some math to index the 1d array of pixels.

These loops could be simplified to a single loop with a counter. I think this is not only more readable, but it has also shown some *slight* speed increases for very large images. My numbers below using the diff of time.h's clock function for a 7096 x 10031 image:

before:
```
brightness 0.932933
replace 0.910925
tint 1.116619
invert 0.785766
```
after:
```
brightness 0.784316
replace 0.782897
tint 0.934186
invert 0.699755
```

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -2649,21 +2649,17 @@ void ImageColorTint(Image *image, Color color)
     float cB = (float)color.b/255;
     float cA = (float)color.a/255;
 
-    for (int y = 0; y < image->height; y++)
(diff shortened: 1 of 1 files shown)


raysan5:
@Minnowo Nice improvement! Thank you very much!

PR #3714, opened 2024-01-07
Enable animation names for the m3d model format.

kolunmi:
Hello! I am not sure why this line was commented out, but this patch seems to work fine. Thank you very much.


raysan5:
@kolunmi thanks for the review!

PR #3715, opened 2024-01-09
Update to latest zig 0.12.0dev while keeping 0.11.0 compatibility

freakmangd:
The recent 0.12.0dev update caused some breakage in the build system.

This method does require the use of `anytype`, but the argument name and type checks should make it obvious what type it expects. This should not break any existing code.

Tested on linux with zig 0.12.0dev and 0.11.0
```
raylib$ zig build
raylib$ zigup run 0.11.0 build
raylib$ cd examples/
raylib/examples$ zig build
raylib/examples$ zigup run 0.11.0 build
```

Once 0.12.0 is released we can remove all of these compatibility functions

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,7 +1,7 @@
 const std = @import("std");
 const raylib = @import("src/build.zig");
 
-// This has been tested to work with zig 0.11.0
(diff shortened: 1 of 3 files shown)


raysan5:
@freakmangd Thanks for the review!

Issue #3717, opened 2024-01-09
[rtext] This font is missing glyphs


tromero:
### Issue description

When I load the ttf for this pixel font (Little League by chevy ray) and draw text using it, a bunch of letters near the end of the alphabet go missing.

I tested about 11 different fonts and this is the only one that appears to have this issue.

In raylib it appears to act as if it's missing the last few lowercase letters? The font file contains upper and lowercase glyphs for all 26 letters, when I inspect it in Font Book on Mac.

Font file provided here for bug repro purposes only, do not redistribute:
[ChevyRay - Little League.ttf.zip](https://github.com/raysan5/raylib/files/13871623/ChevyRay.-.Little.League.ttf.zip)

### Environment

Emscripten web build, running in Brave browser (chrome, effectively) on an M2 Pro Mac Mini.
Also occurs on iOS Safari browser on iPhone.

### Issue Screenshot
![Screenshot 2024-01-09 at 1 16 10 AM](https://github.com/raysan5/raylib/assets/573471/fc53ce06-0577-4c0e-95f0-2e4f234729d7)

### Code Example

```
	font = LoadFontEx("resources/fonts/ChevyRay - Little League.ttf", 7, nullptr, 0);

	float y = 10;
	float verticalSpacing = tileset.font.baseSize * cameraZoom;
	float size = tileset.font.baseSize*cameraZoom;
	float spacing = 1.0f * cameraZoom;

	DrawTextEx(font, TextFormat("%2i FPS", GetFPS()), Vector2{10, y}, size, spacing, DARKGRAY);
	DrawTextEx(font, "Five big quacking zephyrs jolt my wax bed.", Vector2{10, y += verticalSpacing}, size, spacing, DARKGRAY);
	DrawTextEx(font, "AaBbCcDdEeFfGgHhIiJjKkLlMmNnOoPpQqRrSsTtUuVvWwXxYyZz", Vector2{10, y += verticalSpacing}, size, spacing, DARKGRAY);
	DrawTextEx(font, TextFormat("%2ix%2i", GetRenderWidth(), GetRenderHeight()), Vector2{10, y += verticalSpacing}, size, spacing, DARKGRAY);
	DrawTextEx(font, TextFormat("%0.1f DevicePixelRatio", devicePixelRatio), Vector2{10, y += verticalSpacing}, size, spacing, DARKGRAY);
[...]

PR #3719, opened 2024-01-09
[rtext] Adjust font atlas area calculation so padding area is not underestimated at small font sizes.

tromero:
#3717 

The old padding*4 results in around 80px of padding area when the actual amount of padding area was 144px. Adding the padding to the width and height separately instead of only to width should give the exact amount of area after padding, and should better account for fonts with very wide glyphs. With this change I no longer lose glyphs even at impractically small font sizes in the 1-7px range. The 1.2x constant could potentially be reduced now that the area calculation is more accurate.

image


raysan5:
@tromero thanks for the review! This issue has been actually reviewed many times...

PR #3720, opened 2024-01-10
[rtextures] implemented fill color TODO in ImageResizeCanvas()

LievenPetersen:
`ImageResizeCanvas()` accepts a fill color, that was so far unused, with a TODO to implement it.

This is a fairly simple implementation that fills in the entire canvas, with a minor optimization to copy entire rows once the first one is filled.
The implementation is similar to `ImageClearBackground()`, this could possibly be unified in the future.

I was unsure about the desired level of code commenting, so I opted for extensive comments, in line with the contributing guide.

Tested on Linux:
- build
- functionality
- with valgrind (for invalid read/writes)

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -1741,8 +1741,22 @@ void ImageResizeCanvas(Image *image, int newWidth, int newHeight, int offsetX, i
         int bytesPerPixel = GetPixelDataSize(1, 1, image->format);
         unsigned char *resizedData = (unsigned char *)RL_CALLOC(newWidth*newHeight*bytesPerPixel, 1);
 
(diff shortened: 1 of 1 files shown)


LievenPetersen:
I could also upgrade this to either use `ImageClearBackground()` or to not fill the pixels that will be occupied by the old image's content.

raysan5:
@LievenPetersen Thanks for the review!

 > I could also upgrade this to either use ImageClearBackground() or to not fill the pixels that will be occupied by the old image's content.

I'm thinking about that... actually, if image canvas is smaller than current canvas, it wouldn't be needed to fill anything... but I also think if the filling process is fast enough to skip that sequence of checks with offset/size...

I'm merging for now. If it can be optimized in the future, it can be reviewed.

PR #3721, opened 2024-01-10
Fix `ImageDrawRectangleRec`

Bigfoot71:
There seems to be an error in `ImageDrawRectangleRec` where we should subtract the absolute values of `rec.x` and `rec.y` from `rec.width` and `rec.height` when they are negative.

Currently, we are subtracting the negative values, which is equivalent to addition. For example, in the current implementation:
GIF1

If we replace `size -= position` with `size += position`, considering that the position is negative, we correctly subtract the difference as desired. This results _(the visual glitches are from my screen recorder)_:
GIF2

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -3562,8 +3562,8 @@ void ImageDrawRectangleRec(Image *dst, Rectangle rec, Color color)
     if ((dst->data == NULL) || (dst->width == 0) || (dst->height == 0)) return;
 
     // Security check to avoid drawing out of bounds in case of bad user data
-    if (rec.x < 0) { rec.width -= rec.x; rec.x = 0; }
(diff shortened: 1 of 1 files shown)


Bigfoot71:
And for the remaining visible pixel line in the second gif, this can be resolved by replacing the condition:
```c
if (((rec.x + rec.width) < 0) || (rec.y + rec.height < 0)) return;
```
With:
```c
if (((rec.x + rec.width) <= 0) || (rec.y + rec.height <= 0)) return;
```
I can also make the modification if you prefer.

raysan5:
@Bigfoot71 Oh! Good catch! Thanks for the review!

About the second proposed change:

```c
if (((rec.x + rec.width) <= 0) || (rec.y + rec.height <= 0)) return;
```

Did you verify the drawn rectangle is not one pixel wider or higher?

PR #3724, opened 2024-01-11
Update a comment

casavaca:
in #3523 we moved `RecordAutomationEvent()` from after `PollInputEvents()` to before it, but the comment was not updated.


raysan5:
@casavaca Thanks for the comment review!

PR #3727, opened 2024-01-11
Add shared library build option for build.zig and update to zig 0.12.0-dev.2139

candrewlee14:
Using raylib as a shared library was helpful for hot-module reloading, allowing both the main executable and shared game library to link with raylib and then we can reload without segfaulting. I needed the no-sanitize flag to work around an illegal instruction error.

Here is the source for that example: https://github.com/candrewlee14/tetris-raylib

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,6 +2,6 @@ const std = @import("std");
 const raylib = @import("src/build.zig");
 
 // This has been tested to work with zig 0.11.0 and zig 0.12.0-dev.2075+f5978181e
-pub fn build(b: *std.Build) void {
-    raylib.build(b);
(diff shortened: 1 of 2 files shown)


raysan5:
@candrewlee14 It seems this PR is related to this one: https://github.com/raysan5/raylib/pull/3680 but your implementation seems simpler. I'm merging it.

Peter0x44:
@candrewlee14 shouldn't this be defining BUILD_LIBTYPE_SHARED and also passing -fvisibility=hidden?

Peter0x44:
Nevermind, it does that
Not sure how I missed it - my bad

PR #3728, opened 2024-01-13
Make raylib/raygui work better on touchscreen

casavaca:
Hi,

I'm trying raylib on my iPad using emscripten. I found that raylib and raygui don't work well. We need updated mouse position yet it's never updated (always 0,0). As a result, [Cubic-bezier lines](https://www.raylib.com/examples/shapes/loader.html?name=shapes_lines_bezier) doesn't work on iPad.

I don't fully understand how it works on desktop browser, but it may be reasonable to update mouse position of we are using touch screen?

This is not pretty, but after this fix, raylib and raygui work just fine.
If you have better fix, then it's great. Just kindly discard this pull request.


casavaca:
I see. I've updated the comments.

casavaca:
Bump

raysan5:
@casavaca Thanks for the review. Checking the code it shouldn't be a problem to just map it... let's see...

PR #3735, opened 2024-01-16
[rtext] Add BDF font support

stanthesoupking:
Added support for loading BDF fonts (spec version 2.1).

Side-note: in the future we may want to add the parameter `fileType` to `LoadFontData` so it can support more formats than just TTF. BDF font loading was implemented using `LoadBDFFontData`, so it wouldn't be difficult to make the change to support this if we wanted.


raysan5:
@stanthesoupking Thanks for the addition! Really nice!

Issue #254, opened 2017-04-02
[build] Add -D _POSIX_C_SOURCE=200809L to CFLAGS


sanderboer:
Hi,

I'm just starting out with raylib, the emscripten layer is what interests me.
When building raylib, make fails on an implicit function declaration. I can't seem to be able to use nanosleep() (or usleep() ) without this error popping up. 
Apparently, the _POSIX_C_SOURCE macro must be defined with an appropriate value for the POSIX extensions to be available:

`ifeq ($(PLATFORM),PLATFORM_WEB)
    CFLAGS = -D _POSIX_C_SOURCE=200809L -O1 -Wall -std=c99 -s USE_GLFW=3 -s ASSERTIONS=1 --preload-file resources
    #-s ALLOW_MEMORY_GROWTH=1   # to allow memory resizing
    #-s TOTAL_MEMORY=16777216   # to specify heap memory size (default = 16MB)
endif
`

raysan5:
Hi @sanderboer, this part was changed just some days ago... Are you using develop branch?

Maybe @RDR8 could give us some feedback, it was reviewed in PR https://github.com/raysan5/raylib/pull/248

sanderboer:
Hi @raysan5 ,
I just checked out the master branch...

S

RDR8:
Since you're using c99, use _DEFAULT_SOURCE is the flag you, as when compiling PLATFORM=PLATFORM_DESKTOP on Linux.

RDR8:
No idea what is actually going on here. @sanderboer is right, master is broken but the changes made to the development branch at 44de97e are working for me "as is" on Linux. That is, without any _SOURCE type flags. Maybe --profilng CFLAG is the diff.

RDR8:
@sanderboer If that code snippet above is in your makefile, check the syntax of -D_POSIX_C_SOURCE=200809L

RDR8:
@sanderboer If that code snippet above is in your makefile, check the syntax of -D_POSIX_C_SOURCE=200809L

sanderboer:
That code snippet is in my makefile, I added the -D_POSIX_C_SOURCE=200809L for it to compile.
Or do you mean, I'd try it with -D_DEFAULT_SOURCE instead of -D_POSIX_C_SOURCE=200809L ?

sanderboer:
Wait, 
[...]

PR #3737, opened 2024-01-17
[rcore rglfw] Feature Test Macros before include

JayLCypher:
Reference to #254 

Original issue seems resolved by build step (make, cmake, zig), however looking at the source code, it seems the original intent was to include the Feature Test Macros per file. This seems logical for self-documenting purposes and outline requirements for raylib.
However, the macros are defined after includes, so they won't work.
See: [GNU Feature Test Macros](https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html)
> You should define these macros by using ‘#define’ preprocessor directives at the top of your source code files. These directives must come before any #include of a system header file. It is best to make them the very first thing in the file, preceded only by comments.

I've simply moved the rcore _POSIX_C_SOURCE feature test macro to the top and included _XOPEN_SOURCE for readlink function.
Alternative change to rcore would be to define_POSIX_C_SOURCE to 200809L, which removes the need to define _XOPEN_SOURCE >= 500.

These changes allow for compilation with -std=c* (such as -std=c99) without adding -D macros to the build step, for the specially interested. ~~(any tsoding nobbers in chat?)~~

Additionally, added the macros to rglfw, such that the external includes have their respective functions defined.
I went through the other modules and compiled their object files with both -std=c99 and -std=c2x, and found no other required macros. Also ran the Github Actions for Linux, Windows, Web, Android with no issues.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -82,6 +82,19 @@
 *
 **********************************************************************************************/
 
+//----------------------------------------------------------------------------------
+// Feature Test Macros required for this module
+//----------------------------------------------------------------------------------
+#if (defined(__linux__) || defined(PLATFORM_WEB)) && (_XOPEN_SOURCE < 500)
+    #undef _XOPEN_SOURCE
+    #define _XOPEN_SOURCE 500 // Required for: readlink if compiled with c99 without gnu ext.
+#endif
+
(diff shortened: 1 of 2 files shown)


raysan5:
@JayLCypher Thanks for the review, I can't see any potential issue with this change...

PR #3740, opened 2024-01-19
Add frame functions

zap8600:
This PR adds the `StartFrame()` and `EndFrame()` functions, allowing for platforms that need to do some setup before and after frame rendering to be added. An example platform that could now be added is the Meta Quest 2 using OpenXR, which needs `xrStartFrame()` and `xrEndFrame` to be called before and after frame rendering.


src/platforms/rcore_desktop.c
```diff
@@ -996,6 +996,18 @@ void DisableCursor(void)
     CORE.Input.Mouse.cursorHidden = true;
 }
 
+// Start the frame
+void StartFrame(void)
+{
+    TRACELOG(LOG_WARNING, "StartFrame() not implemented on target platform");
```
justgo97, line 1002: I haven't tested this but wouldn't this cause a lot of warnings spam in the console/logs in debug mode?
Probable just leave the function body empty and just add a comment instead.

justgo97:
Also, why not simply write your start frame logic before `BeginDrawing` and end frame logic after `EndDrawing`? For convenience, you can wrap `BeginDrawing` and `EndDrawing` in your application code to achieve the same result.

zap8600:
> Also, why not simply write your start frame logic before `BeginDrawing` and end frame logic after `EndDrawing`? For convenience, you can wrap `BeginDrawing` and `EndDrawing` in your application code to achieve the same result.

OpenXR has those functions that need to be called, and I'd like to keep the platform-specific code in Raylib and away from my application.

orcmid:
@zap8600 
> OpenXR has those functions that need to be called, and I'd like to keep the platform-specific code in Raylib and away from my application.

If you are developing an application using OpenXR, that's far more challenging than those two functions.  Are you expecting to integrate the same way OpenGL and Vulkan are?

raysan5:
@zap8600 Thanks for the PR but I'm afraid this addition can be managed at user side and there is no need to include it in raylib. It really seems to be it belongs to a more high-level engine than a low level library. Also, advance users with this requirement can enable `SUPPORT_CUSTOM_FRAME_CONTROL` for more control over the game loop and this kind of additional functions.

PR #3744, opened 2024-01-20
[rshapes] Fix multisegment Bezier splines.

spelufo:
It seems to me that these functions are wrong, if you step the index by 1 you move to a control point instead of the next segment.

### DIFF
--- a/examples/shapes/shapes_splines_drawing.c
+++ b/examples/shapes/shapes_splines_drawing.c
@@ -63,7 +63,7 @@ int main(void)
     Vector2 *focusedControlPoint = NULL;
     
     // Cubic Bezier control points initialization
-    ControlPoint control[MAX_SPLINE_POINTS] = { 0 };
+    ControlPoint control[MAX_SPLINE_POINTS-1] = { 0 };
     for (int i = 0; i < pointCount - 1; i++)
     {
         control[i].start = (Vector2){ points[i].x + 50, points[i].y };
@@ -88,6 +88,9 @@ int main(void)
         if (IsMouseButtonPressed(MOUSE_RIGHT_BUTTON) && (pointCount < MAX_SPLINE_POINTS))
         {
             points[pointCount] = GetMousePosition();
+            int i = pointCount - 1;
+            control[i].start = (Vector2){ points[i].x + 50, points[i].y };
+            control[i].end = (Vector2){ points[i + 1].x - 50, points[i + 1].y };
             pointCount++;
         }
 
@@ -114,7 +117,7 @@ int main(void)
         if ((splineTypeActive == SPLINE_BEZIER) && (focusedPoint == -1))
         {
             // Spline control point focus and selection logic
-            for (int i = 0; i < pointCount; i++)
+            for (int i = 0; i < pointCount - 1; i++)
             {
                 if (CheckCollisionPointCircle(GetMousePosition(), control[i].start, 6.0f))
                 {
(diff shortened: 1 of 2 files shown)


raysan5:
@spelufo Did you test it? Could you share some code example?

spelufo:
I did, through raylib-go:

```
package main

import rl "github.com/gen2brain/raylib-go/raylib"

func main() {
  rl.InitWindow(800, 600, "raylib [rshapes] example - bezier")
  defer rl.CloseWindow()

  rl.SetTargetFPS(60)

  for !rl.WindowShouldClose() {
    rl.BeginDrawing()
    rl.ClearBackground(rl.RayWhite)
    points := []rl.Vector2{
      {0.0, 0.0}, {50.0, 0.0}, {50.0, 50.0},
      {100.0, 50.0}, {100.0, 100.0}, {250.0, 250.0},
      {500.0, 500.0},
    }

    // Does the wrong thing.
    rl.DrawSplineBezierCubic(points, 5.0, rl.Green)

    // Proposed implementation:
    for i := 0; i < len(points) - 3; i += 3 {
      color := rl.Orange
      p := points[i]
      rl.DrawSplineSegmentBezierCubic(p, points[i + 1], points[i + 2], points[i + 3], 5.0, color)
    }

    rl.EndDrawing()
  }
}
```
Screenshot from 2024-01-30 15-05-41

raysan5:
@spelufo I'm trying this fix with example `shapes_splines_drawing`, replacing current `DrawSplineSegmentBezierCubic()` in a loop by the proposed solution but it doesn't seem to work as expected. Please could you review it?

Drawing code I'm using:

```c
Vector2 fullPoints[14] = {
    points[0],
    control[0].end,
    control[1].start,
    points[1],
    control[1].end,
    control[2].start,
    points[2],
    control[2].end,
    control[3].start,
    points[3],
    control[3].end,
    control[4].start,
    points[4],
    control[4].end
};

DrawSplineBezierCubic(fullPoints, 14, splineThickness, RED);
```

image

The expected result is:

image

spelufo:
I've fixed the example code. I tested that the previous version of `DrawSplineBezierCubic` didn't work right. The new version works with this example code. The sequence of points passed to it must be `points[0], control[0].start, control[0].end, points[1], ..., points[pointCount-1]`, and the length of `control` must always be `pointCount - 1`.

spelufo:
Regarding the "connections", the ugly gaps left on the points between segments, one way to cover them up would be to have `DrawSplineBezierCubic` draw a circle at the points connecting segments of diameter equal to the thickness of the lines. Maybe not the most efficient way to do it but it would cover the kinks.

spelufo:
I've gone ahead and did that. It is ready to merge if you want.

spelufo:
@raysan5 ?

raysan5:
@spelufo Excuse the late response! I'm quite busy lately...

I want to review this code more carefully because I think the current `DrawSplineSegmentBezierCubic();` implementation could not be the best one...

colinwilcox:
Correcting the sequence of points and the iterator seems good. 
I'm not sure I agree with drawing a circle in between each segment though. At the very least it should be optional. 

I'll take a look at the `DrawSplineSegmentBezierCubic()`  implementation later. Do you think it is not performant?

raysan5:
> I'm not sure I agree with drawing a circle in between each segment though. At the very least it should be optional and default off.

@colinwilcox I agree, circle should be in any case, optional, I considered using a flag for it... 

 > I'll take a look at the `DrawSplineSegmentBezierCubic()` implementation later. Do you think it is not performant?
```c
DrawSplineSegmentBezierCubic(Vector2 p1, Vector2 c2, Vector2 c3, Vector2 p4, float thick, Color color);
```

My concerns with the function were the expected sequence of points, they should follow the pattern:

image

After review, it seems so. I also reviewed `DrawSplineBezierCubic()` and it seems it also receives and processes the lit of points as expected. 

So, it seems the only issue is how `shapes_splines_drawing` example organizes the points to be passed to the function; that this PR seems to address.

In conclusion, raylib Cubic Bezier functions seem to be correct but the example could be confusing because it separates `ControlPoints` from function points for Cubic Bezier, and this is done this way because it tries to map the start-end drawing points to other spline types to be able to change between them and show a somewhat same-points curve.

Ok, I'm merging the PR and I will review it with those comments on the example.

PR #3746, opened 2024-01-21
[CORE,RLGL] Fix scale issues when ending a mode

JeffM2501:
There are 3 places in raylib that set the OpenGL modelview matrix scale to the screen scale (for HighDPI).
1 of these is incorrect, BeginMode2d, as the camera is going to do it's own scale.
The other two assume that when we end 2d and 3d mode that we are going back to the screen. That is not always true. BeginMode2d/3d can be used with a render texture and a render texture can be any size and doesn't have to match the screen DPI scale (it's not a screen).

This PR exposes the GL function to get the active draw FBO, and then has raylib use that to know what render target it is restoring to, and if it needs to apply the scale. 

This change should make using a render texture more consistent.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -1734,7 +1734,7 @@ unsigned int rlGetActiveFramebuffer(void)
     glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &fboId);
     return fboId;
 #endif
-	return 0;
+    return 0;
 }
 
 // Disable rendering to texture
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -621,6 +621,7 @@ RLAPI void rlDisableShader(void);                       // Disable shader progra
 // Framebuffer state
 RLAPI void rlEnableFramebuffer(unsigned int id);        // Enable render texture (fbo)
 RLAPI void rlDisableFramebuffer(void);                  // Disable render texture (fbo), return to default framebuffer
+RLAPI unsigned int rlGetActiveFramebuffer(void);        // Returns the active render texture (fbo), 0 for default framebuffer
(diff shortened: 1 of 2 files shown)


JeffM2501:
For testing, I ran several examples, ones that do and don't use render textures and camera modes. I tested them in both high DPI and regular dpi modes and everything continued to work as expected

src/rlgl.h
```diff
@@ -1725,6 +1726,17 @@ void rlEnableFramebuffer(unsigned int id)
 #endif
 }
 
+// return the active render texture (fbo)
+unsigned int rlGetActiveFramebuffer(void)
+{
+#if (defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES3)) && defined(RLGL_RENDER_TEXTURES_HINT)
```
TokyoSU, line 1732: It's not really a problem but in this case if it's OPENGL 33+ or ES3
then you have:
```
return fboId;
return 0;
```
you should use:
```
GLint fboId = 0;
#if (defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES3)) && defined(RLGL_RENDER_TEXTURES_HINT)
glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &fboId);
#endif
return fboId ;
```
or:
```
#if (defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES3)) && defined(RLGL_RENDER_TEXTURES_HINT)
GLint fboId = 0;
glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &fboId);
return fboId ;
#else
return 0;
#endif
```
Just a minor think, it won't change the return anyway.

raysan5:

Issue #3742, opened 2024-01-20
[shapes] False positive with `CheckCollisionPointPoly()`


mllimo:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

The "CheckCollisionPointPoly" function gives me a false positive when I check that two instances of squares form by 4 vertices are colliding particulary when one vertex is inside the other. I have noticed that it happens only when I reach the same or lower "Y" coordinate from the right side or when I surround it in clockwise direction. See the videos below for more clarity.

### Environment

Windows 11 x64, OpenGL 3.3, AMD Radeon RX 6600

### Issue Screenshot

https://github.com/raysan5/raylib/assets/38075437/bce679e1-ada7-4467-b86b-4292ba58c5fe
![debugging](https://github.com/raysan5/raylib/assets/38075437/23a105cb-18ac-4324-93fe-ba27afd144e0)
![verticesrepresentation](https://github.com/raysan5/raylib/assets/38075437/76315a8e-4abe-4ee9-b97d-1fbf913acede)

### Code Example

```cpp
int main()
{
    std::vector<Vector2> vertices_b = {
        { 668.678406, 60.9284592 },
        { 629.318420, 53.7983131 },
        { 622.188721, 93.1581116 },
        { 661.548462, 100.287857 }
    };
   
    // Colliding: 1 
    std::cout << "Colliding: " << CheckCollisionPointPoly({ 100, 100 }, (Vector2*)vertices_b.data(), vertices_b.size()) << '\n';
}
```

raysan5:
@mllimo thanks for reporting! `CheckCollisionPointPoly()` was a contributed function, let's see if someone could take a look to it.

PR #3750, opened 2024-01-21
[shapes] Fix CheckCollisionPointPoly

mllimo:
I tested this changes with the same cases shown in #3742 and goes nice 🙌.


raysan5:
@mllimo Nice! Thanks for the review!

PR #3752, opened 2024-01-22
[RLGL] fixed typo, replaced `th` with `the`

Brian-ED:
Fixed the typo in the comment `We are building he library as ...` by replacing "he" with "the".


raysan5:
@Brian-ED Thanks for the fix!

PR #3753, opened 2024-01-22
[raymath.h] Small code refactor

idircarlos:
The pull request is divided into 2 small commits, so you may don't want to merge both.
1. I identified two functions that have duplicated code, so I refactored to avoid it. You might don't want to merge this change because you might prefer the readability, but I just wanted to give it a try in case you agree with it :)
2. A ternary operator didn't have proper left space as the rest. Some divisions have both spaces also.


raysan5:
@idircarlos Nice! Thanks for the review!

PR #3754, opened 2024-01-22
Add Vector2Dir function to get the direction vector of two given points

Elkantor:
I always add this function anyway on my raylib projects, so I think that could be usefull to get this function directly in raymath.
But it's really not mandatory.
Note: the function name is quite "quick and dirty", but hey, it is probably enough?

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -253,6 +253,14 @@ RMAPI Vector2 Vector2Add(Vector2 v1, Vector2 v2)
     return result;
 }
 
+// Get direction vector of two points
+RMAPI Vector2 Vector2Dir(Vector2 p1, Vector2 p2)
+{
+    Vector2 result = { p2.x - p1.x, p2.y - p1.y };
+
+    return result;
+}
+
 // Add vector and float value
 RMAPI Vector2 Vector2AddValue(Vector2 v, float add)
 {


Peter0x44:
Isn't this just Vector2Subtract with the arguments flipped? What's the need for this?

Elkantor:
> Isn't this just Vector2Subtract with the arguments flipped? What's the need for this?

Yes, totally! Indeed, but actually it's just the name which matters. Anyway, I didn't even took attention to Vector2Substract! Gosh I'm tired haha

idircarlos:
I understand the intention of @Elkantor. Calculating the direction vector of a line (or a plane) is a common operation when working with vectors. Here is my opinion on this. In case we are adding the function, I would: 
- Rename it to `Vector2Direction`, which is a more self-explanatory.
- `Vector2 result = Vector2Subtract(v2, v1);` instead of `Vector2 result = { p2.x - p1.x, p2.y - p1.y };`
- Add the same function for a plane (defined by 3 points)

But this is just my opinion :) I hope it helps.

raysan5:
@Peter0x44 thanks for looking into this, I was aware of some similar function already available.

@Elkantor @idircarlos A `Vector2Direction()` could be useful but probably as a normalized vector, afaik a direction is more useful normalized, right? Not sure if there is any similar alternative.

PR #3755, opened 2024-01-22
[raudio.c] Reformated some comments

idircarlos:
Changed some comments format to align comments style.


raysan5:
@idircarlos Thank you very much! I really like this kind of format improvements! I try to put a lot of care in formating and I always miss some things, usually from PRs... Thanks!

PR #3757, opened 2024-01-23
[rcore] Send raylib logs to stderr by default

Peter0x44:
from man stderr:

       Under  normal circumstances every UNIX program has three streams opened
       for it when it starts up, one for input, one for output,  and  one  for
       printing diagnostic or error messages.

It's a more appropriate place for programs to send logs, as it's a stream meant explicitly for it.

stderr is unbuffered, so it doesn't need to be flushed.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -45,7 +45,7 @@
 #endif
 
 #include <stdlib.h>                     // Required for: exit()
-#include <stdio.h>                      // Required for: FILE, fopen(), fseek(), ftell(), fread(), fwrite(), fprintf(), vprintf(), fclose()
+#include <stdio.h>                      // Required for: FILE, fopen(), fseek(), ftell(), fread(), fwrite(), fprintf(), vfprintf(), fclose()
 #include <stdarg.h>                     // Required for: va_list, va_start(), va_end()
 #include <string.h>                     // Required for: strcpy(), strcat()
 
@@ -147,8 +147,7 @@ void TraceLog(int logType, const char *text, ...)
     unsigned int textSize = (unsigned int)strlen(text);
     memcpy(buffer + strlen(buffer), text, (textSize < (MAX_TRACELOG_MSG_LENGTH - 12))? textSize : (MAX_TRACELOG_MSG_LENGTH - 12));
     strcat(buffer, "\n");
-    vprintf(buffer, args);
-    fflush(stdout);
+    vfprintf(stderr, buffer, args);
 #endif
 
     va_end(args);


raysan5:
@Peter0x44 Is this a change that will work on all raylib supported platforms with no downsides?

orcmid:
@raysan5 @Peter0x44 
>As per my understanding, stderr should be used mostly for error/diagnostic messages but raylib uses `TraceLog()` for all kind of traces, including info.

I think as long as it is something that not meant to be seen by an end-user, especially when the app is launched quietly and no command prompt appears, using `stderr` may be the best option, whether or not there is an error case.  

There might need to be some granularity regarding the purpose  and intended presentation of a log message though.

Peter0x44:
@orcmid I remember Jeff testing this and he could see those messages.
Very strange choice for windows to be making - are you sure of it?

orcmid:
@Peter0x44 
> @orcmid I remember Jeff testing this and he could see those messages. Very strange choice for windows to be making - are you sure of it?

Yes, I verified it with a test build I keep around.  That uses 

```
/link /SUBSYSTEM:WINDOWS /ENTRY:mainCRTStartup kernel32.lib user32.lib shell32.lib winmm.lib gdi32.lib opengl32.lib
```

along with the *.obj files of the main program and of a cache of compiled raylib *.c files.

See the compilation pattern in `raylib\projects\scripts\build-windows.bat`.  The `/ENTRY:mainCRTStartup` might be the culprit.  I don't use logs, so hadn't noticed the issue with `stderr` and this build pattern.

Peter0x44:
/SUBSYSTEM:WINDOWS
Pretty sure it's this. Try CONSOLE instead.

orcmid:
@Peter0x44 
> /SUBSYSTEM:WINDOWS Pretty sure it's this. Try CONSOLE instead.

You got it.  When I use `CONSOLE` and run from a command prompt, I get a ton of log messages.  Even if I launch the .exe directly, I get a console window with all of that output, as well as the app in its own window.

LOL.  I've always used scripts based on the one in the raylib/projects/ folder.  I had no idea there was all this output, even though I have noticed all of the LOG messages in the raylib source code.  

So, in my case, for a production release, I will always use the `/SUBSYSTEM:WINDOWS` and have to come up with something else if I want error messages/notices outside of the app window.  That will be interesting.

Peter0x44:
What this linker flag does is set a byte in the PE header that controls whether windows attached a console to your executable or not, that's all. And if there's no console, there's nowhere for your logs to go.

It has nothing to do with stdout vs stderr, I think you should have been unable to see any message at all for both cases.

Anyway, it's not a point against using stderr. I maintain it's the correct choice, for all sorts of logs that end users aren't supposed to see.

raysan5:
I'm still investigating about this change, most of the log INFO output by raylib is status info, despite some WARNING and some ERROR in very specific situations. I can't see why all the status info should be directed to `stderr` or the benefits it will imply for raylib. Also, I don't know if that's the best behaviour for all the supported platforms.

Some opinions out there: https://stackoverflow.com/questions/4919093/should-i-log-messages-to-stderr-or-stdout

raysan5:
After listening to multiple points of view on a raylib Discord discussion and also on a [Twitter question](https://twitter.com/raysan5/status/1756378993593401658), I decided to stay with the `stdout` output for now. Some of the reasons exposed:

 - Other libraries used by raylib like [GLFW seem to rely on `stdout`](https://github.com/glfw/glfw/blob/master/src/init.c#L343) while SDL log output is platform-dependant, with a preference for `OutputDebugString()` (stdout) on Windows, custom Log output on Android and `stderr` on other platforms. Keeping consistency with GLFW is a good reason to stay on `stdout`.
 - It seems many programmers consider `stderr` specific for error messages while *most* raylib output messages are `LOG_INFO` or esporadically some `LOG_WARNING`... actually there is only one `LOG_ERROR` message in all the raylib library.
 - There is already available a callback function to redirect tracelog as user needs.

PR #3758, opened 2024-01-23
[rmodels]: Fix bug to retrieve material specular map in DrawMesh function, with incorrect index

Elkantor:


raysan5:
@Elkantor Oh! Almost forgot about this issue! Thanks for the review!

PR #3759, opened 2024-01-24
Fix rlSetUniformSampler

veins1:
Fix rlSetUniformSampler not setting sampler uniform if the same texture was already passed to a different sampler uniform


raysan5:
@veins1 Good catch! Thanks for the review!

Issue #17, opened 2015-01-14
Poor CPU Performance


kd7tck:
When running simple text based games, the CPU usage can be as high as 30%. It can be traced back to a poor implementation of system polling with GLFW.

raysan5:
After further investigation, it seems that high (100% or 25% for quad-cores) CPU usage is the expected behavior for most cases in OpenGL/Direct3D games, **it mainly depends on graphics driver implementation**. Game loop just keeps running all the time if not locked in any way. In GLFW3 is possible to glfwWaitEvents() instead of glfwPollEvent() so frame swap can be locked waiting for events and CPU usage falls to ~1% but that's not the desired behaviour in a game. Enabling VSYNC could have a similar effect but it also depends on graphics driver...

A couple of links with further info:

http://stackoverflow.com/questions/18365269/cpu-usage-is-high-when-using-opengl-control-class
http://gamedev.stackexchange.com/questions/19066/why-is-opengl-using-100-cpu-shortly-after-start-up-when-wglswapintervalext-is-e

kd7tck:
Having an option in the lib API for vsync might help allot. When vendor drivers have trouble with vsync, the opengl software can make up for that and override the driver settings.

```c++
#ifdef _WIN32
    // Turn on vertical screen sync under Windows.
    // (I.e. it uses the WGL_EXT_swap_control extension)
    typedef BOOL (WINAPI *PFNWGLSWAPINTERVALEXTPROC)(int interval);
    PFNWGLSWAPINTERVALEXTPROC wglSwapIntervalEXT = NULL;
    wglSwapIntervalEXT = (PFNWGLSWAPINTERVALEXTPROC)wglGetProcAddress("wglSwapIntervalEXT");
    if(wglSwapIntervalEXT)
        wglSwapIntervalEXT(1);
#endif
```

raysan5:
GLFW already uses `WGL_EXT_swap_control` extension to provide VSYNC (where this extension is available) but, again, it can be overriden by graphic driver, forcing on or off. Here it is further information on `Swap Internal`:

https://www.opengl.org/wiki/Swap_Interval

But there is something I can do; I can expose GLFW VSYNC mechanism to raylib users through a flag, to be set before windows creation. That way users can try forcing VSYNC and wait for the driver to respond.

I'm working on it.

kd7tck:
Thank you for the time and effort you have put into this.

raysan5:
[...]

PR #3760, opened 2024-01-26
Added missing "standalone" functions to raudio.c & fixed return bug

sanyokdev:
NOTICE: THIS IS THE SAME COMMIT AS [#17](https://github.com/raysan5/raudio/pull/17) IN RAUDIO. I PUT THIS HERE TO HAVE BOTH REPOS SYNCED.

Hey,
While I updated raudio [#16](https://github.com/raysan5/raudio/pull/16), I also found some bugs & afterthoughts with "raudio.c".

First, as GetFileNameWithoutExt, GetFileName & strprbrk were linked to Raylib itself I had to add them to "raudio.c" under the "RAUDIO_STANDALONE" define for standalone mode to work. Point at which the functions were added:
- (15-04-2023) RadsammyT: "[[raudio] Rewritten ExportWaveAsCode() file saving to be more like rtextures ExportImageAsCode() (raysan5#3013)](https://github.com/sanyokdev/raylib/commit/e2da32e2daf2cf4de86cc1128a7b3ba66a1bab1c)"

Second, I use a lot compiler flags so it found that "SaveFileData" & "SaveFileText" in "raudio.c" were missing return values. Surprisingly a lot of variables inside "raudio.c" were actually using the values returned, that were never actually given.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -2745,9 +2745,19 @@ static bool SaveFileData(const char *fileName, void *data, int dataSize)
 
             fclose(file);
         }
-        else TRACELOG(LOG_WARNING, "FILEIO: [%s] Failed to open file", fileName);
+        else
+        {
+            TRACELOG(LOG_WARNING, "FILEIO: [%s] Failed to open file", fileName);
+            return false;
+        }
     }
-    else TRACELOG(LOG_WARNING, "FILEIO: File name provided is not valid");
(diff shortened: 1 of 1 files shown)


raysan5:
@sanyokdev Thanks for the review! As `raudio` is integrated into raylib this issue was not detected before...

PR #3761, opened 2024-01-26
Added GetTraceLogLevel function

idircarlos:
Just added a (not very important) function that could be useful for some users to get the current log level at runtime.


raysan5:
@idircarlos I think this function is not needed, it's up to the user to set it and register this value.

PR #3762, opened 2024-01-28
[rcamera.h] cleaned away unused macros in rcamera.h

Brian-ED:
I assume these were accidentally left in. they were made in the redesign https://github.com/raysan5/raylib/pull/2563 but the macros were used in the finished pull. I guess a new redesign since then rendered the macros useless.


raysan5:
@Brian-ED Good catch! Definitely those defines are not required any more, they belong to previous implementation.

PR #3769, opened 2024-02-01
Fix segfault in ExportWaveAsCode

IoIxD:
`char *txtData = (char *)RL_CALLOC(waveDataSize * 6 + 2000, sizeof(char));`

assumes every chunk being added to txtData is 6 bytes. This is not always true, sometimes a newline is involved and the data becomes 12 bytes instead, and this can cause a random segfault.

This commit changes `6` to `12`, and explains why in the comment.


raysan5:
@IoIxD thanks for the review, actually those 2000 extra bytes at the end should consider that potential overflow...

IoIxD:
Thats what I thought you tried to do. But in my own testing it is not sufficient.

Issue #3745, opened 2024-01-20
[rtext] `MeasureTextEx()` does not report correct height when using multi line string


karl-zylinski:
### Issue description

When sending in a multi line string to `MeasureTextEx` it does not report the correct height. The text drawn by `DrawTextPro` will often be taller.

### Environment

Windows 10, GL3 (I guess), GTX 2080 Ti

### Issue Screenshot

Drawn multi line text with a dark background color, note that the dark background is less tall than the text:

![image](https://github.com/raysan5/raylib/assets/6352002/d7dcf3aa-86bc-4b55-a320-369ad4c50605)

### Code Example

```
Vector2 ts = MeasureTextEx(some_font, "hi\nho\nha", 20, 0)

text_r := Rectangle {
    0,0,
    ts.x,
    ts.y,
}

DrawRectangleRec(text_r, BLACK)
DrawTextPro(some_font, "hi\nho\nha", {0, 0}, {}, 0, 20, 0, WHITE)
```

I hand translated the code above from Odin to C, so I'm not sure it works out of the box. Also I use a custom font, I'm afraid I don't have time to setup test with non-custom font right.

raysan5:
@karl-zylinski Excuse the late answer, I will try to review it asap. The line-break spacing support was added quite recently, are you using latest raylib version from GitHub master branch?

raysan5:
@karl-zylinski This issue should be fixed with merged PR.

PR #3770, opened 2024-02-01
Fix MeasureTextEx() height calculation

marrony:
For multiline strings `MeasureTextEx()` is returning a much higher value than the actual text height.

Pretty much like `DrawTextEx()` this patch just increment `textLineSpacing` in the height when it sees `\n` and don't apply `scaleFactor` at the end.

Small program to simulate the issue:
```c
#include <stdlib.h>
#include "raylib.h"

int main(void) {
  InitWindow(200, 100, "MeasureTextEx");

  Font font = GetFontDefault();

    BeginDrawing();
    ClearBackground(WHITE);

    const char *text = "Some\nMultiline\nText";
    float fontSize = 20;
    float spacing = 5;

    SetTextLineSpacing(fontSize + 4);
    DrawTextEx(font, text, (Vector2){10, 10}, fontSize, spacing, BLACK);

    // draw red box around text
    Vector2 m = MeasureTextEx(font, text, fontSize, spacing);
    DrawRectangleLinesEx((Rectangle){10, 10, m.x, m.y}, 1, RED);

    EndDrawing();
  }

  CloseWindow();
  return 0;
}
```

Without the fix
Screenshot 2024-01-31 at 5 40 00 PM

With the fix
Screenshot 2024-01-31 at 5 39 25 PM

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1257,7 +1257,7 @@ Vector2 MeasureTextEx(Font font, const char *text, float fontSize, float spacing
     float textWidth = 0.0f;
     float tempTextWidth = 0.0f;     // Used to count longer text line width
 
-    float textHeight = (float)font.baseSize;
+    float textHeight = fontSize;
     float scaleFactor = fontSize/(float)font.baseSize;
 
     int letter = 0;                 // Current character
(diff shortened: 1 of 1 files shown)


raysan5:
@marrony thanks for the review! It should fix issue https://github.com/raysan5/raylib/issues/3745

PR #3771, opened 2024-02-01
[rcore] Fix `GetFileNameWithoutExt()`

oblerion:
```c
puts( GetFileNameWithoutExt("dir/file.name.zip") ); 
// out : "file"
```
[with fix]
```c
puts( GetFileNameWithoutExt("dir/file.name.zip") );
// out : "file.name"
```
Use reverse search '.' for remove extension.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1946,26 +1946,17 @@ const char *GetFileName(const char *filePath)
 // Get filename string without extension (uses static string)
 const char *GetFileNameWithoutExt(const char *filePath)
 {
-    #define MAX_FILENAMEWITHOUTEXT_LENGTH   256
-
-    static char fileName[MAX_FILENAMEWITHOUTEXT_LENGTH] = { 0 };
-    memset(fileName, 0, MAX_FILENAMEWITHOUTEXT_LENGTH);
-
-    if (filePath != NULL) strcpy(fileName, GetFileName(filePath));   // Get filename with extension
-
-    int size = (int)strlen(fileName);   // Get size in bytes
-
-    for (int i = 0; (i < size) && (i < MAX_FILENAMEWITHOUTEXT_LENGTH); i++)
+    const char *filename = GetFileName(filePath);
+    int pos = TextLength(filename);
+    for(int i = pos; i>0; i--)
     {
-        if (fileName[i] == '.')
+        if(filename[i] == '.')
         {
-            // NOTE: We break on first '.' found
-            fileName[i] = '\0';
+            pos = i;
             break;
         }
     }
-
-    return fileName;
+    return TextSubtext(filename,0,pos);
 }
 
 // Get directory for a given filePath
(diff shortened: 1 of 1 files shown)


Peter0x44:
Not entirely related to the PR but this is one of the (many) operations that becomes *so* much easier if you stop using null-terminated strings

```c
#define s8(s) (s8) { (unsigned char*) s, (ptrdiff_t) sizeof(s)-1 }

typedef struct
{
	unsigned char* str;
	ptrdiff_t len;
} s8;

s8 GetExtensionlessFilename(s8 fileName)
{
	s8 ret = fileName;
	for (ptrdiff_t i = ret.len-1; i >= 0; --i)
	{
		--ret.len;
		if (ret.str[i] == '.') return ret;
	}
	return fileName;
}
```

You can eliminate the need to copy the string (and therefore skip the hardcoded size static buffer), and also the O(n) useless strlen call.

It's only an issue if you need to pass the string later on to something that expects null-terminated, which is unfortunately common.

(not suggesting or implying raylib should change its string convention - just wanted to "spread awareness" I guess)

src/rcore.c
```diff
@@ -1946,26 +1946,23 @@ const char *GetFileName(const char *filePath)
...
-        if (fileName[i] == '.')
+        const char *fileName = GetFileName(filePath); // Get filename.ext without path
+        for (int i = (int)strlen(fileName); i>0; i--) // Reverse search '.'
         {
-            // NOTE: We break on first '.' found
-            fileName[i] = '\0';
-            break;
+            if (fileName[i] == '.')
+            {
+                char rfileName[i];
```
raysan5, line 1956: Afaik, some compilers do not allow variable-length allocations on the stack without using a malloc(). Did you try this code on Linux or Web platforms?

src/rcore.c
```diff
@@ -1946,26 +1946,23 @@ const char *GetFileName(const char *filePath)
...
+                strncpy(rfileName, fileName,i);
+                rfileName[i] = '\0';
+                const char* rrfileName = rfileName;
```
raysan5, line 1959: It seems it's returning a pointer to an stack-allocated string...

src/rcore.c
```diff
@@ -1946,26 +1946,23 @@ const char *GetFileName(const char *filePath)
...
+                return rrfileName;
+            }
         }
+        return fileName; // Extension not found
```
raysan5, line 1963: In raylib I tried to avoid, as much as possible, any early return. I prefer to assign required value and wait to the end of the function to return.

src/rcore.c
oblerion: Yes i use linux

src/rcore.c
oblerion: Ok i understand why you do that, i back with fix.

oblerion:
It's fix thank for help Peter0x44 and raysan5

raysan5:
@oblerion thank you very much for the review!

PR #3775, opened 2024-02-02
[rtextures][rlgl] Implement UpdateImageFromTexture

purrie:
I've added ability to update existing image from a texture. The functions in rtextures.c and rlgl.h are pretty much copies of existing functions that create a new image but they use the provided image buffer instead.

I wanted this functionality since my project needs to read rendered pixels in real time for external use and I wanted to avoid allocating memory every frame.

I have successfully tested the addition on GNU/Linux. I don't have access to other systems to do other tests.

Please let me know if any adjustments need to be done. Thank you.

Purrie.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1305,6 +1305,7 @@ RLAPI Image LoadImageAnim(const char *fileName, int *frames);                   
 RLAPI Image LoadImageAnimFromMemory(const char *fileType, const unsigned char *fileData, int dataSize, int *frames); // Load image sequence from memory buffer
 RLAPI Image LoadImageFromMemory(const char *fileType, const unsigned char *fileData, int dataSize);      // Load image from memory buffer, fileType refers to extension: i.e. '.png'
 RLAPI Image LoadImageFromTexture(Texture2D texture);                                                     // Load image from GPU texture data
+RLAPI void UpdateImageFromTexture(Image *image, Texture2D texture);                                      // Load image from GPU texture data into existing image
 RLAPI Image LoadImageFromScreen(void);                                                                   // Load image from screen buffer and (screenshot)
 RLAPI bool IsImageReady(Image image);                                                                    // Check if an image is ready
 RLAPI void UnloadImage(Image image);                                                                     // Unload image from CPU memory (RAM)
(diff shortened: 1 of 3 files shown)


raysan5:
@purrie I prefer to avoid those functions. Users requiring them can implement them at user side. Updating an image from texture is actually a non-recommended practice.

purrie:
@raysan5 That's fair. It is a very specific functionality that is not needed for most people. I should have thought of that before making the PR. Sorry.

I am curious however as to why it is not recommended to update images from textures directly tho.

purrie:
@raysan5 Alright. Thanks for the explanation. That makes sense. In my case, I'm sending part of rendered scene out to a virtual camera so, unfortunately it's not something I can solve with shaders.

raysan5:
@purrie It seems you have a specific use case, did you try using `LoadImageFromTexture()` and blit the required part of the image? did you compare the performance of that approach vs your specific implementation?

purrie:
@raysan5 For full context, I'm making an attempt to create a vtuber/pngtuber type of software. The goal in this use case is to render animated avatar to be displayed as part of a video broadcast such as Twitch or Youtube. 

From what I understand, most other software of this type rely on capturing the program window to display it on the stream but I find this approach unsatisfactory since it mixes user interface which is intended for the broadcaster to manipulate the animated avatar and the rendered output to be displayed on the stream for viewers, so instead, I want to render the avatar and other parts intended to be seen by viewers separately and send it to a virtual camera so the broadcaster can hook just that into the stream without the user interface that's normally displayed in the program window.

The approach I decided on is to render the avatar into a render texture which is then loaded from the GPU into an image buffer so I can process it into a format expected by the virtual camera driver and send it there to be picked up by OBS Studio. The rest of the interface is then rendered on top of the already rendered texture of the avatar in the program window just for the broadcaster.

In normal game or tool software, I imagine the performance of loading a texture into an image is not that important since you normally don't do it that often. In my case tho, I need to render every frame for both the window and the camera and from my measurements, loading the image versus updating existing one is on average about 5 to 8 times slower. With `LoadImageFromTexture()` taking 1.5 to 2.5 ms on average while `UpdateImageFromTexture()` usually being around 0.3 ms on my machine. Which makes sense since the only thing my implementation differs in is absence of malloc call and reliance on user providing the memory buffer instead, which can be allocated just once.

raysan5:
@purrie Good to know your use case. Indeed it seems quite specific to your project.

Issue #1295, opened 2020-06-29
[utils] Support CPU/GPU stats analytics


raysan5:
It could be useful to keep track of some CPU/GPU analytics, like memory usage.

Info already displayed on initialization: GPU vendor, renderer, version, GLSL version

Data loading memory usage:
 - CPU: Images / Sounds / MusicStream / Font.chars
 - GPU: Textures / Shaders / Meshes / Font.texture

Data loading can be registered with:  `RL_MALLOC()` / `RL_CALLOC()` / `RL_FREE()`

```c
enum StatsMemType {
    CPU_RAW_DATA,
    CPU_VERTEX_DATA, 
    CPU_IMAGE, 
    CPU_TEXT, 
    CPU_FONT_DATA,
    CPU_AUDIO_BUFFER, 
    GPU_VERTEX_BUFFER, 
    GPU_TEXTURE, 
    GPU_SHADER, 
    GPU_FBO
} StatsMemType;

// Global memory usage stats
struct StatsMemAlloc {
    int id;
    int type;               // StatsMemType 
    unsigned int size;      // Size in bytes
} StatsMemAlloc;

#define STATS_MEM_RECORD(int type, int size) StatsMemAlloc(type, size)
```

 - GPU memory allocations (VRAM): `glBufferData()`, `glTexImage2D()`, `glShaderSource()`
 - Frame times: `updateTime`, `drawTime`, `idleTime`
 - Draw calls counter (and batching)
 - Textures information (# textures, id, format, samplerState)

This issue is a draft, just some notes. Need to think about the best way to organize/record this information.

```c
typedef struct GpuStats {	// 1 frame? General? → FrameStats?
	int drawsCount;
	int vertexCount;
	int trianglesCount;
	int texturesCount;	// total? used in frame?
	int texturesMemory;	// in bytes
	int vertexMemory;	// in bytes, in CPU/GPU?
	float timeUpdate;	// in seconds
	float timeDraw;		
	float timeIdle;
} GpuStats;
```

Obscure2020:
Most games on Steam list their minimum required RAM and VRAM. It would be handy to have some sort of way to get a raylib program to spit out the maximum amount of RAM and VRAM it ever used while running, so that we know what to put for the minimum required stats for games we publish that were made using raylib. What do you think?

raysan5:
@Obscure2020 Actually that's some of the information intended to be provided by this new feature.

Obscure2020:
[...]

PR #3781, opened 2024-02-04
[core] Create function to retrieve frame busy time (non-sleeping time)

bmbkr:


bmbkr:
I still need to update documentation throughout the repo to include this new function.

That being said, I'm eagerly awaiting your thoughts on this new functionality @raysan5

raysan5:
@bmbkr I prefer to avoid exposing this internal state, most users probably do not need it and advance users can implement it for their needs or just enable `SUPPORT_CUSTOM_FRAME_CONTROL` on `config.h` to have full control over the frame.

Exposing this state would imply exposing other internal states that I think most users do not need at the moment.

bmbkr:
@raysan5 Understood. Perhaps this can be exposed at some point in the future as part of a metrics interface, simply as a ratio of busy to sleep time. Would be particularly useful for performance measurement, although as you said you can just recompile with custom frame control.

raysan5:
@bmbkr there was a new feature in that line in the past (https://github.com/raysan5/raylib/issues/1295) but it required to many changes and coupling and it was discarded...

PR #3782, opened 2024-02-04
premake support

DxrxDev:
I added a basic premake5 file. 

It's only been tested with linux on platform-desktop.


raysan5:
@DxrxDev I prefer to avoid a new build system, it requires additional maintenance work that I can not assume. There is already a separate project in that line in `raylib-extras`: https://github.com/raylib-extras/game-premake

PR #3783, opened 2024-02-04
[rcore] Initial Dreamcast platform support

psxdev:
Hello, these are the initial changes that we are using right now to work with Dreamcast and Naomi boards
Changes including my own patch plus code from MikeDX and suggestion about prefix install by Tchan

We need to decide which ones could be incorporated in main stream and what changes could be done in gldc to ask them fix there.


raysan5:
@psxdev Very nice! Thanks for the addition! I will keep the PR open while checking what things could be directly implemented in main branch.

raysan5:
Reviewed `rlgl` missing OpenGL flags, `rlgl` does not need to be changed anymore.

Also, `rshapes` module will be reviewed on raylib side to allow QUADS-only mode (no LINES required).

PR #3788, opened 2024-02-07
Add camera sensitivity to the Camera3D struct

jakub-swiniarski:
Hello, this PR adds camera sensitivity control to the Camera3D struct. Currently, controlling camera sensitivity requires a custom camera implementation and custom camera functions. I've also updated the examples to showcase how to use the new sensitivity field.
Compiles successfully on GNU/Linux.


raysan5:
@jakub-swiniarski I'm afraid this is a breaking change that I prefer to avoid.

PR #3793, opened 2024-02-10
[CORE] Fix warnings

JeffM2501:
Just another pass on casting and warnings to keep things clean :)


raysan5:
@JeffM2501 Nice! Thanks for the review!

PR #3795, opened 2024-02-12
[rtext] Change strcpy to strncpy to fix buffer overflow

szsam:


raysan5:
@szsam Thanks for the review!

Issue #3665, opened 2023-12-23
[rtextures] `IsTextureReady()` returns false after successful call of `LoadTextureCubemap()`


kolunmi:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

Hello! The title is pretty self-explanatory.

### Environment

Arch linux:
```
INFO: GL: OpenGL device information:
INFO:     > Vendor:   AMD
INFO:     > Renderer: AMD Radeon RX 7900 XTX (radeonsi, navi31, LLVM 16.0.6, DRM 3.54, 6.6.7-arch1-1)
INFO:     > Version:  4.6 (Core Profile) Mesa 23.3.1-arch1.1
INFO:     > GLSL:     4.60
```

### Code Example

```c
#include <stdio.h>
#include <raylib.h>

int main(int argc, char **argv) {
  InitWindow(800, 600, "Test");

  Image img = GenImageColor(600, 100, WHITE);
  Texture tex = LoadTextureCubemap(img, CUBEMAP_LAYOUT_LINE_HORIZONTAL);
  UnloadImage(img);

  bool success = IsTextureReady(tex);
  printf("IsTextureReady() reports a %s!\n", success ? "success" : "failure");

  UnloadTexture(tex);
  CloseWindow();

  return 0;
}
```
yields:
```
INFO: TEXTURE: [ID 3] Cubemap texture loaded successfully (100x100)
IsTextureReady() reports a failure!
```

Thank you for this amazing library!

raysan5:
@kolunmi thanks for reporting! reviewed!

raysan5:
@kolunmi By the way, this issue appeared because there has never been a proper cubemap loading example... :(

amitansu1024:
@raysan5 @kolunmi Can I work on this if somebody isn't working on it already?

raysan5:
@amitansu1024 Feel free, actually, a new `textures_cubemap.c` example would be nice to illustrate `LoadTextureCubemap()` capabilities that probably have not been tested enough.

Dev-Tade:
Has been this solved yet?

raysan5:
@Dev-Tade it does not seem so

PR #3798, opened 2024-02-16
Fix issue #3665

CMDisme:
Set cubemap format in LoadTextureCubemap


CMDisme:
Fixes #3665

raysan5:
@CMDisme how did you test this change if there is no code example to test it?

CMDisme:
I copied the code from the issue

raysan5:
@CMDisme Ok, I see, but I'm afraid that code does not verify if the loaded cubemap is actually valid. And example would be required to review it.

PR #3799, opened 2024-02-17
Fix implicit conversion from float to double then back to float again

jfoscarini:
Proposal to fix unnecessary promotion from float to double.

`implicit conversion increases floating-point precision: 'float' to 'double' [-Wdouble-promotion]`


raysan5:
@jfoscarini Good catch! Thanks for the improvement! 👍🙂

PR #3804, opened 2024-02-19
Fixed loading GLTF animations with 1 frame

BliznyukNM:
When I was working on my 3D project I found an issue that single frame animations (like T-Pose) are not loaded. Upon investigating the reason I realised that frame count calculated based on animation length (which is 0 apparently).

Using some basic combinatorics I added 1 at the end of frame count calculations. It is easy to think of a veeeery small animation (like 0.01 second) which still has to have minimum 2 frames. Animations with 0 length should have at least 1.

Also **fmax** is preventing NaN division result when **tend** and **tstart** are both 0.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5368,7 +5368,7 @@ static bool GetPoseAtTimeGLTF(cgltf_accessor *input, cgltf_accessor *output, flo
         }
     }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@BliznyukNM Nice! Thanks for the review!

PR #3805, opened 2024-02-19
Remove all uses of deps/mingw

Peter0x44:
The purpose of this directory in GLFW is to provide some headers that "mingw.org" doesn't. Raylib has long been unable to build with mingw.org due to using certain symbols that aren't exposed in their headers. (_ftelli64 and _access, among others.). Mingw-w64 already has the necessary headers included, and doesn't need any of these external implementations. For some reason, this also causes the following error when building with Visual Studio's clang:

clang  -c rglfw.c -Wall -D_GNU_SOURCE -DPLATFORM_DESKTOP -DGRAPHICS_API_OPENGL_33 -Wno-missing-braces -Werror=pointer-arith -fno-strict-aliasing  -std=c99 -O1 -Werror=implicit-function-declaration -I. -Iexternal/glfw/include -Iexternal/glfw/deps/mingw In file included from rglfw.c:61:
In file included from ./external/glfw/src/init.c:30: In file included from ./external/glfw/src/internal.h:331: In file included from ./external/glfw/src/platform.h:31: In file included from ./external/glfw/src/win32_platform.h:70: external/glfw/deps/mingw\xinput.h:227:26: error: a parameter list without types is only allowed in a function definition
  227 | void WINAPI XInputEnable(WINBOOL);
      |

If the last -Iexternal/glfw/deps/mingw is removed, the build works fine.

So, this workaround causes other problems, while not actually helping raylib.

https://github.com/glfw/glfw/blob/0bb605cd797e4d63709495f4074ec59362064ab4/src/CMakeLists.txt#L272-L279

GLFW's CMakeLists.txt first checks if either dinput.h or xinput.h are provided by the toolchain, before telling the compiler to search for headers in that directory. For EVERY compiler that can build raylib, this is true.

In summary: This directory causes issues when building with some compilers, and every toolchain that needs this workaround can't build raylib anyway.

### DIFF
--- a/examples/build.zig
+++ b/examples/build.zig
@@ -55,7 +55,6 @@ fn add_module(comptime module: []const u8, b: *std.Build, target: anytype, optim
                 exe.linkSystemLibrary("winmm");
                 exe.linkSystemLibrary("gdi32");
                 exe.linkSystemLibrary("opengl32");
-                exe.addIncludePath(.{ .path = "external/glfw/deps/mingw" });
 
                 exe.defineCMacro("PLATFORM_DESKTOP", null);
             },
--- a/src/Makefile
+++ b/src/Makefile
@@ -438,7 +438,7 @@ INCLUDE_PATHS = -I.
 
 # Define additional directories containing required header files
 ifeq ($(PLATFORM),PLATFORM_DESKTOP)
-    INCLUDE_PATHS += -Iexternal/glfw/include -Iexternal/glfw/deps/mingw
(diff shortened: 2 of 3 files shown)


Peter0x44:
I do not use zig, so I didn't test with it. I'm very confident these changes should make zero difference for zig, but I'd still appreciate testing from a zig user.

raysan5:
@Peter0x44 Great improvement! I was wondering when would be possible to remove it! Thanks!

PR #3810, opened 2024-02-22
Fix SDL multitouch tracking

awfulcooking:
> The fingerId from SDL was used as an index into the CORE.Input.Touch arrays, but it's an opaque / arbitrary int64, way bigger than MAX_TOUCH_POINTS, so the first non-simulated touch event would segfault.

Working on Windows tablet (ARM64 :)

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -956,6 +956,23 @@ void SetMouseCursor(int cursor)
     CORE.Input.Mouse.cursor = cursor;
 }
 
+static void UpdateSDLTouchPoints(SDL_TouchFingerEvent event)
+{
+    CORE.Input.Touch.pointCount = SDL_GetNumTouchFingers(event.touchId);
+
+    for (int i=0; i<CORE.Input.Touch.pointCount; i++)
+    {
+        SDL_Finger *finger = SDL_GetTouchFinger(event.touchId, i);
+        CORE.Input.Touch.pointId[i] = finger->id;
+        CORE.Input.Touch.position[i].x = finger->x * CORE.Window.screen.width;
+        CORE.Input.Touch.position[i].y = finger->y * CORE.Window.screen.height;
+        CORE.Input.Touch.currentTouchState[i] = 1;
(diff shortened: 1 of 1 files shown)


src/platforms/rcore_desktop_sdl.c
```diff
@@ -994,17 +1011,6 @@ void PollInputEvents(void)
...
 
-    // Reset touch positions
-    // TODO: It resets on target platform the mouse position and not filled again until a move-event,
-    // so, if mouse is not moved it returns a (0, 0) position... this behaviour should be reviewed!
-    //for (int i = 0; i < MAX_TOUCH_POINTS; i++) CORE.Input.Touch.position[i] = (Vector2){ 0, 0 };
-
-    // Map touch position to mouse position for convenience
-    // WARNING: If the target desktop device supports touch screen, this behavious should be reviewed!
-    // https://www.codeproject.com/Articles/668404/Programming-for-Multi-Touch
-    // https://docs.microsoft.com/en-us/windows/win32/wintouch/getting-started-with-multi-touch-messages
```
awfulcooking, line 1005: This is done by default in SDL_touch.c anyhow:

SDL_touch.c:34
```c
#define SYNTHESIZE_TOUCH_TO_MOUSE 1
```

awfulcooking: So, it has been restored:

```c
        CORE.Input.Touch.position[0] = CORE.Input.Mouse.currentPosition;
```

I deleted the comments:

>     // WARNING: If the target desktop device supports touch screen, this behavious should be reviewed!

It is hereby reviewed!

>    // https://www.codeproject.com/Articles/668404/Programming-for-Multi-Touch
>    // https://docs.microsoft.com/en-us/windows/win32/wintouch/getting-started-with-multi-touch-messages

These are hangovers from rcore_desktop.c being used as a template for this backend.

GLFW doesn't support touch, so Ray linked to docs for Windows touch as a future reference.

But not relevant here as SDL implemented that :-)

raysan5:
@awfulcooking thanks for the improvement

PR #3812, opened 2024-02-22
[build] Use sdl2-config for SDL paths in Makefile

Sqvid:
The proposed changes use `sdl2-config` to  set the include and library paths for SDL. This allows users to install the SDL2 development files via their preferred method (e.g. package manager) and then just pass PLATFORM=PLATFORM_DESKTOP_SDL to build.

This is my first contribution here so please let me know if I've missed something. Thank you.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -98,8 +98,8 @@ USE_EXTERNAL_GLFW     ?= FALSE
 
 # PLATFORM_DESKTOP_SDL: It requires SDL library to be provided externally
 # WARNING: Library is not included in raylib, it MUST be configured by users
-SDL_INCLUDE_PATH      ?= $(RAYLIB_SRC_PATH)/external/SDL2/include
-SDL_LIBRARY_PATH      ?= $(RAYLIB_SRC_PATH)/external/SDL2/lib
+SDL_INCLUDE_PATH      ?= $(shell sdl2-config --cflags)
+SDL_LIBRARY_PATH      ?= $(shell sdl2-config --libs)
 
 # Use Wayland display server protocol on Linux desktop (by default it uses X11 windowing system)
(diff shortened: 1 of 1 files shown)


Sqvid:
Thank you for the feedback. In the case of SDL2, `sdl2-config` is the way upstream recommends linking to SDL2  so I think it makes sense that it be the default. Users could overwrite the variables if they were savvy enough to clone and compile separately. Lastly, SDL2's `make install` does install `sdl2-config`.

awfulcooking:
For what it's worth.. vcpkg on Windows does not build an sdl2-config.exe by default.

But it does install sdl2.pc files.

Maybe pkgconf is a better option.

Sqvid:
Maybe relying on CMake for this (#3809) is the better way forward then.

awfulcooking:
pkgconf might still be a good option for the Makefile..

It'd mean `make PLATFORM=PLATFORM_DESKTOP_SDL` will just work for most Linux users.

I think even if you `make install` SDL from source.

Sqvid:
> pkgconf might still be a good option for the Makefile..
> 
> It'd mean `make PLATFORM=PLATFORM_DESKTOP_SDL` will just work for most Linux users.
> 
> I think even if you `make install` SDL from source.

I would really like that too, however it would not address the workflow mentioned by @ubkp I think. @ubkp any thoughts?

raysan5:
@Sqvid Thanks for the review but I agree with @ubkp that it adds and external dependency and additional complexity that is not required.

@ubkp Note that this is an open-source project with a big community and multiple contributions, some are better for the project than others. I try to do my best to review them and integrate the ones that I think will improve the project. Simplicity is always my priority.

Issue #3815, opened 2024-02-23
[core] Gamepad rumble with SDL2


GideonSerf:
Could I add a method to support gamepad rumble? I saw issue https://github.com/raysan5/raylib/issues/3122, and while it's still not possible to add rumble support using GLFW, it would be possible to add it for SDL2 via https://wiki.libsdl.org/SDL2/SDL_JoystickRumble. 

The following function prototype could be used:
`void SetGamepadVibration(int gamepad, float leftMotor, float rightMotor, float durationMs)`.  With leftMotor and rightMotor being values between 0.0 and 1.0 and durationMs being a duration in milliseconds. When this function is called on a platform that doesn't support this function, `TRACELOG(LOG_WARNING, "GamepadSetVibration() not available on target platform");` could be called. I think as more platforms are added in the future, this could be a useful function to have.

raysan5:
@GideonSerf Yes! That could be a good addition. But... if time is provided I imagine it would require some global variables to track time, if possible I'd keep it as `SetGamepadVibration(int gamepad, float leftMotor, float rightMotor)` letting the user to manage time and reset the vibration with `SetGamepadVibration(0, 0.0f, 0.0f)`.

GideonSerf:
No I don't think it's an OS API. SDL seems to use a countdown timer to manage when the command to stop the rumble should be sent, as well as sending the rumble command periodically while it should be ongoing. Unfortunately SDL doesn't make it possible to call the function without a duration, but when you call SDL_JoystickRumble with the same vibration values it currently has, it just resets the countdown timer. So perhaps the rumble values can be stored in the Gamepad struct and then resent in PollInputEvents until both values are set to 0?

raysan5:
> So perhaps the rumble values can be stored in the Gamepad struct and then resent in PollInputEvents until both values are set to 0?

I prefer to avoid that route and let the full time control to the user. If I understand correctly, calling `SDL_JoystickRumble()` twice with a `0` duration it just stops the rumble, right?

So, raylib can just define a `MAX_GAMEPAD_VIBRATION_TIME        2.0f     // Time in seconds` and let the user manage other timming by themselfs with:
```c
SetGamepadVibration(0, 0.5f, 0.5f);  // Start vibration

// User manages desired vibration time
timeCounter += GetFrameTime();

[...]

PR #3819, opened 2024-02-24
Gamepad rumble support with SDL2

GideonSerf:
As discussed in issue https://github.com/raysan5/raylib/issues/3815, I added a function `SetGamepadVibration(int gamepad, float leftMotor, float rightMotor)`. I also added a new define, `MAX_GAMEPAD_VIBRATION_TIME 2.0f` to control the maximum duration of the vibration. When the user sends `SetGamepadVibration(gamepad, float 0.0f, float 0.0f)`, it cancels the vibration. Otherwise it continues on for MAX_GAMEPAD_VIBRATION_TIME. 

It seems to work well on my end with an 8BitDo Pro 2 controller set both to Nintendo Switch Pro and Xbox One modes. Perhaps the max time could be changed to something else, but 2 seconds seem to be fine.

Currently the function is set as not returning a value, the SDL function returns -1 if rumble isn't supported on the controller. Should something like that be returned from the SetGamepadVibration function as well?

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -81,6 +81,7 @@
 #define MAX_GAMEPADS                    4       // Maximum number of gamepads supported
 #define MAX_GAMEPAD_AXIS                8       // Maximum number of axis supported (per gamepad)
 #define MAX_GAMEPAD_BUTTONS            32       // Maximum number of buttons supported (per gamepad)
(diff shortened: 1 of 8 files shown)


raysan5:
@GideonSerf Nice addition! Thanks! 👍🙂

PR #3821, opened 2024-02-24
Add parsing for gltf u16 vec4 joints in rmodels.c

GaryMcWhorter:
Follows the format of the previous cases.


raysan5:
@GaryMcWhorter Merging for reference but it requires a future review, I0m adding some comments now.

PR #3823, opened 2024-02-24
Fixes IsTextureReady() returns false after successful call of LoadTextureCubemap()

GaryMcWhorter:
fixes https://github.com/raysan5/raylib/issues/3665

The format of the cubemap was not being set and was defaulted to 0, causing IsTextureReady() to fail.


GaryMcWhorter:
unlike PR https://github.com/raysan5/raylib/pull/3798 this only assigns to format after the cubemap is successfully loaded, using the same format passed into rlLoadTextureCubemap.

Tested locally using a generated linear gradient:

gltf-fix_45TQQEJNgb

PR #3824, opened 2024-02-24
Fix ExportDataAsCode not considering some characters for variable name

lauchimoon:
With a `fileName` argument like `my-image.png.h`, the output inside `my-image.png.h` looks like this:
```c
#define MY-IMAGE.PNG_DATA_SIZE     119076

static unsigned char MY-IMAGE.PNG_DATA[MY-IMAGE.PNG_DATA_SIZE] = ...
```
This commit checks if characters such as `-`, `.`, `?` and `!` are present in `varFileName` and replaces them with `_`, so `my-image.png.h` would be
```c
#define MY_IMAGE_PNG_DATA_SIZE     119076

static unsigned char MY_IMAGE_PNG_DATA[MY_IMAGE_PNG_DATA_SIZE] = ...
```

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -316,7 +316,11 @@ bool ExportDataAsCode(const unsigned char *data, int dataSize, const char *fileN
     // Get file name from path and convert variable name to uppercase
     char varFileName[256] = { 0 };
     strcpy(varFileName, GetFileNameWithoutExt(fileName));
-    for (int i = 0; varFileName[i] != '\0'; i++) if ((varFileName[i] >= 'a') && (varFileName[i] <= 'z')) { varFileName[i] = varFileName[i] - 32; }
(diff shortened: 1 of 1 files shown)


Peter0x44:
I think you have not considered enough. Filenames beginning with '+' are also not okay, as an example.
There are also multiple ExportXAsCode functions where this needs to be considered.
See #3814 as well.

lauchimoon:
Right, I checked Ray's decision on this topic and will close the pr. Thanks!

raysan5:
@lauchimoon I finally decided to implement this change, I merged a similar PR (#3832) but I think yours is better, please, could you resent it?

lauchimoon:
Hi @raysan5, sure thing! If there's any other functions I should modify please lmk :)

lauchimoon:
Will open a new commit because I've made a mess with the conflicts...

PR #3825, opened 2024-02-25
Fix examples linking with CMake and -DPLATFORM=SDL

Peter0x44:
Currently, every example fails linking likeso:

[  3%] Linking C executable audio_mixed_processor
/usr/bin/ld: ../raylib/libraylib.a(raudio.c.o): undefined reference to symbol 'exp@@GLIBC_2.29' /usr/bin/ld: /usr/lib/libm.so.6: error adding symbols: DSO missing from command line collect2: error: ld returned 1 exit status

Apparently, linking libm explicitly is the solution.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -131,6 +131,9 @@ foreach (example_source ${example_sources})
     add_executable(${example_name} ${example_source})
 
     target_link_libraries(${example_name} raylib)
+    if (NOT WIN32)
+        target_link_libraries(${example_name} m)
+    endif()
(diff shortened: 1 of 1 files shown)


Peter0x44:
Yikes. This breaks building on visual studio. Will investigate further later.

Peter0x44:
Okay, libm shouldn't be linked on "WIN32". Should be fine now.

Peter0x44:
The Linux CI build failiure is some network issue, not related.

raysan5:

PR #3826, opened 2024-02-25
Fix building of raylib_opengl_interop on PLATFORM_DESKTOP_SDL

Peter0x44:
I tested this with CMake. I don't know if the examples makefile currently handles this case.
The resulting executable runs and produces the expected output.

NOTE: testing this PR depends on #3825, otherwise this example won't even link due to the issue that PR resolves.

Though I hope it's trivial enough that there aren't problems I missed.

### DIFF
--- a/examples/others/raylib_opengl_interop.c
+++ b/examples/others/raylib_opengl_interop.c
@@ -26,7 +26,7 @@
 
 #include "raylib.h"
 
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 thanks for the review... I'm wondering if this example should belong to raylib, it's platform specific and it seems to give many build issues.

PR #3827, opened 2024-02-25
Update GLFW to 3.4

M374LX:
As the title suggests, this PR updates GLFW to version 3.4.

It seems to work on Linux with both X11 and Wayland (tested on Weston), but we cannot use GLFW's new runtime platform detection at the moment when building raylib with the Makefile instead of CMake, as explained in https://github.com/raysan5/raylib/pull/3827#issuecomment-1963041768. Maybe we can merge the PR as is and check this issue at a different moment.

As for testing on other platforms, I would like to ask for help from other users, although all checks from GitHub Actions are successful as of 6a58cbb.

**Update:** ~It looks like gamepads have stopped working~ (actually, only one of the gamepads I tested has stopped working, which is probably a rare case and we don't need to care).


raysan5:
Thanks for starting this update! It was in my TODO list! But note it requires testing on Windows, Linux and macOS.

planetis-m:
Might need to revert https://github.com/raysan5/raylib/pull/2883 since https://github.com/glfw/glfw/commit/2c3eb75748584c1c9283787e8c40f944fa055d6f removed the dependency on wayland-protocols.

M374LX:
@planetis-m Reverting #2883 is not necessary, as 260a039 contains the necessary changes.

M374LX:
The new GLFW version supports platform detection at runtime, which enables a single executable to run on both X11 and Wayland. However, it looks like raylib cannot offer the same platform detection functionality at the moment because some of GLFW's local (static) functions in different platforms have the same name. As a results, we cannot combine source files from multiple platforms into one, as done in ``rglfw.c``, unless GLFW's developers agree to change the names of these functions.

gen2brain:
@M374LX I just updated GLFW to 3.4 in raylib Go bindings. I did rename some functions in https://github.com/gen2brain/raylib-go/commit/8596a5db729eea07b31804d0421fd5536cbefcd3 though. I know this will affect a few more projects, like https://github.com/go-gl/glfw/issues/393, so I hope it will be solved, but for now, it is not a big deal, at least for me. There are also changes for Wayland generated headers, they are renamed, and there are some additions, I use a script for that https://github.com/gen2brain/raylib-go/commit/5abcc8b79ce4184ebac57f111dfbf40391fa7d64.

Now the single binary runs on both X11 and Wayland, you can check the output:
```
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not provide the window position
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not provide the window position
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not support setting the window position
```

It would be nice to check `glfwGetPlatform` and add that to i.e. `INFO: Platform backend: DESKTOP (GLFW)`.

raysan5:
This is a bit change, it needs some careful review... but it will be merged.

M374LX:
05ba358 implements @gen2brain's suggestion to output the platform selected by GLFW to TRACELOG.

raysan5:
@M374LX @gen2brain  Changes look good to me for now, please, let me know when you think this PR is ready for merge.

Some point to consider:

 - Are gamepad mappings updated? It has been manually updated in the past.
 - Is this PR tested on all GLFW supported platforms? It should be tested on: Windows, Linux, macOS and Web.

M374LX:
The gamepad mappings are supposed to be the ones provided by GLFW.

I am able to test this PR on Linux and can confirm that it works on both X11 and Wayland (despite https://github.com/raysan5/raylib/pull/3827#issuecomment-1963041768), but I prefer to leave the tests on Windows, macOS and Web for someone with access to these systems and Emscripten set up.

Otherwise, I think it is ready to be merged.

raysan5:
@M374LX Thank you very much for this improvement!

PR #3831, opened 2024-02-25
Fix the link error of examples, cmake error, and charset error

### LINKED ISSUE #3818: [build] [examples] `raylib_opengl_interop` linking issue on Windows
### Issue description

Use Cmake to generate NMake makefiles to build shared library, and link error appeared when building.

### Environment

* Platform: desktop
* OS: windows 11 23H2
* OpenGL: version 3.3
* Under Visual Studio 2022 Developer PowerShell v17.9.1

### Issue Screenshot

螢幕擷取畫面 2024-02-24 151834

螢幕擷取畫面 2024-02-24 151944

螢幕擷取畫面 2024-02-24 152215

some translations:
* 無法解析的外部符號: unresolved external symbol
* 傳回碼: return code

Pan-Shin-Z:
fixes #3818 

Fix the link error of `raylib_opengl_interop` and `rlgl_standalone` by adding function implementation.
Set utf-8 as default charset for MSVC; otherwise, there would be an error when encoding `text_codepoints_loading.c`.
Fix the error `add_if_flag_compiles` by removing cache in `AddIfFlagCompiles.cmake`.

### DIFF
--- a/examples/others/raylib_opengl_interop.c
+++ b/examples/others/raylib_opengl_interop.c
@@ -38,6 +38,9 @@
             #include <OpenGL/gl3.h>     // OpenGL 3 library for OSX
             #include <OpenGL/gl3ext.h>  // OpenGL 3 extensions library for OSX
         #else
+            #if defined(_WIN32) && defined(_MSC_VER)
+                #define GLAD_IMPLEMENTATION  // Need self-implementation for MSVC
+            #endif
             #include "glad.h"       // Required for: OpenGL functionality 
         #endif
         #define GLSL_VERSION            330
--- a/cmake/AddIfFlagCompiles.cmake
+++ b/cmake/AddIfFlagCompiles.cmake
@@ -9,4 +9,7 @@ function(add_if_flag_compiles flag)
     set(outcome "compiles")
   endif()
   message(STATUS "Testing if ${flag} can be used -- ${outcome}")
+  # It must be manually removed from the cache; otherwise, the check won't be re-evalated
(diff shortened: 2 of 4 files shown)


examples/others/raylib_opengl_interop.c
```diff
@@ -38,6 +38,9 @@
             #include <OpenGL/gl3.h>     // OpenGL 3 library for OSX
             #include <OpenGL/gl3ext.h>  // OpenGL 3 extensions library for OSX
         #else
+            #if defined(_WIN32) && defined(_MSC_VER)
+                #define GLAD_IMPLEMENTATION  // Need self-implementation for MSVC
```
Peter0x44, line 42: Can you explain why the "self-implementation" is needed?

Pan-Shin-Z: For MSVC, there are only function declarations in glad.h such that the linker can't find the function definitions(or symbols, I'm not familiar with these terms). I need to define `GLAD_IMPLEMENTATION` to offer the function definitions.
No need to define `GLAD_IMPLEMENTATION` in Linux, MSYS and MinGW. I tested it. However, to be honest, I'm not clear about the reasons.

Peter0x44: Something about it doesn't seem right. I think it should not be necessary, on MSVC also, and something else is wrong.

Pan-Shin-Z: I also think so. I found some errors happened after these changes. `raylib_opengl_interop` does not work as expected.

Does I need to close this pull request?

PR #3832, opened 2024-02-25
Add very little sanitization to indentifier names in ExportDataAsCode()

Dev-Tade:
Hello, I decided to add this little sanitization for the '-' characater to the exported identifiers since find it quite useful for me and may help to other raylib users that use '-' in their file names.

Without this change any exported data to a file name containing the character would not compile since it's not a valid character for identifiers, I could have added more characters to be replaced ('@', '#', '%', etc) but i don't think they are really used for the use cases of `ExportDataAsCode()`.

Thanks for your time ❤️


raysan5:
@Dev-Tade Thanks for the review, I was a bit reticent about this change but I will merge this one. Still, it will require some review and also alignment wit all the other `Export*AsCode()` functions in raylib.

Issue #3834, opened 2024-02-25
[build] Short description of the issue/bug/feature


hardliner66:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

Today I worked the whole day on a game in zig + raylib, when suddenly the compiler complains about not finding one of the raylib headers. This was kinda confusing, because it already worked earlier. I tried restarting my environment and deleting and cloning my repository to make sure everything is clean, but it still has the same error.

I'm not 100% sure if the problem is related to raylib directly or if it's a bug in the zig compiler, but I thought I'd start here and find out.

The problem seems to be, that it want's the file `/home/<user>/projects/zig/bug/thirdparty/raylib/src/rcore.c`, but somehow adds the project path to the beginning again and now tries to search for `/home/<user>/projects/zig/bug/home/<user>/projects/zig/bug/thirdparty/raylib/src/rcore.c` which obviously doesn't exist.

### Environment

WSL2 on Windows 11

### Issue Log

```
install
└─ install zig_game
   └─ zig build-exe zig_game Debug native
      └─ zig build-lib raylib Debug native 1 errors
error: unable to check cache: stat file '/home/<user>/projects/zig/bug/home/<user>/projects/zig/bug/thirdparty/raylib/src/rcore.c' failed: FileNotFound
error: the following command failed with 1 compilation errors:
[...]

PR #3835, opened 2024-02-26
fix build paths

hardliner66:
I figured out what the problem for #3834 was and found a solution to fix it as well.

The first problem was, that `srcdir` is an absolut path, but using `.{.path = ...}` is supposed to be relative only. So I instead changed it to use `.cwd_relative` which is badly named, but the appropriate field to use for absolute paths.

The next problem was, that the files added via `addCSourceFilesVersioned` needed to be relativ to the module (raylib itself) so I figured out a way to get the relative path and added that too.

~~Theres still a minor problem with getting the relative path, but that shouldn't take that long to fix.~~

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -1,12 +1,16 @@
 const std = @import("std");
 const builtin = @import("builtin");
 
-// This has been tested to work with zig 0.11.0 and zig 0.12.0-dev.2075+f5978181e
+// This has been tested to work with zig 0.11.0, zig 0.12.0-dev.2075+f5978181e and 0.12.0-dev.2990+31763d28c
 //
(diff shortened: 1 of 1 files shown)


hardliner66:
Fixed the final problem as well. Raylib now compiles on it's own and inside my project without problesm.

raysan5:
@hardliner66 Some experienced zig developer should review this change. I feel it adds an extra level of complexity, not sure if really required.

raysan5:
@hardliner66 Thanks for the detailed explanation. I'm merging it.

hardliner66:
No problem. I analyzed the bug and fixed it, so I should also explain it. But yesterday it was already 2:30am when I created the MR, so my explanation skills weren't the best probably :sweat_smile:

PR #3837, opened 2024-02-26
Change sanitization check for `ExportDataAsCode`

lauchimoon:
I opted to use `isalnum` function since it should handle most cases. It cannot however handle cases of files beginning with numbers.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -320,6 +320,7 @@ bool ExportDataAsCode(const unsigned char *data, int dataSize, const char *fileN
     {
         // Convert variable name to uppercase
         if ((varFileName[i] >= 'a') && (varFileName[i] <= 'z')) { varFileName[i] = varFileName[i] - 32; }
+        // Replace non valid character for C identifier with '_'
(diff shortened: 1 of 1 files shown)


orcmid:
It is good to use isalnum.  Please see https://github.com/raysan5/raylib/pull/3814#issuecomment-1961645534 and subsequent comments for covering all cases and providing a utility function for this.  I still have no idea why capitalization is needed.  I think it is a misplaced application of the preprocessor identifier capitalization practice.  That should be no issue in this use-case.

src/utils.c
```diff
@@ -48,6 +48,7 @@
 #include <stdio.h>                      // Required for: FILE, fopen(), fseek(), ftell(), fread(), fwrite(), fprintf(), vprintf(), fclose()
 #include <stdarg.h>                     // Required for: va_list, va_start(), va_end()
 #include <string.h>                     // Required for: strcpy(), strcat()
+#include <ctype.h>                      // Required for: isalnum()
```
raysan5, line 51: In raylib I try to minimize standard library dependency, I prefer to avoid it.

lauchimoon: Alright, I'll find a workaround, thanks

raysan5: I can review it if you want.

PR #1717, opened 2021-04-09
Enable DRM platform cross compilation support

jonlamb-gh:
A couple minor changes I needed to enable cross compiling     
raylib DRM platform in a Yocto/BitBake environment.           
                                                              
First issue I encountered was dealing with unused result from `chdir()` in `LoadOBJ()`.
>  warning: ignoring return value of 'chdir', declared with attribute warn_unused_result [-Wunused-result]
                                                              
I just tossed in a `LOG_WARNING` if something went wrong.     
Also noticed that it was using `chdir()` directly, which seemed to conflict
with the `#define CHDIR _chdir` setup happening here:
https://github.com/raysan5/raylib/blob/d17c519f88ff33084772263f4585a96051c45cfa/src/models.c#L53-L59         
So I also converted it to use `CHDIR()` instead.              
                                                              
The next issue was related to CMake.                          
The DRM platform was using                                    
> include_directories(/usr/include/libdrm)                    
                                                              
This can be problematic when cross compiling because it doesn't consider `CMAKE_SYSROOT`.
> cc1: warning: include location "/usr/include/libdrm" is unsafe for cross-compilation [-Wpoison-system-directories]                                                  
                                                              
Now it does so conditionally based on `CMAKE_CROSSCOMPILING`.

PR #3839, opened 2024-02-27
Fix DRM cross-compile without sysroot

zuckschwerdt:
Encountering `/usr/include/xf86drm.h:40:10: fatal error: drm.h: No such file or directory`
when doing a cross-compile I found that
```
cmake -DCMAKE_C_FLAGS=-I/usr/include/libdrm …
```
is needed. Some digging revealed that in my case #1717 spoils the libdrm includes.

The unexpected and different situation here: I'm doing a cross-compile on a multi-arch image.
I.e. I run an arm-linux-gnueabihf compiler on aarch64 and the armhf libs are installed side-by-side with the arm64 libs -- no need for a sysroot.

I propose to only drop the libdrm include path if a sysroot is actually used.

### DIFF
--- a/cmake/LibraryConfigurations.cmake
+++ b/cmake/LibraryConfigurations.cmake
@@ -43,7 +43,7 @@ if (${PLATFORM} MATCHES "Desktop")
         if ("${OPENGL_LIBRARIES}" STREQUAL "")
             set(OPENGL_LIBRARIES "GL")
         endif ()
-        
+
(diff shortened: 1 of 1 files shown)


raysan5:
@zuckschwerdt Thanks for the review!

zuckschwerdt:
Kudos!
Using the cmake example project was a completely frictionless experience.
I went from a functioning sketch to feature-complete in 2 days.
Then another day to sort out Github Actions…
My project is https://github.com/zuckschwerdt/ledstrip-ddp-simulator

PR #3840, opened 2024-02-27
[rcore] reduce repetition, use memmove for queue popping

Sqvid:
`GetKeyPressed()` and `GetCharPressed()` repeat virtually all their code with just a different input (`keyPressQueue` instead of `charPressQueue`). This PR tries to reduce that repetition by factoring out the common logic. Additionally I have replaced the raw `for` loop with the built in `memmove()`. We already have `<string.h>` included so using the stdlib function comes at no additional cost.

Tested these changes on Linux 6.6.18


raysan5:
@Sqvid I think this change adds extra unneeded code complexity, I prefer current implementation despite being more repetitive.

PR #3843, opened 2024-02-27
feat: vox_loader normals and new voxels shader

fxgenstudio:
- I added the calculation of the normals for the Vox_loader.h.
- I added voxel_lighting.vs/fs shader.
- I have update models_loading_vox examples.


raysan5:
@fxgenstudio thanks for the addition!

PR #3846, opened 2024-02-29
Add example that shows off LoadRandomSequence

REDl3east:
This example shows off what can be done with the LoadRandomSequence function. The program generates a random sequence of colored rectangles. You can add a rectangle and regenerate by pressing the UP key, remove a rectangle and regenerate with the DOWN key, and to shuffle the rectangles with the SPACE key. 

I had a need to do some simple shuffling in a game I was making and was surprised there was no example already. It also made me think of a possible addition to raylib, that would be some sort of shuffle function. Shuffling is something lots of games use and would be beneficial to the library. It would look something a lot like this: `RLAPI void ShuffleRandomSequence(int *sequence, unsigned int count);`. It would already be beneficial to this example because every time I would want to shuffle I would need to call another LoadRandomSequence, which allocates more memory than it needs to.


orcmid:
There's a more direct way to shuffle the members of a list that does not requires `LoadRandomSequence()` and the creation and release of a transient array.

You may find it to be great fun for your purpose.  It is given as Algorithm P (Shuffling) in section 3.4.2 in The Art of Computer Programming, volume 2: Seminumerical Algorithms.

Conversion to C Language is relatively straightforward.

raysan5:
@REDl3east Thanks for the addition! I'll review it to minimize static functions.

PR #3849, opened 2024-02-29
[build] Fix fix-build-paths

cabarger:
#3835 Broke builds for zig versions 0.11.0 (LATEST STABLE) and  0.12.0-dev.2075+f5978181e when attempting to add support for 0.12.0-dev.2990+31763d28c. This patch supports builds for all three versions.


raysan5:
@cabarger thanks for the review! 🙂

Issue #3850, opened 2024-03-01
[build] Cannot use GLFW runtime platform detection


M374LX:
The big new feature of GLFW 3.4 is runtime platform detection, which enables a single Linux executable to run on both X11 and Wayland. However, raylib cannot use this feature at the moment when built using the Makefile or Zig, although it works when raylib is built using CMake. 

The reason is because the Makefile and Zig use the file ``rglfw.c``, which is not used by CMake, and ``rglfw.c`` combines all of GLFW's source files needed for a single platform selected at build time into one file. However, it is not possible to include multiple platforms into a single source file because GLFW has some static functions with the same name in source files for different platforms. Moreover, GLFW expects the "null" platform to be always present but we cannot include it for the same reason.

Maybe we should get rid of ``rglfw.c`` and have the Makefile and Zig select the GLFW source files to be built.

This problem has been previously reported here: https://github.com/raysan5/raylib/pull/3827#issuecomment-1963041768.

orcmid:
I am concerned that elimination of `rglfw.c` may create difficulties for desktop builds that now include it without using Make or CMake.  E.g., in `raylib/projects/scripts` and elsewhere, including simple builds of examples.

M374LX:
It looks like modifying the bundled GLFW by adding platform-specific suffixes to the functions with conflicting names is an easier solution than removing ``rglfw.c``.

raysan5:
I prefer to keep `rglfw.c`, it amalgamates all GLFW into a single module and it simplifies a lot compilation process.

In any case, how does CMake/GLFW resolve this same issue? Because build process is also supposed to compile and link all those symbols with the same name.

M374LX:
When building with CMake, instead of using ``rglfw.c``, GLFW's ``CMakeLists.txt`` is included along with raylib's own ``CMakeLists.txt``, as seen in https://github.com/raysan5/raylib/blob/master/cmake/GlfwImport.cmake.

Anyway, keeping ``rglfw.c`` and just renaming GLFW's functions with conflicting names is indeed a simpler solution.

PR #3851, opened 2024-03-01
CMake: Remove USE_WAYLAND option

M374LX:
With GLFW now having both X11 and Wayland enabled by default, the ``USE_WAYLAND`` CMake option is no longer necessary.

Setting GLFW's own CMake options ``GLFW_BUILD_X11`` and ``GLFW_BUILD_WAYLAND`` to ``OFF`` allows either to be disabled.

This PR additionally includes a check to prevent both from being disabled.


raysan5:
@M374LX Thanks for the review! It seems this change is generating some issues in the CI build system...

M374LX:
@raysan5 The CI fails with the message:
`` Failed to find wayland-scanner``

This means that the ``wayland-scanner`` utility is missing in the machine that runs the build test, but installing it should fix the problem. On Debian, for instance, the package to be installed is ``libwayland-bin``.

Peter0x44:
https://github.com/raysan5/raylib/blob/25caf14be8d0284aceaa0de28753b3012e71b7bb/.github/workflows/cmake.yml#L89

Adding libwayland-bin here would be the solution, I guess.

But now I'm questioning - doesn't the Makefile need this too?

raysan5:
@M374LX Added `libwayland-bin` to CI system for linux builds.

@Peter0x44 Afair, Makefile does not manage those deps, are expected to be installed in the system by user.

Peter0x44:
@raysan5 I was talking about the ci system for the makefile

raysan5:
@Peter0x44 Oh, ok, I think I already modified it. Just reviewed CMake build CI.

M374LX:
@Peter0x44 There is a deeper issue affecting the Makefile and the Zig build system (#3850), which I plan to review too.

PR #3852, opened 2024-03-01
Remove GLFW mouse passthrough hack and increase GLFW version in CMake

M374LX:
With the ``GLFW_MOUSE_PASSTHROUGH`` constant now available in GLFW 3.4, this hack is no longer necessary.


raysan5:

Peter0x44:
I think you should have left the comments:
```
// Defines and Macros
```
For consistency (even if they are empty)

M374LX:
@Peter0x44 If you notice, there is no "Defines and Macros" comment in files in which this part would be empty. The comment remains only in ``rcore_drm.c`` and ``rcore_web.c``.

PR #3853, opened 2024-03-01
Adding a Draw Model method that uses a transform Matrix instead of individual position , rotation and scale values

Vidrohi:
Added another Draw Model call for a personal project. This method allows the caller to pass a pre made matrix instead of specifying position, rotation etc as in DrawModelEx.

Bit easier to use in scenarios where you want to compute and cache the transform matrix outside and not have to decompose it for every draw call.


raysan5:
@Vidrohi I prefer to avoid a new method, you can do the same just assigning `model.transform` matrix.

Vidrohi:
> @Vidrohi I prefer to avoid a new method, you can do the same just assigning `model.transform` matrix.

All good :) I'll keep it to my fork.

PR #3859, opened 2024-03-05
[MODELS] Fix vertex colors for M3d files with undefined material maps

JeffM2501:
If a m3d mesh has an undefined material map, raylib still allocates a vertex color buffer for it, but does not set any values in it. This means that the colors are black with 0 alpha, so nothing renders. 

This PR fills that buffer with white so that the material texture can be drawn.

Attached is an example file that does not draw without this fix.

[untitled.zip](https://github.com/raysan5/raylib/files/14498630/untitled.zip)

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5716,7 +5716,7 @@ static Model LoadM3D(const char *fileName)
 
         if (!m3d || M3D_ERR_ISFATAL(m3d->errcode))
         {
-            TRACELOG(LOG_WARNING, "MODEL: [%s] Failed to load M3D data, error code %d", fileName, m3d? m3d->errcode : -2);
+            TRACELOG(LOG_WARNING, "MODEL: [%s] Failed to load M3D data, error code %d", fileName, m3d ? m3d->errcode : -2);
             if (m3d) m3d_free(m3d);
             UnloadFileData(fileData);
             return model;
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -5801,15 +5801,22 @@ static Model LoadM3D(const char *fileName)
...
                 // M3D specs only consider vertex colors if no material is provided, however raylib uses both and mixes the colors
-                if ((mi == M3D_UNDEF) || vcolor) model.meshes[k].colors = RL_CALLOC(model.meshes[k].vertexCount*4, sizeof(unsigned char));
+                if ((mi == M3D_UNDEF) || vcolor) model.meshes[k].colors = RL_CALLOC(model.meshes[k].vertexCount * 4, sizeof(unsigned char));
+
+                // If no map is provided and we allocated vertex colors, set them too white
+                if ((mi == M3D_UNDEF) && model.meshes[k].colors != NULL)
+                {
+                    for (int c = 0; c < model.meshes[k].vertexCount * 4; c++)
+                        model.meshes[k].colors[c] = 255;
+                }
```
JeffM2501, line 5819: this is the real fix, sorry about the other changes, my system auto formats to match the rest of the code.

raysan5: Ok, implemented!

PR #3860, opened 2024-03-07
Fix CMake SDL build for rlgl_standalone

RobLoach:
The `rlgl_standalone` example depends on having glfw. On the SDL build, glfw is not embedded directly, so we'll have to link it.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -145,5 +145,11 @@ foreach (example_source ${example_sources})
     endif ()
 endforeach ()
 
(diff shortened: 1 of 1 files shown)


raysan5:
@RobLoach @Peter0x44 thanks for the review, solution looks simple and good enough to be added.

RobLoach:
Thanks for the review, both. Initially I had disabled it, and thought we could link it directly. Perhaps disabling it is the solution that more closely matches the other platforms workaround for this. I'll submit a follow up...

Also, Raysan5, I know you don't like maintaining these CMake files. Feel free to send a ping whenever you need a review.

PR #3862, opened 2024-03-08
Implement IsCursorOnScreen for PLATFORM_DESKTOP_SDL

Peter0x44:
I did some testing and this seems to work okay, but it *is* different to GLFW. GLFW seems to lose "IsCursorOnScreen" for one frame upon pressing alt+tab over the window, without a different window covering it. SDL seems to have a more reasonable behavior for this case. Either way it's an edge case and I don't know the exact way to fix, likely requires internal GLFW changes.


raysan5:
@Peter0x44 thanks for the improvement!

PR #3863, opened 2024-03-09
Fix GLFW runtime platform detection

### LINKED ISSUE #3850: [build] Cannot use GLFW runtime platform detection

M374LX:
Fixes #3850 

In order to allow all of GLFW's source files to be combined into one, as done by ``rglfw.c``, this PR modifies the bundled GLFW by renaming the functions that share the same name in different platforms. Additionally, it disables GLFW's Null platform, which is not used by raylib.

It also removes the ``USE_WAYLAND_DISPLAY`` variable from the Makefile while adding ``GLFW_LINUX_ENABLE_WAYLAND`` and ``GLFW_LINUX_ENABLE_X11``, both set to ``TRUE`` by default.


orcmid:
It is my impression that `src/external/` materials are not maintained on raylib and any changes must come from the origin project of the external dependency.  

If there is to be an exception to that, I trust you confirmed that before all of the work you did.

M374LX:
@orcmid If GLFW's developers agree to avoid functions with the same name in different platforms in a future version of the library, there will be no more need for it to be modified for raylib (related: https://github.com/glfw/glfw/issues/2502 and https://github.com/glfw/glfw/issues/2476).

orcmid:
> @orcmid If GLFW's developers agree to avoid functions with the same name in different platforms in a future version of the library, there will be no more need for it to be modified for raylib

I think it is a policy for raylib that externals are not to be customized and then have to be maintained here as a fork.  I don't know what @raysan5 will say about this particular complicated case.

Do you see an alternative approach that would overcome the difficulty for raylib without touching `src/external`?

raysan5:
@M374LX @orcmid I try to avoid customizing external libraries for raylib but unfortunately it was also don ein the past with other external libraries. I think this is a quite special case and considering that GLFW is bundled into raylib and a related issue has already been opened in GLFW repos, I can consider merging it.

Just waiting a bit for GLFW response on the topic. The only issue I see is the need to review it again after next GLFW release, but it can take some time.

raysan5:
@M374LX thanks for the improvements!

ProIcons:
This breaks the zig build on linux.

PR #3864, opened 2024-03-10
optimisation of the "bone socket" tutorial

ipzaur:
Less matrix operations doing.


raysan5:
@ipzaur that's great! thanks for the review! 🙂

PR #3868, opened 2024-03-13
Update BINDINGS.md

joyousblunder:
Odin's binding were bumped to 5.0 back in November 2023


raysan5:
@joyousblunder Good catch! Thanks for the update!

PR #3869, opened 2024-03-15
[rcore] Fix GetKeyPressed for PLATFORM_DESKTOP_SDL

The-Night-Watch:
The key handling in `PLATFORM_DESKTOP_SDL` was faulty in two ways, which led to `GetKeyPressed` returning incorrect data.

`CORE.Input.Keyboard.keyPressedQueue` was updated only on `SDL_TEXTINPUT`, meaning only text characters were registered as a pressed key, but not function keys (eg. tab, backspace...).

Also on such event, both `CORE.Input.Keyboard.keyPressedQueue` and `CORE.Input.Keyboard.charPressedQueue` were assigned the key's corresponding codepoint, when `CORE.Input.Keyboard.keyPressedQueue` should get the raylib keycode instead.

`CORE.Input.Keyboard.keyPressedQueue` is now updated on `SDL_KEYDOWN` event instead.

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1131,7 +1131,17 @@ void PollInputEvents(void)
             case SDL_KEYDOWN:
             {
                 KeyboardKey key = ConvertScancodeToKey(event.key.keysym.scancode);
-                if (key != KEY_NULL) CORE.Input.Keyboard.currentKeyState[key] = 1;
+
+                if (key != KEY_NULL) {
+                    // If key was up, add it to the key pressed queue
+                    if ((CORE.Input.Keyboard.currentKeyState[key] == 0) && (CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE))
+                    {
+                        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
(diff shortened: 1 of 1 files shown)


RobLoach:
How would one test this? Is there an example I couldt ry it out on?

raysan5:
@RobLoach You need to build raylib for `PLATFORM_DESKTOP_SDL` but not sure if there is any example to test it properly.

@The-Night-Watch Is this change consistent with GLFW?

The-Night-Watch:
@raysan5 Yes the idea was to align to `PLATFORM_DESKTOP` behavior.

@RobLoach Simplest example would be to compare the following on `PLATFORM_DESKTOP` and `PLATFORM_DESKTOP_SDL` before and after the patch:

```
    int key = GetKeyPressed();
    int character = GetCharPressed();

    if (key) TraceLog(LOG_INFO, "key: %i", key);
    if (character) TraceLog(LOG_INFO, "char: %i", character);
    if (IsKeyPressed(KEY_BACKSPACE)) TraceLog(LOG_INFO, "backspace pressed");
}
```

Try typing function keys to see the issue.

The-Night-Watch:
Also, before patch, pressing non text keys (eg. tab, backspace...) would lead to a false negative on `IsKeyPressed(KEY_BACKSPACE)` in `PLATFORM_DESKTOP_SDL`.

raysan5:
@The-Night-Watch thanks for the improvement!

@RobLoach Thanks for reviewing it!

Issue #3841, opened 2024-02-27
[rmodels] invalid rlDisableVertexAttribute call when mesh lacks vertex color data and shader lacks vertex color input


velartrill:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

when drawing a mesh with the DrawMesh() function, a high-severity GL error is raised if the following two conditions are met:
- the shader program does not declare an `in vec4 vertexColor`, or the variable is unused in computing the final output and thus elided by the shader compiler
- the mesh lacks vertex color data

the error as reported by raylib is
```
GL: OpenGL debug message: GL_INVALID_VALUE in glDisableVertexAttribArray(index)
     > Type: ERROR
     > Source = API
     > Severity = HIGH
```

the cause of this is [rmodels.c:1474](https://github.com/raysan5/raylib/blob/a6425fa1b905f410e02def3d2f11f2d968358371/src/rmodels.c#L1474), which calls `rlDisableVertexAttribute()` regardless of whether such an attribute was ever enabled if `mesh.vboId[3] == 0` (i.e. the mesh was never given a VBO for vertex color data)

it is likely this fairly common case was missed due to a lack of a debug context during testing. i don't know enough about the GL state machine to know how likely this is to cause actual rendering problems; i encountered this as one of large number of errors that mesa reported while trying to debug a GPU crash. i am not yet certain if this bug was the culprit, but at the very least it makes debugging more difficult to have trace output clogged with this error.

i would like to stress that this error cannot be (reliably) fixed by simply adding an `in vec4 vertexColor` to the vertex shader. mesa at the very least will elide variables that cannot be used in the output. a viable workaround might be to forcibly insert an empty VBO into the mesh, bypassing the `LoadModel`/`GenMesh` API, but this is a rather discomfiting solution as it relies on the knowledge of internal magic numbers

it would appear that making this call to `rlDisableVertexAttribute()` conditional upon the presence of a vertex color input location in the `Shader` object would trivially solve the problem.

### Environment

nixos unstable (linux 6.1.75), GL 4.5, using mesa software rendering `(LIBGL_ALWAYS_SOFTWARE=1)`. raylib built from master using gcc 13

### Issue Screenshot
n/a

### Code Example

```c
[...]

PR #3871, opened 2024-03-16
[rmodels] Remove redundant call for disabling vertex color attribute

zyperpl:
This PR removes the call to `rlDisableVertexAttribute` for `SHADER_LOC_VERTEX_COLOR` in `DrawMesh` and `DrawMeshInstanced`, because it was being called in cases where this location is not set in the shader. This resulted in an invalid value (-1) being passed to `rlDisableVertexAttribute` which made an **OpenGL error `GL_INVALID_VALUE`**. 

Furthermore the same function is **already being called**  with correct checks (checking if location attribute is set and vboId[3]==0) few lines above (rmodels.c:1450 and rmodels.c:1671 in the original source).

This solves issue https://github.com/raysan5/raylib/issues/3841

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1470,9 +1470,6 @@ void DrawMesh(Mesh mesh, Material material, Matrix transform)
         if (mesh.indices != NULL) rlEnableVertexBufferElement(mesh.vboId[6]);
     }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@zyperpl thanks for the review!

PR #3874, opened 2024-03-17
[PLATFORM DESKTOP] Use relative mouse mode when cursor is disabled

JeffM2501:
When the cursor is disabled we need to put GLFW in relative mouse motion mode. If we don't some platforms (linux) can get drift in the mouse position. 

This PR sets the GLFW mode when we disable/enable the cursor.


raysan5:
@JeffM2501 Interesting issue! Thanks for the review!

PR #3881, opened 2024-03-22
[rtextures] Fix LoadImageFromScreen scaling

proberge-dev:
Before this PR, LoadImageFromScreen was not applying the scale DPI to both width and height. After this PR, the behaviour of LoadImageFromScreen is the same as TakeScreenshot, but returns an image instead. Not sure if this is the appropriate fix, but this was what I was expecting when calling the function. If there needs to be adjustments such as calling ImageResize to get the current width and height instead of the scaled one, let me know.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -681,10 +681,11 @@ Image LoadImageFromTexture(Texture2D texture)
 // Load image from screen buffer and (screenshot)
 Image LoadImageFromScreen(void)
 {
+    Vector2 scale = GetWindowScaleDPI();
     Image image = { 0 };
 
-    image.width = GetScreenWidth();
-    image.height = GetScreenHeight();
+    image.width = GetScreenWidth()*scale.x;
+    image.height = GetScreenHeight()*scale.y;
     image.mipmaps = 1;
     image.format = PIXELFORMAT_UNCOMPRESSED_R8G8B8A8;
     image.data = rlReadScreenPixels(image.width, image.height);


raysan5:
@proberge-dev did you try it? does it work as expected on a HiDPI monitor?

proberge-dev:
Yes, I've tried it on a Macbook Pro. It works as far as I can see. This code is functionally equivalent to the code in TakeScreenshot from rcore.c:

```c
void TakeScreenshot(const char *fileName)
{
#if defined(SUPPORT_MODULE_RTEXTURES)
    // Security check to (partially) avoid malicious code
    if (strchr(fileName, '\'') != NULL) { TRACELOG(LOG_WARNING, "SYSTEM: Provided fileName could be potentially malicious, avoid [\'] character"); return; }

    Vector2 scale = GetWindowScaleDPI();
    unsigned char *imgData = rlReadScreenPixels((int)((float)CORE.Window.render.width*scale.x), (int)((float)CORE.Window.render.height*scale.y));
    Image image = { imgData, (int)((float)CORE.Window.render.width*scale.x), (int)((float)CORE.Window.render.height*scale.y), 1, PIXELFORMAT_UNCOMPRESSED_R8G8B8A8 };

    char path[512] = { 0 };
    strcpy(path, TextFormat("%s/%s", CORE.Storage.basePath, GetFileName(fileName)));
    
    ExportImage(image, path);           // WARNING: Module required: rtextures
    RL_FREE(imgData);

    if (FileExists(path)) TRACELOG(LOG_INFO, "SYSTEM: [%s] Screenshot taken successfully", path);
    else TRACELOG(LOG_WARNING, "SYSTEM: [%s] Screenshot could not be saved", path);
#else
    TRACELOG(LOG_WARNING,"IMAGE: ExportImage() requires module: rtextures");
#endif
}
```

raysan5:
@proberge-dev ok, thanks for the review

PR #3886, opened 2024-03-28
Enable include reordering

marcogmaia:
The purpose of this is for us to be able to do the following:

```c
#include <raymath.h>
#include <rlgl.h>
#include <raylib.h>
```

Right now the only possibility is to include `raylib.h` first.

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -91,48 +91,6 @@
       "description": ""
     },
     {
-      "name": "RL_COLOR_TYPE",
-      "type": "GUARD",
-      "value": "",
-      "description": ""
-    },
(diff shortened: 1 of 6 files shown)


marcogmaia:
@raysan5 I have one question:

Why don't we extract this repeating redefinition to a single source of truth? Something like `internal/common_types.h`?

raysan5:
@marcogmaia Sorry but this change adds extra complexity to the header that I prefer to avoid.

  > Why don't we extract this repeating redefinition to a single source of truth? Something like internal/common_types.h?

I prefer to keep the `raylib.h` header self-contained and not-dependant on other headers.

Issue #3789, opened 2024-02-07
[rcore] DRM Platform and Gamepad Button 0


SirNate0:
### Issue description

These lines use the index of the button from the js device without any mapping. However, it also uses `GAMEPAD_BUTTON_UNKNOWN=0` as the `lastButtonPressed` when the button is released, but button=0 is actually a valid button and not a non-existent button on this platform.

https://github.com/raysan5/raylib/blob/8ba7c61625c1a4eb688fcd95bb95a9e9aea53b1c/src/platforms/rcore_drm.c#L1975-L1981

I see a few ways this could be corrected (unfortunately both would require changes to user's code):

1. Simply increment the index by 1. Button 0 will still be a not-pressable button, like on rcore_desktop.
2. Implement proper button mapping. This would involve an ioctl to get the button map for the joystick. An example of such code can be seen in the jstest code.
3. Switch to libevdev for the gamepad input. This would force button mapping like with the platform_desktop code, as the events would not be condensed to a small array of buttons.

MrMugame:
Yeah so ... ahm ... the drm backend got updated and this is something I definitely overlooked. So all input handling got switched to the evdev api, but I forgot to map the buttons and because the linux event api gampad buttons start somewhere at `0x130` which is much larger than `MAX_GAMEPAD_BUTTONS = 32`, all buttons get irgnored currently.... Going to fix that with a map. Give me a sec.

raysan5:
@SirNate0 Issue should be fixed, please, let me know otherwise.

PR #3888, opened 2024-03-28
Fixing gamepad buttons not working in drm backend

MrMugame:
Fixed the gamepad button problem as described in #3789 and also changed the way axes are enumerated and removed some trailing spaces.


raysan5:
@MrMugame thanks for the quick review!

PR #3891, opened 2024-03-29
Fixes zig build that was broken on #3863

ProIcons:
Not even sure if this should be approached this way, however it's the first step to be fixed :)


ProIcons:
Actually it will not be enough, 
things like: "raylib.linkSystemLibrary("X11");" should be moved only on the X11 case and for wayland the other system libraries should be added.
However i'm not sure which ones there are for Wayland.

raysan5:
@ProIcons Thank you very much for the fixes!

PR #3892, opened 2024-04-01
check for vcvarsall.bat normally before guessing pre-defined paths

derpy54320:
Some people (like me) may have the path to their *vcvarsall.bat* in their system path. This will check for it and just use it directly if available before using the pre-defined directories.

### DIFF
--- a/projects/scripts/build-windows.bat
+++ b/projects/scripts/build-windows.bat
@@ -27,7 +27,11 @@ REM (Prevents calling vcvarsall every time you run this script)
 WHERE cl >nul 2>nul
 IF %ERRORLEVEL% == 0 goto READ_ARGS
 REM Activate the msvc build environment if cl isn't available yet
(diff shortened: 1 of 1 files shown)


orcmid:
I don't think this is an useful approach, and it doesn't always work.

Since Microsoft stopped putting VS information in environment variables, there is a different recommended scheme.

Use one of the Developer Command Prompts that is provided as part of the installation of Visual Studio or Visual Studio Build Tools.   This will cause the search path and all of the environment variables to be set as part of the opened Command Prompt.  `vcvarsall` will also have been run. 
 This will set some environment variables such that a batch script can tell that the build environment is established.   This works no matter where Visual Studio is installed and does not depend on knowing the path to anything, the version being run, etc.

I made an alternative version of `build-windows.bat` that does that and more as `VCrayApp.bat`.  You can find the 0.1.0 beta release (working but unannounced until documentation completed) at
[nfoTools VCrayApp](https://orcmid.github.io/nfoTools/dev/D211101/).

raysan5:
@derpy54320 thanks for the improvement but I agree with @orcmid on this change.

PR #3893, opened 2024-04-01
XML escape descriptions in parser

petelliott:
the current parser is generating invalid xml with attributes containing "<" and ">". this change will escape {><&} in descriptions when outputting XML


raysan5:
@petelliott thanks for the review but I prefer to address this issue on `raylib.h` than increasing parser complexity considering this kind of side cases.

PR #3894, opened 2024-04-02
[rcore] Fix framerate for gif recording

RobLoach:
The framerate for gif recording was triggered every 10 frames. While this is fine, it often resulted in the gif not matching the actual framerate of the application. Frames can often take more or less time depending on rendering, resulting in the gif stuttering and having incorrect speeds. This change updates the calculation of how long the frames should be shown depending on the actual time it took to render the frames. The result is gifs that actually match the application speed.

As an example, see `shaders_custom_uniform` from the [examples](https://www.raylib.com/examples.html)...

### Old

This gif speed is moving about twice as fast as the application actually moves.
old

 ### New

The gif recording now is rendered at the correct speed.
new

### `GIF_RECORD_FRAMERATE`
In addition, this now allows you to change `GIF_RECORD_FRAMERATE`. Changing it to `20`, for instance, will give you a much smoother experience:

screenrec001

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -849,23 +849,33 @@ void EndDrawing(void)
     // Draw record indicator
     if (gifRecording)
     {
+        #ifndef GIF_RECORD_FRAMERATE
         #define GIF_RECORD_FRAMERATE    10
-        gifFrameCounter++;
+        #endif
+        gifFrameCounter += GetFrameTime()*1000;
 
-        // NOTE: We record one gif frame every 10 game frames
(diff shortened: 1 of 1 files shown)


raysan5:
@RobLoach Oh! This is a really nice improvement! Thank you very much! 😄 

It's a sahme that not many users are aware of the GIF recording functionality in raylib, afaik...

RobLoach:
Certainly an underrated feature! We have documentation around for it somewhere, right?

RobLoach:
Hopefully this helps people know it's a thing they can do https://github.com/raysan5/raylib/wiki/How-To:-Screen-Recording

PR #3896, opened 2024-04-04
Fix #3891 breaking builds for Zig v0.11.0

### LINKED ISSUE #3895: [Zig] Zig build fails to compile

 - [x ] I tested it on latest raylib version from master branch
 - [ x] I checked there is no similar issue already reported
 - [ x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [ x] My code has no errors or misuse of raylib

### Issue description
I imported raylib in a zig project, following the guide of this tutorial of the reddit of raylib: https://www.reddit.com/r/raylib/comments/17y0gdx/get_raylib_up_and_running_in_just_4_minutes_using/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

### Environment

Operating System: Arch Linux
OpenGL: Intel TigerLake-LP GT2 [Iris Xe Graphics] 

### Issue Screenshot

image

### Code Example
clone raylib 
execute:
`zig init-exe`

edit build.zig:
```
const std = @import("std");
const raylibSDK = @import("./raylib/src/build.zig");

// Although this function looks imperative, note that its job is to
// declaratively construct a build graph that will be executed by an external
// runner.
pub fn build(b: *std.Build) void {
    // Standard target options allows the person running `zig build` to choose
    // what target to build for. Here we do not override the defaults, which
    // means any target is allowed, and the default is native. Other options
    // for restricting supported target set are available.

    // Standard optimization options allow the person running `zig build` to select
    // between Debug, ReleaseSafe, ReleaseFast, and ReleaseSmall. Here we do not
    // set a preferred release mode, allowing the user to decide how to optimize.

    const raylib = raylibSDK.addRaylib(b, target, optimize, .{});

        .name = "wolfenstein_3d_zig",
        // In this case the main source file is merely a path, however, in more
        // complicated build scripts, this could be a generated file.
        .root_source_file = .{ .path = "src/main.zig" },
        .target = target,
        .optimize = optimize,
    });

    exe.addIncludePath(.{ .path = "./raylib/src/" });
    exe.linkLibrary(raylib);
    // This declares intent for the executable to be installed into the
    // standard location when the user invokes the "install" step (the default
    // step when running `zig build`).
    b.installArtifact(exe);

    // This *creates* a Run step in the build graph, to be executed when another
    // step is evaluated that depends on it. The next line below will establish
    // such a dependency.
    const run_cmd = b.addRunArtifact(exe);

    // By making the run step depend on the install step, it will be run from the
    // installation directory rather than directly from within the cache directory.
    // This is not necessary, however, if the application depends on other installed
    // files, this ensures they will be present and in the expected location.
    run_cmd.step.dependOn(b.getInstallStep());

    // This allows the user to pass arguments to the application in the build
    // command itself, like this: `zig build run -- arg1 arg2 etc`
    if (b.args) |args| {
        run_cmd.addArgs(args);
    }

    // This creates a build step. It will be visible in the `zig build --help` menu,
    // and can be selected like this: `zig build run`
    // This will evaluate the `run` step rather than the default, which is "install".
    const run_step = b.step("run", "Run the app");
    run_step.dependOn(&run_cmd.step);

    // Creates a step for unit testing. This only builds the test executable
    // but does not run it.
    const unit_tests = b.addTest(.{
        .root_source_file = .{ .path = "src/main.zig" },
        .target = target,
        .optimize = optimize,
    });

    const run_unit_tests = b.addRunArtifact(unit_tests);

    // Similar to creating the run step earlier, this exposes a `test` step to
    // the `zig build --help` menu, providing a way for the user to request
    // running the unit tests.
    const test_step = b.step("test", "Run unit tests");
    test_step.dependOn(&run_unit_tests.step);
}
```
build the zig code:

```
zig build run
/media/gus/proyectos/wolfenstein_3d_zig/build.zig:31:21: error: expected type '*Build.Step.C
ompile', found '@typeInfo(@typeInfo(@TypeOf(raylib.src.build.addRaylib__anon_13349)).Fn.retu
rn_type.?).ErrorUnion.error_set!*Build.Step.Compile'
    exe.linkLibrary(raylib);
                    ^~~~~~
/media/gus/proyectos/wolfenstein_3d_zig/build.zig:31:21: note: cannot convert error union to
 payload type
/media/gus/proyectos/wolfenstein_3d_zig/build.zig:31:21: note: consider using 'try', 'catch'
, or 'if'
/usr/lib/zig/std/Build/Step/Compile.zig:672:41: note: parameter type declared here
pub fn linkLibrary(self: *Compile, lib: *Compile) void {
                                        ^~~~~~~~
referenced by:
    runBuild__anon_7136: /usr/lib/zig/std/Build.zig:1638:27
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
/media/gus/proyectos/wolfenstein_3d_zig/raylib/src/build.zig:309:30: error: struct 'child_pr
ocess.ChildProcess' has no member named 'run'
    _ = try std.process.Child.run(.{
            ~~~~~~~~~~~~~~~~~^~~~
/usr/lib/zig/std/child_process.zig:20:26: note: struct declared here
pub const ChildProcess = struct {
                         ^~~~~~

```

iarkn:
The changes brought by #3891 uses `std.process.Child.run` which was renamed from `std.process.Child.exec` in Zig version 0.11.0, resulting in an error when trying to build raylib with Zig version 0.11.0. This commit adds a version check to use the appropriate function names.

Additionally, the `linux_display_backend` configuration option is added so users can set it when running `zig build` with the `-D` option or when using raylib as a dependency.

Tested with Zig version `0.11.0` and version `0.12.0-dev.3522+b88ae8dbd`.

Resolves #3895.

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -254,6 +254,7 @@ pub fn build(b: *std.Build) !void {
         .rshapes = b.option(bool, "rshapes", "Compile with shapes support") orelse defaults.rshapes,
         .raygui = b.option(bool, "raygui", "Compile with raygui support") orelse defaults.raygui,
         .shared = b.option(bool, "shared", "Compile as shared library") orelse defaults.shared,
+        .linux_display_backend = b.option(LinuxDisplayBackend, "linux_display_backend", "Linux display backend to use") orelse defaults.linux_display_backend,
     };
 
     const lib = try addRaylib(b, target, optimize, options);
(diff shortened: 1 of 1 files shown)


raysan5:
@iarkn thanks for the review!

ProIcons:
Thank @iarkn completely missed that on my pr!

PR #3898, opened 2024-04-05
Added `GetScreenSize` function

LiamM32:
I added the `GetScreenSize` function to return the screen width and screen height as a `Vector2`. That's all.


raysan5:
@LiamM32 Thanks but I prefer to avoid that function, I think `GetScreenWidth()` and `GetScreenHeight()` already cover most user cases.

LiamM32:
Oh well. I should mention that I also added the update to `BINDINGS.md`. I decided to add it after noticing.

PR #3900, opened 2024-04-06
Remove redundant axis length calculation

jtainer:


raysan5:
@jtainer thanks for the review!

PR #3901, opened 2024-04-07
Update Makefile RAYLIB_VERSION

Belllg:
Changed the raylib version from 4.2.0 to 5.1-dev, in the makefile from the VScode project to be just like the raylib.h file


raysan5:
@Belllg thanks for the review!

PR #3902, opened 2024-04-07
Update raylib-zig version

Not-Nik:
Been that way for a while, but I don't think I told anyone.


raysan5:
@Not-Nik thanks for the update! 😄

PR #3905, opened 2024-04-09
Update to latest zig and simplify build.zig

freakmangd:
+ `installHeader`'s signature was changed recently, so a versioned function was added for it
+ `srcdir` isn't needed, just adding "src/" before each path is sufficient
+ cabarger's hack seems to not be needed anymore
+ `process.Child.run` and `process.Child.exec` was coerced into a single function

Tested on ubuntu x64, on both 0.12.0-dev.3580+e204a6edb and 0.11.0

PR #3913, opened 2024-04-11
[build.zig] Fix local dependency breakage

freakmangd:
Fix local dependency breakage caused by #3905

Tested on ubuntu x64 with zig 0.11.0 and zig 0.12.0-dev.3632+7fb5a0b18


raysan5:
@freakmangd thanks for the review.

PR #3914, opened 2024-04-12
Changed RLGL_VERSION from "4.5" to "5.0

Mute124:
Line 110 of file rlgl.h is the one in question.


RobLoach:
Makes sense.

raysan5:
@Mute124 @RobLoach Actually waiting for some changes to be merged but it's ok to update version number.

PR #3917, opened 2024-04-13
Reset music.ctxType if loading wasn't succesful

veins1:
Fixes some crashes, e.g. calling StopMusicStream after trying to load Music from a non-existant .mp3 file


raysan5:

PR #3919, opened 2024-04-15
Added missing interpolation types for GLTF animation channels

benjitrosch:
I noticed that the GLTF animation loading only supported linear interpolation, so by adding support for `cgltf_interpolation_type_step` and `cgltf_interpolation_type_cubic_spline` this should cover most cases.

This was tested using `robot.gltf` by importing it into Blender, removing some keyframes, using constant and bezier interpolation, and then rerunning the `models_loading_gltf.exe` example. Step is easy to recognize, and I think the spline looks right as well, but I would love someone with discerning eyes to tell me if it looks accurate.

Here's videos of the exact same animation timeline with different interpolation applied:

- Step interpolation example:

https://github.com/raysan5/raylib/assets/64439681/f4e1d9b9-815b-412c-b62f-d503aea90955

- Bezier interpolation example:

https://github.com/raysan5/raylib/assets/64439681/816d3963-3857-4880-87b8-23068205f640

- Linear example (to compare with the others):

https://github.com/raysan5/raylib/assets/64439681/f154f66a-3a79-4a3e-a296-1048f60ad862

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -955,6 +955,22 @@ RMAPI Vector3 Vector3Lerp(Vector3 v1, Vector3 v2, float amount)
     return result;
 }
 
+// Calculate cubic hermite interpolation between two vectors and their tangents
+// taken directly from: https://en.wikipedia.org/wiki/Cubic_Hermite_spline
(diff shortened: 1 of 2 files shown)


raysan5:
@benjitrosch Nice improvement! Thank you very much!

PR #3920, opened 2024-04-17
Replaced SQUAD interpolation with Cubic Hermite to align with GLTF 2.0 spec

benjitrosch:
This is a follow up PR on #3919.

During that PR, I went researching for a quaternion cubic spline algorithm that could be used, and landed on SQUAD. Turns out, the GLTF 2.0 very specifically [provides an algorithm](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#interpolation-cubic) that is says to use (which looks to me like Cubic Hermite Spline):

```
Let ak, vk, and bk be the in-tangent, the property value, and the out-tangent of the k-th frame respectively.

The interpolated sampler value vt at the timestamp tc is computed as follows.

vt=(2t3−3t2+1)∗vk+td(t3−2t2+t)∗bk+(−2t3+3t2)∗vk+1+td(t3−t2)∗ak+1
```

However, Cubic Hermite Spline showed flipping/spinning when I was doing tests due to quaternion double-cover. To alleviate this, I added some logic to check the dot product and negate them (and also normalized them).

After this change I'm happy to say it looks pretty good and should be technically more "correct" and compliant with what GLTF 2.0 expects!

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -956,7 +956,7 @@ RMAPI Vector3 Vector3Lerp(Vector3 v1, Vector3 v2, float amount)
 }
 
 // Calculate cubic hermite interpolation between two vectors and their tangents
-// taken directly from: https://en.wikipedia.org/wiki/Cubic_Hermite_spline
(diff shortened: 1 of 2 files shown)


raysan5:
@benjitrosch Nice! thanks for the review!

PR #3921, opened 2024-04-17
Build specific example using -DBUILD_EXAMPLE Cmake flag

benjitrosch:
As the title says, this allows users to build only one example rather than all of them using CMake by passing a BUILD_EXAMPLE flag:

```cmake
cmake -B build -DBUILD_EXAMPLE=models_loading_gltf
```

Making a change to the core Raylib library will cause all of the examples to rebuild which can take a long time. This would be a helpful timesaver to contributors who use CMake through the terminal, or with an IDE like VSCode or Neovim etc. rather than using the Visual Studio solution.


Peter0x44:
I can't help but feel there has to be a better way to implement this in cmake

benjitrosch:
@Peter0x44 what did you have in mind?

Peter0x44:
This commit is not needed, and probably should be reverted.
This already works likeso:
```
$ cmake --build . -t core_basic_window
[ 85%] Built target glfw
[ 85%] Building C object raylib/CMakeFiles/raylib.dir/rtextures.c.o
[ 85%] Linking C static library libraylib.a
[100%] Built target raylib
[100%] Linking C executable core_basic_window
[100%] Built target core_basic_window
```
You can pick a specific target and it will rebuild only that target and its dependencies.
This option is not necessary or useful.

raysan5:
@Peter0x44 Oh! Thanks for investigating this improvement!

PR #3922, opened 2024-04-19
Revert "Build specific example using -DBUILD_EXAMPLE Cmake flag"

benjitrosch:
Thanks to @Peter0x44 seems this is possible with the following steps:

```cmake
cmake
cmake --build . -t core_basic_window
```


Peter0x44:
LGTM.

PR #3923, opened 2024-04-19
Fix window not initializing on primary monitor on GLFW backend

eldskald:
The way the current code worked was by calling `GetCurrentMonitor()`, which would always return the monitor at position (0,0). This isn't the primary monitor on all platforms, on Linux in particular it isn't the case.

This isn't the case on the SDL backend, after calling `InitWindow()` the window would always show up on the primary monitor. Even on the GLFW backend, if the full screen flag was set it would attempt to put it on the primary monitor as it would call `glfwGetPrimaryMonitor()` to do it, so for consistency's sake we should do it on windowed mode too.

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -1518,10 +1518,16 @@ int InitPlatform(void)
     else 
     {
         // Try to center window on screen but avoiding window-bar outside of screen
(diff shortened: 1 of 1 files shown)


raysan5:
@eldskald thanks for the improvement and the formatting review! 🙂

PR #3935, opened 2024-04-24
[rlgl] Fix incorrect matrix multiplication order in `rlMultMatrixf`

bohonghuang:
Hello Ray! When playing with Raylib's matrix stack, I've noticed that the behavior of `rlMultMatrixf` might be inconsistent with that of `glMultMatrixf`. Let's first take a look at Khronos' explanation for `glTranslatef`:

> glTranslate produces a translation by x y z. The current matrix (see glMatrixMode) is multiplied by this translation matrix, with the product replacing the current matrix, as if glMultMatrix were called with the following matrix for its argument:
>
> 1 0 0 x
> 0 1 0 y
> 0 0 1 z
> 0 0 0 1

This means `glTranslate` can be achieved by calling `glMultMatrix` using a translation matrix. This implies that in Raylib, the following function can be used to replace `rlTranslatef`:

```c
void rlTranslatef_(float x, float y, float z) {
  rlMultMatrixf(MatrixToFloat(MatrixTranslate(x, y, z)));
}
```

Now, let's compare the behavior of the built-in `rlTranslatef` in RLGL (which can serve as a correct reference due to its extensive use) with the version implemented above using `rlMultMatrixf`:

```c
#include "raylib.h"
#include "rlgl.h"
#include "raymath.h"

void rlTranslatef_(float x, float y, float z) {
  rlMultMatrixf(MatrixToFloat(MatrixTranslate(x, y, z)));
}

int main(void)
{
  const int screenWidth = 800;
  const int screenHeight = 450;

  InitWindow(screenWidth, screenHeight, "rlMultMatrixf Test");
  
  Camera camera = {
    .position = (Vector3){ 10.0f, 10.0f, 10.0f },
    .target = (Vector3){ 0.0f, 0.0f, 0.0f },
    .up = (Vector3){ 0.0f, 1.0f, 0.0f },
    .fovy = 45.0f,
    .projection = CAMERA_PERSPECTIVE,
  };

  SetTargetFPS(60);

  while (!WindowShouldClose())
  {
    BeginDrawing();

      ClearBackground(RAYWHITE);

      BeginMode3D(camera);

        DrawGrid(100, 1.0);

        rlPushMatrix();

        rlLoadIdentity();

          rlScalef(2.0f, 2.0f, 2.0f);

          rlTranslatef(2.0f, 0.0f, 0.0f);

          DrawCube(Vector3Zero(), 0.5f, 0.5f, 0.5f, Fade(GREEN, 0.5f));

        rlLoadIdentity();

          rlScalef(2.0f, 2.0f, 2.0f);

          rlTranslatef_(2.0f, 0.0f, 0.0f);

          DrawCube(Vector3Zero(), 0.5f, 0.5f, 0.5f, Fade(RED, 0.5f));

        rlPopMatrix();

      EndMode3D();

    EndDrawing();
  }
  
  CloseWindow();

  return 0;
}
```

As shown in the result, the green cube indicates correct behavior (produced by `rlTranslatef`), meaning the cube along with the coordinates were scaled after translation. However, the red one indicates incorrect behavior (produced by `rlMultMatrixf`) where only the cube itself was scaled without any change to the coordinates. Clearly, there's an issue with the multiplication order, namely:

```c
rlScalef(2.0f, 2.0f, 2.0f);
rlTranslatef_(2.0f, 0.0f, 0.0f);
```

Which was actually supposed to be:

```c
rlTranslatef(2.0f, 0.0f, 0.0f);
rlScalef(2.0f, 2.0f, 2.0f);
```

Mathematically speaking, in the existing Raylib code, `rlMultMatrixf` is implemented to perform a right multiplication on the current matrix (following Raylib's matrix multiplication convention, which is actually left multiplication mathematically):

```c
*RLGL.State.currentMatrix = rlMatrixMultiply(*RLGL.State.currentMatrix, mat);
```

While functions like `rlTranslatef` are implemented as:

```c
*RLGL.State.currentMatrix = rlMatrixMultiply(matTranslation, *RLGL.State.currentMatrix);
```

Mathematically, left multiplying a matrix cannot usually be equivalently done by right multiplying another matrix.

Here is the result of this PR fixing the code above:

It's worth mentioning that this oversight, along with its correction, does not affect Raylib's existing other functionalities at all. This is because I have checked that almost every call to `rlMultMatrixf` within Raylib is preceded by a `rlLoadIdentity`, meaning both left and right multiplying the identity matrix equals the original matrix.


bohonghuang:
> @bohonghuang thank you very much for the review and the detailed report and excuse me for taking so long to review it! Thanks!

That's fine. Glad I could help.

PR #3941, opened 2024-04-28
make RGFW a custom platform

ColleagueRiley:
This adds the custom platform for RGFW.

~~it also removes GLAD.h and loads opengl 3.3 directly in RLGL.h~~


src/rcore.c
```diff
@@ -992,18 +994,18 @@ void BeginMode3D(Camera camera)
     if (camera.projection == CAMERA_PERSPECTIVE)
     {
         // Setup perspective projection
-        double top = RL_CULL_DISTANCE_NEAR*tan(camera.fovy*0.5*DEG2RAD);
+        double top = rlGetCullDistanceNear()*tan(camera.fovy*0.5*DEG2RAD);
```
raysan5, line 997: Those changes are already in the raylib master branch, please, could you remove them from the list?

src/Makefile
```diff
@@ -694,6 +706,11 @@ rshapes.o : rshapes.c raylib.h rlgl.h
 rtextures.o : rtextures.c raylib.h rlgl.h utils.h
 	$(CC) -c $< $(CFLAGS) $(INCLUDE_PATHS)
 
+RGFW.o:
```
raysan5, line 709: Is it possible to avoid compiling this unit? One of the benefits of RGFW would be compiling it directly into `rcore_desktop_rgfw.c`.

examples/models/models_skybox.c
```diff
@@ -208,7 +208,7 @@ static TextureCubemap GenTextureCubemap(Shader shader, Texture2D panorama, int s
     rlEnableShader(shader.id);
 
     // Define projection matrix and send it to shader
-    Matrix matFboProjection = MatrixPerspective(90.0*DEG2RAD, 1.0, RL_CULL_DISTANCE_NEAR, RL_CULL_DISTANCE_FAR);
+    Matrix matFboProjection = MatrixPerspective(90.0*DEG2RAD, 1.0, rlGetCullDistanceNear(), rlGetCullDistanceFar());
```
raysan5, line 211: This change shouldn't be in this changelist...

examples/Makefile
```diff
@@ -416,6 +416,31 @@ ifeq ($(PLATFORM),PLATFORM_DESKTOP_SDL)
         LDLIBS += -latomic
     endif
 endif
+ifeq ($(PLATFORM),PLATFORM_DESKTOP_RGFW)
+    ifeq ($(PLATFORM_OS),WINDOWS)
+        # Libraries for Windows desktop compilation
+        LDLIBS = -lopengl32 ..\src\libraylib.a -lgdi32 -lgdi32 -lXinput -lopengl32 -lShcore
```
raysan5, line 422: Why are `-lXinput` and `-lShcore` required? Also, `-lopengl32` is linked twice while it does not seem to be required with GLFW.

src/rmodels.c
```diff
@@ -5137,18 +5129,18 @@ static Model LoadGLTF(const char *fileName)
                             if (attribute->component_type == cgltf_component_type_r_32f)  // vec2, float
                             {
                                 // Init raylib mesh texcoords to copy glTF attribute data
-                                texcoordPtr = RL_MALLOC(attribute->count*2*sizeof(float));
+                                texcoordPtr = (float *)RL_MALLOC(attribute->count*2*sizeof(float));
```
raysan5, line 5132: Those changes are already on raylib master branch, they shouldn't be in this changelist.

raysan5:
Added some notes, it seems some changes from raylib master branch get into the changelist. I also note `glad` removal implies a considerable review of `rlgl`, I think it should be on a separate PR for proper review.

examples/Makefile
ColleagueRiley: Xinput is used for controller input and I believe Shcore is used for monitor functions.

The main reason why GLFW doesn't need to link with -lopengl32 twice is because it loads WGL directly into itself. I do not like that design.

src/Makefile
ColleagueRiley: I don't know if that's possible due to naming conflicts with winapi.

ColleagueRiley:
I replied to two of your comments and I agree that the RLGL part should be in its own PR.

src/Makefile
```diff
@@ -694,6 +706,11 @@ rshapes.o : rshapes.c raylib.h rlgl.h
 rtextures.o : rtextures.c raylib.h rlgl.h utils.h
 	$(CC) -c $< $(CFLAGS) $(INCLUDE_PATHS)
 
+RGFW.o: 
+	cp external/RGFW.h RGFW.c
```
raysan5, line 710: Does `cp` work on Windows? Also MSVC needs to be considered, a Visual Studio Project is included with raylib.

src/rcore.c
```diff
@@ -3090,8 +3092,8 @@ void InitTimer(void)
     // However, it can also reduce overall system performance, because the thread scheduler switches tasks more often
     // High resolutions can also prevent the CPU power management system from entering power-saving modes
     // Setting a higher resolution does not improve the accuracy of the high-resolution performance counter
-#if defined(_WIN32) && defined(SUPPORT_WINMM_HIGHRES_TIMER) && !defined(SUPPORT_BUSY_WAIT_LOOP) && !defined(PLATFORM_DESKTOP_SDL)
-    timeBeginPeriod(1);                 // Setup high-resolution timer to 1ms (granularity of 1-2 ms)
+#if defined(_WIN32) && defined(SUPPORT_WINMM_HIGHRES_TIMER) && !defined(SUPPORT_BUSY_WAIT_LOOP) && !defined(PLATFORM_DESKTOP_SDL) && !defined(PLATFORM_DESKTOP_RGFW)
+   timeBeginPeriod(1);                 // Setup high-resolution timer to 1ms (granularity of 1-2 ms)
```
raysan5, line 3096: Just a detail: it should be 4 spaces.

src/rcore.c
```diff
@@ -3614,4 +3616,4 @@ const char *TextFormat(const char *text, ...)
     return currentBuffer;
 }
 
-#endif // !SUPPORT_MODULE_RTEXT
+#endif // !SUPPORT_MODULE_RTEXT
```
raysan5, line 3619: I should end with a line-break, like all the other files.

src/Makefile
ColleagueRiley: I believe cp should work with make because make is a gnu tool. 

Although I'll double check that later.

ColleagueRiley: > Does `cp` work on Windows? Also MSVC needs to be considered, a Visual Studio Project is included with raylib.

Does the makefile even support msvc?

orcmid: A number of things are being inter-confused here.  First, [make](https://en.wikipedia.org/wiki/Make_(software)) originated in 1976 on Unix and various enhancements and variations have been made since.  

The Visual Studio Build Tools and Visual Studio supply the tool [nmake](https://learn.microsoft.com/en-us/cpp/build/reference/nmake-reference?view=msvc-170), which also preceded Linux (but maybe not GNU).

It is the case that `nmake` accepts makefiles, although there might need to be differences in some setting macros and accommodation of differences with respect to other `make` forks.

Whether the makefile you are speaking of is specific to use of gcc and has any GNU/Linux-tool dependencies, like the Unix `cp` command, it all depends.

ColleagueRiley: Thank you for explaining

ColleagueRiley:
All the comments have been resolved. (I'm not sure if you've seen this yet)

raysan5:
> Compile RGFW directly with Rcore_desktop_Rgfw.h via RGFW_IMPLEMENTATION

@ColleagueRiley Wow! Does it really work??? I didn't know collisions could be avoided that way!

ColleagueRiley:
Yep, Rob posted about it here! https://github.com/raysan5/raylib/discussions/3945#discussioncomment-9277152

raysan5:
@ColleagueRiley I'm trying to compile `core_basic_window` example with VS202 on `PLATFORM_DESKTOP_RGFW` and I get this issue:
image

raysan5:
When I try to compile with `w64devkit (GCC 13.2.0)`, I got some warnings but `libraylib.a` is correctly generated:
image

but when I try to compile a `core_basic_window` example I got this:

image

ColleagueRiley:
GLAD should work with RGFW now!

raysan5:
@ColleagueRiley Testing it again and I'm still getting this issue on `x86` configuration with MSVC

ColleagueRiley:
> @ColleagueRiley Testing it again and I'm still getting this issue on `x86` configuration with MSVC
> 
> image
> 
> 

Ensure you're using the newest pull.

raysan5:
@ColleagueRiley Updated to latest PR version and now I'm getting this issue:
image

ColleagueRiley:
> @ColleagueRiley Updated to latest PR version and now I'm getting this issue:
> 
> image
> 
> 

This is because Raylib needs to link with winmm. You should already be linking with winmm I think.

raysan5:
> This is because Raylib needs to link with winmm. You should already be linking with winmm I think.

Yes, I am, just reviewed it but issue is similar to previous one, for some reason the symbols are not detected despite linking with the appropiate libraries...

In any case, is `timeBeginPeriod` really required for `RGFW`? I see that `PLATFORM_DESKTOP_SDL` just skips that line...

ColleagueRiley:
> In any case, is `timeBeginPeriod` really required for `RGFW`? I see that `PLATFORM_DESKTOP_SDL` just skips that line...

I'm not sure why SDL skips that line. Both RGFW and GLFW use that line. For some reason without that line the timing is off on windows. I'm not sure why this didn't happen to you before, but this is not an issue with this PR.

ColleagueRiley:
> I'm not sure why SDL skips that line. Both RGFW and GLFW use that line. For some reason without that line the timing is off on windows. I'm not sure why this didn't happen to you before, but this is not an issue with this PR.

Also, I don't think I got that issue in my tests so I could be an issue with your setup.

ColleagueRiley:
> > This is because Raylib needs to link with winmm. You should already be linking with winmm I think.
> 
> Yes, I am, just reviewed it but issue is similar to previous one, for some reason the symbols are not detected despite linking with the appropiate libraries...
> 
> In any case, is `timeBeginPeriod` really required for `RGFW`? I see that `PLATFORM_DESKTOP_SDL` just skips that line...

That's because SDL uses it's own timer system Iirc... 
GLFW and RGFW use winapi's timer system.

raysan5:
@ColleagueRiley I reviewed it again and it keeps failing. When I change to `PLATFORM_DESKTOP` (GLFW) it works ok but `PLATFORM_DESKTOP_RGFW` complaints about that missing symbol. WinMM is passed to the linker. It only happens for `x86`.

ColleagueRiley:
See if these fix the issue

raysan5:
> See if these fix the issue. This issue didn't happen for me, but this might fix it.

Just applied the same fix on my end and it worked. I don't know why it is not required for GLFW...

ColleagueRiley:
> > See if these fix the issue. This issue didn't happen for me, but this might fix it.
> 
> Just applied the same fix on my end and it worked. I don't know why it is not required for GLFW...

Maybe the GLFW header defines it somewhere. I'm not sure.

ColleagueRiley:
```
  #if defined(SUPPORT_WINMM_HIGHRES_TIMER) && !defined(SUPPORT_BUSY_WAIT_LOOP)
        // NOTE: Those functions require linking with winmm library
        unsigned int __stdcall timeBeginPeriod(unsigned int uPeriod);
        unsigned int __stdcall timeEndPeriod(unsigned int uPeriod);
    #endif
    ```
    
    https://github.com/raysan5/raylib/blob/3d9aafed3b5d183f35a1e8060d0985475101f6db/src/platforms/rcore_desktop.c#L68

raysan5:
@ColleagueRiley Good catch! That functionality should be probably moved from `rcore.c` to every specific platform...

ColleagueRiley:
> @ColleagueRiley Good catch! That functionality should be probably moved from `rcore.c` to every specific platform...

That would be silly

raysan5:
@ColleagueRiley Please, let me know when ready to merge!

ColleagueRiley:
Should be ready to merge now.

raysan5:
@ColleagueRiley Thank you very much for all the work put on integrating this new platform! Merged! 😄

PR #3942, opened 2024-04-28
[raylib.h] Made comments on raylib.h match those present in rcamera.h

lima-limon-inc:
This is a follow up to this PR: https://github.com/raysan5/raylib/pull/3938

This would make the comments in `raylib.h` match the ones present in `rcamera.h`.

Here are the `rcamera.h` comments: https://github.com/raysan5/raylib/blob/master/src/rcamera.h#L119-L125

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -913,11 +913,11 @@ typedef enum {
 
 // Camera system modes
 typedef enum {
-    CAMERA_CUSTOM = 0,              // Custom camera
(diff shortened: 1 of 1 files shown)


raysan5:
@lima-limon-inc thanks for the review!

lima-limon-inc:
> @lima-limon-inc thanks for the review!

On the contrary @raysan5; thank you!

Have a nice week!

PR #3950, opened 2024-05-01
[rcore_desktop] Fix 3693 initial window geometry

### LINKED ISSUE #3693: [rcore_desktop] Window sizes and position with multiple monitor configurations

using current raylib v5.0 release

### Issue description

I'm having this odd issue where if the game is in a release configuration aka `Windows (/SUBSYSTEM:WINDOWS), /ENTRY: mainCRTStartup` If I start the game on my primary monitor it behaves as expected but if I try to open the game while on a different monitor, the game is running but there is no window... yet if I drag the raylib.dll over the exe to open it will trigger the window.
doing it with the console works.

### Environment

Windows 11, OpenGL 3.3.0 , RTX 3080

### Issue Screenshot

https://github.com/raysan5/raylib/assets/113463662/c3c747b6-11d8-45d8-b84a-4973ae83e17a

### Code Example

Just a really simple example.
```
#include <raylib.h>

int main(void) {
    const int screenWidth = 800;
    const int screenHeight = 450;
    InitWindow(screenWidth, screenHeight, "Simple Game");
    SetTargetFPS(60);

        BeginDrawing();
        ClearBackground(BLUE);
        DrawText("Hello World!", 32,32,23,WHITE);
        EndDrawing();
    }

    CloseWindow();

    return 0;
}
```

dertseha:
This pull request fixes #3693 .

Tested under MS Win11, with `core_basic_window.c` sample under `raylib\projects\scripts\`, using a three-monitor setup (see linked issue for monitor layout).

This pull request causes the following regarding window initialization, when started via file-explorer without console on the three screens:
* windowed (`InitWindow(>0, >0, ...)`): window is of requested size, always centred on the monitor where file-explorer is
* fullscreen (`SetConfigFlags(FLAG_FULLSCREEN_MODE);`): Always a fullscreen window on the primary monitor
* windowed fullscreen (`InitWindow(0, 0, ...)`): window is of monitor size, filling the monitor where file-explorer is *)

### Approach

* Fullscreen applications use the primary monitor, exclusively. This is as per recommendation of comment from `glfwCreateWindow()`, which states `"Unless you have a way for the user to choose a specific monitor, it is recommended that you pick the primary monitor."`
* Non-fullscreen applications will have their windows created in two flavours:
  * a) User specified an explicit dimension
  * b) User requested implicit monitor size by specifying `width=0, height=0` (which mimics the behaviour of fullscreen apps)

Either way, non-fullscreen applications will be centred on the monitor where they are created.

The way this solution does this is by querying the monitor after creating the window, and then derive necessary sizes. This also needed a special case of implicit-monitor-sized windows to be created with an initial size of 1x1 pixels, as GLFW prohibits using 0x0.

### Further changes (fixes)

The change of #3923 broke window positioning of a windowed fullscreen app on the primary monitor (on MS Windows): It ended up positioning the window at a very high Y coordinate. This is because [the line](https://github.com/raysan5/raylib/pull/3923/files#diff-769a8a9fb419056857448648271703ed813ed161157a25ace24ce449ae18643aR1528)
```
int posY = monitorY + (monitorHeight - CORE.Window.screen.height)/2;
```
has an integer overflow in case `monitorHeight` is less than `CORE.Window.screen.height` - since the second is an unsigned integer, the first one is treated as such as well, leading to an underflow to a very high unsigned integer value. Halving this then brings it back into possible ranges of a positive signed integer, which is then taken.

How is this possible? The intent is to avoid the taskbar, so `monitorHeight` is describing only the workable area, which is thus less than the screen height.

Side-note: This integer conversion issue is even highlighted in my IDE. There are several such notices across the codebase.

### *) Unknown detail

On my MS Windows machine, creating a windowed fullscreen application on the primary monitor creates a window that has the size of the monitor, yet still shows the taskbar. So, the game considers a larger area to work in, but not all is visible, hidden behind the taskbar.
This is even the case before my change (and after the fix regarding positioning).
Is this the intention? This may require an additional fix.

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -1371,12 +1371,13 @@ int InitPlatform(void)
     // REF: https://github.com/raysan5/raylib/issues/1554
     glfwSetJoystickCallback(NULL);
 
+    GLFWmonitor *monitor = NULL;
     if (CORE.Window.fullscreen)
     {
         // According to glfwCreateWindow(), if the user does not have a choice, fullscreen applications
         // should default to the primary monitor.
 
-        GLFWmonitor *monitor = glfwGetPrimaryMonitor();
+        monitor = glfwGetPrimaryMonitor();
         if (!monitor)
         {
           TRACELOG(LOG_WARNING, "GLFW: Failed to get primary monitor");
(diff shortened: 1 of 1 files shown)


raysan5:
@dertseha Thank you very much for all the work put on this issue, I'll review it as soon as possible.

raysan5:
@dertseha Merged! Thank you very much for reviewing this issue! This is a great improvement! Thanks!

PR #3955, opened 2024-05-01
Allow raymath before raylib

noodlecollie:
In one of my projects, it'd be really useful to allow the `raymath.h` header to be included before `raylib.h`. Currently this is not possible without throwing errors regarding types already being defined before `raylib.h` encounters them.

This PR adds some `#ifdef` brackets around the definitions so that `raymath.h` can define them if it was included first. It also fixes a number of MSVC warnings in various code files. I didn't go through the examples to fix warnings there because there are a lot, and I though it'd make this PR very noisy. You might want to double check that the casts make sense, particularly the one from `int` to `size_t` to `void*` - I think it should be fine since that'd essentially be what the compiler is doing underneath, but best to check.

### DIFF
--- a/examples/others/raymath_vector_angle.c
+++ b/examples/others/raymath_vector_angle.c
@@ -10,10 +10,9 @@
 *   Copyright (c) 2023 Ramon Santamaria (@raysan5)
 *
 ********************************************************************************************/
- 
-#include "raylib.h"
 
 #include "raymath.h"
+#include "raylib.h"
 
 //------------------------------------------------------------------------------------
 // Program main entry point
@@ -30,7 +29,7 @@ int main(void)
     Vector2 v0 = { screenWidth/2, screenHeight/2 };
     Vector2 v1 = Vector2Add(v0, (Vector2){ 100.0f, 80.0f });
     Vector2 v2 = { 0 };             // Updated with mouse position
-    
+
     float angle = 0.0f;             // Angle in degrees
     int angleMode = 0;              // 0-Vector2Angle(), 1-Vector2LineAngle()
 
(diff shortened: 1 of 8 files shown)


raysan5:
@noodlecollie Sorry but I prefer to avoid polluting `raylib.h` with additional checks fro raymath for an specific use case.

noodlecollie:
Forgive me if I'm misunderstanding, but isn't the function of the definitions like `#define RL_VECTOR3_TYPE` in order to allow the header to work properly with `raymath.h`? If so, doesn't this header then already have knowledge of raymath? It seems unusual that the includes would work in one order but not the other, especially since raymath would arguably be the lower-level dependency.

For context, my use case is that `raymath.h` is in my library's public headers and treated as inline, but `raylib.h` is purely internal. This means that if pretty much any of my public library headers is used before `raylib.h`, things break because of multiple definition errors. I'm limited in what I can do because the math header has to be separate from the rest of raylib, so if this PR doesn't go anywhere I'll have to just make edits myself.

raysan5:
@noodlecollie You are right on your assumptions but I still prefer to avoid referencing it on `raylib.h` to cover a specific use case. I recommend patching it locally for the specific project needs.

orcmid:
@raysan5 I cringe at local patching of any of someone else's maintained libraries.  It creates maintenance issues when there is an important update of a supported release and it makes my application code brittle and not portable.

With regard to `raymath.h` specifically, my impression is that it can be used completely standalone from raylib, something I am going to exploit in some tutorials on math for graphical applications.

@noodlecollie If you `#include "raymath.h"` without `RAYMATH_STANDALONE`, you do not need to`#include "raylib.h"`.  `raymath.h` will do that for you and there should be no conflicts.  If there are, that ***is*** a bug.

Issue #3946, opened 2024-04-30
[core] RPI Opengl - Wrong DRM connector?


colesnicov:
### Issue description

I got have a same display and have has same the problem like his one: https://www.reddit.com/r/raylib/comments/ykmke0/raylib_on_1280x400_hdmi_screen_w_fbconrotate3/

When I run core_basic_window, everything is fine, well almost. I had to edit the examples, change the size (width) of the window in them from 800 px to a smaller one. My display is in a vertical position, so it is 400 px wide. when I try to create a window in the program that is larger than the screen resolution, instead of one window, 4 small ones are displayed, which in addition slowly disappear in black, and then only some text and letters are visible... <s>When I adjusted the window size to 400px, everything works.</s> When I adjusted the window size to 300px, so everything works. 

Here is the output at 300px width:

```
denis@raspberrypi:~/dev/raylib/examples $ ./core/core_basic_window 
INFO: Initializing raylib 5.1-dev
INFO: Platform backend: NATIVE DRM
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
INFO:     > raudio:.... loaded (optional)
INFO: DISPLAY: No graphic card set, trying platform-gpu-card
INFO: DISPLAY: Failed to open platform-gpu-card, trying card1
INFO: DISPLAY: Failed to open graphic card1, trying card0
INFO: DISPLAY: Selected DRM connector mode 384x1280 (384x1280p@62)
INFO: DISPLAY: Upscaling required: Screen size (300x450) smaller than display size (384x1280)
INFO: DISPLAY: Device initialized successfully
INFO:     > Display size: 384 x 1280
INFO:     > Screen size:  300 x 450
INFO:     > Render size:  300 x 450
INFO:     > Viewport offsets: 0, 550
WARNING: GetCurrentMonitor() not implemented on target platform
WARNING: GetMonitorWidth() not implemented on target platform
WARNING: GetCurrentMonitor() not implemented on target platform
WARNING: GetMonitorHeight() not implemented on target platform
WARNING: SetWindowPosition() not available on target platform
INFO: GL: Supported extensions count: 59
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Broadcom
[...]

PR #3956, opened 2024-05-02
Implementing GetMonitorWidth/Height for DRM

gabriel-marques:
Hello :)

I would like to propose an implementation for the following functions for the DRM platform : 

- int GetMonitorWidth(int monitor)
- int GetMonitorHeight(int monitor)
- int GetMonitorPhysicalWidth(int monitor)
- int GetMonitorPhysicalHeight(int monitor)
- const char *GetMonitorName(int monitor)

Although the monitor selection is not supported for the drm platform I kept the argument to avoid having multiple function prototypes.

I hope it complies to your coding rules :)

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -389,18 +389,16 @@ Vector2 GetMonitorPosition(int monitor)
 int GetMonitorWidth(int monitor)
 {
     int width = 0;
-    if (monitor > 0)
+
+    if (monitor != 0)
     {
         TRACELOG(LOG_WARNING, "GetMonitorWidth() implemented for first monitor only");
     }
-    else
+    else if ((platform.connector) && (platform.modeIndex >= 0))
     {
-        if ((platform.connector) && (platform.modeIndex >= 0))
-        {
-            width = platform.connector->modes[platform.modeIndex].hdisplay;
-        }
+        width = platform.connector->modes[platform.modeIndex].hdisplay;
     }
-
+    
     return width;
 }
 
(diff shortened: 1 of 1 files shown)


colesnicov:
Hello. What would it be good for? In addition to the fact that I can find out these dimensions... At the moment we have encountered such an unpleasant dilemma regarding DRM: https://github.com/raysan5/raylib/issues/3946. Can you please tell me about it?

orcmid (changes requested):
In `GetMonitorName(int monitor)`

for consistent behavior with the unimplemented case, you should use

```
const char name = "";
```

Also, to avoid many warnings about the non-zero `monitor`, you might want to do something about the non-zero value case, providing something like "unknown" or "unavailable".  This will also avoid compiler warnings about `monitor` not being used :).

It looks like it might be valuable to have a `GetMonitorCount()` or similar as part of `raylib.h` and have that always return `1` for DRM. 

Isn't there a `raylib.h` way to `GetCurrentMonitor()` or something like that?  (I am being lazy.)

colesnicov:
I dont know yet. I just tuk here for few days. tests, tests, tests...

gabriel-marques:
Hello @orcmid :)

Thanks for the suggested changes of consistency, I will apply them :)
By the way, there is both `GetMonitorCount()` and  `GetCurrentMonitor()` which although log that they are not implemented, returns the correct value (because the actual implementation is limited to one screen only)

```C
// Get number of monitors
int GetMonitorCount(void)
{
    TRACELOG(LOG_WARNING, "GetMonitorCount() not implemented on target platform");
    return 1;
}

// Get number of monitors
int GetCurrentMonitor(void)
{
    TRACELOG(LOG_WARNING, "GetCurrentMonitor() not implemented on target platform");
    return 0;
}
```

raysan5:
@gabriel-marques Implementation looks good to me. Please, let me know when this PR is ready for merging!

gabriel-marques:
Hello,

As @orcmid suggested I changed the `getMonitorName` to return "" as default and also I changed others functions to also evaluate negative monitors.

raysan5:
@gabriel-marques Thanks! Implementation looks good to me! Merging!

PR #3957, opened 2024-05-03
Expose FBO aspect

pfeodrippe:
So we can customize the 3d mode if desired.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -988,6 +988,7 @@ RLAPI int GetScreenWidth(void);                                   // Get current
 RLAPI int GetScreenHeight(void);                                  // Get current screen height
 RLAPI int GetRenderWidth(void);                                   // Get current render width (it considers HiDPI)
 RLAPI int GetRenderHeight(void);                                  // Get current render height (it considers HiDPI)
+RLAPI float GetCurrentAspect();                                   // Get current aspect (from the current FBO)
 RLAPI int GetMonitorCount(void);                                  // Get number of connected monitors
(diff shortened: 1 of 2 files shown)


pfeodrippe:
For more context (and an example), I'm using this to read GLTF metadata and set the Camera rotation \o Let me know if this is okay.

raysan5:
@pfeodrippe Is it possible to do the same with `GetRenderWidth()` and `GetRenderHeight()`? I would prefer to avoid adding this specific function.

pfeodrippe:
Hi @raysan5, if you say that you could use GetRenderWidth and GetRenderHeight at https://github.com/raysan5/raylib/blob/f1007554a0a8145060797c0aa8169bdaf2c1c6b8/src/rcore.c#L989 instead of referring to currentFbo, then yes, let me know if this will always be correct.  

Thank you very much for the lib and review :D

pfeodrippe:
@raysan5 So are they the same thing based on your 👍 ? Can we use, for example, CORE.Window.currentFbo.width and GetRenderWidth() interchangeably?

raysan5:
@pfeodrippe Afaik, there could be a difference when using RenderTextures but it should be analized more carefully. Still, I prefer not to expose the function on its current form. In any case, I'd prefer to update render size for RenderTexture if requried.

pfeodrippe:
> @pfeodrippe Afaik, there could be a difference when using RenderTextures but it should be analized more carefully. Still, I prefer not to expose the function on its current form. In any case, I'd prefer to update render size for RenderTexture if requried.

Thanks o/

PR #3960, opened 2024-05-04
[raudio] Volume, Pan and Pitch getters

LeandroSQ:
Following the current methods to set `Volume`, `Pan` and `Pitch` this pull request implements a getter for the same properties.

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -11189,6 +11189,17 @@
       ]
     },
     {
+      "name": "GetSoundVolume",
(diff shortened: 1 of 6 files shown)


raysan5:
@LeandroSQ I'm afraid I'm not merging this PR, despite following the academic styling of get/set pair. It's up to the user to manage the values they set, so getters shouldn't be needed.

LeandroSQ:
I understand your point, 
I just thought that since we have this pattern on the `master` volume, the same style could be applied on audio streams for symmetry.
Just like #3434 

Thank you once more for this amazing work!

Issue #3889, opened 2024-03-29
[raudio] Failing to load a MP3-File causes a double free


FishingHacks:
### Issue description

When using `LoadMusicStream()` on a file that is not valid mp3, raylib crashes with `free(): double free detected in tcache 2`.

I tried digging a bit deeper and upon further (tho not very detailed) inspection, i found out that `drmp3_init_file` closes the file if it could not successfully load ([`dr_mp3.h:3536`](https://github.com/raysan5/raylib/blob/master/src/external/dr_mp3.h#L3536)). `pclose` is later called again by `drmp3_uninit` ([`raudio.c:1448:54`](https://github.com/raysan5/raylib/blob/master/src/raudio.c#L1448)). `drmp3_uninit` seems to be causing the issue as removing it fixes it. I also noticed that `drmp3_uninit` calls `drmp3__free_from_callbacks`, but that shouldnt be called anyway because `drmp3_init_internal` calls it when it fails ([`dr_mp3.h:2854:9`](https://github.com/raysan5/raylib/blob/master/src/external/dr_mp3.h#L2854)).

### Environment

Platform: Desktop

### Code Example

```c
#include "raylib/src/raylib.h"
#include "stdio.h"

int main(void) {
    InitAudioDevice();
    if (!IsAudioDeviceReady()) {
        printf("couldn't initialize the audio device :notlikethis:");
        return -1;
    }

    Music stream = LoadMusicStream("music.mp3");

    PlayMusicStream(stream);

    while (IsMusicStreamPlaying(stream)) {
        UpdateMusicStream(stream);
    }

    UnloadMusicStream(stream);

    CloseAudioDevice();

    return 0;
}
```

FishingHacks:
Checking other formats, the .wav format also has the same double free error message, so i suspect its the same issue with `drwav_uninit`. The QOA Format segfaults, i didn't investigate yet why tho

RobLoach:
Able to reproduce a similar issue with the `audio_music_stream` example if you delete `resources/country.mp3`.

```
WARNING: FILEIO: [resources/country.mp3] Music file could not be opened
INFO: TIMER: Target time per frame: 33.333 milliseconds
audio_music_stream: /home/rob/Documents/raylib/src/external/dr_mp3.h:3775: drmp3_seek_to_start_of_stream: Assertion `pMP3 != ((void *)0)' failed.
Aborted (core dumped)
```

raysan5:
[...]

PR #3966, opened 2024-05-07
Fixes for loading Music

veins1:
Fix for #3889
Fixes for QOA crashes.
Memory leak fix on .wav loading.


raysan5:
@veins1 Nice! Thanks for the review! I like it!

PR #3967, opened 2024-05-07
Draft: Replace GLAD.h with internal OpenGL loading inside RLGL

ColleagueRiley:
Cuts out the GLAD.h dependency by loading only the used OpenGL functions by hand.


ColleagueRiley:
This does not properly compile with MSVC yet because MSVC does not have `GL/glext.h`.

ColleagueRiley:
This doesn't seem to compile with MSVC properly because MSVC doesn't like how I'm doing typedef PROC defines...

JeffM2501:
What is the advantage of this? Simply making rlgl harder to maintain? glad is one header

raysan5:
> What is the advantage of this? Simply making rlgl harder to maintain? glad is one header

Actually the advantage was minimize init time and make `rlgl.h` truly self-contained and portable, avoiding glad dependency. Note that for OpenGL ES 2.0 backend (PLATFORM_DRM, PLATFORM_ANDROID) no glad is used, only required extensions are manually loaded.

But I agree the additional code is considerable...

PR #3969, opened 2024-05-08
(PLATFORM_DESKTOP_RGFW) Fix Makefile issues (RGFW) (linux) (macOS)

ColleagueRiley:
Sorry, I should've done some tests for this and this should've been included in the last PR.

Found some issues with compiling using the shared method for PLATFORM_DESKTOP_RGFW.


raysan5:
@ColleagueRiley Thanks for the review!

PR #3970, opened 2024-05-09
Do not use "nanosleep" on windows at all (PLATFORM_DESKTOP_RGFW)

ColleagueRiley:
An issue was found the RGFW platform while a user was trying to create bindings wherein mingw did not define `nanosleep`.

It turns out I was wrong in assuming that all distros of mingw support `nanosleep`. 
This fixes the issue by never using nanosleep when compiling for windows. 

Source issue: https://github.com/ColleagueRiley/RGFW/issues/13


raysan5:
@ColleagueRiley thanks for the review!

PR #3971, opened 2024-05-10
remove #define RGFWDEF and make the #undefs only happen for their OS (RGFW)

ColleagueRiley:
While working on RGFW support, before I found out I could use RGFW with Raylib without compiling RGFW on it's own I needed to add this line `#define RGFWDEF`. 

This line makes it so the RGFW functions are not inline'd. This could cause potential performance issues for RGFW. I also made it so the `#defs` and `#undefs` required to get RGFW with raylib only happen if they're needed for that OS. This is a very minor change but it prevents any unforeseen issues.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -592,12 +592,12 @@ endif
 ifeq ($(PLATFORM),PLATFORM_DESKTOP_RGFW)
     ifeq ($(PLATFORM_OS),WINDOWS)
         # Libraries for Windows desktop compilation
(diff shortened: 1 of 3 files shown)


ColleagueRiley:
This should hopefully be the last `PLATFORM_DESKTOP_RGFW` related PR I do for a while.

raysan5:
@ColleagueRiley Nice! Thank you very much for the improvement! 👍

PR #3974, opened 2024-05-11
Fix parsing of empty parentheses

Filyus:
This fix allows to parse `()` like `(void)` in the function parameters.
Another way to fix is to add `void` into following lines:
https://github.com/raysan5/raylib/blob/f62202198e35161a9fe394a14bbd1a7c655107ad/src/rlgl.h#L594-L595 these lines are from the commit: https://github.com/raysan5/raylib/commit/4b0e25d3af79ce4c5cd94e8279473764d969cca3

### DIFF
--- a/parser/raylib_parser.c
+++ b/parser/raylib_parser.c
@@ -1006,8 +1006,14 @@ int main(int argc, char* argv[])
             {
                 funcEnd = c + 2;
 
-                // Check if previous word is void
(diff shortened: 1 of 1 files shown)


raysan5:
@Filyus Actually those functions need to be reviewed. C requires defining `void` if function should not receive parameters, empty parenthesis mean any number of input parameters.

raysan5:
@Filyus I'm merging in case of that malformed side-case but it shouldn't be required.

PR #3977, opened 2024-05-12
[examples] Use logarithmic scaling for 2d examples with zoom functionality

myQwil:
Makes zoom increments feel more natural. Otherwise, increments can feel too small when zoomed in, or too big when zoomed out.

### DIFF
--- a/examples/core/core_2d_camera_mouse_zoom.c
+++ b/examples/core/core_2d_camera_mouse_zoom.c
@@ -63,10 +63,9 @@ int main ()
             camera.target = mouseWorldPos;
 
             // Zoom increment
-            const float zoomIncrement = 0.125f;
-
-            camera.zoom += (wheel*zoomIncrement);
-            if (camera.zoom < zoomIncrement) camera.zoom = zoomIncrement;
+            float scaleFactor = 1.0f + (0.25f * fabsf(wheel));
+            if (wheel < 0) scaleFactor = 1.0f / scaleFactor;
+            camera.zoom = Clamp(camera.zoom * scaleFactor, 0.125, 64);
         }
 
         //----------------------------------------------------------------------------------


rmn20:
You can also implement something similar with powf, like here:
```
float scaleFactor = 1.125f;
// *= scaleFactor, if you scroll up
// /= scaleFactor, if you scroll down
// if you scroll multiple times in one frame, the zoom will be faster
camera.zoom *= powf(scaleFactor, GetMouseWheelMove());
// add clamping probably?
```
This way you also get natural feeling increments. This solution seems more straightforward to me, but maybe the expf solution has some advantages that I fail to see.

raysan5:
@rmn20 Personally I prefer this solution, it seems more simple to me. Also, I prefer to apply it to only one of the examples (`core_2d_camera_mouse_zoom`), just as an example, while keeping the others simpler.

myQwil:
> maybe the expf solution has some advantages that I fail to see.

Basically, the `pow(x, y)` function is often implemented as `exp(log(x) * y)`. So to make the code more efficient, we can store the `log(x)` part of that calculation in a variable instead of having to calculate it every time.

myQwil:
If it's just the one example that this change is being applied to, I'd also like to propose using a click-and-drag motion instead of the mouse wheel:
```diff
@@ -48,9 +48,8 @@ int main ()
             camera.target = Vector2Add(camera.target, delta);
         }
 
-        // Zoom based on mouse wheel
-        float wheel = GetMouseWheelMove();
-        if (wheel != 0)
+        // Zoom based on mouse left click
+        if (IsMouseButtonPressed(MOUSE_BUTTON_LEFT))
         {
             // Get the world point that is under the mouse
             Vector2 mouseWorldPos = GetScreenToWorld2D(GetMousePosition(), camera);
@@ -61,9 +60,11 @@ int main ()
             // Set the target to match, so that the camera maps the world space point 
             // under the cursor to the screen space point under the cursor at any zoom
             camera.target = mouseWorldPos;
-
+        }
+        if (IsMouseButtonDown(MOUSE_BUTTON_LEFT))
+        {
             // Zoom increment
-            float scaleFactor = 1.25f * wheel;
+            float scaleFactor = 1.0f + 0.01f * GetMouseDelta().x;
             if (scaleFactor < 0) scaleFactor = 1.0f / -scaleFactor;
             camera.zoom = Clamp(camera.zoom * scaleFactor, 0.125, 64);
         }
```

raysan5:
@myQwil Actually examples are intended for users to learn so we can add several approaches to show users the different possibilities. I'm merging this PR and adding the proposed alternative.

PR #3978, opened 2024-05-12
[raymath] Add extern "C" to raymath header for C++

JeffM2501:
This just adds a C extern to raymath if it's compiled as C++ so that C++ compilers and linters don't generate warnings about initializer lists.


Peter0x44:
This doesn't help for GCC. Same -Wmissing-field-initializers warnings get generated, turning off name mangling won't do anything for that.

Peter0x44:
What I said also applies to clang, it generates the same.

Unfortunately, I think this has no clean solution
the only one that really would solve this is having a macro like:
```cpp
#ifdef __cplusplus
#define RAYMATH_STRUCT_ZERO {}
#else
#define RAYMATH_STRUCT_ZERO {0}
#endif
```

and using it in place of {0} throughout raymath.h. Unfortunately that's ugly and invasive, and also stops you from being able to directly copy individual functions out of raymath.h for your own modification

raysan5:
@JeffM2501 wow! I missed that one! thanks for the update!

JeffM2501:
This PR needs to be reverted, it breaks debug builds in MinGW-W64 when raymath is included from a cpp file and links raylib static. As @Peter0x44 pointed out, it doesn't fix warnings, so it's not a big deal to revert. Please see PR #3985 for the revert.

PR #3979, opened 2024-05-12
feat(zig): add `opengl_version` option

alexmozaidze:
Added `opengl_version` option to `src/build.zig`.

I was manual-testing it on a Linux machine, and it works well. However, I am unsure of how correct the code is for target-specific OpenGL options. Currently, it prioritizes user-defined option and doesn't do anything linking-related so the user handles it himself (like linking GLESv2).


raysan5:
@alexmozaidze thanks for the review! 😄

PR #3981, opened 2024-05-14
[parser] Don't crash for files that don't end in newlines

Peter0x44:
The parser assumes all lines end in newlines, but sometimes this isn't true. Check for a null terminator along with '\n' when stripping leading spaces.


raysan5:

PR #3983, opened 2024-05-15
Zig build script - LazyPath.path has been deprecated, using b.path()

CosmicBagel:
This works in zig 0.12 (current stable), LazyPath.path has been removed in zig 0.13 (development)

Currently building raylib is broken in 0.13, this change fixes that while not breaking building with current stable 0.12


raysan5:
@CosmicBagel nice! thanks for the review! 🙂

PR #3985, opened 2024-05-16
[RAYMATH] Revert Extern 'C' in raymath

JeffM2501:
This PR reverts the change made in PR #3978
This change for some reason (may be a bug) breaks links in debug builds of MinGW-W64 in projects that include raymath.h from a c and cpp file.
Until we can figure this out, we should not have the extern "C" in raymath so that stuff works.
As @Peter0x44 pointed out the change in PR #3978 does not actually fix any warnings, so there is no problem undoing it.


raysan5:

PR #3988, opened 2024-05-17
[rcore] Draft: OpenGL 4.6 Support

impact112:
This pull request is a draft for adding OpenGL 4.6 support in Raylib.

OpenGL 4.6 offers additional features such as DSA, SPIR-V and NOERROR context.

I currently implemented basic support for opengl46 context creation and a config flag for enabling noerror context.

DSA implentations of rlgl functions are not yet implemented. The first part of rlgl where i plan to add DSA support is the batching system.

Testing was only done on Linux with Mesa and Nvidia 550.


raysan5:
I'm afraid I'm not adding the additional OpenGL version, it seems it's only intended for the NOERROR_CONTEXT but it adds an additional version to maintain for that small benefit, only useful for some very specific situations.

PR #3990, opened 2024-05-19
[rmodels] Fix out of bounds error in GenMeshTangents for loop.

sgalindo:
Found a bug in the GenMeshTangents function when calling it with a generated plane mesh. Pointers to tan1 and tan2 go out of bounds since the int i would increment up to vertexCount instead of vertexCount - 3.


raysan5:
@sgalindo Good catch! Thanks for the review!

PR #3994, opened 2024-05-21
fix WaveCrop() and use frames instead of samples

### LINKED ISSUE #3993: [raudio] WaveCrop() is broken
`initSample` is treated as being in both \[frames\] and \[frames * channels\] units. This was probably an oversight in 7f6cd93 as `initSample` is still multiplied by `wave->channels` in the call to `memcpy(3)` which renders the function unusable.

Although I think it would make more sense for `initSample` and `finalSample` to be measured in \[frames\] (and renamed to `initFrame` and `endFrame`), as I can't think of a reason to support splitting frames, because with interleaved channel data, you may end up getting different starting and ending offsets for different channels, and the channels' data would be shifted, i.e: the n'th channel would have the data for the k'th channel, with `k = (n + initSamples) % wave->channels`.

I can make a PR for this if needed, but unless there's a good reason for using samples, I'd go ahead and use frames instead.

ListeriaM:
fix #3993

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1274,17 +1274,17 @@ Wave WaveCopy(Wave wave)
     return newWave;
 }
 
-// Crop a wave to defined samples range
+// Crop a wave to defined frames range
 // NOTE: Security check in case of out-of-range
-void WaveCrop(Wave *wave, int initSample, int finalSample)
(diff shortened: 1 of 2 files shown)


raysan5:
@ListeriaM Thanks for the review!

PR #3997, opened 2024-05-21
[rcore] Call SDL_GL_SetSwapInterval() after GL context creation

### LINKED ISSUE #3982: [rcore] uncapped framerate not working properly with SDL platform

### Issue description

(this is on latest master commit)

When compiling with the SDL platform, removing/not using SetTargetFPS does not seem to really work: framerate is actually matching the screen refresh rate.

- Take any example, just remove SetTargetFPS(60)
- Then right after BeginDrawing(), add DrawFPS(10, 10);

When doing this with the GLFW platform, you basically get unlocked/uncapped framerate: in my case it can be more than 10000-20000 FPS depending on the example.

When doing this exact same thing, but with SDL platform, I get 115-120 FPS, which is really close to my screen refresh rate (120 HZ).

### Environment

- Ubuntu 23.10 x86_64 under Gnome
- AMD Ryzen 9 7900 + AMD Radeon 5700XT
- SDL Platform

### Issue Screenshot

With GLFW platform
image

With SDL platform
image
image

### Code Example

- Just take for instance textures_image_loading.c
- Comment out SetTargetFPS
- Add DrawFPS
- Build first with GLFW, then with SDL
- Observe completely different framerate

JupiterRider:
Fixes #3982

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1499,11 +1499,6 @@ int InitPlatform(void)
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
     }
 
-    if (CORE.Window.flags & FLAG_VSYNC_HINT)
-    {
-        SDL_GL_SetSwapInterval(1);
-    }
-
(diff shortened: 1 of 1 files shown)


raysan5:
@JupiterRider Nice! Thanks for the review!

PR #4002, opened 2024-05-23
[rmodels] Multiplication of colors in `DrawModelEx` which can be simplified

Bigfoot71:
The multiplication of colors in `DrawModelEx` can be simplified by performing 4 integer divisions rather than 8 float conversions/divisions.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -3546,10 +3546,10 @@ void DrawModelEx(Model model, Vector3 position, Vector3 rotationAxis, float rota
         Color color = model.materials[model.meshMaterial[i]].maps[MATERIAL_MAP_DIFFUSE].color;
 
         Color colorTint = WHITE;
-        colorTint.r = (unsigned char)((color.r*tint.r)/255);
-        colorTint.g = (unsigned char)((color.g*tint.g)/255);
-        colorTint.b = (unsigned char)((color.b*tint.b)/255);
-        colorTint.a = (unsigned char)((color.a*tint.a)/255);
+        colorTint.r = (unsigned char)(((int)color.r*(int)tint.r)/255);
+        colorTint.g = (unsigned char)(((int)color.g*(int)tint.g)/255);
+        colorTint.b = (unsigned char)(((int)color.b*(int)tint.b)/255);
+        colorTint.a = (unsigned char)(((int)color.a*(int)tint.a)/255);
 
         model.materials[model.meshMaterial[i]].maps[MATERIAL_MAP_DIFFUSE].color = colorTint;
         DrawMesh(model.meshes[i], model.materials[model.meshMaterial[i]], model.transform);
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -3546,10 +3546,10 @@ void DrawModelEx(Model model, Vector3 position, Vector3 rotationAxis, float rota
         Color color = model.materials[model.meshMaterial[i]].maps[MATERIAL_MAP_DIFFUSE].color;
 
         Color colorTint = WHITE;
-        colorTint.r = (unsigned char)((((float)color.r/255.0f)*((float)tint.r/255.0f))*255.0f);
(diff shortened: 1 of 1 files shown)


JayLCypher:
I suppose that the cast back down to unsigned char already truncates the values so this seems like a good simplification to me. Since the implicit conversion of integer promotion will be signed int, it will most likely hold the value of 255 * 255 in any scenario, which makes the division also fit within the truncation. 

Explicit conversion of color.r * tint.r to unsigned int is up to preference.

**TL;DR:** LGTM FWIW :+1: 
(Set up a little [GodBolt](https://godbolt.org/z/z8ssofY1P) link to check the various cast and div outputs.)

raysan5:
@Bigfoot71 @JayLCypher I'm afraid I don't understand this change... `color.r*tint.r` clearly overflows, just letting the CPU handle that overflow. Seems dangerous for the potential performance improvement...

Bigfoot71:
@raysan5, as mentioned by @JayLCypher, the expression `color.r*tint.r` undergoes integer promotion and is thus implicitly converted to a signed integer. Afterwards, the result is divided by 255. A simple demonstration `(255*255)/255 == 255` shows that no overflow of type `unsigned char` is possible as long as `color.r` and `tint.r` are of this type.

To maintain the same level of "clarity" as the previous version, we can write `(unsigned char)((int)(color.r*tint.r)/255)` to explicitly indicate what will be implicitly done.

raysan5:
> To maintain the same level of "clarity" as the previous version, we can write (unsigned char)((int)(color.r*tint.r)/255) to explicitly indicate what will be implicitly done.

@Bigfoot71 Personally I think is clearer: `(unsigned char)(((int)color.r*(int)tint.r)/255)`, the cast CPU is supposed to do is implicit in the code.

Bigfoot71:
> @Bigfoot71 Personally I think is clearer: `(unsigned char)(((int)color.r*(int)tint.r)/255)`, the implicit cast CPU is supposed to be doing is explicit in the code.

@raysan5 Indeed, I wanted to quickly write my response, sorry. The change has been made. Personally, I have never seen any criticism of this method. It's up to you whether to choose it or not.

I would choose it personally, however the clarity of understanding of raylib is also very important, and you are the best to choose from that! :)

raysan5:
@Bigfoot71 Thanks for the review! Actually avoiding float usage is a good optimization, wondering if it can be applied in some other raylib functions doing similar color processing, maybe in textures module...

JayLCypher:
> > To maintain the same level of "clarity" as the previous version, we can write (unsigned char)((int)(color.r*tint.r)/255) to explicitly indicate what will be implicitly done.
> 
> @Bigfoot71 Personally I think is clearer: `(unsigned char)(((int)color.r*(int)tint.r)/255)`, the implicit cast CPU is supposed to be doing is explicit in the code.

Casting is more of a compiler/language operation for the most part, until you hit certain CPU specifications on register loading. Some types of CPUs (iirc ARM) only has one register width, and so there are no "implicit upcasting" theoretically. Signed integers also don't "exist" for CPUs. This is where the types *type*_fast*width*_t, *type*_least*width*_t and other vendor-implementation-defined type variables came into place around C99. See [Fixed Width Integer Types](https://en.cppreference.com/w/c/types/integer).

Other than that, implicit casts should be guaranteed compatible, so making them explicit comes up to preference (I prefer implicit because I get warned when I do ~~bad~~ incorrect stuff). Remember that in C, these are **All-Powerful** C-style casts, so there are potential dangers where you may end up writing an explicit conversion where none was required or the original type changed and suddenly you have a silent bug. The compilers *should* be able to optimize for wasted casts, but errant casts are mostly silent errors. Old advice said to always cast the result of malloc, but it turns out that was a cause for more bugs, and letting the compiler handle all pointer implicit casts from void * is better.

If implicit conversions are of note, I also recommend (for debug builds) of Raylib to add -Wconversion flags (for GCC and Clang). This flag is not covered by -Wall, -Wextra or -Wpedantic. (I also personally roll -Wshadow just in case).
This will highlight many implicit conversions.

Good merge @Bigfoot71 ! :)

PR #4003, opened 2024-05-24
fix WaveCrop(): update wave->frameCount

ListeriaM:
I missed this in my last PR, but the wave resulting from the cropping had the `frameCount` of the old wave.
also allow `finalFrame = wave->frameCount' as the range of frames does not include it.


raysan5:
@ListeriaM oh! good catch! thanks for the fix!

PR #4011, opened 2024-05-29
[parser] MemoryCopy() calls: Prevent buffer overflow by replacing har…

avx0:
…d-coded arguments

In future, if a dev edits the second arg and miscalulates the corresponding 3rd arg, there will be a buffer overflow or the string (2nd arg) will be cut short. This commit prevents that.

### DIFF
--- a/parser/raylib_parser.c
+++ b/parser/raylib_parser.c
@@ -202,9 +202,12 @@ int main(int argc, char* argv[])
 {
     if (argc > 1) ProcessCommandLine(argc, argv);
 
-    if (inFileName[0] == '\0') MemoryCopy(inFileName, "../src/raylib.h\0", 16);
-    if (outFileName[0] == '\0') MemoryCopy(outFileName, "raylib_api.txt\0", 15);
-    if (apiDefine[0] == '\0') MemoryCopy(apiDefine, "RLAPI\0", 6);
+    const char *raylibhPath = "../src/raylib.h\0";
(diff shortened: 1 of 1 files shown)


avx0:
MemoryCopy() does not do input validation on its args. 

1. Is `count == size(srcPtr)`?
I wrote some code to set `count` equal to `size(srcPtr)`, ignoring the user's input of `count` (this prevents out-of-bounds write / buffer overflow if user inputs `count > size(srcPtr)`.  I don't know the reliability of my size() function, please test it and suggest modifications.

```
// my edit -- begin
int size(char *a)
{
     int sz = 0;
     if (sizeof(a) != 8) return sizeof(a);
     while (a[sz] != '\0') sz++;
     return sz;
}
// my edit -- end

static void MemoryCopy(void *dest, const void *src, unsigned int count)
{
     char *srcPtr = (char *)src;
     char *destPtr = (char *)dest;

// my edit -- begin
     count = size(srcPtr);
// my edit -- end

// out-of-bounds write
     for (unsigned int i = 0; i < count; i++) destPtr[i] = srcPtr[i];
}
```

2. ` size(destPtr) >? size(srcPtr)`
This issue still remains. There is no reliable way of checking it.

raysan5:
@avx0 I'm merging this change but note that I usually expect the user to provide reliable data, I can't consider all the breaking situations resulting from a  missuse of the library, it only adds **code complexity and maintenance burden**, and only to cover a very minimum number of use-case situations that hardly ever will happen.

PR #4012, opened 2024-05-29
[build.zig] Fix Zig emscripten build

### LINKED ISSUE #4010: [build] Zig emscripten build broken?

### Issue description

I think #3983  broke the Zig 0.12.0 build. If I pass `--sysroot` with an absolute path I now get
```
thread 637865 panic: sub_path is expected to be relative to the build root, but was this absolute path: '/home/lilith/Developer/emsdk/upstream/emscripten/cache/sysroot/include'. It is best avoid absolute paths, but if you must, it is supported by LazyPath.cwd_relative
```
If I pass `--sysroot` with relative paths
```
thread 637149 panic: reached unreachable code
/home/lilith/.zigup/cache/0.12.0/files/lib/std/debug.zig:403:14: 0x10b5bfd in assert (build)
    if (!ok) unreachable; // assertion failure
             ^
/home/lilith/.zigup/cache/0.12.0/files/lib/std/fs.zig:280:11: 0x10f0b3a in openDirAbsolute (build)
    assert(path.isAbsolute(absolute_path));
          ^
``` 
due to the `std.fs.openAbsoluteDir` verifying that a path is absolute or not. I think the changes in #3983 are well intended but are maybe too soon. Upgrading to Zig 0.13.0 when it is released will likely require more work than just changing one line...

I have a commit in a fork I am happy to PR that just reverts the change in #3983. I have tested it and it seems to work fine.

Edit: to suggest a compromise, one could also use something other than `openAbsoluteDir` to keep some version compat with newer 0.13.0-dev releases.

### Environment

Cloned master branch and built with Zig version 0.12.0 with
```
zig build -Dtarget=wasm32-emscripten --sysroot "/home/lilith/Developer/emsdk/upstream/emscripten"
```

dylanlangston:
This pull request fixes #4010. In Zig's master branch the path to the emsdk headers must be a relative path, as mentioned https://github.com/ziglang/zig/pull/19623. Previously, in versions >= 0.12, this path was absolute.

This fix splits the code based on the current Zig version. If the current version is greater than 0.12.0 then we use a relative path and otherwise we use an absolute path. _**If upgrading to Zig's master branch this is a breaking change and the path supplied for the `--sysroot` must be relative to Zig's build directory.**_

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -198,13 +198,15 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
                 @panic("Pass '--sysroot \"$EMSDK/upstream/emscripten\"'");
             }
 
-            const cache_include = std.fs.path.join(b.allocator, &.{ b.sysroot.?, "cache", "sysroot", "include" }) catch @panic("Out of memory");
-            defer b.allocator.free(cache_include);
-
-            var dir = std.fs.openDirAbsolute(cache_include, std.fs.Dir.OpenDirOptions{ .access_sub_paths = true, .no_follow = true }) catch @panic("No emscripten cache. Generate it!");
-            dir.close();
-
-            raylib.addIncludePath(b.path(cache_include));
+            if (comptime builtin.zig_version.minor > 12) {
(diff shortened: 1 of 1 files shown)


raysan5:
@dylanlangston thanks for the review!

PR #4013, opened 2024-05-29
[rtext] Added cast to ExportFontAsCode output to fix C++ compiler errors

DarkAssassin23:
Adds casting to the `ExportFontAsCode()` auto-generated `LoadFont` function to fix the following compiler errors when using C++: 
`error: assigning to 'Rectangle *' from incompatible type 'const Rectangle[]' `
`error: assigning to 'GlyphInfo *' from incompatible type 'const GlyphInfo[]`


raysan5:
After reviewing the function more carefully, I just removed the `const` from the static arrays.

PR #4015, opened 2024-05-29
[rtextures] review color tint functions

Bigfoot71:
As requested here: https://github.com/raysan5/raylib/pull/4002#issuecomment-2137166212
I have only modified `ImageColorTint` and `ColorTint`.

Reading through `rtextures.c`, I believe in other cases we could also avoid floating-point conversions, but for now, I have only modified the color multiplication. I can work on this further if it interests you.


raysan5:
@Bigfoot71 Thanks! Nice improvement!

PR #4018, opened 2024-05-30
[rshapes]Circle line collision function

kai-z99:
Added a function to check collision between a line defined by p1, p2 and a circle defined by its position and radius. This code is tested in https://github.com/kai-z99/raylib/blob/circleLineDev/examples/shapes/shapes_collision_area.c

This function is intended to a standalone collision detection function as well as possibly a helper function for a polygon-circle collision function.

The algorithm is based on https://www.jeffreythompson.org/collision-detection/line-circle.php

It is also similar to this one: https://github.com/raysan5/raylib/pull/4000 (which is closed)

Screenshot 2024-05-29 213049
Screenshot 2024-05-29 213032
Screenshot 2024-05-29 213011


raysan5:
@kai-z99 Thanks for the addition! I try to minimize this kind of functions on raylib, they probably belong to a physics engine more than raylib but adding this one for convenience. Thanks!

kai-z99:
@raysan5 Thanks! I was planning to make a polygon to circle collision function, would that be appropriate? I've created a few games in raylib so far and I always felt it would be super handy as opposed to making a custom one for the game or introducing physics lib.

PR #4020, opened 2024-05-31
[build] Fix examples not building with gestures system disabled

Sprixitite:
Build no longer fails with `-DSUPPORT_GESTURES_SYSTEM=OFF` and `-DBUILD_EXAMPLES=ON`


raysan5:
@Sprixitite I don't understand this PR, why do those examples fail on compile if that flag is enabled? It should never happen.

Sprixitite:
When attempting to build with CMake (3.27.2) on Pop!_OS 22.04 running the commands:
```bash
mkdir build && cd build
cmake .. -DCUSTOMIZE_BUILD=ON -DSUPPORT_GESTURES_SYSTEM=OFF && make
```
Fails with linker errors when linking the relevant examples, as `rgestures.h` is not included due to the build option

raysan5:
@Sprixitite Merging, thanks for the review!

PR #4022, opened 2024-06-01
[rmodels] Send full matModel to shader in DrawMesh

### LINKED ISSUE #4005: [rmodels] "matModel" shader uniform is not multiplied with the push/pop matrix stack
### Issue description

While making a transform hierarchy, I noticed that `rmodels.c` DrawMesh function does not set the "matModel" matrix shader uniform relative to the push/pop matrix stack. See here: https://github.com/raysan5/raylib/blob/9d67f4734b59244b5b10d45ce7c8eed76323c3b5/src/rmodels.c#L1392-L1398

"matModel" uniform is set to the model's _local_ transform, then right after, a matModel variable is created,  accumulated with push/pop matrix stack, and used for "matNormal" and "mvp" uniforms

This seems unintuitive, I expect "matModel" uniform to be the model space to world space transform, the same used calculate the "mvp" uniform, the same as the matModel variable.

However, I can work around this by getting the correct matModel in the shader by way of `mat4 matModel = inverse(transpose(matNormal));`

### Issue Screenshot

Based on the Basic Lighting example, I have a player model that gets drawn after using `rlPushMatrix()` and `rlTranslatef(player_position)`, it is incorrectly lit, as if it was positioned at `{0, 0, 0}`.

image

It should look like this:

image

denovodavid:
Fix #4005


raysan5:
@denovodavid nice! thanks for the review!

PR #4025, opened 2024-06-01
update IQM loader

JettMonstersGoBoom:
Will try and load the texture from the same folder as the model. 
Also fixes animation names being corrupt memory, they are now the animation names from the file.


JettMonstersGoBoom:
loading the texture is likely a breaking change actually. since it's been busted this long. feel free to refuse. the animation names one is still valid. I can PR that alone.

PR #4026, opened 2024-06-01
LoadModelAnimationsIQM: fix corrupted animation names

JettMonstersGoBoom:
Correctly copies animation names from IQM animation to raylib animation.


raysan5:
@JettMonstersGoBoom Thanks for the review!

PR #4029, opened 2024-06-02
LoadIQM: attempt to load texture from IQM at loadtime.

JettMonstersGoBoom:
tries to load the texture with the base path of the original IQM file, relative paths should work.
LoadIQM currently does not load textures. this change attempts to resolve this.


JettMonstersGoBoom:
admittedly this could be seen as a breaking change, as it SHOULD load textures but currently doesn't. 
so folks have worked around this feature.

PR #4035, opened 2024-06-03
Add GenModelMipmaps()

paulmelis:
I was a bit surprised to find that a textured model loaded from (say) a glTF file did not use mipmaps automatically. The required code to generate and enable mipmaps isn't a lot, but seems to be an operation one would want to be easy to do. Hence this function.


paulmelis:
Included the wrong code

PR #4036, opened 2024-06-03
Add GenModelMipmaps() for generating mipmaps and setting filtering settings

paulmelis:
I was a bit surprised to find that a textured model loaded from (say) a glTF file did not use mipmaps automatically. The required code to generate and enable mipmaps isn't a lot, but seems to be an operation one would want to be easy to do. Hence this function.


raysan5:
@paulmelis This function could be implemented at user side if required but it also adds coupling between `rtextures` and `rmodels` modules. I prefer to avoid/minimize coupling as much as possible.

PR #4037, opened 2024-06-04
[rmodels] Initial work to correctly handle the node hierarchy in a glTF file

paulmelis:
Having to use glTF files without node hierarchies and without transformations applied to the meshes is really a big limitation. Also, the current handling is not really correct, as it completely ignores the node hierarchy and thus might lead to meshes not getting instanced more than once. 

This patch adds using the nodes in the glTF file to drive the process of creating meshes, instead of only processing the list of meshes and ignoring the node hierarchy. Transforms defined in the nodes (including parent-child effects) are applied to the vertices/normals/tangents of the meshes in the leafs. A separate raylib Mesh is generated for each leaf node (and still generates one Mesh per primitive in a glTF mesh).

Static meshes with more complex hierarchy in terms of parenting and object transforms seem to work fine in my tests. 

Haven't tried any glTF files with animations yet, but those are almost certainly broken. It's also the part where I don't know off-hand how to handle that, so any tips on what is currently going on with respect to bones and animation are appreciated.

Btw, scenes in the glTF files are also ignored (which can contain subsets of nodes, and thus meshes), but that's something I ignore here as well.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4969,9 +4969,14 @@ static Model LoadGLTF(const char *fileName)
             cgltf_mesh *mesh = node->mesh;
             if (!mesh)
                 continue;
-            primitivesCount += mesh->primitives_count;
+
+            for (unsigned int p = 0; p < mesh->primitives_count; p++)
+            {
+                if (mesh->primitives[p].type == cgltf_primitive_type_triangles)
+                    primitivesCount++;
+            }
         }
-        TRACELOG(LOG_DEBUG, "    > Primitives count based on hierarchy: %i", primitivesCount);
+        TRACELOG(LOG_DEBUG, "    > Primitives (triangles only) count based on hierarchy : %i", primitivesCount);
 
         // Load our model data: meshes and materials
         model.meshCount = primitivesCount;
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4895,7 +4895,8 @@ static Model LoadGLTF(const char *fileName)
                      PBR specular/glossiness flow and extended texture flows not supported
           - Supports multiple meshes per model (every primitives is loaded as a separate mesh)
           - Supports basic animations
-          - Node hierarchy and mesh transforms are applied, but the hierarchy is flattened.
+          - Transforms, including parent-child relations, are applied on the mesh data, but the 
+            hierarchy is not kept (as it can't be represented).
           - Mesh instances in the glTF file (i.e. same mesh linked from multiple nodes)
             are turned into separate raylib Meshes.
 
@@ -4907,6 +4908,8 @@ static Model LoadGLTF(const char *fileName)
               > Texcoords: vec2: float
               > Colors: vec4: u8, u16, f32 (normalized)
               > Indices: u16, u32 (truncated to u16)
+          - Scenes defined in the glTF file are ignored. All nodes in the file
+            are used.
 
     ***********************************************************************************************/
 
@@ -5089,17 +5092,17 @@ static Model LoadGLTF(const char *fileName)
             if (!mesh)
                 continue;
 
-            cgltf_float world_transform[16];
-            cgltf_node_transform_world(node, world_transform);
+            cgltf_float worldTransform[16];
+            cgltf_node_transform_world(node, worldTransform);
 
-            Matrix world_matrix = {
-                world_transform[0], world_transform[4], world_transform[8], world_transform[12],
-                world_transform[1], world_transform[5], world_transform[9], world_transform[13],
-                world_transform[2], world_transform[6], world_transform[10], world_transform[14],
-                world_transform[3], world_transform[7], world_transform[11], world_transform[15]
+            Matrix worldMatrix = {
+                worldTransform[0], worldTransform[4], worldTransform[8], worldTransform[12],
+                worldTransform[1], worldTransform[5], worldTransform[9], worldTransform[13],
+                worldTransform[2], worldTransform[6], worldTransform[10], worldTransform[14],
+                worldTransform[3], worldTransform[7], worldTransform[11], worldTransform[15]
             };
 
-            Matrix world_matrix_normals = MatrixTranspose(MatrixInvert(world_matrix));
+            Matrix worldMatrixNormals = MatrixTranspose(MatrixInvert(worldMatrix));
 
             for (unsigned int p = 0; p < mesh->primitives_count; p++)
             {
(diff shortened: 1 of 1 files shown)


raysan5:
@paulmelis Thank you very much for working on this feature! It's a nice addition to the library!

About glTF animation, I did not implemented it but you can check `LoadModelAnimationsGLTF()`; in raylib side, it uses a Bone system to define skeleton, where every vertex is assigned a maximum of 4-bones influence. Then there is a Transform matrix for every bone for every key-frame. There is no interpolation implemented, animation just consists of transforming on CPU side all affected vertices by every Bone group and re-uploading transformed vertices data to GPU.

No GPU-skinning is supported due to limitations on some platforms supported by raylib.

paulmelis:
Am I reading https://github.com/raysan5/raylib/blob/e1379afb01e5ad758b3928d055e418c2fb511d2a/src/rmodels.c#L5513-L5519 correctly in that `animVertices` and `animNormals` are always allocated, regardless if any animation is present? That potentially doubles memory usage for mesh vertices and normals.

raysan5:
> that animVertices and animNormals are always allocated, regardless if any animation is present? That potentially doubles memory usage for mesh vertices and normals.

@paulmelis Mmmh... it seem so...

paulmelis:
I updated the animation data part as well. The robot animation in `examples/shaders_shadowmap` still seems to work.

raysan5:
> I updated the animation data part as well. The robot animation in examples/shaders_shadowmap still seems to work.

Nice! Thanks for the review! Just let me know when this improvement is ready for review and merge!

paulmelis:
I've tested it with a few glTF models, including with more complex hierarchies. I saw no weird behaviour. I'd say it's ready for going into "beta testing" state

giovancris:
Hey @paulmelis great work on this! 

I noticed that using this model https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/CesiumMan/glTF-Binary/CesiumMan.glb we have some issues!

I had no time to debug this one, but I think it might be useful to report it here :)

paulmelis:
> I noticed that using this model https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/CesiumMan/glTF-Binary/CesiumMan.glb we have some issues!
> 
> I had no time to debug this one, but I think it might be useful to report it here :)

It looks like this is an issue with `LoadModelAnimations()` in raylib. That code isn't touched by this pull request, though. So this particular file triggers a bug in the current raylib animation loading code, which I'm not familiar with. It's best to create a separate bug report for this case.

```
$ cat load_model_animations.c 
#include "raylib.h"

int main(void)
{
    InitWindow(1024, 512, "");

    int count;
    ModelAnimation* robotAnimations = LoadModelAnimations("CesiumMan.glb", &count);
}

$ gcc -o load_model_animations load_model_animations.c -lraylib
$ gdb ./load_model_animations
(gdb) run
Starting program: /home/melis/concepts/battlegroup-raylib/tests/load_model_animations 
...
INFO: Initializing raylib 5.1-dev                                                                                                                                                                                  
INFO: Platform backend: DESKTOP (GLFW)
...
INFO: FILEIO: [CesiumMan.glb] File loaded successfully

Program received signal SIGSEGV, Segmentation fault.
__strncpy_avx2 () at ../sysdeps/x86_64/multiarch/strncpy-avx2.S:85
85              VMOVU   (%rsi), %VMM(0)                                                                                                                                                                            
(gdb) bt
#0  __strncpy_avx2 () at ../sysdeps/x86_64/multiarch/strncpy-avx2.S:85
#1  0x00007ffff7cc3434 in LoadModelAnimations () from /usr/lib/libraylib.so.500
#2  0x000055555555519f in main () at load_model_animations.c:8
```

VitoTringolo:
Yeah I saw this one! And just for testing I added a NULL check to go ahead!

Adding the check I was able to correctly load the model and with l the original raylib code the animation looks good with your changes something goes wrong! 

I think should be related to the world transform matrix we use to transform the vertices, because if that matrix is set to identity the animation come back to be normal!

raysan5:
@paulmelis Afair, animation names support was added quite recently in a PR, probably not extensively tested.

giovancris:
Hey yeah I saw this one but I just bypassed with NULL check!

I was referring to the wrong animation. I think we are having a problem with the worldMatrix we are using to transform the vertices, if you set this matrix to identity the animation is OK. 

In the original code the animation for the this model looks good!

paulmelis:
> I was referring to the wrong animation. I think we are having a problem with the worldMatrix we are using to transform the vertices, if you set this matrix to identity the animation is OK.

Do you have more details (or a code snippet) to show what you mean, as it's not entirely clear to me which worldMatrix you mean?

paulmelis:
So I'm still not sure if what you're seeing is fully caused by my changes to the glTF handling. When I re-export CesiumMan.glb from Blender and load it **into current raylib** (so without my patches for glTF) I'm seeing the same distortions and weird movement. And similar for some other models (e.g. https://sketchfab.com/3d-models/model-47a-loggerhead-sea-turtle-c438e81e796d41d9a6ae4cc147ef8d4f). 

There's also different aspects that influence each other:

* Current glTF import (without these patches) is more or less broken, as it ignores the node hierarchy in the glTF file, leading to incorrect relative mesh placement and instancing. This impacts all use of more complex glTF files (i.e. beyond a single mesh in a file), regardless if they contain animation data. The reason CesiumMan.glb actually worked so far was that it only contains a single mesh, which gets deformed correctly through the bone hierarchy.
* Current import of glTF animations also does not appear to be correct in all cases (the examples above)
* Raylib in itself does not support a true node hierarchy when it comes to meshes/models.
* Due to the previous point the patches in this pull request work around this limitation by transforming all mesh data in leaf nodes by their net local-to-world transform. That's actually quite a hack, but it's about the only thing possible to get complex glTF models to import correctly (as long as there's no node hierarchies in Raylib). It's also causing some of the issues you're seeing with animated models.

I'm actually wondering if it is even possible to correct support models with complex hierarchies **and** having animation data without some form of node hierarchy.

giovancris:
Ah OK! I see, and thank you for the explanation :)

MrScautHD:
AMAZING WORK!

raysan5:
@paulmelis Is this PR ready for merge (despite the current limitations with nodes-hierarchy)?

paulmelis:
Hi @raysan5, I'd say it's a definite improvement for importing glTF models, as you no longer need to flatten the hierarchyi in the file. I don't fully know about the impact on animated models, but I suspect current support isn't perfect either (i.e. #4055). So I would suggest indeed merging this and then we can look at improving animation support on top of it later.

raysan5:
@paulmelis Merged! Thank you very much for all the effort put on this improvement!

Issue #1217, opened 2020-04-23
Raylib doesn't build with windows.h


ZelCloud:
Raylib has a lot of same named functions to windows.h such as Rectangle, LoadImage, DrawText etc etc. Could we get some kind of workaround for this?

scottmudge:
Use CMake to generate a MSVS project file. I have just confirmed it compiles fine on Windows.

You'll tend to get those errors if you attempt to compile the raylib C files inside a separate C++ project.

ZelCloud:
Hey @scottmudge , it compiles fine on windows by itself, the problem is when you're trying to compile it with the windows.h header for native window functions.

raysan5:
@ZelCloud you have three options to avoid conflicts:

 1. Definine the required symbols from `windows.h` instead of including the full header.

 2. Compile the `windows.h` with functions using it on an independent module/code-unit (a separate .c file to generate a separate .o) and link with it.

 3. Try to avoid conflics with `windows.h` defines:
```c
#if defined(_WIN32)
// To avoid conflicting windows.h symbols with raylib, some flags are defined
// WARNING: Those flags avoid inclusion of some Win32 headers that could be required
// by user at some point and won't be included...
//-------------------------------------------------------------------------------------

// If defined, the following flags inhibit definition of the indicated items.
#define NOGDICAPMASKS     // CC_*, LC_*, PC_*, CP_*, TC_*, RC_
#define NOVIRTUALKEYCODES // VK_*
#define NOWINMESSAGES     // WM_*, EM_*, LB_*, CB_*
#define NOWINSTYLES       // WS_*, CS_*, ES_*, LBS_*, SBS_*, CBS_*
#define NOSYSMETRICS      // SM_*
#define NOMENUS           // MF_*
#define NOICONS           // IDI_*
#define NOKEYSTATES       // MK_*
#define NOSYSCOMMANDS     // SC_*
#define NORASTEROPS       // Binary and Tertiary raster ops
#define NOSHOWWINDOW      // SW_*
#define OEMRESOURCE       // OEM Resource values
#define NOATOM            // Atom Manager routines
#define NOCLIPBOARD       // Clipboard routines
#define NOCOLOR           // Screen colors
#define NOCTLMGR          // Control and Dialog routines
#define NODRAWTEXT        // DrawText() and DT_*
#define NOGDI             // All GDI defines and routines
#define NOKERNEL          // All KERNEL defines and routines
#define NOUSER            // All USER defines and routines
[...]

Issue #243, opened 2017-03-09
[build] raylib doesn't build under mingw-w64


Phyllostachys:
I tried building raylib 1.6 in MinGW-w64 but there are name conflicts with the wingdi system that MinGW-w64 exposes. I tried taking that include path out but ```GL/gl.h```, as required by GLFW3, is in the w32api folder path which includes the ```wingdi.h``` files.

The log looks like this:
```bash
$ make
gcc -c core.c -O1 -Wall -std=gnu99 -fgnu89-inline -Wno-missing-braces -I. -Iexternal -Iexternal/glfw3/include -Iexternal/openal_soft/include -DPLATFORM_DESKTOP -DBUILDING_STATIC
In file included from /usr/include/w32api/windows.h:71:0,
                 from /usr/include/w32api/GL/gl.h:13,
                 from external/glfw3/include/GLFW/glfw3.h:168,
                 from core.c:71:
/usr/include/w32api/wingdi.h:3049:28: error: ‘Rectangle’ redeclared as different kind of symbol
   WINGDIAPI WINBOOL WINAPI Rectangle(HDC hdc,int left,int top,int right,int bottom);
                            ^~~~~~~~~
In file included from core.c:45:0:
raylib.h:345:3: note: previous declaration of ‘Rectangle’ was here
 } Rectangle;
   ^~~~~~~~~
In file included from /usr/include/w32api/windows.h:72:0,
                 from /usr/include/w32api/GL/gl.h:13,
                 from external/glfw3/include/GLFW/glfw3.h:168,
                 from core.c:71:
/usr/include/w32api/winuser.h:2210:29: error: conflicting types for ‘CloseWindow’
   WINUSERAPI WINBOOL WINAPI CloseWindow (HWND hWnd);
                             ^~~~~~~~~~~
In file included from core.c:45:0:
raylib.h:623:12: note: previous declaration of ‘CloseWindow’ was here
 RLAPI void CloseWindow(void);                                     // Close Window and Terminate Context
            ^~~~~~~~~~~
In file included from /usr/include/w32api/windows.h:72:0,
                 from /usr/include/w32api/GL/gl.h:13,
                 from external/glfw3/include/GLFW/glfw3.h:168,
                 from core.c:71:
/usr/include/w32api/winuser.h:3609:25: error: conflicting types for ‘ShowCursor’
   WINUSERAPI int WINAPI ShowCursor(WINBOOL bShow);
                         ^~~~~~~~~~
[...]

PR #4039, opened 2024-06-06
[rlgl] Rename near, far variables

jgabaut:
Hi, I found a couple more instances of the `near` and `far` names being used in `rlgl.h`.

I don't really know if this could incure the same problem I found with `raymath.h`, but I tought it might be useful to also rename this one.

This relates to this old issue that mentions these names:
- https://github.com/raysan5/raylib/issues/243

I would also mention that I'm not yet using the suggested setup to select includes from `windows.h`, from here:
- https://github.com/raysan5/raylib/issues/1217

Which may fix most of this kind of issues, should anyone incur in the same warnings as [in here.](https://www.reddit.com/r/raylib/comments/1d8m52k/question_why_does_including_raymathh_make_my/)


raysan5:
@jgabaut Sure! those functions were added recently

PR #4040, opened 2024-06-06
The example always showed a xbox controller, never a ps3 controller

KonradGrande:


raysan5:
@KonradGrande thanks for the review!

PR #4042, opened 2024-06-06
fix(cmake): infer CMAKE_MODULE_PATH in super-build

fruzitent:
Raylib as a library should not override the `CMAKE_MODULE_PATH' variable. This is the way it is used in [glfw](https://github.com/raysan5/raylib/blob/5.0/src/external/glfw/CMakeLists.txt#L62).

As it also depends on `find_package()` calls which expect system-wide installed dependencies, this prevents you from using custom `Find<target>.cmake` files to build them in-source.

In CMake this is a bit hacky, but in e.g. meson this is natively supported [behavior](https://mesonbuild.com/External-Project-module.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ cmake_policy(SET CMP0054 NEW)
 
 # Directory for easier includes
 # Anywhere you see include(...) you can check <root>/cmake for that file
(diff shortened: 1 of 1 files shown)


raysan5:
@fruzitent thanks for the review! Please note that I do not maintain the CMake build system, in case this PR results in some unexpected behaviour, it is not tested by me.

PR #4043, opened 2024-06-07
Bazel support

zaucy:
Hello! I was wondering if it would be acceptable to bring in some bazel files. This would make consuming raylib as a bazel user much easier. The goal would be to submit raylib to the [BCR](https://registry.bazel.build/) so that a user may add raylib as easily as this:

```python
bazel_dep(name = "raylib", version = "5.0")
```

and then in their `BUILD.bazel` files they can refer to raylib like so:

```python
cc_binary(
  name = "my_cool_game",
  srcs = [...],
  deps = [
    # .. other deps ...
    "@raylib",
  ],
)
```

NOTE: if having bazel files in your repo is undesirable I can submit to the BCR with patches instead.

Let me know what you think!

## TODO

- [ ] add bazel files for models examples
- [ ] add CI for testing bazel builds
- [ ] add archive generation on release https://blog.bazel.build/2023/02/15/github-archive-checksum.html
- [ ] add web support
- [ ] add macOS support
- [ ] add android support
- [ ] add build files for raylib parser


asdqwe:
Hello,

Would it be possible to group all those files on a `bazel/` or `build/bazel/` directory? Or something like that?

At the moment we have build files for different build systems (make, cmake, zig) spread all over the place, it's becoming a mess.

Best regards.

zaucy:
Unfortunately not. Files are only detectable by bazel if they are underneith a bazel package (the folder where a `BUILD.bazel` file exists.)

If it is a real burden to keep these bazel files where they are currently then what I could do is remove all the bazel files in the `examples` directory (although that would be a real shame for anyone coming to the repository to try raylib with bazel) and add the main `MODULE.bazel` and `BUILD.bazel` file into the release archive when raylib cuts a release. That would have a major downside making the repository unusable with [git_override](https://bazel.build/rules/lib/globals/module#git_override) and would only be usable once submitted to the BCR, which would also make testing more difficult.

raysan5:
@zaucy Sorry but I prefer to avoid polluting the repo with more specific binding/building files.

aaylward:
@zaucy are you still considering adding bazel support with patches on the BCR side?

zaucy:
Yes I do. Since the BCR supports 'overlay' style submissions now it is much easier to test as well. I just haven't gotten around to it yet.

PR #4046, opened 2024-06-07
[rlgl] Reviewed `rlUpdateShaderBuffer()`, it can not update the size

MrScautHD:
With this change `rlUpdateShaderBuffer` updates the size of the buffer.

### DIFF
--- a/examples/others/rlgl_compute_shader.c
+++ b/examples/others/rlgl_compute_shader.c
@@ -104,7 +104,7 @@ int main(void)
         else if (transfertBuffer.count > 0)  // Process transfert buffer
         {
             // Send SSBO buffer to GPU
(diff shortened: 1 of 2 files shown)


raysan5:
@MrScautHD I'm afraid I don't understand this PR, it looks confusing to me on its intended behaviour... It should be reviewed more carefully...

MrScautHD:
> @MrScautHD I'm afraid I don't understand this PR, it looks confusing to me on its intended behaviour... It should be reviewed more carefully...

After you set the size in loadbuffer you can not anymore change the size in the buffer update

raysan5:
> After you set the size in loadbuffer you can not anymore change the size in the buffer update

And that's the expected behaviour, `rlUpdateShaderBuffer` is sued to **update** an existent pre-defined shader buffer, not change its size.

MrScautHD:
:/

PR #4049, opened 2024-06-09
Make addRaylib use options.opengl_version

carverdamien:


raysan5:
@carverdamien Not an expert on Zig build system but it seems a reasonable change.

carverdamien:
Me too. I had to make this change to use 2.1 on RPi4. The default version doesn't work. Would you like to see how I'm using this?

PR #4051, opened 2024-06-09
[SHAPES] Make functions that draw point arrays take them as const

JeffM2501:
This PR makes a small change to the draw functions that take arrays of vector2, and vector3 to take the pointers as const pointers not mutable pointers.
This is logically const correct, as the points will not be changed during the function.
This also makes it easier to store the data in vectors in C++.

This should not change any existing behavers or code since the mutable pointers will be const cast.


src/raylib.h
```diff
@@ -1292,7 +1292,7 @@ RLAPI bool CheckCollisionCircleRec(Vector2 center, float radius, Rectangle rec);
 RLAPI bool CheckCollisionPointRec(Vector2 point, Rectangle rec);                                         // Check if point is inside rectangle
 RLAPI bool CheckCollisionPointCircle(Vector2 point, Vector2 center, float radius);                       // Check if point is inside circle
 RLAPI bool CheckCollisionPointTriangle(Vector2 point, Vector2 p1, Vector2 p2, Vector2 p3);               // Check if point is inside a triangle
-RLAPI bool CheckCollisionPointPoly(Vector2 point, Vector2 *points, int pointCount);                      // Check if point is within a polygon described by array of vertices
+RLAPI bool CheckCollisionPointPoly(Vector2 point, const Vector2 *points, int pointCount);                      // Check if point is within a polygon described by array of vertices
```
raysan5, line 1295: Hi @JeffM2501, please, could you review the comments alignment?

raysan5:
@JeffM2501 I was a bit hesitant about this change, I remember it was already proposed in the past and rejected... but let's add it now. Thanks for the review!

PR #4053, opened 2024-06-09
[rmodels] `LoadBoneInfoGLTF()`, add check for animation name being NULL

VitoTringolo:
Same issue described here https://github.com/raysan5/raylib/pull/4037 with this model https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/BrainStem/glTF-Binary/BrainStem.glb

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4864,7 +4864,7 @@ static BoneInfo *LoadBoneInfoGLTF(cgltf_skin skin, int *boneCount)
     for (unsigned int i = 0; i < skin.joints_count; i++)
     {
         cgltf_node node = *skin.joints[i];
-        strncpy(bones[i].name, node.name, sizeof(bones[i].name));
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -4864,7 +4864,9 @@ static BoneInfo *LoadBoneInfoGLTF(cgltf_skin skin, int *boneCount)
     for (unsigned int i = 0; i < skin.joints_count; i++)
     {
         cgltf_node node = *skin.joints[i];
-        strncpy(bones[i].name, node.name, sizeof(bones[i].name));
+        if (node.name != NULL) {
```
raysan5, line 4867: Please, could you avoid the brackets, in raylib one line statements use no-brackets:
```c
if (node.name != NULL) strncpy(bones[i].name, node.name, sizeof(bones[i].name));
```

raysan5:
@VitoTringolo thanks for the review!

PR #4054, opened 2024-06-09
Update `Makefile`

lzralbu:
Since the flag `CUSTOM_CFLAGS` should allow users customize things, in particular the optimization level, I figured it should come after other default ones.

Based on: "If you use multiple -O options, with or without level numbers, the last such option is the one that is effective."

Source: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html


raysan5:
@lzralbu thanks for the review!

PR #4056, opened 2024-06-11
Fix fragPosition using wrong matrix in lighting_instancing.vs

karl-zylinski:
fragPosition was multiplied by mvp*instanceTransform, but it should only be multiplied by instanceTransform. Compare to lighting.vs, there we only use mvp for gl_Position, but matModel for the fragPosition.


raysan5:
@karl-zylinski Oh! Good catch! Thanks for the review! 😄

PR #4059, opened 2024-06-12
Fix rlgl.h standalone/implementation

TokyoSU:
Hello it's my first contribution to this project, i did found this error when compiling with #define RLGL_IMPLEMENTATION:
- The function rlGetMatrixProjectionStereo and rlGetMatrixViewOffsetStereo have RLAPI for the function declaration and cause error when compiling using RLGL_IMPLEMENTATION.


raysan5:
@TokyoSU Good catch! Thanks for the review!

PR #4062, opened 2024-06-13
[raudio] fix mapping of wave data in LoadWaveSamples()

### LINKED ISSUE #4060: [raudio] Incorrect mapping of wave data in `LoadWaveSamples()`
The current implementation will map `wave.data[i]` to the wrong range if `wave.sampleSize` is 8 or 16:
```c
if (wave.sampleSize == 8) samples[i] = (float)(((unsigned char *)wave.data)[i] - 127)/256.0f;
else if (wave.sampleSize == 16) samples[i] = (float)(((short *)wave.data)[i])/32767.0f;
else if (wave.sampleSize == 32) samples[i] = ((float *)wave.data)[i];
```

If it's 16, a value of `wave.data[i] = -32768` (`SHRT_MIN`) will yield a sample value outside of the expected range of [-1..1], so the divisor should be `32768.0f`. If it's 8 the divisor should be `128.0f` (currently the output range is [-0.5..0.5]).
Of course these can never yield a sample value of `1.0f`, but that is expected given the asymmetry of the input range.

Regarding the offset of 127 with `wave.sampleSize = 8`:
In my testing, an offset of 128 yielded better results (this is also the value used by libsndfile). That is to say a value of `wave.data[i] = 128` would output no sound, but using `wave.data[i] = 127` I could hear a popping sound at the start and end of the track, which makes sense given that the range of `signed char` is [-128..127], so a value of `wave.data[i] = 0` (the minimum) should correspond to -128. Here's the code I used to test this:

```c
#include <raylib.h>

int
main(void)
{
    int i;
    Wave wave;

    wave.frameCount = 44100;
    wave.sampleRate = 44100;
    wave.sampleSize = 8;
    wave.channels = 2;
    wave.data = MemAlloc(wave.frameCount*wave.channels*wave.sampleSize/8);

    if (!IsWaveReady(wave))
        TraceLog(LOG_FATAL, "failed to initialize wave");

    for (i = 0; i < wave.frameCount*wave.channels; i++)
        ((unsigned char *)wave.data)[i] = 127;
    ExportWave(wave, "127.wav");

    for (i = 0; i < wave.frameCount*wave.channels; i++)
        ((unsigned char *)wave.data)[i] = 128;
    ExportWave(wave, "128.wav");

    MemFree(wave.data);
}
```
afterwards listen to `127.wav` and `128.wav`.

If this seems reasonable I could make a PR adjusting all 3 mentioned values.

ListeriaM:
fix #4060

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1304,8 +1304,8 @@ float *LoadWaveSamples(Wave wave)
 
     for (unsigned int i = 0; i < wave.frameCount*wave.channels; i++)
     {
-        if (wave.sampleSize == 8) samples[i] = (float)(((unsigned char *)wave.data)[i] - 127)/256.0f;
-        else if (wave.sampleSize == 16) samples[i] = (float)(((short *)wave.data)[i])/32767.0f;
+        if (wave.sampleSize == 8) samples[i] = (float)(((unsigned char *)wave.data)[i] - 128)/128.0f;
+        else if (wave.sampleSize == 16) samples[i] = (float)(((short *)wave.data)[i])/32768.0f;
         else if (wave.sampleSize == 32) samples[i] = ((float *)wave.data)[i];
     }
 


raysan5:
@ListeriaM thanks for the review!

PR #4063, opened 2024-06-13
[build] zig: run examples from their directories

myQwil:
This way, when we run an example using the `zig build [example_name]` command, it will be able to find the resource folder associated with that section.


raysan5:
@myQwil Thanks for the improvement! I'm afraid I'm not expert with Zig, anyone with more experience than me on this build system should review this PR.

Gamer-Kold:
looks good to me!

Not-Nik:
+1

Not-Nik (approved):

raysan5:
@myQwil thanks for the improvement! @Not-Nik @Gamer-Kold thanks for the review!

Issue #3998, opened 2024-05-21
[rmodels] Read uninitialized values in `GenMeshTangents()`


ListeriaM:
This commit (c4a51a3) will skip the last iteration if `mesh->vertexCount % 3 == 0` (I suppose this is the normal case), which results in the last values for `tan1` and `tan2` being uninitialized. In any case this reads uninitialized memory from the end of `tan1` and `tan2` in the next loop regardless of the value of `mesh->vertexCount`. Did you mean to use `i <= mesh->vertexCount - 3`? Maybe a better option would be to assert that `mesh->vertexCount % 3 == 0` and keep the old loop, since otherwise the last 1 or 2 values from `tan1` and `tan2` will always be uninitialized, which is probably not what you want.

\cc @sgalindo

raysan5:
@ListeriaM Thanks for the further review, I didn't notice it...

PR #4066, opened 2024-06-14
[rmodels] Read uninitialized values in GenMeshTangents() - fix bounding case

kai-z99:
reference to issue: https://github.com/raysan5/raylib/issues/3998

Fix the problem where last 3 values are uninitialized in GenMeshTangents(). Also give a warning when this function will generate uninitialized values.


raysan5:
@kai-z99 thanks for the review!

PR #4069, opened 2024-06-14
[build] Fix some warnings on web builds and remove some redundant flags.

lzralbu:


cmake/LibraryConfigurations.cmake
```diff
@@ -59,7 +59,7 @@ if (${PLATFORM} MATCHES "Desktop")
 elseif (${PLATFORM} MATCHES "Web")
     set(PLATFORM_CPP "PLATFORM_WEB")
     set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -s USE_GLFW=3 -s ASSERTIONS=1 --profiling")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} --profiling")
```
Peter0x44, line 62: I was wondering what this did:
https://emscripten.org/docs/tools_reference/emcc.html?highlight=profiling
I think users can control debug builds on their own, by passing this option manually, or perhaps `-DCMAKE_BUILD_TYPE=Debug`
Is there a reason to preserve it still I don't know about?
Perhaps this should just be removed completely?

Peter0x44 (commented):
One minor nit, otherwise LGTM.

cmake/LibraryConfigurations.cmake
lzralbu: > I think users can control debug builds on their own, by passing this option manually

Agreed.

I got no good reason to keep it. Should I remove it then?

Peter0x44: Please do.

raysan5:
@lzralbu thanks for the update. @Peter0x44 thanks for the review

Issue #3763, opened 2024-01-28
[rcore] GetScreenWidth() and GetScreenHeight() do not return current values


Sqvid:
Please, before submitting a new issue verify and check:

 - [X ] I tested it on latest raylib version from master branch
 - [X ] I checked there is no similar issue already reported
 - [ X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [ ?] My code has no errors or misuse of raylib

### Issue description

From my understanding of the function descriptions in the raylib cheatsheet, I would assume GetScreenWidth() and similar functions would return up-to-date values for the window. However, it seems the values returned are those the window was initialised with in InitWindow().

### Environment

*Provide your Platform, Operating System, OpenGL version, GPU details where you experienced the issue.*
Platform: Desktop, Linux 6.6.11, OpenGL version string: 4.6 (Compatibility Profile) Mesa 23.3.2, AMD Ryzen 7 7840HS w/ Radeon 780M Graphics

### Code Example

```C
#include <stdio.h>

#include "raylib.h"

int main(void) {
	SetTraceLogLevel(LOG_WARNING);

	InitWindow(10, 10, "Test");
	SetWindowSize(400, 400);

	printf("ScreenWidth: %d, ScreenHeight: %d\n", GetScreenWidth(), GetScreenHeight());
	printf("RenderWidth: %d, RenderHeight: %d\n", GetRenderWidth(), GetRenderHeight());

	CloseWindow();
	return 0;
}
```

output:

```
WARNING: GLFW: Error: 65539 Description: Invalid window hint 0x0002000D
ScreenWidth: 10, ScreenHeight: 10
RenderWidth: 10, RenderHeight: 10
```

Sqvid:
I want to add that ```SetWindowState(FLAG_WINDOW_RESIZABLE);``` did not change the output. I verified that the flag was set correctly with ```IsWindowState(FLAG_WINDOW_RESIZABLE);```.

raysan5:
@Sqvid I tested the example on Windows and it works as expected:

![image](https://github.com/raysan5/raylib/assets/5766837/60c77d33-2238-44e9-a41b-f7cfdc46aaed)

Looking at this line:

 > WARNING: GLFW: Error: 65539 Description: Invalid window hint 0x0002000D

[...]

PR #4074, opened 2024-06-17
[platform_desktop] fixed issue with GetScreenWidth/GetScreenHeight

CrackedPixel:
There is a known issue that people are having where `GetScreenWidth`/`GetScreenHeight` are returning incorrect values. It only uses the one set during initialization and does not work if you change the screen size. I believe this section of code is missing, and I copied the format from `SetWindowMaxSize` directly above it

`PLATFORM_DEKSTOP_SDL` does the same thing: https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_sdl.c#L661

Note: I don't know about `CORE.Window.render.width`/height or where that one should be added (definitely a separate PR) but it has the same problem and i believe same solution

https://github.com/raysan5/raylib/issues/3763
https://github.com/raysan5/raylib/discussions/3764

### DIFF
--- a/src/platforms/rcore_desktop.c
+++ b/src/platforms/rcore_desktop.c
@@ -666,6 +666,9 @@ void SetWindowMaxSize(int width, int height)
 // Set window dimensions
 void SetWindowSize(int width, int height)
 {
+    CORE.Window.screen.width = width;
+    CORE.Window.screen.height = height;
+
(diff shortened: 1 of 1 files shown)


raysan5:
@CrackedPixel thanks for the review. At first sight it seems logical, wondering why those lines were not there... 🤔

PR #4076, opened 2024-06-19
Build issue: Header conflict with dirent.h on linux

SparklesReal:
When building on linux using and using the header dirent.h it instead tries to use external/dirent.h from raylib and that results in a conflict.

Solution: renamed the external/dirent.h to external/raylibdirent
It now compiles on both Windows and Linux.

[build] In file included from /media/sparkles/Games n stuff1/CODING/RaylibGuiCreator/extern/libtinyfiledialogs/tinyfiledialogs.c:117:
[build] /media/sparkles/Games n stuff1/CODING/RaylibGuiCreator/extern/raylib/src/external/dirent.h:83:10: fatal error: io.h: No such file or directory
[build] 83 | #include <io.h> // _findfirst and _findnext set errno iff they return -1
[build] | ^~~~~~
[build] compilation terminated.


SparklesReal:
I was just dumb and had my Cmake set to include raylib/external for some reason, sorry.

PR #4077, opened 2024-06-19
[rtext] Don't return default font if LoadFontEx fails

Peter0x44:
It is currently impossible to check a font loaded successfully with IsFontReady because LoadFontEx will always return a valid font.

DrawTextEx has this check:
if (font.texture.id == 0) font = GetFontDefault();  // Security check in case of not valid font

So anyone relying on the default font as a fallback for fonts failing to load should still be covered.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -394,7 +394,6 @@ Font LoadFontEx(const char *fileName, int fontSize, int *codepoints, int codepoi
 
         UnloadFileData(fileData);
     }
-    else font = GetFontDefault();
 
     return font;
 }


CrackedPixel:
@Peter0x44 do you think something like this should be applied to loading shaders as well?

raysan5:
@Peter0x44 @CrackedPixel raylib has some fallbacks for convenience, change this behaviour could imply reviewing several assets loading functions... It's not a one-function trivial change.

Peter0x44:
About the shaders, I didn't think about reviewing them, but it can be done at a later time.
i believe this change is self-contained and should be okay.

raysan5:
Just for completion, here it is a list with the result of loading several types of assets (from files) and failing doing it:

 - `Image`: Returns invalid image, `image.data = NULL`
 - `Texture2D`: Returns invalid texture, `texture.id = 0`
 - `Font`: _Fallback to `GetFontDefault()`_ --> Return invalid font -> **Does it CRASH anything?**
 - `Mesh`: Returns invalid vertex data, `mesh.vertex = NULL`, `mesh.vaoId = 0`
 - `Shader`: _Fallback to default shader `RLGL.State.defaultShaderId`_
 - `Material`: _Fallback to `LoadMaterialDefault()`_
 - `Wave`: Returns invalid wave, `wave.data = NULL`
 - `Sound`: Returns invalid sound, `sounds.stream.buffer = NULL`
 - `Music`: Returns invalid music, `music.stream.buffer = NULL`

@Peter0x44 Please, could you review if failing on font loading, with no fallback, any example crashes?

raysan5:
I tested the examples and it seems nothing crashes if font is not found or not properly loaded, it's just not displayed.

Peter0x44:
Thanks! Sorry I didn't get to testing sooner but I reviewed most relevant functions taking Font and they seemed okay.

PR #4079, opened 2024-06-19
[build] CMake: support OpenGL ES3 in `LibraryConfigurations.cmake`

manuel5975p:
Only put ES2 as a fallback if `GRAPHICS` is not specified.


cmake/LibraryConfigurations.cmake
```diff
@@ -58,7 +58,11 @@ if (${PLATFORM} MATCHES "Desktop")
 
 elseif (${PLATFORM} MATCHES "Web")
     set(PLATFORM_CPP "PLATFORM_WEB")
-    set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
+    if(NOT GRAPHICS)
+        set(GRAPHICS "GRAPHICS_API_OPENGL_ES2")
+    elseif(${GRAPHICS} MATCHES "GRAPHICS_API_OPENGL_ES3")
+        set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sFULL_ES3=1")
```
Peter0x44, line 64: For GLFW this is put in target_link_options, rather than CMAKE_EXE_LINKER_FLAGS.
Could you do that instead?

Peter0x44 (changes requested):

cmake/LibraryConfigurations.cmake
manuel5975p: Changed it

Peter0x44:
LGTM! approved.

Peter0x44:
Actually, one last nit:
emscripten documents the option as `-sFULL_ES3` and not `-sFULL_ES3=1`

```
To enable OpenGL ES 3.0 emulation, specify the emcc option -sFULL_ES3 when linking the final executable (.js/.html) of the project.
```

Change that and all should be fine.

manuel5975p:
Doesn't another part of the emscripten doc say that -sOPTION is exactly equivalent to -sOPTION=1?

Issue #3962, opened 2024-05-04
[rtextures] DDS / DXT1, DXT3 and DXT5: unit 0 GLD_TEXTURE_INDEX_2D is unloadable


aboeglin:
### Issue description
I have a sprite sheet that works fine when loaded as png but fails when loaded as DXT. I've tried different softwares to generate it. Texture packer, gimp and online converters all leads to the same issue, the texture seems to load fine:
```
INFO: FILEIO: [./assets/spritesheets/barbarian_jump.dds] File loaded successfully
INFO: IMAGE: Data loaded successfully (1473x1473 | DXT3_RGBA | 1 mipmaps)
INFO: TEXTURE: [ID 7] Texture loaded successfully (1473x1473 | DXT3_RGBA | 1 mipmaps)
```
I've also tried DXT1, 3 and 5 compression, all leading to the same issue:
```
UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
```

### Environment
```
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Apple
INFO:     > Renderer: Apple M1 Pro
INFO:     > Version:  4.1 Metal - 86
INFO:     > GLSL:     4.10
```

### Example file causing the issue
Here is one such image: [barbarian_jump.dds.zip](https://github.com/raysan5/raylib/files/15211062/barbarian_jump.dds.zip).

raysan5:
@aboeglin I'm testing the shared image and I get the following info about it, please could you confirm it is correct:
```
INFO: IMAGE: Data loaded successfully (1473x1473 | DXT5_RGBA | 1 mipmaps)
```

raysan5:
Definitely the issue is related to image not being POT, I tried to resize canvas to POT and now it loads correctly. 
Could you test the attached file?
[barbarian_scaled_to_pot.zip](https://github.com/raysan5/raylib/files/15212334/barbarian_scaled_to_pot.zip)

aboeglin:
@raysan5 yes correct, the provided image was compressed with DXT5. And your POT version does work although I'm pretty sure I tried that when exporting from texture packer already. Is it expected that non POT DXT compressed images can't be bound?

raysan5:
> Is it expected that non POT DXT compressed images can't be bound?

[...]

PR #4080, opened 2024-06-20
add warning log multiple of 4

CrackedPixel:
Conversation taken from https://github.com/raysan5/raylib/issues/3962

Tried to match the LOG format from other ones in the file

Tested to make sure the issue was the correct reason
Tested to make sure the warning displays

It's a long-known issue on Unity and such https://forum.unity.com/threads/crunch-compression-width-height-multiple-of-4-issue.538306/


raysan5:
@CrackedPixel Thanks! Nice improvement! Actually I think there was some related issue...

PR #4082, opened 2024-06-20
Miniaudio thread initialization callbacks

### LINKED ISSUE #4081: [raudio] `MA_ON_THREAD_ENTRY` and `MA_ON_THREAD_EXIT` are not exposed

### Issue description

These preprocessor macros allow users of miniaudio to initialize thread-local state in audio threads. For example, this may be required for some language runtimes to be usable in audio stream callbacks.

KislyjKisel:
Fixes #4081

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -3113,6 +3113,28 @@
           "name": "frames"
         }
       ]
+    },
+    {
+      "name": "AudioThreadEntryCallback",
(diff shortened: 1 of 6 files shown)


raysan5:
@KislyjKisel I'm afraid I prefer not to expose threading capabilities from raylib. Users with that specific need can use miniaudio directly or modify raylib for their needs.

PR #4088, opened 2024-06-22
Update rlgl.h

lzralbu:


raysan5:
I'm afraid this change is not aligned with Matrix memory layout:
```c
// Matrix, 4x4 components, column major, OpenGL style, right-handed
typedef struct Matrix {
    float m0, m4, m8, m12;  // Matrix first row (4 components)
    float m1, m5, m9, m13;  // Matrix second row (4 components)
    float m2, m6, m10, m14; // Matrix third row (4 components)
    float m3, m7, m11, m15; // Matrix fourth row (4 components)
} Matrix;
```
Note that float generated array actually trasposes raylib Matrix definition.

lzralbu:
Yeah, but there is no need to manually transpose the matrix since `glUniformMatrix4fv` already does that if `true`(or non-zero I guess) is specified in the [third parameter](https://docs.gl/gl3/glUniform), as I did.

And at least a bunch of 3D/mesh/shader examples worked the same as before.

raysan5:
@lzralbu Oh! Sorry, I didn't noticed it. You are right. Still, I prefer to keep it as it is, I think it's more visual for new users.

PR #4089, opened 2024-06-22
Fixed LoadCodepoints returning a freed pointer when codepointCount is zero

Minnowo:
When calling `int *LoadCodepoints(const char *text, int *count)`. If the value of `codepointCount` is zero, the function returns a freed pointer.

It now returns null instead.


raysan5:
@Minnowo Good catch! Thanks for the review!

PR #4090, opened 2024-06-22
[build] CMake: pass -sFULL_ES3 instead of -sFULL_ES3=1

manuel5975p:


raysan5:
@manuel5975p Thanks for the review!

PR #4095, opened 2024-06-23
[Build] Fix warnings when building in VS 2022

JeffM2501:
This PR fixes warnings when building in Visual Studio 2022. Mostly typecast warnings.


examples/core/core_smooth_pixelperfect.c
```diff
@@ -69,18 +69,18 @@ int main(void)
...
-        cameraY = cosf(GetTime())*30.0f;
+        cameraX = (sinf((float)GetTime())*50.0f) - 10.0f;
+        cameraY = cosf((float)GetTime())*30.0f;
 
         // Set the camera's target to the values computed above
         screenSpaceCamera.target = (Vector2){ cameraX, cameraY };
 
         // Round worldSpace coordinates, keep decimals into screenSpace coordinates
-        worldSpaceCamera.target.x = (int)screenSpaceCamera.target.x;
+        worldSpaceCamera.target.x = truncf(screenSpaceCamera.target.x);
```
raysan5, line 79: Is it really needed to use `truncf()`?

projects/VS2022/examples/shapes_splines_drawing.vcxproj
```diff
@@ -202,7 +202,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;PLATFORM_DESKTOP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>WIN32;_CRT_SECURE_NO_WARNINGS;_DEBUG;_CONSOLE;PLATFORM_DESKTOP;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
raysan5, line 205: Why is this flag added to this project? Is it required? Is it included in the other example projects?

src/rlgl.h
```diff
@@ -3945,7 +3945,9 @@ void rlSetVertexAttribute(unsigned int index, int compSize, int type, bool norma
     // Additional types (depends on OpenGL version or extensions):
     //  - GL_HALF_FLOAT, GL_FLOAT, GL_DOUBLE, GL_FIXED,
     //  - GL_INT_2_10_10_10_REV, GL_UNSIGNED_INT_2_10_10_10_REV, GL_UNSIGNED_INT_10F_11F_11F_REV
-    glVertexAttribPointer(index, compSize, type, normalized, stride, (void *)offset);
+
+    size_t offsetNative = offset;
+    glVertexAttribPointer(index, compSize, type, normalized, stride, (void *)offsetNative);
```
raysan5, line 3950: Nice! Thanks for this one! It was on my TODO list!

src/rtextures.c
```diff
@@ -3632,10 +3632,10 @@ void ImageDrawTriangle(Image *dst, Vector2 v1, Vector2 v2, Vector2 v3, Color col
 {
     // Calculate the 2D bounding box of the triangle
     // Determine the minimum and maximum x and y coordinates of the triangle vertices
-    int xMin = (v1.x < v2.x)? ((v1.x < v3.x)? v1.x : v3.x) : ((v2.x < v3.x)? v2.x : v3.x);
-    int yMin = (v1.y < v2.y)? ((v1.y < v3.y)? v1.y : v3.y) : ((v2.y < v3.y)? v2.y : v3.y);
-    int xMax = (v1.x > v2.x)? ((v1.x > v3.x)? v1.x : v3.x) : ((v2.x > v3.x)? v2.x : v3.x);
-    int yMax = (v1.y > v2.y)? ((v1.y > v3.y)? v1.y : v3.y) : ((v2.y > v3.y)? v2.y : v3.y);
+    int xMin = (int)((v1.x < v2.x) ? ((v1.x < v3.x) ? v1.x : v3.x) : ((v2.x < v3.x) ? v2.x : v3.x));
```
raysan5, line 3635: Please, could you review formating, in raylib I always put the `?` next to the previous statement with no space.

src/rtextures.c
```diff
@@ -3730,9 +3730,9 @@ void ImageDrawTriangleEx(Image *dst, Vector2 v1, Vector2 v2, Vector2 v3, Color c
     }
 
     // Calculate the initial barycentric coordinates for the top-left point of the bounding box
-    int w1Row = (xMin - v2.x)*w1XStep + w1YStep*(yMin - v2.y);
-    int w2Row = (xMin - v3.x)*w2XStep + w2YStep*(yMin - v3.y);
-    int w3Row = (xMin - v1.x)*w3XStep + w3YStep*(yMin - v1.y);
+    int w1Row = (int)((xMin - v2.x) * w1XStep + w1YStep * (yMin - v2.y));
```
raysan5, line 3733: Please, could you review formating, in raylib I always use the `*` and `/` operators without spaces around.

raysan5:
@JeffM2501 Thanks for the improvements! I added some comments about minor details.

examples/core/core_smooth_pixelperfect.c
JeffM2501: yes, because the result of target uses floats. 
The input is a float and the destination is a float. the use of the int cast here is just to truncate, thus the right thing to do is to call trunc, not cast. To use the cast we'd have to do 
```worldSpaceCamera.target.x = (float)(int)(screenSpaceCamera.target.x);```
and that loos way worse.

projects/VS2022/examples/shapes_splines_drawing.vcxproj
JeffM2501: Because this project uses sscanf, and without this there will be a security warning. Any project that uses printf/scanf needs to do this.
the raylib project itself has defines for this. any example that uses the functions directly needs the same.

src/rtextures.c
JeffM2501: Fixed

src/rtextures.c
JeffM2501: Fixed

projects/VS2022/examples/shapes_splines_drawing.vcxproj
raysan5: I can't find the usage of `sscanf()` (or `printf()`/`scanf()`) in the example code. But I imagine it's because the inclusion of `<stdlib.h>` for the `NULL` pointer definition.

raysan5:
@JeffM2501 Thank you very much for the review! I really like those PRs that keep the code clean!

PR #4097, opened 2024-06-23
[rtextures] Adding `ImageDrawLineEx` function

Bigfoot71:
## Description
I propose adding the `ImageDrawLineEx` function with the same signature as `DrawLineEx` from `rshapes.c`, enabling line rendering with specified thickness on an `Image`. This might not be the most efficient method used here, but it is the most concise, understandable, and maintainable, and it is fully functional.

In this PR, I have also revised the `ImageDrawLine` function to be shorter and more readable in my opinion, thus more maintainable as well. It avoids the use of `float` types entirely, performing all calculations using integers and utilizing a bit-shift trick for slope calculation.

Additionally, I took the liberty of modifying the `ImageDrawLineV` function to round input coordinates to the nearest integer using `(int)(xfloat + 0.5f)`, which I find preferable. However, I can revert this change to the original state if there are objections to this approach.

## Basic Example
```c
#include "raylib.h"
#include <math.h>

int main(void)
{
    InitWindow(800, 600, "Lines");

    Image image = GenImageColor(800, 600, BLACK);
    Texture texture = LoadTextureFromImage(image);

    Vector2 p1 = { 0 };
    Vector2 p2 = { 0 };

    while (!WindowShouldClose())
    {
        BeginDrawing();

        ImageClearBackground(&image, BLACK);

        p1.x = 400 + cosf(GetTime())*1000;
        p1.y = 300 - sinf(GetTime())*1000;
        p2.x = 400 + cosf(GetTime() + PI)*1000;
        p2.y = 300 - sinf(GetTime() + PI)*1000;

        ImageDrawLineEx(&image, p1, p2, 8, BLUE);
        ImageDrawLineEx(&image, (Vector2) { 0, 0 }, (Vector2) { 800, 600 }, 8, RED);
        ImageDrawLineEx(&image, (Vector2) { 800, 0 }, (Vector2) { 0, 600 }, 8, RED);

        UpdateTexture(texture, image.data);

        EndDrawing();
    }

    CloseWindow();

    return 0;
}
```

## GIF Example
screenrec001

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1373,6 +1373,7 @@ RLAPI void ImageDrawPixel(Image *dst, int posX, int posY, Color color);         
 RLAPI void ImageDrawPixelV(Image *dst, Vector2 position, Color color);                                   // Draw pixel within an image (Vector version)
 RLAPI void ImageDrawLine(Image *dst, int startPosX, int startPosY, int endPosX, int endPosY, Color color); // Draw line within an image
 RLAPI void ImageDrawLineV(Image *dst, Vector2 start, Vector2 end, Color color);                          // Draw line within an image (Vector version)
+RLAPI void ImageDrawLineEx(Image *dst, Vector2 start, Vector2 end, int thick, Color color);              // Draw a line defining thickness within an image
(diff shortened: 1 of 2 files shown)


raysan5:
@Bigfoot71 Nice! Thanks for the addition!

PR #4098, opened 2024-06-24
created DrawRectangleRoundedPro and DrawRectangleRoundedLinesPro

brccabral:
DrawRectangleRoundedLinesPro and DrawRectangleRoundedLinesPro have options to selected what corners to be rounded. Existing code won't break because the default options are all `true`


raysan5:
@brccabral I'm afraid I prefer not to merge this functions, I try to keep raylib functions [under 7 parameters](https://github.com/raysan5/raylib/wiki/raylib-syntax-analysis#16-functions-parameters) in size and those functions have too many parameters for raylib. 

In any case, they can be implemented in a separate example, in user code; that's actually the intended use for `rlgl`.

Issue #3231, opened 2023-08-08
[core] `FLAG_WINDOW_RESIZABLE` has no effect on `PLATFORM_WEB`


Peter0x44:
### Issue description
FLAG_WINDOW_RESIZABLE has no effect on PLATFORM_WEB

### Issue Screenshot
![Screenshot_20230808_012136](https://github.com/raysan5/raylib/assets/49879065/bd9d5c2b-50c9-4655-8e16-c9b3d6c68f3c)

### Code Example

```c
#include "raylib.h"

#ifdef __EMSCRIPTEN__
    #include <emscripten/emscripten.h>
#endif

const int screenWidth = 800;
const int screenHeight = 450;

void UpdateDrawFrame(void);


int main(void)
{
    SetConfigFlags(FLAG_WINDOW_RESIZABLE);
    InitWindow(screenWidth, screenHeight, "web resizing example");

#ifdef __EMSCRIPTEN__
    emscripten_set_main_loop(UpdateDrawFrame, 0, 1);
#else
    SetTargetFPS(60);

    while (!WindowShouldClose())
    {
        UpdateDrawFrame();
    }
#endif

    CloseWindow();

    return 0;
}

void UpdateDrawFrame(void)
{
    static Vector2 ballPosition = (Vector2) { screenWidth/2.0f, screenHeight/2.0f };
    static Vector2 ballSpeed = (Vector2) { 6.0f, 4.0f };
    const float ballRadius = 10.0f;

    ballPosition.x += ballSpeed.x;
    ballPosition.y += ballSpeed.y;

    if (((ballPosition.x + ballRadius) > GetScreenWidth()) || ((ballPosition.x - ballRadius) < 0)) ballSpeed.x *= -1.0f;
    if (((ballPosition.y + ballRadius) > GetScreenHeight()) || ((ballPosition.y - ballRadius) < 0)) ballSpeed.y *= -1.0f;

    BeginDrawing();

        ClearBackground(GREEN);

        DrawCircleV(ballPosition, ballRadius, RED);

    EndDrawing();
}
```

compile with:
`emcc resize_example.c -I path/to/raylib/src ./libraylib.a -s USE_GLFW=3 -s WASM=1 --shell-file minshell.html -o resize_example.html`

The following 3 lines seem to do nothing:
https://github.com/raysan5/raylib/blob/db55bed72b45c97616f0cfaf26448475f80174d5/src/rcore.c#L6176-L6178

[...]

PR #4104, opened 2024-06-25
[web] Fix undesired scrollbars on shell files

jspast:
While testing my game, I noticed scrollbars were visible when compiling with minshell.html. First, I thought my RenderTexture was not scaling correctly on the web. After some tests, I discovered that a small change to the minshell.html file completely fixes the problem.

The issue has already been described [here](https://github.com/raysan5/raylib/issues/3231#issuecomment-1713026923).

And even the solution has already been described [here](https://github.com/raysan5/raylib/issues/3231#issuecomment-1713043728) but not merged.

I also added `background-color: black` to the body so that areas not rendered (happens when resizing the window or when not scaling the game) are black instead of white, which I assume should be the default.

Then I checked shell.html and found that it always shows a horizontal scrollbar, even with no RenderTexture scaling. This can also be fixed directly on the shell.html file.

So hopefully these changes make the shell files behave as expected.

## Tests

(For minshell.html I used a modified version of core_basic_window that scales using a RenderTexture)

minshell.html before:
minshell html before

minshell.html after:
minshell html after

shell.html before:
shell html before

shell.html after:
shell html after

### DIFF
--- a/src/minshell.html
+++ b/src/minshell.html
@@ -34,8 +34,12 @@
     <link rel="shortcut icon" href="https://www.raylib.com/favicon.ico">
 
     <style>
-        body { margin: 0px; }
-        canvas.emscripten { border: 0px none; background-color: black; }
+        body { 
+          margin: 0px; 
+          overflow: hidden; 
+          background-color: black;
+        }
+        canvas.emscripten { border: 0px none; background-color: black;}
     </style>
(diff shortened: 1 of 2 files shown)


raysan5:
@jspast thanks for the review! I just fixed it for one of my projects today! Still wondering the root of this issue, why does it happen?

jspast:
Thanks @raysan5 for the merge! To be completely honest, I am also not sure! I was just able to find a solution and thought it was a good idea to help people like me who have no experience with web development.

PR #4109, opened 2024-06-26
[build.zig] Leverage `dependencyFromBuildZig`

InKryption:
Also ran zig fmt.

Using dependencyFromBuildZig, the build script can now directly reference itself as a dependency in helper functions, without having to rely on the naming convention or for the user to match up the names.


raysan5:
@InKryption Thanks for the review!

PR #4107, opened 2024-06-26
Fix GLTF animations

VitoTringolo:
Fix issues described in https://github.com/raysan5/raylib/issues/4055

PR #4112, opened 2024-06-27
rmodels.c, `LoadImageFromCgltfImage()` : fix base64 padding support

SuperUserNameMan:
This should fix the issue related to `.gltf` embeded textures in base64 format, by ignoring the optional `=` padding and calculating the data size in bytes correctly.

Before this patch, `cgltf_load_buffer_base64()` would fail, and the `.gltf`model would not load embedded textures.

Should not interfere with pending https://github.com/raysan5/raylib/pull/4107


raysan5:
@SuperUserNameMan Thanks for the review!

PR #4113, opened 2024-06-27
[build.zig] pass the real build.zig file

InKryption:
Many apologies, #4109 was not thoroughly tested enough to catch this blunder.

`@This()` was naively passed to `dependencyFromBuildZig` while inside a file that is not actually the build file, causing a panic when actually used. Passing `@import("../build.zig")` fixes this.


raysan5:
@InKryption thanks for the review!

PR #4115, opened 2024-06-28
[EXAMPLES] Default Skybox example to not use HDR

JeffM2501:
The skybox example tries to use a HDR texture, but raylib no longer builds with HDR support by default, so this example doesn't work in the base configuration.

This PR changes the example to default to the PNG skybox, and makes a note that you must build raylib with HDR support in order to use the HDR texture in the example.


raysan5:
@JeffM2501 thanks for the review! 😄

PR #4117, opened 2024-06-28
Added more DrawPlane and DrawGrid functions

Atan-D-RP4:
Thought the DrawPlanes and DrawGrid functions should have variants that supported more than one Drawing plane in 3D mode. So I added them in. Though I suspect that the lack of too much variety in this case is intentional to keep the API simple. 

But it's a little less hassle if they are already built-in sometimes, you know. So I went ahead with this PR.

Thank you, for your time in reviewing it.


raysan5:
@Atan-D-RP4 Thanks for the proposed functions but I'm afraid I'm not merging them. I try to minimize the number of functions added to the library because every new function implies and extra maintenance cost. Any user with the need of those specific functions can implement them on their code directly using `rlgl`, that's the reason `rlgl.h` exists and it's distributed along raylib.

Atan-D-RP4:
Yeah, I guessed as much. Thanks for the review still. And your time.

PR #4118, opened 2024-06-30
[rmodels] Return true if no need to interpolate to avoid log flooding

VitoTringolo:


raysan5:
@VitoTringolo thanks for the review!

PR #4120, opened 2024-07-01
[rtextures] advance k in LoadImageColors

brccabral:
Some formats are not advancing k to get pixels values

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -3111,6 +3111,7 @@ Color *LoadImageColors(Image image)
                     pixels[i].b = 0;
                     pixels[i].a = 255;
 
+                    k += 1;
                 } break;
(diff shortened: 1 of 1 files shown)


raysan5:
@brccabral Thanks for the review!

**SIDE NOTE**: I added a `TODO` note with a possible improvement: instead of always returning `PIXELFORMAT_UNCOMPRESSED_GRAYSCALE` pixelformat, it could return other single-channel formats like `PIXELFORMAT_UNCOMPRESSED_R16` and `PIXELFORMAT_UNCOMPRESSED_R32` when the image provided to extract channel is a  16bit/channel or 32bit/channel data. So, no data is lost in those cases (and it's up to the user to convert them to lower resolution if desired).

PR #4121, opened 2024-07-01
[build.zig] Make emscripten build compatible with Zig 0.13.0

myQwil:
When attempting to export a project for the web with emscripten and Zig 0.13.0, I would get the following error:
```
std.debug.panic("sub_path is expected to be relative to the build root, but was this absolute path: '{s}'. It is best avoid absolute paths, but if you must, it is supported by LazyPath.cwd_relative", .{
                       ^
raylib/src/build.zig:213:45: 0x11dd8dd in compileRaylib (build)
                raylib.addIncludePath(b.path(cache_include));
``` 

The proposed change has shown to work for both Zig versions 0.12.0 and 0.13.0.

Also added `.zig-cache/` to .gitignore, as the name of the cache folder gets changed in 0.13.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,7 @@ GRTAGS
 GTAGS
 
 # Zig programming language
+.zig-cache/
 zig-cache/
 zig-out/
 build/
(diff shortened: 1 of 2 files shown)


raysan5:
@myQwil thanks for the review!

PR #4122, opened 2024-07-01
[raudio] check audio device state early

brccabral:
I was getting "frame count" warning message, but the real reason was that I forgot to call `InitAudioDevice()`. I feel that having this check early would better describe the problem.


raysan5:
@brccabral I'm afraid I'm not merging this PR, I prefer to let the user do those checks and decouple audio loading from device state.

PR #4123, opened 2024-07-01
[build.zig] Link all required X11 deps

Cloudef:
The raylib includes glfw header which pulls in bunch of X11 includes. To make sure these X11 includes are in search path we need to linkSystemLibrary them as well.

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -122,6 +122,13 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
                     .X11 => {
                         raylib.defineCMacro("_GLFW_X11", null);
                         raylib.linkSystemLibrary("X11");
+                        raylib.linkSystemLibrary("Xcursor");
+                        raylib.linkSystemLibrary("Xrender");
+                        raylib.linkSystemLibrary("Xrandr");
+                        raylib.linkSystemLibrary("Xinerama");
+                        raylib.linkSystemLibrary("Xi");
+                        raylib.linkSystemLibrary("Xext");
+                        raylib.linkSystemLibrary("Xfixes");
                     },
                     .Wayland => {
                         raylib.defineCMacro("_GLFW_WAYLAND", null);


src/build.zig
```diff
@@ -122,6 +122,13 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
                     .X11 => {
                         raylib.defineCMacro("_GLFW_X11", null);
                         raylib.linkSystemLibrary("X11");
+                        raylib.linkSystemLibrary("Xcursor");
+                        raylib.linkSystemLibrary("Xrender");
+                        raylib.linkSystemLibrary("Xrandr");
+                        raylib.linkSystemLibrary("Xinerama");
+                        raylib.linkSystemLibrary("Xi");
+                        raylib.linkSystemLibrary("Xext");
```
raysan5, line 130: Is this library really requried?

src/build.zig
```diff
@@ -122,6 +122,13 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
...
+                        raylib.linkSystemLibrary("Xfixes");
```
raysan5, line 131: Is this library really required?

src/build.zig
Cloudef: yes

src/build.zig
Cloudef: yes

src/build.zig
Cloudef: XInput2.h includes X11/extensions/Xge.h which is part of Xext

src/build.zig
Cloudef: Xinput2.h includes X11/extensions/Xfixes.h which is part of Xfixes

raysan5:
@Cloudef What version of Linux are you using? I reviewed the raylib required dependencies and in my list I only have:
```
-lGL -lX11 -lXrandr -lXinerama -lXi -lXcursor -lm -lpthread -ldl -lrt
```
And some of them can actually be ignored. depending on the Linux distribution and the compiler.

Afaik, this issue was not reported before by any Zig developer, I'd like to ask Zig community for a second review.

Cloudef:
NixOS, on FHS distros you are unlikely to hit this issue because compilers and build systems also search /usr/lib, /usr/include and so on

gen2brain:
@Cloudef The dependencies you added are build requirements, i.e. header files that are only required during building, you are not supposed to link all those libraries in the final build.

Cloudef:
I know, but afaik this is the only way to tell zig to use pkg-config and include the header search paths. Not sure if it actually links the libraries if the symbols don't end up in the final binary.

Cloudef:
answer seems to be no:
```
~/d/p/knucklebones main• ❱ ldd zig-out/bin/knucklebones
	linux-vdso.so.1 (0x00007f491e6b0000)
	libm.so.6 => /nix/store/dbwp0scbb0rk78m636sb7cvycz8xzgyh-glibc-2.39-52/lib/libm.so.6 (0x00007f491e5c7000)
	libc.so.6 => /nix/store/dbwp0scbb0rk78m636sb7cvycz8xzgyh-glibc-2.39-52/lib/libc.so.6 (0x00007f491e3d9000)
	/nix/store/dbwp0scbb0rk78m636sb7cvycz8xzgyh-glibc-2.39-52/lib/ld-linux-x86-64.so.2 => /nix/store/dbwp0scbb0rk78m636sb7cvycz8xzgyh-glibc-2.39-52/lib64/ld-linux-x86-64.so.2 (0x00007f491e6b2000)
```

raysan5:
@Cloudef I'm afraid I'm not merging this PR, it seems some of those libraries are indeed not required on linking, probably the issue is somewhere else or it can be addressed in another way.

Cloudef:
I don't quite understand. Even the libx11 dep here isn't used runtime, only build time (though this may depend on glfw version). And AFAIK linkSystemLibrary is the build.zig blessed way to basically use pkg-config and bring in the C dependency (be it only headers or the lib).

Problem is that for both non-FHS distros and sandboxed builds raylib build will be broken if the headers aren't in the search path.

This is the output you'll see without this PR:
https://gist.github.com/Cloudef/4e5a61b7a470275cf55a7540471442f0

Cloudef:
I also noticed the drm build is broken as well and fixed it here <https://github.com/Cloudef/raylib/commit/f6960a2d02321c432b4fbf42b1f9cefd8bc01ab5>

Issue #1525, opened 2021-01-11
[physac] MSVC compiler cannot compile this file correctly


object71:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] My code has no errors or misuse of raylib

### Issue description

When compiling with MSVC (generated projects with CMake) I and include physac.h I get the following errors:

`D:\Projects\raylib-cpp\vendor\raylib\src\physac.h(410): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax`

When any of these initializers are met: `(Vector2){ 0.0f, 0.0f }`

This is already solved in raylib.h by the following code

```cpp
#if defined(__cplusplus)
    #define CLITERAL(type)      type
#else
    #define CLITERAL(type)      (type)
#endif

// ...

CLITERAL(Vector2){ 0.0f, 0.0f }
```

### Environment

Happening on windows with VS2019 tools installed. Projects generated in cmake and also while trying to build raylib-cpp with all examples - but it is failing on raylib side not on the language port.

### Code Example

Just clone raylib-cpp and build with examples without tests.

raysan5:
@object71 Yeah, `physac` library requires some review... Feel free to send a PR with this improvement.

raysan5:
@object71 Reviewed physac library, please, let me know if it works.

object71:
@raysan5 It is fine now

PR #4127, opened 2024-07-01
[raymath.h] use CLITERAL for MSVC

brccabral:
I had same error as #1525 , `error C4576`. I applied the same solution which is to use `CLITERAL`. Example `rlgl_standalone.c` does not import `raylib.h`, so I had to add double check for CLITERAL definition.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -77,6 +77,17 @@
     #endif
 #endif
 
+// NOTE: MSVC C++ compiler does not support compound literals (C99 feature)
+// Plain structures in C++ (without constructors) can be initialized with { }
+// This is called aggregate initialization (C++11 feature)
+#ifndef CLITERAL
+    #if defined(__cplusplus)
+        #define CLITERAL(type)      type
+    #else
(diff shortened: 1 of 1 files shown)


raysan5:
@brccabral thanks but I prefer to add additional macros just for that use case, in all `raymath` I already avoided compound-literals usage.

brccabral:
fyi, just in case someone else has the same problem.

The error was
`C:\Program Files\raylib\include\raymath.h(2580,34): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion`

I was getting it because I was targeting the Raylib installation path directly
```cmake
target_include_directories(${PROJECT_NAME} PRIVATE ${RAYLIB_PATH}/include)
target_link_directories(${PROJECT_NAME} PRIVATE ${RAYLIB_PATH}/lib)
```
Removing those lines and using `find_package` solves the error without any modification to `raymath.h`.  
```cmake
find_package(raylib REQUIRED)
```

On Windows, make sure `C:\path\to\raylib` is in your `%PATH%`

PR #4129, opened 2024-07-03
Change SDL_Joystick to SDL_GameController

FrankyBalu:
With SDL_Joystick my game controller wasn't working properly. That's why I changed it to SDL_GameController. Gamepads are now mapped correctly.


raysan5:
@FrankyBalu Thanks for the improvement!

PR #4135, opened 2024-07-07
[rshapes] Give CheckCollisionPointCircle() its own implementation

kai-z99:
Originally, CheckCollisionPointCircle() calls CheckCollisionCircles() with a circle radius 0. This change simplifies and gives this function its own implementation without relying on another function call.


raysan5:
@kai-z99 thanks for the review, it looks good to me.

PR #4136, opened 2024-07-07
Fix Reddit badge in README

ninadsachania:
# Problem

Currently, the Reddit badge in README displays "subreddit is private" even when it's not.

This may mislead newcomers.

Proof:

image

# Solution

I created this static badge that could work as a replacement:

[Reddit Static Badge](https://www.reddit.com/r/raylib/)

Clicking on it takes you to the subreddit.

It does not have the live subscriber count, but atleast it does not mislead people.


raysan5:
@ninadsachania thanks for the improvement!

PR #4139, opened 2024-07-08
Add workaround for NetBSD

NishiOwO:


raysan5:
@NishiOwO thanks for the review!

PR #4143, opened 2024-07-08
[GLFW backend] `WindowSizeCallback()` should not handle DPI scale. GLFW already do that

### LINKED ISSUE #3972: [rcore] `ToggleFullscreen()` not working as expected with HighDPI mode enabled
when enabling the highDPI flag in raylib (SetConfigFlags(FLAG_WINDOW_HIGHDPI);) and using ToggleFullscreen(); the fullscreen Window is twice the size it should be and you only see the top left quarter of your software. Toggling again will half the window size each time you toggle fullscreen until the window is gone.

image

steps to repdroduce:
SetConfigFlags(FLAG_WINDOW_HIGHDPI); on a high dpi monitor
call ToggleFullscreen(); 

I am using Windows 11

SuperUserNameMan:
This fix #3972 on my Linux desktop.

Awaiting confirmation on Windows platform.

MACOS not impacted.

If `FLAG_WINDOW_HIGHDPI` is set, `InitPlatform()` will ask GLFW to handle the resizing of window content area based on the monitor content scale using : ` glfwWindowHint(GLFW_SCALE_TO_MONITOR, GLFW_TRUE); `

So `WindowSizeCallback()` should not try to handle DPI rescale a second time.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1664,23 +1664,9 @@ static void WindowSizeCallback(GLFWwindow *window, int width, int height)
     if (IsWindowFullscreen()) return;
 
     // Set current screen size
-#if defined(__APPLE__)
+
     CORE.Window.screen.width = width;
     CORE.Window.screen.height = height;
-#else
-    if ((CORE.Window.flags & FLAG_WINDOW_HIGHDPI) > 0)
-    {
-        Vector2 windowScaleDPI = GetWindowScaleDPI();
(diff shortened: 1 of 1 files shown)


raysan5:
@SuperUserNameMan @veins1 Looks good to me, thanks for the review!

SuperUserNameMan:
This PR improved the issue, but it was not the REAL solution.

I think i've found where the REAL problem was from.
I'll come back with the REAL solution with a new PR soon.

PR #4144, opened 2024-07-08
update RGFW to RGFW 1.0

ColleagueRiley:
I would also like to know if you think the event loop should be changed to event callbacks since RGFW now supports both.


raysan5:
@ColleagueRiley Thanks for the update!

PR #4148, opened 2024-07-09
[Shapes] Remove duplicate color calls in DrawGrid

JeffM2501:
Draw Gird called rlColor3f 4 times in a row, that's not needed since it's all batched. This PR cleans that up


raysan5:

PR #4150, opened 2024-07-10
[build.zig] GLFW Platform Detection Support

InventorXtreme:
While tinkering with Zig and raylib, I noticed that the Zig-compiled version of raylib was running through XWayland. I determined that the build.zig file would, by default, only support X11, and could be manually switched to Wayland, but could not use both via runtime platform detection. 

I resolved this problem by adding a new value, "Both" to the LinuxDisplayBackend enum and replacing the switch statement that was responsible for including the libraries with a set of if statements. This change allows for both backends to run off of one executable compiled through Zig.  I also changed the default support mode from X11 only to both.

I tested the changes on my machine and was able to run the same binary natively in both Wayland and X11.

Apologies if anything is funky with my submission, this is my first time contributing code to another project.

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -117,33 +117,30 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
 
                 raylib.addLibraryPath(.{ .cwd_relative = "/usr/lib" });
                 raylib.addIncludePath(.{ .cwd_relative = "/usr/include" });
-				
-				if (options.linux_display_backend == .X11 or options.linux_display_backend == .Both) {
+                if (options.linux_display_backend == .X11 or options.linux_display_backend == .Both) {
 
                         raylib.defineCMacro("_GLFW_X11", null);
                         raylib.linkSystemLibrary("X11");
-				}
+                }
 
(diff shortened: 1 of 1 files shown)


Gamer-Kold:
This looks good to me on initial glance but I'll probably wanna try building this myself to see if this works on my machine.

A little bit surprised this is all it takes for runtime detection of display protocol. Does it automatically fall back to X11 if Wayland support is enabled?

InventorXtreme:
It seems that as long as both X11 and Wayland are enabled for glfw and the appropriate libraries are linked, it figures it out. I roughly copied this from the make file, which I knew had worked in the past so I think this is all it takes.

If the build is compiled with the "both" option, it will automatically fall back to X11, but if it is only compiled with Wayland support, it will not be able to fall back because the X11 support will not be enabled and the libraries will not be linked. I tested the same executable compiled with the new "both" option and it ran natively on both platforms without a recompile. The user can of course still choose to support only one platform if they would prefer.

Gamer-Kold:
Tried it; got the same executable working on both X11 and Wayland natively. This looks good to me.

raysan5:
@InventorXtreme thanks for the improvement! 

@Gamer-Kold thanks for the review!

PR #4154, opened 2024-07-11
[rlgl] `rlGetShaderBufferSize` seems to always return 0

kal39:
### Description

The `rlGetShaderBufferSize` function seems to always return 0, even after uploading, processing and downloading data with an SSBO (all successfully). Looking at the code, the function is implemented using `glGetInteger64v`:

```c
// Get SSBO buffer size
unsigned int rlGetShaderBufferSize(unsigned int id) {
    long long size = 0;
    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id);
    glGetInteger64v(GL_SHADER_STORAGE_BUFFER_SIZE, &size);
    return (size > 0)? (unsigned int)size : 0;
}
```

After looking around a bit, it looks like `glGetBufferParameteri64v` is probably the correct function to be calling in order to get the size of the SSBO, see below:

```c
// Get SSBO buffer size
unsigned int rlGetShaderBufferSize(unsigned int id) {
    long long size = 0;
    glBindBuffer(GL_SHADER_STORAGE_BUFFER, id);
    glGetBufferParameteri64v(GL_SHADER_STORAGE_BUFFER, GL_BUFFER_SIZE, &size);
    return (size > 0)? (unsigned int)size : 0;
}
```

With this implementation, the actual size of the buffer is correctly returned. From what I can tell, `glGetInteger64v` is used to query the maximum value of some property (I'm not exactly sure). Is the current implementation the intended behaviour, or should it actually return the current size of the buffer?

### Environment

Windows 11, desktop:

```
INFO:     > Renderer: NVIDIA GeForce RTX 3060 Laptop GPU/PCIe/SSE2
INFO:     > Version:  4.3.0 NVIDIA 546.30
INFO:     > GLSL:     4.30 NVIDIA via Cg compiler
```

### Code Example

```c
#include <stdio.h>
#include <stdlib.h>

#include "raylib.h"
#include "rlgl.h"

int main(void) {
    InitWindow(800, 600, "pathtracer");

    unsigned int target_size = 100;
    unsigned int buffer = rlLoadShaderBuffer(target_size, NULL, RL_DYNAMIC_COPY);

    unsigned int actual_size = rlGetShaderBufferSize(buffer);
    printf("target size: %u, actual size: %u\n", target_size, actual_size);

    rlUnloadShaderBuffer(buffer);
    CloseWindow();
}
```

With the `glGetInteger64v` implementation, the code always outputs `target size: 100, actual size: 0`, but with the `glGetBufferParameteri64v` implementation, it outputs `target size: 100, actual size: 100`.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -4401,7 +4401,7 @@ unsigned int rlGetShaderBufferSize(unsigned int id)
 
 #if defined(GRAPHICS_API_OPENGL_43)
     glBindBuffer(GL_SHADER_STORAGE_BUFFER, id);
-    glGetInteger64v(GL_SHADER_STORAGE_BUFFER_SIZE, &size);
+    glGetBufferParameteri64v(GL_SHADER_STORAGE_BUFFER, GL_BUFFER_SIZE, &size);
 #endif
 
     return (size > 0)? (unsigned int)size : 0;


raysan5:
@kal39 Good catch! Thanks for the improvement! Actually, it seems previous implementation required `glGetInteger64i_v()` instead of `glGetInteger64v()` as per [the specs](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGet.xhtml).

PR #4156, opened 2024-07-12
[Core] Add GetRandomFloat to API

REDl3east:
I've had to implement this function for my use cases many times in the past and I think it would be beneficial to add it along side GetRandomValue. I also went ahead and updated the GetRandomValue example to include an example of the GetRandomFloat in use.


raysan5:
@REDl3east I prefer to avoid adding this function, it can be implemented on user side with provided ones.

Issue #4019, opened 2024-05-30
[rcore] KeyboardKey only supports American keyboard 


PizzaGame01:
I'm using a Europe (German) keyboard.
Its happend in [Raylib-cs](https://www.nuget.org/packages/Raylib-cs), but becaus its a port of Raylib i think this problem is still in original Raylib.

If i do this:
```cs
if (Raylib.IsKeyDown(KeyboardKey.Z))
{
    Console.WriteLine("Called!");
}
```
Its happen nothing by pressing Z, but if i press Y the if statement is executing.

This problem is also by ``int key = Raylib.GetKeyPressed();``

```cs
int key = Raylib.GetKeyPressed();
Console.WriteLine((KeyboardKey)key).ToString());
```

Clicked: Y
Output: Z

MrScautHD:
I can agree.

JupiterRider:
I think this is behavior intended. See https://github.com/raysan5/raylib/issues/2822

raysan5:
@PizzaGame01 I'm afraid there is not an easy solution to this problem, the keyboard mapping should be managed at GLFW, that raylib uses or at user side. Do you know of another possible solution?

raysan5:
If anyone finds this issue and knows of a simple multi-platform solution to address it, please, comment

MrScautHD:
> @PizzaGame01 I'm afraid there is not an easy solution to this problem, the keyboard mapping should be managed at GLFW, that raylib uses or at user side. Do you know of another possible solution?

has SDL 2 the same problem?

JeffM2501:
The problem statement in this issue is flawed, that is why there is confusion.

The keyboard key (KEY_) enum is not the keyname, it's the physical keycode. That enum was named after the American layout, because the enum has to be named something.

The enum can not and should not change based on keyboard layout. it represents physical buttons.
A FPS game on a dvorak keyboard would not use WASD, it'd use ,AEO, because they are in the same physical location.
A raylib game should not be expected to know the layout when checking for physical keys, it should simply check for KEY_W, KEY_A, KEY_D, KEY_S. 
Text input is mapped using GetCharPressed and is handled correctly.
What we are missing is a way to get the key name from the platform so that a game can show the mapped key name in things like settings and control bindings.

There should be an API where the user can ask "What is the name of KEY_W", and on dvorak it would return ',', or whatever the characater code is for the keyboard mapping.

[...]

PR #4161, opened 2024-07-13
Adding GetKeyName(int key)

MrScautHD:
Fixing this issue: https://github.com/raysan5/raylib/issues/4019


raysan5:
@MrScautHD Thanks for the addition!

PR #4163, opened 2024-07-15
Update BINDINGS.md: dray binding supports raylib 5.0

redthing1:


raysan5:
@redthing1 Thanks for the update!

PR #4164, opened 2024-07-15
[rcore] `InitWindow()` should check `InitPlatform()` returned value

SuperUserNameMan:
In case `PlatformInit()` fails, `InitWindow()` should quit properly and set `CORE.Window.ready = false`.

Also fix `rcore_template.c` which returned `false` instead of `-1`, as `false` might be defined to `0` in `raylib.h`.

Note that if the user forget to check `IsWindowReady()` before continuing, the program might still crash in a segfault if `InitPlatform()` failed.

@raysan5 : could we change `void InitWindow()` to `bool InitWindow()` instead of hoping the user think about checking `IsWindowReady()` in case of `InitPlatform()` failure ?

### DIFF
--- a/src/platforms/rcore_template.c
+++ b/src/platforms/rcore_template.c
@@ -476,7 +476,7 @@ int InitPlatform(void)
     if (platform.device == EGL_NO_DISPLAY)
     {
         TRACELOG(LOG_WARNING, "DISPLAY: Failed to initialize EGL device");
-        return false;
+        return -1;
     }
 
     // Initialize the EGL device connection
@@ -484,7 +484,7 @@ int InitPlatform(void)
     {
         // If all of the calls to eglInitialize returned EGL_FALSE then an error has occurred.
         TRACELOG(LOG_WARNING, "DISPLAY: Failed to initialize EGL device");
-        return false;
+        return -1;
     }
 
     // Get an appropriate EGL framebuffer configuration
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -631,10 +631,10 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    if ( InitPlatform() != 0 )
+    int result = InitPlatform();
+    if (result == -1 )
     {
         TRACELOG(LOG_ERROR, "Platform backend: Window initialization failed.");
-        CORE.Window.ready = false;
         return;
     }
     //--------------------------------------------------------------
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -631,7 +631,12 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    InitPlatform();
+    if ( InitPlatform() != 0 )
+    {
+        TRACELOG(LOG_ERROR, "Platform backend: Window initialization failed.");
+        CORE.Window.ready = false;
+        return;
+    }
     //--------------------------------------------------------------
 
     // Initialize rlgl default data (buffers and shaders)
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -632,7 +632,7 @@ void InitWindow(int width, int height, const char *title)
     // Initialize platform
     //--------------------------------------------------------------
     int result = InitPlatform();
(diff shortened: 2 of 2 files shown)


src/rcore.c
```diff
@@ -631,7 +631,12 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    InitPlatform();
+    if ( InitPlatform() != 0 )
```
raysan5, line 634: I try to avoid this kind of structure on raylib, also, not following conventions, raylib does not place spaces between parenthesis. In any case it should be:
```c
int result = InitPlatform();
if (result == -1)  // Platform initialization failed
{
    TRACELOG(LOG_WARNING, "WINDOW: Platform initialization failed");
    //CORE.Window.ready = false;  // If platform has not been correctly initialized, this should be already "false"
    return;  // In araylib I try to avoid/minimize early returns
}
```

src/platforms/rcore_template.c
```diff
@@ -476,15 +476,15 @@ int InitPlatform(void)
     if (platform.device == EGL_NO_DISPLAY)
     {
         TRACELOG(LOG_WARNING, "DISPLAY: Failed to initialize EGL device");
-        return false;
+        return -1;
```
raysan5, line 479: Actually, afaik, there is no clear convention defined on the function return values when used for errors detection... I tried to avoid that approach for raylib but I can consider it...

raysan5:
@SuperUserNameMan There is no clear convention on the function return values when used for errors detection... Actually, I tried to avoid that approach for most raylib functions (depite other libraries like SDL use it). I can consider it but it will require analyzing afected functions and conventions more carefully. For example, for `InitWindow()`, I like the `void` return approach for simplicity, maybe error codes could be queried a-part (like OpenGL does)... or, in case, `InitWindow()` returned an `int` what other uses it could have, other than a pure error message, i.e. for future multi-window id support.

All those questions require a review and redesign of the raylib library conventions and I prefer to consider them carefully.

SuperUserNameMan:
> maybe error codes could be queried a-part (like OpenGL does).

Just for thinking, here is the funny bug i encountered lately with the `glfwGetError()` way of doing :

```C
    if (posX < monitorX) posX = monitorX;
    if (posY < monitorY) posY = monitorY;

    glfwSetWindowPos(platform.handle, posX, posY); 

    glfwMakeContextCurrent(platform.handle); 

    result = glfwGetError(NULL);

    if (result != GLFW_NO_ERROR)
    {
        TRACELOG(LOG_FATAL, "PLATFORM: Failed to initialize graphics device");
        glfwDestroyWindow(platform.handle);
        glfwTerminate();
        return -1;
    }
```

On MS Windows, MacOS and X11 it would work as expected. But on Wayland, it would fail and terminate.

As I did not expect `glfwSetWindowPos()` to fail for any reason, i did not even consider checking for `glfwGetError()` just after.
But on Wayland, `glfwSetWindowPos()` is not supported and triggers a GLFW error.

This error is minor and could be considered just as a warning that should not cause the app to terminate, but, when `glfwGetError()` is checked just after the `glfwMakeContextCurrent()`, the error triggered by `glfwSetWindowPos()` was still in there, and the app terminated.

All of that because :
1) i did not expect a so basic function to fail and trigger an error ;
2) i thought `glfwGetError()` was automatically cleared each time a new glfw function was called ;
3) i should have read the GLFW doc more carefully for each and every GLFW function i used.

---

What I mean is that this `glfwGetError()` way of a handling errors might be confusing and counter intuitive compared the rest of the Raylib API. 

Also, it would requires to document which function might trigger an error, and what error should we expect to find (like the GLFW doc does).

---

Regarding what i initially suggested with returning a `bool` or a `int` to check the success of a function, i'm not sure it's better than the `glfwGetError()` way of doing neither. Because a lazy or hurry or over-intuitive programmer might still forget to check the returned value, or believe there is none returned since the compiler would not complain.

---

Another solution  I was thinking about would be something like that :

```C
void InitWhatever( int a, int b, int *error );
```

It would be impossible for the user to forget about the error as the compiler would complain about a missing argument.
Also, the `*error` could be cleared by each function,

PR #4170, opened 2024-07-18
[raymath] Type-Generic RayMath (tgraymath.h)

JayLCypher:
Tags: [TEACH] [CONVENIENCE]

TL;DR: Type-Generic RayMath function-like macros in the spirit of tgmath.h

Since the standard library has tgmath for type-generic math, thought it was only fitting to create type-generic raymath! 💯 

Probably not wanted or needed but offered as convenience. Allows omitting specific type for a bunch of Vector operations and some operations also supporting floats and numbers.

Most things are limited, but I decided to use defaults for the variants which also includes floats. This will allow usage of any number type, but I suspect -Wconversion will complain if someone aren't using compatible float types on float functions. Trying to set up _Generic for compatible type will literally explode the macros.
~~_Generic is ass.~~

For Raylib as a teaching tool, I suspect there are examples here of nested _Generics.

Includes a preprocessor warning in case someone tries to use the file in a project that is compiled with versions less than C11. (-Wcpp)

[Test file](https://gist.githubusercontent.com/JayLCypher/8ee7f5a5c33feae6e68d3799d374a929/raw/0eb53e0c8e8e06e2e213ae24710d2118e0978593/test_tgraymath.c)

### DIFF
--- a/src/tgraymath.h
+++ b/src/tgraymath.h
@@ -3,6 +3,10 @@
 #include "raymath.h"
 #endif
 
+#if __STDC_VERSION__ < 201112L
+#warning _Generic is a C11 directive and requires at least C11 compatibility.
+#endif
+
 void tgrarted(void);
 
 //----------------------------------------------------------------------------------
--- a/src/tgraymath.h
+++ b/src/tgraymath.h
@@ -1,13 +1,10 @@
-
-#ifndef RAYMATH_H
 #include "raymath.h"
-#endif
+
+void rl_tgempty(void);
 
 #if __STDC_VERSION__ < 201112L
 #warning _Generic is a C11 directive and requires at least C11 compatibility.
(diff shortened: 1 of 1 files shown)


raysan5:
@JayLCypher Interesting but I'm afraid that's out of scope for raylib, it should be moved to a separate repo.

JayLCypher:
Alright, I have created a repo for it just to have it up.
If you ever set raymath up as it's own thing, feel free to include tgraymath.h

Thanks for looking at it so quick Ray!

raysan5:
@JayLCypher Thanks for your comprehension. Please, could you share the repo for reference?

JayLCypher:
[tgraymath.h repository](https://github.com/JayLCypher/tgraymath.h)

PR #4173, opened 2024-07-20
[raudio] Improve AudioCallback (API breaking)

RealBitdancer:
**Changes in this PR**

- **AudioCallback Signature Update:** Changes the signature of `AudioCallback` to return the number of frames processed by the callback. This is useful when `AudioStream` looping is disabled, and fewer than the requested frames have been copied to the output buffer. Additionally, a `void* userData` argument has been added to associate custom data with the `AudioStream`, simplifying the creation of custom audio data loaders.

- **API Signature Changes:** The signatures of `SetAudioStreamCallback`, `AttachAudioStreamProcessor`, and `AttachAudioMixedProcessor` have been updated to accept an additional `void* userData` argument, which is forwarded to the corresponding callback.

- **New Function to enable/disable looping:** A new function, `SetAudioStreamLooping`, has been added to enable toggling looping on and off.

- **Example Updates:** Examples using the mentioned API functions have been updated to reflect the new signatures.

### DIFF
--- a/examples/audio/audio_mixed_processor.c
+++ b/examples/audio/audio_mixed_processor.c
@@ -21,7 +21,7 @@ static float averageVolume[400] = { 0.0f };   // Average volume history
 //------------------------------------------------------------------------------------
 // Audio processing function
 //------------------------------------------------------------------------------------
-void ProcessAudio(void *buffer, unsigned int frames)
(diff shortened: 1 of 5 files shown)


raysan5:
@RealBitdancer Thanks for the effort but no plans to break the API at the moment.

PR #4177, opened 2024-07-22
A better default that saves the whopping amount of 28KB on the final …

lzralbu:
…bundle.

The users that really need the [features emulated by `-sFULL_ES3`](https://emscripten.org/docs/porting/multimedia_and_graphics/OpenGL-support.html#opengl-es-2-0-3-0-emulation) can add the flag to their build without impacting all other raylib users.


raysan5:
@lzralbu Nice improvement! Thanks! I actually put a lot of care on OpenGL ES 2.0 usage to minimize emulation-paths when compiled to Web.

PR #4178, opened 2024-07-23
Add default vertex/fragment shader to OpenGL ES 3.0 based on the ones…

lzralbu:
… from OpenGL 3.3.

I was getting errors when using raylib's default shaders on WebGL 2 due to version mismatch and this PR fixes it.


raysan5:
@lzralbu thanks for the improvement! Afaik, the OpenGL ES 3.0 / WebGL 2 has been barely tested... testing, improvements and feedback are really welcome!

PR #4179, opened 2024-07-24
Warning on invalid image data

Jutastre:
Added some log warnings.
LoadImage will call with NULL fileType if file extension is missing.


raysan5:
@Jutastre thanks for the review!

PR #4182, opened 2024-07-26
[build] Add build flags for optional image format support in build.zig

nat3Github:
problem: 
no option to easily enable file format support like .svg, .jpg... in the build.zig options.

solution: 
add options like support_fileformat_bmp to build.zig. 
these default to false but when set to true the compile step will define the respective macro.

positive : 
user has the option to select image file format support in zig build

negative : 
if default options change one must also revisit the build.zig and apply the respective changes.


raysan5:
@nat3Github Adding those flags would imply adding the rest of `config.h` flags, I prefer to avoid it.

PR #4184, opened 2024-07-26
[rcore] fix gamepad axis movement and its automation event recording

maxmutant:
While working with Automation Events, I recognized that my gamepad input was not recorded correctly. Negative axis movements (e.g. stick going left/up) weren't recorded at all.

Furthermore, `GetGamepadAxisMovement` seems to snap trigger axis values to 0.0 if they are in the region [-0.1, 0.1]. I think the delta/drift check was intended to ignore values in between the range [-1.0, -0.9] instead (since -1.0 is the default state for triggers).

This PR addresses both of these issues.

**Remark:**
With my changes `GetGamepadAxisMovement` always returns the default value for each axis, even if a gamepad is not attached. This makes more sense in my opinion, but it is a behavior change for triggers.
Please let me know if you don't want that, and I'll change it back to always return 0.0 if the gamepad isn't available.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2949,13 +2949,22 @@ int GetGamepadAxisCount(int gamepad)
     return CORE.Input.Gamepad.axisCount[gamepad];
 }
 
+static float GetGamepadAxisMovementDefault(int axis)
+{
+    return (axis == GAMEPAD_AXIS_LEFT_TRIGGER || axis == GAMEPAD_AXIS_RIGHT_TRIGGER) ? -1.0f : 0.0f;
+}
+
 // Get axis movement vector for a gamepad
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -2957,13 +2957,22 @@ int GetGamepadAxisCount(int gamepad)
     return CORE.Input.Gamepad.axisCount[gamepad];
 }
 
+static float GetGamepadAxisMovementDefault(int axis)
```
raysan5, line 2960: I prefer to avoid adding new functions. Please, could you use that line directly on target function?

maxmutant: Sure, no problem. Inlined it.

raysan5:
@maxmutant Thanks for the improvement! :)

PR #4189, opened 2024-07-27
[GLFW] No longer set the RL_TEXTURE_FILTER_LINEAR when high dpi flag is enabled

SoloByte:
I am not sure why this was done in the first place. If I am missing something here and there is a good reason for doing this, please tell me.

Right now when the font texture filter is set to linear when the high dpi flag is enabled the default font looks very blurry.
The screenshots were made on a 4k monitor on Windows 11 with the `HIGH_DPI `flag enabled.

## Linear Filter Enabled / Before I changed it (Blurry)
raylib-default-font-linear-filter-enabled

## Linear Filter Disabled / How it looks now with the changes (Not Blurry)
raylib-default-font-linear-filter-disabled

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -668,15 +668,7 @@ void InitWindow(int width, int height, const char *title)
     SetShapesTexture(texture, (Rectangle){ 0.0f, 0.0f, 1.0f, 1.0f });    // WARNING: Module required: rshapes
     #endif
 #endif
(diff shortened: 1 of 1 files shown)


raysan5:
@SoloByte I'm merging it because it removes code and I like it but note that visual quality is relative to users, afair, the filter was added because some users preferred the slightly blurry text than the irregular nearest-neighbourgh scaling mode.

SoloByte:
Thanks, okay good to know. But it is still possible to set the default font filter to linear if the effect is wanted, correct ?

PR #4190, opened 2024-07-27
[rcore][desktop_glfw] Keeping CORE.Window.position properly in sync with glfw window position

SoloByte:
CORE.Window.position is now properly kept in sync with the glfw window position.
- WindowPosCallback added
- SetWindowPos now updates CORE.Window.position as well
- Fullscreen & Borderless Fullscreen now properly user CORE.Window.position for setting CORE.Window.prevposition

Those changes should not change any behavior other than fixing `ToggleFullscreen()` not restoring the window to the correct previous window position. 

---

FYI: There will be another PR for fixing the remaining problems with fullscreen & borderless fullscreen
- restoring window size correctly and using already existing CORE.Window.prevscreen for it
- BorderlessFullscreen using glfw SetWindowMonitor function as well to fix borderless fullscreen on macOS
- some small misc fixes.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -618,8 +618,9 @@ void SetWindowMonitor(int monitor)
         {
             TRACELOG(LOG_INFO, "GLFW: Selected monitor: [%i] %s", monitor, glfwGetMonitorName(monitors[monitor]));
 
-            const int screenWidth = CORE.Window.screen.width;
-            const int screenHeight = CORE.Window.screen.height;
+            // Here the render width has to be used again in case high dpi flag is enabled
+            const int screenWidth = CORE.Window.render.width;
+            const int screenHeight = CORE.Window.render.height;
             int monitorWorkareaX = 0;
             int monitorWorkareaY = 0;
             int monitorWorkareaWidth = 0;
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1572,12 +1572,17 @@ int InitPlatform(void)
         int monitorWidth = 0;
         int monitorHeight = 0;
         glfwGetMonitorWorkarea(monitor, &monitorX, &monitorY, &monitorWidth, &monitorHeight);
-
-        int posX = monitorX + (monitorWidth - (int)CORE.Window.screen.width)/2;
-        int posY = monitorY + (monitorHeight - (int)CORE.Window.screen.height)/2;
+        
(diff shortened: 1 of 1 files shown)


casperbear:
There is a small issue in RayLib 5 when calling ToggleFullscreen() changes the angle Camera3D looks at. (And calling ToggleFullscreen() again changes the angle back). I wonder if this fix fixes that issue as well.

casperbear:
@raysan5 @SoloByte With the raylibdll.lib built from https://github.com/SoloByte/raylib/tree/glfw-window-pos-callback the bug with Camera3D angle jumping when switching between resolutions is indeed fixed. I tested it a 3d project. So this PR can be considered as a bug fix, not a just-in-case improvement. The relevant code I tested it with:
```
inline constexpr int SCREEN_WIDTH = 1920;
inline constexpr int WINDOWED_SCREEN_HEIGHT = 1080;
inline constexpr int FULL_SCREEN_HEIGHT = 1200;

if ((IsKeyDown(KEY_LEFT_ALT) || IsKeyDown(KEY_RIGHT_ALT)) && (IsKeyPressed(KEY_ENTER) || IsKeyPressed(KEY_KP_ENTER)))
        {
            if (IsWindowFullscreen())
            {
                ToggleFullscreen();
                screenWidth = SCREEN_WIDTH;
                screenHeight = WINDOWED_SCREEN_HEIGHT;
                SetWindowSize(screenWidth, screenHeight);
            }
            else
            {
                screenWidth = SCREEN_WIDTH;
                screenHeight = FULL_SCREEN_HEIGHT;
                SetWindowSize(screenWidth, screenHeight);
                ToggleFullscreen();
            }
        }
```

SoloByte:
Thanks for testing this :) I did not think this would fix something right away ^^

SoloByte:
The new commit should set `CORE.Window.position` correctly from the start in `InitPlatform()`. I also had to fix the code for centering the window on start-up when the high dpi flag is enabled so that `CORE.Window.position` would be set correctly as well.

@casperbear You could recheck it, it should not have changed anything (hopefully).

casperbear:
@SoloByte yes, Camera3D angle is still preserved when changing from windowed to fullscreen and back with RayLib from commit https://github.com/raysan5/raylib/pull/4190/commits/7bd4888a9dbbe0c42b7b85240ec201a391515ee4
In both cases: when windowed window is centered on the desktop and not centered on the desktop.
My setup is Windows 7 64-bit and one 1920x1200 monitor.

raysan5:
@SoloByte The improvement looks good to me, thanks for the review and sorry for the late review.

SoloByte:
> @SoloByte The improvement looks good to me, thanks for the review and sorry for the late review.

Thanks, and no worries because of the delay :)

Issue #4185, opened 2024-07-26
[rtextures] `LoadImageSvg()` causes heap buffer overflow when reading from file


spartan-engi:
### Issue description

compiling raylib with address sanitizer (-fsanitize=address) and calling `LoadImageSvg()` accuses a "heap-buffer-overflow" error

### Environment

**Platform:** Desktop
**Operating System:** Windows 10
**OpenGL:**
  Vendor:   Intel
  Renderer: Intel(R) HD Graphics 3000
  Version:  3.1.0 - Build 9.17.10.4459
  GLSL:     1.40 - Intel Build 9.17.10.4459

### Issue Screenshot

![image](https://github.com/user-attachments/assets/b9f34645-3abf-443d-ae85-eb85bb245966)

### Code Example

when compiling with sanitizer, even the `textures_svg_loading.c` example gives out this error
but this is smallest program that also triggers it is:
````
#include "raylib.h"
int main(void)
{
    Image test = LoadImageSvg("examples/textures/resources/test.svg", 2500, 2484);
    return 0;
}
````

---

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

spartan-engi:
After a bit more investigation, `nanosvg` expects the input to `nsvgParse()` to be a NULL terminated string,
and it seems that `LoadFileData()` doesn't insert a EOF character

Thus, the function simply reads past the given input buffer, onto some random heap memory
and i guess most of the time that is '\0' ?

Honestly i have no idea how this function appears to work normally

Though, it might be the cause of some very infrequent crashes

CrackedPixel:
@spartan-engi 
Are you still able to reproduce this with the latest `master` branch? (raylib recompiled with svg support)

raysan5:
There is a related open PR, it looks a bit cumbersome to me but I can't think of a better solution... I will continue the discussion there.

PR #4191, opened 2024-07-27
[rtextures] fix buffer overflow on loading .svg file

spartan-engi:
simple fix for #4185 

Makes sure the input to `nsvgParse()` function is NULL terminated, to avoid buffer overflow

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -353,6 +353,14 @@ Image LoadImageSvg(const char *fileNameOrString, int width, int height)
         {
             fileData = LoadFileData(fileNameOrString, &dataSize);
             isSvgStringValid = true;
+
+            // make sure that fileData is a NULL terminated String
+            // as required by nsvgParse()
+            if (fileData[dataSize-1] != '\0')
+            {
+                fileData = RL_REALLOC(fileData, dataSize+1);
+                fileData[dataSize] = '\0';
+            }
         }
         else
         {
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -291,11 +291,21 @@ Image LoadImage(const char *fileName)
     // Loading file to memory
     int dataSize = 0;
     unsigned char *fileData = LoadFileData(fileName, &dataSize);
+    const char *fileType = GetFileExtension(fileName);
+
+    #if defined(SUPPORT_FILEFORMAT_SVG)
+    if ((strcmp(fileType, ".svg") == 0) || (strcmp(fileType, ".SVG") == 0))
+    {
+        fileData = RL_REALLOC(fileData, dataSize+1);
+        fileData[dataSize] = '\0';
+        dataSize++;
+    }
+    #endif
 
     // Loading image from memory data
     if (fileData != NULL)
     {
-        image = LoadImageFromMemory(GetFileExtension(fileName), fileData, dataSize);
+        image = LoadImageFromMemory(fileType, fileData, dataSize);
 
         UnloadFileData(fileData);
     }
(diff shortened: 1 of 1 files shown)


planetis-m:
Not OPs fault but this function has terrible API. `fileNameOrString` ?? There should be two separate functions.

spartan-engi:
> Any reason to not just use `nsvgParseFromFile`?

Given the existence of the custom callback function for loading files (`LoadFileDataCallback`), and that that `nsvgParseFromFile` wasn't used from the beginning
I presumed the idea was to centralize file I/O to the internal functions
So if the user wants to do something weird with I/O, or for some reason raylib I/O gets updated, it's easier to change

> Not OPs fault but this function has terrible API. `fileNameOrString` ?? There should be two separate functions.

I could understand the idea behind having only one function for parsing both file and raw string of SVG,
It wouldn't make things terribly more complicated, and is rather elegant

But given this thing is something more educational, this is probably a bad idea, obscures the operation

The functionality should be broken into two functions, to keep parity with `LoadImage` and `LoadImageFromMemory`
Thought i feel that that might be out of scope for this PR?

spartan-engi:
I just realized `LoadImage` is also able to load SVGs,
verified that it had the same problem and extended the fix for it.

The fact that the generic function can also load SVGs does explain why `LoadImageSvg` is the way it is,
it is just the alternative if you need a specific scale, thus, it is compacted to avoid clutter

raysan5:
@spartan-engi thanks for the review. About the specific function, note that raylib API has no file-format-specific function.

src/rtextures.c
```diff
@@ -291,11 +291,21 @@ Image LoadImage(const char *fileName)
     // Loading file to memory
     int dataSize = 0;
     unsigned char *fileData = LoadFileData(fileName, &dataSize);
+    const char *fileType = GetFileExtension(fileName);
+
+    #if defined(SUPPORT_FILEFORMAT_SVG)
```
raysan5, line 296: I prefer to avoid this hack to consider a specific format situation, it breaks the funtion symmetry and organization.

src/rtextures.c
```diff
@@ -353,6 +363,14 @@ Image LoadImageSvg(const char *fileNameOrString, int width, int height)
         {
             fileData = LoadFileData(fileNameOrString, &dataSize);
             isSvgStringValid = true;
+
+            // make sure that fileData is a NULL terminated String
+            // as required by nsvgParse()
+            if (fileData[dataSize-1] != '\0')
```
raysan5, line 369: What are the changes that an Svg string does not finish with ' \0' ?

raysan5:
@spartan-engi I'm cosnidering removing SVG support all together...

src/rtextures.c
spartan-engi: A svg that is pre compiled into the program
aka: `char* fileSVG = "<svg`...
will have absolutely no problem, as internal strings are automatically NULL terminated by the compiler

however, actual files, not so much.
I managed to test Inkscape and Krita, both open source
neither outputs files without an end byte

Given that .svg is technically a XML file type, I don't think a EOF byte is in the specification

src/rtextures.c
spartan-engi: no problem, I can see a few options to this.

1. I'm not sure if this would break symmetry as well, but moving all of SVG functionality  from `LoadImage` `LoadImageFromMemory` to `LoadImageSvg` would be very simple

2. nanosvg.h could be altered, so it's functions take a file_data and length parameter instead of relying on the data string being formatted.
This gets rid of the `realloc()` calls, and makes the functions safer in my opinion.

3. `LoadImage` can be made symmetric again by pushing the buffer re allocation call to `LoadImageFromMemory`.
Originally I avoided doing this, as a non-allocated string could be passed to `LoadImageFromMemory`, and trying to reallocate that would exit the program very 'inelegantly',
But an argument can be made that the chance of a non-allocated string somehow being not NULL terminated is very low

madelfagam:
> @spartan-engi I'm cosnidering removing SVG support all together...

Oh, no. I would ask to add support for processing animated SVGs, but there is still a mess with the implementation standards :D

src/rtextures.c
raysan5: Very good alternatives list! Despite I don't like changing external libraries I think the best option for raylib is 2.

My other alternative to avoid this hack is just removing SVG support, users with this requirement could just implement it themselfs, actually an example could be added in `raylib-extras`.

`LoadImageSvg()` is really and UGLY function and afaik, the only function in all the library file-format specific, that is something I tried to avoid in raylib...

Definitely, moving SVG support to an example seems the best solution to me...

spartan-engi: On trying to apply option 2, a small problem appeared

The issue is fixed with `LoadImage()`, `LoadImageFromMemory()`,
and `LoadImageSvg()` behaves nicely when loading from files

But `LoadImageSvg()` from a string still has the issue that the input needs to be a NULL terminated string. Since the function has no parameter for the size of the string

The tweak does 'shrink' the issue down to only one use case though:
`LoadImageSvg(LoadFileData())`

raysan5:
I decided to remove SVG support from raylib and move the required code to a separate example, the `LoadImageSvg()` API is ugly (and the only format-dependant function in the raylib API); also, as this issue states, it could generate issues requiring ugly hacks. Also, it adds extra maintenance burden. And also, it's the only format that requries several external libraries to work. Also, the format is disabled by default so most bindings probably don't care about it.

For all those reasons I'm moving the functionality and the example `raylib-extras`. I really think there is no need to be in raylib.

PR #4196, opened 2024-07-30
replace unicode characters by ascii characters in jar_xml

RicoP:
MSVC keeps giving me the warning

`jar_xm.h(1404): warning C4819: The file contains a character that cannot be represented in the current code page (932). Save the file in Unicode format to prevent data loss`

The issue where the unicode characters. I replaced them with ascii characters to avoid the warning.


raysan5:
@RicoP Good review! Thanks!

PR #4197, opened 2024-07-30
[rlgl] use GLint64 for glGetBufferParameteri64v

rnpnr:
Hi Ray,

When compiling raylib on Linux with `OPENGL_VERSION=4.3` the build fails because `GLint64` is defined as `long int` which is not the same type as `long long`. This will fix the issue while maintaining compatibility with Win32 where `long long` is equivalent to `GLint64`.


raysan5:
@rnpnr Thansk for the review!

PR #4202, opened 2024-08-01
[rlgl] Adding warnings in case OpenGL 4.3 is not enabled

MaximKn1:
Hi
I added warnings for the functionality related to OpenGL 4.3 features in case GRAPHICS_API_OPENGL_43 is not defined. 

Personally, been struggling to make compute shaders work, only to discover that I am not on the correct API version (as the default one is 3.3). I think it is better to add some logging to notify about incorrect OpenGL version. Currently 4.3 related functions are virtually empty if you are not using 4.3 and you can call them even if you are on 3.3. Shader compiler even says it managed to compile a compute shader when using 3.3.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -4083,6 +4083,8 @@ unsigned int rlCompileShader(const char *shaderCode, int type)
             //case GL_GEOMETRY_SHADER:
         #if defined(GRAPHICS_API_OPENGL_43)
             case GL_COMPUTE_SHADER: TRACELOG(RL_LOG_WARNING, "SHADER: [ID %i] Failed to compile compute shader code", shader); break;
+        #else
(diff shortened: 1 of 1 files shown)


raysan5:
@MaximKn1 Some of those warnings could spam the log if the functions are called every frame, please, remove the ones from functions that can potentially be called on a frame basis.

MaximKn1:
@raysan5 Removed logging as you suggested, except rlBindImageTexture, because as far as I can tell, there is no way to tell if proper OpenGL version is used for this functionality before calling it.

raysan5:
@MaximKn1 Thanks for the review!

raysan5:
@MaximKn1 I'm afraid this PR breaks raylib compilation for web, please, could you review it?

PR #4203, opened 2024-08-01
More performant point cloud rendering with DrawModelPoints

satchelfrost:
Currently, raylib has DrawPoint3D which works well for most things, but has a few issues:

1. technically draws lines not points
2. starts to become less performant as point cloud size increases

Instead I added a two new functions DrawModelPoints and DrawModelPointsEx. They follow the same pattern as DrawModelWires and DrawModelWiresEx, but instead temporarily sets rlEnablePointMode and rlDisableBackfaceCulling (which it then reverts after calling DrawModel). The following example provided demonstrates the performance difference between DrawModelPoints and DrawPoint3D for various point cloud sizes.

With my card (6700xt) 1,000,000 points are easily rendered @60 FPS using DrawModelPoints, whereas the same example with DrawPoint3D is unusable < 10 FPS. Your mileage may vary depending on the graphics card, but overall DrawModelPoints should be more performant than DrawPoint3D.


raysan5:
@satchelfrost Thanks for the addition, please, note that new examples should follow examples template, including the rules and conventions: https://github.com/raysan5/raylib/blob/master/examples/examples_template.c

examples/models/models_point_rendering.c
```diff
@@ -0,0 +1,177 @@
...
+    Vector3 position = {0.0f, 0.0f, 0.0f};
+    bool useDrawModelPoints = true;
+    bool numPointsChanged = false;
+    int numPoints = 1000;
+    Mesh mesh = GenPoints(numPoints);
+    Model model = LoadModelFromMesh(mesh);
+    //--------------------------------------------------------------------------------------
+
+    // Main game loop
+    while(!WindowShouldClose()) {
```
raysan5, line 55: Not following raylib coding conventions

examples/models/models_point_rendering.c
```diff
@@ -0,0 +1,177 @@
...
+    // De-Initialization
+    //--------------------------------------------------------------------------------------
+    UnloadModel(model);
+    CloseWindow();
+    //--------------------------------------------------------------------------------------
+    return 0;
+}
+
+// Random float between 0 and 1
+float RandFloat()
```
raysan5, line 143: Please, could you remove this function and calculate it inline?

raysan5:
Please, note that the following files should be updated when adding a new example, it's planned to create some script to automatize part of this process but not available yet.

     - raylib/examples/<category>/<category>_example_name.c
     - raylib/examples/<category>/<category>_example_name.png
     - raylib/examples/<category>/resources/*.*
     - raylib/examples/Makefile
     - raylib/examples/Makefile.Web
     - raylib/examples/README.md
     - raylib/projects/VS2022/examples/<category>_example_name.vcxproj
     - raylib/projects/VS2022/raylib.sln
     - raylib.com/common/examples.js
     - raylib.com/examples/<category>/<category>_example_name.html
     - raylib.com/examples/<category>/<category>_example_name.data
     - raylib.com/examples/<category>/<category>_example_name.wasm
     - raylib.com/examples/<category>/<category>_example_name.js

satchelfrost:
Ah, sorry about that, I will make those changes.

On a separate note I did some more investigation with the point rendering. On my machine rendering 10,000,000 points, I noticed that when I zoomed in, the frame rate started to drop. Which doesn't really make any sense for a point cloud.

So I opened renderdoc and found that even though the points are being shown in the final image, you can actually see the triangles in wireframe in renderdoc.

using_rlEnablePointMode_renderdoc

As a test, instead of using rlEnablePointMode(), I modified rlDrawVertexArray() to use GL_POINTS. Visually there is no difference in the final image. But when I zoom in at 10,000,000 points there is no performance drop, and its a solid 60 FPS, and the renderdoc image shows only points being drawn.

using_gl_points_renderdoc

Also just to see how far I could push it on my machine I was able to get 100,000,000 points at roughly 50 FPS. Kind of looks like a blob though.

blob

So my question is given that it is even more performant to just use GL_POINTS instead of just rlEnablePointMode(), would you like me to modify the current code to use this method instead? Perhaps a new function rlDrawVertexArrayPoints()? The current method works and doesn't mess with the API too much, it's just not as performant as it could be.

raysan5:
@satchelfrost I'm not sure how to address this improvement, not only because it's platform-dependant but also because if `rlDrawVertexArray()` uses GL_POINTS internally, it would use x3 vertex for any model, considering the data provided are usually 3-vertex per triangle...

satchelfrost:
In that case, I will stick to the previous approach which is still more performant than DrawPoint3D. I've updated the makefiles as well. I still need to figure out how to generate the vcxproj file since I don't use visual studio and it looks like a simple copy paste from one of the other examples won't work because a guid needs to be generated (currently using Linux). If you're okay with the current example then I will make a request on the raylib.com repo for the web version.

raysan5:
@satchelfrost Thanks for the review! Don't worry about the VS project, I'll try to address it myself.

PR #4204, opened 2024-08-01
Update raymath.h

ruby-bang:
just change the name of the variable to suit its function

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -469,7 +469,7 @@ RMAPI Vector2 Vector2Rotate(Vector2 v, float angle)
 }
 
 // Move Vector towards target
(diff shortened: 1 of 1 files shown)


src/raymath.h
```diff
@@ -469,7 +469,7 @@ RMAPI Vector2 Vector2Rotate(Vector2 v, float angle)
 }
 
 // Move Vector towards target
-RMAPI Vector2 Vector2MoveTowards(Vector2 v, Vector2 target, float maxDistance)
+RMAPI Vector2 Vector2MoveTowards(Vector2 v, Vector2 target, float speed)
```
asdqwe, line 472: Hello,

I'm assuming this was submited by mistake, because the only thing this does is break `Vector2MoveTowards`.

Best regards.

PR #4206, opened 2024-08-02
[rtextures] Promote calculation to long long in GetPixelDataSize() to allow for larger image sizes

noname22:
Fix to allow working with image sizes >= 8192 x 8192 x 32bit by promoting calculation of image sizes to long long during calculation.
Without fix GetPixelDataSize() yields a negative result due to an int overflow with image sizes >= 8192 x 8192 x 32bit.
Fix allows for sizes up to ~23k x 23k x 32bit without overflowing the size int.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -5403,7 +5403,9 @@ int GetPixelDataSize(int width, int height, int format)
         default: break;
     }
 
-    dataSize = width*height*bpp/8;  // Total data size in bytes
+    // Total data size in bytes
+    // NOTE: promote to long long during calculation to avoid overflows before dividing by 8
+    dataSize = width*height*(long long)bpp/8;
 
     // Most compressed formats works on 4x4 blocks,
     // if texture is smaller, minimum dataSize is 8 or 16


raysan5:
@noname22 Thanks for the review but I actually prefer to address it in another way:
```c
char bytesPerPixel = bpp/8;
dataSize = width*height*bytesPerPixel;  // Total data size in bytes
```

raysan5:
@noname22 I just reviewed it, thanks for reporting!

noname22:
float (32bit) is not great for large integers, it can only describe integers up to 2^24 with precision, @raysan5
```
    float bytesPerPixel = (float)bpp/8.0f;
    dataSize = (int)(bytesPerPixel*width*height); // Total data size in bytes
```
With this implementation (the current in master), the expression gets promoted to float, since bytesPerPixel is a float. It suffers quite a lot from said precision issues.

I wrote a small test program to try different texture sizes and it found thousands of combinations where the current logic produces too small sizes.

Consider the following code:

```
#include <stdio.h>

{   
    int width = 4097;
    int height = 4097;
    int bpp = 32;

    int dataSize, dataSize2, dataSize3;
    
    // current implementation in master
    float bytesPerPixel = (float)bpp/8.0f;
    dataSize = (int)(bytesPerPixel*width*height); // Total data size in bytes

    // double implementation
    double bytesPerPixel2 = bpp/8.0;
    dataSize2 = (int)(bytesPerPixel2*width*height);
    
    // original PR implementation
    dataSize3 = width*height*(long long)bpp/8;
        
    printf("%d x %d x %dbpp\n", width, height, bpp);
    printf("size (current master): %d\n", dataSize);
    printf("size (double):         %d\n", dataSize2);
    printf("size (original PR):    %d\n", dataSize3);

    return 0;
}
```

which produces the output:

```
size (current master): 67141632
size (double):         67141636
size (original PR):    67141636
```

As you can see the float implementation is missing 4 bytes in the size for this 4097 x 4097 x 32bpp image.

Using a double instead of a float solves the issue, since that can accurately describe integer values up to 2^56.
My original PR which promotes the expression to a 64bit integer also works as intended.

noname22:
The safest way to do this would probably be a ceilinged double. It would guarantee that the buffer is *at least* the size it needs to be with esoteric texture formats, such as bpp 3 or 5 (or anything under 8).

ie.

```
    double bytesPerPixel = bpp/8.0;
    dataSize = (int)ceil(bytesPerPixel*width*height);
```

Imagine a 7 x 7 RL_PIXELFORMAT_COMPRESSED_PVRT_RGBA image. That's 7 x 7 x 4 bits, which is 196 bits or 24.5 bytes.

The code above produces a size of 25, which is a large enough buffer to contain the image.
All other proposed and implemented expressions will yield a size of 24, which is not large enough to hold the image.

asdqwe:
Hello,

Upvoting @noname22 's suggested double implementation:
```
    // double implementation
    double bytesPerPixel2 = bpp/8.0;
    dataSize2 = (int)(bytesPerPixel2*width*height);
```

Best regards.

PR #4209, opened 2024-08-04
update make for examples

CrackedPixel:
These were missing. Added in alphabetical order


raysan5:
@CrackedPixel Nice! Thanks for the review. Note that `core_basic_window_web` was kept out on pourpose because it's the same as `core_basic_window`

PR #4210, opened 2024-08-04
Changed raylib.h to improve tooltips in modern IDEs

KyleDickersonComposer:
### Tooltips fail to display correct information in VSCode and Nvim
- I moved the documentation comments in raylib.h to directly above a given function definition. 
- Doing so allows modern IDEs to display the correct information in tooltips.

### An example of incorrect information in Nvim and VSCode
- The ClearBackground function incorrectly displays the text "Drawing related functions".
- Should say: "Set background color (framebuffer clear color)"
incorrect-tooltip-Nvim
incorrect-tooltip-VSCode

### Using my updates to raylib.h
- The tooltips now display helpful information.
correct-tooltip-Nvim
correct-tooltip-VSCode

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -935,11 +935,16 @@ typedef enum {
 
 // Callbacks to hook some internal functions
 // WARNING: These callbacks are intended for advanced users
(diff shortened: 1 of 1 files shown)


raysan5:
@KyleDickersonComposer Thanks but no plans to change it. I find this version way longer and more difficult to read.

Issue #4208, opened 2024-08-04
[WIP] [Examples] Update progress for v5.5


CrackedPixel:
* All below Build/Works is for Desktop. Will probably add more columns to be explicit later
* **2024-08-03:** All builds are successful. Need to test if they work properly
* **2024-08-04** Tested Audio and Others
* **2024-08-04** Tested Text
* **2024-08-04** Tested Shapes
* **2024-08-04** Tested Models
* **2024-08-04** Tested Shaders
* **2024-08-05** Tested Textures
* **2024-08-05** Tested Core

## Audio
| Name                   | Build | Works | Notes              |
|------------------------|-------|-------|--------------------|
| audio_mixed_processor  |✅|✅| Visual "average volume" does not work. Fixed in https://github.com/raysan5/raylib/pull/4214 (merged)|
| audio_module_playing   |✅|✅|                    |
| audio_music_stream      |✅|✅|                    |
| audio_raw_stream           |✅|✅| PNG needs updated|
| audio_sound_loading     |✅|✅|                    |
| audio_sound_multi          |✅|✅|                    |
| audio_stream_effects     |✅|⚠️| Delay effect lowers volume -- intentional?|

## Core
| Name                           | Build | Works | Notes          |
|--------------------------------|-------|-------|--------------------|
| core_2d_camera                            |✅|✅|                    |
| core_2d_camera_mouse_zoom  |✅|✅| PNG needs updated|
| core_2d_camera_platformer       |✅|✅| Camera clamp to screen edge broken when going off map. I don't think it needs updated, though|
| core_2d_camera_split_screen     |✅|✅|                    |
| core_3d_camera_first_person     |✅|✅| Some camera modes do not support the controls shown on-screen. PNG needs updated. Being worked on in https://github.com/raysan5/raylib/pull/4221|
| core_3d_camera_free                   |✅|✅|                    |
| core_3d_camera_mode               |✅|✅|                    |
| core_3d_camera_split_screen     |✅|✅|                    |
| core_3d_picking                            |✅|✅| PNG needs updated. Maybe add something about how the red line is the ray? It's very confusing -- can discuss if needed|
| core_automation_events             |✅|✅|                    |
| core_basic_screen_manager       |✅|✅|                    |
| core_basic_window                       |✅|✅|                    |
| core_basic_window_web              |✅|✅|                    |
| core_custom_frame_control        |✅|✅|                    |
| core_custom_logging                   |✅|✅|                    |
[...]

PR #1470, opened 2020-12-25
Fix use of GuiSliderBar in examples.

badlydrawnrod:
Fixed a couple of examples as they weren't compiling. Happy Christmas!

PR #4211, opened 2024-08-05
[examples] Fix some examples

kai-z99:
Fix issues where settings values are not displaying on shapes_draw_circle_sector, shapes_draw_rectangle_rounded, and shapes_draw_ring. The issue also notes that the PNGs are rotated 90 deg clockwise compared to the examples which is true, although I believe the samples are accurate and the PNGs are not so I did not change it.

https://github.com/raysan5/raylib/issues/4208

### DIFF
--- a/examples/shapes/shapes_draw_circle_sector.c
+++ b/examples/shapes/shapes_draw_circle_sector.c
@@ -63,11 +63,11 @@ int main(void)
 
             // Draw GUI controls
             //------------------------------------------------------------------------------
-            GuiSliderBar((Rectangle){ 600, 40, 120, 20}, "StartAngle", NULL, &startAngle, 0, 720);
-            GuiSliderBar((Rectangle){ 600, 70, 120, 20}, "EndAngle", NULL, &endAngle, 0, 720);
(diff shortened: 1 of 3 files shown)


CrackedPixel:
It looks like this was removed by @badlydrawnrod in https://github.com/raysan5/raylib/pull/1470

Not sure why the values were set to `NULL`. Perhaps by mistake?

This PR seems good and puts it back how the original example worked. I agree with the PNG as well and think the PNG just needs updated

kai-z99:
> It looks like this was removed by @badlydrawnrod in #1470
> 
> Not sure why the values were set to `NULL`. Perhaps by mistake?
> 
> This PR seems good and puts it back how the original example worked. I agree with the PNG as well and think the PNG just needs updated

kai-z99:
Yeah that odd, It seems that the GUISliderBar() function was changed and took different parameters at that timne which broke it?

CrackedPixel (approved):

CrackedPixel:
> Yeah that odd, It seems that the GUISliderBar() function was changed and took different parameters at that timne which broke it?

I think it was just overlooked in #1470 when updating the example

raysan5:
@kai-z99 Thanks for the review!

PR #4212, opened 2024-08-05
[miniaudio] Fixing miniaudio and Makefile for NetBSD

NishiOwO:


src/Makefile
```diff
@@ -467,7 +467,7 @@ INCLUDE_PATHS = -I.
 ifeq ($(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW)
     INCLUDE_PATHS += -Iexternal/glfw/include
     ifeq ($(PLATFORM_OS),BSD)
-        INCLUDE_PATHS += -I/usr/local/include
+        INCLUDE_PATHS += -I/usr/local/include -I/usr/pkg/include -I/usr/X11R7/include
```
raysan5, line 470: Those paths look highly irregular, why are they needed? Is there no other option?

src/external/miniaudio.h
```diff
@@ -36078,9 +36078,15 @@ static ma_result ma_context_get_device_info_from_fd__audio4(ma_context* pContext
         ma_uint32 channels;
         ma_uint32 sampleRate;
 
+#ifdef __NetBSD__
```
raysan5, line 36081: Actually those changes should be on miniaudio side because they will be ovewritten when updating miniaudio...

NishiOwO: already merged into dev branch of miniaudio, so it is for until we get new release

src/Makefile
NishiOwO: because netbsd installs packages under /usr/pkg

raysan5:
Merging! Note miniaudio will be override on next update!

PR #4213, opened 2024-08-05
[rtext] add ExportFontAsCodeEx()

rnpnr:
Hi All,

This PR introduces a function to export font as code without ever opening a window. This allows me to include the required font file with my project's source code and export it as an automatically generated header from a build script on a system without a GPU present.

Any comments/review is appreciated!

Commit message follows:

ExportFontAsCode() loads image data from a font that is already uploaded to the GPU; this means that it requires a full graphics context/window. ExportFontAsCodeEx() loads the relevant font data into an atlas itself without ever creating a texture. This makes it possible to export font data as part of a build step when compiling a project when there is no GPU present (e.g. github CI).

One minor change is that ExportFontAsCode() no longer converts the name to PascalCase. Its up to the user to pass the name in in that format if they want it.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1452,6 +1452,7 @@ RLAPI Image GenImageFontAtlas(const GlyphInfo *glyphs, Rectangle **glyphRecs, in
 RLAPI void UnloadFontData(GlyphInfo *glyphs, int glyphCount);                               // Unload font chars info data (RAM)
 RLAPI void UnloadFont(Font font);                                                           // Unload font from GPU memory (VRAM)
 RLAPI bool ExportFontAsCode(Font font, const char *fileName);                               // Export font as code file, returns true on success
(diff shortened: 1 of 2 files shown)


raysan5:
@rnpnr I'm afraid this function is intended for a very specific use case but it breaks the consistency with the other `Export*()` and `Export*AsCode()` functions; It also requires a considerable amount of input parameters. I also adds static functions and functions calling functions, something that I try to minimize in raylib.

I'm afraid I'm not merging this PR.

rnpnr:
Hi @raysan5, I understand. I'm happy to work on something you would find more acceptable. I just want a way to export a font without opening a window/initializing a graphics context. Perhaps a way of loading a `Font` without uploading it to the GPU? `IsFontReady()` already accounts for the `Font` missing a valid texture id.

PR #4214, opened 2024-08-06
Update audio mixed processor

CrackedPixel:
This fixes the display issue. The average volume is a float such as 0.03, so the cast to int was making it always 0 and having no height


raysan5:
@CrackedPixel thanks for the improvement!

PR #4217, opened 2024-08-06
[build.zig] X11-only systems fail builds by default

lnc3l0t:
- this PR references the https://github.com/raysan5/raylib/pull/4150#issuecomment-2271084090

I added a check to find the program `wayland-scanner`, if it's not found the following error log is printed and the build fails:
```
 Wayland may not be installed on the system.
 You can switch to X11 in your `build.zig` by changing `Options.linux_display_backend`
```

- **NOTE:** in zig 0.14.0 it will be possible to add a failure message directly to a failed step (https://github.com/ziglang/zig/pull/20312). It is not needed, but I reference it here for the future.


raysan5:
@lnc3l0t thanks for the improvement!

PR #4221, opened 2024-08-07
[Examples] Update core_3d_camera_first_personCamera

kai-z99:
Regarding issue https://github.com/raysan5/raylib/issues/4208 for core_3d_camera_first_person.

It doesn't mention that SPACE and LCTRL only work in CAMERA_FREE mode. So I added some text to clarify. Im not sure if adding functionallity to SPACE AND LCTRL was the intention for CAMERA_FIRST_PERSON, CAMERA_THIRD_PERSON, and CAMERA_ORBITAL? Because that would make CAMERA_FREE and CAMERA_FIRST_PERSON function identical if im not mistaken?

Also it never mentions that the zoom functionality is not availible in CAMERA_FIRST_PERSON, so I propose the change that zooming in CAMERA_FIRST_PERSON changes your fovy which mimics how zoom works first-person gameplay. It just resets back to the default value of 60.0f once you change cameraMode. That will make CAMERA_FIRST_PERSON differ a little more from CAMERA_FREE.

### DIFF
--- a/examples/core/core_3d_camera_first_person.c
+++ b/examples/core/core_3d_camera_first_person.c
@@ -65,6 +65,7 @@ int main(void)
         {
             cameraMode = CAMERA_FREE;
             camera.up = (Vector3){ 0.0f, 1.0f, 0.0f }; // Reset roll
+            camera.fovy = 60.0f;
         }
 
         if (IsKeyPressed(KEY_TWO))
--- a/examples/core/core_3d_camera_first_person.c
+++ b/examples/core/core_3d_camera_first_person.c
@@ -117,22 +117,22 @@ int main(void)
             }
         }
 
-        //Camera zooming
-        float wheelMove = GetMouseWheelMove();
-        bool zoomIn = IsKeyDown(KEY_KP_ADD);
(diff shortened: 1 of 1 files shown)


casperbear:
Can attest that I did wonder about SPACE and LCTRL several months ago when started to learn RayLib. Thought they were for jumping and sneaking, but were broken. :) Never realized they were for free camera.

CrackedPixel:
I think this has way too many changes that aren't necessary. Why not display different controls based on the camera mode and leave it at that? Or something

kai-z99:
@CrackedPixel For me its just makes it more consistant that each mode can zoom. And it stops the free camera being the same as the first person camera. 

I updated the pr to include your idea of displaying different controls. based on camera mode.  I still think adding zooming to the first-person mode makes sense but I comment it out for now.

raysan5:
@kai-z99 why that code is commented? Add it or remove it.

kai-z99:
@raysan5 I was wondering if you would rather have that feature or remove it?

raysan5:
@kai-z99 in that case, I prefer to leave the example as it is.

PR #4223, opened 2024-08-07
[Examples] update core input mouse

CrackedPixel:
Added list of colours and the mouse button on-screen for people to easier try

image


raysan5:
@CrackedPixel I prefer to avoid showing that much text...

PR #4225, opened 2024-08-07
[Examples] update shaders basic pbr

CrackedPixel:
Scale was not correct. Maybe they model was updated at some point? Not sure. But it looks perfect at 0.25 now

image


raysan5:
@CrackedPixel Mmmh... that's strange... in any case, thanks for review!

PR #4227, opened 2024-08-07
[Examples] update shapes rectangle scaling

CrackedPixel:
This allows to visually see the border and triangle when hovering over the rectangle. It also retains the border when resizing. Much nicer UX

https://github.com/user-attachments/assets/9045ea76-3edf-4ab1-9c40-6f08b958a30f


raysan5:
@CrackedPixel It seems it is mixing the black border with the red border depending on the scale, personally I find it more confusing. I prefer to keep it as it is.

PR #4228, opened 2024-08-07
[Examples] update text format text

CrackedPixel:
- changed "elapsed" to "frame", since thats what its measuring
- added a real "elapsed", which also shows the dynamic text updating

https://github.com/user-attachments/assets/7ebf6422-f52c-4416-b338-68b2ad914646


raysan5:
@CrackedPixel Elapsed time is the amount of time that passes from the start of an event to the finish of the event, in videogames is usually referred to frame time, so that naming is correct for videogames industry standards, referring to frame time.

PR #4229, opened 2024-08-07
[Examples] update text input box

CrackedPixel:
10 FPS causes input issues. 60 is rock solid and most of the other examples are also set to 60 so i think this makes sense


raysan5:

PR #4232, opened 2024-08-08
[Examples] updated example versions

CrackedPixel:
Updated example versions

### DIFF
--- a/examples/shaders/shaders_lightmap.c
+++ b/examples/shaders/shaders_lightmap.c
@@ -7,6 +7,8 @@
 *
 *   NOTE: Shaders used in this example are #version 330 (OpenGL 3.3).
 *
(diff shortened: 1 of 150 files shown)


examples/shaders/shaders_lightmap.c
```diff
@@ -7,6 +7,8 @@
 *
 *   NOTE: Shaders used in this example are #version 330 (OpenGL 3.3).
 *
+*   Example last time updated with raylib 5.5
```
CrackedPixel, line 10: This one did not have an original version :shrug:

raysan5:
@CrackedPixel In most cases, that's missleading. The version of the examples refers to the last raylib version used when creating/updating that example, it's a good reference for me, instead of setting everything to latest.

PR #4233, opened 2024-08-08
[Examples] change texture2d to texture

CrackedPixel:
[Texture2D is just an alias for Texture](https://github.com/raysan5/raylib/blob/97c02b2425225982da6f3569d50fb490fab53c10/src/raylib.h#L275) and i think it looks cleaner with just "Texture". It also makes it clear that there is NOT a Texture3D, etc

This pattern is also used in existing examples such as https://github.com/raysan5/raylib/blob/97c02b2425225982da6f3569d50fb490fab53c10/examples/shaders/shaders_simple_mask.c#L67

### DIFF
--- a/examples/core/core_input_gamepad.c
+++ b/examples/core/core_input_gamepad.c
@@ -38,8 +38,8 @@ int main(void)
 
     InitWindow(screenWidth, screenHeight, "raylib [core] example - gamepad input");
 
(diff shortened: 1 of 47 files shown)


raysan5:
@CrackedPixel As mentioned, I prefer not to change that naming.

PR #4234, opened 2024-08-08
[Examples] update examples missing unloadtexture

CrackedPixel:
We want to set a good example, right? :wink:


raysan5:
@CrackedPixel Oh! Really good catches! Thanks!

casperbear:
@CrackedPixel Btw, in shaders_lightmap.c there is no corresponding UnloadMaterial for "Material material = LoadMaterialDefault();" I've been wondering if it's even needed, and if it is, then should it be called before or after UnloadShader

PR #4237, opened 2024-08-08
[Examples] change rendertexture2d to rendertexture

CrackedPixel:
[RenderTexture2D is just an alias for RenderTexture](https://github.com/raysan5/raylib/blob/97c02b2425225982da6f3569d50fb490fab53c10/src/raylib.h#L288) and i think it looks cleaner with just "RenderTexture". It also makes it clear that there is NOT a RenderTexture3D, etc


raysan5:
@CrackedPixel I'm afraid I prefer to keep the `RenderTexture2D`, it's aligned with `Texture2D` (that it's also an alias of `Texture`), it's more explicit name for users.

PR #4239, opened 2024-08-08
Fix warnings

JeffM2501:
Just a few typecast warnings before 5.5


raysan5:

PR #4242, opened 2024-08-09
Fix build.zig and use zig fmt

freakmangd:
+ `std.mem.split` is deprecated, `splitScalar` seems like the intended choice here
+ zig files should be formatted according to `zig fmt`


raysan5:
@freakmangd thanks for the update!

PR #4243, opened 2024-08-09
[rlgl] Fixed compilation for OpenGL ES

MaximKn1:
Changed the conditions to avoid using GL_COMPUTE_SHADER if OpenGL ES is used. Alternatively, we can use RL_COMPUTE_SHADER instead as it is always defined, but I guess it is not recommended.


raysan5:
@MaximKn1 Thanks for the review

PR #4244, opened 2024-08-09
[Projects]CMakeLists.txt for raylib version issue

ourabigdev:
i dont know but i do think that raylib 5.5 is not out yet and putting it instead of 5.0 break the cmake project but when using raylib target version 5.0 it works

### DIFF
--- a/projects/CMake/CMakeLists.txt
+++ b/projects/CMake/CMakeLists.txt
@@ -5,7 +5,7 @@ project(example)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Dependencies
-set(RAYLIB_VERSION 5.5)
+set(RAYLIB_VERSION 5.0)
 find_package(raylib ${RAYLIB_VERSION} QUIET) # QUIET or REQUIRED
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 please, could you take a look to this issue?

raysan5:
@ourabigdev last raylib release was very dramatic due to not updating that value. I prefer to keep it broken for some weeks than forgetting about updating it again and wait another year or more...

ourabigdev:
> @ourabigdev last raylib release was very dramatic due to not updating that value. I prefer to keep it broken for some weeks than forgetting about updating it again and wait another year or more...

so you say that raylib 5.5 is releasing soon  because keeping it broken can reduce the productivity of the user making him modify the cmake file i think it would be better to fix it and then modify it when the version release

raysan5:
@ourabigdev I'm not the maintainer of CMake, personally I don't like that build system. Last release nobody update it and it has been broken for a year. Considering that fact, I try to avoid that happening again.

Peter0x44:
@ourabigdev the current 5.0 release tag is broken and reports itself as 4.5. Knowing that, it's better that no one forgets to bump the version for next release.

PR #4252, opened 2024-08-13
Change some global variables to have internal linkage

lzralbu:


raysan5:
@lzralbu thanks for the review... mmmh, I don't know how I missed that, not sure if there was any reason for those variables not to be static... 🤔

PR #4257, opened 2024-08-16
Fix raylib version in CMake template

shaperones0:
Raylib 5.5 isn't out yet unfortunately. So the cmake configure fails.


raysan5:
@shaperones0 I'm afraid I'm leaving it broken for a while: https://github.com/raysan5/raylib/pull/4244

PR #4259, opened 2024-08-17
Update RGFW

ColleagueRiley:
Quality of life update:
    more opengl options (core vs comp, disable double buffer, samples defaults to 0)
    deleted unneeded functions

(Bug fixes)
    raw input data shouldn't be flipped


raysan5:
@ColleagueRiley thanks for the update!

PR #4276, opened 2024-08-23
Fixing GenImagePerlinNoise() being stretched, if Image is not rectangular

Bugsia:
I have noticed that when the width and height of GenImagePerlinNoise() are not the same, then the noise will get stretched. So I have added a compensation for it.
Now even if the sides aren´t the same length, the noise won´t get distorted, but rather extended.

Here an image to demonstrate:
On the left is my fix being applied. The noise looks normal.
On the right the current function is being used. The noise is stretched.
fixedPerlinNoise


raysan5:
@Bugsia thanks for the review!

PR #4277, opened 2024-08-23
[build] CMake: Delete BuildOptions.cmake

Peter0x44:
This file seems to not do anything useful. From what I can tell the OSX_FATLIB option sets CMAKE_OSX_ARCHITECTURES to "x86_64;i386". This doesn't account for the arm that apple now has, as well as 32 bit support being completely removed, and I think it's entirely reasonable to expect users to pass the necessary architectures they want themselves. It's possible this could break some users who rely on this, but I sincerely doubt anyone does. The solution is trivial either way (put -DCMAKE_OSX_ARCHITECTURES="i386;x86_64" on the command line yourself)

The second part of BuildOptions.cmake claims to set PLATFORM to "Web" if the emscripten toolchain file is used (if (EMSCRIPTEN)), but it does not work correctly anyway. Currently, glfw searches for wayland and x11 libraries and fails likeso:

CMake Error at /usr/share/cmake/Modules/FindPkgConfig.cmake:645 (message):
  The following required packages were not found:
   - wayland-client>=0.2.7
   - wayland-cursor>=0.2.7
   - wayland-egl>=0.2.7
   - xkbcommon>=0.5.0

Call Stack (most recent call first):
  /usr/share/cmake/Modules/FindPkgConfig.cmake:873 (_pkg_check_modules_internal)
  src/external/glfw/src/CMakeLists.txt:163 (pkg_check_modules)

Considering this code doesn't work as described, it's okay to delete it. I think a better check should be implemented, but that is for a different PR.


raysan5:
@Peter0x44 Thanks for the review! PRs deleting code are the best ones! 😄

PR #4279, opened 2024-08-23
[raudio] Support 24-bit FLACs in `LoadMusicStreamFromMemory`

konstruktor227:
When loading FLAC files using `LoadMusicStreamFromMemory`, 24-bit files play either silence (Windows) or extreme noise (Linux, macOS), while 16-bit files work all fine.

While debugging this I stumbled on #4058, which recently fixed the same problem in `LoadMusicStream` by forcing conversion to 16-bit.

Here simply copying the same fix to `LoadMusicStreamFromMemory`, which fixes all the issues with 24-bit files for me.


raysan5:
@konstruktor227 Nice! Thanks for the review! 😄

PR #4280, opened 2024-08-24
[RCORE] Update comments on fullscreen and boderless window to describe what they do

JeffM2501:
There is a large amount of confusion regarding what ToggleFullscreen and ToggleBorderlessWindow do (3 questions today alone).
The current comments are insufficient, the term 'Fullscreen' no longer means what it used to in the days of CRTs, and the expectation is that the function changes the size of the window, but it does not.

This PR adds a description to each function saying what it does exactly, to help reduce this confusion for what is now becoming an increasingly poorly named function.


raysan5:

PR #4282, opened 2024-08-24
(rcore_desktop_rgfw.c) fix errors when compiling with mingw

ColleagueRiley:
`MultiByteToWideChar` should be defined for msvc AND mingw.

`win->src.window` is now used for all platforms


raysan5:
@ColleagueRiley Thanks for the review!

PR #4284, opened 2024-08-24
rlgl.h: glint64 did not exist before OpenGL 3.2

Tchan0:
Compilation breaks on rlgl.h for early OpenGL versions. 
Glint64 did not exist on those versions (< OpenGL 3.2)


raysan5:
@Tchan0 Good catch! Thanks for the review!

PR #4286, opened 2024-08-24
Examples makefiles: align /usr/local with /src Makefile

Tchan0:
In the /src Makefile, /usr/local can be overriden by the user.
Align the /example Makefile & Makefile.Web with that.


raysan5:
@Tchan0 thanks for the review!

PR #4290, opened 2024-08-26
[rcore] created SetWindowShouldClose

brccabral:
SetWindowShouldClose can be used to indicate the app/game to end the `while (!WindowShouldClose())` main loop


raysan5:
@brccabral I think there is no need for this function. Users can control the main loop with a custom variable if required.

PR #4294, opened 2024-08-28
Fix missing equal sign

hanaxars:
I just noticed there is a missing equal sign.


raysan5:
@hanaxars Oh! Good catch! Thanks for the review!

PR #4296, opened 2024-08-30
[rlgl] [rshapes] [rtextures] Fixes for disabled SUPPORT_QUADS_DRAW_MODE

Destructor17:
While trying to make raylib work in WASI and OpenGL ES, i figured out that my OpenGL ES driver does not supports quads draw mode. But disabling SUPPORT_QUADS_DRAW_MODE leads to following bugs:
- Many elements are still using RL_QUADS mode.
- Many elements are displayed with default texture.

I tried a few examples and managed to fix RL_QUADS bug for them (that was hard due to way debugger in wasmtime works). Fixes are straightforward, but it is likely that RL_QUADS is still used in some raylib modules and examples. BTW there is an alternative way to solve this problem: emulating RL_QUADS using triangles and pushing 2 additional vertices every 4th call of rlVertex3f. I guess it worth investigating.

Honestly i don't know how textures worked before, but in my case i had to make raylib restore texture(which is usually set by rlSetTexture before rlBegin call) when draw mode switches (inspired by https://github.com/raysan5/raylib/pull/2018)


Destructor17:
It looks like the texture bug is known. textures_polygon.c, there is a `// Texturing is only supported on RL_QUADS` comment. Restore texture on rlBegin call makes this example work with triangles well.

PR #4302, opened 2024-09-01
[rcore] Add filtering folders to `LoadDirectoryFilesEx()`/`ScanDirectoryFiles()`

foxblock:
Right now LoadDirectoryFiles and LoadDirectoryFilesEx only return either all folders and all files (when no filter is passed) or a list of files filtered by file extension.
For use in a file browser (like raygui\examples\custom_file_dialog) it is useful to gather a list of files with a particular extension and also all folders in a location (so the user can navigate further).

This change implements this by adding a FILTER_FOLDER define.
The define can be used as a filter and will match all folders. The contents of the define are chosen as to never match a file extension on Linux and Windows.

So for example you can do
`LoadDirectoryFilesEx(%RAYLIB_SRC_DIR%, FILTER_FOLDER";.c", false);`
which results in
```
..\raylib\src\external
..\raylib\src\platforms
..\raylib\src\raudio.c
..\raylib\src\rcore.c
..\raylib\src\rglfw.c
..\raylib\src\rmodels.c
..\raylib\src\rshapes.c
..\raylib\src\rtext.c
..\raylib\src\rtextures.c
..\raylib\src\utils.c
```

As a side effect this fixes one "bug" (at least I assume it was not intended behaviour):
LoadDirectoryFiles(Ex) would previously return folders which also matched the file extension. So `LoadDirectoryFilesEx(RAYLIB_SRC_DIR, ".c", false);` would match a folder with a name like "code.c". 
With this change it does not do that any more (only if FILTER_FOLDER would have been passed as well).

Existing calls to LoadDirectoryFilesEx should not be affected otherwise and return the same results as before.

Btw. I did not see a place to add tests, but I am happy to provide some if needed.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -194,6 +194,10 @@
 #define MAGENTA    CLITERAL(Color){ 255, 0, 255, 255 }     // Magenta
 #define RAYWHITE   CLITERAL(Color){ 245, 245, 245, 255 }   // My own White (raylib logo)
 
+// Filter string used in ScanDirectoryFiles, ScanDirectoryFilesRecursively and
+// LoadDirectoryFilesEx to include directories in the result array
+#define FILTER_FOLDER "/DIR"
+
 //----------------------------------------------------------------------------------
 // Structures Definition
 //----------------------------------------------------------------------------------
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -194,10 +194,6 @@
 #define MAGENTA    CLITERAL(Color){ 255, 0, 255, 255 }     // Magenta
 #define RAYWHITE   CLITERAL(Color){ 245, 245, 245, 255 }   // My own White (raylib logo)
 
-// Filter string used in ScanDirectoryFiles, ScanDirectoryFilesRecursively and
-// LoadDirectoryFilesEx to include directories in the result array
-#define FILTER_FOLDER "/DIR"
-
 //----------------------------------------------------------------------------------
 // Structures Definition
 //----------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/raylib.h
```diff
@@ -194,6 +194,10 @@
 #define MAGENTA    CLITERAL(Color){ 255, 0, 255, 255 }     // Magenta
 #define RAYWHITE   CLITERAL(Color){ 245, 245, 245, 255 }   // My own White (raylib logo)
 
+// Filter string used in ScanDirectoryFiles, ScanDirectoryFilesRecursively and
+// LoadDirectoryFilesEx to include directories in the result array
+#define FILTER_FOLDER "/DIR"
```
raysan5, line 199: I prefer to avoid exposing this define, it can be defined internally in `rcore` and just document as an option in the function

foxblock: I can make the change, no problem.
Just to make sure: People would call it like this (with the "/DIR" option mentioned in documentation)?
LoadDirectoryFilesEx("somepath", "/DIR;.c", false);

The only minor issue I see with this, is that documentation has to be updated separately from the code then. If you ever want to change the "/DIR" string to something else, you will have to make sure to edit the documentation accordingly. This would also break existing code, while using an exposed define would always "just work".
Personally I would rather expose the define, but it's your library so you make the call.

If you are okay with that, I will make the requested change.

raysan5: We could keep the define but internal to `rcore` module and configurable **only at compile time** through `config.h`, like many other options in raylib.

foxblock: Okay I made the requested changes and pushed them to the branch.

raysan5:
@foxblock Thank you very much for the improvement, I will think about the best tag name for directories... but `/DIR` looks good to me for now.

Issue #4304, opened 2024-09-03
[raylib_parser] Segmentation fault


chrisws:
I get `Segmentation fault (core dumped)` when running the command

`/raylib_parser --format JSON --input ../src/raylib.h --output raylib_api.json`

The problem is related to the length of the comment for the following in raylib.h

`RLAPI Font LoadFontEx(const char *fileName, int fontSize, int *codepoints, int codepointCount);`

If I remove the last 6 characters of the comment, the parser completes but the output JSON is a bit broken. I needed to remove further comment lines to get valid JSON output.

raysan5:
@chrisws feel free to send a PR increasing comments buffer size

PR #4306, opened 2024-09-04
[raylib_parser] Fix seg fault with long comment lines

chrisws:
#4304


raysan5:
@chrisws thanks for the fix!

PR #4308, opened 2024-09-06
Change implicit conversion to explicit conversion to remove warning

masnm:
I was using raylib in one of my project which i was compiling with clang and clang showed an warning about the implicit conversion. So making a PR to get rid of the warning and to help the project in a negligible way.


raysan5:
@masnm Thanks for the review but note this change could be overwritten on `par_shapes` update, this change should be sent to the library main repo.

PR #4312, opened 2024-09-10
fix for hardcoded index values in vboID array

JettMonstersGoBoom:
changing any of the #defines in CONFIG.H related to vboId could cause issues when rendering.

( wanted to add more attributes for a test and ran into this issue )


raysan5:
@JettMonstersGoBoom Good improvement! I don't know why it was not already done this way...

PR #4313, opened 2024-09-11
Add a way to get GLFWwindow and RGFW_window pointers

sse2:
Right now, there's no way to get the underlying GLFWwindow structure ([unless you're on Linux](https://github.com/raysan5/raylib/blob/fd961deba76c7c4939440fa6e6574eb86e68c5fb/src/platforms/rcore_desktop_glfw.c#L714)). This is especially useful when trying to get something like ImGui to work with Raylib without using any of the existing wrappers (espeically since they don't support docking & multiple viewports properly).

This could possibly be expanded for other platforms too, such as Android & SDL. For the rcore_desktop_sdl platform, GetWindowHandle returns the SDL_Window pointer, so perhaps this could return the GLContext?

Don't know if this functionality is actually wanted by anyone but myself, but opening just in case.

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -424,6 +424,12 @@ void *GetWindowHandle(void)
     return NULL;
 }
 
+void *GetPlatformHandle(void) 
+{
+    TRACELOG(LOG_WARNING, "GetPlatformHandle() not implemented on target platform");
+    return NULL;
+}
+
 // Get number of monitors
 int GetMonitorCount(void)
 {
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -721,6 +721,11 @@ void *GetWindowHandle(void)
     return NULL;
 }
 
+void *GetPlatformHandle(void) 
+{
+    return platform.handle;
+}
+
 // Get number of monitors
 int GetMonitorCount(void)
 {
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -723,7 +723,7 @@ void *GetWindowHandle(void)
 
 void *GetPlatformHandle(void) 
 {
-    return platform.handle;
+    return (void*)platform.handle;
 }
 
 // Get number of monitors
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -541,6 +541,12 @@ void *GetWindowHandle(void)
 #endif
 }
 
+void *GetPlatformHandle(void) 
+{
+    TRACELOG(LOG_WARNING, "GetPlatformHandle() not implemented on target platform");
+    return NULL;
+}
+
 // Get number of monitors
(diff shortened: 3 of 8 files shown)


mfbulut:
This might be a useful feature, but the naming is confusing because there is already a function called GetWindowHandle.

- GetWindowHandle → Returns the platform-specific handle
- GetPlatformHandle → Returns the window handle

The naming should be reversed, but doing so would break compatibility with older versions. (I'm not a maintainer I just wanted to point this out.)

raysan5:
@sse2 I'm afraid I'm not merging this change, it seems very platform-specific to expose a specific library structure for a specific platform. I think most users do not need it.

ColleagueRiley:
> @sse2 I'm afraid I'm not merging this change, it seems very platform-specific to expose a specific library structure for a specific platform. I think most users do not need it.

You should be able to get the API structure on any platform imo. This would help a lot with handling things Raylib does not support or handling events.

One problem I heard of recently regarding Raylib events is that it's impossible to know the order events happen. For example if you check if the mouse button was press and a key was press or if the "a" key was pressed then the "backspace" key was pressed you cannot know which event happened first. I might be wrong about that last example, but if I'm not then being able to get the API's window structure would allow the user to handle events themselves if they need to. 

It's a relatively simple feature that would enable Raylib users to handle a wider range of features.

raysan5:
@ColleagueRiley Users with those specific needs could add this feature on their forks, as you say is a relative simple feature.

ColleagueRiley:
> @ColleagueRiley Users with those specific needs could add this feature on their forks, as you say is a relative simple feature.

Adding an extra layer of work the user has to do for no good reason seems silly. I don’t think the user should have to manage their own fork for something that, in my opinion, should be in base Raylib. 

It would make a lot of sense for this to be a feature in Raylib because this would allow the user to accomplish way more things that they otherwise couldn’t do with base Raylib.

raysan5:
@ColleagueRiley Sorry, no plans to add it to raylib.

CasparSchoolderman:
Managed to add this functionality myself by:

Add in rcore_desktop_glfw.c, add the following function to allow access to the GLFW window handle:
void* getGLFWwindowHandle(void)
{
    return (void*)platform.handle;
}

And added the following into raylib.h:
RLAPI void* getGLFWwindowHandle(void);

In the main code you can now do the following which makes IMGUI work:
    // Setup Platform/Renderer backends
    GLFWwindow* window;
    window = (GLFWwindow*) getGLFWwindowHandle();

    ImGui_ImplGlfw_InitForOpenGL(window, true);
    ImGui_ImplOpenGL3_Init(glsl_version);

CasparSchoolderman:
> > @ColleagueRiley Users with those specific needs could add this feature on their forks, as you say is a relative simple feature.
> 
> Adding an extra layer of work the user has to do for no good reason seems silly. I don’t think the user should have to manage their own fork for something that, in my opinion, should be in base Raylib.
> 
> It would make a lot of sense for this to be a feature in Raylib because this would allow the user to accomplish way more things that they otherwise couldn’t do with base Raylib.

See a few comments below on how to implement.

PR #4325, opened 2024-09-16
[raymath] Add Vector2Limit

vinibispo:
My idea came from https://p5js.org/reference/p5.Vector/limit/
Let me know if this is appreciated and what else I need to do, thanks :)


raysan5:
@vinibispo I'm afraid I'm not adding this function for now, it would require an additional Vector3Limit() and QuaternionLimit() for consistency... but still it seems for quite specific use-cases.

vinibispo:
I could add for both missing if you want it, but I understand if this PR won't be merged given these specific use cases

PR #4332, opened 2024-09-20
Projects: Fix CMake example project 

Ridge3Dproductions:
This adds missing link flags inside the CMake example without these link flags the HTML5 build will not work.

The snippet in the Cmake file was taken from the [wiki](https://github.com/raysan5/raylib/wiki/Working-for-Web-(HTML5)#23-using-cmake)


raysan5:
@Ridge3Dproductions Thanks for the review!

PR #4334, opened 2024-09-21
[rlgl] Fix rlgl standalone defaults

asdqwe:
Ref: [src/config.h#L124](https://github.com/raysan5/raylib/blob/e5d0cc978ae933ccccd41758d50d1cd57e3474d1/src/config.h#L124)


raysan5:
@asdqwe Good catch! Thanks for the review!

PR #4340, opened 2024-09-24
DRAFT: Fix isGpuReady flag on android

Mercotui:
This commit sets the isGpuReady flag when the android platform code has initialized the GPU.

This seemingly duplicates the code that is used in `rcore.c`.


raysan5:
@Mercotui Thanks for the review!

PR #4344, opened 2024-09-27
[SHAPES] Add more detail to comment for DrawPixel

JeffM2501:
This PR adds some more detail to the comment for DrawPixel and DrawPixelV to make it clear that they use geometry and may have an impact on rendering performance.


raysan5:
@JeffM2501 Thanks for the improvement!

Issue #4346, opened 2024-09-29
[rmodels] Errors during LoadMesh on MacOS


TillWege:
### Issue description

Since #4321 uploading Meshes on MacOS will randomly fail and cause the running program to crash.

### Environment

- MacBook Pro M1 16gb running MacOS 15

### Issue Screenshot

<img width="631" alt="image" src="https://github.com/user-attachments/assets/64f448c8-1f09-4fe0-b210-af3921eb5b7a">

### Code Example

```c++
#define MODEL_FOLDER "models/"

int main()
{
	const int screenWidth = 1280;
	const int screenHeight = 720;

	InitWindow(screenWidth, screenHeight, "Window");

	std::vector<Model> models;

	for (const auto& entry : std::filesystem::directory_iterator(MODEL_FOLDER)) {
		std::string path = entry.path().string();
		std::string fileName = entry.path().filename().string();
		if (IsFileExtension(fileName.c_str(), ".obj")) {
			Model model = LoadModel(path.c_str());
			models.push_back(model);
		}
	}
}
```

Running this leads to it crashing within the first ~4 models it tries to load. It doesn't seem to change whether its .obj or .gltf files.

JeffM2501:
It's possible that the mac may only support 8 VBOs, the code should check GL for the max and not try to load them all if possible.
We may need to tie the GPU skinning feature to a flag that apple platforms do not set, or perhaps remove one of the other less used VBOs

deDug:
This could be the same problem as https://github.com/raysan5/raylib/issues/4343 where loading mesh a bad define means the wrong amount of memory is allocated for VBo indices. 
You could try changing `MAX_MESH_VERTEX_BUFFERS` in `config.h` to 9 and see if that fixes it.

TillWege:
Yeah you seem to be right. When I change it to 7 the crash doesnt appear.

JeffM2501:
I just tested on an intel mac, and can verify that yes, the mac only supports 8 VBOs. 
We will have to not support GPU meshes on the mac, and move those buffers to the end.
This includes meshes that don't even use animations, the change to use 9 VBOs just strait up breaks meshes on Mac OS.

JeffM2501:
PR https://github.com/raysan5/raylib/pull/4348 should disable this feature for MacOS

stevegwh:
[...]

PR #4348, opened 2024-09-29
[MODELS] Disable GPU skinning for MacOS platform

JeffM2501:
The Apple OpenGL drivers for MacOS does not seem to support more than 8 VBOs per mesh, so the new feature to add two additional VBOs to support GPU skinning breaks all model loading on the Mac, even if the model is not animated, simply due to the fact that it tries to allocate the index buffer at ID 8 

This PR disables these VBOs for the apple platform via a new #define in rlgl RL_SUPPORT_MESH_ANIMATION_VBO
This can allow us to disable the buffers for other platforms that may have the same issues, if apple is not the only vendor with this issue. I moved the optional buffers to buffer indexes 7 and 8, making the index buffer be 6, so that they can still be used on the Mac.

I have also added notes to functions and examples that say that the feature is not supported on apple platforms due to limitations in the driver.

Oh and I also added the GPU skinning example to MSVC.

I tested this by validating that Meshes did not load on a mac (2011 intel mac) before the change and do load after the change. I also validated that GPU skinning does still work on windows (Hence adding the example to MSVC projects).


JeffM2501:
This should resolve issue #4346

raysan5:
@JeffM2501 Thank you very much for reviewing this issue! That was a good catch! Sadly, there is always some kind of nuance with macOS... Pinging @orangeduck that probably was not aware of this macOS limitation.

orangeduck:
Yes sadly I was not aware of this limitation :( - thanks a lot for this @JeffM2501

PR #4351, opened 2024-09-30
fix: fixes issue while downloading the binaries from gh

OctoD:
The resource at the URL 
https://github.com/raysan5/raylib/archive/refs/tags/5.5.tar.gz Cannot be found, so the download fails and the lib cannot be added via URL.


raysan5:
@OctoD A new version is being released soon, and it needs to be changed to 5.5.

PR #4352, opened 2024-09-30
[rlgl] Complements the #4348 GPU skinning fix

asdqwe:
Rationale: although it may not get defined in [config.h#L124-L132](https://github.com/raysan5/raylib/blob/0e7bcd5639729fac1df19ed2d0622ff3daab59e3/src/config.h#L124-L132), it ends up always getting defined on [rlgl.h#L350-L355](https://github.com/raysan5/raylib/blob/0e7bcd5639729fac1df19ed2d0622ff3daab59e3/src/rlgl.h#L350-L355) defaults, which is more noticeable when using rlgl standalone.

CC @JeffM2501 for review.


raysan5:
@asdqwe Thanks for the improvement! 

Actually, I'm considering exposing a new config flag for `RL_SUPPORT_MESH_GPU_SKINNING`.

asdqwe:
@raysan5 That would be super helpful, specially for 2D oriented builds that wouldn't use GPU skinning.

PR #4353, opened 2024-10-01
[MODELS] Better fix for GPU skinning issues

JeffM2501:
Looks like the core issue with GPU skinning was a missmatch between the VBO array and the #define for the max VBO array size.
This PR fixes that issue and re-enables GPU skinning for the mac. I have verified that it works on a mac.

I did keep, but rename the #define to enable GPU skinning in case someone needs to disable it for some reason.


raysan5:
Nice! Thanks for the review Jeff!

PR #4354, opened 2024-10-01
Fixes `GetClipboardText()` memory leak for `PLATFORM_DESKTOP_SDL`

asdqwe:
`SDL_GetClipboardText()` must be freed with `SDL_free()` ([doc](https://wiki.libsdl.org/SDL2/SDL_GetClipboardText#return-value)), but that can't/shouldn't be done to a returned `const`.

Fix can be tested with:
```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        if (IsKeyPressed(KEY_ONE)) { TraceLog(LOG_INFO, "%s", GetClipboardText()); }

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -58,6 +58,13 @@
 #endif
 
 //----------------------------------------------------------------------------------
+// Defines and Macros
+//----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


raysan5:
@asdqwe Great improvement! Thanks!

PR #4358, opened 2024-10-02
Allow Zig build script to change desktop backend

### LINKED ISSUE #4299: [build] Add an option to `build.zig` to support different `rcore` platforms
The `build.zig` script currently doesn't provide an option to switch between the different desktop platform implementations.

I would add one myself, but I lack time to create and test that functionality.

ref: Not-Nik/raylib-zig#143

Not-Nik:
Allow Zig build script to change desktop backend using a new `platform` option, which replaces the previous `platform_drm` option. I don't have an extensive testing setup, so someone else should at least run the build once too.

fixes #4299 (closing unresolved issues is weird?)
cc @ryupold 
ref Not-Nik/raylib-zig#143

### DIFF
--- a/src/build.zig
+++ b/src/build.zig
@@ -40,6 +40,17 @@ pub fn addRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
     return raylib;
 }
 
+fn setDesktopPlatform(raylib: *std.Build.Step.Compile, platform: PlatformBackend) void {
+    raylib.defineCMacro("PLATFORM_DESKTOP", null);
+
+    switch (platform) {
+        .glfw => raylib.defineCMacro("PLATFORM_DESKTOP_GLFW", null),
+        .rgfw => raylib.defineCMacro("PLATFORM_DESKTOP_RGFW", null),
+        .sdl => raylib.defineCMacro("PLATFORM_DESKTOP_SDL", null),
+        else => {}
+    }
+}
+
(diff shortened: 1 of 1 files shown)


raysan5:
@Not-Nik Oh! This is a really nice addition! Thanks! 😄

HaxSam:
I tried to use the new `platform` option but without any success.
When I try to use SDL raylib it prints out `Platform backend: DESKTOP (GLFW)`

My assumption why it does not work is because of the line `raylib.defineCMacro("PLATFORM_DESKTOP", null);`
after stetting `PLATFORM_DESKTOP` in `src/rcore.c` `PLATFORM_DESKTOP_GLFW` gets always set as default
https://github.com/raysan5/raylib/blob/415a98965a9d8ca1538e890367f27ec08c95f05d/src/rcore.c#L511C1-L513C7
```c
#if defined(PLATFORM_DESKTOP)
    #define PLATFORM_DESKTOP_GLFW
#endif
```

I cloned the repo and tried my assumption and I get now the error that `rcore_desktop_sdl.c` cant find `SDL.h`
This looks more like that the platform really changed.

@raysan5 should I create a separate issue for that?

raysan5:
@HaxSam you can just send a PR with the fix or open an issue...

HaxSam:
@raysan5 I'm going to make a PR after I figured out how to include SDL when changing the platform on zig

PR #4359, opened 2024-10-02
[build] CMake: Fix GRAPHICS check

zyperpl:
Currently, CMake does not set and check the `SUGGESTED_GRAPHICS` variable correctly when checking for the overwritten `GRAPHICS` variable. As a result, the warning message is never shown. 

This PR fixes the issue and also replaces one instance of tabs with spaces to match the rest of the code.


raysan5:
@zyperpl Thanks for the review!

PR #4247, opened 2024-08-11
Make camera movement independant of framerate

hanaxars:
Instead of moving camera with constant speed per frame, speed is multiplied with delta time before movement.

PR #4362, opened 2024-10-04
[rcamera] updated camera speeds with GetFrameTime

CrackedPixel:
Starting from https://github.com/raysan5/raylib/pull/4247

Did the same type of logic for the other camera speeds to factor in the `GetFrameTime()`

The other PR solved the issue when using WASD but arrow keys, E, and Q were still affected

### DIFF
--- a/src/rcamera.h
+++ b/src/rcamera.h
@@ -444,11 +444,17 @@ void UpdateCamera(Camera *camera, int mode)
     bool lockView = ((mode == CAMERA_FREE) || (mode == CAMERA_FIRST_PERSON) || (mode == CAMERA_THIRD_PERSON) || (mode == CAMERA_ORBITAL));
     bool rotateUp = false;
 
+    // Camera speeds based on frame time
+    float cameraMoveSpeed = CAMERA_MOVE_SPEED*GetFrameTime();
+    float cameraRotationSpeed = CAMERA_ROTATION_SPEED*GetFrameTime();
+    float cameraPanSpeed = CAMERA_PAN_SPEED*GetFrameTime();
+    float cameraOrbitalSpeed = CAMERA_ORBITAL_SPEED*GetFrameTime();
+
     if (mode == CAMERA_CUSTOM) {}
     else if (mode == CAMERA_ORBITAL)
     {
         // Orbital can just orbit
-        Matrix rotation = MatrixRotate(GetCameraUp(camera), CAMERA_ORBITAL_SPEED*GetFrameTime());
+        Matrix rotation = MatrixRotate(GetCameraUp(camera), cameraOrbitalSpeed);
         Vector3 view = Vector3Subtract(camera->position, camera->target);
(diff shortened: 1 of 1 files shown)


src/rcamera.h
```diff
@@ -444,34 +444,40 @@ void UpdateCamera(Camera *camera, int mode)
...
+    float cameraMoveSpeed = CAMERA_MOVE_SPEED*GetFrameTime();
+    float cameraRotationSpeed = CAMERA_ROTATION_SPEED*GetFrameTime();
+    float cameraPanSpeed = CAMERA_PAN_SPEED*GetFrameTime();
+    float cameraOrbitalSpeed = CAMERA_ORBITAL_SPEED*GetFrameTime();
+
     if (mode == CAMERA_CUSTOM) {}
     else if (mode == CAMERA_ORBITAL)
     {
         // Orbital can just orbit
-        Matrix rotation = MatrixRotate(GetCameraUp(camera), CAMERA_ORBITAL_SPEED*GetFrameTime());
```
CrackedPixel, line 451: this one had `GetFrameTime()` included in-line. should we do this for all the lines instead?

src/rcamera.h
```diff
@@ -481,7 +487,6 @@ void UpdateCamera(Camera *camera, int mode)
         }
 
         // Keyboard support
-        float cameraMoveSpeed = CAMERA_MOVE_SPEED*GetFrameTime();
```
CrackedPixel, line 484: This is what the other PR added

raysan5:
@CrackedPixel I'm merging tthose changes for now but I'm afraid this functionality should be redesigned in some other way, currently it couples the `rcamera.h` module with raylib that was something I tried to avoid. 

The original idea was making the camera module self-contained to allow it being used as an stand-alone library... but latest big redesign of the module already implied a high-coupling with `raymath.h`...

At the moment, `UpdateCameraPro()` is the only function decoupled from `raylib.h`, where user provides camera movement/rotation/zoom values on the function call.

CrackedPixel:
> @CrackedPixel I'm merging tthose changes for now but I'm afraid this functionality should be redesigned in some other way, currently it couples the `rcamera.h` module with raylib that was something I tried to avoid.
> 
> The original idea was making the camera module self-contained to allow it being used as an stand-alone library... but latest big redesign of the module already implied a high-coupling with `raymath.h`...
> 
> At the moment, `UpdateCameraPro()` is the only function decoupled from `raylib.h`, where user provides camera movement/rotation/zoom values on the function call.

What if instead of `rcamera` handling different modes (CAMERA_ORBIT, etc), there is a separate (maybe extras) that shows how to do this behavior and its removed from rcamera? So then rcamera wouldnt need input/speeds/GetFrameTime at all. People would have to implement their own but they could reference some sort of extras for how to do orbit, first person, etc

I know it will need a lot of consideration and planning from you, but im just throwing out ideas hahaha

PR #4356, opened 2024-10-02
Typo fix

FluxFlu:
Seemed to be a typo. Hope this helps.

PR #4363, opened 2024-10-04
[examples] Typo fix -- taken from FluxFlu

CrackedPixel:
Taken from https://github.com/raysan5/raylib/pull/4356


raysan5:

Issue #4347, opened 2024-09-29
[rlgl][rshapes] rlBegin discards texture when drawing mode switches


Destructor17:
### Issue description

When using GL version other then 1.1 and rlBegin is called with mode which differs from previous one, it sets textuteID to RLGL.State.defaultTextureId. It is a problem because texture is set using rlSetTexture function BEFORE rlBegin is called. Possible solution is to preserve textureID whet drawing mode switches, as it is done in https://github.com/raysan5/raylib/pull/4296. Surprisingly, many functions in rshapes, such as DrawEllipse(and maybe in other modules) depends on this wrong behavior: they don't set texture while hoping it is already set to default one(ID 1). It means that  fixing this bug in rlBegin will uncover a lot of other bugs in rshapes (this is one of reasons why I closed my PR).

The brightest example is [textures_polygon](https://github.com/raysan5/raylib/blob/master/examples/textures/textures_polygon.c). The current implementation works but  has the following comment:
```
// Texturing is only supported on RL_QUADS
```
and uses quads QUADS where two last vertices are same instead of triangles.

### Environment

GLFW, Linux, OpenGL ES 3.2
I guess this bug may be reproduced at any OpenGL version other than 1.1

### Issue Screenshot

See code example. This is without preserving textureID (polygon is white)
![image](https://github.com/user-attachments/assets/6a5d6fee-284f-4a07-ab11-1da9d9ca36fc)

And this is with preserving (ellipse is missing)
![image](https://github.com/user-attachments/assets/c4ef2a0e-f318-420e-9b6f-f40d29db21c3)

### Code Example

Code of modified DrawTexturePoly. Note DrawEllipse and RL_TRIANGLES
```
void DrawTexturePoly(Texture2D texture, Vector2 center, Vector2 *points, Vector2 *texcoords, int pointCount, Color tint)
{
    DrawEllipse(GetScreenWidth()/2.0f, GetScreenHeight()/2.0f, GetScreenWidth()/4.0f, GetScreenHeight()/4.0f, RED);
    rlSetTexture(texture.id);

    // Texturing is only supported on RL_QUADS
    rlBegin(RL_TRIANGLES);

        rlColor4ub(tint.r, tint.g, tint.b, tint.a);

        for (int i = 0; i < pointCount - 1; i++)
        {
            rlTexCoord2f(0.5f, 0.5f);
            rlVertex2f(center.x, center.y);

            rlTexCoord2f(texcoords[i].x, texcoords[i].y);
[...]

PR #4364, opened 2024-10-04
[rlgl] Preserve texture on mode switching

Destructor17:
`rlBegin` resets texture to default when drawing mode switches. This PR makes `rlBegin` preserve texture. Since many routines expect texture to be default if not explicitly specified, `rlEnd` now resets texture to default.

Fixes https://github.com/raysan5/raylib/issues/4347

### DIFF
--- a/examples/textures/textures_polygon.c
+++ b/examples/textures/textures_polygon.c
@@ -88,9 +88,9 @@ int main(void)
         BeginDrawing();
 
             ClearBackground(RAYWHITE);
-
+ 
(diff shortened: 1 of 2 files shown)


raysan5:
@Destructor17 I can't remember now but I'm sure there was a reason why texture reseting was not done on `rlEnd()`, I think it was related to batch buffer filling in the middle of drawing before reaching rlEnd() but not sure. I prefer to avoid this change for now but we can keep the PR open for further investigation and testing.

raysan5:
@Destructor17 This change broke the batching system, reverted.

PR #4369, opened 2024-10-06
[Makefile] updated makefile to disable wayland by default

CrackedPixel:
GLFW has a lot of caveats and different behavior for Wayland compared to X11 on Linux systems
By disabling it as default, we can avoid these issues but still allow Wayland for those who want it explicitly and accept the caveats/changes

Idea taken from @Peter0x44

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -109,8 +109,9 @@ RAYLIB_MODULE_RAYGUI_PATH ?= $(RAYLIB_SRC_PATH)/../../raygui/src
 # Use external GLFW library instead of rglfw module
 USE_EXTERNAL_GLFW     ?= FALSE
 
-# Enable support for both Wayland and X11 by default on Linux when using GLFW
-GLFW_LINUX_ENABLE_WAYLAND  ?= TRUE
+# Enable support for X11 by default on Linux when using GLFW
+# NOTE: Wayland is disabled by default, only enable if you are sure
+GLFW_LINUX_ENABLE_WAYLAND  ?= FALSE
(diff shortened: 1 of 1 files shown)


asdqwe:
My two cents:
1. For better or worst the major distros already migrated to Wayland as default ([ref](https://en.wikipedia.org/wiki/Wayland_(protocol)#Desktop_Linux_distributions)) and depending on Xwayland is unreliable since it's far from perfect and may not even be available.
2. Unless all the major distros revert course (which is unlikely), migrating to Wayland is unavoidable. So it's probably better to start migrating earlier while Xwayland is still kept around for compatibility, then rushing later when it may no longer be maintainted.
3. Yes, some funcionality we're used to have may be lost (control being passed over to the Window Manager instead of the program), but ignoring Wayland won't work on the long term.

raysan5:
@CrackedPixel I think someone commented about it on Discord with some reasons to keep the Wayland option active but I can't find the message. Not sure what to do, I'd need more opinions on this change...

CrackedPixel:
> @CrackedPixel I think someone commented about it on Discord with some reasons to keep the Wayland option active but I can't find the message. Not sure what to do, I'd need more opinions on this change...

Maybe you're thinking this one? They said they wanted a flag to enable/disable, but that already exists. It would just be good to disable by default since glfw has a lot of problems with wayland

image

raysan5:
@CrackedPixel thanks for this cahnge and thanks to @asdqwe for the detailed report on the topic.

PR #4370, opened 2024-10-07
removed old comment

CrackedPixel:
"data" parameter does not exist


raysan5:
@CrackedPixel Wow! I can't believe that comment was still there! Afair, that was changed +10 years ago! 😮

PR #4373, opened 2024-10-08
Some update to gltf loading

HarryDC:
Hit some small snags when importing from kenneys assets this is just a small update.

- Allows animations to be imported even if they affect more `skins` than expected, this allows users to decide if the degraded animations work for them or not 
- Allows `unsigned char` as an mesh index type 

It slightly changes the import macro to correctly cast to the expected type, the old macro is overloaded to maintain the old functionality

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5226,17 +5226,19 @@ static Model LoadGLTF(const char *fileName)
     ***********************************************************************************************/
 
     // Macro to simplify attributes loading code
-    #define LOAD_ATTRIBUTE(accesor, numComp, dataType, dstPtr) \
+    #define LOAD_ATTRIBUTE(accesor, numComp, srcType, dstPtr) LOAD_ATTRIBUTE_CAST(accesor, numComp, srcType, dstPtr, srcType) 
+
+    #define LOAD_ATTRIBUTE_CAST(accesor, numComp, srcType, dstPtr, dstType) \
     { \
         int n = 0; \
-        dataType *buffer = (dataType *)accesor->buffer_view->buffer->data + accesor->buffer_view->offset/sizeof(dataType) + accesor->offset/sizeof(dataType); \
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -5226,17 +5226,19 @@ static Model LoadGLTF(const char *fileName)
     ***********************************************************************************************/
 
     // Macro to simplify attributes loading code
-    #define LOAD_ATTRIBUTE(accesor, numComp, dataType, dstPtr) \
+    #define LOAD_ATTRIBUTE(accesor, numComp, srcType, dstPtr) LOAD_ATTRIBUTE_CAST(accesor, numComp, srcType, dstPtr, srcType) 
+
+    #define LOAD_ATTRIBUTE_CAST(accesor, numComp, srcType, dstPtr, dstType) \
```
HarryDC, line 5231: @raysan5 As mention in discord this would have to change to something like  

```
if (dstPtr == NULL) { 
  dstPtr = RL_ALLOC(accessor->count * numComp * sizeof(dstType); 
}
```
This would reduce a lot of the 4-6 line `if/else` branches to 1 Line 
```
LOAD_ATTRIBUTE(accessor, numpComp, srcType, dstPtr, dstType)
```

src/rmodels.c
```diff
@@ -5697,23 +5699,25 @@ static Model LoadGLTF(const char *fileName)
                         // Load unsigned short data type into mesh.indices
                         LOAD_ATTRIBUTE(attribute, 1, unsigned short, model.meshes[meshIndex].indices)
                     }
+                    else if (attribute->component_type == cgltf_component_type_r_8u)
+                    {
+                        // Init raylib mesh indices to copy glTF attribute data
```
HarryDC, line 5704: With allocation inside the macor, this would be one line

raysan5: Looks good to me.

raysan5:
@HarryDC Thanks for the improvement! Added a note on proposed macro change: looks fine to me.

PR #4375, opened 2024-10-10
build.zig: Improve logic

sagehane:
Currently, the project was failing to build when trying `zig build` inside the `src/` dir instead of the root dir, due to hardcoding the path *from* the relative dir to `"src"`, such that the compiler will try to look for `$PROJECT_ROOT/src/src/<path>` instead of `$PROJECT_ROOT/src/<path>`. I tried to fix all such instances of `"src"` with the parent directory of `@src().file`.

Also, exposed the `Options` type so people can do more things with it.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -9,3 +9,6 @@ pub fn build(b: *std.Build) !void {
 // expose helper functions to user's build.zig
 pub const addRaylib = raylib.addRaylib;
 pub const addRaygui = raylib.addRaygui;
+
+// expose options for compiling
+pub const RaylibOptions = raylib.Options;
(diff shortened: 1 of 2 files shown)


raysan5:
@sagehane Thank you very much for the review!

Not-Nik:
This breaks the build for me on Zig 0.13, because the path is no longer relative. Platform is macOS 14.6.1

raysan5:
@sagehane please, could you review @Not-Nik issue?

sagehane:
Yes, I'll try. I'm sorry, seems this breaks for 0.12.0 too (which is the advertised supported version). There were other confusing aspects of the build system so I'll take a look.

PR #4377, opened 2024-10-11
superfluous code removed

johannes-el:


raysan5:
@johannes-el thanks but I try to avoid/minimize early returns in raylib, so the result variables.

johannes-el:
Ok I understand. Albeit the MB to MiB change is reasonable in my opinion. Note: This was the **2.222** closed pull request :)

PR #4380, opened 2024-10-13
build.zig: Fix `@src` logic and a few things

sagehane:
Motivation and context:
https://github.com/raysan5/raylib/pull/4375#issuecomment-2408950056
https://github.com/raysan5/raylib/pull/4375#issuecomment-2408998315

@Not-Nik, I'm pretty sure this works but can you check just in case?

PR #4381, opened 2024-10-13
build.zig: Fix `@src` logic and a few things

sagehane:
This should be the proper version of https://github.com/raysan5/raylib/pull/4380. Compared to the previous version, line 68 has been changed and I deleted some printf expression I forgot to remove.


raysan5:
@sagehane thanks for the review!

PR #4382, opened 2024-10-14
Make sure DrawRectangleLinesEx accounts for square tiles

Joshalosh:
Currently when the thickness is greater than the width or the height a comparison between the rec.width and rec.height is made. The line thickness then becomes half of whatever is smaller. At the moment this doesn't account for if the rec.width and rec.height are equal. Which could potentially be pretty common for people making tile-based games. ( I accidentally ran into this )


Joshalosh:
You're very welcome :)

PR #4383, opened 2024-10-14
build.zig: Fix options.config type issue

yuval-herman:
Fixes the options struct config optional type that is not supported from 0.13.
I simply changed it into a slice type and changed the check into a length check.


raysan5:
@yuval-herman Thanks for the review!

PR #4390, opened 2024-10-16
[rcore] added sha1 implementation

CrackedPixel:
I kept it as close to the MD5 style as much as possible, using the [pseudo code](https://en.wikipedia.org/wiki/SHA-1) shared via discord for reference


raysan5:
@CrackedPixel thanks for the addition! 👍😄

PR #4392, opened 2024-10-17
[rcore] [SDL] Fix #4388

### LINKED ISSUE #4388: [rcore][SDL] `GetWorkingDirectory()` returns application directory after initialization
This is an issue with the SDL backend.

Before Raylib is inititalized, GetWorkingDirectory() returns the correct working directory.  After initialization, it instead returns the same as GetApplicationDirectory().  This also causes resources not to load if your application is running from somewhere other than the current directory.

asdqwe:
Removes the `CHDIR` so `PLATFORM_DESKTOP_SDL` is in line with `PLATFORM_DESKTOP_GLFW` ([ref](https://github.com/raysan5/raylib/blob/d6399622a0280bc8ea458104b946f1860d928156/src/platforms/rcore_desktop_glfw.c#L1638-L1641)). Fixes #4388.


raysan5:
@asdqwe Nice! Thank you very much for the review!

PR #4395, opened 2024-10-19
[rtextures] ImageDraw(): Don't try to blend images without alpha

Not-Nik:
This PR enables ImageDraw() to work with all formats that are *not* compressed and *don't* have an alpha channel. This specifically allows loading cubemaps again, which before did not work for any layouts except `CUBEMAP_LAYOUT_LINE_VERTICAL`, possibly since 589d606000a61aaf9d7f14089a83a2bb30c3e714, because the function attempts to restructure cubemap images into a vertical line.


raysan5:
@Not-Nik Thanks for the review!

PR #4399, opened 2024-10-20
[rl_gputex] Correctly load mipmaps from DDS files

### LINKED ISSUE #3446: [gputex] Compressed DDS textures loading problems
### Issue description

Raylib fails to load specific compressed DDS texture files, resulting in crashing or rendering black textures (with no console output about it). The textures that i'm trying to load are valid as they can be opened in other programs. The textures still don't work even after opening them in an image editing program and resaving (with the same compression settings). So the issue seems to be in raylib itself.

After a bit of searching, i found [this](https://github.com/raysan5/raylib/blob/master/src/external/rl_gputex.h#L264) bit of code, which crashed the app when loading some textures. The code computes the `data_size` using some `header->pitch_or_linear_size` variable. After taking a look at the texture in the hex editor, it appears that `data_size` is equal to `0x400000 * 2 = 0x800000` (bc the texture has 12 > 1 mipmaps) which is bigger than the file itself (which is `0x5555F0`)! 

image
^ this is the `results in crashing but works after fixing the crash.dds` file from the archive below.

So i fixed it by doing this (0x80 = "DDS " 4 char code + sizeof(dds_header)):

image

Now no textures were crashing anymore, but some of them just appeared black (`results in crashing, black after fixing the crash.dds`), i have no idea why it may happen and how to fix it.

The raylib log says that everything is loaded successfully (with all textures) (i have a custom log callback):
```
[DEBUG] [RL] FILEIO: [...path...\forraylib\results in crashing, black after fixing the crash.dds] File loaded successfully
[DEBUG] [RL] IMAGE: Data loaded successfully (2048x1024 | DXT1_RGB | 12 mipmaps)
[DEBUG] [RL] TEXTURE: [ID 23] Texture loaded successfully (2048x1024 | DXT1_RGB | 12 mipmaps)
```

Here is an archive of the textures i've tested (check the file names): [forraylib.zip](https://github.com/raysan5/raylib/files/13056505/forraylib.zip)

### Environment

Windows 11, an nvidia videocard, OpenGL `3.3.0 NVIDIA 537.13`, GLSL `3.30 NVIDIA via Cg compiler`, raylib `5.0-dev` (latest commit)

### Issue Screenshot

Loading `results in crashing, black after fixing the crash.dds`:

image

Loading `results in crashing but works after fixing the crash.dds`:

image

### Code Example

```cpp
Mesh mesh  = {0};

// ... setting up mesh (vertices, normals, uvs, etc.), no issues here

UploadMesh(&mesh, false);
auto model = LoadModelFromMesh(mesh);
auto tex = LoadTexture("path to texture"); // FILEIO log says that the file is loaded successfully
model.materials[0].maps[MATERIAL_MAP_DIFFUSE].texture = tex;

// ... using the 'model' later:

DrawModel(model, {0, 0, 0}, 1.f, WHITE);
```

Thanks in advance! My discord if needed: jaan2897

Not-Nik:
Correctly calculate the size of all mipmaps when loading compressed and uncompressed DDS files. This properly fixes #3446 and cleans up the mess #3483 made. The root issue arose from the fact that mipmap sizes were calculated incorrectly. I'm going to write the math down here in case anyone cares.

## How the original mistake happened

When working with "normal" linear numbers, the following applies:

```c
(x) + (x / 2) + (x / (2 ^ 2)) + (x / (2 ^ 3)) + ... + (1)
```

is equal to

```c
(x + x) - 1
```

So, with actual numbers

```c
1024 + 512 + 256 + 128 + 64 + 32 + 16 + 8 + 4 + 2 + 1 = (1024 + 1024) - 1 = 2047
```

While mipmaps dimensions do half for every level, i.e. level0 is 1024x1024, level1 is 512x512, and so on; this actually means that the number of pixels (and thus the number of bytes) is quarted.

While, in plain notation, the formula looks quite similar, the following does **NOT** hold:

```c
(x^2) + ((x/2) ^ 2) + ((x / (2^2)) ^ 2) + ... + (1) = ((x^2) + (x^2) - 1)
```

So

```c
(1024^2) + (512^2) + (256^2) + (128^2) + ... + 1 != (1024 + 1024) - 1 = 2047
```

## How I fixed it

This does not mean, however, that there is no shorthand for this kind of calculation. The actual formula is

```c
((x * x) + ((x * x) / 3)) - 1
```

The proof is out of the scope of this PR, but intuitively, we do the same calculation as with linear numbers, just that we skip every other summand. Starting from one, clearly:

```c
(1^2)                 =  1 = (1) + floor((1) / 3)

(1^2) + (2^2)         =  5 = (2^2) + floor((2^2) / 3)

// With linear numbers (**wrong**): 4 + 2 + 1 = 7
// With every other term omitted : 4 + 1 = 5


(1^2) + (2^2) + (4^2) = 21 = (4^2) + floor((4^2) / 3)

// With linear numbers (**wrong**): 16 + 8 + 4 + 2 + 1 = 31
// With every other term omitted : 16 + 4 + 1 = 21
```

Here we can see that the omitted terms, always add up to one third of the actual result (with rounding)

Analogously, the same logic applies to non-square textures, so the size with mipmaps is

```c
(width * height * bpp) + (width * height * bpp / 3)
// or
(top_level_size) + (top_level_size / 3)
```

### DIFF
--- a/src/external/rl_gputex.h
+++ b/src/external/rl_gputex.h
@@ -185,6 +185,7 @@ void *rl_load_dds_from_memory(const unsigned char *file_data, unsigned int file_
                 if (header->ddspf.flags == 0x40)        // No alpha channel
                 {
                     int data_size = image_pixel_size*sizeof(unsigned short);
+                    if (header->mipmap_count > 1) data_size = data_size + data_size / 3;
                     image_data = RL_MALLOC(data_size);
 
                     memcpy(image_data, file_data_ptr, data_size);
@@ -196,6 +197,7 @@ void *rl_load_dds_from_memory(const unsigned char *file_data, unsigned int file_
                     if (header->ddspf.a_bit_mask == 0x8000)     // 1bit alpha
                     {
                         int data_size = image_pixel_size*sizeof(unsigned short);
+                        if (header->mipmap_count > 1) data_size = data_size + data_size / 3;
                         image_data = RL_MALLOC(data_size);
 
                         memcpy(image_data, file_data_ptr, data_size);
@@ -215,6 +217,7 @@ void *rl_load_dds_from_memory(const unsigned char *file_data, unsigned int file_
                     else if (header->ddspf.a_bit_mask == 0xf000)   // 4bit alpha
                     {
                         int data_size = image_pixel_size*sizeof(unsigned short);
+                        if (header->mipmap_count > 1) data_size = data_size + data_size / 3;
                         image_data = RL_MALLOC(data_size);
 
                         memcpy(image_data, file_data_ptr, data_size);
@@ -236,6 +239,7 @@ void *rl_load_dds_from_memory(const unsigned char *file_data, unsigned int file_
             else if ((header->ddspf.flags == 0x40) && (header->ddspf.rgb_bit_count == 24))   // DDS_RGB, no compressed
             {
                 int data_size = image_pixel_size*3*sizeof(unsigned char);
+                if (header->mipmap_count > 1) data_size = data_size + data_size / 3;
                 image_data = RL_MALLOC(data_size);
 
                 memcpy(image_data, file_data_ptr, data_size);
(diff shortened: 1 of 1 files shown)


raysan5:
@Not-Nik Looks good! Thank you very much for the review!

PR #4400, opened 2024-10-20
correction of comments

Bigfoot71:
The indication of locations for bone ids and bone weights did not correspond to their default values ​​in config.h


raysan5:
@Bigfoot71 Thanks for the review! I actually double checked it yesterday somewhere else!

PR #4401, opened 2024-10-20
[examples] removed extra update command

CrackedPixel:
Brought up by another user and then confirmed by Jeff, this call to `UpdateModelAnimationBoneMatrices` is not needed as it's performed in the loop before any rendering and can be removed


raysan5:
@CrackedPixel Merging... but actually I'd keep the one removed, that's the reason why *update* is separated from *draw* code!

CrackedPixel:
@raysan5 do you want me to flip it? i can make a new PR to move the Update outside the drawing section

PR #4406, opened 2024-10-21
build.zig: Better specify Linux dependencies

sagehane:
This is an updated list of Linux dependencies that Raylib should require according to Nixpkgs. I think it's likely to be accurate but it does warrant testing. In particular, I would like to know whether the fact that `libdrm` and `libgbm` depends on `pkg-config` is a NixOS-quirk. It could also just be that NixOS is more/less faithful to upstream than other distros. In practice, what should matter is obviously the convention of major distros. I'm only touching the Zig build system as that's what I know best.

`flake.nix` used to test dependencies:
```nix
{
  inputs.flake-utils.url = "github:numtide/flake-utils";

  outputs = {
    self,
    nixpkgs,
    flake-utils,
  }:
    flake-utils.lib.eachDefaultSystem (
      system: let
        pkgs = nixpkgs.legacyPackages.${system};
      in {
        #devShells.default = pkgs.stdenv.mkDerivation {
        devShells.default = pkgs.stdenvNoCC.mkDerivation {
          name = "raylib";

          nativeBuildInputs = with pkgs; [
            pkg-config # Needed with stdenvNoCC

            #-Dlinux_display_backend=Wayland
            wayland-scanner
            libxkbcommon
            wayland

            #-Dlinux_display_backend=X11
            xorg.libX11
            xorg.libXcursor
            xorg.libXi
            xorg.libXinerama
            xorg.libXrandr

            #-Dplatform=drm
            pkg-config # Needed for libdrm
            libdrm
            libglvnd
            mesa
          ];
        };
      }
    );
}
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -186,16 +186,16 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
         .linux => {
             if (options.platform != .drm) {
                 try c_source_files.append("src/rglfw.c");
-                raylib.linkSystemLibrary("GL");
-                raylib.linkSystemLibrary("rt");
-                raylib.linkSystemLibrary("dl");
-                raylib.linkSystemLibrary("m");
-
-                raylib.addLibraryPath(.{ .cwd_relative = "/usr/lib" });
-                raylib.addIncludePath(.{ .cwd_relative = "/usr/include" });
                 if (options.linux_display_backend == .X11 or options.linux_display_backend == .Both) {
                     raylib.defineCMacro("_GLFW_X11", null);
                     raylib.linkSystemLibrary("X11");
+                    raylib.linkSystemLibrary("Xcursor");
+                    raylib.linkSystemLibrary("Xext");
+                    raylib.linkSystemLibrary("Xfixes");
+                    raylib.linkSystemLibrary("Xi");
+                    raylib.linkSystemLibrary("Xinerama");
+                    raylib.linkSystemLibrary("Xrandr");
+                    raylib.linkSystemLibrary("Xrender");
                 }
 
                 if (options.linux_display_backend == .Wayland or options.linux_display_backend == .Both) {
@@ -208,8 +208,6 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
                     };
                     raylib.defineCMacro("_GLFW_WAYLAND", null);
                     raylib.linkSystemLibrary("wayland-client");
-                    raylib.linkSystemLibrary("wayland-cursor");
-                    raylib.linkSystemLibrary("wayland-egl");
                     raylib.linkSystemLibrary("xkbcommon");
                     waylandGenerate(b, raylib, "wayland.xml", "wayland-client-protocol");
                     waylandGenerate(b, raylib, "xdg-shell.xml", "xdg-shell-client-protocol");
@@ -228,14 +226,8 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
                     raylib.defineCMacro("GRAPHICS_API_OPENGL_ES2", null);
                 }
 
-                raylib.linkSystemLibrary("EGL");
-                raylib.linkSystemLibrary("drm");
                 raylib.linkSystemLibrary("gbm");
-                raylib.linkSystemLibrary("pthread");
-                raylib.linkSystemLibrary("rt");
-                raylib.linkSystemLibrary("m");
-                raylib.linkSystemLibrary("dl");
-                raylib.addIncludePath(.{ .cwd_relative = "/usr/include/libdrm" });
+                raylib.linkSystemLibrary2("libdrm", .{ .use_pkg_config = .force });
 
                 raylib.defineCMacro("PLATFORM_DRM", null);
                 raylib.defineCMacro("EGL_NO_X11", null);


build.zig
```diff
@@ -228,14 +227,9 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
...
-                raylib.linkSystemLibrary("drm");
-                raylib.linkSystemLibrary("gbm");
-                raylib.linkSystemLibrary("pthread");
-                raylib.linkSystemLibrary("rt");
-                raylib.linkSystemLibrary("m");
-                raylib.linkSystemLibrary("dl");
-                raylib.addIncludePath(.{ .cwd_relative = "/usr/include/libdrm" });
+                raylib.linkSystemLibrary("GLESv2");
+                raylib.linkSystemLibrary2("gbm", .{ .use_pkg_config = .force });
+                raylib.linkSystemLibrary2("libdrm", .{ .use_pkg_config = .force });
```
sagehane, line 232: While the rest of the PR is kind of bikeshedding, this part definitely isn't. Trying to add `/usr/include/libdrm` to the build system is a hack because:
1. Not all distros support FHS (like NixOS and Guix)
2. Some people may be using custom directories like `/opt`
3. The headers that build system needs according to https://github.com/raysan5/raylib/blob/5.0/src/platforms/rcore_drm.c#L63-L64 should both be in the root dir of libdrm for at least NixOS and [Arch](https://archlinux.org/packages/extra/x86_64/libdrm/files/), making this useless as far as I can tell.

@Yerdun: I would appreciate if you could tell me what distro you use and whether this PR works with:
1. Removing the `.use_pkg_config = .force` for `gbm`.
2. Removing the `.use_pkg_config = .force` for `libdrm`.
3. Renaming `libdrm` with `drm`.
  - I have no idea why this one doesn't work on NixOS, tbh

Yerdun: Using EndeavourOS (Arch based distro that directly uses the official Arch repos) and Zig 0.12.1:

- [X] Removing the `.use_pkg_config = .force` for `gbm`.

- [X] Removing the `.use_pkg_config = .force` for `libdrm`.

- [ ] Renaming `libdrm` with `drm`. *(I'm interpreting this as replacing `"libdrm"` in line 232 with `"drm"`)*

```
/tmp/raylib % /usr/bin/zig build -Dplatform=drm
thread 5465 panic: pkg-config failed for library drm
/usr/lib/zig/std/Build/Step/Compile.zig:1166:50: 0x116e996 in make (build)
                                            panic("pkg-config failed for library {s}", .{system_lib.name});
                                                 ^
/usr/lib/zig/std/Build/Step.zig:182:13: 0x11491ce in make (build)
    s.makeFn(s, prog_node) catch |err| switch (err) {
            ^
/usr/lib/zig/compiler/build_runner.zig:951:31: 0x10fb3e1 in workerMakeOneStep (build)
    const make_result = s.make(&sub_prog_node);
                              ^
/usr/lib/zig/std/Thread/Pool.zig:93:39: 0x10c1d52 in runFn (build)
            @call(.auto, func, closure.arguments);
                                      ^
/usr/lib/zig/std/Thread/Pool.zig:132:18: 0x1183ee4 in worker (build)
            runFn(&run_node.data);
                 ^
/usr/lib/zig/std/Thread.zig:408:13: 0x1147a0d in callFn__anon_14565 (build)
            @call(.auto, f, args);
            ^
/usr/lib/zig/std/Thread.zig:1226:30: 0x10fac17 in entryFn (build)
                return callFn(f, self.fn_args);
                             ^
/usr/lib/zig/c.zig:239:13: 0x12b6210 in clone (c)
            asm volatile (
            ^
???:?:?: 0xaaaaaaaaaaaaaaa9 in ??? (???)
Unwind information for `???:0xaaaaaaaaaaaaaaa9` was not available, trace may be incomplete
```

Since you mentioned the header files are in the same location on Nix and Arch, I can spin up a distrobox with a different distro of your choice if that would help.

sagehane: Thanks for testing, but my dumb instructions kind of made it pointless. Of course the first two would work without `.force`. What I really meant is if you could try with `.use_pkg_config = .no`.

The third should be tested *without* forcing pkg-config, so replacing the `libdrm` line with `raylib.linkSystemLibrary("drm");` is fine. It is nice to know that `libdrm` works with `pkg-config` but `drm` doesn't, although I kind of knew that as Arch clearly provides a `libdrm.pc` file.

---

>... I can spin up a distrobox with a different distro of your choice if that would help.

Maybe Fedora? Eh nvm, from my experience, they usually have weird package *names* but the internal files are probably fine.

Looking at other distros, I see nothing suspicious.

https://rpmfind.net/linux/RPM/fedora/updates/40/x86_64/Packages/l/libdrm-devel-2.4.123-1.fc40.x86_64.html.
https://pkgs.alpinelinux.org/contents?name=libdrm-dev&repo=main&branch=edge&arch=x86_64

So, yeah, I think it's quite likely to be fine if both NixOS and Arch passes.

sagehane: Okay, I understand the situation better now. First of all, I needed `pkg-config` for `gbm` because I was being dumb and using `stdenvNoCC`. It's nice as it makes me need to supply all necessary dependencies, but it also has the side effect of always requiring `pkg-config`.

Next, I now get why the original build system tries to add `/usr/include/libdrm` to the search path. It's because `xf86drm.h` depends on `drm.h`, even though `drm.h` is actually located at `libdrm/drm.h`. This is probably one of those weird upstream issues where they depend on `pkg-config` or some custom include path logic. `pkg-config` should be the "correct" way to handle this issue because it adds the expected directory to the include path:
```sh-session
$ pkg-config --cflags-only-I libdrm
-I/nix/store/c29mg2gxp1ah7w2f5vz70hyh0k6w4pxk-libdrm-2.4.122-dev/include -I/nix/store/c29mg2gxp1ah7w2f5vz70hyh0k6w4pxk-libdrm-2.4.122-dev/include/libdrm
```

I think I'm happy with this as I've now documented that there's an aspect of the build system depending on `pkg-config`.

Yerdun: Hi, sorry for the late reply, I've only found the time to take a look at this now. I'm assuming this discovery means there's no longer any need to test compiling. Sorry I didn't get to do that other thing you asked, and let me know if there's anything more I could help with

sagehane: Yeah, I think I'm satisfied as you don't seem to be getting any compile errors either, and I kind of understand the root cause of the problem now.

raysan5:
@sagehane thanks for the review!

PR #4408, opened 2024-10-21
[rlgl] Updated instanced rendering support loading

rapha-s:
Hi!

The the function loading for this was kinda broken before, with an impossible condition (unless you were using ANGLE). This fixes that and includes the GL_NV versions of the extensions, that are supported by all modern NVIDIA GPUs when running on Desktop using its GLES mode (that is a better alternative to ANGLE on systems that support it).

I opted by only marking the support flag if GL_XXX_instanced_arrays is present, but still loading the function pointers if only GL_XXX_draw_instanced is, just in case the user isn't using the vertex attribute functionality and for some reason the GPU only supports one of the extensions (although I can't find instanced of that being a thing, if one is supported, both seems to be). If that's not OK, let me know the proper wanted behavior, so I can change it.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -2472,25 +2472,37 @@ void rlLoadExtensions(void *loader)
         }
 
         // Check instanced rendering support
-        if (strcmp(extList[i], (const char *)"GL_ANGLE_instanced_arrays") == 0)         // Web ANGLE
-        {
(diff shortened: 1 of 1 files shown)


rapha-s:
Oh, ok!  (≧∀≦)ゞ Sorry that the formatting got in wrong!

raysan5:
@rapha-s No problem! Thanks! 👍 😄

PR #4413, opened 2024-10-22
moved update out of draw area

CrackedPixel:
Moved the update stuff out of the drawing area. Can close PR if not necessary

Discussion from https://github.com/raysan5/raylib/pull/4401


raysan5:
@CrackedPixel Excellent! That's how I like the code organized! 😄

PR #4415, opened 2024-10-22
[rcore] [web] Simplify `EmscriptenResizeCallback()`

asdqwe:
So it doesn't depend on macro'd JS functions. Changes can be tested with:
```
// Note: usage of `minshell.html` is recommended since it won't force canvas CSS overrides

#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

    SetWindowState(FLAG_WINDOW_RESIZABLE);

        BeginDrawing();
        ClearBackground(RAYWHITE);

        DrawText("Resize the browser window manually", 20, 20, 20, BLACK);
        DrawText(TextFormat("canvas size: %i x %i", GetScreenWidth(), GetScreenHeight()), 20, 40, 20, BLACK);

        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -324,8 +324,8 @@ void MaximizeWindow(void)
         platform.unmaximizedWidth = CORE.Window.screen.width;
         platform.unmaximizedHeight = CORE.Window.screen.height;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@asdqwe Nice! Thanks for the review!

PR #4407, opened 2024-10-21
[build][Zig] Make more flags from `src/config.h` configurable

sagehane:
This PR does not depend on #4406.

Currently, only flags containing the word `SUPPORT` is override-able via `-Dconfig`. This was the case as there was no code to check for what value they were set to, and all `SUPPORT` flags have a value of 1. This is also not ideal because it'll cause unexpected behaviour if the convention of the current `src/config.h` changes.

With this PR, everything that isn't defined between `#ifdef` or `#ifndef` and `#endif` can be passed or configured. I didn't bother evaluating flags between those as they already seem to be meant to be configured and truly evaluating them would be very annoying.

List of flags that are now passed by default when passing `-Dconfig`:
<details><pre>
-DSUPPORT_MODULE_RSHAPES=1
-DSUPPORT_MODULE_RTEXTURES=1
-DSUPPORT_MODULE_RTEXT=1
-DSUPPORT_MODULE_RMODELS=1
-DSUPPORT_MODULE_RAUDIO=1
-DSUPPORT_CAMERA_SYSTEM=1
-DSUPPORT_GESTURES_SYSTEM=1
-DSUPPORT_RPRAND_GENERATOR=1
-DSUPPORT_MOUSE_GESTURES=1
-DSUPPORT_SSH_KEYBOARD_RPI=1
-DSUPPORT_WINMM_HIGHRES_TIMER=1
-DSUPPORT_PARTIALBUSY_WAIT_LOOP=1
-DSUPPORT_SCREEN_CAPTURE=1
-DSUPPORT_GIF_RECORDING=1
-DSUPPORT_COMPRESSION_API=1
-DSUPPORT_AUTOMATION_EVENTS=1
-DMAX_FILEPATH_CAPACITY=8192
-DMAX_FILEPATH_LENGTH=4096
-DMAX_KEYBOARD_KEYS=512
-DMAX_MOUSE_BUTTONS=8
-DMAX_GAMEPADS=4
-DMAX_GAMEPAD_AXIS=8
-DMAX_GAMEPAD_BUTTONS=32
-DMAX_GAMEPAD_VIBRATION_TIME=2.0f
-DMAX_TOUCH_POINTS=8
-DMAX_KEY_PRESSED_QUEUE=16
-DMAX_CHAR_PRESSED_QUEUE=16
-DMAX_DECOMPRESSION_SIZE=64
-DMAX_AUTOMATION_EVENTS=16384
-DRL_DEFAULT_BATCH_BUFFERS=1
[...]

PR #4417, opened 2024-10-22
build.zig: Re-add OpenGL to Linux deps

sagehane:
Of course, being dumb, I forgot to test the damn thing with an actual binary and got hit by missing OpenGL issues at runtime. *Sigh*

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -186,8 +186,10 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
         .linux => {
             if (options.platform != .drm) {
                 try c_source_files.append("src/rglfw.c");
+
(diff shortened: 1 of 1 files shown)


raysan5:
@sagehane Thanks for the update... please, note that I'm not testing all those changes and a new raylib release is around the corner, build.zig should be as much stable as possible...

sagehane:
I'm very sorry about all the sloppiness on my part. AFAIK, there is one more bug in the build system at https://github.com/raysan5/raylib/pull/4407#pullrequestreview-2382221857 and that should be everything. I'll extract that to a separate PR and focus on reviewing the PRs more carefully and understanding this repo until the next release. Everything else would classify as enhancements over the existing setup and wouldn't be required. Aside from that, the fact that I haven't been able to check how Raygui works with the current build system is slightly concerning, but I don't think I would have introduced breaking errors.

PR #4423, opened 2024-10-23
Grammar fix in CONTRIBUTING.md

cypressru:
"Can you write some tutorial/example?"
changed to 
"Can you write some tutorials/examples?"

Alternatively we could do 
"Can you write a tutorial/example?"

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,7 @@ Hello contributors! Welcome to raylib!
 Do you enjoy raylib and want to contribute? Nice! You can help with the following points:
 
 - `C programming` - Can you write/review/test/improve the code? 
-- `Documentation/Tutorials/Example` - Can you write some tutorial/example?
+- `Documentation/Tutorials/Example` - Can you write some tutorials/examples?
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -5,7 +5,7 @@ Hello contributors! Welcome to raylib!
 Do you enjoy raylib and want to contribute? Nice! You can help with the following points:
 
 - `C programming` - Can you write/review/test/improve the code? 
-- `Documentation/Tutorials/Example` - Can you write some tutorial/example?
+- `Documentation/Tutorials/Example` - Can you write some tutorials/examples?
```
CrackedPixel, line 8: > Alternatively we could do
> "Can you write a tutorial/example?"


I like that version better

Secondly, the `Documentation/Tutorials/Example` can probably be changed to `Examples` (plural) as well

(imo)

cypressru: I actually like "Can you write a tutorial/example?" better as well.

raysan5:
@cypressru thanks for the review!

PR #4425, opened 2024-10-23
[RTEXTURES] Remove the panorama cubemap layout option

JeffM2501:
The panorama cube map option was not implemented, so this PR removes it from the enumeration in raylib.h so that users do not get confused and try to use it.
I left a todo in the code for some aspiring developer to finish.


raysan5:

PR #4428, opened 2024-10-24
[examples] Use free camera mode in model shader example

juliohq:
This PR changes the model shader example to use free camera mode (`CAMERA_FREE`).


raysan5:
@juliohq thanks for the review

PR #4429, opened 2024-10-24
[rtextures/rlgl] Load mipmaps for cubemaps

Not-Nik:
Load all mipmaps for cubemaps to the GPU and enable the proper filters for them. This is especially important for image based lighting (IBL), where for different roughness levels, different pre-filtered environment maps are stored in mipmaps. Calculating these filters is rather expensive, so it is often done offline and the pre-filtered cubemaps are shipped with the game.

The implementation did require some reworking; notably `rlLoadTextureCubemap` gains a new `mipmapCount` parameter, and `ImageMipmaps` no longer regenerates already existing mipmaps.

I structured the PR to be able to just be rebased onto master if you want all the different changes to reflect in the tree.

### DIFF
--- a/examples/models/models_skybox.c
+++ b/examples/models/models_skybox.c
@@ -191,7 +191,7 @@ static TextureCubemap GenTextureCubemap(Shader shader, Texture2D panorama, int s
     // STEP 1: Setup framebuffer
     //------------------------------------------------------------------------------------------
     unsigned int rbo = rlLoadTextureDepth(size, size, true);
-    cubemap.id = rlLoadTextureCubemap(0, size, format);
+    cubemap.id = rlLoadTextureCubemap(0, size, format, 1);
 
     unsigned int fbo = rlLoadFramebuffer();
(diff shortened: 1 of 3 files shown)


raysan5:
@Not-Nik Thanks for the improvement!

raysan5:
@Not-Nik I'm afraid this PR breaks `models_skybox` example, it crashes on execution at `line 91: UnloadImage(img);`. Please, can you review it?

raysan5:
I'm afraid I'm reverting it in the meantime, I'm planning a release in the following hours and I prefer to avoid breaks.

raysan5:
I just put a quick fix to avoid the crash in the `models_skybox` example but I noticed this change generates artifacts in the texture.

Not-Nik:
The fix should comment out both mipmap generations, other wise OpenGL will try to read mipmaps that aren't properly generated

Not-Nik:
I found the bug, I'll create a fixup PR in 20 minutes

PR #4430, opened 2024-10-24
[Examples] Add shadow map example to MSVC projects

JeffM2501:
The shadow map example did not have a visual studio project. This PR adds it in.


raysan5:

PR #4431, opened 2024-10-24
[Shaders] Have the base fragment shader use the vertex color

JeffM2501:
This PR has the base fragment shader for the examples use the vertex color in addition to the tint color. This provides a better example of what the default shader is doing.
I also added some more comments to the finalColor computation to explain what is happening.

If having the base shader ignore vertex color is intended in order to keep the shader simple, please reject this PR.

### DIFF
--- a/examples/shaders/resources/shaders/glsl330/base.fs
+++ b/examples/shaders/resources/shaders/glsl330/base.fs
@@ -20,6 +20,9 @@ void main()
 
     // NOTE: Implement here your fragment shader code
 
-    finalColor = texelColor*colDiffuse;
+    // final color is the color from the texture 
+    //    times the tint color (colDiffuse)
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 It was not added because, afair, usual meshes (obj) did not support vertex colors, so the base shader did not work as expected if multiplied by fragColor.

It was omitted for that reason but it can be at least commented the case, or add a commented line considering it, just in case vertColor is provided.

JeffM2501:
I believe all the meshes should support vertex color now, as well as anything from the batch system, so I don't think this will cause an issue.
The default shaders in rlgl do this same computation.
So if there was an issue with fragColor, it would show up in all the examples.

raysan5:
@JeffM2501 Yeah, agree... it was implemented long time ago so probably it was already addressed in other places of the code.

Issue #4371, opened 2024-10-07
[rcore] Overview of Wayland, Xwayland and X11 support on Linux


asdqwe:
> [!NOTE]
> \- This post contains my opinions and observations, YMMV.
> \- Posting as an issue because it's a recurrent subject, but feel free to close this.

### Context:
1. [Many Linux distributions](https://en.wikipedia.org/wiki/Wayland_(protocol)#Desktop_Linux_distributions) are undergoing a migration from the [X Window System (X11)](https://en.wikipedia.org/wiki/X_Window_System) to the [Wayland communication protocol](https://en.wikipedia.org/wiki/Wayland_(protocol)).
2. However, a series of issues are emerging due to the architectural differences between both systems. E.g.: Wayland transfering some window control (like positioning) from the application to the window manager.
3. These architectural choices make Wayland diverge from how other desktop operating systems handle windows.
4. Since supporting Wayland requires rebuilding all graphical applications for it, [Xwayland](https://wayland.freedesktop.org/xserver.html) was provided as a backward compatibility solution to run X clients under Wayland.

### Native Wayland limitations on GLFW:
Referencing the [GLFW documentation](https://www.glfw.org/docs/latest/group__window.html):

| GLFW function: | Wayland limitation: | Use in raylib: |
| :--- | :--- | :--- |
| `glfwSetWindowIcon` | _"Wayland: There is no existing protocol to change an icon, the window will thus inherit the one defined in the application's desktop file. This function will emit GLFW_FEATURE_UNAVAILABLE."_ | `SetWindowIcon` |
| `glfwGetWindowPos` | _"Wayland: There is no way for an application to retrieve the global position of its windows. This function will emit GLFW_FEATURE_UNAVAILABLE."_ | `GetCurrentMonitor`<br>`GetWindowPosition` |
| `glfwSetWindowPos` | _"Wayland: There is no way for an application to set the global position of its windows. This function will emit GLFW_FEATURE_UNAVAILABLE."_ | `ToggleBorderlessWindowed`<br>`SetWindowPosition`<br>`SetWindowMonitor` |
| `glfwSetWindowSizeLimits` | _"Wayland: The size limits will not be applied until the window is actually resized, either by the user or by the compositor."_ | `SetWindowMinSize`<br>`SetWindowMaxSize` |
| `glfwSetWindowAspectRatio` | _"Wayland: The aspect ratio will not be applied until the window is actually resized, either by the user or by the compositor."_ | Not used. |
[...]

PR #4432, opened 2024-10-24
[build] CMake: Don't build for wayland by default

Peter0x44:
This is to align with the behavior of #4369, see #4371 for rationale on disabling Wayland by default.


raysan5:

PR #4435, opened 2024-10-25
[shapes] Add `shapes_rectangle_advanced ` example implementing a `DrawRectangleRoundedGradientH` function

evertonse:
Edit: Add the ``shapes_rectangle_advanced`` examples that implements ``DrawRectangleRoundedGradientH``  as sugested.
#### Show Case:
Showcase Image

Tried to follow all convetions I could find for implementing an example.


raysan5:
@evertonse Thanks for the improvement but I'm afraid I'm not adding this function to raylib API, users with that level of requirement should implement the function on their end instead. There are too just many possibilities for a rectangle to cover all requirements:

 - Border, border width
 - Rounded corners, every one with independent value
 - Gradient, vertical/horizontal/specific-angle, multi-color
 - Corner vertex color with interpolation
 - Internal padding

A function to cover all those possibilities would require too many parameters and keeping adding more functions, I don't think is the best approach.

But, you can provide a `shapes_rectangle_advanced` example implementing this custom function using `rlgl`. It would be very useful for users with those advance rectangle drawing rectangles to have a reference implementation and learn how to do it on their end.

evertonse:
> But you can provide a `shapes_rectangle_advanced` example implementing this custom function using `rlgl`. It would be very useful for users with those advanced rectangle drawing capabilities to have a reference implementation and learn how to do it on their end.

Thank you for the quick response, Raysan. I've done just that and updated the PR to accommodate it. Let me know if you need any further adjustments.

Can I ask something? Would you be willing to accept a PR for getting an image from the clipboard that only works for Windows? I have implemented something like this:

```c
// The user would have to call free on the pointer if the clipboard is not an image; it returns NULL.
char* pixels = GetClipboardImage(&width, &height, &data_size);
img = LoadImageFromMemory(".bmp", pixels, data_size);
```
 I can easily change it into this:
```c
Image image = GetClipboardImage();
```

The problem is that I don't have enough knowledge to make it work on X11 or Wayland, just for Windows. I also don't know the philosophy of the raylib project regarding having a function that only works for a single platform. I was thinking that maybe someone else could implement it for more platforms in an incremental way.

---

raysan5:
@evertonse Thank you very much for the review and all the work put on it! Sorry for not mentioning it earlier but would you mind making the example be 800x450 resolution, all raylib examples follow that same size to better fit in the web.

About the `GetClipboardImage()`, I'd love to have that function, together with `SetClipboardImage(Image image)` but that functionality should be provided by GLFW library... afaik, SDL already provides it...

Probably the function should be `Image GetClipboardImage()`... despite it couples `rcore` with `rtext` module...

In any case, feel free to send a PR for further review.

raysan5:
@evertonse thank you very much for this great example! 😄

PR #4437, opened 2024-10-26
Fix example makefile

NishiOwO:


NishiOwO:
wait a sec, broke git
will remake pr

PR #4439, opened 2024-10-26
[rtextures] LoadTextureCubemap(): Copy image before generating mipmaps, to avoid dangling re-allocated pointers

Not-Nik:
`LoadTextureCubemap()`, takes `image` as a copy, not a reference.`ImageMipmaps()` re-allocates the `data` pointer of an image, making the old pointer invalid. This means that calls to `LoadTextureCubemap()` would almost always render the image unusable, resulting in a use-after-free bug, e.g. when unloading the image.

My apologies, fixup for #4429


raysan5:
@Not-Nik thanks for the review. Did you verify the `models_skybox` artifacts on the corners?

Not-Nik:
Yes, both with and without HDR works for me!

PR #4440, opened 2024-10-26
Fix OpenGL

NishiOwO:
Package one break ; removed /usr/pkg flags, it is used by only NetBSD anyway.


raysan5:
@NishiOwO thanks for the review!

PR #4444, opened 2024-10-26
Add an assert on custom frame control support

Mercotui:
This prevents users from getting a attacked by angry frames when they try this example without reading the warnings in the source code.

### DIFF
--- a/examples/core/core_custom_frame_control.c
+++ b/examples/core/core_custom_frame_control.c
@@ -27,12 +27,20 @@
 ********************************************************************************************/
 
 #include "raylib.h"
+#include "config.h"
+#include "assert.h"
 
 //------------------------------------------------------------------------------------
 // Program main entry point
 //------------------------------------------------------------------------------------
 int main(void)
 {
+    #if !defined(SUPPORT_CUSTOM_FRAME_CONTROL)
+    // This examaple requires the SUPPORT_CUSTOM_FRAME_CONTROL feature.
+    // Please edit your raylib config.h and recompile raylib.
+    assert(false);
+    #endif
+
     // Initialization
     //--------------------------------------------------------------------------------------
     const int screenWidth = 800;


examples/core/core_custom_frame_control.c
```diff
@@ -27,12 +27,20 @@
 ********************************************************************************************/
 
 #include "raylib.h"
+#include "config.h"
+#include "assert.h"
```
JeffM2501, line 31: Since this is a system header it should use <> not ""

examples/core/core_custom_frame_control.c
```diff
@@ -27,12 +27,20 @@
...
 //------------------------------------------------------------------------------------
 // Program main entry point
 //------------------------------------------------------------------------------------
 int main(void)
 {
+    #if !defined(SUPPORT_CUSTOM_FRAME_CONTROL)
+    // This examaple requires the SUPPORT_CUSTOM_FRAME_CONTROL feature.
+    // Please edit your raylib config.h and recompile raylib.
+    assert(false);
+    #endif
```
JeffM2501, line 42: Why is assert needed? would not a printf and return -1 here be the same?
assert is not use any other raylib code.

raysan5:
@Mercotui thanks but I'm not adding two extra libraries and an assert just because the user can not read the example instructions.

examples/core/core_custom_frame_control.c
Mercotui: true :+1:

examples/core/core_custom_frame_control.c
Mercotui: Inside the library assert might be to harsh, 
but considering this example application can not function without that define, I would use an assert.

But whatever the codebase follows is fine for me

Mercotui:
Naja as I personally experienced: 
running some examples to explore the project,
it is very unpleasant to accidentally run this one. 

The whole screen flashes black and white at a high framerate. 
Sure, that is the expected result, but it still hurts my eyes.

As for libraries: `assert.h` is ANSI-C and `config.h` is from raylib.

Anyway I could update the code to `return -1;` as Jeff mentioned.

PR #4445, opened 2024-10-27
[rcore][SDL] Fix disabled cursor mouse inputs

electrolys:
The mouse inputs gotten from GetMouseDelta on the SDL2 platform of Raylib currently do not match the GLFW platform at all, and this patch fixes that. Now First Person cameras can actually work on the SDL2 platform.

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1025,7 +1025,7 @@ void PollInputEvents(void)
     CORE.Input.Mouse.currentWheelMove.y = 0;
 
     // Register previous mouse position
-    if (platform.cursorRelative) CORE.Input.Mouse.currentPosition = (Vector2){ 0.0f, 0.0f };
+    if (platform.cursorRelative) CORE.Input.Mouse.previousPosition = (Vector2){ 0.0f, 0.0f };
     else CORE.Input.Mouse.previousPosition = CORE.Input.Mouse.currentPosition;
 
     // Reset last gamepad button/axis registered state
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1025,7 +1025,11 @@ void PollInputEvents(void)
     CORE.Input.Mouse.currentWheelMove.y = 0;
 
     // Register previous mouse position
-    if (platform.cursorRelative) CORE.Input.Mouse.previousPosition = (Vector2){ 0.0f, 0.0f };
+    if (platform.cursorRelative)
+    {
+        CORE.Input.Mouse.previousPosition = (Vector2){ 0.0f, 0.0f };
+        CORE.Input.Mouse.currentPosition = (Vector2){ 0.0f, 0.0f };
+    }
     else CORE.Input.Mouse.previousPosition = CORE.Input.Mouse.currentPosition;
 
     // Reset last gamepad button/axis registered state


JeffM2501:
LGTM

asdqwe:
AFAIK `cursorRelative`'s `CORE.Input.Mouse.previousPosition` is zero'd at the `SDL_MOUSEMOTION` event: https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1235
Tested `CAMERA_FIRST_PERSON` and `GetMouseDelta()` on `core_3d_camera_first_person` at current master branch (0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc) with `SDL2-2.30.7` and it appears to be working properly here.
<details>
<summary>Test diff and screenshot:</summary>

``` diff
diff --git a/examples/core/core_3d_camera_first_person.c b/examples/core/core_3d_camera_first_person.c
index 35b18ace..9124cb8d 100644
--- a/examples/core/core_3d_camera_first_person.c
+++ b/examples/core/core_3d_camera_first_person.c
@@ -178,6 +178,8 @@ int main(void)
             DrawText("- Camera mode keys: 1, 2, 3, 4", 15, 60, 10, BLACK);
             DrawText("- Zoom keys: num-plus, num-minus or mouse scroll", 15, 75, 10, BLACK);
             DrawText("- Camera projection key: P", 15, 90, 10, BLACK);
+            Vector2 mouseDelta = GetMouseDelta();
+            DrawText(TextFormat("MOUSE DELTA %f %f", mouseDelta.x, mouseDelta.y), 15, 120, 20, BLACK);
 
             DrawRectangle(600, 5, 195, 100, Fade(SKYBLUE, 0.5f));
             DrawRectangleLines(600, 5, 195, 100, BLUE);
```

img
</details>

electrolys:
i had to zero both the currentPosition and the previous in that if statement (previously it was just current) to make the cursor act correctly
the previous position also has to be zeroed there cause otherwise the mouse delta remains after you stop moving the mouse cause there isnt a mouse movement

electrolys:
at high framerates also the mouse inputs seem correct on the current master but when you get to like 20fps they are extremely bad

asdqwe:
When `cursorRelative` is enabled ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1231)), any generated `SDL_MOUSEMOTION` event ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1229)) will always zero `previousPosition` ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1235)). When `PollInputEvents` runs ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1011)) the `previousPosition` will never be filled ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1029)) while `cursorRelative` is active ([ref](https://github.com/raysan5/raylib/blob/0e1fc33c5c85a81fb1397e1f98f6bed68519e8fc/src/platforms/rcore_desktop_sdl.c#L1028)). So `previousPosition` will continue to be zero'd. I can't see how zero'ing what that was already zero'd is changing something.

raysan5:
I can't neither see the usefulness of this change... 🤔

PR #4446, opened 2024-10-27
[raymath]add method to support only adding x or y to vector2

cmanlh:
Add below method to raymath.h for convenience

```C
// Add vector and x float value
RMAPI Vector2 Vector2AddX(Vector2 v, float x)
{
    Vector2 result = { v.x + x, v.y };

    return result;
}

// Add vector and y float value
RMAPI Vector2 Vector2AddY(Vector2 v, float y)
{
    Vector2 result = { v.x, v.y + y };

    return result;
}
```

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -264,6 +264,22 @@ RMAPI Vector2 Vector2AddValue(Vector2 v, float add)
     return result;
 }
 
+// Add vector and x float value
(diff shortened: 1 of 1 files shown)


raysan5:
@cmanlh thanks but I prefer to minimize adding functions and those wants seem semple enough to be implemented at user side.

PR #4447, opened 2024-10-27
[rcore][GLFW] Fix warning on windows in MSVC

JeffM2501:
There is an error with the forward declaration of the winMM functions when building on MSVC. Inconsistent DLL linkage.
This is not a critical warning, so this PR disables it.
This PR also removes the timeBeginPeriod declaration as it is no longer needed.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -65,8 +65,9 @@
 
     #if defined(SUPPORT_WINMM_HIGHRES_TIMER) && !defined(SUPPORT_BUSY_WAIT_LOOP)
         // NOTE: Those functions require linking with winmm library
-        unsigned int __stdcall timeBeginPeriod(unsigned int uPeriod);
+#pragma warning( disable : 4273 )
         unsigned int __stdcall timeEndPeriod(unsigned int uPeriod);
+#pragma warning(default : 4273) 
     #endif
 #endif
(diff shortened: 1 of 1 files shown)


asdqwe:
AFAIK `timeBeginPeriod` may still be necessary, as its operation changed on _Windows 10, version 2004_ (see _Remarks_ section, third paragraph, of the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod)), which was on [May 2020](https://en.wikipedia.org/wiki/Windows_10,_version_2004), not that long ago. There's a lot of hardware still running older versions of Windows (e.g.: automation, industry, etc).

JeffM2501:
> AFAIK `timeBeginPeriod` may still be necessary, as its operation changed on _Windows 10, version 2004_ (see _Remarks_ section, third paragraph, of the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod)), which was on [May 2020](https://en.wikipedia.org/wiki/Windows_10,_version_2004), not that long ago. There's a lot of hardware still running older versions of Windows (e.g.: automation, industry, etc).

It's not called in that file, thus it does not need to be forward declared, it'll still be linked by the lib if some other internal file uses it, like GLFW itself.

asdqwe:
I see, thanks.

PR #4449, opened 2024-10-29
Add src/raygui.c to .gitignore

sleeptightAnsiC:
Hi,

While building with `RAYLIB_MODULE_RAYGUI=TRUE`, Makefile generates additional build artifact here: `./src/raygui.c` Said file isn't excluded from git repo which is a bit annoying when embeding raylib inside of another repository.

This PR fixes mentioned issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,6 @@ docgen_tmp/
 
 # Parser stuff
 parser/raylib_parser
+
+# Raygui auto-generated source file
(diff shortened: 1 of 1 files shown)


raysan5:
@sleeptightAnsiC this file is geenrated in a very specific configuration and only required on some examples, I prefer not to add it to .gitignore, it could cause confusions in the future.

PR #4450, opened 2024-10-30
[rshapes]Add function to draw rotated ellipse

cmanlh:
```C
RLAPI void DrawEllipseRotation(int centerX, int centerY, float radiusH, float radiusV, Color color, float angle);
RLAPI void DrawEllipseLinesRotation(int centerX, int centerY, float radiusH, float radiusV, Color color, float angle);
```
And add example to show drawing ellipse at the same time

shapes_draw_ellipse


raysan5:
@cmanlh I'd prefer not to populate the shapes drawing functions with more specific-use-case functions, it would imply reviewing many functions to support rotations, not only ellipses. If possible, please, could you move those functions to the example to be used as a base for custom implementations, instead of being added to raylib?

cmanlh:
That's ok

Issue #3929, opened 2024-04-22
[rcore_desktop] Upon entering fullscreen, uses original window size


OilyFishMan:
### Issue description

Upon running the procedure `ToggleFullscreen()`, raylib reports "GetScreenWidth()" and "GetScreenHeight()" as values set before `SetWindowSize(x, y)`. However, without `ToggleFullscreen()`, the original values are kept for one frame then immediately restored to correct full resolution.

### Environment

Platform backend: DESKTOP (GLFW)
OS: Fedora 39 x86_64
Desktop: Gnome X11
OpenGL version: 4.6
GPU: Integrated

### Issue Screenshot

[Screencast from 2024-03-24 20-24-06.webm](https://github.com/raysan5/raylib/assets/139824102/d515f9a9-5590-4b3f-9a11-99f58d9e7282)
[Screencast from 2024-04-22 18-26-19.webm](https://github.com/raysan5/raylib/assets/139824102/57b8dc58-5b8d-4ed6-958d-af2c5db4f69c)

### Code Example

```c
#include <raylib.h>

int main(void)
{
    SetTraceLogLevel(LOG_ERROR);
    SetTargetFPS(60);
    InitWindow(1, 1, "My Application's Title");
    {
        int display = GetCurrentMonitor();
        SetWindowSize(GetMonitorWidth(display), GetMonitorHeight(display));
        ToggleFullscreen();
    }
    while (!WindowShouldClose()) {
        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

MagiRomanya:
It seems that the actual resizing of the window contents is done in the WindowSizeCallback after the first call to EndDrawing() in the game loop.
However, there is this line that prevents CORE.Window.screen.width and height to be updated in the callback:
https://github.com/raysan5/raylib/blob/2e774432090db1be697d2e4e5f49ea91ca038cd3/src/platforms/rcore_desktop.c#L1625
Commenting this line solved the issue for me, but I presume is there for a reason.

OilyFishMan:
That's really strange. To me this definitely doesn't seem like expected behavior.

raysan5:
> Commenting this line solved the issue for me, but I presume is there for a reason.

There should probably be a comment there explaining that reason... I don't know at the moment...

archewashi:
I ran the sample code under msys2 and Fedora.
[...]

PR #4451, opened 2024-10-30
fix the issue with GetScreenWidth/GetScreenHeight that was identified on other platforms

archewashi:
Like #4074 I found the same defect on other platforms. 

P.S. I'm not sure if there will be an issue, but logically it should be added.
#3929


archewashi:
#3929

raysan5:
@archewashi thanks for the review!

Tottel:
Hey, this broke compiling for WebAssembly :(

PR #4452, opened 2024-10-30
[rcore] [web] Fix `SetWindowSize()` for `PLATFORM_WEB`

asdqwe:
Reverts #4451 for `PLATFORM_WEB`.

The window size is properly set by `WindowSizeCallback` ([ref](https://github.com/raysan5/raylib/blob/3bad05aa7bd319eaaae937928b396b46f4886e3e/src/platforms/rcore_web.c#L1394)) after updating the viewport, render and fbo.

<details>
<summary>Test case:</summary>

```
// Note: usage of `minshell.html` is recommended since it won't force canvas CSS overrides

#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

    SetWindowState(FLAG_WINDOW_RESIZABLE);

    while (!WindowShouldClose()) {

        if (IsKeyPressed(KEY_ONE)) SetWindowSize(400, 200);

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText(TextFormat("window size %i x %i", GetScreenWidth(), GetScreenHeight()), 20, 20, 20, BLACK);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

PR #4453, opened 2024-10-31
fix typo in #4451

archewashi:
I had a typo that caused the compilation to fail.

I believe it should be added because the timing of when `WindowSizeCallback` is called may be platform-dependent, similar to #3929.

<details>

<summary>Test case same as #4452</summary>

```c
// Note: usage of `minshell.html` is recommended since it won't force canvas CSS overrides

#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

    SetWindowState(FLAG_WINDOW_RESIZABLE);

        if (IsKeyPressed(KEY_ONE)) SetWindowSize(400, 200);

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText(TextFormat("window size %i x %i", GetScreenWidth(), GetScreenHeight()), 20, 20, 20, BLACK);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

</details>

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -667,6 +667,9 @@ void SetWindowMaxSize(int width, int height)
 // Set window dimensions
 void SetWindowSize(int width, int height)
 {
+    CORE.Window.screen.width = width;
+    CORE.Window.screen.height = height;
+
     glfwSetWindowSize(platform.handle, width, height);
 }
 


asdqwe:
1. There's more involved in resizing a window (canvas on web) than just setting a `CORE.Window.screen` value. The viewport, render and fbo must be updated, the actual size could be scale dependent, there's possible interactions with fullscreen modes, etc.

2. The event should be triggered immediatly after the resize and thus handled accordingly, as you can see running the test case (there was a reason why I added `DrawText(TextFormat("window size %i x %i", GetScreenWidth(), GetScreenHeight()), 20, 20, 20, BLACK);` there in the first place).

3. `PLATFORM_WEB` is not like other platforms as it's mostly web browser dependent. The web browser will ultimately have final say when it will execute anything request of it, so timing is never trully assured regardless and client-side must take that into consideration when developing for web.

4. `CORE.Window.screen` values should only be updated after the window (canvas on web) is resized because there's always a chance where that could fail (incidentally, the callback ensures that success condition).

raysan5:
@asdqwe agree.

PR #4456, opened 2024-11-01
[examples] drawing rotation shape in shapes_custom_rotation example

cmanlh:
shapes_custom_rotaion


raysan5:
@cmanlh Thanks for the addition! Please, could you use a 800x450 resolution for example to align with other raylib examples?

Also, as far as possible, could you implement the required systems required in [`raylib_examples_template`](https://github.com/raysan5/raylib/blob/master/examples/examples_template.c#L42)?

PR #4460, opened 2024-11-01
Update rmodels.c - 'fix' for GenMeshSphere artifact

MikiZX1:
When creating a new sphere mesh with high number of slices/rings the top and bottom parts of the generated sphere are removed. This happens because the triangles in those parts, due to high resolution, end up being very small and are removed as part of the 'par_shapes' library's optimization. Adding par_shapes_set_epsilon_degenerate_sphere(0.0); before generating the sphere mesh sets the threshold for removal of small triangles is removed and the sphere is returned to raylib correctly.


raysan5:
@MikiZX1 thanks for the review! 👍😄

Issue #4454, opened 2024-10-31
[raymath] `MatrixDecompose()` sometimes gives incorrect results when combining scale and rotation


waveydave:
### Issue description

MatrixDecompose seems to return incorrect scale and rotation information, for certain ranges of values. (I'm currently only concerned with z-rotations, so that is all I've been testing.)

### Environment

PC, Windows 11

### Code Example

Tested against latest master, b8c0842.

```
#include <raylib.h>
#include <raymath.h>

// parameters: z-rotation angle and scale
void test_decompose_rotz(float in_angle, float sx, float sy, float sz) {
    // Create the tranformation matrix
    Matrix m = MatrixScale(sx, sy, sz);

    // EDIT: my original implementation was incorrect (see discussion below)
    // m = MatrixMultiply(m, MatrixRotateZ(in_angle));
    // FIX:
    m = MatrixMultiply(MatrixRotateZ(in_angle), m);

    // Extract the scale and rotation info
    Vector3 translation, scale;
    Quaternion rotation;
    MatrixDecompose(m, &translation, &rotation, &scale);
    Vector3 axis;
    float out_angle;
    QuaternionToAxisAngle(rotation, &axis, &out_angle);
    out_angle = copysignf(out_angle, axis.z);

    // Check it matches the inputs
    if (FloatEquals(in_angle, out_angle) && FloatEquals(sx, scale.x) &&
        FloatEquals(sy, scale.y) && FloatEquals(sz, scale.z)) {
        TraceLog(LOG_INFO, "Pass: angle %.2f, scale(%.2f, %.2f, %.2f)", in_angle, sx, sy, sz);
    } else {
        TraceLog(LOG_ERROR, "Fail: angle %.2f, scale(%.2f, %.2f, %.2f) -> angle %.2f, scale(%.2f, %.2f, %.2f)",
                 in_angle, sx, sy, sz, out_angle, scale.x, scale.y, scale.z);
    }
}

int main ()
{
    test_decompose_rotz(0.5, 1, 1, 1); // pass
    test_decompose_rotz(2.5, 1, 1, 1); // pass
    test_decompose_rotz(0.5, 3, 3, 5); // pass
    test_decompose_rotz(2.5, 3, 3, 5); // fail -> angle 0.66, scale(3.00, 3.00, 5.00)
    test_decompose_rotz(0.5, 3, 4, 5); // pass (NB my original test showed this as failing, see FIX above)
[...]

PR #4461, opened 2024-11-03
[raymath] Fix for issue 4454, `MatrixDecompose()` gave incorrect output for certain scale and rotations

waveydave:
For transform matrices with certain ranges of rotation angles and scaling factors, `MatrixDecompose()` would give incorrect answers.

I narrowed down the issue to where the scaling factors are removed from the cloned transform, before the rotation is extracted: the scale factors needed to be taken into account for extra matrix members.

The models_loading_gltf example still works correctly after this fix.

Further tests and discussion here: https://github.com/raysan5/raylib/issues/4454


raysan5:
@waveydave Thanks for the review!

Issue #4464, opened 2024-11-04
[rcore_web] Canvas is incorrectly scaled when window size is applied


deathbeam:
The deployed example works fine: https://www.raylib.com/examples/core/core_2d_camera_mouse_zoom.html

Latest master (https://github.com/raysan5/raylib/commit/281ee51aff02c9e454891b75d6fdfead8ffccb0f) behaviour is input gets translated incorrectly and on initial load the canvas is not scaled and positioned incorrectly in bottom-left corner as well. This is because of width: 100% in template that tries to scale the canvas, but this most likely fails to scale the underlying window:

![image](https://github.com/user-attachments/assets/704c4e9a-f26e-4c7c-8e22-334f969b4267)

See https://github.com/raysan5/raylib/pull/4455 for original discussion.

Reproduction steps are simply using latest emsdk master, then navigating to raylib `src/` directory, running:

```sh
make PLATFORM=PLATFORM_WEB -B
```

then navigating to `examples/` directory and running:

```sh
make PLATFORM=PLATFORM_WEB -B
```

and then running:

```sh
emrun --port 8080 core/core_2d_camera_mouse_zoom.html
```

raysan5:
@deathbeam thanks for reporting this new issue, I'm reviewing it but I can't find the cause of the issue, even tried reverting to some raylib 5.0 parts but issue also happens! Could it maybe be related to emscripten or some browser change in the way canvas is managed??? I'm a bit lost at the moment... 😢

Mute124:
If you reverted it, then it is probably an issue with emscripten. It could also just be their browser itself.

raysan5:
After some investigation, tahnks to raylib Discord community, we found that issue seems related to some change in emscripten. Building with `emsdk 3.1.43` it works as expected.

Trying to find at what point the issue was introduced. One potential change reported is on [emsdk 3.1.51](https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3151---121323).

After further testing it was verified that it breaks from `3.1.50` -> `3.1.51`

asdqwe:
Maybe related: [[wasm64] Fix offscreen canvas and offscreen framebuffer #20678](https://github.com/emscripten-core/emscripten/pull/20678) happened on Nov 13, 2023 (between 3.1.49 and 3.1.50).

raysan5:
[...]

PR #4466, opened 2024-11-05
[shell] [web] Temporary workaround for `shell.html` `<canvas>` CSS width

asdqwe:
Ref: #4464

Temporary workaround for `canvas.emscripten { width: 100%; }` CSS not working with current emscripten (`3.1.70`).
**Revert this change when the issue is fixed upstream.**

1. Found out that the `<canvas>` width can be resized with CSS, but only after it's loaded.
2. The issue could be related to `$GLFW_Window: function` that now takes the framebuffer's width/height as separated parameters ([ref1](https://github.com/emscripten-core/emscripten/blob/dab87027494b547358253a0336259ac629b80d89/src/library_glfw.js#L40), [ref2](https://github.com/emscripten-core/emscripten/blob/dab87027494b547358253a0336259ac629b80d89/src/library_glfw.js#L1114-L1115)).
3. Hooked this workaround to `Module.setStatus` because it was the only Module option I found that runs (consistently) after the canvas finishes loading.

### DIFF
--- a/src/shell.html
+++ b/src/shell.html
@@ -79,7 +79,7 @@ jwE50AGjLCVuS8Yt4H7OgZLKK5EKOsLviEWJSL/+0uMi7gLUSBseYwqEbXvSHCec1CJvZPyHCmYQffaB
       canvas.emscripten {
         border: 0px none;
         background: black;
-        width: 100%;
+        /* width: 100%; */
       }
 
       .spinner {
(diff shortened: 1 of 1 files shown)


asdqwe:
Unfortunately (upon more testing) this workaround only handles the canvas size. The mouse input is still wrong (unscaled). Closing the PR because the workaround is insufficient.

PR #4468, opened 2024-11-05
Improve cross-compilation with zig builds

deathbeam:
- Add xcode_frameworks when cross compiling for mac (ref: https://github.com/hexops/mach/blob/main/build.zig)
- Add emsdk when cross compiling for wasm (ref: https://github.com/floooh/sokol-zig/blob/master/build.zig)

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 /// Minimum supported version of Zig
-const min_ver = "0.12.0";
+const min_ver = "0.13.0";
 
 comptime {
     const order = std.SemanticVersion.order;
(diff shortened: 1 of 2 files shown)


deathbeam:
For testing:

```
zig build -Doptimize=ReleaseFast -Dtarget=wasm32-emscripten
zig build -Doptimize=ReleaseFast -Dtarget=x86_64-macos
```

should just work :tm:

raysan5:
@deathbeam I don't like specially adding a separate .zig.zon file, more files usually mean more maintenance work and more breaking possibilities... but I'm merging it.

deathbeam:
> @deathbeam I don't like specially adding a separate .zig.zon file, more files usually mean more maintenance work and more breaking possibilities... but I'm merging it.

Yea if there was an option to put it to the build.zig i would do it but to use the zig dependency manager the zig.zon file is sadly necessary (at least afaik). but benefit is that when you use raylib as dependency in zig for other projects they will also get the 2 dependencies resolved at same time and cross-compilation should work for them out of the box as well (when set up properly).

PR #4469, opened 2024-11-07
Add TextInvert and TextClear functions

Cyrodwd:
Added the ‘TextInvert’ function to revert a text string, and the ‘TextClear’ function to clean up its contents. This is to give more functions for string manipulation in raylib. Any suggestions for changes, additions, etc., I will take them into account.


raysan5:
@Cyrodwd I'm afraid those functions are very use-case specific, I prefer not to add them into main library, it would imply more maintenance work.

PR #4470, opened 2024-11-07
UpdateModelAnimation speedup

JettMonstersGoBoom:
If we borrow from the GPU skinned animation code, we can just multiply the vertex by the matrix * weight and shave a chunk of CPU time.
over 2X speedup in a complex test.


raysan5:
@JettMonstersGoBoom Very nice! Thanks for the review!

PR #4473, opened 2024-11-09
[raymath] Clang Vector Extension support

jopadan:
- Replace src/raymath.h Vector type definitions with ext_vector_type ones for clang compiler to activate use
`cmake -DCMAKE_C_COMPILER=clang`
- silence example type cast warning

**TODO**
- check for Vector3 storage size change from 12 to 16 bytes complications

### DIFF
--- a/examples/models/models_point_rendering.c
+++ b/examples/models/models_point_rendering.c
@@ -159,9 +159,9 @@ Mesh GenMeshPoints(int numPoints)
     // https://en.wikipedia.org/wiki/Spherical_coordinate_system
     for (int i = 0; i < numPoints; i++)
     {
-        float theta = PI*rand()/RAND_MAX;
-        float phi = 2.0f*PI*rand()/RAND_MAX;
-        float r = 10.0f*rand()/RAND_MAX;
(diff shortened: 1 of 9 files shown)


raysan5:
@jopadan Did you benchmark the benefits of this change with some sample code?

It seems interesting but I'm concerned of the potential implications of this change. `raymath` is a base library for raylib, used by multiple modules.

raysan5:
I'm afraid the changes volume is too high and too risky for merging, just to consider a specific compiler.

orcmid:
My impression is that compiler command-line optimization options are a better way to rely on particular hardware capabilities and also deal with their conditional presence at run-time.

The use of data structures that are recognized as optimizable are a limitation that might distort the source code and operation in the absence of actual special hardware features though.

We have enough to deal with concerning platform differences and surrendering to compiler differences sort of flies in the face of the idea that a feature of C Language is portability across platforms.

In any case, I favor the conclusion of @raysan5

jopadan:
For benchmark results see #4486 and continue this PR topic there

PR #4475, opened 2024-11-10
[build][Zig] Remove `addRaylib` and fix `raygui` builds when using raylib as dep

deathbeam:
- addRaylib just duplicates what adding raylib as dependency already does
  so it do not needs to exist
- move raygui build to standard build process when flag is enabled. this
  works correctly when using raylib as dependency and having raygui as
  dependency as well. problem with previous approach was that raygui was in
  options but it was doing nothing because you had to also use addRaygui for
  it to be effective

For how the dependency should be included normally in other projects using zig builds, its as simple as this (on top of including raylib in build.zig.zon)

```zig

// This is the whole thing
const raylib_dep = b.dependency("raylib", .{ .target = target, .optimize = optimize, .shared = true });
const raylib = raylib_dep.artifact("raylib");
b.installArtifact(raylib)
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -11,36 +11,6 @@ comptime {
         @compileError("Raylib requires zig version " ++ min_ver);
 }
 
-// NOTE(freakmangd): I don't like using a global here, but it prevents having to
-// get the flags a second time when adding raygui
-var raylib_flags_arr: std.ArrayListUnmanaged([]const u8) = .{};
-
-// This has been tested with zig version 0.13.0
-pub fn addRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.builtin.OptimizeMode, options: Options) !*std.Build.Step.Compile {
-    const raylib_dep = b.dependencyFromBuildZig(@This(), .{
-        .target = target,
-        .optimize = optimize,
-        .raudio = options.raudio,
-        .rmodels = options.rmodels,
-        .rshapes = options.rshapes,
-        .rtext = options.rtext,
-        .rtextures = options.rtextures,
-        .platform = options.platform,
-        .shared = options.shared,
-        .linux_display_backend = options.linux_display_backend,
-        .opengl_version = options.opengl_version,
-        .config = options.config,
-    });
-    const raylib = raylib_dep.artifact("raylib");
-
-    if (options.raygui) {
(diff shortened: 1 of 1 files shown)


raysan5:
@deathbeam thanks for the review! I like PRs that remove code lines!

@sagehane maybe you can also review those changes?

sagehane:
I never used `addRaylib` because it's not idiomatic as far as the Zig package manager is supposed to be used, so I am in favour of removing it. I've not used raygui so I really can't comment on it, and the only other thing to note is the removal of the global flags array and moving around a bit of code. It looks good to me.

I will note that it is a breaking change in that `addRaylib` will no longer work, but the functionality exposed is the same and people will just be encouraged to do things the more idiomatic way.

deathbeam:
Re: breaking change: this is list of projects that are using addRaylib in build.zig:

https://github.com/search?q=addRaylib+path%3Abuild.zig&type=code&p=1

A lot of it is coming from having raylib build.zig vendored so its not even being used. And outisde of 1 or 2 repos everything else is already not working on latest master (latest master addRaylib requires try before so everything that doesnt use it wont work).

Also probably the biggest consumer of build.zig (i assume) is using the dependency properly (+ some boilerplate for raygui which should be unnecessary after this change too):

https://github.com/Not-Nik/raylib-zig/blob/devel/build.zig#L103

raysan5:
@sagehane @deathbeam thanks for the further review!

fourtytwelve:
@deathbeam this is just a bad idea tbh

AnthonyAltieri:
> Re: breaking change: this is list of projects that are using addRaylib in build.zig:
> 
> https://github.com/search?q=addRaylib+path%3Abuild.zig&type=code&p=1
> 
> A lot of it is coming from having raylib build.zig vendored so its not even being used. And outisde of <5 repos everything else is already not working on latest master (latest master addRaylib requires try before so everything that doesnt use it wont work, which is another side effect of having custom dependency function that only causes issues).
> 
> Also probably the biggest consumer of build.zig (i assume) is using the dependency properly (+ some boilerplate for raygui which should be unnecessary after this change too):
> 
> https://github.com/Not-Nik/raylib-zig/blob/devel/build.zig#L103
> 
> So I think impact of the change is minimal

Can you link some documentation about the recommended approach?

PR #4477, opened 2024-11-10
[build]: Replace fatal ``errors`` with ``conditional warnings`` for Clipboard Image Support

evertonse:
Fixes [#4459](https://github.com/raysan5/raylib/pull/4459#issuecomment-2466821215)

Conditional compilation warnings instead of warnings to inform developers about the necessary dependencies when enabling clipboard image support (e.g. SUPPORT_MODULE_RTEXTURES, specific image formats , etc ...).  

This allows for more flexible configuration while still alerting users, although subtely.

NOTE: [6.10.7 C standard](https://open-std.org/jtc1/sc22/wg14/www/docs/n3096.pdf) says that any unrecognized ``#pragma`` should be ignores, so it should be no problem is case of less common C compilers.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -512,10 +512,27 @@ const char *TextFormat(const char *text, ...);              // Formatting of tex
     #define PLATFORM_DESKTOP_GLFW
 #endif
 
+// We're using `#pragma message` because `#warning` is not adopted by MSVC.
 #if defined(SUPPORT_CLIPBOARD_IMAGE)
-    #if !defined(SUPPORT_FILEFORMAT_BMP) || !defined(STBI_REQUIRED) || !defined(SUPPORT_MODULE_RTEXTURES)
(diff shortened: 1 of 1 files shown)


raysan5:
@evertonse Merging this review but maybe it will be redesigned, I usually try to avoid/minimize `#pragma` errors/warnings in raylib... but I need to think about it...

evertonse:
> @evertonse Merging this review but maybe it will be redesigned, I usually try to avoid/minimize `#pragma` errors/warnings in raylib... but I need to think about it...

That's completely fine.

A better solution might be to pull all the dependencies inside ``rcore.c`` itsself, [as I suggested here](https://github.com/raysan5/raylib/pull/4459#issuecomment-2468549210), because it seems that building with ``cmake -DCUSTOMIZE_BUILD`` it doesn't pull ``config.h``, I might be wrong but it seems that way.

raysan5:
@evertonse `config.h` didn't force any flags, just define the user desired ones (but lately some conditionals were added).

The approach usually used on raylib was just checking required flag availability at a function level and fail gracefully with a `TraceLog(LOG_WARNING, ...)` message. It's up to the user to enable required/desired flags.

PR #4478, opened 2024-11-10
[rModels] Fix bad triangles in DrawCylinder

JeffM2501:
The lower Y cap of the cylinder in DrawCylinder has a small mistake and does not draw properly.
this PR fixes it up to render correctly.
image


raysan5:

PR #4480, opened 2024-11-10
[rcore][RGFW] Add new backend option: `PLATFORM_WEB_RGFW` and update RGFW

ColleagueRiley:
Adds `PLATFORM_WEB_RGFW` and updates RGFW

RGFW Updates:
- fix bugs and refine gamepad support, and rename joystick to gamepad (X11) (WinAPI) (HTML5)
- fix window resize bug (X11)
- fix use of ctypes (Visual Studio)
- Fix warnings


ColleagueRiley:
Issues thus far: 

- The examples compile fine until:

```c
core/core_basic_window_web.c:18:10: fatal error: 'raylib.h' file not found
   18 | #include "raylib.h"
```

I'm not sure why this is happening for this example but not the others.

- The mouse wheel is not working correctly 

- more testing has to be done

raysan5:
@ColleagueRiley Thank you very much for working on this new platform backend! Exiting to have a GLFW alternative for `PLATFORM_WEB`! Keep up the great work!

ColleagueRiley:
The only remaining issue with the web platform is the fullscreen toggle.

Makefile
```diff
@@ -0,0 +1,8 @@
+EXAMPLE = core/core_input_gamepad
```
raysan5, line 1: It seems this file shouldn't be in the PR

ColleagueRiley: Yea, not sure how it got added. I'll fix that.

raysan5:
@ColleagueRiley I've been reviewing this PR. Some concerns:

 - I see now we have `PLATFORM_WEB` (defaults to GLFW) and `PLATFORM_WEB_RGFW`. I'm thinking if the currrent approach with the filters in Makefile can be somewhat simplified in some way... maybe not... just thinking...
 - There are several formating detailes in the PR, not following raylib coding conventions, including some TABs.

Thinking about the best way to add this new Web platform backend option to minimize friction... no need to change anything for now (beside the formating details).

ColleagueRiley:
@raysan5 In my opinion the current approach is already pretty simply, an alternative could be to define a variable like `USES_WEB` when `PLATFORM_WEB` or `PLATFORM_WEB_RGFW` is defined and use that instead of checking for `PLATFORM_WEB` or `PLATFORM_WEB_RGFW` each time. Although I don't think that way is necessarily better or worse, it's just up to how you want it to be formatted.

ColleagueRiley:
@raysan5 I fixed the formatting issues I found, let me know if I'm missing anything.

raysan5:
@ColleagueRiley Thank you very much for all the work put into this improvement! Really happy that now we have an alternative to GLFW for Web platform building! I'm sure it will be really helpful in the future! Thanks!

PR #4481, opened 2024-11-11
[rcore] Add basic support for scancodes

JeffM2501:
Currently raylib does not expose the scancode data for keyboard input received from the platform. This makes it VERY hard to build keyboard bindings that work with different keyboard mappings and localizations. They KeyboardKey enum is just the physical keys and does not take the keyboard mapping into account. For most game input this is fine, as you want the physical key, but there are use cases where you need local mapping of non-text keys.

GetCharPressed handles the mapping, but only for printable characters.

Any user that wants the 'logical' key that is provided by the keyboard mapping as no way to get this data.
The data was passed to the keyboard input functions of most platforms, but the raylib key system has no place to store it.
In SDL this scancode is the 'virtual key'

This PR adds GetKeyPressedPro that takes an optional scancode to fill out, allowing people to see the localized key for each event.
Platforms that don't provide a scancode set the code to -1.

I don't know if this should go into 5.5 or if we should discuss the API for the future. This is just the smallest possible change to raylib to expose the data that it's currently ignoring. I am just posting this so it doesn't get lost.


raysan5:
@JeffM2501 I see the issue but I want to review this change more carefully, waiting after 5.5 release for further discussion.

JeffM2501:
> @JeffM2501 I see the issue but I want to review this change more carefully, waiting after 5.5 release for further discussion.

Absolutely understandable

src/rcore.c
```diff
@@ -268,6 +268,7 @@ __declspec(dllimport) unsigned int __stdcall timeEndPeriod(unsigned int uPeriod)
 //----------------------------------------------------------------------------------
 typedef struct { int x; int y; } Point;
 typedef struct { unsigned int width; unsigned int height; } Size;
+typedef struct { int keycode; int scancode; } KeyCodes;
```
raysan5, line 271: I'd call this struct `KeyInfo` to avoid confusion with `keycode`

src/raylib.h
```diff
@@ -1177,6 +1177,7 @@ RLAPI bool IsKeyReleased(int key);                            // Check if a key
 RLAPI bool IsKeyUp(int key);                                  // Check if a key is NOT being pressed
 RLAPI int GetKeyPressed(void);                                // Get key pressed (keycode), call it multiple times for keys queued, returns 0 when the queue is empty
 RLAPI int GetCharPressed(void);                               // Get char pressed (unicode), call it multiple times for chars queued, returns 0 when the queue is empty
+RLAPI int GetKeyPressedPro(int* scanCode);                       // Get key pressed (keycode), and optional scancode, call it multiple times for keys queued, returns 0 when the queue is empty
```
raysan5, line 1180: As per raylib design, I try to minimize return-pointers when possible. I prefer a ` int GetKeyPressedRaw(void)` function to note the raw nature (scancode) of returned key.

src/platforms/rcore_android.c
```diff
@@ -1225,7 +1225,8 @@ static int32_t AndroidInputCallback(struct android_app *app, AInputEvent *event)
             {
                 CORE.Input.Keyboard.currentKeyState[key] = 1;   // Key down
 
-                CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
+                CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount].keycode = key;
+                CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount].scancode = -1;
```
raysan5, line 1229: It seems scancode can be read with NDK [AKeyEvent_getScanCode()](https://developer.android.com/ndk/reference/group/input#group___input_1ga4a0a846b7a195aeb290dfcd2250137d9).

src/rcore.c
```diff
@@ -3163,28 +3165,37 @@ bool IsKeyUp(int key)
...
         CORE.Input.Keyboard.keyPressedQueueCount--;
     }
 
     return value;
 }
 
+// Get the last key pressed
+int GetKeyPressed(void)
+{
+    return GetKeyPressedPro(NULL);
```
raysan5, line 3196: Considering `GetKeyPressed()` is very short, I think code lines can be "duplicated" for `GetKeyPressedRaw()`.

src/platforms/rcore_desktop_rgfw.c
```diff
@@ -1001,7 +1001,8 @@ void PollInputEvents(void)
                     // If key was up, add it to the key pressed queue
                     if ((CORE.Input.Keyboard.currentKeyState[key] == 0) && (CORE.Input.Keyboard.keyPressedQueueCount < MAX_KEY_PRESSED_QUEUE))
                     {
-                        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount] = key;
+                        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount].keycode = key;
+                        CORE.Input.Keyboard.keyPressedQueue[CORE.Input.Keyboard.keyPressedQueueCount].scancode = event->keyCode;
```
raysan5, line 1005: I'm curious, why key `scancode` is actually `event->keyCode`? @ColleagueRiley

src/raylib.h
JeffM2501: The problem with making a function that just returns the scan code is that this pulls from the events. Calling that function will eat the event and prevent you from beeing able to get the Keyboard Key too.
There may be cases where the user wants both (such as for a "press a key" control config dialog), they need the KeyboardKey to use later for polling, but the scancode to be able to show the mapped key name.

The other option is to return the KeyInfo structure by value.

JeffM2501:
Per our discussion on discord. 
GLFW calls the KeyboardKey value "key" and the mapped value ScanCode
SDL calls the KeyboardKey value scancode, and the mapped value the  "SDL virtual keysym" https://wiki.libsdl.org/SDL2/SDL_Keysym
This is very confusing.
I think we should come up with our own names for these things and let the platform map as needed.

orcmid (commented):
I think the pair struct is valuable.  `mappedcode` seems awkward (and I wonder about the absence of camelCase).

For me, I think of it as scanCode versus keyTop.  And of course, that would normally depend on a keyboard binding, with a QWERTY ANSI keyboard the default.   So that's murky too, since shift-state is also reflected.  On the other hand, it might be more helpful for people to understand at least the default situation.  Not proposing anything, just musing about the awkwardness of this situation.

ColleagueRiley:
@raysan5 After looking into GLFW's code with Jeff, I found that GLFW only supports the use of scancodes, but not keycodes. 

The `scancode` data is the source API's scancode, where as the `key` data is GLFW's API scancode.

JeffM2501:
Yes, after good discussion and research, it has been shown that GLFW doesn't support a mapped value, so this entire PR is pointless while GLTF is the default backend. I am going to close it and open a discussion where we can talk about what we want raylib to expose and come up with use cases.

orcmid:
> I think the pair struct is valuable. `mappedcode` seems awkward (and I wonder about the absence of camelCase).
> 
> For me, I think of it as scanCode versus keyTop. And of course, that would normally depend on a keyboard binding, with a QWERTY ANSI keyboard the default. So that's murky too, since shift-state is also reflected. On the other hand, it might be more helpful for people to understand at least the default situation. Not proposing anything, just musing about the awkwardness of this situation.

@JeffM2501 
I await the Discussion to be opened.  Meanwhile, I keep obsessing about this complicated situation.  I think there are 2+ cases and tensions with an application's usability (UX) and internationalization (i18n).  In particular, dialogs that specify keypresses for actions (e.g., the famous WASD on US-ANSI keyboards and similar ones based on Roman alphabets).  Accessibility (a11y) might also factor in, along with IME.

I am tempted to suggest that "mapped code" should be "keyedChar" (depending on type, since might not be u8, maybe "keyedGlyph" but might be more accurate but not so understandable) that provides coordination with what an user sees themselves keying and what the application recognizes as keyed if reflected in a message and especially in an action.

Somehow, the default case should just work in alignment with the original impetus for raylib (namely BGI) and the idea of simplicity.  

The question then becomes how to address some of these more-expanded use cases and computer configurations while being as simple as possible (but no simpler) in the main and in line with raylib as a learning instrument for graphical applications.

JeffM2501:
I have created a discussion topic here
https://github.com/raysan5/raylib/discussions/4595

PR #4482, opened 2024-11-11
[Examples] Fix colors in lighting shaders

JeffM2501:
There is a small regression with the lighting example/shader from previous versions with the new 5.5 example,
5.0 used meshes since the batch did not support normals. This meant that the tint was passed to the shader as the 'colDiffuse' uniform.
uses shapes, since the batch system does now support normals. Shapes do't set colDiffuse from the tint, they apply the tint as vertex colors.

The example lighting fragment shader does not use the fragColor, only colDiffuse, so this means that the tint is now ignored where it was not before.

This PR fixes the example lighting shaders to use both frag and diffuse colors so they work with shapes and meshes.


raysan5:
@JeffM2501 Nice! Thanks for the review!

PR #4485, opened 2024-11-13
[BUILD][ZIG]: remove raygui from build options and re add addRaygui as a function

kimierik:
- Removed raygui option from build.zig since it is not working as intended
- Re add addRaygui function

Function addRaygui is not the idiomatic way of including dependencies in zig, but currently using the raygui flag would give the following error. This would happen because the build system is looking for the raygui dependancy in from raylibs own [build.zig.zon](https://github.com/raysan5/raylib/blob/master/build.zig.zon) file.
 ```
thread 27788 panic: no dependency named 'raygui' in '/home/kimi/.cache/zig/p/1220eb09ded36376449dc82f259286d0ad17440081117e36bb41604504a0c0ac19c3/build.zig.zon'. All packages used in build.zig must be declared in this file
```
Having the addRaygui function saves the user from adding non intuitive code into their own build.zig file.
  
The following is required for adding raygui to a project as a dependancy (Assuming raylib and raygui are defined dependencies in build.zig.zon)
```zig
const std = @import("std");
const rl = @import("raylib");

pub fn build(b: *std.Build) void {

        .name = "example",
        .root_source_file = b.path("src/main.zig"),
        .target = target,
        .optimize = optimize,
    });

    const raylib_dep = b.dependency("raylib", .{});
    const raylib = raylib_dep.artifact("raylib");
    rl.addRaygui(b, raylib, b.dependency("raygui", .{}));

    exe.linkLibrary(raylib);
    b.installArtifact(exe);
}
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -298,19 +298,18 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
         .flags = raylib_flags_arr.items,
     });
 
-    if (options.raygui) {
-        const raygui_dep = b.dependency(options.raygui_dependency_name, .{});
-
-        var gen_step = b.addWriteFiles();
-        raylib.step.dependOn(&gen_step.step);
-
-        const raygui_c_path = gen_step.add("raygui.c", "#define RAYGUI_IMPLEMENTATION\n#include \"raygui.h\"\n");
-        raylib.addCSourceFile(.{ .file = raygui_c_path, .flags = raylib_flags_arr.items });
-        raylib.addIncludePath(raygui_dep.path("src"));
-        raylib.installHeader(raygui_dep.path("src/raygui.h"), "raygui.h");
-    }
-
     return raylib;
+}
+
+pub fn addRaygui(b: *std.Build, raylib: *std.Build.Step.Compile, raygui_dep: *std.Build.Dependency) void {
+    var gen_step = b.addWriteFiles();
+    raylib.step.dependOn(&gen_step.step);
+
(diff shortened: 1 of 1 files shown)


diwalkerdev:
Sharing in case this is useful to anyone in the future. Probably user error, but I couldn't find anything else that I was doing wrong.

When using `addRaygui`, `raygui.c` fails to compile, as it cannot find `raylib.h`.

This seems to be the case for master fde3779, but it was also happening for a slightly older version of raylib before I updated to the latest master (I'm not sure of the hash number, but I remember seeing the min zig version supported was 0.12 in `build.zig`).

For my testing, I used the example `build.zig` provided by @kimierik in the first comment and a minimal raylib window example.

The solution was to add:
```zig
rl.addRaygui(b, raylib, b.dependency("raygui", .{}));
raylib.addIncludePath(raylib.getEmittedIncludeTree()); // <== Line to add
``` 
I doubt this is the most idiomatic way of solving this problem, but it seems to work. It seems that raylib is not aware of its own include path during compilation, and since raygui depends on raylib, the compilation fails.

Unrelated to raylib, but I also found that I needed to force zig to download the latest master, by breaking the hash number in the zon file. This causes zig to redownload of the zip file. I then needed to fix the hash once zig complained the hash number was wrong. This is possibly the expected behaviour, but was not what I expected, as I thought it would track the latest master.

kimierik:
> I doubt this is the most idiomatic way of solving this problem, but it seems to work. It seems that raylib is not aware of its own include path during compilation, and since `raygui.h` depends on `raylib.h`, the compilation fails.

This was an oversight on my part and as far as i know this only happens in windows builds. I have tried to fix this in #4489.

PR #4486, opened 2024-11-13
Math supporting Clang Vector Extension

jopadan:
**Changes**
- Restructure Raymath and simplify most of the code to use already existing inline functions without changing the exposed interface
- Add minor missing Vector/Quaternion functions
- silence example type cast warning
- add Vector4.r0-Vector4.r3 rows to matrix type
- use vector type definitions with ext_vector_type if CMAKE_C_COMPILER=clang
- minor fixes for examples see **Possible conflicts**
- move C++ operators to raymath.hpp

**Test/Examples**
- Every example included runs without a flaw adapted to vector extensions

**Possible conflicts**
- Vector3 storage size change from 12 to 16 bytes
- initializer list .= element not working for ext_vector_type
- address of vector element

**Notes**
I am sorry to open another PR after the earlier closed #4473 that should solve all of the concerns with the one before.
How do I benchmark the changes with examples? Do I have to disable GLFW vsync in texture_bunnymark or is there another existing benchmark program?
Please don't just reject and close it without reading the code.

### DIFF
--- a/examples/models/models_point_rendering.c
+++ b/examples/models/models_point_rendering.c
@@ -102,15 +102,15 @@ int main()
                     for (int i = 0; i < numPoints; i++)
                     {
                         Vector3 pos = {
-                            .x = mesh.vertices[i*3 + 0],
-                            .y = mesh.vertices[i*3 + 1],
-                            .z = mesh.vertices[i*3 + 2],
+                            mesh.vertices[i*3 + 0],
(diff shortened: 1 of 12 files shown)


raysan5:
@jopadan Thanks for the time put on this PR but it includes many changes that are not desired for raylib, as already commented on previous PR. I'm not adding compiler specific code for `clang` and I prefer to keep the code simpler, as it is now.

PR #4489, opened 2024-11-14
[build][Zig] Fix `raygui` inclusion in windows cross compilation

kimierik:
## Problem:  
Cross compiling to windows when adding Raygui would error due to how zig handles the C \#include macro.
## Solution:  
Added explicit include path for raylib.h for the raylib dependency when adding raygui.
  
## Extra Info:  
- Build and executable tested only on my local machine.
- Only tested with argument  -Dtarget=x86_64-windows

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -302,12 +302,14 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
 }
 
 pub fn addRaygui(b: *std.Build, raylib: *std.Build.Step.Compile, raygui_dep: *std.Build.Dependency) void {
+    const raylib_dep = b.dependencyFromBuildZig(@This(), .{});
     var gen_step = b.addWriteFiles();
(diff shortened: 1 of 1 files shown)


deathbeam:
wouldnt it be better to just get rid of addRaygui and provide the raygui dependency reference as parameter in options (e.g the approach i started, but this part wasnt finished)? That should be possible I think. Instead of trying to make addRaygui work.

kimierik:
> Benefit of having the raygui dep in build.zig.zon directly is also that user will always have actual functional build instead of potentionally mismatched raylib and raygui versions

I dont think having raygui in the build.zig.zon is that good. The hash would have to be updated whenever raygui updates. And it could still lead to non functional builds if raylib updates and the raygui version has not.
  
I agree that addRaygui is not that good of a solution to the problem and a better alternative should be considered.

raysan5:
@kimierik thanks for the review!

PR #4492, opened 2024-11-14
[Build MSVC] Warnings and missing projects

JeffM2501:
Fixes a few casting warnings... [I wish people would check types when they submit examples. :( ]
shapes_rectangle_advanced was in the wrong solution dir in the MSVC project
Added the virtual controls example back to the project instead of it being a dangling project reference.

This is all just small MSVC fixes to get ready for release.


raysan5:
@JeffM2501 thank you very much! just in time! updating the release package!

PR #4494, opened 2024-11-15
Fixing an OBJ loader bug that fragmented the loaded meshes

zet23t:
The nextShapeEnd integer is a pointer in the OBJ data structures. The faceVertIndex is a vertex index counter for the mesh we are about to create. Both integers are not compatible, which causes the code to finish the meshes too early, thus writing the OBJ data incompletely into the target meshes.

It wasn't noticed because for the last mesh, it process all remaining data, causing the last mesh to contain all remaining triangles.

This would have been noticed if the OBJ meshes used different textures for each mesh.

mesh-debugging


raysan5:
@zet23t Great! Thank you very much for reviewing this issue anf for the fast fix! 👍😄

PR #4496, opened 2024-11-15
[rmodels] Improve OBJ vertex data precision and lower memory usage by ExportMesh()

mikeemm:
Using ExportMesh() to save a mesh as an OBJ currently creates broken models due to very low floating-point precision, both for vertex position, normals and UVs.

Here's a model I made, imported into a raylib program:

And here's the same model, first imported into the program, then exported using ExportMesh(), then imported back:

I guess vertex positions could be scaled x1000 before exporting to account for the precision loss, but neither vertex normals nor UVs can be treated the same way. I don't see any reason why these values cannot be increased; I replaced the relevant numbers with the precision values that Blender's OBJ exporter uses, which seems to give me a visually identical model to the original upon next import.

The exporter also greatly overestimates the amount of memory necessary to be allocated to store the data in, which could potentially lead to issues when attempting to export very big meshes. Wanting to lower the memory allocation to the minimum necessary, while trying to add the least amount of extra code as well as avoiding polluting the codebase with extra functions, I made a macro function declaration inside the exporter and replaced much of the code there with that new function. The code compiles and works fine in both Windows and HTML5; however I don't know if macro functions are acceptable in this project, and perhaps a different approach might be preferable to the one I ended up taking. If so, I totally understand; I'll change it as required.


raysan5:
@mikeemm I prefer to keep the same function structure as other similar functions and avoid that `TEXT_APPEND` macro. I the problem is the scale of the buffer, incrementing its size to fit more decimals would be enough, no need to redesign all the function.

mikeemm:
@raysan5 I understand. I have reverted the macro changes and modified the initial allocation instead, please take a look whenever you'd like.

PR #4511, opened 2024-11-18
BINDINGS: raylib-cpp supports raylib 5.5

RobLoach:
[raylib-cpp](https://github.com/RobLoach/raylib-cpp) has been updated to use the new raylib 5.5 functions. This is a great release! Thanks so much.


raysan5:
@RobLoach Super-fast update! Thank you Rob! 👍😄

PR #4512, opened 2024-11-19
Update BINDINGS.md: raylib-python-cffi is on v5.0

villares:
According to https://github.com/electronstudio/raylib-python-cffi


raysan5:
@villares Thanks for the review!

villares:
> @villares Thanks for the review!

My pleasure... it looks like they want to get to 5.5, let's see how it goes and I'll send another PR!

PR #4519, opened 2024-11-20
Update BINDINGS.md for raylib-py 5.5

overdev:


raysan5:
@overdev thansk for the update!

PR #4520, opened 2024-11-20
[raymath] Add cross product function

uwiwiow:
while working on a personal project using raylib, i noticed that raymath doesnt have a function for calculating the cross product of 2D vectors, so i implemented a simple function for this purpose in my project, and i think it could be a useful addition to the raymath module.


raysan5:
@uwiwiow thanks for the addition!

Issue #4521, opened 2024-11-20
[raudio] SeekMusicStream seeks to incorrect value


amigash:
### Issue description

Using SeekMusicStream with a value of 0.0 (or very close, such as 0.01) causes a jump to the end of the song (visible with GetMusicTimePlayed)

### Environment

Windows 11

### Code Example

```
(Load and start playing song)
...
TraceLog(LOG_INFO, "GetMusicTimePlayed: %f", GetMusicTimePlayed(song));
SeekMusicStream(song, 0.0);
TraceLog(LOG_INFO, "GetMusicTimePlayed: %f", GetMusicTimePlayed(song));
UpdateMusicStream(song);
TraceLog(LOG_INFO, "GetMusicTimePlayed: %f", GetMusicTimePlayed(song));                   
```                                             
Output:

```
INFO: GetMusicTimePlayed: 0.810000
INFO: GetMusicTimePlayed: 255.672592
INFO: GetMusicTimePlayed: 255.672592
```
In this example, the song length is 255.733337.

nk521:
Can reproduce. I used the "Music stream processing effects" example and changed pause key to do what your code does. Upon pausing, the music seeks to the end and then back to `0.0`. I tested it using Raylib 5.5 on windows 11.

raysan5:
Issue should be fixed.

PR #4523, opened 2024-11-20
Fix for #4521 (Incorrect GetMusicTimePlayed)

veins1:


raysan5:
@veins1 Nice! Thanks for the review!

Issue #4522, opened 2024-11-20
[rcore] Incorrect monitor resolution


xroberx:
### Issue description

The monitor of my laptop has a resolution of 2560x1600. I have configured scaling to 1.25x through Xfce, which sets virtual resolution to 3328x2080 as can be seen in the output of the xrandr command. However, when I call GetMonitorWidth()/GetMonitorHeight(),  I get the physical resolution of the monitor (2560x1600), not the virtual one. Also, GetWindowScaleDPI() returns 1. 

SDL3 returns the virtual resolution on X11, so I don't think it's a misconfiguration on my part. 

### Environment

Hardware: Ryzen 7 8845HS laptop with Radeon 780M integrated graphics, 2560x1600 pixels monitor
Software: Linux 6.11.9, XFCE 4.18, Xorg-server 21.1.4

### Issue Screenshot

![xrandr](https://github.com/user-attachments/assets/ab082fa2-2dd5-468d-9581-f44b15c3ef97)

### Code Example

`
InitWindow(WIDTH, HEIGHT, "Test");
int w = GetMonitorWidth(0);
int scale = GetWindowScaleDPI();
`

asdqwe:
Those are the correct values. `GetMonitorWidth()`, `GetMonitorHeight()` are returning the exact sizes provided by GLFW ([ref](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_glfw.c#L852-L854), [ref](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_glfw.c#L871-L873)) or SDL ([ref](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_sdl.c#L957-L958), [ref](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_sdl.c#L974-L975)), as they should.

If you want to factor in arbitrary user defined/applied scales you'll have to multiply it on client-side. GLFW has it ([ref](https://www.glfw.org/docs/3.3/group__monitor.html#gad3152e84465fa620b601265ebfcdb21b)), SDL3 has it ([ref](https://wiki.libsdl.org/SDL3/SDL_GetDisplayContentScale)), SDL2 doesn't ([ref](https://wiki.libsdl.org/SDL2/SDL_GetDisplayContentScale)). But nobody implemented a `GetMonitorScale()` yet for raylib, although it probably should be added at some point.

xroberx:
[...]

PR #4525, opened 2024-11-20
[rcore] Add `GetMonitorScale()`

asdqwe:
Ref: #4522

Necessary to allow the user to factor in arbitrary user defined/applied monitor scales ([GLFW ref](https://www.glfw.org/docs/3.3/group__monitor.html#gad3152e84465fa620b601265ebfcdb21b), [SDL3 ref](https://wiki.libsdl.org/SDL3/SDL_GetDisplayContentScale), [Web ref](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio)).
Note: this is different from `GetWindowScaleDPI` ([GLFW ref](https://www.glfw.org/docs/3.3/group__window.html#gaf5d31de9c19c4f994facea64d2b3106c), [SDL3 ref](https://wiki.libsdl.org/SDL3/SDL_GetWindowDisplayScale)).
The PR can be tested with:
```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        const Vector2 monitorScale = GetMonitorScale(GetCurrentMonitor());

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText(TextFormat("monitor scale %fx%f", monitorScale.x, monitorScale.y), 20, 20, 20, BLACK);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -480,6 +480,13 @@ int GetMonitorRefreshRate(int monitor)
     return 0;
 }
 
+// Get specified monitor scale
+Vector2 GetMonitorScale(int monitor)
+{
+    TRACELOG(LOG_WARNING, "GetMonitorScale() not implemented on target platform");
+    return 0;
+}
+
 // Get the human-readable, UTF-8 encoded name of the selected monitor
 const char *GetMonitorName(int monitor)
 {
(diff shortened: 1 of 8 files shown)


asdqwe:
After further review `GetMonitorScale()` will basically end up returning the same as `GetWindowScaleDPI()`, so I guess no need for this function afterall.

PR #4526, opened 2024-11-21
[rcore] [web] Add implementation to `GetWindowScaleDPI()` for `PLATFORM_WEB`

asdqwe:
Ref: [doc](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio), #4525. The PR can be tested with:

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        const Vector2 scale = GetWindowScaleDPI();

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText(TextFormat("scale %fx%f", scale.x, scale.y), 20, 20, 20, BLACK);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -771,8 +771,11 @@ Vector2 GetWindowPosition(void)
 // Get window scale DPI factor for current monitor
 Vector2 GetWindowScaleDPI(void)
 {
-    TRACELOG(LOG_WARNING, "GetWindowScaleDPI() not implemented on target platform");
-    return (Vector2){ 1.0f, 1.0f };
+    // NOTE: Returned scale is relative to the current monitor where the browser window is located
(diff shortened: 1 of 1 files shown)


raysan5:
@asdqwe Thank you very much for the addition!

BTW, I saw your closed PR about `GetMonitorScale()` but you were to fast at closing it; yes, it was same as `GetWindowScaleDPI()`. Actually, I'm thinking if `GetMonitorScaleDPI()` could be a better naming for it. Something to consider in a future...

As a side note, there is an [open Discussion](https://github.com/raysan5/raylib/discussions/4507) to refactor fullscreen modes that could be somewhat related to High-DPI windowing. Definitely not an easy problem considering all the possible configurations and platforms.

PR #4531, opened 2024-11-22
fixed platform change for zig build file

HaxSam:
Like already discussed here https://github.com/raysan5/raylib/pull/4358#issuecomment-2488242077 I made it possible to change the platform over the zig build.
I also saw that it wasn't possible to change if raudio, rmodels, and so on is enabled or not, I also made a fix for that, and It's now customizable over the zig build.

It would be nice if someone who already did some changes to the zig build could look into this PR.


raysan5:
@HaxSam thanks for the review!

PR #4535, opened 2024-11-23
[raymath] fix C++ operator overloads

ListeriaM:
some functions overloaded the wrong operator or multiplied instead of dividing


raysan5:
@ListeriaM thanks for the review!

PR #4537, opened 2024-11-24
Add cross product overload

joako1721:
Added operator^ overload for cross product, chosen for its visual similarity.


raysan5:
@joako1721 Is this operator a standard among math libraries?

joako1721:
The standard in math is the '×' symbol, but since it can't be overloaded, '^' was chosen. However, I now see this could cause confusion.

raysan5:
@joako1721 thanks, I prefer not to merge it if it is not a standard convention between math libraries.

PR #4539, opened 2024-11-25
[raudio] Fixed buffer overflow when loading WAV files

CalebHeydon:
This PR fixes a heap buffer overflow in raudio that can be triggered by a WAV file with a very large frame count. This is the result of two separate issues. The first is an integer overflow in the call to malloc since the unsigned integer can be overflowed when multiplied by the maximum number of channels (256). The second is using the 64-bit value for frame count even though it is cast to an unsigned int when allocating the buffer.

To fix this, I cast the frame count to a size_t in the malloc call, and I used 32-bit value for the number of frames to be written to the buffer. I recognize that issues like these aren't critical since the assets are almost always trusted, but it is nice to fix in case anyone wants to incorporate user-generated content into their game.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -801,10 +801,10 @@ Wave LoadWaveFromMemory(const char *fileType, const unsigned char *fileData, int
             wave.sampleRate = wav.sampleRate;
             wave.sampleSize = 16;
             wave.channels = wav.channels;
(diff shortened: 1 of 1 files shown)


orcmid:
I'm not equipped to confirm the correctness of the patch.  I do applaud the motivation and approach.

PR #4546, opened 2024-11-27
The Jaylib binding is at 5.5

glowiak:
Noticed that the entry's been outdated for quite a while.


raysan5:
@glowiak Nice! Thanks for the update!

Issue #4411, opened 2024-10-21
[config.h] Consider changing how defines are used in raylib to support configuration from build systems


planetis-m:
**Description**:
Currently, raylib uses `#if defined(SUPPORT_X)` to conditionally enable/disable certain features like file format support. However, setting a define to `0` from the build system (e.g., `-DSUPPORT_FILEFORMAT_PNG=0`) doesn't actually disable the feature because the macro is still considered "defined." 

This limitation makes it difficult to selectively disable certain features without manually editing `config.h`, especially when working with different build systems or when trying to configure raylib via command-line flags alone.

**Proposed Change**:
Switch the current `#if defined(SUPPORT_FILEFORMAT_X)` checks to something like `#if SUPPORT_FILEFORMAT_X == 1`. This would allow features to be properly toggled off by defining them as `0` from the build system.

Example:

```c
#ifndef SUPPORT_FILEFORMAT_PNG
#define SUPPORT_FILEFORMAT_PNG 1
#endif

#if SUPPORT_FILEFORMAT_PNG
    // PNG support code
#endif
```

This would enable features to be turned off using `-DSUPPORT_FILEFORMAT_PNG=0` in the build system, without requiring edits to the default configuration file.

**Discussion Summary**:
This topic was discussed on the raylib Discord server. The current method was found to be limiting for [build systems that rely on command-line defines](https://github.com/raysan5/raylib/pull/4407). While the change is agreed to be useful, it was also acknowledged that it's better suited for a future release after the current planned version has been completed.

znichola:
I think this could be achieved by passing the macro with the [`-U` flag](https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#index-U) to undefine it, at least this is available with gcc. Not sure how easy this would be to make portable, but it does seem like a lower impact solution.

planetis-m:
@znichola no that won't work, we have already tried that. Here's a test you can try:

```c
#include <stdio.h>

#ifndef EXAMPLE_OPTION
#define EXAMPLE_OPTION  1
#endif

int main(void)
{
#if EXAMPLE_OPTION == 1
    puts("option enabled");
#endif
}
```

znichola:
Ah yes of course I see that's dumb of me, I was thinking from the angle of disabling flags that were added with `-D` (if for instance using a raylib wrapper and not wanting to modify its build script), but that is not the case with the config.h.

sagehane:
[...]

PR #4548, opened 2024-11-28
[RCORE] Make some config defines settable by build systems

JeffM2501:
The way config.h and it's 'SUPPORT_FEATURE' flags are setup right now makes it impossible for build systems to change these flags without fully redefining every feature in config.h. This is not easy for people to do, and is counter to raylib's simple design.

This PR takes 3 of those flags and reverses the sense of them, turning them from SUPPORT_ flags to DISABLE_ flags.
* Screen Capture
* GIF recording
* High Resolution Timer

This means that by default raylib will enable the code that implements these features, unless the build system defines the specific DISABLE_ feature flag.

By reversing these defines, it makes it very easy for people to simply add the things they want to turn off to a makefile or project. Raylib has gone from 'add stuff to it' to 'remove stuff you don't need from it'.

I simply picked 3 flags that made the most sense to disable, if this concept is accepted we can apply it additional features.
The big ones here are screen capture and gif recording since they use a specific key that people may want to use for other features (this is my use case)

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -43,29 +43,42 @@
 //------------------------------------------------------------------------------------
 // Camera module is included (rcamera.h) and multiple predefined cameras are available: free, 1st/3rd person, orbital
 #define SUPPORT_CAMERA_SYSTEM           1
+
 // Gestures module is included (rgestures.h) to support gestures detection: tap, hold, swipe, drag
 #define SUPPORT_GESTURES_SYSTEM         1
+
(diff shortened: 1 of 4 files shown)


raysan5:
@JeffM2501 I'm afraid thsi change breaks consistency. I prefer to look for a more consistent solution to this issue, probably following discussion here https://github.com/raysan5/raylib/issues/4411

JeffM2501:
I was unaware of that issue, I will make a PR that does that. I have a need to be able to disable features.
The solution in that PR works for me, we just need to do something.

PR #4553, opened 2024-11-29
Update CONTRIBUTING.md + FAQ.md: small misprints

Mossieur-Patate:
Fixed two small misprints.

1. Before: ‘You can check other issues to see how is being done!‘
After: ‘You can check other issues to see how it's being done!

2. Before: ‘I would say "simplicity" and "enjoyment" at a really low level of coding but actually is up to the user to discover it, to try it and to see if it fits their needs.’
After: ‘I would say "simplicity" and "enjoyment" at a really low level of coding but actually it is up to the user to discover it, to try it and to see if it fits their needs.’

(I suspect these errors are tied to how Spanish can work with implicit pronouns, by the way. ;))


raysan5:
@Mossieur-Patate Thanks for the reviews!

PR #4554, opened 2024-11-29
[Config.h] Make all the defines in config.h be override-able by the build system

JeffM2501:
Per issue #4411 this PR makes all the config defines first check if they have been set by the build system, and doesn't just check for the define, but instead check the value.
This allows the config.h values to be individually set by the build system.

### DIFF
--- a/examples/shapes/shapes_rectangle_advanced.c
+++ b/examples/shapes/shapes_rectangle_advanced.c
@@ -61,7 +61,7 @@ void DrawRectangleRoundedGradientH(Rectangle rec, float roundnessLeft, float rou
     const Vector2 centers[4] = { point[8], point[9], point[10], point[11] };
     const float angles[4] = { 180.0f, 270.0f, 0.0f, 90.0f };
 
-#if defined(SUPPORT_QUADS_DRAW_MODE)
+#if SUPPORT_QUADS_DRAW_MODE
     rlSetTexture(GetShapesTexture().id);
     Rectangle shapeRect = GetShapesTextureRectangle();
 
--- a/src/config.h
+++ b/src/config.h
@@ -32,52 +32,107 @@
 // Module selection - Some modules could be avoided
 // Mandatory modules: rcore, rlgl, utils
 //------------------------------------------------------------------------------------
+#ifndef SUPPORT_MODULE_RSHAPES
 #define SUPPORT_MODULE_RSHAPES          1
+#endif
+
+#ifndef SUPPORT_MODULE_RTEXTURES
 #define SUPPORT_MODULE_RTEXTURES        1
+#endif 
+
+#ifndef SUPPORT_MODULE_RTEXT
 #define SUPPORT_MODULE_RTEXT            1       // WARNING: It requires SUPPORT_MODULE_RTEXTURES to load sprite font textures
+#endif
+
+#ifndef SUPPORT_MODULE_RMODELS
 #define SUPPORT_MODULE_RMODELS          1
+#endif
+
+#ifndef SUPPORT_MODULE_RAUDIO
 #define SUPPORT_MODULE_RAUDIO           1
+#endif
 
 //------------------------------------------------------------------------------------
 // Module: rcore - Configuration Flags
 //------------------------------------------------------------------------------------
 // Camera module is included (rcamera.h) and multiple predefined cameras are available: free, 1st/3rd person, orbital
+#ifndef SUPPORT_CAMERA_SYSTEM
 #define SUPPORT_CAMERA_SYSTEM           1
+#endif
+
 // Gestures module is included (rgestures.h) to support gestures detection: tap, hold, swipe, drag
+#ifndef SUPPORT_GESTURES_SYSTEM
 #define SUPPORT_GESTURES_SYSTEM         1
+#endif
+
 // Include pseudo-random numbers generator (rprand.h), based on Xoshiro128** and SplitMix64
+#ifndef SUPPORT_RPRAND_GENERATOR
 #define SUPPORT_RPRAND_GENERATOR        1
+#endif
+
 // Mouse gestures are directly mapped like touches and processed by gestures system
+#ifndef SUPPORT_MOUSE_GESTURES
 #define SUPPORT_MOUSE_GESTURES          1
+#endif
+
 // Reconfigure standard input to receive key inputs, works with SSH connection.
+#ifndef SUPPORT_SSH_KEYBOARD_RPI
 #define SUPPORT_SSH_KEYBOARD_RPI        1
+#endif 
+
 // Setting a higher resolution can improve the accuracy of time-out intervals in wait functions.
 // However, it can also reduce overall system performance, because the thread scheduler switches tasks more often.
+#ifndef SUPPORT_WINMM_HIGHRES_TIMER
 #define SUPPORT_WINMM_HIGHRES_TIMER     1
-// Use busy wait loop for timing sync, if not defined, a high-resolution timer is set up and used
-//#define SUPPORT_BUSY_WAIT_LOOP          1
+#endif
+
+// Use busy wait loop for timing sync, if 0, a high-resolution timer is set up and used (Default off)
+#ifndef SUPPORT_BUSY_WAIT_LOOP
+#define SUPPORT_BUSY_WAIT_LOOP          0
+#endif
+
 // Use a partial-busy wait loop, in this case frame sleeps for most of the time, but then runs a busy loop at the end for accuracy
+#ifndef SUPPORT_PARTIALBUSY_WAIT_LOOP
 #define SUPPORT_PARTIALBUSY_WAIT_LOOP    1
+#endif
+
 // Allow automatic screen capture of current screen pressing F12, defined in KeyCallback()
+#ifndef SUPPORT_SCREEN_CAPTURE
 #define SUPPORT_SCREEN_CAPTURE          1
+#endif 
+
 // Allow automatic gif recording of current screen pressing CTRL+F12, defined in KeyCallback()
+#ifndef SUPPORT_GIF_RECORDING
 #define SUPPORT_GIF_RECORDING           1
+#endif
+
 // Support CompressData() and DecompressData() functions
+#ifndef SUPPORT_COMPRESSION_API
 #define SUPPORT_COMPRESSION_API         1
+#endif
+
 // Support automatic generated events, loading and recording of those events when required
+#ifndef SUPPORT_AUTOMATION_EVENTS
 #define SUPPORT_AUTOMATION_EVENTS       1
-// Support custom frame control, only for advanced users
+#endif
+
+// Support custom frame control, only for advanced users (Default off)
 // By default EndDrawing() does this job: draws everything + SwapScreenBuffer() + manage frame timing + PollInputEvents()
 // Enabling this flag allows manual control of the frame processes, use at your own risk
-//#define SUPPORT_CUSTOM_FRAME_CONTROL    1
+#ifndef SUPPORT_CUSTOM_FRAME_CONTROL
+#define SUPPORT_CUSTOM_FRAME_CONTROL    0
+#endif
 
 // Support for clipboard image loading
 // NOTE: Only working on SDL3, GLFW (Windows) and RGFW (Windows)
+#ifndef SUPPORT_CLIPBOARD_IMAGE
 #define SUPPORT_CLIPBOARD_IMAGE    1
+#endif
 
 // NOTE: Clipboard image loading requires support for some image file formats
 // TODO: Those defines should probably be removed from here, I prefer to let the user manage them
-#if defined(SUPPORT_CLIPBOARD_IMAGE)
+
+#if SUPPORT_CLIPBOARD_IMAGE
     #ifndef SUPPORT_MODULE_RTEXTURES
         #define SUPPORT_MODULE_RTEXTURES 1
     #endif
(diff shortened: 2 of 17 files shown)


Peter0x44:
Are you sure this is enough? Anything you pass on the command line will still get redefined in config.h itself.
This lets you turn on extra things, but not turn off things, and not change the value of anything.

JeffM2501:
> Are you sure this is enough? Anything you pass on the command line will still get redefined in config.h itself.
> 

Yes because config.h now checks if the label is not already defined first before defining it, so things on the command line will be defined before the preprocessor runs and config.h will not replace them. This ensures that all things are still defined. I have tested this in msvc and get expected results, I can build raylib with a define in the build system only to disable screenshots. Please test it in other systems.

What this won't do is disable things at runtime, since it's all still baked into the compile with preprocessor directives.

Peter0x44:
I missed the change to config.h because GitHub collapsed it, sorry! I will test this with gcc/make on Linux later.

Peter0x44:
I tested this with:
`make CUSTOM_CFLAGS="-DSUPPORT_FILEFORMAT_PNG=0` and I was then unable to load PNGs as expected
LGTM

raysan5:
@JeffM2501 I see the usefulness of this big change but trying to understand what problem it is **currently** addressing. As per the issue description:

 > This limitation makes it difficult to selectively disable certain features without manually editing config.h, especially when working with different build systems or when trying to configure raylib via command-line flags alone.

Yeah, that's right but, afaik, there have been no real use-case issues related since `config.h` was introduced. Also, this change, makes `config.h` bigger and less readable for new users. For my specific use cases, the tools I develop, I just overwhite `config.h` (defining `EXTERNAL_CONFIG_FLAGS`) and define the flags I want, depending on the build system I use ([VS2022 example](https://github.com/raysan5/rfxgen/blob/master/projects/VS2022/raylib/Directory.Build.props), [Linux/Web](https://github.com/raysan5/rfxgen/blob/master/.github/workflows/linux.yml#L18)). I have to see how this change affects my pipelines because it could require updating +40 projects...

I don't know... I got the feeling this redesign addresses a potential issue that nobody really had... at least yet.

JeffM2501:
I am trying to make a game where I allow the user to map items to F12, I can not because the built in screen capture and recording is using that key. That is my real world case.

JeffM2501:
You can still overwrite config.h and it will be the same as before. This just means you do not have to if you want another option. You are the author of raylib so it's not that big of a deal for you to just make a new config for each project, You always know what is in config.h. But if someone does not want to fork raylib, or does not know everything that is in raylib, that makes it much harder for them to keep raylib up to date.

raysan5:
@JeffM2501 Thanks for the further explanation, just one last concern, if I define `make -DSUPPORT_FILEFORMAT_KTX`, does it default to 1 and gets enabled? Or is it required to `make -DSUPPORT_FILEFORMAT_KTX=1`?

JeffM2501:
> @JeffM2501 Thanks for the further explanation, just one last concern, if I define `make -DSUPPORT_FILEFORMAT_KTX`, does it default to 1 and gets enabled? Or is it required to `make -DSUPPORT_FILEFORMAT_KTX=1`?

It is required that you set the value to 1. 
#define NAME does not add any default value, since #define is just a string replacement, The previous version simply checked if the define existed, not what it's value is. This was the core problem, there was no way to change it's value once it's defined. #define NAME defines NAME  as nothing, so when anything that uses NAME, it gets replaced with nothing.
so #if NAME comes out to just #if and is a syntax error.

Peter0x44:
@JeffM2501 @raysan5  That's not true!!! I just tested myself
```c
#include <stdio.h>
int main(void)
{
#if EXAMPLE_OPTION
                puts("option enabled");
#else
                puts("option disabled");
#endif
}
```
```
$ gcc test.c && ./a.exe
option disabled
$ gcc test.c -DEXAMPLE_OPTION && ./a.exe
option enabled
$ gcc test.c -DEXAMPLE_OPTION=0 && ./a.exe
option disabled
$ gcc test.c -DEXAMPLE_OPTION=1 && ./a.exe
option enabled
```

It doesn't need to be =1 to enable, no compromise necessary

raysan5:
@Peter0x44 thanks for the test! that's really great to know! Is that behaviour consistent between compilers or it is actually "Undefined Behaviour" and could be managed in other ways by other compilers? I'm trying to find this info but I couldn't...

UPDATE: ChatGPT tells me:
```
Yes, if you define a macro using the -D flag without specifying a value, it implicitly evaluates to 1. This behavior follows the convention of the C preprocessor.
```

Peter0x44:
https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html
> -D name
> Predefine name as a macro, with definition 1.

raysan5:
**Testing ChatGPT, quite impressive! 😄 **

---

The choice between checking if a flag is defined (`#ifdef`/`#if defined()`) or checking its value (`#if CUSTOM_FLAG`) depends on your specific use case and coding style. Here's a breakdown of the two approaches and their implications:

---

### **1. Checking if a Flag is Defined (`#ifdef` or `#if defined()`)**

#### Syntax:
```c
#ifdef CUSTOM_FLAG
// Code for enabled feature
#endif
```

Or:
```c
#if defined(CUSTOM_FLAG)
// Code for enabled feature
#endif
```

#### Behavior:
- The feature is considered "enabled" if `CUSTOM_FLAG` is defined, regardless of its value.
- The flag does not need to have a numeric value.
- If `CUSTOM_FLAG` is not defined, the code is excluded.

#### Use Case:
- **When you only care whether a flag is defined or not.**
- Ideal for simple on/off toggles without requiring numeric or value-based logic.

#### Pros:
- Simple and clean for basic feature toggles.
- Does not require explicitly setting the value of the flag.

#### Cons:
- Cannot easily support more complex configurations (e.g., different modes).
- The code behavior depends solely on whether the flag exists, not its value.

---

### **2. Checking a Flag's Value (`#if CUSTOM_FLAG`)**

#### Syntax:
```c
#if CUSTOM_FLAG
// Code for enabled feature
#endif
```

#### Behavior:
- The feature is considered "enabled" if `CUSTOM_FLAG` evaluates to a truthy value (non-zero).
- The flag must either be defined with a numeric value (`-DCUSTOM_FLAG=1`) or default to `1` when defined with `-DCUSTOM_FLAG`.

#### Use Case:
- **When you might want to assign the flag a specific value, or use it in numeric or multi-level configurations.**
- Ideal for more flexible setups, e.g., enabling multiple modes or levels of a feature.

#### Pros:
- Supports more nuanced configurations (e.g., `-DFEATURE_MODE=2` for mode-specific behavior).
- Explicitly checking values makes the intent clearer when toggling features programmatically.

#### Cons:
- If the macro is not defined, you need to ensure it defaults to `0` to prevent compilation errors. Use:
  ```c
  #ifndef CUSTOM_FLAG
  #define CUSTOM_FLAG 0
  #endif
  ```

---

### **Comparison**

| Aspect                    | `#if defined(CUSTOM_FLAG)`         | `#if CUSTOM_FLAG`              |
|---------------------------|-------------------------------------|---------------------------------|
| **Simplicity**            | Good for simple on/off toggles     | Slightly more complex          |
| **Flexibility**           | Limited to "defined" vs "not"      | Can use numeric values         |
| **Error Handling**        | No issues if undefined             | Requires `#ifndef` fallback    |
| **Multi-level Config**    | Not supported                      | Supported                      |
| **Code Clarity**          | Clearly indicates "defined check"  | Explicit about value logic     |

---

### **Best Practice**

1. **For simple toggles** (e.g., enable/disable a feature):  
   Use `#if defined(CUSTOM_FLAG)` or `#ifdef CUSTOM_FLAG`.  

2. **For configurable features** (e.g., modes, levels):  
   Use `#if CUSTOM_FLAG` and assign meaningful values during compilation (e.g., `-DFEATURE_LEVEL=2`).  

3. **To ensure robustness**:  
   Combine both patterns by defining defaults for macros in your code to avoid accidental undefined behavior:
   ```c
   #ifndef CUSTOM_FLAG
   #define CUSTOM_FLAG 0
   #endif

   #if CUSTOM_FLAG
   // Feature is enabled
   #endif
   ```

This way, your code is safer and can handle cases where the macro is not defined.

JeffM2501:
image
This is exactly what this PR does.

sleeptightAnsiC:
> I see the usefulness of this big change but trying to understand what problem it is currently addressing.

This change would be really useful for people who vendor their dependencies with something like [git-submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) or [Cmake's FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html), and for someone who wants to support multiple build configurations in the project. In those cases editing config file by hand really sucks™ and current approaches when dealing with this are not so great (you either had to manage patches yourself, or have a script that edits the config).

> Is that behaviour consistent between compilers

According to [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/c99.html) - `-D name[=value]   Define name as if by a C-language #define directive. If no = value is given, a value of 1 shall be used.` - so this should work the same across GCC, Clang, Emscripten, TCC, and other smaler compilers. MSVC also seems to follow this rule according to [their docks](https://learn.microsoft.com/en-us/cpp/build/reference/d-preprocessor-definitions?view=msvc-140) but I'm not sure for how long. Generally, this is well defined for `-D` to default to `1`. (oh gosh, watch out for LLMs, those gaslight A LOT)

--- 

However, change is super big and there is already an option to exclude config from build process. You can even do this at command line with build system e.g. these allow you to inject library without changing any file `-DEXTERNAL_CONFIG_FLAGS -include raylib/src/config.h -USUPPORT_FILEFORMAT_PNG -USUPPORT_MODULE_RAUDIO -DSUPPORT_FILEFORMAT_BMP`. Most compilers and build systems should support equivalents of `-U` and `-D` options, though `-include` is NOT very compliant and I don't see TCC supporting it. A quick workaround is to use separate file with all -D -U and use it with `@file` options. This is also not compliant but seems to be better supported.

Peter0x44:
Are there any concerns left holding this up?

raysan5:
@Peter0x44 All my tools pipelines depend `config.h` flags customization and I have to make sure nothing breaks. Also it adds more code complexity to the config file multiplying the number of lines and reducing, imho, readability for new users. Also, afaik, all checks for flags in raylib (platform, os, arch...) use the defined() convention.

If the main concern for this change is that some default defined flags are problematic (screenshot capture), they can just be reviewed, instead of redesigning all the flags system approach.

Still, I want to review it more carefully. It's a big change.

decromo:
Chiming in on the readability aspect: I'd suggest keeping `config.h` like it is now on master, but adding a suffix to all macro names (e.g. `_DEFAULT`). Then you could apply those default values at the bottom if they were not overridden. Example:
```c
...
// Camera module is included (rcamera.h) and multiple predefined cameras are available: free, 1st/3rd person, orbital
#define SUPPORT_CAMERA_SYSTEM_DEFAULT           1
// Gestures module is included (rgestures.h) to support gestures detection: tap, hold, swipe, drag
#define SUPPORT_GESTURES_SYSTEM_DEFAULT         1

// rest of header

#ifndef SUPPORT_CAMERA_SYSTEM
#define SUPPORT_CAMERA_SYSTEM SUPPORT_CAMERA_SYSTEM_DEFAULT   // apply default value only if it hasn't been overridden
#endif
#ifndef SUPPORT_GESTURES_SYSTEM
#define SUPPORT_GESTURES_SYSTEM SUPPORT_GESTURES_SYSTEM_DEFAULT
#endif
...
```
You keep the top of the header as a readable list of the given options with their default values and helpful comments, and all the ugly preprocessor stuff is at the bottom.
Although, using `_DEFAULT` as a suffix leads to some stupidly named macros (`RL_DEFAULT_*_DEFAULT` and `LOAD_DEFAULT_FONT_DEFAULT`), so we would need to come up with a better suffix (not a huge issue all considered).

Now, this does come at the cost of a little complexity: I imagine that if a user wanted to change, for instance, the maximum number of gamepads, they might try to define `MAX_GAMEPADS_DEFAULT` (since that's what they see in the header); of course that wouldn't work as they need to know to define just `MAX_GAMEPADS` (i.e. to elide the suffix). This could be explained somewhere (like a comment at the top of the file), but the more seamless solution would be doing the opposite strategy:
```c
...
#define MAX_GAMEPADS                    4       // Maximum number of gamepads supported

// rest of header

#ifdef MAX_GAMEPADS_OVERRIDE                    // when overriding from the build system...
#ifdef MAX_GAMEPADS                             // if present, remove previous definition...
#undef MAX_GAMEPADS
#endif
#define MAX_GAMEPADS MAX_GAMEPADS_OVERRIDE      // apply the overridden value.
#endif
...
```
This way, all of `config.h` (and probably its behaviour too) stays exactly the same, that ugly business gets added at the bottom out of sight, and the nuisance of dealing with suffixes (or it could be a prefix instead!) is exclusively confronted by the build system (which is a good tradeoff IMO since it is a new feature) e.g. `gcc -DMAX_GAMEPADS_OVERRIDE=5 main.c -lraylib`.

I haven't tested it in raylib, but I've used some similar strategies for my smaller utility headers; let me know if you enjoy this direction so I can try it out here.

Peter0x44:
I think this PR is long overdue a merge. It would be great for raylib 6.0

raysan5:
Config system has been reviewed, details on the commit https://github.com/raysan5/raylib/commit/5361265a7d5d1ef662077350ef3db9747333b95b.

PR #4556, opened 2024-11-30
Update documentation for Vector2Angle

mrjonjonjon:
Clarified that the angle is signed and that positive angles appear visually clockwise. I am coming from Unity and their [Vector2.Angle](https://docs.unity3d.com/ScriptReference/Vector2.Angle.html) is unsigned so this confused me for some time. This might help others too.


raysan5:
@mrjonjonjon thanks for the review

PR #4557, opened 2024-11-30
[rmodels] Fix null pointer dereference in LoadImageFromCgltfImage

CalebHeydon:
This fixes a null pointer dereference when loading a texture from a GLB file. The fix is just to check if the buffer_view is NULL before checking if the data pointer is NULL.


raysan5:
@CalebHeydon thank for the review

PR #4561, opened 2024-12-01
use unused variable. Fixes #4560

### LINKED ISSUE #4560: [rlgl] `rlDrawVertexArrayInstanced()` doesn't use the offset variable.
rlDrawVertexArrayInstanced doesn't use the offset variable.
https://github.com/raysan5/raylib/blob/58fe34d9cc2dd1929efb864e5222f25253c4ad89/src/rlgl.h#L3959

Instead, glDrawArraysInstanced uses for it's "first" parameter ([reference](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDrawArraysInstanced.xhtml)) always zero. 

Should zero be  "count" instead? 

So instead of 

```c

void rlDrawVertexArrayInstanced(int offset, int count, int instances)
{
#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
    glDrawArraysInstanced(GL_TRIANGLES, 0, count, instances);
#endif
}
```

It should be 

```c
void rlDrawVertexArrayInstanced(int offset, int count, int instances)
{
#if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
    glDrawArraysInstanced(GL_TRIANGLES, offset /*<--!*/, count, instances);
#endif
}
```

RicoP:
Fixes #4560

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -3959,7 +3959,7 @@ void rlDrawVertexArrayElements(int offset, int count, const void *buffer)
 void rlDrawVertexArrayInstanced(int offset, int count, int instances)
 {
 #if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
-    glDrawArraysInstanced(GL_TRIANGLES, 0, count, instances);
+    glDrawArraysInstanced(GL_TRIANGLES, offset, count, instances);
 #endif
 }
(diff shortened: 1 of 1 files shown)


raysan5:
@RicoP thanks for the review!

PR #4563, opened 2024-12-02
[rmodels] Fix crash when NULL is passed to LoadImageFromCgltfImage

CalebHeydon:
This fixes a crash that can happen when null is passed to LoadImageFromCgltfImage. There is a case where the argument is null in LoadGLTF. A possible fix is to check if the it is null in LoadImageFromCgltfImage.


raysan5:
@CalebHeydon thanks for the review... note taht this function is only used internally to the module, I think this case is very unprobable to happen.

PR #4564, opened 2024-12-02
[rmodels] Fixed null pointer dereference in LoadGLTF

CalebHeydon:
This is the last I've found so far :)

I was able to get it to crash here when loading a file:
image

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5664,7 +5664,7 @@ static Model LoadGLTF(const char *fileName)
                 }
 
                 // Load primitive indices data (if provided)
-                if (mesh->primitives[p].indices != NULL && mesh->primitives[p].indices->buffer_view != NULL)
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -5664,7 +5664,7 @@ static Model LoadGLTF(const char *fileName)
                 }
 
                 // Load primitive indices data (if provided)
-                if (mesh->primitives[p].indices != NULL)
+                if (mesh->primitives[p].indices != NULL && mesh->primitives[p].indices->buffer_view != NULL)
```
raysan5, line 5667: Please, could you put parenthesis around the individual conditions? I just reviewed it yesterday in many files...

CalebHeydon: done

raysan5:
@CalebHeydon thanks for the review!

PR #4565, opened 2024-12-02
Added new Jai bindings to BINDINGS.md

ahmedqarmout2:
Added new Jai bindings for Raylib version 5.5 and latest Jai version


ahmedqarmout2:
@raysan5 thanks for the feedback
Jai does not have any official website/online documentations yet, as it is still in beta. This is the best we have so far. I will update the documentations link if I find better documentations.
I am still adding more examples, however, I've complied and tested the examples I've ported

PR #4570, opened 2024-12-03
[documentation] Updating keybindings versions

legendaryredfox:
Went through the repos for the bindings and updated some of them. Repos that have not been updated in a while may have remained outdated.

I would like to point out that both raylib-lua and rayboo are 404, but I did not know if it was ok to remove them


raysan5:
@legendaryredfox thanks for the review! I think 404 can be removed.

PR #4572, opened 2024-12-03
[documentation] Removing 404 repos from bindings

legendaryredfox:
Removed 2 repos that give 404 when accessed


raysan5:
@legendaryredfox thanks for the review! 👍

PR #4574, opened 2024-12-04
[rcore] fix ShaderUniformDataType to match rlgl

danilwhale:
This PR fixes bug with missing unsigned vectors from `rlShaderUniformDataType` in `ShaderUniformDataType` and causing `SAMPLER2D` from raylib to become `UINT` in rlgl.


danilwhale:
https://github.com/raysan5/raylib/pull/4577

PR #4576, opened 2024-12-05
Add LoadModelEx and UploadModel functions

devdad:
To enable custom model and mesh batching in projects that use the official raylib and instancing, it is essential to have control over whether or not meshes are automatically uploaded to the GPU. In my opinion, this is the cleanest solution for achieving that flexibility.

New functions added:
// Load model from files (mesh and material) without automatically uploading vertex data to the GPU
Model LoadModelEx(const char *fileName, bool upload) 

// Upload model data to the GPU (RAM and/or VRAM) 
void UploadModel(const Model *model, bool dynamic)

The existing LoadModel function now calls LoadModelEx with upload = true.

Additionally, fixed LoadOBJ to match the behavior of other load functions by not automatically uploading meshes.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1552,8 +1552,10 @@ RLAPI void DrawGrid(int slices, float spacing);                                 
 
 // Model management functions
 RLAPI Model LoadModel(const char *fileName);                                                // Load model from files (meshes and materials)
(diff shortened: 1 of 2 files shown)


raysan5:
@devdad I'm afraid I don't like this approach, I try to avoid functions with that kind of `bool` flags.

devdad:
@raysan5 ok np, but do you have some better suggestion, what we have now it is clearly not good too, because it force meshes to be uploaded straight away and to do work around is so hard because we can't call static LoadOBJ and other similar functions.

PR #4577, opened 2024-12-05
Make sure ShaderUniformDataType matches rlShaderUniformDataType

RicoP:
It seems like with version 5.5 ShaderUniformDataType and rlShaderUniformDataType went out of sync with the addition of RL_SHADER_UNIFORM_UIXXX caused  SHADER_UNIFORM_SAMPLER2D to be interpreted as RL_SHADER_UNIFORM_UINT in rlSetUniform.

Depending on how lenient your specific OpenGL version is that might work fine, but in my case this caused issues.

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -2591,8 +2591,28 @@
           "description": "Shader uniform type: ivec4 (4 int)"
         },
         {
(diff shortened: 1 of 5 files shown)


raysan5:
@RicoP Thanks for the review! I missed those additions on `raylib.h` side

PR #4580, opened 2024-12-05
[rlgl] Add support for glPolygonOffset()

mikeemm:
Added the following functions into `rlgl.h`;

```
void rlEnablePolygonOffsetFill(void);
void rlDisablePolygonOffsetFill(void);
void rlEnablePolygonOffsetLine(void);
void rlDisablePolygonOffsetLine(void);
void rlEnablePolygonOffsetPoint(void);
void rlDisablePolygonOffsetPoint(void);
void rlSetPolygonOffset(float factor, float units);
```

This allows us to call `glPolygonOffset()`, which helps minimize z-fighting when we want to draw lines or other shapes along the surface of a mesh.


raysan5:
@mikeemm I prefer to avoid that set of functions on `rlgl`, anyone with that specific need can use OpenGL directly.

PR #4583, opened 2024-12-07
[rlgl] Small cleaning

Bigfoot71:
I simply removed the `int format` parameter from `rlGenTextureMipmaps()`, which was unused, as well as the `rlTextureFilter` enum, which was also unused in `rlgl.h` and wasn't referenced anywhere else in the library. In any case, `TextureFilter` replaces it.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -474,6 +474,18 @@ typedef enum {
     RL_PIXELFORMAT_COMPRESSED_ASTC_8x8_RGBA        // 2 bpp
 } rlPixelFormat;
 
+// Texture parameters: filter mode
+// NOTE 1: Filtering considers mipmaps if available in the texture
+// NOTE 2: Filter is accordingly set for minification and magnification
+typedef enum {
+    RL_TEXTURE_FILTER_POINT = 0,        // No filter, just pixel approximation
+    RL_TEXTURE_FILTER_BILINEAR,         // Linear filtering
+    RL_TEXTURE_FILTER_TRILINEAR,        // Trilinear filtering (linear with mipmaps)
+    RL_TEXTURE_FILTER_ANISOTROPIC_4X,   // Anisotropic filtering 4x
+    RL_TEXTURE_FILTER_ANISOTROPIC_8X,   // Anisotropic filtering 8x
+    RL_TEXTURE_FILTER_ANISOTROPIC_16X,  // Anisotropic filtering 16x
+} rlTextureFilter;
+
 // Color blending modes (pre-defined)
 typedef enum {
     RL_BLEND_ALPHA = 0,                 // Blend textures considering alpha (default)
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -474,18 +474,6 @@ typedef enum {
     RL_PIXELFORMAT_COMPRESSED_ASTC_8x8_RGBA        // 2 bpp
 } rlPixelFormat;
 
-// Texture parameters: filter mode
-// NOTE 1: Filtering considers mipmaps if available in the texture
-// NOTE 2: Filter is accordingly set for minification and magnification
-typedef enum {
-    RL_TEXTURE_FILTER_POINT = 0,        // No filter, just pixel approximation
-    RL_TEXTURE_FILTER_BILINEAR,         // Linear filtering
-    RL_TEXTURE_FILTER_TRILINEAR,        // Trilinear filtering (linear with mipmaps)
-    RL_TEXTURE_FILTER_ANISOTROPIC_4X,   // Anisotropic filtering 4x
-    RL_TEXTURE_FILTER_ANISOTROPIC_8X,   // Anisotropic filtering 8x
-    RL_TEXTURE_FILTER_ANISOTROPIC_16X,  // Anisotropic filtering 16x
-} rlTextureFilter;
-
 // Color blending modes (pre-defined)
 typedef enum {
     RL_BLEND_ALPHA = 0,                 // Blend textures considering alpha (default)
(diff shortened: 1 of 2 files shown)


src/rlgl.h
```diff
@@ -474,18 +474,6 @@ typedef enum {
...
-// NOTE 2: Filter is accordingly set for minification and magnification
-typedef enum {
-    RL_TEXTURE_FILTER_POINT = 0,        // No filter, just pixel approximation
-    RL_TEXTURE_FILTER_BILINEAR,         // Linear filtering
-    RL_TEXTURE_FILTER_TRILINEAR,        // Trilinear filtering (linear with mipmaps)
-    RL_TEXTURE_FILTER_ANISOTROPIC_4X,   // Anisotropic filtering 4x
-    RL_TEXTURE_FILTER_ANISOTROPIC_8X,   // Anisotropic filtering 8x
-    RL_TEXTURE_FILTER_ANISOTROPIC_16X,  // Anisotropic filtering 16x
-} rlTextureFilter;
-
```
asdqwe, line 488: Please don't remove the `rlTextureFilter`.
Although it's not used on `raylib`, it's used when using `rlgl` standalone.

Bigfoot71: I committed the changes locally, but before pushing, could you let me know to what extent it is being used? I searched thoroughly but couldn't find any usage in `rlgl`

asdqwe: When using `rlgl` standalone (where raylib is not present, so no `raylib.h`, no `rtextures`, etc) it's used with (to set the correct) `rlTextureParameters()` similarly to `SetTextureFilter()` ([ref](https://github.com/raysan5/raylib/blob/602583bcdbc46f317915c811a7e8e2416e65bbc7/src/rtextures.c#L4360)).

raysan5: Sincerely, I don't know... I usually like to simplify code and function parameters... but I agree that `SetTextureFilter()` uses them and it could be a reference for `rlgl` users.

About the `format` parameter in `rlGenTextureMipmaps()`, it was kept there for a future implementation for OpenGL 1.1 that required the data size in memory... but implementation was removed at some point in raylib live...

asdqwe: I'd like to point a parallel with Blend Mode.
The [BlendMode](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/raylib.h#L902-L911) enum mirrors [rlBlendMode](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rlgl.h#L493-L502C3).
All [BeginBlendMode()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rcore.c#L1144-L1147) / [EndBlendMode()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rcore.c#L1150-L1153) do is call [rlSetBlendMode()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rlgl.h#L2101-L2136).
And [rlSetBlendMode()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rlgl.h#L2108-L2130) handles the switch from there.

The Texture Filter is basically the same.
The [TextureFilter](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/raylib.h#L868-L875) enum mirrors [rlTextureFilter](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rlgl.h#L483-L490).
All [SetTextureFilter()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rtextures.c#L4360-L4423) does is calls to [rlTextureParameters()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rtextures.c#L4369-L4420).
But it was chosen to handle the switch at [SetTextureFilter()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rtextures.c#L4362-L4422) instead of on `rlgl` (thus, incidentally, not using `rlTextureFilter`).

It would be nice to keep `rlTextureFilter` at least for the sake of structural consistency since Texture Filtering could very well be implemented inside `rlgl`, similarly to [rlSetBlendMode()](https://github.com/raysan5/raylib/blob/83f6f3dabeb14b8f5268bf3abdcc898f50cfeb17/src/rlgl.h#L2101-L2136).

Bigfoot71: Alright, understood, I just canceled the change

Bigfoot71:
I've just restored the `rlTextureFilter` enum following our conversation, so the commit only removes the unused `int format` parameter from `rlGenTextureMipmaps`.

It's up to you to decide whether it's better to keep this parameter in case it might be reused in the future.

src/rlgl.h
asdqwe: Thank you.

asdqwe:
Side note: if the `format` parameter will be required for expanding the `OpenGL 1.1` implementation, it could end up being necessary for some retro-console platform support ([ref](https://github.com/raysan5/raylib/blob/aeb33e6301fd614e1f38aed66ab6b219a824df3c/HISTORY.md?plain=1#L498)). So, if it gets removed, maybe it's a good idea to leave a note/comment mentioning that `OpenGL 1.1` will need that parameter (just to aid future reimplemenations, if/when necessary).

PR #4584, opened 2024-12-07
[examples] improve core_input_virtual_controls example

danilwhale:
Improved `core_input_virtual_controls` example a lot:
- Reviewed code to follow code conventions and match other examples (e.g. use `Vector2` instead of `float[2]`)
- Move buttons to enum for easier extension
- Fixed desktop input (buttons activating even when you don't press them)
- Added colored labels for buttons
- Fixed button label disappearing when you press it
- Replaced player rectangle with ball (to match other input examples)


raysan5:
@danilwhale thanks for the review, I'm merging it but notice it adds extra complexity that can be avoided, no need to get the stri ngs as separate objects or even all the buttons positions (just the middel point). I'll try to review it at some point.

PR #4589, opened 2024-12-09
[documentation] Adding a related project

legendaryredfox:
Even though the creator states this project is not 1:1 bindings, the project was clearly inspired on raylib, and uses many of its libraries


raysan5:
@legendaryredfox thanks for the addition... it's not a binding per-se but it's related, so I'm adding it..

PR #4594, opened 2024-12-12
[BUILD] Disable the threading example in MSVC for release and debug builds

JeffM2501:
This PR disables the threading example in MSVC for the debug and release targets. The project was already disabled for the DLL targets just not the static targets, this seems like a simple oversight. The project will not build on MSVC (and is commented as having that limitation). This simply lets MSVC build the solution without errors for static builds like it does for DLL builds.


raysan5:

PR #4596, opened 2024-12-12
[rtext] Removed inaccurate comment about negatives not being supported with TextToFloat and TextToInt Methods

hexmaster111:
I was going to add negative support for these methods, however, after investigation, I've found that it is actually already supported.

[rtext.c:1444](https://github.com/raysan5/raylib/blob/1f704be4e4aba6b25e56bde3c30cc136b6e94049/src/rtext.c#L1444)


raysan5:
@hexmaster111 oh! good catch! thanks for the review!

PR #4598, opened 2024-12-12
[rcore] [SDL2] Add implementation for `FLAG_WINDOW_ALWAYS_RUN`

### LINKED ISSUE #4588: [rcore] `GetFrameTime()` counts time since application was minimized
### Issue description

I'm using Raylib 5.0 with Rust but I'm presuming this is not a Rust specific bug. If it is then I'm sorry for wasting your time. If you minimize a Raylib application, when you unminimize the app, GetFrameTime() will return the total amount of time that has passed since you minimized the app. I feel like this probably shouldn't happen, anything that's moving based on frame time will move way too far when opening a minimized application, so you could probably clip through walls and stuff. 

### Environment

Raylib 5.0 with Rust (raylib-rs)
Windows 10
Platform backend: DESKTOP (GLFW)
Renderer: Intel(R) HD Graphics 530

### Code Example

Just print GetFrameTime() to the console if it's more than one second, so something like: 

```
if GetFrameTime() > 1.0 {
  printf(GetFrameTime());
}
```

asdqwe:
See https://github.com/raysan5/raylib/pull/4598#issuecomment-2554483960.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1739,7 +1739,12 @@ static void WindowContentScaleCallback(GLFWwindow *window, float scalex, float s
 static void WindowIconifyCallback(GLFWwindow *window, int iconified)
 {
     if (iconified) CORE.Window.flags |= FLAG_WINDOW_MINIMIZED;  // The window was iconified
-    else CORE.Window.flags &= ~FLAG_WINDOW_MINIMIZED;           // The window was restored
+    else
+    {
+        CORE.Window.flags &= ~FLAG_WINDOW_MINIMIZED;           // The window was restored
+
+        if ((CORE.Window.flags & FLAG_WINDOW_ALWAYS_RUN) == 0) CORE.Time.previous = GetTime();
+    }
 }
 
 // GLFW3 WindowMaximize Callback, runs when window is maximized/restored
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1341,6 +1341,9 @@ int InitPlatform(void)
     if ((CORE.Window.flags & FLAG_WINDOW_TOPMOST) > 0) glfwWindowHint(GLFW_FLOATING, GLFW_TRUE);
     else glfwWindowHint(GLFW_FLOATING, GLFW_FALSE);
 
+    // Enable FLAG_WINDOW_ALWAYS_RUN by default
+    if ((CORE.Window.flags & FLAG_WINDOW_ALWAYS_RUN) == 0) CORE.Window.flags |= FLAG_WINDOW_ALWAYS_RUN;
+
     // NOTE: Some GLFW flags are not supported on HTML5
     if ((CORE.Window.flags & FLAG_WINDOW_TRANSPARENT) > 0) glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_TRUE);     // Transparent framebuffer
     else glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_FALSE);  // Opaque framebuffer
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1341,9 +1341,6 @@ int InitPlatform(void)
     if ((CORE.Window.flags & FLAG_WINDOW_TOPMOST) > 0) glfwWindowHint(GLFW_FLOATING, GLFW_TRUE);
     else glfwWindowHint(GLFW_FLOATING, GLFW_FALSE);
 
-    // Enable FLAG_WINDOW_ALWAYS_RUN by default
-    if ((CORE.Window.flags & FLAG_WINDOW_ALWAYS_RUN) == 0) CORE.Window.flags |= FLAG_WINDOW_ALWAYS_RUN;
-
     // NOTE: Some GLFW flags are not supported on HTML5
     if ((CORE.Window.flags & FLAG_WINDOW_TRANSPARENT) > 0) glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_TRUE);     // Transparent framebuffer
     else glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_FALSE);  // Opaque framebuffer
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1499,6 +1499,8 @@ void PollInputEvents(void)
                             if ((CORE.Window.flags & SDL_WINDOW_MAXIMIZED) > 0) CORE.Window.flags &= ~SDL_WINDOW_MAXIMIZED;
                         }
                         #endif
+
+                        if ((CORE.Window.flags & FLAG_WINDOW_ALWAYS_RUN) == 0) CORE.Time.previous = GetTime();
                     } break;
 
                     case SDL_WINDOWEVENT_HIDDEN:
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -571,7 +571,7 @@ void SetWindowState(unsigned int flags)
     }
     if (flags & FLAG_WINDOW_ALWAYS_RUN)
     {
-        TRACELOG(LOG_WARNING, "SetWindowState() - FLAG_WINDOW_ALWAYS_RUN is not supported on PLATFORM_DESKTOP_SDL");
+        CORE.Window.flags |= FLAG_WINDOW_ALWAYS_RUN;
     }
     if (flags & FLAG_WINDOW_TRANSPARENT)
     {
@@ -658,7 +658,7 @@ void ClearWindowState(unsigned int flags)
     }
     if (flags & FLAG_WINDOW_ALWAYS_RUN)
     {
-        TRACELOG(LOG_WARNING, "ClearWindowState() - FLAG_WINDOW_ALWAYS_RUN is not supported on PLATFORM_DESKTOP_SDL");
+        CORE.Window.flags &= ~FLAG_WINDOW_ALWAYS_RUN;
     }
     if (flags & FLAG_WINDOW_TRANSPARENT)
     {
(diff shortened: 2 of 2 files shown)


raysan5:
@asdqwe Actually, stop running is my expected behaviour when minimizing the Window, I think it's the optimal option, so being default. 

About the other platforms, the ones that have the minimizing option I think implement similar alternatives (afair, Android also pauses the Activity and Web browsers pauses tabs activity when minimized/unselected tab).

I prefer to let the user configure it for the sub-optimal use case that is `ALWAYS_RUN`, but maybe the time counter can be reseted/paused if the flag is active.

Also, about the othe Desktop backends (SDL, RGFW), probably behaviour should be the same, just forgot about it when added.

Just thinking... what is the default behaviour for games on Windows?

asdqwe:
@raysan5 I'm not aware of any (big) game that fully stops processing like that when minimizing because that breaks timers, gameplay sincronization, animations, networking, etc. Most games just lower the framerate when minimized or unfocused (some even allow to set specific _foreground fps_ and _background fps_ values). Which was why I thought this was a bug when debugging #4588. Basically everything with timers break if this happens.

asdqwe:
@raysan5 I don't know how Android specifically handles it. But on `PLATFORM_WEB` (where we have little control of the execution), when, for example, we change tabs, the browser continues to process the game/canvas just lowering the frequency of it.

<details>
<summary>This example can test/confirm that behavior:</summary>
<br>

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");

    SetWindowState(FLAG_WINDOW_ALWAYS_RUN);

    SetTargetFPS(60);

        TraceLog(LOG_INFO, "GetTime():%f", GetTime());

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
     }
    CloseWindow();
    return 0;
}
```

When running it, after 1 second I changed tabs. Waited 10 seconds then returned. Here's the screenshot of the console log where it shows the execution dropping to once a second:

img
</details>

asdqwe:
@raysan5 Although I don't think that should be the default behavior, I've reverted my previous commit on `PLATFORM_DESKTOP_GLFW` and, instead, added implemenation for `FLAG_WINDOW_ALWAYS_RUN` on `PLATFORM_DESKTOP_SDL` which will now behave, by default, like `PLATFORM_DESKTOP_GLFW` and stop processing until some event happen while minimized.

This PR was tested with `SDL2.30.10` on `Linux Mint 22.0`. `SDL3` wasn't tested but should work since `SDL_WaitEvent()` still appears to be compatible in this use case ([doc](https://wiki.libsdl.org/SDL2/SDL_WaitEvent), [doc](https://wiki.libsdl.org/SDL3/SDL_WaitEvent)).

<details>
<summary>The PR can be tested with:</summary>
<br>

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        if (IsKeyPressed(KEY_ONE)) { MinimizeWindow(); TraceLog(LOG_INFO, "Minimized"); }
        if (IsKeyPressed(KEY_TWO)) { SetWindowState(FLAG_WINDOW_ALWAYS_RUN); TraceLog(LOG_INFO, "Set FLAG_WINDOW_ALWAYS_RUN"); }
        if (IsKeyPressed(KEY_THREE)) { ClearWindowState(FLAG_WINDOW_ALWAYS_RUN); TraceLog(LOG_INFO, "Clear FLAG_WINDOW_ALWAYS_RUN"); }

        TraceLog(LOG_INFO, "GetTime():%f.", GetTime());

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
     }
    CloseWindow();
    return 0;
}
```
</details>

Regarding `PLATFORM_WEB`, I don't think it will be possible to implement `FLAG_WINDOW_ALWAYS_RUN` there since the web browser has control over the main loop execution  ([ref](https://github.com/raysan5/raylib/blob/bdfbd6e8cc77b755d6552f90028ba8262d79ab92/examples/core/core_basic_window_web.c#L47C5-L47C29), [ref](https://emscripten.org/docs/api_reference/emscripten.h.html#c.emscripten_set_main_loop)), which is likely why `glfwWaitEvents()` doesn't work there ([ref](https://github.com/raysan5/raylib/blob/bdfbd6e8cc77b755d6552f90028ba8262d79ab92/src/platforms/rcore_web.c#L1082-L1084)).

raysan5:
@asdqwe thank you very much for the testing and detailed explanations, I'm still thinking about it and the best way to address it. Maybe letting control to the user is the right way... but I'm afraid many users would not control it and just let the application running at full speed on background... but maybe more advanced/professional users would prefer to control it... 🤔 

The other option is letting raylib to control it, like now, but instead of halting, reducing framerate... but in that case, probably better to give full control to user...

asdqwe:
> but I'm afraid many users would not control it and just let the application running at full speed on background...

That's true, specially if the user forgot to `SetTargetFPS()` or disabled `vsync`.

Yeah, I think you're right, the current default behavior is a good safety net and by the time the user need to control it, a simple `SetWindowState(FLAG_WINDOW_ALWAYS_RUN);` easily solves it.

<details>
<summary>Besides, if the user wants a different background framerate, it can easily be done like this:</summary>
<br>

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

    SetWindowState(FLAG_WINDOW_ALWAYS_RUN);
    int toggle = 0;

        if (!IsWindowFocused() || IsWindowHidden()) {
            if (!toggle) {
                SetTargetFPS(10);
                toggle = 1;
            }
        } else {
            if (toggle) {
                SetTargetFPS(60);
                toggle = 0;
            }
        }

        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawFPS(20, 20);
        EndDrawing();
     }
    CloseWindow();
    return 0;
}
```

Volcanic-Penguin:
Since I was the one who posted issue https://github.com/raysan5/raylib/issues/4588 which led to this issue being opened I figured I should give my two cents. I don't really care that much if the game stops or not when you minimize the game, but I do think it's bad if you minimize the game for 5 minutes and then when you open the game again `GetFrameTime()` returns 300 as it can cause all kinds of unexpected bugs. 

As a developer I currently have two options to get around this.

A) Set `FLAG_WINDOW_ALWAYS_RUN` to `true` to avoid the issue entirely. 
B) Replace all of my calls to `GetFrameTime()` with a wrapper function that calls `GetFrameTime()` and then caps the returning float  to `0.1` or something like that.

raysan5:
@Volcanic-Penguin yeah, definitely `GetFrameTime()` should be reseted if `not FLAG_WINDOW_ALWAYS_RUN` and window is minimized/iconified.

asdqwe:
In my honest opinion, `SetWindowState(FLAG_WINDOW_ALWAYS_RUN);` should be the way to go, it's a clean unintrusive self-contained call (just a flag to keep track of).

I really don't like the idea of starting to tamper with anything from `CORE.Time`, which is basically creating another problem (inconsistency) trying to workaround something else.

asdqwe:
@raysan5 As requested ([ref](https://github.com/raysan5/raylib/pull/4598#issuecomment-2552185241)), I've added a reset for `GetFrameTime()` when `FLAG_WINDOW_ALWAYS_RUN` is not set and the window was iconified (to be precise, when the window is raised) for `PLATFORM_DESKTOP_GLFW` and `PLATFORM_DESKTOP_SDL`. [c8f0f7c](https://github.com/raysan5/raylib/pull/4598/commits/c8f0f7c47dff4bda402f71298fb5fa101b485ec9) was the least friction solution I could find, since `CORE.Time.frame` will then be calculated normally upon returning ([ref](https://github.com/raysan5/raylib/blob/03ff864087a3ca46a60cd64ef709650cd8d824f5/src/rcore.c#L948-L953)).

This PR was tested with `GLFW` and `SDL2.30.10` on `Linux Mint 22.0`. `SDL3` wasn't tested but should work since these changes are agnostic.

**Disclaimer:** I don't like changes like [c8f0f7c](https://github.com/raysan5/raylib/pull/4598/commits/c8f0f7c47dff4bda402f71298fb5fa101b485ec9) and I cannot predict if it will cause any side effects on the future. But these changes were tested here and, so far, it works as expected.

<details>
<summary>The current PR can be tested with:</summary>
<br>

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        if (IsKeyPressed(KEY_ONE))   { MinimizeWindow();                         TraceLog(LOG_INFO, "MINIMIZED");      }
        if (IsKeyPressed(KEY_TWO))   { SetWindowState(FLAG_WINDOW_ALWAYS_RUN);   TraceLog(LOG_INFO, "ALWAYS RUN ON");  }
        if (IsKeyPressed(KEY_THREE)) { ClearWindowState(FLAG_WINDOW_ALWAYS_RUN); TraceLog(LOG_INFO, "ALWAYS RUN OFF"); }

        TraceLog(LOG_INFO, "GetTime():%f   GetFrameTime():%f", GetTime(), GetFrameTime());

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```
</details>

asdqwe:
I don't mint it being there and being the default, it's a good enough "guardrail" for learning/new users.
All I'm saying is I just hope adding handling like [c8f0f7c](https://github.com/raysan5/raylib/pull/4598/commits/c8f0f7c47dff4bda402f71298fb5fa101b485ec9) don't become a habit.

asdqwe:
As requested ([ref](https://github.com/raysan5/raylib/pull/4598#issuecomment-2543828388), [ref](https://github.com/raysan5/raylib/pull/4598#issuecomment-2552185241)), the PR was updated to instead add implementation for `FLAG_WINDOW_ALWAYS_RUN` on `PLATFORM_DESKTOP_SDL` and also reset the `GetFrameTime()` when `FLAG_WINDOW_ALWAYS_RUN` is not set and the window was iconified on `PLATFORM_DESKTOP_GLFW` and `PLATFORM_DESKTOP_SDL`.

<details>
<summary>PR can be tested with:</summary>
<br>

```
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        if (IsKeyPressed(KEY_ONE))   { MinimizeWindow();                         TraceLog(LOG_INFO, "MINIMIZED");      }
        if (IsKeyPressed(KEY_TWO))   { SetWindowState(FLAG_WINDOW_ALWAYS_RUN);   TraceLog(LOG_INFO, "ALWAYS RUN ON");  }
        if (IsKeyPressed(KEY_THREE)) { ClearWindowState(FLAG_WINDOW_ALWAYS_RUN); TraceLog(LOG_INFO, "ALWAYS RUN OFF"); }

        TraceLog(LOG_INFO, "GetTime():%f   GetFrameTime():%f", GetTime(), GetFrameTime());

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```

raysan5:
@asdqwe Nice! Thank you very much for the review!

PR #4600, opened 2024-12-14
[rlgl] Added a way to get information about the amount of draw calls

gk646:
Non invasive way to get information about the amount of draw calls made during each tick.
Its under the custom frame control features as its done automatically per default if no custom frame control is chosen.
Accessor exist so it can be adapted to custom frame control

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1088,6 +1088,7 @@ RLAPI int GetFPS(void);                                           // Get current
 RLAPI void SwapScreenBuffer(void);                                // Swap back buffer with front buffer (screen drawing)
 RLAPI void PollInputEvents(void);                                 // Register all input events
 RLAPI void WaitTime(double seconds);                              // Wait for some time (halt program execution)
+RLAPI int GetDrawCalls(void);                                     // Returns draw calls made during the last tick
 
 // Random values generation functions
 RLAPI void SetRandomSeed(unsigned int seed);                      // Set the seed for the random number generator
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -955,6 +955,7 @@ void EndDrawing(void)
         CORE.Time.frame += waitTime;    // Total frame time: update + draw + wait
     }
 
+    rlSwapDrawCallsCounter();
     PollInputEvents();      // Poll user events (before next frame update)
 #endif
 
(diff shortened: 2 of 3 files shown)


asdqwe:
What would be the use case of this call (besides debugging)?

Also, I don't think this should be called by default on the standard frame control.
It's non-essential and most use cases probably won't have an use for it.

gk646:
> What would be the use case of this call (besides debugging)?
> 
> Also, I don't think this should be called by default on the standard frame control. It's non-essential and most use cases probably won't have an use for it.

Well its just information you should have access to if you want. Its good for checking how many draw calls are made in different setups (using atlas vs none, ...).
It just in the default so it can be accessed easily and works right out of the box. 
Many things you dont usually need that are more expensive are done in there aswell (Automation events, gif recordings).
But it would be fine just to add it to rlgl.

asdqwe:
> Well its just information you should have access to if you want. Its good for checking how many draw calls are made in different setups (using atlas vs none, ...).
> It just in the default so it can be accessed easily and works right out of the box.

I think that the less `rlgl` has to carry, the better. It's the one part of raylib that is hard to trim of non-essential calls.

> Many things you dont usually need that are more expensive are done in there aswell (Automation events, gif recordings).

But all those can be removed easily by just not defining them on `config.h`. I can't remove this proposed change unless I edit `rlgl` manually.

gk646:
I just think that its information a rendering abstraction should expose in an official way. 
Also were arguing over two integer assignments. It will just be inlined. Those are 2 cpu cycles. 
If your modifying defines you most likely also know enough to remove the call if you want to.
As raylib is also aimed towards newcomers, thus most users will not change the default defines.
Also I my impression from raylib is that it doesnt optimized single digit cpu cycles if you get a valid new feature.

asdqwe:
Learning/new users have no need for this. And by the time they start working directly on rendering they can do this themselves. `rlgl` runs on really old and/or limited hardware, the less bloat it has by default, the better.

raysan5:
@gk646 I'm afraid I agree with @asdqwe mentioned points, most users don't need this function and it implies adding more globals, new functions and calls on rcore module. Also, this function seems to only consider draw calls accumulated on batching system, models drawing issue independent calls and depending on user code there could be more variables involved (like forcing batching flush). Users requiring this data, would probably also require VRAM memory usage and [probably way more](https://github.com/raysan5/raylib/issues/1295).

At this point I prefer to avoid exposing it.

PR #4602, opened 2024-12-14
[rmodels] Optimized `UpdateModelAnimationBones()` function

Kirandeep-Singh-Khehra:
- Updating bones only once instead for each `mesh`.
 - Updating only one `model.meshes[].boneMatrices` and then using deep copy for other meshes instead of calculating for each bone in each mesh.
 - Tried to keep asserts and condition at their places.

**Other points:**
 - Makes it a clean base/template/reference for other/new bone updation functions. Because if using this as template or modifying it, then some calculations done in one mesh can affect bones in other mesh in next iteration(doubles the effect in for next mesh) If some new variable used in loops is declared outside loop.
 - This new template is helpful for me to create new functions to update bones of model for animation system I'm working on.

 - [x] **Tested and is working fine for me.**

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -2270,38 +2270,62 @@ void UpdateModelAnimationBones(Model model, ModelAnimation anim, int frame)
     {
         if (frame >= anim.frameCount) frame = frame%anim.frameCount;
 
+        // Get first mesh which have bones
+        int firstMeshWithBones = -1;
+
         for (int i = 0; i < model.meshCount; i++)
         {
             if (model.meshes[i].boneMatrices)
             {
                 assert(model.meshes[i].boneCount == anim.boneCount);
-
-                for (int boneId = 0; boneId < model.meshes[i].boneCount; boneId++)
+                if (firstMeshWithBones == -1)
                 {
-                    Vector3 inTranslation = model.bindPose[boneId].translation;
-                    Quaternion inRotation = model.bindPose[boneId].rotation;
-                    Vector3 inScale = model.bindPose[boneId].scale;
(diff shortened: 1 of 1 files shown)


raysan5:
@Kirandeep-Singh-Khehra Thanks for the review. Did you profile this change to verify the optimization gain? Also, could it happen that only one model mesh has bones info and not others? Won't this solution implement one specific mesh transformations to all other meshes? 🤔

Kirandeep-Singh-Khehra:
> Did you profile this change to verify the optimization gain?

Hi Ray, I have not profiled for optimization gain but i can explain in terms of time complexity. Initially matrix transformations were calculated in **O(mn)** (*`m`* is number of meshes and *`n`* is number of bones.) but after this change it will be **O(n)* (*`n`* is number of bones). As **all the calculations were resulting in same bone matrix array for each mesh**. So, instead of calculating it for each bone of each mesh i tried to calculate the bone matrices of only first mesh which have bone Matrices and then copying the bone matrix to other meshes if they already have it.

> Also, could it happen that only one model mesh has bones info and not others?

Asserts are at same place as they were before. And there is a `if (model.mesh[i].boneMatrices)` check while selecting the index of first mesh with bones and while copying it to other meshes.
**Yes it is checking the presence of bone matrices before updating them.**

>  Won't this solution implement one specific mesh transformations to all other meshes? 🤔

Yes it will. As it was the same thing happening before. So, **instead of calculating same boneMatrix array for each mesh i am calculating it only once**.

raysan5:
@Kirandeep-Singh-Khehra thanks for your answers and the details, now I better see the improvement.

PR #4603, opened 2024-12-14
install rcamera.h

saxofon:
Small fix to export the rcamera header file.


raysan5:
@saxofon not sure if many users would need it but meging...

saxofon:
@raysan5 thanks! Not sure how projects like this https://github.com/saxofon/3D-testing.git could be built without it...  ;-)

raysan5:
> Not sure how projects like this https://github.com/saxofon/3D-testing.git could be built without it... ;-)

@saxofon Adding the library manually for the specific project needs, thousands of other projects just don't need it.

dc1248:
Thanks! I've just been kludging around it.

PR #4612, opened 2024-12-17
add SetGamepadVibration to rgfw and template platform

JupiterRider:
The implementation of the function `SetGamepadVibration` was missing.


raysan5:
@JupiterRider thanks for the review!

Issue #4604, opened 2024-12-14
[rlgl] Compile error for Android with OpenGL ES 3


Bigfoot71:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

I tried to build raylib for Android with the flag `GRAPHICS_API_OPENGL_ES3`, but I got this error:
```console
ld.lld: error: undefined symbol: glDrawBuffersEXT
```

For the line [1915 (rlgl.h)](https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L1915):
```c
#if defined(GRAPHICS_API_OPENGL_ES3)
            glDrawBuffersEXT(count, buffers);      ///< L1915
#else
            glDrawBuffers(count, buffers);
#endif
```

I simply replaced `glDrawBuffersEXT` with `glDrawBuffers`, just like the line below, and the compilation succeeded and everything works as expected.

Is there a reason to use `glDrawBuffersEXT` here? Because it seems to me that this is an extension for OpenGL ES 2 originally, am I wrong?

Bigfoot71:
Sorry, I really have a problem with titles :skull:

Bigfoot71:
To elaborate on what I was thinking -> https://registry.khronos.org/OpenGL/extensions/EXT/EXT_draw_buffers.txt

It does seem that this is an extension for OpenGL ES 2.0, inspired by the features of OpenGL ES 3.0, unless I missed something.

asdqwe:
You're correct, that `#if` is (very) wrong.
`OpenGL ES1` and `OpenGL ES2` should use `glDrawBuffersEXT()`.
`Open GL ES3` and `OpenGL 2.0` to `OpenGL 4.5` should use `glDrawBuffers()`
Refs: [1](https://emscripten.org/docs/optimizing/Optimizing-WebGL.html#migrating-to-webgl-2), [2](https://registry.khronos.org/OpenGL-Refpages/es3.0/), [3](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDrawBuffers.xhtml).

Bigfoot71:
Well, given the conditions at the top of the function, it seems that we just need to remove the extra `#if defined(GRAPHICS_API_OPENGL_ES3)` without causing any issues

asdqwe:
[...]

PR #4617, opened 2024-12-18
Fix `shaders_deffered_render.c` for OpenGL ES 3

Bigfoot71:
This fixes an incomplete framebuffer issue due to the use of a texture format not supported in ES 3.
This commit also adds more information on how to manage deferred rendering.
See: https://github.com/raysan5/raylib/issues/4604#issuecomment-2550665295


raysan5:
@Bigfoot71 thank you very much for the review!

PR #4622, opened 2024-12-20
setting MAX_LEVEL based on actual mipcount input

JettMonstersGoBoom:
if the user supplies less mips than would be generated normally the texture rendering will break. 
for example quake1&2 and halflife have 4 pre-generated mips included with the data, viewing those textures without this fix would result in visual errors.


raysan5:
@JettMonstersGoBoom merging but... is this parameter supported in all OpenGL required versions by the function?

JettMonstersGoBoom:
> @JettMonstersGoBoom merging but... is this parameter supported in all OpenGL required versions by the function?

it looks like it's supported in GLES2.0 too actually, you could safely move outside of the #ifdef I think. I don't have the setup to test that. I put it there because MIPS are automatically enabled by the filter change above.

PR #4623, opened 2024-12-20
[rtext] Centralize Text

nircoe:
Added two methods into rtext.c (and the methods declarations into raylib.h) -

1. CentralizeText (for default font)
2. CentralizeTextEx (for specific font)

Used the same logic as in "MeasureText" and "MeasureTextEx" methods:
In the "CentralizeText" method - check if the default font has been loaded, 
If it doesn't - return the zero vector. 
If it does - modify the fontSize, calculate the spacing, and then call the "CentralizeTextEx" method to calculate the desired textPosition.


raysan5:
@nircoe I'm afraid I try to minimize new functions additions unless they are really needed, those functions can be implemented on user size if required.

PR #4624, opened 2024-12-20
[rtext] Fix default font alpha on Big Endian systems

Fancy2209:
Alpha was broken for rtext's default font on Big Endian
Before
imagem
After
imagem

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -247,7 +247,11 @@ extern void LoadFontDefault(void)
                 // we must consider data as little-endian order (alpha + gray)
                 ((unsigned short *)imFont.data)[i + j] = 0xffff;
             }
+            #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
             else ((unsigned short *)imFont.data)[i + j] = 0x00ff;
+            #else
+            else ((unsigned short *)imFont.data)[i + j] = 0xff00;
+            #endif
         }
 
         counter++;
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -247,11 +247,11 @@ extern void LoadFontDefault(void)
                 // we must consider data as little-endian order (alpha + gray)
                 ((unsigned short *)imFont.data)[i + j] = 0xffff;
             }
-            #if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
-            else ((unsigned short *)imFont.data)[i + j] = 0x00ff;
-            #else
-            else ((unsigned short *)imFont.data)[i + j] = 0xff00;
-            #endif
+            else 
+            { 
+                ((unsigned char *)imFont.data)[(i + j)*sizeof(short)] = 0xFF;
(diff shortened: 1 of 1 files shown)


raysan5:
@Fancy2209 It seems this patch addresses a specific use case for a specific hardware arquitecture... won't endianness affect other parts of raylib library? 🤔

Fancy2209:
I tested some examples and they work, but I still have to test loading assets

Also it isn't just PowerPC/PowerPC64  




I'll search for further issues with Big Endian
By the way, it isn't just PowerPC that's big endian, it's just the more common one

Peter0x44:
Big endian is uncommon, but worth supporting. This will break all big endian systems, like powerpc linux, and sparc linux. Not that they are common, but it still matters to some users.

Fancy2209:
> Big endian is uncommon, but worth supporting. This will break all big endian systems, like powerpc linux, and sparc linux. Not that they are common, but it still matters to some users.

You mean this issue affects them too or that this PR breaks them?

Peter0x44:
I mean the issue affects them. I think sparc and powerpc are the most relevant BE systems

raysan5:
@Fancy2209 @Peter0x44 what is the cost of supporting big endian in raylib? It can require many changes and additional code complexity for a very specific set of platforms... if it implies only a couple of checks, it can be considered...

Fancy2209:
I have to test all examples and find issues so it might take a bit
According to an article Peter0x44 code should always be endian independent anyways so it shouldn't be too hard to patch
I'll do a torough testing spree of all tests in a bit, fix the missing tests and report back

raysan5:
@Fancy2209 ok, nice! 👍😄

orcmid:
@raysan5 keep in mind that endian-ness only crops up when one presumes which end bits are in for numerical types with size larger than one byte.  The peculiar case, the one addressed here, is when an int is intended to pack two single-byte values: forming a mini-struct.   I trust that is a very unusual case in raylib and libraries it depends upon.

Peter0x44:
It implies no extra complexity, just a few minor edge cases need some extra care and attention
This case can actually be rewritten not to need the ifdef by writing both bytes of the short separately
Something like:

```c
((unsigned char *)imFont.data)[(i + j)*sizeof(short)] = 0x00;
((unsigned char *)imFont.data)[(i + j)*sizeof(short) + 1] = 0xFF;
```

Fancy2209:
> It implies no extra complexity, just a few minor edge cases need some extra care and attention This case can actually be rewritten not to need the ifdef by writing both bytes of the short separately Something like:
> 
> ```c
> ((unsigned char *)imFont.data)[(i + j)*sizeof(short)] = 0xFF
> ((unsigned char *)imFont.data)[(i + j)*sizeof(short) + 1] = 0x00;
> ```

Done, I still have to test loading Data though

raysan5:
@Fancy2209 please, let me know when ready! did you test some examples to see other possible related point to review?

Peter0x44:
This change only applies to loading the default font and is fine on its own.
There might be other issues with endianness to review but there is no reason not to merge this one.

PR #4628, opened 2024-12-21
Update BINDINGS.md: Raylib-cs now supports 5.5

JupiterRider:
Raylib-cs now supports raylib 5.5.
Source: https://github.com/chrisdill/raylib-cs/pull/279


raysan5:
@JupiterRider Thanks for the update!

PR #4633, opened 2024-12-23
Update BINDINGS.md

james2doyle:
Add some missing bindings for 5.5


raysan5:
@james2doyle thanks for the update! Wow! So many Lua bindings! 😄

james2doyle:
@raysan5 Yeah, and these are just the ones that aren't wrapped in engines!

Issue #4581, opened 2024-12-06
[rmodels] Potential problem with transforming normals during animations


Bigfoot71:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [ ] My code has no errors or misuse of raylib

### Issue description

There seems to be a potential error in the normal transformation process within the `UpdateModelAnimation()` function. I noticed that the shading looked strange when using my lighting shader during an animation, which led me to suspect the normals were being handled incorrectly. After some investigation, I checked the code responsible for transforming the normals in [rmodels.c:2358](https://github.com/raysan5/raylib/blob/master/src/rmodels.c#L2358):

```c
// Normals processing
// NOTE: We use meshes.baseNormals (default normal) to calculate meshes.normals (animated normals)
if (mesh.normals != nullptr) {
    animNormal = (Vector3){ mesh.normals[vCounter], mesh.normals[vCounter + 1], mesh.normals[vCounter + 2] };
    animNormal = Vector3Transform(animNormal, surface.mesh.boneMatrices[boneId]);
    mesh.animNormals[vCounter] += animNormal.x * boneWeight;
    mesh.animNormals[vCounter + 1] += animNormal.y * boneWeight;
    mesh.animNormals[vCounter + 2] += animNormal.z * boneWeight;
}
```

I observed that the normals are being transformed like this:
```c
animNormal = Vector3Transform(animNormal, surface.mesh.boneMatrices[boneId]);
```

But shouldn’t we be transforming the normals using only the rotational part of the matrix, by obtaining the inverse transpose? 
```c
animNormal = Vector3Transform(animNormal, MatrixTranspose(MatrixInvert(surface.mesh.boneMatrices[boneId])));
```

Or maybe I’m missing something else to properly solve this issue, which could explain why the function was written this way.

---

Here’s an example with the current function (look at the robot's hands):
<img src="https://github.com/user-attachments/assets/e95b5dd6-e1b4-4158-a3ec-549a292cab77" width="500" />

And here’s an example with the inverse-transpose transformation applied:
<img src="https://github.com/user-attachments/assets/90899a63-cdaf-40cf-be8c-44dc5b9caaf8" width="500" />

Bigfoot71:
Sorry for the title, I'm tired :fearful:

raysan5:
[...]

PR #4634, opened 2024-12-23
[rmodels] Fix normal transform in `UpdateModelAnimationBones`

Bigfoot71:
As requested here: https://github.com/raysan5/raylib/issues/4581#issuecomment-2560162214

I fixed the function by transforming the normals using only the rotational part of the transformation matrix, obtained via its inverse transpose.

I also simplified the inversion of the quaternion `inRotation`, which was being calculated twice within the same scope.

**Edit**: the examples `models_animation`, `models_bone_socket`, and `models_loading_gltf` were successfully (re)tested.


Bigfoot71:
Sorry for the force push, I just renamed the commit messages because I made a mistake, and they would have poorly represented the history. Nothing else was changed.

raysan5:

PR #4635, opened 2024-12-24
resolved a few segfaults with animation system

JettMonstersGoBoom:
NULL checking added


raysan5:
@JettMonstersGoBoom thanks for the review!

kilometer3604:
Sadly this fix is not included in something like version 5.5.1. That requires maintaining multi branchs though.

PR #4637, opened 2024-12-24
Update RGFW

ColleagueRiley:
Bug fixes:
* When enabling raw input mode, set _lastMousePoint to the screen center (Linux / X11) 
* Lagging scroll value (Linux / X11) 
* Input lag in raw input mode (Windows 10)
* Casting warnings
* Remove unnecessary static keywords

Features: 
* A function rename for consistency
* Change 'keycode' to 'key', 'key' is the scancode for the event
* Add 'keyChar', the character of the keycode of the event. 
* Rename `RGFW_Quote to `RGFW_Apostrophe`


ColleagueRiley:
This is marked as a draft because there was a change in the code that handles how raw mouse input works. I want to test and ensure that this does not make the camera examples on each OS before this is pulled.

raysan5:
@ColleagueRiley wow! big review! merging!

PR #4638, opened 2024-12-25
Update version number for the nim wrapper

planetis-m:


raysan5:
@planetis-m thanks for the update!

PR #4640, opened 2024-12-26
Fix `IsWindowFocused()` on web.

marionauta:
Currently the web platform uses `glfwSetWindowFocusCallback` which seems to do nothing regarding the game canvas' focus. This PR adds emscripten focus/blur callbacks targeting the canvas.

Sample project [focus.zip](https://github.com/user-attachments/files/18254710/focus.zip)

```c
#include <raylib.h>

void gameupdate(void) {
    BeginDrawing();
    ClearBackground(WHITE);
    if (IsWindowFocused()) {
        DrawText("Focused", 10, 10, 50, BLACK);
    } else {
        DrawText("Unfocused", 10, 10, 50, BLACK);
    }
    EndDrawing();
}

int main(void) {
  InitWindow(400, 400, "focus events");
  emscripten_set_main_loop(gameupdate, 0, 1);
  CloseWindow();
  return 0;
}
```

https://github.com/user-attachments/assets/ceec3e5a-d787-49bb-b749-1fbde211691f


raysan5:
@marionauta Thanks for the improvement! As [per the sources](https://github.com/emscripten-core/emscripten/blob/main/src/library_glfw.js#L1706), it seems the function is set but I can't see how it is used...

PR #4644, opened 2024-12-27
[Build] Add platform files to MSVC project for editing

JeffM2501:
This PR adds the platform files to the MSVC project so that they show up for editing (and searching), but do not get built since they are handled via a #include.

This just lets MSVC find the files so that features like search and replace and refactoring tools will work on the files. It changes no functionality.


raysan5:

PR #4646, opened 2024-12-28
[Examples] Add support for normal transformation to the GPU skinning example

JeffM2501:
The GPU skinning example only transformed the vertex data in the shader, it did not show how to also transform the normal data. Even though the normals are not used by the fragment shader, I feel it is important to have an example of how to transform the normals by the skin transforms for people that do want to eventually use lighting on GPU skinned meshes.

This PR transforms the normal and passes it to the fragment shader.


raysan5:
@JeffM2501 nice! thanks for the review!

PR #4649, opened 2024-12-29
Fix leaks in LoadIQM() and LoadModelAnimationsIQM()

peter15914:
Add calls to UnloadFileData() before return in cases of invalid IQM file.


raysan5:
@peter15914 Good catch! Thanks for the review! This is the reason I try to avoid the early-returns in raylib, but this function is not doing it... maybe reviewing it...

PR #4652, opened 2024-12-31
[build] CMake: Don't build examples using audio if audio is disabled

Peter0x44:


raysan5:

PR #4653, opened 2025-01-01
[rcore] Added GetRandomFloat

hexmaster111:
I found myself re-writing this func through many projects.


raysan5:
@hexmaster111 It has been discussed multiple times, this function can be very easely implemented at user side for users requiring it, no need to be in raylib.

PR #4656, opened 2025-01-03
Fix: Typo in core example

rexept:


raysan5:
@rexept thanks for the fix!

PR #4657, opened 2025-01-04
[rcore] [GLFW] Fix camera initial position

### LINKED ISSUE #4654: [rcore] Initial position reported by `GetMouseDelta()` incorrect

### Issue description

I'm having an issue with camera update on MacOS: when the app starts, and I first move the mouse, GetMouseDelta() is very far away. After doing some experiments I figured out that it's assuming that the mouse is initially in the center of the screen. When I start the app having previously moved the mouse to the center of the screen, the behaviour is OK.

I've build from source **core_3d_camera_free**, and the issue is reproduced as described above.

### Environment

MacBook M1 Pro, MacOS Sequoia 15.1.1, OpenGL Version:  4.1 Metal - 89.3.

### Issue Screenshot

Just touched my mouse and got this:
image

### Code Example

**examples/core_3d_camera_free** reproduces the issue.

asdqwe:
Fix #4654.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1017,6 +1017,9 @@ void EnableCursor(void)
 // Disables cursor (lock cursor)
 void DisableCursor(void)
 {
+    // Reset mouse position within the window area before disabling cursor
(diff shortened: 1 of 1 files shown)


raysan5:
@asdqwe thanks for the review!

Issue #4601, opened 2024-12-14
[rshapes] `DrawRectangleRoundedLines()` results not aligned with `DrawRectangleRounded()`


Kaehvaman:
- [*] I tested it on latest raylib version from master branch
 - [*] I checked there is no similar issue already reported
 - [*] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [*] My code has no errors or misuse of raylib

### Issue description

DrawRectangleRoundedLines draws weird unequal corners and is taller and wider than DrawRectangleRounded, which draws identical corners.
DrawRectangleRoundedLinesEx with lineThick != 1 works as expected, so the issue is linked to lineThick of 1 pixel.

### Environment

Windows 10
Visual Studio 2022
MSVC
C language
raylib.lib from v5.5 release
INFO: GL: OpenGL device information:
INFO:     > Renderer: AMD Radeon RX 6750 XT
INFO:     > Version:  3.3.0 Core Profile Context 24.10.1.241007
INFO:     > GLSL:     4.60

### Issue Screenshot
![Screenshot_2](https://github.com/user-attachments/assets/4f397742-b6bb-4592-8c17-b277dd7a4add)

### Code Example
Rectangle roundRect = { 100, 250, 185, 36 };
DrawRectangleRounded(roundRect, 0.5f, 6, BLACK);
DrawRectangleRoundedLines(roundRect, 0.5f, 6, ORANGE);

asdqwe:
Drawing lines has been a hit or miss for a while. It has been reviewed a few times (check #3884, #4075, #4130, #4261), but YMMV.

raysan5:
I think the issue is related to an off-by-one-pixel on lines calculations and I'm afraid that's a GPU-driver dependant issue, very difficult to address. Related to this, we have `DrawRectangleLines()` that still fails after many many changes and tries to solve the off-by-one-pixel issue...

Kaehvaman:
Wow, never expected a driver to break such a simple thing

Kaehvaman:
Updating to 24.12.1 driver didn't fix the issue :(

Bigfoot71:
@raysan5 I get exactly the same result, seemingly pixel-perfect, on:
```console
INFO:     > Vendor:   NVIDIA Corporation
INFO:     > Renderer: NVIDIA GeForce GTX 980/PCIe/SSE2
INFO:     > Version:  3.3.0 NVIDIA 550.120
[...]

PR #4666, opened 2025-01-08
[rshapes] Fix pixel offset issue with line drawing

Bigfoot71:
I propose here a fix for `DrawRectangleRoundedLinesEx` as discussed in this issue: https://github.com/raysan5/raylib/issues/4601.

The method used to calculate the points corresponds to this in pseudocode:
```c
vec2 tl = coord + 0.5
vec2 br = coord + size - 0.5
```

I was only able to test it on the following configurations:

- **NVIDIA GTX 980**  
  - Driver: 550.120  
  - System: Linux Mint 21.3

- **NVIDIA Quadro NVS 160M**  
  - Driver: 340.108  
  - System: Linux Mint 21.3
  
It would be nice to be able to test it on AMD and Intel hardware.

___

The same approach can also be applied to `DrawRectangleLines`.

However, as demonstrated in the issue linked above, this function applies an offset using `zoomFactor`, which causes other alignment issues and leads to incorrect rendering.

To obtain correct results with a scale of (1, 1) for `DrawRectangleLines`, it is currently necessary to eliminate the addition and subtraction of `zoomFactor` and follow the method described above.

Nevertheless, I have conducted several tests to address the offset with "non-identity" scaling, but, strangely, I have not yet achieved any convincing results.

___

**Additional note**: The current application of the `zoomFactor` offset in `DrawRectangleLines` only considers `matView.m0`, but shouldn’t it also take `matView.m5` into account for non-uniform scaling?

___

**I would like to gather opinions on the question of `DrawRectangleLines` before completing this draft.**

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1141,8 +1141,7 @@ void DrawRectangleRoundedLinesEx(Rectangle rec, float roundness, int segments, f
     }
 
     float stepLength = 90.0f/(float)segments;
-    const float outerRadius = radius + lineThick - 0.5f;
-    const float innerRadius = radius - 0.5f;
+    const float outerRadius = radius + lineThick, innerRadius = radius;
 
     /*
     Quick sketch to make sense of all of this,
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1141,7 +1141,8 @@ void DrawRectangleRoundedLinesEx(Rectangle rec, float roundness, int segments, f
     }
 
     float stepLength = 90.0f/(float)segments;
-    const float outerRadius = radius + lineThick, innerRadius = radius;
+    const float outerRadius = radius + lineThick - 0.5f;
+    const float innerRadius = radius - 0.5f;
 
     /*
     Quick sketch to make sense of all of this,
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 thanks for the review! I tested on my laptop, on Windows 10, with Intel integrated and NVIDIA cards, including the Web build. Everything seems to work ok... still, should be picked with a grain of salt: Raspberry Pi, Android, DRM, macOS... too many GPU drivers doing magic...

image

About `DrawRectangleLines()`, I never liked the zoomFactor approach... too much math involved...

raysan5:
Maybe with this nice improvement it's about time to consider rounded controls for `raygui`? 🤔

Bigfoot71:
I realized that I had made small errors, which I have just corrected.

The current version seems to work for any given positions and dimensions, including those with decimal parts.

I also tried many different approaches to solve the issue of line overlap with the underlying shape mentioned by Kaehvaman, but without success.

Here’s an example:  
image

These issues stem from the rendering of the semicircles and seem very difficult to fix.

What ultimately seemed the most logical to me was to start a fix in the loop calling sin/cos for rendering the semicircles, but my attempts were starting to become far too complex to be taken seriously...

The current proposal therefore draws the straight lines on the sides with one less pixel at each end to leave a one-pixel margin for the semicircles and allow them to be drawn with one extra pixel around the shape, avoiding overlap between the lines.

As you mentioned, @raysan5, the lines are indeed drawn with one extra pixel around the shape, so even if this were corrected, we would still see pixels from the shape visible outside the lines, in the current version, at least..

I believe the current version is the most optimal for now.

---

> About `DrawRectangleLines()`, I never liked the zoomFactor approach... too much math involved...

This should still be easily adjustable, if not for this additional offset issue...

I won’t deny that this problem is giving quite a workout to a lot of neurons...

raysan5:
@Bigfoot71 Looks great, really, let me when ready for merge.

Also, feel free to send a PR for `DrawRectangleLines()`, in any case current "solution" could be commented and explained, instead of just being removed, to keep a reference for all the fixes that have been tried along the years...

Bigfoot71:
I just wanted to test it on Android, here’s the setup:  
```console
> Vendor:   Imagination Technologies  
> Renderer: PowerVR Rogue GE8320  
> Version:  OpenGL ES 3.2 build 1.11@5425693  
> GLSL:     OpenGL ES GLSL ES 3.20 build 1.11@5425693  
```

The result is fine:  
image

With transparency, however, on Android, it's interesting to note that two lines seem to overlap at their ends on the semicircle at the top right, and the overlap with the shape is not the same as on PC.

Android:  
image

PC:  
image

**So, here we have indeed factual proof that the behavior can changes depending on the hardware.**

However, in both cases, without transparency, the result seems fine and should be for the vast majority of cases.

___

@raysan5 You can merge if that suits you!

raysan5:
@Bigfoot71 Great improvement! Thanks! 👍🙂

orcmid:
@Bigfoot71 @raysan5 
> I propose here a fix for `DrawRectangleRoundedLinesEx` as discussed in this issue: #4601.
> 
> The method used to calculate the points corresponds to this in pseudocode:
> 
> ```c
> vec2 tl = coord + 0.5
> vec2 br = coord + size - 0.5
> ```
Assuming that *tl* and *br* are **float** values, I don't think the pseudocode is reliable.  It doesn't actually complete the rounding to integral values.

A more explicit one, that does not rely on what the floating-point instructions do and how rounding may vary by configuration would be

```
vec2.tl = floor(coord + 0.5);
vec2.br = ceiling(coord - 0.5);
```
unless you are attempting to have pixel fractions somehow.   Then there's even more to deal with.  (I see remarks about where a pixel's coordinates are presumed to land in the pixel's geometry but that melts my brain.)

Bigfoot71:
> unless you are attempting to have pixel fractions somehow. Then there's even more to deal with. (I see remarks about where a pixel's coordinates are presumed to land in the pixel's geometry but that melts my brain.)

@orcmid Yes, the desired result is to have lines that are as close as possible to the rounded rectangle drawn using the same dimensions and coordinates. And providing coordinates rounded to integers (floor, ceiling, nearest, etc.) will not work here.

The code used for rounded rectangles here is a bit more complex than my pseudocode, and the fix is indeed not complete, as the rendering of quarter circles in lines seems more complex to make it work in all scenarios, and the rest remains fairly simple anyway.

The current state, however, seems satisfactory for most of the given inputs for now, hoping that it will remain so across all platforms.

orcmid:
@Bigfoot71 
> > unless you are attempting to have pixel fractions somehow. Then there's even more to deal with.
> 
> @orcmid Yes, the desired result is to have lines that are as close as possible to the rounded rectangle drawn using the same dimensions and coordinates. And providing coordinates rounded to integers (floor, ceiling, nearest, etc.) will not work here.

In that case, what do the +- 0.5 accomplish?

PR #4669, opened 2025-01-09
[rshapes] Fix pixel offset issue with `DrawRectangleLines`

Bigfoot71:
As requested here (https://github.com/raysan5/raylib/pull/4666#issuecomment-2578790857), here is the fix for `DrawRectangleLines`.

And good news! This fix also handles scale transformations!

There was initially a small error in applying the offsets, which has been corrected.

But most importantly, strangely, the function `rlGetMatrixModelview()` seems to only return the first matrix of the transformation stack, which could be considered just the view matrix if we want to put it that way...

However, by calling `rlGetMatrixTransform()`, we seem to get the actual 'modelView' matrix, which corresponds to all the accumulated transformation matrices on the stack.

Here’s a little example:
```c
BeginDrawing();  // rlLoadIdentity();

rlPushMatrix();
rlScalef(2, 5, 1);

Matrix mv = rlGetMatrixModelview();    //< m0 = 1 | m5 = 1
Matrix t = rlGetMatrixTransform();     //< m0 = 2 | m5 = 5

rlPopMatrix();
```

So, I simply replaced the call to `rlGetMatrixModelview()` with `rlGetMatrixTransform()`, fixed how the offsets are applied, and added consideration for `m5` in non-uniform scales along both axes.

___

Maybe `rlGetMatrixModelview()` is misleading in its name? It's up to you to tell me.

Also, maybe this adjustment could be applied to all other line drawing functions? I’m not sure how relevant that could be, so feel free to share your thoughts on that too.

If that’s the case, for `DrawRectangleRoundedLinesEx()`, it would still require some testing to be sure of what we are doing.

___

Here is an example with a 10x10 square scaled using `rlScalef(5, 2, 1);`:
image

The red is at a transparency of `127`, so we can see the black underneath.

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -807,22 +807,25 @@ void DrawRectangleGradientEx(Rectangle rec, Color topLeft, Color bottomLeft, Col
 // but it solves another issue: https://github.com/raysan5/raylib/issues/3884
 void DrawRectangleLines(int posX, int posY, int width, int height, Color color)
 {
-    Matrix mat = rlGetMatrixModelview();
-    float zoomFactor = 0.5f/mat.m0;
+    Matrix mat = rlGetMatrixTransform();
+    float xOffset = 0.5f/mat.m0;
+    float yOffset = 0.5f/mat.m5;
+
     rlBegin(RL_LINES);
         rlColor4ub(color.r, color.g, color.b, color.a);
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 thanks for the review, looks good, let's hope it keeps working for a long time... 😄

PR #4672, opened 2025-01-09
[rlgl] Remove unused 'format' parameter in rlGenTextureMipmaps

Not-Nik:
Unless I had a major oversight somewhere, that parameter isn't used, so let's leave it out entirely.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -757,7 +757,7 @@ RLAPI void rlUpdateTexture(unsigned int id, int offsetX, int offsetY, int width,
 RLAPI void rlGetGlTextureFormats(int format, unsigned int *glInternalFormat, unsigned int *glFormat, unsigned int *glType); // Get OpenGL internal formats
 RLAPI const char *rlGetPixelFormatName(unsigned int format);              // Get name string for pixel format
 RLAPI void rlUnloadTexture(unsigned int id);                              // Unload texture from GPU memory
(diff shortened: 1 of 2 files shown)


Bigfoot71:
I had already proposed it some time ago and the final opinion is that it was better to keep it
See: https://github.com/raysan5/raylib/pull/4583

> About the format parameter in rlGenTextureMipmaps(), it was kept there for a future implementation for OpenGL 1.1 that required the data size in memory... but implementation was removed at some point in raylib live...

Not-Nik:
So we had a 1.1 implementation, then it was removed and now we're waiting for something to happen so we can add it again? Regardless, unless there's solid plans to add a 1.1 implementation, I reckon it would be OK to just remove the parameter(s) for now and add them again once the implementation is actually there.

asdqwe:
As @Bigfoot71 mentioned, it was debated [before](https://github.com/raysan5/raylib/pull/4583#issuecomment-2532643414). Although `1.1` support was removed (at some point) from `rlGenTextureMipmaps`, `1.1` support itself was [readded](https://github.com/raysan5/raylib/blob/master/CHANGELOG#L140) for `5.5` to support old consoles. So, in my opinion, it's better to leave it as a reference.

raysan5:
As commented, we are keeping this parameter just in case it can be required on a future implementation (OpenGL 1.1, software render, old consoles...)

PR #4675, opened 2025-01-10
Update xcode-frameworks dependency for latest zig

mobiuscog:
When building on Mac via zig, the xcode-frameworks dependency is out of date and therefore fails to build.

This PR updates that dependency to the latest version, which builds correctly.

It should only affect macs. I have only been able to test it on an M1 mac, but I do not think there is any architecture difference in this regard.


raysan5:
@mobiuscog thanks for the review!

PR #4676, opened 2025-01-10
Reviewed shaders_deferred_render

veins1:
Fixed: g-buffer textures binding
Fixed: Clearing screen with white would leak onto g-buffer textures 
Reviewed comments


raysan5:
@veins1 thank you very much for reviewing this nice example!

PR #4678, opened 2025-01-10
Changed things to implement correct usage of RMA (or MRA) textures

Plasmakatana:
RMA (MRA) textures provide Red channel for Roughness, Green channel for Metalness, and Blue channel for Ambient Occlusion.
Made changes to this example to implement this texture correctly.

### DIFF
--- a/examples/shaders/shaders_basic_pbr.c
+++ b/examples/shaders/shaders_basic_pbr.c
@@ -22,7 +22,7 @@
 #if defined(PLATFORM_DESKTOP)
     #define GLSL_VERSION            330
 #else   // PLATFORM_ANDROID, PLATFORM_WEB
-    #define GLSL_VERSION            100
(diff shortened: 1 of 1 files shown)


examples/shaders/shaders_basic_pbr.c
```diff
@@ -22,7 +22,7 @@
 #if defined(PLATFORM_DESKTOP)
     #define GLSL_VERSION            330
 #else   // PLATFORM_ANDROID, PLATFORM_WEB
-    #define GLSL_VERSION            100
+    #define GLSL_VERSION            120
```
raysan5, line 25: raylib uses OpenGL ES 2.0 by default for Web/Android, it requires GLSL 100

raysan5:
@Plasmakatana I checked the shader and it expects to receive the MRA combined in a single texture.

Issue #4673, opened 2025-01-09
[rshapes] `DrawRectangleRounded` draws a regular rectangle if width or height < 1


teatwig:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

Since `Rectangle` accepts floats for `width` and `height` it was a bit surprising to me that you can't draw rounded rectangles that have either be smaller than 1.
See: https://github.com/raysan5/raylib/blob/master/src/rshapes.c#L887

My rectangle is drawn with world coordinates between 0.0 and 1.0, which are then scaled to the full size with Camera2D.

If I remove the check it works without issues, so I assume this is a leftover from some earlier behavior?

### Environment

Linux, raylib 5.5

### Code Example

I'm using the Elixir bindings, but it's really just any rectangle with width or height < 1:

```elixir
        draw_rectangle_rounded(
          %S.Rectangle{
            x: 0.0,
            y: 0.0
            width: 0.5,
            height: 0.5
          },
          0.5,
          3,
          %S.Color{r: 0, g: 0, b: 0, a: 0}
        )
```

asdqwe:
Sometimes those checks are there to prevent artifacts due to the wide range of OpenGL versions raylib supports. Probably requires some testing across `GRAPHICS` versions and `PLATFORM` before removing.

raysan5:
@teatwig Please, could you post some image of the resulting drawn rectangle? Afaik, it's not possible to really draw a sub-pixel element... but I could be wrong...

teatwig:
I tested it again with the latest commit from master (08b089f620c361e68f82782463c9e9a318d15d1f).

This is my program:
```c
int main(void)
{
    const int width = 800;
    const int height = 600;

    InitWindow(width, height, "rounded rect test");

    Camera2D camera = { 0 };
    camera.zoom = height;

    Rectangle rect = { 0.1f, 0.1f, 0.8f, 0.8f };

    while (!WindowShouldClose())
    {
        BeginDrawing();
            ClearBackground(RAYWHITE);
            BeginMode2D(camera);

                DrawRectangleRounded(rect, 0.4f, 3, BLACK);

            EndMode2D();
        EndDrawing();
    }

    CloseWindow();
    return 0;
}
```

Unchanged raylib:
[...]

PR #4683, opened 2025-01-12
[rshapes] Allow `DrawRectangleRounded()` to draw rectangles with a size < 1

teatwig:
As discussed in #4673 this removes the fallback to `DrawRectangle()` if the width or height is < 1.

Rounded rectangles with a size < 1 can already be drawn without issues if the check is removed and this behavior is desirable in cases where rounded rectangles are placed in between fractional world coordinates.


raysan5:
@teatwig thanks for the review!

PR #4689, opened 2025-01-14
Update shaders_mesh_instancing.c

MikiZX1:
adding a missing  'locs[SHADER_LOC_MATRIX_MODEL] = GetShaderLocationAttrib(shader, "instanceTransform");'


raysan5:
@MikiZX1 That data is not needed to be bound, it's already passed to `DrawMeshInstanced()`

PR #4690, opened 2025-01-14
Using canvas id provided by Module for event binding

zet23t:
This change is replacing the hardcoded "#canvas" element references in rcore_web to allow using canvas elements that use different names (which is necessary when using multiple canvas elements on one page).

Also adding a cursor hiding example to test this functionality in browser.


raysan5:
@zet23t Very nice improvement! Thanks for the review! 😄

PR #4691, opened 2025-01-15
Update BINDINGS.md for official V language binding change

mobiuscog:
The bindings from @EmmaTheMartian were moved to official status, so the link is updated.  The name was normalized to be consistent with other 'vendored' names in other languages.


spytheman (approved):
Excellent.

raysan5:
@mobiuscog thanks for the update!

Issue #4695, opened 2025-01-17
[raylib] Buffer Overflow In Initialization


SNAIDY1:
I couldn't find anyone with this problem And i tried to see any solution but nothing.
And this issue is not only in raylib, even in raygui.

`./program`
```
INFO: Initializing raylib 5.5
INFO: Platform backend: DESKTOP (GLFW)
INFO: Supported raylib modules:
INFO: > rcore:..... loaded (mandatory)
INFO: > rlgl:...... loaded (mandatory)
INFO: > rshapes:... loaded (optional)
INFO: > rtextures:. loaded (optional)
INFO: > rtext:..... loaded (optional)
INFO: > rmodels:... loaded (optional)
INFO: > raudio:.... loaded (optional)
INFO: DISPLAY: Device initialized successfully
INFO: > Display size: 1920 x 1080
INFO: > Screen size: 800 x 450
INFO: > Render size: 800 x 450
INFO: > Viewport offsets: 0, 0
INFO: GLAD: OpenGL extensions loaded successfully
INFO: GL: Supported extensions count: 401
INFO: GL: OpenGL device information:
INFO: > Vendor: NVIDIA Corporation
INFO: > Renderer: NVIDIA GeForce RTX 3050/PCIe/SSE2
INFO: > Version: 3.3.0 NVIDIA 565.77
INFO: > GLSL: 3.30 NVIDIA via Cg compiler
INFO: GL: VAO extension detected, VAO functions loaded successfully
INFO: GL: NPOT textures extension detected, full NPOT textures supported
INFO: GL: DXT compressed textures supported
INFO: GL: ETC2/EAC compressed textures supported
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```

And This is The Code :

```
#include "raylib.h"

int main() {
    InitWindow(800, 450, "Minimal Test");
    SetTargetFPS(60);

    while (!WindowShouldClose()) {
        BeginDrawing();
        ClearBackground(RAYWHITE);
        DrawText("Hello, Ray!", 190, 200, 20, LIGHTGRAY);
        EndDrawing();
    }

    CloseWindow();
    return 0;
}
```

asdqwe:
1. Which OS?
2. Did you test with current master branch?
3. Did you set any specific `GRAPHICS`?
4. How are you compiling it?

SNAIDY1:
>     1. Which OS? : Arch Linux
>     2. Did you test with current master branch?:No
[...]

PR #4697, opened 2025-01-17
Fix for buffer overflow

hanaxars:
Fix for #4695 and also removing hardcoded values.


raysan5:
@hanaxars I prefer to avoid the `sizeof()` approach and just use a `MAX_GAMEPAD_NAME_LENGTH` define, line other properties.

PR #4703, opened 2025-01-18
Updted the comment for SetShaderValueTexture to reflect the differenc…

pejorativefox:
Updated the comment for SetShaderValueTexture to reflect the difference between  SetShaderValue using sampler2d and SetShaderValueTexture is the automatic bind of the texture.


JeffM2501 (approved):

raysan5:
@veins1 Agree that it could lead to some confusion, I prefer to keep current description

JeffM2501:
think it should say that it does push the binding.
 there is a confusion between calling SetShaderValueTexture  vs just setting the sampler.
 The comment should say why you want to use one over the other.

raysan5:
@JeffM2501 Ok, it's a minor detail, if it could sound more clear to users, it's fine.

PR #4710, opened 2025-01-19
Support separate dir for screenshots

saxofon:
Make it possible to steer where screenshots are stored.
Default to current behaviour (storing them at same path as the running process) to not break current users of raylib.

If one wants to change the location, we call SetScreenshotPath() before we call TakeScreenshot() like :

```
...
    SetScreenshotPath("/mnt/my-nifty-screenshots-directory");
    TakeScreenshot("my-nifty-screenshot.png");
...
```

SetScreenshotPath is "sticky" so next time we take screenshot it uses same path.


raysan5:
@saxofon I prefer not to add extra functions to maintain in this case, if you need that level of control use the `LoadImageFromScreen()` + `ExportImage()` on user side.

Issue #4712, opened 2025-01-19
[rcore_web] The sleep in WindowShouldClose makes all examples stutter in Firefox


karl-zylinski:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

All the web examples are stuttering in Firefox.

### Environment

Example: https://www.raylib.com/examples/core/loader.html?name=core_custom_frame_control

### Issue Screenshot

Below is a video of how the example runs in Firefox. The example stutters quite a lot. Note how the red circle is "jerking" / "slowing up and  down". It's not smooth.
https://github.com/user-attachments/assets/5ead441a-02e1-4a12-915f-17049ef0ed23

In chrome it is almost smooth, but stutters a bit sometimes as well. You can compare chrome and firefox side-by-side to get a good idea of the difference. It's almost butter smooth in chrome. I also got reports that it is not very smooth in Safari, but I haven't been able to verify that.

### Code Example

It's probably due to this code:
```
bool WindowShouldClose(void)
{
    // Emterpreter-Async required to run sync code
    // https://github.com/emscripten-core/emscripten/wiki/Emterpreter#emterpreter-async-run-synchronous-code
    // By default, this function is never called on a web-ready raylib example because we encapsulate
    // frame code in a UpdateDrawFrame() function, to allow browser manage execution asynchronously
    // but now emscripten allows sync code to be executed in an interpreted way, using emterpreter!
    emscripten_sleep(16);
    return false;
}
```
I understand that "real games" should use `emscripten_set_main_loop` and never call `WindowShuoldClose` (because they are not using a "desktop update loop"). However, the examples nowadays use the "desktop update loop" and rely on this sleep (the comment is outdated I think?). However: It does make them stutter and run poorly in several browsers. Could perhaps the sleep be lowered to 1?

raysan5:
@karl-zylinski Thanks for reporting! Related code has been reviewed. Please, could you confirm examples build with ASYNCIFY not run without stuttering on Firefox and Chrome browsers? 

@CrackedPixel Please, could you also verify that? Thanks!

PR #4713, opened 2025-01-19
[rcore_web] removed hardcoded sleep

CrackedPixel:
conversation from https://github.com/raysan5/raylib/issues/4712

There are 2 types of web-builds

1) Using `asyncify` -- the additional `emscripten_sleep(16)` is removed, EndDrawing still pauses properly
2) Using `set_main_loop` -- the additional `emscripten_sleep(16)` is removed, EndDrawing still adds an additional pause from the browser (i think should be handled in a different PR)


raysan5:

Issue #4717, opened 2025-01-21
[WEB] Makefile: building with RAYLIB_LIBTYPE=SHARED silently creates STATIC library


sleeptightAnsiC:
If you build raylib with src/Makefile, with PLATFORM_WEB and RAYLIB_LIBTYPE=SHARED, it will silently create STATIC library, instead of creating shared library or throwing an error.

If I understand right, shared libraries are not really a thing with Emscripten (it's just a fake static lib imitating `.so` for portability reasons): https://emscripten.org/docs/compiling/Building-Projects.html#faux-dynamic-linking
However, Makefile should not silently create STATIC lib when explicitly asked for SHARED.
It should EITHER: throw an error and exit OR create fake `.so` as asked.

I can easily patch it and propose PR, just tell me which behavior do you want (probably error would be better).

The code that dictates current behavior is here (it ignores `$(RAYLIB_LIBTYPE)` entirely): https://github.com/raysan5/raylib/blob/d48b8afbb5b01241bc85017d168965bdd5d2b6d8/src/Makefile#L664-L670

### Environment
```
~ $ uname -a
Linux MAL200424 6.12.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Fri, 10 Jan 2025 00:39:41 +0000 x86_64 GNU/Linux
~ $ emcc --version
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.72-git (437140d149d9c977ffc8b09dbaf9b0f5a02db190)
```

### Code Example
```
~ $ git clone https://github.com/raysan5/raylib
~ $ cd raylib/src
~/raylib/src $ make TARGET_PLATFORM=PLATFORM_WEB RAYLIB_LIBTYPE=SHARED
[...]
emar rcs ../src/libraylib.web.a rcore.o rshapes.o rtextures.o rtext.o utils.o  rmodels.o raudio.o
raylib library generated (libraylib.web.a)!
```

Peter0x44:
Emscripten does not really support shared libs anyway.

https://emscripten.org/docs/compiling/Building-Projects.html

See "Faux shared libraries" on this page.

I'm wondering, what exactly is your use case?

sleeptightAnsiC:
Din't you read the whole issue @Peter0x44 ?

I linked to the same page as you did, and mentioned the same information:

[...]

PR #4718, opened 2025-01-21
[build] Makefile: throw an error when trying to build SHARED library

sleeptightAnsiC:
[see linked issue for more context]

When asking Makefile to create SHARED library for WEB
```
~/raylib/src $ make TARGET_PLATFORM=PLATFORM_WEB RAYLIB_LIBTYPE=SHARED
```
it would instead silently create STATIC library thus not fulfilling the request as expected

This commit adds an error in this case and stops further execution which fixes this issue.

This is not consistent with Cmake, because Cmake throws the warning and does not stop, but Cmake can easily recover from this case and people probably does not even notice it. However, Makefile is something that you have to handle yourself and you have to recover from any issues so having an error and aborting with exit code 1 is more expected. Otherwise people may spend a lot of time debugging Makefile in order to understand what's even going on.

Fixes: https://github.com/raysan5/raylib/issues/4717

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -665,6 +665,9 @@ raylib: $(OBJS)
 ifeq ($(TARGET_PLATFORM),$(filter $(TARGET_PLATFORM),PLATFORM_WEB PLATFORM_WEB_RGFW))
     # Compile raylib libray for web
     #$(CC) $(OBJS) -r -o $(RAYLIB_RELEASE_PATH)/lib$(RAYLIB_LIB_NAME).bc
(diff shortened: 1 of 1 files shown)


Peter0x44:
I don't think this should be a fatal error, since that just causes inconsistency. A warning alone is fine.

raysan5:
@sleeptightAnsiC I agree with @Peter0x44 that a warning is enough.

raysan5:
@sleeptightAnsiC Merging and reviewing it myself as a warning. Thanks for the improvement.

PR #4720, opened 2025-01-23
[rglfw] fix: suppress warning: "_GNU_SOURCE" redefined

sleeptightAnsiC:
When compiling rglfw.c with recent versions of GCC or Clang following warning appears:

```
$ git clone https://github.com/raysan5/raylib.git
$ cd raylib/src
$ make rglfw.o CC=gcc
gcc  -c rglfw.c -Wall -D_GNU_SOURCE -DPLATFORM_DESKTOP_GLFW -DGRAPHICS_API_OPENGL_33 -Wno-missing-braces -Werror=pointer-arith -fno-strict-aliasing -std=c99 -fPIC -O1 -Werror=implicit-function-declaration -D_GLFW_X11  -I.  -Iexternal/glfw/include
In file included from rglfw.c:99:
external/glfw/src/posix_poll.c:27:9: warning: "_GNU_SOURCE" redefined
   27 | #define _GNU_SOURCE
      |         ^~~~~~~~~~~
<command-line>: note: this is the location of the previous definition
```
This PR suppresses said warning by checking, if _GNU_SOURCE has NOT been defined, and only then defining it. It does not change any behavior, just suppresses the warning.

Raylib already handles similar cases elsewhere in it's source tree (e.g. in [rcore.c](https://github.com/raysan5/raylib/blob/7bfc8e8ca75882de434c601c4294ca1774b69278/src/rcore.c#L159-L161)) so this is nothing new.

### DIFF
--- a/src/external/glfw/src/posix_poll.c
+++ b/src/external/glfw/src/posix_poll.c
@@ -24,7 +24,9 @@
 //
 //========================================================================
 
-#define _GNU_SOURCE
+#if !defined(_GNU_SOURCE)
+    #define _GNU_SOURCE
+#endif
 
 #include "internal.h"
 
--- a/src/external/glfw/src/wl_window.c
+++ b/src/external/glfw/src/wl_window.c
@@ -25,7 +25,9 @@
 //
 //========================================================================
 
-#define _GNU_SOURCE
+#if !defined(_GNU_SOURCE)
+    #define _GNU_SOURCE
+#endif
 
 #include "internal.h"
 


raysan5:
@sleeptightAnsiC I'm afraid this issue belongs to an external library and should be addressed in GLFW repo instead of raylib to avoid being rewritten when updating to next GLFW version.

sleeptightAnsiC:
Hi @raysan5 ,

This issue does not belong to GLFW upstream because Raylib is the one causing it. Some of our build systems define `_GNU_SOURCE` at command line thus triggering this. It simply does not occur on their side and they rejected similar contributions (see https://github.com/glfw/glfw/issues/2133 and https://github.com/glfw/glfw/issues/2076 )

Perhaps, this can be fixed in our build systems? I've noticed only two affected: [src/Makefile](https://github.com/raysan5/raylib/blob/49d37b035f6d764d6bafdc835b5496fded3edcc0/src/Makefile#L328) and [build.zig](https://github.com/raysan5/raylib/blob/49d37b035f6d764d6bafdc835b5496fded3edcc0/build.zig#L84). For comparison, Cmake does not define it and works just fine. So maybe `-D_GNU_SOURCE` is a relic of the past, obsolete in current versions of Raylib? I've noticed that most sources (re)define it anyway when needed.

Hope you're doing well.
Best regards.

raysan5:
@sleeptightAnsiC Still, the change applies to GLFW and it will be overriden on  update. Afair, @Peter0x44 added? that flag on the past for some specific use-case, maybe he can provide more details and if it is still required.

Peter0x44:
I can't remember why I needed it, I think it was for accessing clock functions from C99.
C11 has them as standard so it's not needed there.

The warning is preferable to patching glfw, imo, I hope there's a better solution.
Ideally feature test macros go on the command line if possible.

sleeptightAnsiC:
It's a bit strange that Makefile and zig builds can trigger this but Cmake does not. Is Cmake just not defining it or suppressing with something like `-Wno-builtin-macro-redefined` ? @Peter0x44 

It would be nice if all builds system were consistent here, either all defining `_GNU_SOURCE` at command line or all not doing it. Current difference could cause some cryptic issues in the future where functions depending on specific glibc behavior would become buggy in non-glibc environments.

Anyway, I agree this PR was wrong.

Peter0x44:
@sleeptightAnsiC I think I investigated this before, it's because cmake uses `-std=gnu99` or so by default, which defines _GNU_SOURCE
Setting CMAKE_C_EXTENSIONS to off should make it fail.
I haven't tested this but will do so as soon as I can.

And suppressing the warning on the command line is not a good idea imo.

sleeptightAnsiC:
> And suppressing the warning on the command line is not a good idea imo.

No no, I'm not saying it should be suppressed, just was wondering if it was.

> it's because cmake uses -std=gnu99 or so by default, which defines _GNU_SOURCE

Well, Makefile and build.zig are also using gnu99, but they redefine _GNU_SOURCE unlike Cmake.

Peter0x44:
https://github.com/raysan5/raylib/blob/49d37b035f6d764d6bafdc835b5496fded3edcc0/src/Makefile#L334
The makefile only uses it for web and rgfw web
So that's not exactly true.

I did not check build.zig.

sleeptightAnsiC:
Ahh, I see, sorry. 

zig seems to be using `-std=gnu99` unconditionally and also defining `-D_GNU_SOURCE` at the same time.

https://github.com/raysan5/raylib/blob/49d37b035f6d764d6bafdc835b5496fded3edcc0/build.zig#L81-L88

Peter0x44:
I think this requires a bit more research to figure how to properly solve and also isn't particularly urgent.
Do you mind opening an issue for this to discuss?

I can also do it if needed.

(And slight apology for my comments on our last interaction, figured it was worth a mention)
I'll be more attentive from now.

sleeptightAnsiC:
> Do you mind opening an issue for this to discuss?

sure, I'll open it in a moment

> (And slight apology for my comments on our last interaction, figured it was worth a mention)
> I'll be more attentive from now.

No worries. I tried push my point of view really hard. For this I also apologize.

PR #4723, opened 2025-01-24
[examples] Update `shaders_mesh_instancing`, binding matrix location

MikiZX1:
Due to this missing line of code the example seems is not going to work: shader.locs[SHADER_LOC_MATRIX_MODEL] = GetShaderLocationAttrib(shader, "instanceTransform");

### DIFF
--- a/examples/shaders/shaders_mesh_instancing.c
+++ b/examples/shaders/shaders_mesh_instancing.c
@@ -73,6 +73,7 @@ int main(void)
     Shader shader = LoadShader(TextFormat("resources/shaders/glsl%i/lighting_instancing.vs", GLSL_VERSION),
                                TextFormat("resources/shaders/glsl%i/lighting.fs", GLSL_VERSION));
     // Get shader locations
+    shader.locs[SHADER_LOC_MATRIX_MODEL] = GetShaderLocationAttrib(shader, "instanceTransform");
     shader.locs[SHADER_LOC_MATRIX_MVP] = GetShaderLocation(shader, "mvp");
     shader.locs[SHADER_LOC_VECTOR_VIEW] = GetShaderLocation(shader, "viewPos");
 


MikiZX1:
Hi Ray,
I just saw we did this alread in  'Update shaders_mesh_instancing.c #4689 '
But users on discord keep reporting this line is fixing the issue they have. You maybe mean the 'MVP' is automatically sent to the shader? Anyway, sorry for this double PR, I'm leaving it as is... you will know better what to do.

raysan5:
@MikiZX1 I think there is some confusion here, I tried compiling the example with GCC 14 and VS2022, on Windows 10 and it works as expected.

image

image

Please, could you provide the details on the environment it fails?

raysan5:
Reviewed the issue carefully and as per my understanding that line is not needed, `DrawMeshInstanced()` [binds the instancing matrix data to `SHADER_LOC_VERTEX_INSTANCE_TX`](https://github.com/raysan5/raylib/blob/master/src/rmodels.c#L1736), that is actually the location point used by that function. 

`SHADER_LOC_MATRIX_MODEL` is **not used** by `DrawMeshInstanced()`.

Please, correct me if I'm missing some point.

MikiZX1:
Hello Ray,
You must be right, I just report what I'm told makes the example work.

If you wish to triple check, please have a look here:
https://github.com/raysan5/raylib/blob/139de05e9d68145422b0d625c23a7d7505b8762b/examples/shaders/resources/shaders/glsl330/lighting_instancing.vs#L9C4-L9C8

-'instanceTransform' is the name used in the glsl shader and as far I know raylib does not obtain the glsl location of 'instanceTransform' automatically?
-I haven't checked deeply but I believe the function rlLoadVertexBuffer does not obtain the glsl location of the attribute 'instanceTransform' either. 

I'll dig into source to find this (at this time I don't understand  where is the glsl identifier 'instanceTransform' linked with the corresponding OpenGL storage by raylib)?

Since it is not done by raylib, the example needs to do it itself. (I currently believe this)

I cannot tell why it works for you. At this time it appears like undefined behavior to me.
Again, I am not able to understand what is happening so please forgive me, I'll do my best not to waste your time in the future, especially now during these months during which you are working on the university as well. It is definitively not my place to correct you.

Thanks for all!
Best wishes,
Miki

MikiZX1:
I've just gone through the repo and I do see what you mean. The location of 'instanceTransform' is obtained by rcore.h and rlgl.h
So I don't understand what is going on here. OK to close the issue, I will try understand better what is happening.  Maybe the users reporting the issue were using an older version of raylib. idk
Sorry for the trouble and thanks for all.
Kind regards
Miki

PR #4724, opened 2025-01-24
[rcore] [SDL2] Fix gamepad event handling by adding joystick instance id tracking

### LINKED ISSUE #4722: [rcore] Gamepad no longer detected after removing and adding when using SDL2 backend
### Issue description

I am using SDL2 as my backend for my project, because GLFW doesn't support controller rumble. I've noticed that while the project is running if I unplug my controller and plug it back in it no longer works (sometimes it takes more than one unplug/plug back in). I've tried this also with my 8BitDo Pro 2 controller without it's battery plugged in, so I don't think it's a bluetooth issue. I tested this with the GLFW backend and this problem never occurs. I also tested working with just SDL2 directly and I don't think it's an SDL2 problem.

### Environment

Windows, Desktop. SDL2 2.30.10, Raylib 5.5

### Issue Screenshot
Here I start the .exe, plug in PS4 Controller, unplug, plug it back in (still works), and then when I unplug this time raylib still thinks the controller is plugged in. On plugging in any controller after this no gamepad input works.
https://github.com/user-attachments/assets/2e7dff57-0e9a-4832-9bab-9eca6942ed21

### Code Example

I've attached a zip of my project.
[raylib_c_test.zip](https://github.com/user-attachments/files/18527356/raylib_c_test.zip)

Below is the main file. 

```
#include "sdl2/raylib.h"
#include <math.h>
#include <stdio.h>
#include <stdlib.h>

// Program main entry point
int main(void)
{
    // Initialization
    const int screenWidth = 800;
    const int screenHeight = 450;
    int test = 999;
    char message[20];


    InitWindow(screenWidth, screenHeight, "raylib [core] example - basic window");

    SetTargetFPS(60);               // Set our game to run at 60 frames-per-second

    // Main game loop
    while (true)    // Detect window close button or ESC key
    {
        // Update
        // TODO: Update your variables here
        test = GetGamepadButtonPressed();
        sprintf(message, "%d", test);
        const char* name = GetGamepadName(0);
        const char* name2 = GetGamepadName(1);
        const char* name3 = GetGamepadName(2);
        const char* name4 = GetGamepadName(3);

        // Draw
        BeginDrawing();

            ClearBackground(BLACK);

            DrawText(message, 190, 140, 20, WHITE);

            DrawText("1:", 190 - 20, 200, 20, WHITE);
            DrawText(name, 190, 200, 20, WHITE);
            DrawText("2:", 190 - 20, 230, 20, WHITE);
            DrawText(name2, 190, 230, 20, WHITE);
            DrawText("3:", 190 - 20, 260, 20, WHITE);
            DrawText(name3, 190, 260, 20, WHITE);
            DrawText("4:", 190 - 20, 290, 20, WHITE);
            DrawText(name4, 190, 290, 20, WHITE);

        EndDrawing();
    }

    // De-Initialization

    return 0;
}

```

asdqwe:
Fixes #4722 (both the gamepad connect/disconnect issue and the gamepad button handling).

<details><summary>Explanation of the issue:</summary>
<br>

The SDL2 [doc mentions](https://wiki.libsdl.org/SDL2/SDL_JoyDeviceEvent) `Sint32 which;       /**< The joystick device index for the ADDED event, instance id for the REMOVED event */`, and that's what's happening here. During `SDL_JOYDEVICEADDED` `event.jdevice.which` will hold the `device_index`. But during `SDL_JOYDEVICEREMOVED` `event.jdevice.which` holds the `instance_id`, which are different things.

The problem was that raylib is using the `device_index` to fill/order the `platform.gamepad[]` array. However, it should track the `instance_id` instead (on a different var) to known which gamepad is which.
</details>

This PR can be tested with the `core_input_gamepad.c` [example](https://github.com/raysan5/raylib/blob/master/examples/core/core_input_gamepad.c). It was tested on Linux (Mint 22.0) with SDL2 (2.30.11).

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1693,6 +1693,7 @@ void PollInputEvents(void)
                         CORE.Input.Gamepad.ready[i] = false;
                         memset(CORE.Input.Gamepad.name[i], 0, MAX_GAMEPAD_NAME_LENGTH);
                         platform.gamepadId[i] = -1;
+                        break;
                     }
                 }
             } break;
@@ -1726,8 +1727,15 @@ void PollInputEvents(void)
 
                 if (button >= 0)
                 {
-                    CORE.Input.Gamepad.currentButtonState[event.jbutton.which][button] = 1;
-                    CORE.Input.Gamepad.lastButtonPressed = button;
+                    for (int i = 0; i < MAX_GAMEPADS; i++)
+                    {
+                        if (platform.gamepadId[i] == event.jbutton.which)
+                        {
+                            CORE.Input.Gamepad.currentButtonState[i][button] = 1;
+                            CORE.Input.Gamepad.lastButtonPressed = button;
+                            break;
+                        }
+                    }
                 }
             } break;
             case SDL_CONTROLLERBUTTONUP:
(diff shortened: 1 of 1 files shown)


raysan5:
@asdqwe Thank you very much for the review!

PR #4730, opened 2025-01-26
[rtext] BMFont loading, support RGB and RGBA image formats 

GithubPrankster:
With this PR, if the BMFont file points to images that aren't grayscale, they will now load and be converted. This allows the images to be, for example, .qoi ones which are solely RGBA.

~~I'm curious if failing the format check should throw a log warning, let me know and I'll add that.~~ EDIT: Subsequently added a log warning.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -2227,26 +2227,34 @@ static Font LoadBMFont(const char *fileName)
     {
         imFonts[i] = LoadImage(TextFormat("%s/%s", GetDirectoryPath(fileName), imFileName[i]));
 
-        if (imFonts[i].format == PIXELFORMAT_UNCOMPRESSED_GRAYSCALE)
-        {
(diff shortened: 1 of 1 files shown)


raysan5:
@GithubPrankster I'm seeing multiple points of improvement on this PR. In case of GRAYSCALE only, it can be converted to GRAYSCALE+ALPHA but in case of an image with already provided Alpha, it can use the full RGB color.

Also, the change do not follow raylib coding conventions: parenthesis on conditions, brackets alignment, operations symbols spacing...

GithubPrankster:
> but in case of an image with already provided Alpha, it can use the full RGB color.

So it should just be a copy of the image?

raysan5:
@GithubPrankster I'm afraid I'm not merging this change, the approach of picking up R channel in any case (using stride) is not the desired behaviour.

Issue #3651, opened 2023-12-18
[rcore] Gamepad not working on macOS


NitroPlum:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description
Gamepad input not working. I tried 2 different controllers.

- Xbox One Series X connected via USB
   - IsGamepadAvailable(0) returns false.
- Nintendo Switch Pro Controller connected via USB
  - IsGamepadAvailable(0) returns true.
  - No button input is ever detected

### Environment

MacOSX Ventura on an M2 Mac.

### Issue Screenshot
Can't really show this.

### Code Example
```c
#include "raylib.h"
#include <stdio.h>

int main(void)
{
    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "Gamepads Broken");

    SetTargetFPS(60);               

    while (!WindowShouldClose()) 
    {
        BeginDrawing();

            ClearBackground(RAYWHITE);

            if(IsGamepadAvailable(0)) {
                DrawText("FOUND GAMEPAD TEST", 190, 275, 20, PURPLE);
                
                if(IsGamepadButtonDown(0, GAMEPAD_BUTTON_RIGHT_FACE_DOWN)) {
                    DrawText("BUTTON PRESSED", 190, 350, 20, PURPLE);
                }
            }

        EndDrawing();
    }
    CloseWindow();        

    return 0;
}
```
*Provide minimal reproduction code to test the issue. Please, format the code properly and try to keep it as simple as possible, just focusing on the experienced issue.*

NitroPlum:
After GLFW, I also tried the SDL backend. Is there a way for me to fix the binding issue locally @ubkp ?

NitroPlum:
Oh, another bit of info that may help. I did try in the browser examples on the Raylib site and both controllers work there.

SoloByte:
I have similar problems. I am using the c# bindings of raylib on MacOSX 14.2.1 Sonoma and gamepad input is no longer working. Raylib detects the connected gamepads but does not trigger any button/ trigger/ axis functions.

I have tested Xbox Elite Controller 2 & Xbox Series X Controller. 

- I can connect them fine to my mac
- The input works with the raylib example in the browsers
[...]

PR #4733, opened 2025-01-27
[rcore] [GLFW] [SDL] Add `GetGamepadGUID()

asdqwe:
Ref: #3651

I know that adding new functions should be avoided to keep the library lean, but, in my opinion, this one is necessary.

Currently gamepads are identified by device index (which is just the connected order), joystick id (which is generic/incremental for internal use only) and name (which is not unique and repeatable between controllers).

GUIDs, which are missing, are what actually identify controller for the user-side and the reference to link it to the correct mappings globally. Since controller mappings are outside the library responsability, it's up to the user to provide the necessary mappings, and the user can't do that without checking/confirming the GUID. Without this information, a controller may fail silently due to differences in GUID (e.g.: different firmware versions, etc) and there's no way for the user to learn that (which, incidentally, is one of the issues happening on #3651).

<details><summary>This PR was tested with Linux Mint 22.0 for `PLATFORM_DESKTOP_GLFW` and `PLATFORM_DESKTOP_SDL` with `SDL2` (2.30.11) with the following test case:</summary><br>

``` c
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);

        TraceLog(LOG_INFO, "");

        int joy1 = IsGamepadAvailable(0);
        TraceLog(LOG_INFO, "joy1:%i", joy1);
        if (joy1) {
            TraceLog(LOG_INFO, "name:%s", GetGamepadName(0));
            TraceLog(LOG_INFO, "guid:%s", GetGamepadGUID(0));
        }

        BeginDrawing();
        ClearBackground(RAYWHITE);
        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```
</details>

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -624,6 +624,13 @@ int SetGamepadMappings(const char *mappings)
     return 0;
 }
 
+// Get gamepad GUID
+const char *GetGamepadGUID(int gamepad)
+{
+    TRACELOG(LOG_WARNING, "GetGamepadGUID() not implemented on target platform");
+    return "";
+}
+
 // Set gamepad vibration
 void SetGamepadVibration(int gamepad, float leftMotor, float rightMotor, float duration)
 {
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1084,6 +1084,12 @@ int SetGamepadMappings(const char *mappings)
     return glfwUpdateGamepadMappings(mappings);
 }
 
+// Get gamepad GUID
+const char *GetGamepadGUID(int gamepad)
+{
+    return glfwGetJoystickGUID(gamepad);
+}
+
 // Set gamepad vibration
 void SetGamepadVibration(int gamepad, float leftMotor, float rightMotor, float duration)
 {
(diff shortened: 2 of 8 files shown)


zet23t:
So the purpose of this change is to provide a way to uniquely identify controllers and using a GUID, but most platforms seem to not support getting this information? At least it doesn't seem like JavaScript allows it and for Android I couldn't find information either. I looked up if Unity supports controller GUIDs, but I could not find much about this either.

If the only platform this will work on is SDL, does it really make sense to give a raylib user the false sense that using this function is going to be useful?

asdqwe:
`PLATFORM_ANDROID` and `PLATFORM_WEB` have their own its internal mapping, it's not (and has never been) an issue there.
`PLATFORM_DRM` needs this, but someone else can implement it.
AFAIK Unity ([ref](https://docs.unity3d.com/6000.0/Documentation/Manual/embedded-linux-deploy.html)) and UE ([ref](https://forums.unrealengine.com/t/why-does-unreal-engine-use-sdl2/366360)) use SDL (to some capacity), but have their own handling so their users don't have to.

Without this info, you'll end up having to go native GLFW or native SDL to get basic information that would have solved #3651 a year ago.

asdqwe:
On a second though, maybe raylib doesn't need it. You guys can figure it out.

PR #4734, opened 2025-01-28
Fixing resize and mouse input coordinates

zet23t:
Also replacing hardcoded canvas id references with module variable usages

I tested these changes on my somewhat non-standard blog (using different canvas ids and using a different init logic). Testing it with the window flags example is something I want to do before merging this.

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -185,8 +185,8 @@ void ToggleFullscreen(void)
         else if (CORE.Window.flags & FLAG_BORDERLESS_WINDOWED_MODE) enterFullscreen = true;
         else
         {
-            const int canvasWidth = EM_ASM_INT( { return document.getElementById('canvas').width; }, 0);
-            const int canvasStyleWidth = EM_ASM_INT( { return parseInt(document.getElementById('canvas').style.width); }, 0);
+            const int canvasWidth = EM_ASM_INT( { return Module.canvas.width; }, 0);
+            const int canvasStyleWidth = EM_ASM_INT( { return parseInt(Module.canvas.style.width); }, 0);
             if (canvasStyleWidth > canvasWidth) enterFullscreen = false;
             else enterFullscreen = true;
         }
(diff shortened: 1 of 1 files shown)


asdqwe:
Please try to make individual PRs for each subject. Mixing everything in a single PR makes it harder to test and review.

Also, I don't understand why these mouse position changes are necessary. Tested the mouse position on current master branch (d46ba9a67176f49c3badc031a8f0445b142b5880) with `shell.html` and `minshell.html` on `Chromium 130.0` and `Firefox 131.0` on `Linux Mint 22.0` and the positions are correct (note that mouse position is relative to the canvas not absolute to the screen).

<details><summary>Test case:</summary><br>

``` c
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);
        BeginDrawing();
        ClearBackground(RAYWHITE);

        const int mx = GetMouseX();
        const int my = GetMouseY();
        DrawText(TextFormat("Mouse Pos %i %i", mx, my), 20, 20, 20, BLACK);
        DrawCircle(mx, my, 5.0f, RED);

        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```
</details>

<details><summary>Screenshots:</summary><br>

img1

img2

</details>

zet23t:
All right. I will try your test in my setup to see how it behaves on my machine and in my non standard setup. Eventually there is another hard coded canvas ID somewhere.

zet23t:
The problems start when embedding the canvas into a website and enabling the resizable flag.
```
    SetConfigFlags(FLAG_WINDOW_RESIZABLE);
```

See the gif how the provided test behaves, when setting that flag. I will also add a test for embedded canvas elements, I think there are then issues with fullscreen modes too.

raylib-resizable-bug

asdqwe:
That has nothing to do with mouse position. That's happening because of the manual resize implementation used for `PLATFORM_WEB`. Here's a far simpler fix: #4738

PR #4735, opened 2025-01-29
Replacing hardcoded canvas id references with module variable usages

zet23t:
Should also have the benefit of being slightly faster than obtaining the canvas element via ID.


raysan5:
@zet23t Thanks for the review!

PR #4738, opened 2025-01-29
[rcore] [web] Fix mouse position on resize callback

asdqwe:
Fix for this issue: https://github.com/raysan5/raylib/pull/4734#issuecomment-2622138249

<details><summary>Test case tested on shell.html and minshell.html on Chromium 130.0 and Firefox 131.0 on Linux Mint 22.0 with Emscripten 4.0.1:</summary><br>

``` c
#include "raylib.h"
int main(void) {
    InitWindow(800, 450, "test");
    SetTargetFPS(60);
        BeginDrawing();
        ClearBackground(RAYWHITE);

        if (IsKeyPressed(KEY_ONE)) { ToggleFullscreen(); }
        if (IsKeyPressed(KEY_TWO)) { ToggleBorderlessWindowed(); }

        if (IsKeyPressed(KEY_R)) { SetWindowState(FLAG_WINDOW_RESIZABLE); }

        const int mx = GetMouseX();
        const int my = GetMouseY();
        DrawText(TextFormat("Mouse Pos %i %i", mx, my), 20, 20, 20, BLACK);
        DrawCircle(mx, my, 5.0f, RED);

        EndDrawing();
    }
    CloseWindow();
    return 0;
}
```
</details>

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -162,13 +162,13 @@ bool WindowShouldClose(void)
     // REF: https://emscripten.org/docs/porting/asyncify.html
 
     // WindowShouldClose() is not called on a web-ready raylib application if using emscripten_set_main_loop()
-    // and encapsulating one frame execution on a UpdateDrawFrame() function, 
+    // and encapsulating one frame execution on a UpdateDrawFrame() function,
     // allowing the browser to manage execution asynchronously
 
     // Optionally we can manage the time we give-control-back-to-browser if required,
     // but it seems below line could generate stuttering on some browsers
     emscripten_sleep(12);
-    
+
     return false;
 }
 
@@ -1683,19 +1683,9 @@ static EM_BOOL EmscriptenResizeCallback(int eventType, const EmscriptenUiEvent *
     if (height < (int)CORE.Window.screenMin.height) height = CORE.Window.screenMin.height;
     else if ((height > (int)CORE.Window.screenMax.height) && (CORE.Window.screenMax.height > 0)) height = CORE.Window.screenMax.height;
 
-    emscripten_set_canvas_element_size(GetCanvasId(), width, height);
-
-    SetupViewport(width, height); // Reset viewport and projection matrix for new size
-
-    CORE.Window.currentFbo.width = width;
-    CORE.Window.currentFbo.height = height;
-    CORE.Window.resizedLastFrame = true;
-
     if (IsWindowFullscreen()) return 1;
 
-    // Set current screen size
-    CORE.Window.screen.width = width;
-    CORE.Window.screen.height = height;
+    glfwSetWindowSize(platform.handle, width, height);
 
     // NOTE: Postprocessing texture is not scaled to new size
 


zet23t:
I can confirm that this fixes the mouse cursor issue, but the resizing is still breaking the web page: When the `SetWindowState(FLAG_WINDOW_RESIZABLE);` is set and the canvas is embedded in a web page, upon return from fullscreen mode, the canvas element has a different size and overlaps other elements. I don't know why it behaves this way and why the changes I did (https://github.com/raysan5/raylib/pull/4734) are handling this switch correctly.

zet23t:
Here are the files I use to test this pull request: 

https://gist.github.com/zet23t/9b98d9f22a20e5c056a374dd4413278b

asdqwe:
> upon return from fullscreen mode

Which fullscreen mode? `ToggleFullscreen()` or `ToggleBorderlessWindowed()`?

zet23t:
Both modes break the original canvas size (or don't restore it properly).

zet23t:
No. My test case is slightly different as it has a canvas that is integrated on a website. Here is the start situation (resizable true)
grafik

And here is what happens when the browser window is resized:
grafik

It is not even necessary to go to full screen even. The reason why the testcases on your machine don't show problems is, that upon window size change (either by triggered by fullscreen toggle or window size change), the canvas uses the entire browser's client area. If the canvas is using this size anyway, there's no problem. But when the canvas is part of a website, it breaks due to assigning the canvas the current client size area as new canvas size.

asdqwe:
Ok, I see. The problem then is that, at the moment, raylib wasn't (really) meant to be deployed like that <sup>1</sup>. Usually you'd encase the exported `shell.html`/`index.html` on some `<iframe>` (e.g.: like done on [itch.io](https://itch.io/), etc). Then set the `<iframe>` `width/height` to the value you need (even a percentage, or proportion). Fullscreen/BorderlessWindowed also wasn't designed for that use case (at all).

Supporting your use case would require a bigger refactor to `PLATFORM_WEB` (not only resize, but also maximize, fullscreen, borderlesswindowed, etc) and testing it on both ways of deployment (directly and inside an `<iframe>`, to not break it for everyone deploying raylib on sites like [itch.io](https://itch.io/)).

I'm updating the title since such refactor is out of scope of this PR and this fix is still needed for the current master branch.

<sup>1</sup> Canvas is not a window, but, for convenience, it was manually abstracted like one for some things (e.g.: resize and maximize).

zet23t:
Well, the changes I did do fix these issues altogether, even though the handling is a bit awkward.

asdqwe:
As I mentioned, you would still need to refactor a lot to completely support this use case. Also keep in mind that raylib has to work with both `minshell.html` and `shell.html` (I believe @raysan5 uses it for his tools or some deployment). But feel free to PR that if you think it's a better solution.

raysan5:
@asdqwe @zet23t Thank you very much for working on this issue! Please gave me some days to review it properly. I'll try to setup some quick examples to illustrate my use-case.

The specific use case I found the issue with was on some of my tools that allow a RESIZABLE window, the canvas covers all the browser window and the tools gets resized responsively with the canvas. Here it is an example: https://www.raylibtech.com/rtexviewer/

Note the main toolbar gets resized with the browser window. This release works as expected because it was build with a previous `library_glfw.js` version not including the "HighDPI support improvement". Currently in the mouse is completely off.

On the other side, there are the examples that use CSS to set `width: 100%`, in that case mouse is scaled accordingly to be reported considering canvas scaling.

Despite it worked as expected before the HighDPI change, I fear new implementation could not work on both cases... or maybe I'm not properly understanding all the parts involved... I need to review it more carefully.

zet23t:
Honestly, when I tried to fix the issues I've had, it crossed more than once my mind to bypass GLFW bindings and bind directly to JS using emscripten. This 2nd abstraction layer in between makes this quite a black box what is going on. I don't know though how much work that would be. In the approach I chose, I decided to ignore the GLFW mouse input feedback and use emscripten's callback instead. But yeah, this is quite a mess, unfortunately.

raysan5:
> Honestly, when I tried to fix the issues I've had, it crossed more than once my mind to bypass GLFW bindings and bind directly to JS using emscripten. This 2nd abstraction layer in between makes this quite a black box what is going on.

Completely agree! Also considered it several times! Latest update of `library_glfw.js` gave me lot of dissatisfaction...

Afaik, that's what @ColleagueRiley did for `RGFW.h`, just using a custom implementation directly.

ColleagueRiley:
@raysan5 Iirc platform_web already implements a lot of the backend on its own. It would still require some work and would force you to maintain an extra layer of Raylib, but it could be worth it. 

Also for RGFW, I'm thinking about dropping Emscripten callbacks in favor of directly calling the JS callbacks just to be less Emscripten dependent. It would also be nice if I could drop the Emscripten from RGFW entirely, but I don't think that's possible.

asdqwe:
Yeah, that's a better option.

PR #4739, opened 2025-01-29
[rmodels] Separate GLTF roughness and metallic channels on load

Not-Nik:
Create separate textures for roughness and metallic values when loading GLTF/GLB files. No one interjected, so I just did the implementation of my proposal.

This doesn't break any examples, because the only GLTF+PBR example we have, loads its own texture files manually. I did test this with my own setup, where this creates the same behaviour as if doing the separation in an image editing tool before running, and then loading the textures, after loading the image.

For more info see #4715

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5347,7 +5347,33 @@ static Model LoadGLTF(const char *fileName)
                     Image imMetallicRoughness = LoadImageFromCgltfImage(data->materials[i].pbr_metallic_roughness.metallic_roughness_texture.texture->image, texPath);
                     if (imMetallicRoughness.data != NULL)
                     {
-                        model.materials[j].maps[MATERIAL_MAP_ROUGHNESS].texture = LoadTextureFromImage(imMetallicRoughness);
+                        Image imMetallic, imRoughness;
+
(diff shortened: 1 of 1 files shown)


raysan5:
@Not-Nik Nice redesign, I think it's more consistent with raylib this way and will avoid confusions, despite the performance penalty involved...

Just a side note, format is not following raylib code conventions, I'm just reviewing it myself now.

raysan5:
@Not-Nik While reviewing the function I got a doubt about the channels used by `imMetallicRoughness`, why are they `G` and `B`? Why not `RG`? It's always that way by GLTF specs? Also, I think `GetImageColor()` call can be avoided...

Not-Nik:
> Just a side note, format is not following raylib code conventions, I'm just reviewing it myself now.

Ah, dammit. I always try to adhere, but my formatter is configured differently, and it messes is up sometimes.

> It's always that way by GLTF specs?

Yes, the spec defines the values to be in these channels. I use the `GetImageColor()` in case of different formats (8/16/32 bit colors, un/compressed).

https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_material_pbrmetallicroughness_metallicroughnesstexture

PR #4741, opened 2025-01-31
[rtext] fix: misuse of cast in GetCodepointCount

sleeptightAnsiC:
I was really wondering what is going on here :D I believe this code tried initially to out-cast 'const' specifier but this is not needed here at all. Currently, it just confuses whoever reads this so I changed it. The old code would also trigger -Wcast-qual warning on some compilers.

Tested on Linux with latest branch.


raysan5:
@sleeptightAnsiC Thanks for the review! I don't know why it was coded that way... 🤔

PR #4742, opened 2025-01-31
[platforms/glfw] fix: GetGamepadButtonPressed() can return analog trigger button values

whaleymar:
Currently GetGamepadButtonPressed() cannot return `GAMEPAD_BUTTON_LEFT_TRIGGER_2` or `GAMEPAD_BUTTON_RIGHT_TRIGGER_2` when using GLFW for input because GLFW treats them as analog values. When their analog value exceeds 0.1, Raylib sets the binary button state to 1 but doesn't update `lastButtonPressed`. This is probably an oversight, so I've fixed that in this PR.


raysan5:
@whaleymar Good catch! Thanks for the review!

PR #4744, opened 2025-01-31
[raudio] AddAudioStreamProcessor: enable closures and similar user context use cases by providing a "context pointer"

goto40:
I played with your wonderful library after I listened to a recent podcast with an interview with you :-).

I encountered a problem while trying to attach a callback in rust as audio stream processor. While it was possible to add it via free functions I had some problems when trying to attach a closure.

It would be much easier when the audio stream processor callback signature would include a context pointer (`void* user_data`) like in other APIs like pthread. With such a context pointer a context data structure can easily be attached to a callback. This helps disambiguating the callbacks from each other.
 
I tried to propose a non-breaking extension to allow in addition to the existing callbacks to include a new callback with a context pointer.

I would be happy to discuss this proposal - and hopefully to find a solution to my problem.


raysan5:
@goto40 I'm afraid it was a design decision to avoid a pointer to user data to keep the API as much simple as possible for new users, that's one of the main goals of raylib. Actually, afair, this possibility was also proposed in the past and not accepted; users with advance audio needs could implement this specific function for the specific use case.

goto40:
I understand. What a pity. This makes it hard for languages like C++ and Rust to easily wrap the C-call with safe wrappers. But I will find a solution for my use case. Thank you very much for the explanation.

PR #4748, opened 2025-02-01
[build][cmake] Set PLATFORM to Web by default when configuring web builds with emcmake

gilzoide:
This PR improves the CMake web build by setting the `PLATFORM` to `Web` by default when configuring builds using `emcmake cmake`.
This solution is the same as 9821725c6bafeaf4153ab8604e692dfe13cbff09 wanted to achieve, but for it to work the value must be set before `enum_option(PLATFORM ...)` is called, that's why it didn't work before.

My only doubt is if it's best to just set the default value like I did, or if it makes sense to force PLATFORM to Web and only call `enum_option(PLATFORM ...)` in the `else()` branch, since Emscripten cannot build raylib with any other value of PLATFORM set.

### DIFF
--- a/CMakeOptions.txt
+++ b/CMakeOptions.txt
@@ -2,6 +2,10 @@
 include(CMakeDependentOption)
 include(EnumOption)
 
+if(EMSCRIPTEN)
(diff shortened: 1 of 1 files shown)


raysan5:
@gilzoide Thanks for the review!

PR #4753, opened 2025-02-06
doc: audio stream processor, number of channels

goto40:
Thank you for your nice library.

I stumbled over the usage of audio processors. Initially I processed a mono-audio file and was surprised to receive stereo data. After analyzing the code I realized that it always provides two channels as float (see #4743). 

I added additional documentation. Could you have a critical look and help to clarify the usage?
Thank you


raysan5:
@goto40 Oh, good catch, thanks for the review

PR #4755, opened 2025-02-07
[Examples] Added single file option to examples

CrackedPixel:


CrackedPixel:
@raysan5 what do you think of this? it will make single .html files that can be viewed without a web server

examples/Makefile.Web
```diff
@@ -59,6 +59,8 @@ USE_WAYLAND_DISPLAY   ?= FALSE
 BUILD_WEB_ASYNCIFY    ?= TRUE
 BUILD_WEB_SHELL       ?= $(RAYLIB_PATH)/src/shell.html
 BUILD_WEB_HEAP_SIZE   ?= 134217728
+BUILD_WEB_SINGLE_FILE ?= TRUE
+BUILD_WEB_FILE_MODE   ?= --preload-file
```
CrackedPixel, line 63: a single file requires `--embed-file` instead

examples/Makefile.Web
```diff
@@ -59,6 +59,8 @@ USE_WAYLAND_DISPLAY   ?= FALSE
 BUILD_WEB_ASYNCIFY    ?= TRUE
 BUILD_WEB_SHELL       ?= $(RAYLIB_PATH)/src/shell.html
 BUILD_WEB_HEAP_SIZE   ?= 134217728
+BUILD_WEB_SINGLE_FILE ?= TRUE
```
CrackedPixel, line 62: default to false?

raysan5:
@CrackedPixel Definitely single-html is an interesting option... but it also seems a non-standard approach. I prefer to not expose that option for the users for now... It seems very user-specific for certain situations...

Still, I think some note/comment could be added somewhere, maybe on emscripten parameters details section, to explain users this is a possibility.

PR #4758, opened 2025-02-10
[rcore] `ScanDirectoryFiles()` and `ScanDirectoryFilesEx()` creates dependency on module `rtext`

mannikim:
the functions `ScanDirectoryFiles()` and `ScanDirectoryFilesEx()` both use the function `TextFindIndex()` which is defined in rtext.c, and requires the flags `SUPPORT_MODULE_RTEXT` and `SUPPORT_TEXT_MANIPULATION` to be defined

this pull request introduces #if directives to prevent the use of TextFindIndex by disabling filtering directories in these functions

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2243,6 +2243,7 @@ const char *GetApplicationDirectory(void)
 // NOTE: Base path is prepended to the scanned filepaths
 // WARNING: Directory is scanned twice, first time to get files count
 // No recursive scanning is done!
+// WARNING: filtering directories requires module rtext and SUPPORT_TEXT_MANIPULATION
 FilePathList LoadDirectoryFiles(const char *dirPath)
 {
     FilePathList files = { 0 };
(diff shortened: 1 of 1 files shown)


sleeptightAnsiC:
Shouldn't this just use a LibC equivalent of `TextFindIndex()` which would be slightly modified `strstr()` ?

The rest of ScanDirectoryFiles/Recursively() already uses LibC calls from `<string.h>` so someone made sure to decouple it from rtext and probably forgot about this one single function.

Adding `#ifdef` around single branch would only silently break those as `files->count` will not be properly incremented.

mannikim:
> Shouldn't this just use a LibC equivalent of `TextFindIndex()` which would be slightly modified `strstr()` ?

that would be a good alternative, since it would retain the directory filtering feature. i was afraid of changing the code too much

> Adding `#if` around single branch would only silently break those as `files->count` will not be properly incremented.

the pointer only increments if the filter string contains `DIR`. since the idea was to deactivate this feature if SUPPORT_TEXT_MANIPULATION is off, it is working as intended, as you wouldn't be able to compile without it anyways
also, i couldn't find any use of this feature through the rest of the library, so there is nothing breaking internally

raysan5:
@sleeptightAnsiC Considering rcore is already dependant on `string.h`, decoupling it from `rtext` would be a better addition than this PR.

mannikim:
@raysan5 i made a new pull request fixing it, see #4764

@sleeptightAnsiC thanks for the opportunity :)

closing it now

PR #4759, opened 2025-02-12
[rtext] `TextSubtext()` fixes

veins1:
Fix buffer write overflow
Fix reading past the end of text


raysan5:

PR #4761, opened 2025-02-12
[rmodels] fix: Add Stride Argument for rlSetVertexAttribute calls inside UploadMesh

cellomath:
The vertex attribute pointer does not increment for meshes that have been uploaded with this function.
The vertex pointer should increment by default by the size of the vertex data.


raysan5:
@cellomath As per OpenGL specs:

 > If stride is 0, the generic vertex attributes are understood to be tightly packed in the array. The initial value is 0.

I think that's correct.

PR #4763, opened 2025-02-13
[rmodels] Improve default normal and tangent values (UploadMesh)

Bigfoot71:
### Description: 
This PR updates the default normal and tangent values used when a mesh lacks corresponding buffers.  

#### Changes:
- **Default normal:** Changed from `(1, 1, 1)` to **`(0, 0, 1)`** to ensure a properly normalized default normal, aligned with the standard convention of pointing out of the surface.  
- **Default tangent:** Changed from `(0, 0, 0, 0)` to **`(1, 0, 0, 1)`** to provide a valid and reliable fallback for constructing the bitangent in shaders.  

#### Benefits:
- Avoids potential problems when calculating the bitangent from a tangent that has a zero value.
- You can achieve **optional normal mapping** without requiring extra checks/shaders or conditional branches in shaders.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1307,7 +1307,7 @@ void UploadMesh(Mesh *mesh, bool dynamic)
     {
         // Default vertex attribute: normal
         // WARNING: Default value provided to shader if location available
-        float value[3] = { 1.0f, 1.0f, 1.0f };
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Looks good! Thanks for the improvement!

PR #4764, opened 2025-02-13
[rcore] fix: use `strstr()` instead of `TextFindIndex()` to decouple module from `rtext`

mannikim:
this fixes issue addressed at #4758, but preserving functionality. 
in short, module `rcore` cannot be compiled independently from module `rtext` since it uses the function `TextFindIndex()`, implemented at `rtext`.


raysan5:
@mannikim Great improvement! Thank you very much!

PR #4768, opened 2025-02-15
[rlgl] Added rlWriteTexturePixels to support work with externally allocated buffers

KhanKotyano:
Added rlWriteTexturePixels, whitch is rlReadTexturePixels implementation without internal memory allocations, to support dynamic conversion from the Texture to the Image, without a need for free() on previous memory buffer


raysan5:
@KhanKotyano I prefer to avoid that function, afair, there is no function in raylib/rlgl that fills a user provided buffer.

PR #4769, opened 2025-02-15
[rcore_android] Implementation for `GetWindowScaleDPI` and monitor physical dimensions

Bigfoot71:
Implementation for the missing functions:
- `GetWindowScaleDPI()`
- `GetMonitorPhysicalWidth()`
- `GetMonitorPhysicalHeight()`

Only a small issue with `GetMonitorPhysicalWidth` and `GetMonitorPhysicalHeight` on some devices, where the value seems to be underestimated.

I tested it on a Samsung Galaxy with satisfactory results, but on a Xiaomi device (Redmi A9), the result was underestimated by about ~0.5 inches...

So I add a note about it.

This PR follows a request on raymob: https://github.com/Bigfoot71/raymob/issues/35

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -460,17 +460,21 @@ int GetMonitorHeight(int monitor)
 }
 
 // Get selected monitor physical width in millimetres
+// NOTE: It seems to return a slightly underestimated value on some devices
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Thanks for the implementations!

PR #4772, opened 2025-02-16
Fix typo in rshapes.c

vict-Yang:
Fix typo: "betweeen" → "between"


raysan5:
@vict-Yang Good catch! Thanks for the fix! 👍🙂

PR #4774, opened 2025-02-17
[raylib.h] Added space after comma

Brian-ED:
args for ImageResizeNN were missing a space in the arguments after a comma, i just added one.


raysan5:
@Brian-ED Good catch! Thanks for the review! I love those small improvements that make the library so consistent!

PR #4775, opened 2025-02-17
[rlgl] Implementing `rlReloadVertexBuffer()` and `rlReloadVertexBufferElement()`

mikeemm:
Currently, a VBO/EBO cannot be resized after being loaded with `rlLoadVertexBuffer()` or its element counterpart. Using `rlUpdateVertexBuffer()` only works if your new vertex data is as big or smaller than what was initially allocated; not if it is bigger. The buffer's _usage_ is also not modifiable this way.

At the moment, the raylib-only solution is to use `rlUnloadVertexBuffer` on the old buffer and load it anew with `rlLoadVertexBuffer()`, while also recreating any VAOs using these VBO/EBOs. Implementing a solution within your own code doesn't seem to be possible out of the box considering that the OpenGL functions it requires don't seem to be exposed by Raylib.

With these changes we get `rlReloadVertexBuffer()` and `rlReloadVertexBufferElement()`, which reload an existing vertex buffer without requiring new bindings. I believe this is a really simple addition that can be really helpful in cases when you may want to modify parts of an existing mesh without destroying and recreating it completely, or when you're handling vertex buffer objects on your own without using Raylib's structures directly.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -738,6 +738,8 @@ RLAPI void rlSetTexture(unsigned int id);               // Set current texture f
 RLAPI unsigned int rlLoadVertexArray(void);             // Load vertex array (vao) if supported
 RLAPI unsigned int rlLoadVertexBuffer(const void *buffer, int size, bool dynamic); // Load a vertex buffer object
 RLAPI unsigned int rlLoadVertexBufferElement(const void *buffer, int size, bool dynamic); // Load vertex buffer elements object
+RLAPI void rlReloadVertexBuffer(unsigned int id, const void *buffer, int size, bool dynamic); // Reload an existing vertex buffer object
+RLAPI void rlReloadVertexBufferElement(unsigned int id, const void *buffer, int size, bool dynamic); // Reload an existing vertex buffer elements object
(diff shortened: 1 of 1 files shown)


raysan5:
@mikeemm I'm afraid I prefer to avoid those extra functions, I think not many users need them and user with that specific level of requirement can do the OpenGL calls directly.

mikeemm:
@raysan5 That's fair enough. Though, may I ask what is the recommended way to access these functions outside of Raylib's code? I saw the suggestion somewhere to use `#include "external/glfw/include/GLFW/glfw3.h"`, and that works for a bunch of functions, but not all of them; for example, not the ones involved in `rlLoadVertexBuffer()`. I'm not sure how to use them without modifying Raylib directly.

casperbear:
@mikeemm Chances are you will find that you want to build your own RayLib build with small changes here and there anyways.

mikeemm:
@casperbear That's what I'm currently doing, however I would prefer staying up to date with the original repo if I could just implement small functions like these on my own code without modifying Raylib itself.

The option to use OpenGL directly keeps being brought up in cases where the proposed functionality is deemed too complex for Raylib, but functions like these don't seem to be exposed by raylib at least in the ways I've tried to add them, so I was wondering if there was a specific way that we are meant to enable access to OpenGL functions for our personal use alongside Raylib?

Issue #4776, opened 2025-02-18
[rcore] GLFW does not use Raylib-provided RL_*ALLOC macros for memory allocations


sleeptightAnsiC:
Raylib has a support for providing custom heap memory allocators by defining `RL_MALLOC` / `RL_CALLOC` / `RL_REALLOC` / `RL_FREE` macros. GLFW supports this too, but with function pointers passed into [glfwInitAllocator()](https://www.glfw.org/docs/latest/intro_guide.html#init_a,llocator) call. Currently, raylib does not use glfwInitAllocator anywhere, and so GLFW always uses default LibC allocators, which is undesired.

I noticed that there is some code in `src/platforms/rcore_desktop_glfw.c` that was meant to support this, but it is currently commented-out with a TODO note:
https://github.com/raysan5/raylib/blob/11090ab6cbc9a58dc640c2d712d1d467457c2e98/src/platforms/rcore_desktop_glfw.c#L1290-L1305

Related to https://github.com/raysan5/raylib/issues/4751

[I'll make a PR for this in a moment]

Issue #4751, opened 2025-02-05
[*] Concerns about safety, errors handling and inconsistencies


raysan5:
This issue is related to a [reddit post](https://www.reddit.com/r/raylib/comments/1ieqejd/some_concerns_and_general_question_about_error/) about errors handling and inconsistencies.

raylib original approach was a hard focused on **simplicity**, trying to avoid any extra-code that could pollute the functions, leaving the security data checks to the users and expecting the data to be provided to most functions to be valid data.

But raylib has been +11 years in development and in that time some data validation checks have been added here and there resulting in some inconsistencies... and polluting the functions code with some things I originally tried to avoid (i.e early-returns).

Still, I tried to follow some principles in the library:

 - Users are the owners of data and valid data should be provided to raylib functions.
 - raylib does not use `assert()` or `exit()`.
 - In case of issues with data, priority is logging a warning message and trying to return gracefully (with valid default data or at least something to be validated at user side - i.e. zero-initialized struct)
 - raylib does not (usually) use error return codes. (But lately some were added to file-system functions)
 - raylib text functions always expect `\0` terminated strings, it's up to the user to make sure that happens.
 - raylib does not validate memory allocations returned pointers, the chances for a failure on memory allocation are extremely low but, in any case, users can define their own allocator macros.
 - In raylib I tried to minimize early returns from functions
 - In case some function return an error-code, I prefer to accumulate it into a `result` variable and use it for checking.

This issue is open to analyze if there could be some of those tentative rules not applied in some functions or if there are some inconsistencies.

orcmid:
One probably-unhelpful comment: 

When "user" is used above, the developer of a raylib-based application is meant.  The problem is when that application is distributed to end-users and the application fails in a place where log messages, even if actually produced, are invisible and unknown to the casual user.  It is for this situation that defensive programming is valuable.  

Also, raylib-internal operations may also need to be defensively programmed, as indicated in some places already, where a problem is not protectable above the API, it being related to under-the-covers failures.   

That could also apply to unchecked inputs from above that violate API pre-conditions.  Perhaps the pre-conditions should be carefully explicit.

The worry should be about defective distributed executables.  That's without considering the creation or corruption of an app for malicious purposes.

sleeptightAnsiC:
EDIT: I'm hiding this. I's a huge and scary wall of text that I'm pretty sure is partially outdated. Some of the information is repeated in later messages.

(I'm the author of the reddit post)

[...]

PR #4777, opened 2025-02-18
[rcore_desktop_glfw.c] fix: make sure that GLFW uses RL_*ALLOC macros

sleeptightAnsiC:
Raylib allows for providing custom allocators via macros. GLFW supports this too, but via function pointers. Let's make sure that GLFW uses those Raylib macros, by wrapping them in function calls and setting them up inside of InitPlatform(). This is possible because of glfwInitAllocator() and GLFWallocator. (see linked issue for more context)

I hope that code notes speak for itself but let me know what do you think, and if this is the correct approach or not.
 
Tested on x86_64 Linux with GNU Make and Cmake, with GCC and TCC, by compiling and running some examples.

Reference: https://www.glfw.org/docs/latest/intro_guide.html#init_allocator
Fixes: https://github.com/raysan5/raylib/issues/4776
Relates-to: https://github.com/raysan5/raylib/issues/4751

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -86,6 +86,8 @@
     #include "GLFW/glfw3native.h"       // Required for: glfwGetCocoaWindow()
 #endif
 
+#include <stddef.h>  // Required for: size_t
+
 //----------------------------------------------------------------------------------
 // Types and Structures Definition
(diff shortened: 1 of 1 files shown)


raysan5:
@sleeptightAnsiC thanks for this review! That TODO has been there for a long time! 👍😄

PR #4779, opened 2025-02-20
[raudio] Fixed memory leak on early-return of WaveFormat func

deckarep:
In the case of a failure within miniaudio on the function: `ma_convert_frames`, the dynamic memory allocated for the `data` variable will leak on the early return.


raysan5:
@deckarep Good catch! Thanks for the improvement!

deckarep:
Thanks @raysan5! But I realized you have a standalone rep for raudio as well. Does this PR actually need to be applied there? If so I can do that.

PR #4783, opened 2025-02-21
Update RGFW and ensure it's consistent with the GLFW backend

ColleagueRiley:
- implements a lot of missing features 
- ensures features are implemented correctly:

eg. exclusive fullscreen creates an exclusive fullscreen, borderless creates non-exclusive fullscreen (not just a regular no-border window)

- Update RGFW to 1.6-dev

- link IOKit on MacOS, this is required for gamepad support.


raysan5:
@ColleagueRiley Definitely a huge update! Thanks for the multiple improvements! I'm considering switching to RGFW as default backend on next release! Great work!

PR #4786, opened 2025-02-22
[PROTOTYPE WIP] implemented avif support for testing

CrackedPixel:
@raysan5 what do you think of the concept? Of course it will need more testing etc. Maybe its out of scope or conflicting licenses etc and could be used for a reference for someone else only.

To use: 
* clone [libavif](https://github.com/AOMediaCodec/libavif) into `/src/external` (copying SDL pattern)
* rebuild with `SUPPORT_FILEFORMAT_AVIF` and `LIBAVIF_ENABLE` (copying SDL/etc format)

Everything works the same:
```cpp
Texture avifTexture = LoadTexture("texture.avif");
```

avif is much smaller than png with high quality. Loading speeds vary drastically from the image, some 3x faster some 2x slower

avif uses BSD license

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -121,6 +121,15 @@ SDL_INCLUDE_PATH      ?= $(RAYLIB_SRC_PATH)/external/SDL2/include
 SDL_LIBRARY_PATH      ?= $(RAYLIB_SRC_PATH)/external/SDL2/lib
 SDL_LIBRARIES         ?= -lSDL2 -lSDL2main
 
+# AVIF support: It requires libavif to be provided externally
+# WARNING: Library is not included in raylib, it MUST be configured by users
+LIBAVIF_ENABLE          ?= TRUE
+LIBAVIF_ROOT_PATH       ?= $(RAYLIB_SRC_PATH)/external/libavif
+LIBAVIF_INCLUDE_PATH    ?= $(LIBAVIF_ROOT_PATH)/include
+LIBAVIF_SRC_PATH        ?= $(LIBAVIF_ROOT_PATH)/src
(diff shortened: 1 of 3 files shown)


CrackedPixel:
Converting to draft as this is NOT meant to be merged, I just want to know what you think of the concept only

loftafi:
avif support would be awesome. One of my biggest annoyances with Godot is it doesnt support avif and I was already thinking that if I use raylib I can probably convert to using avif. (My app has about 600mb of jpg data, and avif would allow me to get it down to perhaps 300-400mb.)

raysan5:
@CrackedPixel I'm afraid that this implies an additional external dependency, raylib only support file-formats contained in portable single-file header-only libraries.

CrackedPixel:
> @CrackedPixel I'm afraid this change implies an additional external dependency and raylib only support file-formats contained in portable single-file header-only libraries.

great example for anyone who needs it, then :stuck_out_tongue_winking_eye:

CrackedPixel:
> avif support would be awesome. One of my biggest annoyances with Godot is it doesnt support avif and I was already thinking that if I use raylib I can probably convert to using avif. (My app has about 600mb of jpg data, and avif would allow me to get it down to perhaps 300-400mb.)

can you contact me on discord? i can probably get that 600mb jpeg to 100mb png for you

casperbear:
@CrackedPixel Trying to integrate it, but cannot figure out how to do it. I don't use cmake, I just compile VS2022\raylib.sln
If you have time to answer my question, do you think it is enough to add avif.h / avif.c to raylib solution or code in other c files in https://github.com/AOMediaCodec/libavif is also needed and I should somehow build lib file from there and make raylib use it?

PR #4787, opened 2025-02-22
Amend comment to explain what LoadCodepoints is doing.

loftafi:
As a beginner, it was not clear to me what the purpose of the LoadCodepoints function is. I am trying to help past me.


raysan5:
@loftafi Thanks for the review.

PR #4793, opened 2025-02-25
Fix this wording

jordan4ibanez:
A very small fix.


raysan5:
@jordan4ibanez 77 files modified is not a small fix... still, it's ok...

PR #4795, opened 2025-02-25
update RGFW to 1.6

ColleagueRiley:
- update RGFW to 1.6

- fixes mistakes in `rcore_desktop_rgfw.c`

- loads/unloads gamepads when the `RGFW_gamepadConnected` `RGFW_gamepadDisconnected` event happens instead of when they first send input.


raysan5:
@ColleagueRiley thanks for the review!

PR #4796, opened 2025-02-25
example resize fixes

jopadan:
- fix screen size not being updated at window resize event
- fix size dependent font/rectangle/object draw


jopadan:
@raysan5 Added images for one example.

jopadan:
Is there any chance of getting a response?

raysan5:
@jopadan As I told you those are too many changes for a single PR. I need some time to review them properly. Still wondering why all those changes are required.

jopadan:
@raysan5 All these changes are required and only relate to examples to fix none adaptive viewport resize.
There are unnecessary calls to `GetScreenWidth()/GetScreenHeight()` that should be avoided adding event callbacks in order to circumvent performance penalties.

nezvers:
Looks like only examples are modified. Mostly just replacing `screenWidth` with `GetScreenWidth() and same for height.

raysan5:
I'm afraid I'm not merging this PR, too big and find it confusing. If it is related to HighDPI, it should/could be addressed internally in some other way.

jopadan:
It is in no way related to HIGH_DPI mentioned in #4836. As mentioned earlier it only effects examples and corrects cosmetic HUD size updates. I discovered `IsWindowResized()` now and will update the examples with it

PR #4798, opened 2025-02-25
[rcore][RGFW] bug fixes

ColleagueRiley:
- ensure functions that use `platform.window` check `platform.window` is valid first 
- ensure fullscreen toggling is working properly 
- ensure raylib responds to `RGFW_quit` properly 
- update RGFW
- handle `RGFW_windowMaximized`, `RGFW_windowMinimized` and `RGFW_windowRestored` events
- properly handle raylib's `FLAG_WINDOW_UNFOCUSED` flag with `RGFW_windowFocusOnShow `
- remove out-of-date code

### DIFF
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -121,6 +121,7 @@ void CloseWindow(void);
 typedef struct {
     RGFW_window *window;                // Native display device (physical screen connection)
     RGFW_monitor mon;
+    double base;
 } PlatformData;
 
 //----------------------------------------------------------------------------------
@@ -256,7 +257,7 @@ bool InitGraphicsDevice(void);   // Initialize graphics device
 // Check if application should close
 bool WindowShouldClose(void)
 {
-    if (CORE.Window.shouldClose == false && platform.window != NULL)
+    if (CORE.Window.shouldClose == false)
         CORE.Window.shouldClose = RGFW_window_shouldClose(platform.window);
     if (CORE.Window.ready) return CORE.Window.shouldClose;
     else return true;
@@ -265,8 +266,6 @@ bool WindowShouldClose(void)
 // Toggle fullscreen mode
 void ToggleFullscreen(void)
 { 
-    if (platform.window == NULL) return;
-
     if (!CORE.Window.fullscreen)
     {
         // Store previous window position (in case we exit fullscreen)
@@ -278,7 +277,7 @@ void ToggleFullscreen(void)
         CORE.Window.flags |= FLAG_FULLSCREEN_MODE;
 
         RGFW_monitor_scaleToWindow(platform.mon, platform.window);
-        RGFW_window_setFullscreen(platform.window, 1);
+       // RGFW_window_setFullscreen(platform.window, 1);
     }
     else
     {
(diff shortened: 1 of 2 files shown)


sleeptightAnsiC:
> ensure functions that use platform.window check platform.window is valid first

This is something similar that rcore_desktop_glfw.c does inside of [WindowShouldClose()](https://github.com/raysan5/raylib/blob/eb8a343e313967a51cb302ac9bb1206a05727d13/src/platforms/rcore_desktop_glfw.c#L150) but no other function has those kind of checks. Other backends also don't have them. If this gets accepted, can we also port it to SDL, GLFW, etc? @raysan5 

(also linking this to https://github.com/raysan5/raylib/issues/4751 )

src/platforms/rcore_desktop_rgfw.c
```diff
@@ -351,7 +360,7 @@ void SetWindowState(unsigned int flags)
 {
     CORE.Window.flags |= flags;
 
-    if (flags & FLAG_VSYNC_HINT)
+    if (flags & FLAG_VSYNC_HINT && platform.window != NULL)
```
sleeptightAnsiC, line 363: Can't this just return at the very top, just like other *Window functions do in this PR: `if (platform.window == NULL) return;` ?

raysan5:
@ColleagueRiley @sleeptightAnsiC In raylib I tried to avoid/minimize this kindbof checks, spreaded all around the code on all functions. Can it be checked only on InitWindow()?

src/platforms/rcore_desktop_rgfw.c
raysan5: Personally I'm not a fan of early-returns, I tried to avoid them for years but in the last few years some were added, there is an open issue to consider this kind of "safety" decisions.

raysan5:
> Currently, there is a problem with raylib's initialization where, if a platform cannot provide Window and/or OpenGL context, the whole application will crash deep inside of rcore backend code, without any possibility to recover.

@sleeptightAnsiC Could it be just validated if the process (window+context creation) worked properly and just show a TraceLog() message showing the issue (like most raylib functions approach) if not initialized and then just set `CORE.Window.ready` to false for further user check?

sleeptightAnsiC:
> @sleeptightAnsiC Could it be ...

Replied under linked issue.

ColleagueRiley:
If the window is not valid or is invalidated, there will be a confusing crash. 

This check is there to avoid the crash, but I agree that this is not how it should be done. Instead of the if-statement, there should be an assert on each function `platform.window` so they crash in an obvious way if they are used before initialization or after deinitialization.  

If you think there shouldn't be any asserts or checks, I will remove them. But they could be useful if something goes wrong. 

That other issue is not relevant for RGFW internally, RGFW fallbacks to a basic OpenGL context if it fails to load the accelerated context. The crash will still happen because RLGL does not use a fallback, but that could be easily added. 

As for `SetWindowState`:  It assumes that the user might try setting the state before the window is created and modifies flags that can only take effect when creating the window.  that may not be a correct assumption.

src/platforms/rcore_desktop_rgfw.c
```diff
@@ -303,8 +308,7 @@ void ToggleFullscreen(void)
 // Toggle borderless windowed mode
 void ToggleBorderlessWindowed(void)
 {
-    if (platform.window == NULL)
-        return;
+    if (platform.window == NULL) return;
```
raysan5, line 311: I prefer to avoid adding that check to all the functions, if that check is required I think it should be implemented in RGFW library side.

ColleagueRiley: Alright, they exist on the RGFW side as asserts.

ColleagueRiley:
I accidentally pushed my copy of the samples that I was testing on.

Issue #4797, opened 2025-02-25
[cmake] Default lib target


brccabral:
Hi,
The default build type when build raylib is STATIC libraries.
But when added to a project, the raylib-config.cmake expects raylib_USE_STATIC_LIBS OFF.
I know it is just a flag, but would be nice if they were consistent with each other.

Thanks,
Bruno

raysan5:
@brccabral Agree. raylib prioritizes static build and linkage. Feel free to send a PR to review it.

PR #4799, opened 2025-02-26
[cmake] Use STATIC lib ON by default

brccabral:
Default target when building raylib itself is STATIC, but projects that use raylib was SHARED.
Changed raylib-config.cmake to consider STATIC as the default linkage.
Related to #4797


raysan5:
@brccabral thanks for the update!

Issue #4801, opened 2025-02-28
[rcore] `InitWindow()` can cause a crash when Window and/or OpenGL context failed to initialize / `IsWindowReady()` always returns true, meaning there is no way to recover


sleeptightAnsiC:
First mentioned here https://github.com/raysan5/raylib/pull/4798#issuecomment-2690141258 : 

> Currently, there is a problem with raylib's initialization where, if a platform cannot provide Window and/or OpenGL context, the whole application will crash deep inside of rcore backend code, without any possibility to recover.

Given following code:

```c
#include "raylib.h"

int main(void)
{
	InitWindow(800, 600, "sample text");

	// exit if Window/OpenGL context creation failed
	if (!IsWindowReady()) {
		return 1;
	}

	// main loop
	while (!WindowShouldClose()) {
		BeginDrawing();
		ClearBackground(GRAY);
		EndDrawing();
	}
	CloseWindow();

	return 0;
}
```

and compiled as follows:

```sh
$ git clone https://github.com/raysan5/raylib.git
$ cd raylib/src
$ make RAYLIB_BUILD_MODE=DEBUG
$ cc main.c -o  main -O0 -g -I. -L. -lraylib -lm -lGL
$ ./main
```

I expect that this program will exit with error code 1, if InitWindow() failed to initialize Window/OpenGL context. This is exactly what note about [IsWindowReady()](https://github.com/raysan5/raylib/blob/eb8a343e313967a51cb302ac9bb1206a05727d13/src/raylib.h#L976) suggests.

However, this is not what is happening right now. Application simply crashes, either somewhere inside of InitWindow() or somewhere later during main loop. There are multiple cases where this can happen (please, click at them in order to see the details):

<details><summary>[ Case 1: old GPU trying to load newer OpenGL context ]</summary>
<p>

```sh
$ ./main
INFO: Initializing raylib 5.6-dev
INFO: Platform backend: DESKTOP (GLFW)
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
INFO:     > raudio:.... loaded (optional)
WARNING: GLFW: Error: 65543 Description: GLX: Failed to create context: GLXBadFBConfig
[...]

PR #4803, opened 2025-02-28
[rcore] fix crash in InitWindow, due to unchecked result of InitPlatform

sleeptightAnsiC:
Check the result of InitPlatform(), if it isn't 0, report the Error and return. This prevent crashes and allows for gracefully aborting or recovering by checking IsWindowReady().

I decided to use `LOG_ERROR` here, because this is a fatal error in most cases, and it will cause a crash when ignored. However, I'm not so sure about `SYSTEM` prefix. There was no good choice for it. It must be very generic because InitPlatform() can fail in multiple cases.

Tested on native x86_64 Linux, and on x86_64 MacOS 13.6.9 running in VM. With GLFW in both cases.

Partially-fixes: https://github.com/raysan5/raylib/issues/4801

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -685,7 +685,7 @@ void InitWindow(int width, int height, const char *title)
 
     if (result != 0)
     {
-        TRACELOG(LOG_ERROR, "SYSTEM: Failed to initialize Platform");
+        TRACELOG(LOG_WARNING, "SYSTEM: Failed to initialize Platform");
         return;
     }
     //--------------------------------------------------------------
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -681,7 +681,10 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    InitPlatform();
+    if(InitPlatform() != 0) {
+        TRACELOG(LOG_ERROR, "SYSTEM: Failed to initialize Platform");
+        return;
+    }
     //--------------------------------------------------------------
 
     // Initialize rlgl default data (buffers and shaders)
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -681,7 +681,10 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    if(InitPlatform() != 0) {
+    int result = InitPlatform();
+
+    if (result != 0)
+    {
         TRACELOG(LOG_ERROR, "SYSTEM: Failed to initialize Platform");
         return;
     }


src/rcore.c
```diff
@@ -681,7 +681,10 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    InitPlatform();
+    if(InitPlatform() != 0) {
```
raysan5, line 684: In raylib I tried to avoid this format approach, also code is not following raylib code conventions. I'd prefer to do it:
```c
int result = InitPlatform();

if (result != 0) 
{
    TRACELOG(LOG_ERROR, "SYSTEM: Failed to initialize Platform");
    return;
}
```

sleeptightAnsiC: @raysan5 pushed suggested change as requested

src/rcore.c
```diff
@@ -681,7 +681,12 @@ void InitWindow(int width, int height, const char *title)
 
     // Initialize platform
     //--------------------------------------------------------------
-    InitPlatform();
+    int result = InitPlatform();
+
+    if(result != 0) {
```
raysan5, line 686: Please, could you follow raylib code conventions?

src/rcore.c
```diff
@@ -681,7 +681,12 @@ void InitWindow(int width, int height, const char *title)
...
+        TRACELOG(LOG_ERROR, "SYSTEM: Failed to initialize Platform");
```
raysan5, line 687: Afair, raylib always prioritizes LOG_WARNING over LOG_ERROR

src/rcore.c
sleeptightAnsiC: Ah I see, it's about trailing `{`, right?
Sorry, I just didn't notice this.

src/rcore.c
sleeptightAnsiC: > I decided to use LOG_ERROR here, because this is a fatal error in most cases, and it will cause a crash when ignored.

If this place fails, the whole library is pretty much unusable. In this state, it's only possible to exit the application or use very limited functionality of raylib (like RTEXT or some file IO utilities from RCORE) but anything related to GPU/Window will crash.

Is LOG_WARNING still preferred here? Just want to make sure.

src/rcore.c
sleeptightAnsiC: it should be
```
if (result != 0)
{
```
(space after `if` and newline before `{`)

sleeptightAnsiC: Fixed by force pushing (no need for another commit, since the last one was also a style change). Apologies for not noticing this the first time.

src/rcore.c
sleeptightAnsiC: Fixed by another commit, so we won't have to wait for each-others response. I still think ERROR may be more preferred here, considering the context, but you're the boss here :smile: 

In any case, PR can be merged from commit that still uses the ERROR.

sleeptightAnsiC:
Sorry for inconvenience @raysan5
I simply didn't notice all issues in invalid code style, the first time you pointed it out.
Everything should be resolved now.

raysan5:
@sleeptightAnsiC Thanks for the quick review! About the LOG_WARNING, I always try to give the user ownership over errors, as well as input provided info.

orcmid:
> @sleeptightAnsiC Thanks for the quick review! About the LOG_WARNING, I always try to give the user ownership over errors, as well as input provided info.

Who is "the user" in this context?

src/rcore.c
orcmid: @raysan5 @sleeptightAnsiC 
> If this place fails, the whole library is pretty much unusable. In this state, it's only possible to exit the application or use very limited functionality of raylib (like RTEXT or some file IO utilities from RCORE) but anything related to GPU/Window will crash.

Thinking out loud:

Launch a shell/terminal script that displays a message that the application cannot run and then pauses for confirmation before closing.  Use specific one for user language if the app supports more than one.

If there's a log to display, it can be MOREd in the script.

Sounds like an interesting side project and maybe something to add in examples :).  Depends on availability of shell processing though.

PR #4804, opened 2025-02-28
[desktop_glfw] fix InitPlatform crash caused by glfwCreateWindow...

sleeptightAnsiC:
...returning NULL. This was causing a crash few lines later.
This was described in detail here: https://github.com/raysan5/raylib/issues/4801#issuecomment-2691201072

Note that rcore_web.c has a very similar code path, but unlike in rcore_desktop_glfw.c, WEB version checks for null in all branches before doing anything. 

Tested on native x86_64 Linux, and on x86_64 MacOS 13.6.9 running in VM. With GLFW in both cases.

Partially-fixes: https://github.com/raysan5/raylib/issues/4801


raysan5:
@sleeptightAnsiC looks good, thanks for the review.

Issue #4809, opened 2025-03-04
[rcore] More spline functions


AmityWilder:
### Description

It would be helpful to have spline functions for velocity--to sample a curve's direction--and calculating bounds--for collision/rendering optimization. I think these would fit in nicely alongside the `GetSplinePoint_` functions.

`GetSplineVelocityLinear`\
`GetSplineVelocityBezierQuad`\
`GetSplineVelocityBezierCubic`\
`GetSplineAccelerationBezierQuad`\
`GetSplineAccelerationBezierCubic`\
`GetSplineJoltBezierCubic`

`GetSplineBoundsBezierLinear`\
`GetSplineBoundsBezierQuad`\
`GetSplineBoundsBezierCubic`

`GetSplineCurvatureBezierCubic`\
`GetSplineNearestTLinear`

It might also be nice to have spline drawing with variable thickness along the curve, since this can't currently be accomplished without some pretty advanced trickery.

I have already written the code for these as part of my Raylib vector art program and thought they might be valuable for others as a contribution to Raylib.
The variable thickness one needs a little work because it doesn't translate 1:1 from object-oriented Rust directly into C (I will continue to work on that further), but it is already working pretty well as a basic calligraphic effect.

![Image](https://github.com/user-attachments/assets/ba86226c-e9f8-4ba9-8731-6fd674fe212b)

raysan5:
@AmityWilder Those improvements are really nice but I'm afraid they are growing considerably, maybe they should be moved to a separate library... let me review them carefully...

raysan5:
@AmityWilder As commented in the open PR, I think that moving the points-generation math and triangles-generation to its own single-file header-only library would be great, so `rshapes` module could include that library to implement the spline-drawing functions but the header would also be available to users with more advance spline-generation/management requirements. That library would also be platform/engine agnostic and portable.

I'm closing this issue but please, let's continue the discussion in the open PR: https://github.com/raysan5/raylib/pull/4810

PR #4810, opened 2025-03-04
More spline functions

AmityWilder:
Relating to #4809

### Added math functions

`GetSplineVelocityLinear`\
`GetSplineVelocityBezierQuad`\
`GetSplineVelocityBezierCubic`\
`GetSplineAccelerationBezierQuad`\
`GetSplineAccelerationBezierCubic`\
`GetSplineJoltBezierCubic`

`GetSplineBoundsBezierLinear`\
`GetSplineBoundsBezierQuad`\
`GetSplineBoundsBezierCubic`

`GetSplineCurvatureBezierCubic`\
`GetSplineNearestTLinear`

### Added draw function

`DrawSplineSegmentBezierCubicVar`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -179,6 +179,19 @@ WIP: Last update with commit from 02-Nov-2024
 [rshapes] REVIEWED: `CheckCollisionCircleRec()` (#3584) by @ubkp
 [rshapes] REVIEWED: Add more detail to function comment (#4344) by @Jeffery Myers
 [rshapes] REVIEWED: Functions that draw point arrays take them as const (#4051) by @Jeffery Myers
+[rshapes] ADDED: `DrawSplineSegmentBezierCubicVar()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineVelocityLinear()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineVelocityBezierQuad()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineVelocityBezierCubic()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineAccelerationBezierQuad()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineAccelerationBezierCubic()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineJoltBezierCubic()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineBoundsBezierLinear()` (#TBD) by @AmityWilder
+[rshapes] ADDED: `GetSplineBoundsBezierQuad()` (#TBD) by @AmityWilder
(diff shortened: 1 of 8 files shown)


raysan5:
@AmityWilder I'm afraid this PR has become too big with too many additions, increasing complexity. Despite useful, probably many of those spline-specific functions are not required for most raylib users and they increase API complexity and maintenance cost... 

I can add the 2 new draw functions but all the `GetSpline*()` functions should be moved to a separate spline-specific header-only library or just included in the example as static functions for now.

What do you think? Please, let me know to move this PR forward...

EDIT: An ideal option would be to have a separate single-file header-only library, maybe `rsplines.h` that could implement the generation of spline-vertex-triangle data, independently of raylib-rlgl, like a pure-spline-math library, so raylib could use that library on the `DrawSpline*()` functions and users could include it if further spline functionality was required.

AmityWilder:
> @AmityWilder I'm afraid this PR has become too big with too many additions, increasing complexity. Despite useful, probably many of those spline-specific functions are not required for most raylib users and they increase API complexity and maintenance cost...
> 
> I can add the 2 new draw functions but all the `GetSpline*()` functions should be moved to a separate spline-specific header-only library or just included in the example as static functions for now.
> 
> What do you think? Please, let me know to move this PR forward...
> 
> EDIT: An ideal option would be to have a separate single-file header-only library, maybe `rsplines.h` that could implement the generation of spline-vertex-triangle data, independently of raylib-rlgl, like a pure-spline-math library, so raylib could use that library on the `DrawSpline*()` functions and users could include it if further spline functionality was required.

That sounds alright with me, would you like me to filter this PR down to just those draw functions?

AmityWilder:
@raysan5 
> EDIT: An ideal option would be to have a separate single-file header-only library, maybe rsplines.h that could implement the generation of spline-vertex-triangle data, independently of raylib-rlgl, like a pure-spline-math library, so raylib could use that library on the DrawSpline*() functions and users could include it if further spline functionality was required.

How does something like this look? (based a good chunk of it off of raygui and copied over the already-implemented functions from the PR with the addition of optional 1D and 3D variations of them&mdash;I'm not opposed to changing the names of them before release)
https://github.com/AmityWilder/rsplines

raysan5:
@AmityWilder Thanks! It looks great!

Let me see how it can be integrated into `DrawSpline*()` to request the points...

raysan5:
@AmityWilder I've reviewed this PR again and I'm afraid I'm not merging it. The added splines functionality is actually fantastic, `rsplines` cover most spline requirements for any project but I think it's too much for raylib. It would imply redesigning all splines functions to request the segments using `rsplines` and it would add a new dependency that at this moment I prefer to avoid.

Still I think `rsplines` could be great library for anyone with advance splines requirements. As an improvement for the library, I'd remove raylib types dependency, providing its own types, to make the library portable and self-contained.

In any case, thanks for the all the work put into this library and thanks for your patience with this PR.

PR #4811, opened 2025-03-04
[rtext] Fix `stb_truetype` composite glyph scaling logic

AshishBhattarai:
The logic for handling composite glyph scaling doesn't match the truetype specs, hence the results are different when a glyph in a ttf file defines composite glyh scales. I couldn't really find a ttf font with a glyph that has this exact issue, but the specs allows both uniform and non-uniform scaling for composite glyphs.

Part of the spec i am referring to
image

Synthetic tests (intentionally add scale to glyph component, spec allows it)
Left is FontForger, Right is raylib
image

With this fix applied
image

### DIFF
--- a/src/external/stb_truetype.h
+++ b/src/external/stb_truetype.h
@@ -1863,11 +1863,11 @@ static int stbtt__GetGlyphShapeTT(const stbtt_fontinfo *info, int glyph_index, s
                stbtt_vertex* v = &comp_verts[i];
                stbtt_vertex_type x,y;
                x=v->x; y=v->y;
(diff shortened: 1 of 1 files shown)


raysan5:
@AshishBhattarai I'm afraid this change belongs to `stb_truetype` library, it should be patched in original repo instead of raylib side.

AshishBhattarai:
> @AshishBhattarai I'm afraid this change belongs to `stb_truetype` library, it should be patched in original repo instead of raylib side.

just opened a pr on stb
https://github.com/nothings/stb/pull/1756

raysan5:
@AshishBhattarai thanks, just merging this change but note it can be overwritten on library update...

PR #4812, opened 2025-03-05
Fix GetCurrentMonitor comments.

Kaluub:
Hello, my editor was pulling the function description from the raylib source and the comment for GetCurrentMonitor was not accurate which was confusing. This is just a small fix to replace the comment with what is in raylib.h.


RobLoach (approved):
Yuppers. Good fix.

raysan5:
@Kaluub Good catch! Thanks for the fix!

PR #4817, opened 2025-03-06
[Raymath] fix compiler warning missing initializer for member

schwieni:
See [https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmissing-field-initializers] for details. Somehow the current gcc doesn't like code like `Vector2 result = { 0 };`. This is an issue if you are using -Werror=missing-field-initializers, since the code won't compile at all.


schwieni:
@raysan5 thanks for the review. I understand and I can change that no problem.
Is there a reason why you closed this PR? I could have added another commit to this PR with the changes following your review.
Anyhow, I can make another PR.
Regards

PR #4818, opened 2025-03-06
[examples] Potential for new example: `text_styled_texts`

SultansOfCode:
Hello!

I have added the ability to customize text color and background color with special characters inside the text (pretty much like mIRC does)

You can use the functions `DrawTextStyled` and `DrawTextStyledEx` almost as you would use `DrawText` and `DrawTextEx` respectively

The changes are: instead of passing the Color, you will pass an array of Colors and the size of this array

Inside the text you can use `\003` to set the foreground color, `\004` to set the background color, `\015` to reset them to their default values (first color in array, if any, as foreground color [or black, if none], and transparent background) and `\022` to swap foreground and background colors

Also added `MeasureTextStyled` and `MeasureTextStyledEx` to correctly measure them as `MeasureText` and `MeasureTextEx` would for plain texts

If any adjustment is needed, either to the code or to the PR, please, let me know

Thank you

### DIFF
--- a/examples/text/text_styled.c
+++ b/examples/text/text_styled.c
@@ -1,0 +1,70 @@
+/*******************************************************************************************
+*
+*   raylib [text] example - Text styled
+*
+*   Example complexity rating: [★☆☆☆] 1/4
+*
+*   Example originally created with raylib 5.5, last time updated with raylib 5.5
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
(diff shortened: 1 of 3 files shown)


raysan5:
@SultansOfCode Thanks for the nice effect! I'm afraid this implementation is too custom to be added to raylib for all users, the styling mechanism should be a user-decision, not imposed by raylib.

Still, it can be directly implemented into a single example intended for users looking for an implementation sample or as a base solution to implement their own.

raysan5:
I'm not adding this functionality to raylib main library but I'll try to create a custom example from it.

Issue #4814, opened 2025-03-06
Raylib doesn't build with Zig 0.14.0


10aded:
### Issue description

Raylib fails to build with `zig build` for the current standard version of Zig, `0.14.0`.

One of both of `build.zig` and `build.zig.zon` need to be updated to work with Zig `0.14.0`.

### Issue Screenshot

```
PS C:\Users\10aded\Documents\Temp\Raylib-Curr> zig version
0.14.0
PS C:\Users\10aded\Documents\Temp\Raylib-Curr> zig build
C:\Users\10aded\Documents\Temp\Raylib-Curr\build.zig.zon:2:13: error: expected enum literal
    .name = "raylib",
            ^~~~~~~~
PS C:\Users\10aded\Documents\Temp\Raylib-Curr>
```

demilich1:
Are you sure about this? It seems to work for me. Here are the steps I took (using zig 0.14.0 ofc):

1. Create new empty directory
2. use `zig init` to initialize a new Zig project
3. `zig fetch --save=raylib https://github.com/raysan5/raylib/archive/f1385f3.zip` (most recent commit as of this writing)
4. Add the following block to `build.zig`:
 ```
// Raylib include
const raylib_dep = b.dependency("raylib", .{ .target = target, .optimize = optimize });
const raylib = raylib_dep.artifact("raylib");
exe.linkLibrary(raylib);
```
5. Add example code to `main.zig` to open a window
6. `zig build run` -> Works

I think the error message is actually complaining about your `build.zig.zon`?

**Comment by blogscot** [comment]
@10aded I came across this problem too. As I only had one dependency, I solved it by deleting the `build.zig.zon` file (Zig 0.14.0 has made some changes to how the build system works) and regenerated a new version using `zig init`, and finally fetched the dependency again.

**Comment by 10aded** [comment]
> Are you sure about this?

@demilich1 Yes. The default `build.zig.zon` file in *Raylib* is flawed: try the following on your system; it'll almost surely fail for you too:

```
$ mkdir Raylib_Temp
$ cd Raylib_Temp
$ git clone --depth 10 https://github.com/raysan5/raylib.git .
$ zig version
0.14.0
$ zig build
C:\Users\10aded\Documents\Temp\Raylib_Temp\build.zig.zon:2:13: error: expected enum literal
    .name = "raylib",
            ^~~~~~~~
$
```

[...]

PR #4819, opened 2025-03-07
Update build.zig.zon to use the new zig v0.14 version

zewenn:
The new zig version changed the `build.zig.zon` file. Not updating to the new standard causes the projects to automatically re-fetch the package on every build (resulting in 10-15 second build times).

I have updated the `build.zig.zon` to use zig v0.14.0 and the appropriate [fingerprint](https://ziglang.org/download/0.14.0/release-notes.html#New-Package-Hash-Format).

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,7 +1,9 @@
 .{
-    .name = "raylib",
+    .name = .raylib,
     .version = "5.5.0",
-    .minimum_zig_version = "0.13.0",
+    .minimum_zig_version = "0.14.0",
+
+    .fingerprint = 0x13035e5cb8bc1ac2,
 
     .dependencies = .{
         .xcode_frameworks = .{


raysan5:
@zewenn thanks for the review!

10aded:
Great, this closes #4814.

However, per the documentation when creating a package via `zig init`, it is highly recommended that for security reasons a comment accompany the fingerprint line.

From `build.zig.zon` when generated via `zig init`:

```
// Together with name, this represents a globally unique package
// identifier. This field is generated by the Zig toolchain when the
// package is first created, and then *never changes*. This allows
// unambiguous detection of one package being an updated version of
// another.
//
// When forking a Zig project, this id should be regenerated (delete the
// field and run `zig build`) if the upstream project is still maintained.
// Otherwise, the fork is *hostile*, attempting to take control over the
// original project's identity. Thus it is recommended to leave the comment
// on the following line intact, so that it shows up in code reviews that
// modify the field.
.fingerprint = 0x6f178c1f29db3887, // Changing this has security and trust implications.
```

The fingerprint line should have the comment

`// Changing this has security and trust implications.`

I'll submit a pull request with this change soon.

10aded:
PR made: #4827

Issue #4820, opened 2025-03-07
[examples] some sources use `#include <raylib.h>` with `< >` instead of `" "`


sleeptightAnsiC:
I was testing few changes applied to raylib by compiling all examples with Cmake, and I noticed that some examples use older system-wide installation of raylib instead of the edited local one. 

After taking a look at those examples, it appears they use `#include <raylib.h>` with `< >` instead of `" "` which causes said issue. Quick grep search shows there are 4 examples affected by this:
```sh
$ git clone https://github.com/raysan5/raylib.git
$ cd raylib/examples
$ grep -r 'include *<raylib\.h>' .
./textures/textures_image_channel.c:#include <raylib.h>
./shapes/shapes_draw_circle_sector.c:#include <raylib.h>
./shapes/shapes_draw_ring.c:#include <raylib.h>
./shapes/shapes_draw_rectangle_rounded.c:#include <raylib.h>
```

All examples should use `" "` includes.

Tested with latest raylib pulled from git on Linux:
```sh
$ uname -a
Linux MAL200424 6.13.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 27 Feb 2025 18:09:44 +0000 x86_64 GNU/Linux
```

(I'll create a patch for this in a sec)

PR #4821, opened 2025-03-07
[examples] fix: use quotation marks when including raylib.h

sleeptightAnsiC:
Fixes: https://github.com/raysan5/raylib/issues/4820


raysan5:
@sleeptightAnsiC Good catch! Thanks for the fix!

PR #4822, opened 2025-03-07
Add call-back user data pointer

sockentrocken:
The purpose of this pull request is to add a void pointer to user-data to every call-back in RL. The idea is that an user can set a pointer to, for instance, an asset manager in their code, which may only be living within the scope of a function, and not on the global scope. (Rust is an example of a language with a RL binding project which does not allow global state, or at the very least, does not encourage it.) Here's an example of setting a pointer to a .ZIP file handle to retrieve data from it, without relying on global state.
[raylib-test.zip](https://github.com/user-attachments/files/19131570/raylib-test.zip)

```c
#include <stdio.h>
#include "raylib.h"

// Use ZIP library (https://github.com/kuba--/zip) for this demonstration.
#include "zip/zip.h"

// Create our LoadFileDataCallback call-back function.
unsigned char* call_back(void* userData, const char *fileName, int *dataSize)
{
	// Retrieve the open ZIP handle we made in main().
	struct zip_t *user = (struct zip_t *) userData;

	void *buf = NULL;
	size_t bufsize;

	// Read data out of the ZIP file.
	{
	    zip_entry_open(user, fileName);
	    {
	        zip_entry_read(user, &buf, &bufsize);
	    }
	    zip_entry_close(user);
	}

	// Set the size of the buffer.
	*dataSize = bufsize;
	
	// Return file data from the ZIP file.
	return buf;
}

int main(void)
{
    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "raylib [core] example - basic window");

    SetTargetFPS(60);

    // Open our sample ZIP file. The ZIP file has only one file in it: texture.png.
	struct zip_t *zip = zip_open("foo.zip", 0, 'r');

    // Cast our open ZIP file handle to a void pointer. We will get this back in the call-back.
	void* pointer = (void*) zip;

	// Set call-back with our own custom user data-pointer.
    SetLoadFileDataCallback(pointer, call_back);

    // Load texture.png from the ZIP file.
    Texture texture = LoadTexture("texture.png");

    while (!WindowShouldClose())
    {
        BeginDrawing();

            ClearBackground(RAYWHITE);

            DrawText("Congrats! You created your first window!", 190, 200, 20, LIGHTGRAY);

        EndDrawing();
    }

    UnloadTexture(texture);
    zip_close(zip);

    CloseWindow();

    return 0;
}
```

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -950,11 +950,11 @@ typedef enum {
 
 // Callbacks to hook some internal functions
 // WARNING: These callbacks are intended for advanced users
-typedef void (*TraceLogCallback)(int logLevel, const char *text, va_list args);  // Logging: Redirect trace log messages
-typedef unsigned char *(*LoadFileDataCallback)(const char *fileName, int *dataSize);    // FileIO: Load binary data
-typedef bool (*SaveFileDataCallback)(const char *fileName, void *data, int dataSize);   // FileIO: Save binary data
-typedef char *(*LoadFileTextCallback)(const char *fileName);            // FileIO: Load text data
-typedef bool (*SaveFileTextCallback)(const char *fileName, char *text); // FileIO: Save text data
+typedef void (*TraceLogCallback)(void* userData, int logLevel, const char *text, va_list args);  // Logging: Redirect trace log messages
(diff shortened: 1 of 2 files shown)


raysan5:
@sockentrocken I'm afraid I'm not breaking current API

PR #4823, opened 2025-03-07
[Raymath] fix compiler warning missing initializer for member

schwieni:
See [https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wmissing-field-initializers] for details. Somehow the current gcc doesn't like code like `Vector2 result = { 0 };`. This is an issue if you are using -Werror=missing-field-initializers, since the code won't compile at all.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -391,7 +391,7 @@ RMAPI Vector2 Vector2Divide(Vector2 v1, Vector2 v2)
 // Normalize provided vector
 RMAPI Vector2 Vector2Normalize(Vector2 v)
 {
-    Vector2 result = { 0 };
+    Vector2 result = { 0.0f, 0.0f };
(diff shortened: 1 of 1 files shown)


raysan5:
@schwieni Current initialization mechanism is valid in C.

sleeptightAnsiC:
@schwieni 

This seems like an issue with your build setup or a bug in GCC itself.
Documentation clearly says the warning should NOT appear in this case:

```
In C this option does NOT warn about the universal zero initializer ‘{ 0 }’:

struct s { int f, g, h; };
struct s x = { 0 };
```

Are you sure that you were compiling raylib sources as C (not C++) ?
What was your GCC version `gcc -v`? Could you provide some repro steps?

Asking, because I have GCCbugzilla access, and if this is an actual GCC issue, I could simply report it for you.

PR #4826, opened 2025-03-08
build.zig.zon: update hashes

david-vanderson:
zig 0.14 updated the hash format.  Using the old hash forces a refetch on each zig build.  See https://github.com/ziglang/zig/issues/23051 for explanation.


raysan5:
@david-vanderson thanks for the update

david-vanderson:
Thank you!

Issue #4828, opened 2025-03-08
[examples] Overlapping spline controls fight for mouse focus


AmityWilder:
### Description

The `shapes_splines_drawing.c` example has a bug where hovering over a control point while one is selected confuses the program and makes the points fight over which one should receive focus/selection.

This is because the example only checks if the mouse is *down*, not *pressed*, and also doesn't prevent the focus search from occurring while a point is already selected.

### Issue Screenshot

![Image](https://github.com/user-attachments/assets/171b0fd8-e8d9-4f14-9c01-53c14a6f5f4d)

AmityWilder:
Evidence of fix:

![Image](https://github.com/user-attachments/assets/05bd3ed1-e09e-44b9-b9b9-cada4a162717)

raysan5:
Fixed by https://github.com/raysan5/raylib/pull/4829

PR #4829, opened 2025-03-08
[examples] Reorganize some conditions to fix overlap bugs

AmityWilder:
Move some if statements to surround for loops instead of being inside them, and add additional conditions for ensuring selection is inactive before looking for a point to focus.

Related to #4828


raysan5:
@AmityWilder Very nice! Thanks for the review!

PR #4831, opened 2025-03-11
[rgl] Add Support for Integer Attributes in GPU via glVertexAttribIPointer

SnoopyPlayz:
Hello!

I've added a small but useful function that allows sending an int to the GPU instead of a float by using glVertexAttribIPointer instead of glVertexAttribPointer.

This is particularly helpful for cases involving bit-packing, as OpenGL automatically converts integers to floats unless explicitly handled with glVertexAttribIPointer. This change ensures that integer data remains intact without unintended conversions.


raysan5:
@SnoopyPlayz I'm afraid OpenGL ES 2.0 does not support `glVertexAttribIPointer()` function so I prefer to avoid adding this function.

Also, using `glVertexAttribPointer()` for integer types with no normalization should be enough for most use case situations with no lose of information.

PR #4833, opened 2025-03-12
[rcore] Use snprintf to prevent buffer overflow in path construction

MykBamberg:
In some cases, specific user configurations may allow for file/directory paths that exceed the set path limit. In such cases the current implementation of `ScanDirectoryFiles` would cause a buffer overflow.

This change makes sure that the path string does not exceed the buffer size, failing gracefully with the following warning.
```
WARNING: FILEIO: Path longer than 4096 characters ([very long path]...)
WARNING: FILEIO: Read files count do not match capacity allocated
```


raysan5:
@MykBamberg Thanks for the improvement. Note that I'm reviewing the implementation, I prefer to avoid the `sizeof()` approach for sizes.

Issue #4834, opened 2025-03-12
[bug] Resizing the window breaks HIGHDPI scaling with values from GetScreenWidth() and GetScreenHeight()


NiamhNightglow:
**WARNING: Please, read this note carefully before submitting a new issue:**

It is important to realise that **this is NOT A SUPPORT FORUM**, this is for reproducible BUGS with raylib ONLY.

There are lots of generous and helpful people ready to help you out on [raylib Discord forum](https://discord.gg/raylib) or [raylib reddit](https://www.reddit.com/r/raylib/).

Remember that asking for support questions here actively takes developer time away from improving raylib.

---

Please, before submitting a new issue verify and check:

 - [ x ] I tested it on latest raylib version from master branch
 - [ x ] I checked there is no similar issue already reported
 - [ x ] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [ x ] My code has no errors or misuse of raylib

### Issue description

*Briefly describe the issue you are experiencing (or the feature you want to see added to raylib). Tell us what you were trying to do and what happened instead. Remember, this is not the best place to ask questions. For questions, go to [raylib Discord server](https://discord.gg/raylib).*

The reported values from GetScreenWidth() and GetScreenHeight() are incorrect after resizing the window.  Prior to resizing the window you get the dimensions for before you apply DPI scaling.  After a resize you get the actual pixel width after applying DPI scaling.  This means to account for this you currently need to catch IsWindowResized() and latch the value and apply a scaling factor based on GetWindowScaleDPI() only if the window has been resized.

### Environment

*Provide your Platform, Operating System, OpenGL version, GPU details where you experienced the issue.*
Windows 11, DPI 150%/1.5, default quickstart backend

### Issue Screenshot

*If possible, provide a screenshot that illustrates the issue. Usually an image is better than a thousand words.*
Uhhh.... Idk if a screenshot would help much with this?

### Code Example

*Provide minimal reproduction code to test the issue. Please, format the code properly and try to keep it as simple as possible, just focusing on the experienced issue.*
I made a fork here that shows the issue.  Just run it and resize the window: https://github.com/NiamhNightglow/raylib-quickstart

raysan5:
Issue malformed.

PR #4836, opened 2025-03-12
[rcore] Correctly handle window size on resize in auto-scaled HIGHDPI environment

JeffM2501:
This PR fixes the issue reported in #4834
When we get a resize event from the platform, we are just setting the screen size to the raw data returned by the event.
But if the user enabled the HIGHDPI flag on startup, we scaled the render and window size by the DPI scale. This means that on startup you get one set of sizes for the window, but after a resize, you swap from emulated unscaled values up to scaled values, and GetScreenSize no longer matches the viewport size.

This PR unscales the screen size on resize for the 3 desktop platforms so that the screen size stays in the same 'unscaled' space as the viewport.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1753,8 +1753,14 @@ static void WindowSizeCallback(GLFWwindow *window, int width, int height)
 
     if (IsWindowFullscreen()) return;
 
-    // Set current screen size
+	// if we are doing automatic DPI scaling, then the "screen" size is divided by the window scale
+	if (IsWindowState(FLAG_WINDOW_HIGHDPI))
+	{
+		width = (int)(width / GetWindowScaleDPI().x);
+		height = (int)(height / GetWindowScaleDPI().y);
+	}
 
+    // Set current screen size
     CORE.Window.screen.width = width;
     CORE.Window.screen.height = height;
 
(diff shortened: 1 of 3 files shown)


NiamhNightglow:
I just pulled this branch and tested it in my copy of the quickstart, and this does appear to resolve the issues I was seeing with the GetScreenSize functions not returning the scaled values.  The values remained scaled when I built my game with this version even with the FLAG_WINDOW_HIGHDPI flag set after rescaling the window.

Peter0x44:
This is indenting with tabs instead of spaces, you can tell since github displays with a default tabstop of 8.

JeffM2501:
> This is indenting with tabs instead of spaces, you can tell since github displays with a default tabstop of 8.

fixed

raysan5:

nicbarker:
Just a heads up that this commit causes resizing to break with `FLAG_WINDOW_HIGHDPI` enabled on mac OSX.
Window size is initially reported correctly, then on first resize returns incorrectly values.

JeffM2501:
The mac does lie on it's DPI scale, the change probably needs to not happen for mac desktop platforms.. thanks apple...

ntnon:
> The mac does lie on it's DPI scale, the change probably needs to not happen for mac desktop platforms.. thanks apple...

Will you consider fixing this?

PR #4837, opened 2025-03-12
[rcore] Support window flags with initialization issues

Andersama:
This may not be the correct approach, however this appears to work. The idea is that before modifying `CORE.Window.flags` when first creating the window we keep a copy of the flags in order to call `SetWindowState` after initialization has completed, which should behave as if `MaximizeWindow` or `MinimizeWindow`, or conceptually any other flag modifying function were called after `InitWindow`.

This pull request only performs this for the windows platform, modify as needed in the switch statement at the end for others.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1351,6 +1351,8 @@ int InitPlatform(void)
     // additionally auto iconify restores the hardware resolution of the monitor if the window that loses focus is a fullscreen window
     glfwWindowHint(GLFW_AUTO_ICONIFY, 0);
 
+    // Keep flags to restore after initialization (for flags not supported on initialization)  
+    unsigned int originalCoreWindowFlags = CORE.Window.flags;
     // Check window creation flags
     if ((CORE.Window.flags & FLAG_FULLSCREEN_MODE) > 0) CORE.Window.fullscreen = true;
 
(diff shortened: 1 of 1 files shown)


JeffM2501:
I think you can do this for all the GLFW platforms, not just windows.
I think you should also check the other desktop platforms and see if they need the same fix.

raysan5:
@Andersama Sorry, I don't understand the use case of this change... user should know the flags they are modyfying...

JeffM2501:
> @Andersama Sorry, I don't understand the use case of this change... user should know the flags they are modifying

The use case is this.

```
SetConfigFlags(FLAG_WINDOW_MINIMIZED);
InitWindow(.....);
```
this does not maximize the window, because InitWindow strips those flags

you must do 

```
InitWindow(.....);
SetWindowState(FLAG_WINDOW_MINIMIZED); // or call MaximizeWindow();
```

`InitPlatform `does not support `FLAG_WINDOW_MINIMIZED`, so even if you set a flag, it gets cleared with no indication it does not work. This leaves the window in a confusing state where a flag was requested but was ignored.

The idea was to have the internal platform call SetWindowState() after the window is initialized by the platform so that it can apply any of the flags that the window initialization can't or doesn't support.

That way when you set these flags at config time, they are actually applied and not ignored silently.

Andersama:
@raysan5  @JeffM2501's summary pretty much covers it. The only thought I had about doing this was that since this is mostly do to with opening a window, the code could be move to the calling function instead*, which I think would make more sense. But I think its inside this call that most of the work for opening a window is happening anyway so that's where I made the modification.

Also, I wasn't sure if there were other reasons why these flags were ignored in this way.

raysan5:
@JeffM2501 @Andersama thanks for the explanation, I see the use case.

PR #4839, opened 2025-03-14
 Converting int to char

aidonmaster:
Fixes warning: narrowing conversion caused by trying to convert int to unsigned char


raysan5:
@aidonmaster good catch! thanks!

PR #4844, opened 2025-03-18
[rshapes] Changed startAngle and endAngle comparison to compare against FLT_EPSILON instead of 0 in DrawCircleSector, DrawCircleSectorLines, DrawRing, DrawRingLines functions

theundergroundsorcerer:
This change adds early return condition in the following  drawing functions:
`DrawCircleSector`, `DrawCircleSectorLines`, `DrawRing`, `DrawRingLines`.

Instead of comparing for equality, the comparisons are done against `FLT_EPSILON`.
**(Perhaps for consistency, it should be done against `EPSILON` in `raymath.h`, however `raymath.h`
header is not imported in `rshapes.c` and other functions (the ones that check against collisions for example) compare against `FLT_EPSILON` as well)**

Benefits: Prevents problems related to to numerical stability. Comparing floating numbers is ill advice in general.

I've tested these changes on a Mac  and they don't break the build.
In addition, I didn't notice any changes in the behavior of the following examples:
`shapes_draw_ring`, `core_input_gestures_web`, `shapes_draw_circle_segment`, `raymath_vector_angle` that use the affected functions.

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -285,7 +285,8 @@ void DrawCircleV(Vector2 center, float radius, Color color)
 // Draw a piece of a circle
 void DrawCircleSector(Vector2 center, float radius, float startAngle, float endAngle, int segments, Color color)
 {
-    if (startAngle == endAngle) return;
+    if (FLT_EPSILON < endAngle - startAngle && endAngle - startAngle < FLT_EPSILON) return;
+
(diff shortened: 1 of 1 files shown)


raysan5:
@veins1 Thanks for the further details on this behaviour. In that case I prefer to keep the simple equal check and avoid involving `FLT_EPSILON` in the process.

PR #4850, opened 2025-03-22
[rcore] fix: TRACELOG upon successfully changing directory

sleeptightAnsiC:
I find it quite confusing that `InitWindow()` logs CWD during initialization, but the actual `ChangeDirectory()` does not do this upon successfully changing directory. This patch adds one line with `TRACELOG()` for said condition. It uses the same message as the one from InitWindow. Tested on Linux by building and running some examples with Cmake.


raysan5:
@sleeptightAnsiC thanks for the improvement!

PR #4851, opened 2025-03-23
added control and alt checking

barodapride:
Will probably be rejected to keep the API bloat low but I wanted to learn if I can do a pull request.


raysan5:
@barodapride I'm afraid those keys can already be checked with provided functions.

PR #4856, opened 2025-03-25
[build] remove examples/build.zig, incorporate into main build.zig

marler8997:
This is a different way to solve the same issue that https://github.com/raysan5/raylib/pull/4854 solves.   That PR keeps the examples build separate and connects it to the raylib build via a zon file, this PR instead removes the second build.zig in the examples directory and incorporates it into the main build.zig.

My personal preference would be this PR but both solve the underlying issue.

Note that running `zig build` will still only build the raylib library, I've added a separate `examples` target that will build all the examples (what running `zig build` in the examples directory would have done before).

NOTE: I also updated the examples build configuration so that we only add the extra include paths for the examples that need it.


raysan5:
@marler8997 Great! Thanks for the review!

PR #4859, opened 2025-03-26
Update rcore_desktop_sdl.c

MikiZX1:
raylib app crashing when started and a gamepad is already connected to the PC (even if the gamepad is not used in the app). I only tested this with a gamepad that has a layout which is not recognized. Using SDL3 as backend.


raysan5:
@MikiZX1 Oh! Good catch! It looks like a typo, thanks for the review!

PR #4860, opened 2025-03-26
Avoid divide by 0 in core_custom_frame_control example

marler8997:


raysan5:

PR #4861, opened 2025-03-26
[examples] Added new shader example: `shader_depth_texture`

luis605:
Some raylib users may find it useful, for post processing and shadows stuff, to have an example that renders the depth texture to screen.


JeffM2501:
This PR does not add the example to any build systems, it should include makefile additions and visual studio projects

luis605:
@JeffM2501 I have removed shaders_view_depth compilation when Android is specified and added the file to the example's Makefile.

JeffM2501:
What about the MSVC project? You should add the example to all the standard builds.

PR #4862, opened 2025-03-26
Uncomment SetTargetFPS in core_window_flags

marler8997:
I think this line was accidently commented out in this commit:

3d1ae3500c731aa37d8f887cd165db480c4890eb

The FPS limit is needed to get the desired wait time for the "hide window" test, which uses the frame counter to hide the window for 3 seconds. On my machine without this limit it runs at over 1000 FPS and it appears like the hide window state is broken.

I also added some text that tells the user that it only hides the window for 3 seconds so they're not surprised when the window automatically reappears.

### DIFF
--- a/examples/core/core_window_flags.c
+++ b/examples/core/core_window_flags.c
@@ -52,7 +52,7 @@ int main(void)
 
     int framesCounter = 0;
 
-    //SetTargetFPS(60);               // Set our game to run at 60 frames-per-second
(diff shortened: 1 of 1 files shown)


raysan5:
@marler8997 Note that some examples contain commented code, just to be used as a reference for the user or illustrate some alternative functionality. Probably it was the case, but it's ok to enable it...

PR #4863, opened 2025-03-27
Prevent immediate window restore from minimize in core_window_flags

marler8997:
The core_window_flags example automatically restores the window from being minimized after 3 seconds. However, it only resets the frameCounter if the window is minimized through the app from inputting KEY_N.  This means if you miminize the window by other means (like pressing the minimize button) then the example will immediately restore the window because the frame counter wasn't reset. I've fixed this by setting the frameCounter back to 0 once it's expired. I also added a note in the UI that the example automatically restores the window so it's not a surprise.


raysan5:
@marler8997 thanks for the review

PR #4864, opened 2025-03-27
[build] only include rglfw.c for glfw platform

marler8997:
It looks like raylib doesn't require rglfw.c if you're using the RGFW backend.  I'm guessing the same is true for SDL but I haven't tested. Excluding this one file brings the raylib library down from 6.9 MB to 6.3 MB for RGFW.

However, one of the examples requires the symbols from rglfw.c, to accomodate this I added a function that will check whether raylib has already included rglfw and if not include it for that one example.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -179,7 +179,11 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
     raylib.addIncludePath(b.path("src/platforms"));
     switch (target.result.os.tag) {
         .windows => {
-            try c_source_files.append("src/rglfw.c");
+            switch (options.platform) {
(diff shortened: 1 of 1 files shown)


raysan5:
@marler8997 Oh! Good catch! Still, 6.3MB seems quite big to me, `libraylib.a` compiled with w64devkit for PLATFORM_DESKTOP with GLFW is about 2MB for me.

marler8997:
Ah, might be because that's a debug build.  I get 1.8 MB for a ReleaseSmall build via zig.  Looks like a ReleaseFast build jumps back up to 6.7 MB though.  A bit surprising ReleaseSmall is so much smaller.

PR #4865, opened 2025-03-27
[examples] Adding `core_high_dpi` example

marler8997:
This example enables HIGHDPI and renders a scale showing how the logical size compares to the pixel size of the window.

image

Screenshot 2025-03-27 at 11 16 57 AM


raysan5:
@marler8997 Would you mind renaming the example to `core_high_dpi`, all raylib examples follow at-least-3-words pattern on their names. Also, the example should be added to the VS2022 solution. Here the example contribution details: https://github.com/raysan5/raylib/blob/master/examples/examples_template.c

marler8997:
Ok, updated the name and added to the VS project. I don't have VS to test it so hopefully the CI covers it?

raysan5:
@marler8997 thanks for the addition! 😄

PR #4866, opened 2025-03-28
Improve description of RestoreWindow

marler8997:
Restore window currently says it sets the window state to:

   "not minimized/maximized"

However, if a window is maximized and then minimized, it's typical that it would restore back to being maximized, which is what seems to happen from my testing. I've reworded the description to better reflect this behavior.


raysan5:

PR #4868, opened 2025-03-31
core_window_flags example: add borderless windowed toggle

marler8997:
Adds a toggle for borderless windowed mode to the core_window_flags example with the `[B]` key.
image
image


raysan5:
@marler8997 nice! thanks for the review!

PR #4870, opened 2025-03-31
Add Android Target to build.zig

lumenkeyes:
Added an Android target to the zig build script, mostly adapting existing work from the c Makefile.
This allows a user to easily build with only Android-supported libraries.
Library paths are automatically inferred as much as possible. Unfortunately, that also means a lot of ugly runtime string concatenation lol.

Tested on Linux and Windows. 
Also: my first PR!


raysan5:
@lumenkeyes thanks for the addition!

PR #4871, opened 2025-04-01
small build.zig fix

lumenkeyes:
Fixes an issue (that I created, sorry) with the generated `android-libc.txt` file when using raylib as a dependency in another zig project. Tested on Linux and Windows.


raysan5:
@lumenkeyes Thanks for the review!

Issue #4872, opened 2025-04-01
[rcore][SDL2] First touch position gets overwritten with mouse position when touch is not moved


zet23t:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

The touch position of first touch point is seemingly overwritten by some other coordinate when the touch point (=my finger) is not moving. When multiple touch points exists, the coordinates of the other touch points are correct. I modified the rcore_desktop_sdl.c file with logging to illustrate the issue:

Project code:

```c
    for (int i=0; i < GetTouchPointCount(); i++)
    {
        Vector2 pointerPosition = GetTouchPosition(i);
        TraceLog(LOG_INFO, "Touch %f %f %d %d", pointerPosition.x, pointerPosition.y,
            GetTouchX(), GetTouchY());
    }
```

Modified rcore_desktop_sdl.c file for debugging, in UpdateTouchPointSDL function:

```c
    CORE.Input.Touch.pointCount = SDL_GetNumTouchFingers(event.touchId);

    for (int i = 0; i < CORE.Input.Touch.pointCount; i++)
    {
        SDL_Finger *finger = SDL_GetTouchFinger(event.touchId, i);
        CORE.Input.Touch.pointId[i] = finger->id;
        CORE.Input.Touch.position[i].x = finger->x*CORE.Window.screen.width;
        CORE.Input.Touch.position[i].y = finger->y*CORE.Window.screen.height;
        TRACELOG(LOG_WARNING, "CORE.INPUT touch index %i: id=%i x=%f y=%f", i, finger->id, finger->x*CORE.Window.screen.width, finger->y*CORE.Window.screen.height);
        CORE.Input.Touch.currentTouchState[i] = 1;
    }
```

Output:

```
WARNING: CORE.INPUT touch index 0: id=14863 x=681.000000 y=427.000000
WARNING: CORE.INPUT touch index 1: id=14864 x=692.000000 y=251.999985
WARNING: CORE.INPUT touch index 0: id=14863 x=681.000000 y=427.000000
WARNING: CORE.INPUT touch index 1: id=14864 x=692.000000 y=251.999985
INFO: Touch 681.000000 427.000000 681 427
INFO: Touch 692.000000 251.999985 681 427
[...]

PR #4873, opened 2025-04-01
[rcore][SDL2] First touch position is overwritten with mouse pos

zet23t:
I believe it makes sense to only do this when there are no known touch points. I am not even sure if this should be done at all.

See https://github.com/raysan5/raylib/issues/4872 for more information on the bug this attempts to fix.


raysan5:
@zet23t thank you very much for the review, looks good to me

PR #4879, opened 2025-04-07
[cmake] set correct description for verbose option

brccabral:
Fixed wrong copy/paste


raysan5:
@brccabral thanks for the review!

PR #4881, opened 2025-04-09
[examples] Add 3D sound spatialization example

Bigfoot71:
This PR adds a basic example demonstrating 3D sound spatialization.

The example includes:  
- A function `SetSoundPosition(Camera listener, Sound sound, Vector3 position, float maxDist)`  
- A sphere rotating around the scene's origin, serving as the sound source  
- A 3D camera in `CAMERA_FREE` mode  

The `SetSoundPosition` function updates the sound's panning based on its position relative to the camera's position and orientation.

It also applies attenuation based on the distance to the listener, the emission radius (`maxDist`), and includes additional damping if the sound source is located behind the listener.

___

This follows up on the discussion in the wishlist thread for raylib 6.0: https://github.com/raysan5/raylib/discussions/4660#discussioncomment-12777056


Bigfoot71:
I set the difficulty to 2 stars, but I'm really not sure.

With some basic high school-level math, it might actually be just 1 star, I think.

raysan5:
@Bigfoot71 Hey! That's a really good example! Actually, there are not many audio examples, new examples like this one are very welcome!

Would you mind adding the example to build systems following the [`examples_template`](https://github.com/raysan5/raylib/blob/master/examples/examples_template.c) requirements?

Bigfoot71:
I updated the makefiles and the README.

As for VS2022, I can't do it right away. I checked if it can be done manually, but it looks like there are some hashes generated in `raylib.sln`

I'll take care of it as soon as I can!

raysan5:
@Bigfoot71 Thanks for the review! I think you can just set a random hash for the new VS2022 project to be added to solution.

raysan5:
@Bigfoot71 I'm merging it, I will review the missing parts.

PR #4887, opened 2025-04-11
build.zig: run zig fmt

marler8997:
Formats `build.zig` with `zig fmt build.zig`.  It was formatted before but fell out with the recent changes to add android.


raysan5:
@marler8997 oh! very nice! thanks for the review!

Issue #4888, opened 2025-04-12
[rmodels] `UpdateModelAnimationBones()` seems to apply scale animation incorrectly


mUnicorn:
Hello!
First of all, thank you for amazing library, and all the hard work supporting it.
Second, I'm pretty inexperienced in 3d related programming, so my assumptions might be off, please feel free to correct me if i'm wrong.

### Issue description

It appears that the scaling in [UpdateModelAnimationBones](https://github.com/raysan5/raylib/blob/master/src/rmodels.c#L2268C6-L2268C31) applied incorrectly. 

There's two observations:
#### No. 1
The scale of a bone applied without relation to bone rotation.

I have an example, where the animation is done in blender. Animation scales the cube with the bone horizontally (by blender's global y axis).

<img width="409" alt="Image" src="https://github.com/user-attachments/assets/ac86af9b-6eae-4c4d-83c1-742988e2c3d3" />

When exported as glb and imported in raylib project, essentially by doing:
```c
Model model = LoadModel(filename);
// ...
ModelAnimation * animations = LoadModelAnimations(filename, &animation_count);
// ...
UpdateModelAnimationBones(model, animations[0], animation_frame);
// ...
DrawModel(model, (Vector3){0.0f, 0.0f, 0.0f}, 1.0f, WHITE);
```
 the scale animation appears to be vertical

<img width="410" alt="Image" src="https://github.com/user-attachments/assets/b150f123-b74b-4868-a097-4d6b5b531fb8" />

(BTW I've tested glb in [gltf viewer](https://gltf-viewer.donmccurdy.com/), where animation appears as expected)

By patching the [transformation part of the UpdateModelAnimationBones](https://github.com/raysan5/raylib/blob/master/src/rmodels.c#L2292-L2312), applying transformations in a more dumb way:
```c
            // 1. Calculate bind pose matrix
            Transform * bind_transform = &model.bindPose[boneId];
            Matrix bind_matrix = 
              MatrixMultiply(
                MatrixMultiply(
                  MatrixScale(bind_transform->scale.x, bind_transform->scale.y, bind_transform->scale.z),
                  QuaternionToMatrix(bind_transform->rotation)
                ),
                MatrixTranslate(bind_transform->translation.x, bind_transform->translation.y, bind_transform->translation.z)
              );

            // 2. Calculate animation pose matrix
            Transform * target_transform = &anim.framePoses[frame][boneId];
            Matrix target_matrix =
              MatrixMultiply(
[...]

PR #4896, opened 2025-04-16
[rmodels] Fix bones animation scale

mUnicorn:
As discussed in #4888 this fixes scale-related issues during skeletal animation updates. 
Could you, please, take a look


raysan5:
@mUnicorn Thanks for the review, it looks ok to me but this change should be tested with several models and configurations to make sure it works on all cases...

mUnicorn:
@raysan5 Thank you! If you like, I can test some cases, if you provide some guidelines.

PR #4898, opened 2025-04-17
[RAUDIO] Better computation of the default audio stream buffer size

JeffM2501:
The audio stream subbuffer size was computed based on the device sample rate, but not the device sample size and channels.
this PR uses that data to compute the buffer size so we get a better default that is not too small on devices with more than 8 bits per sample.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -2104,8 +2104,12 @@ AudioStream LoadAudioStream(unsigned int sampleRate, unsigned int sampleSize, un
     // The size of a streaming buffer must be at least double the size of a period
     unsigned int periodSize = AUDIO.System.device.playback.internalPeriodSizeInFrames;
 
-    // If the buffer is not set, compute one that would give us a buffer good enough for a decent frame rate
-    unsigned int subBufferSize = (AUDIO.Buffer.defaultSize == 0)? AUDIO.System.device.sampleRate/30 : AUDIO.Buffer.defaultSize;
+    // If the buffer is not set, compute one that would give us a buffer good enough for a decent frame rate at the device bit size/rate
+    int deviceBitsPerSample = AUDIO.System.device.playback.format;
+    if (deviceBitsPerSample > 4)  deviceBitsPerSample = 4;
+    deviceBitsPerSample *= AUDIO.System.device.playback.channels;
+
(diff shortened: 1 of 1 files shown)


src/raudio.c
```diff
@@ -2104,8 +2104,12 @@ AudioStream LoadAudioStream(unsigned int sampleRate, unsigned int sampleSize, un
     // The size of a streaming buffer must be at least double the size of a period
     unsigned int periodSize = AUDIO.System.device.playback.internalPeriodSizeInFrames;
 
-    // If the buffer is not set, compute one that would give us a buffer good enough for a decent frame rate
-    unsigned int subBufferSize = (AUDIO.Buffer.defaultSize == 0)? AUDIO.System.device.sampleRate/30 : AUDIO.Buffer.defaultSize;
+    // If the buffer is not set, compute one that would give us a buffer good enough for a decent frame rate at the device bit size/rate
+    int deviceBitsPerSample = AUDIO.System.device.playback.format;
+    if (deviceBitsPerSample > 4)  deviceBitsPerSample = 4;
```
raysan5, line 2109: @JeffM2501 Is that correct? It is limiting the bitsPerSample to 4.

src/raudio.c
```diff
@@ -2104,8 +2104,12 @@ AudioStream LoadAudioStream(unsigned int sampleRate, unsigned int sampleSize, un
...
+    deviceBitsPerSample *= AUDIO.System.device.playback.channels;
+
+    unsigned int subBufferSize = (AUDIO.Buffer.defaultSize == 0) ? (AUDIO.System.device.sampleRate / 30 * deviceBitsPerSample) : AUDIO.Buffer.defaultSize;
```
raysan5, line 2112: The / and * shouldn't have spaces around.

raysan5:
@JeffM2501 Just added some review comments after merging...

src/raudio.c
JeffM2501: yes, the enum uses value 4 and 5 to differentiate between 32 bit ints and floats, both formats are 4 bytes pers sample. there is no 5 byte sample size, 4 is the max. Stereo will have 2 channels and get handled by code after this.

PR #4900, opened 2025-04-17
[Examples] Remove duplicate resource file

JeffM2501:
The gltf example had a duplicate res file in it's project and would not build, this PR removes the extra file from the build so there is not a duplicate icon resource.


raysan5:
@JeffM2501 Good catch! Actually, checking it more carefully, I think the issue is the other way round, I added that `examples.rc` to avoid linking with the resource defining raylib library data for the application... I'm changing all projects references!

PR #4878, opened 2025-04-07
[rlgl] Preserve current mode and textureId when switching mode or texture

brccabral:
Related to #4874 , #4347 , #4364 
I tested both `textures_polygon` and `textures_bunnymark`, they both work. I tested some others too, didn't see any problem, but feel free to comment.

Issue #4874, opened 2025-04-02
[rlgl] Textures not working when using `RL_TRIANGLES`


raysan5:
As per example `textures_poligon.c` it should [work as expected](https://www.raylib.com/examples/textures/loader.html?name=textures_polygon).

Texture should be assigned to vertex triangles.

This issue could be related to some recent changes that broke batching system and were reverted here: https://github.com/raysan5/raylib/commit/46cd07d2c74742fc7fbd9b0f2fb25cd78cd81b4c

Possible review steps: Compare raylib 5.5 `rlgl` to current master.

raysan5:
One solution is reordering the `SetTexture()` because `rlBegin()` resets the current texture to default on mode changing...
```c
    rlBegin(RL_TRIANGLES);
        rlSetTexture(texture.id);
        rlColor4ub(tint.r, tint.g, tint.b, tint.a);

        for (int i = 0; i < pointCount - 1; i++)
        {
            rlTexCoord2f(0.5f, 0.5f);
            rlVertex2f(center.x, center.y);

            rlTexCoord2f(texcoords[i].x, texcoords[i].y);
            rlVertex2f(points[i].x + center.x, points[i].y + center.y);

            rlTexCoord2f(texcoords[i + 1].x, texcoords[i + 1].y);
            rlVertex2f(points[i + 1].x + center.x, points[i + 1].y + center.y);
        }
    rlEnd();
```

brccabral:
I don't have a solution, but I noticed that both `rlSetTexture` and `rlBegin` advances `RLGL.currentBatch->drawCounter` if `rlCheckRenderBatchLimit`. Maybe only one of them should do it...

sleeptightAnsiC:
> This issue could be related to some recent changes that broke batching system and were reverted here: https://github.com/raysan5/raylib/commit/46cd07d2c74742fc7fbd9b0f2fb25cd78cd81b4c

I bisected between [5.5](https://github.com/raysan5/raylib/commit/c1ab645ca298a2801097931d1079b10ff7eb9df8) and https://github.com/raysan5/raylib/commit/1c4aa1378f61e24b2699b24ce05d5a16db64f380, and it showed me the same commit:
```
46cd07d2c74742fc7fbd9b0f2fb25cd78cd81b4c is the first bad commit
commit 46cd07d2c74742fc7fbd9b0f2fb25cd78cd81b4c (HEAD)
Author: Ray <raysan5@gmail.com>
[...]

PR #4905, opened 2025-04-20
[rlgl] introduced rlStartBatch

brccabral:
I was thinking about #4878 , in the solution I had to call `rlSetTexture` before _every_ `rlBegin`, therefore I though that they should merge into one call and avoid problems in the future, so I created `rlStartBatch`.  
I think that if I use `rlBegin` I would forget `rlSetTexture`, that's why I created `rlStartBatch`, so I don't forget to set both _mode_ and _textureId_.  
`rlEndBatch` only exists because sometimes there are calls to `rlSetTexture(0)` (maybe merge with `rlEnd()`?).  
   
I tested bunnymark in RenderDoc, it looks fine (60 fps with max objects).  
  
So, two options  
1. PR #4878 , and tell users to **always** call `rlSetTexture/rlBegin` together
2. this PR, tell users to call `rlStartBatch` and maybe deprecate `rlSetTexture/rlBegin` in the future...

### DIFF
--- a/examples/models/models_draw_cube_texture.c
+++ b/examples/models/models_draw_cube_texture.c
@@ -103,8 +103,6 @@ void DrawCubeTexture(Texture2D texture, Vector3 position, float width, float hei
     float y = position.y;
     float z = position.z;
 
-    // Set desired texture to be enabled while drawing following vertex data
-    rlSetTexture(texture.id);
 
     // Vertex data transformation can be defined with the commented lines,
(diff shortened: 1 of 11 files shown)


raysan5:
@brccabral I prefer to avoid that addition, raylib follows the Begin/End convention on drawing and I prefer to keep it as is.

brccabral:
If the function name is the issue I can just rename it, the idea is more important, to not forget to set `mode` and `textureId` at the begin to avoid bug #4874 , but it is your call. Saludos

PR #4906, opened 2025-04-22
[rlgl] Proposed fix for default near/far clipping range

Bigfoot71:
I've noticed the default near and far clipping planes were recently changed, and I'd like to suggest a fix

The current range `[0.001, 10000.0]` feels excessive and hard to justify

The near plane is too close, which can cause depth buffer precision issues and visual artifacts, while a far value of `10000.0` seems unnecessarily high for a default here

If the intent was to increase the far plane, a more reasonable default would be `[0.05, 4000.0]`, offering a better balance between depth precision and visible range

For special cases, `rlSetClipPlanes` already exists to override these defaults anyway

___

Here is a visual example using shadow mapping, where the scene depth (from the camera's point of view) was projected with a near/far range of `[0.001, 10000.0]`:

https://github.com/user-attachments/assets/1633d1d6-1e55-417f-b608-d25a2196bd18

And here is an example with `[0.05, 4000.0]`:

https://github.com/user-attachments/assets/a21d169f-0855-434c-b8ef-83c9b8e704ee

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -56,8 +56,8 @@
 *
 *       #define RL_MAX_MATRIX_STACK_SIZE             32    // Maximum size of internal Matrix stack
 *       #define RL_MAX_SHADER_LOCATIONS              32    // Maximum number of shader locations supported
-*       #define RL_CULL_DISTANCE_NEAR             0.05     // Default projection matrix near cull distance
-*       #define RL_CULL_DISTANCE_FAR            4000.0     // Default projection matrix far cull distance
+*       #define RL_CULL_DISTANCE_NEAR              0.05    // Default projection matrix near cull distance
+*       #define RL_CULL_DISTANCE_FAR             4000.0    // Default projection matrix far cull distance
 *
 *       When loading a shader, the following vertex attributes and uniform
 *       location names are tried to be set automatically:
@@ -237,7 +237,7 @@
     #define RL_CULL_DISTANCE_NEAR                 0.05      // Default near cull distance
 #endif
 #ifndef RL_CULL_DISTANCE_FAR
-    #define RL_CULL_DISTANCE_FAR                4000.0     // Default far cull distance
+    #define RL_CULL_DISTANCE_FAR                4000.0      // Default far cull distance
 #endif
 
 // Texture parameters (equivalent to OpenGL defines)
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 note that near/far values were reviewed recently to work on Orange Pi with PLATFORM_DRM, when I tried with lower ranges I got a lot of clipping in the platform due to the depth buffer bit-limitations. Also increases depth buffer requested to be at least 24 bit. Have you tried this fix in some low-end OpenGL ES device?

Bigfoot71:
@raysan5 Ah, was it a fix for 16-bit buffers?

Why not apply this range of `0.001/10000.0` based on the default depth buffer precision?
Because it might also become problematic in the long run for even more cases

I don't have a device that only supports 24-bit depth buffers, but I can try forcing the creation of a 16-bit buffer and see.

Bigfoot71:
@raysan5 I just ran some tests with a 16-bit depth buffer set via `glfwWindowHint(GLFW_DEPTH_BITS, 16);` using GLFW

Here is an example with the current planes `0.001 / 10000.0`:

https://github.com/user-attachments/assets/ac9b11b7-1eb8-4384-8835-f796bd3a5777

And here is an example with `0.05 / 4000.0` as suggested:

https://github.com/user-attachments/assets/140fc88f-f1f6-4628-bbfa-9762aee7e251

The scene consists of a 1000x1000 plane and a 1x1x1 cube

raysan5:
@Bigfoot71 thanks for the further review, I'm merging those new values, let's hope it works on most devices. Now raylib defaults to 24 bits depth buffers so it should be good on most devices...

This improvement maybe also addresses https://github.com/raysan5/raylib/issues/4917

Bigfoot71:
@raysan5 What I proposed will definitely work on a 24-bit depth buffer as well, and even offers better stability than the _old old_ `0.01 / 1000.0`, with the added benefit of a much farther far plane.

If any other issues are reported regarding this, it should be possible to handle it adaptively during `rlgl` initialization with something like:

```c
GLint depthBits;
glGetIntegerv(GL_DEPTH_BITS, &depthBits);

switch (depthBits) {
case 16:
    rlSetClipPlanes(0.5, 400.0);
    break;
case 24:
case 32:
    rlSetClipPlanes(0.05, 4000.0);
    break;
}
```

PR #4907, opened 2025-04-22
[example] Closes #4901, Fix Problems in 'text_draw_3d.c'

### LINKED ISSUE #4901: [examples] BaseSize is divided twice in example 'draw 3d' in 'text'
I was trying to use code of DrawText3D() and MeasureText3D() in 'draw 3d' example in 'text' part like following and notice that the size of the text is significantly smaller then DrawText(). Also, the X, Z value as return value in MeasureText3D() was different to MeasureTextEx().

This is the native code in example copied from DrawTextCodepoint3D():

    int index = GetGlyphIndex(font, codepoint);
    float scale = fontSize/(float)font.baseSize;

    // Character destination rectangle on screen
    // NOTE: We consider charsPadding on drawing
    position.x += (float)(font.glyphs[index].offsetX - font.glyphPadding)/(float)font.baseSize*scale;
    position.z += (float)(font.glyphs[index].offsetY - font.glyphPadding)/(float)font.baseSize*scale;

    // Character source rectangle from font texture atlas
    // NOTE: We consider chars padding when drawing, it could be required for outline/glow shader effects
    Rectangle srcRec = { font.recs[index].x - (float)font.glyphPadding, font.recs[index].y - (float)font.glyphPadding,
                         font.recs[index].width + 2.0f*font.glyphPadding, font.recs[index].height + 2.0f*font.glyphPadding };

    float width = (float)(font.recs[index].width + 2.0f*font.glyphPadding)/(float)font.baseSize*scale;
    float height = (float)(font.recs[index].height + 2.0f*font.glyphPadding)/(float)font.baseSize*scale;

Notice that value added in position X and Z, it's _OringinalSize / BaseSize * ScaleFactor_, aka _OringinalSize / BaseSize * FontSize / BaseSize_ since _ScaleFactor_ defined in line two is _FontSize / BaseSize_, therefore the basesize is not supposed to be divided again I guess.

The result became normal on my test case after removing the second division.

ZeanKey:
Fixed multiple problems mentioned in #4901 , now the _DrawText3D()_, _MeasureTextWave3D(_), _DrawTextWave3D()_ won't be affected by the base size of the font, which will show same effect as directly invoke DrawText() in 3D drawing. 

This could be the **expected effect** of the 'text_draw_3d.c' as the creator of the example mentioned that they suppose _MeasureTextEx()_ could work but didn't:
```
// Measure a text in 3D. For some reason `MeasureTextEx()` just doesn't seem to work so i had to use this instead.
```
As the unexpected scale is removed now _MeasureTextEx()_ can normally function in program so I replaced _MeasureText3D()_ with it.

Also, the "\~\~" markdown as command in wave text used to be counted for space and calculate character number in line in typer, which is now fixed by let counter only add up in characters not '\n' or '\~' in "\~\~".

The line space and space now would not be scaled by scale factor as these arguments works in 2D text functions, and removed duplicated accumulation for space in _MeausreTextWave3D()_.

### DIFF
--- a/examples/text/text_draw_3d.c
+++ b/examples/text/text_draw_3d.c
@@ -107,7 +107,7 @@ int main(void)
 
     // Use the default font
     Font font = GetFontDefault();
-    float fontSize = 8.0f;
+    float fontSize = 0.8f;
     float fontSpacing = 0.5f;
     float lineSpacing = -1.0f;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@ZeanKey Thanks for the review! I'm not fully convinced if this should be the expected behaviour but the truth is that it looks good and seems more aligned to raylib 2d text behaviour.

Issue #4908, opened 2025-04-25
[rcore] Wayland: GLFW framebuffer scale is handled incorrectly


gfaster:
Please, before submitting a new issue verify and check:

 - [X] I tested it on latest raylib version from master branch
 - [X] I checked there is no similar issue already reported
 - [X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [X] My code has no errors or misuse of raylib

### Issue description

This is effectively a followup to #4815.

GLFW 4.3 now handles fractional scaling correctly by default, namely changing `GLFW_COCOA_RETINA_FRAMEBUFFER` to `GLFW_SCALE_FRAMEBUFFER` and enabling it by default. much of the handling the Raylib to handle when screen size doesn't equal render size is redundant and breaks rendering.

I'm including a workaround that goes back to pre-4.3 behavior. I also wrote a more complete fix, but I don't have the means to test it, so I figured going back to old behavior would be preferable for now.

### Environment

```
OS: NixOS 25.05pre785698.b024ced1aac2 (Warbler)
Kernel: Linux 6.13.11
Display (LG HDR 4K): 3840x2160 @ 60 Hz (as 2259x1271) in 27"
Display (XB271HU): 2560x1440 @ 144 Hz (as 2229x1253) in 27"
WM: KWin (Wayland)
CPU: Intel(R) Core(TM) i7-7700 (8) @ 4.20 GHz
GPU: NVIDIA GeForce RTX 2080 Rev. A
```
```
INFO: GL: Supported extensions count: 395
INFO: GL: OpenGL device information:
INFO:     > Vendor:   NVIDIA Corporation
INFO:     > Renderer: NVIDIA GeForce RTX 2080/PCIe/SSE2
INFO:     > Version:  3.3.0 NVIDIA 570.133.07
INFO:     > GLSL:     3.30 NVIDIA via Cg compile
```

Cmake command:
```
cmake .. -DUSE_EXTERNAL_GLFW=ON -DBUILD_EXAMPLES=OFF -DCUSTOMIZE_BUILD=ON -DBUILD_SHARED_LIBS=ON
```

### Issue Screenshot

Current behavior is identical to the screenshot in #4815 

![Image](https://github.com/user-attachments/assets/48d65ad1-0bbe-4d37-bf01-c41c2e4043b3)

### Code Example

```c
#include "raylib.h"
#include <GLFW/glfw3.h>
#include <stdio.h>

[...]

PR #4909, opened 2025-04-25
Revert GLFW_SCALE_FRAMEBUFFER to pre GLFW 4.3 behavior

gfaster:
workaround for #4908

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1381,6 +1381,12 @@ int InitPlatform(void)
     if ((CORE.Window.flags & FLAG_WINDOW_TRANSPARENT) > 0) glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_TRUE);     // Transparent framebuffer
     else glfwWindowHint(GLFW_TRANSPARENT_FRAMEBUFFER, GLFW_FALSE);  // Opaque framebuffer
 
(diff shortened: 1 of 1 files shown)


raysan5:
@gfaster Thanks for the review. Shouldn't it be adapted to GLFW 4.3 current behaviour instead of patching it?

gfaster:
>  Shouldn't it be adapted to GLFW 4.3 current behaviour instead of patching it?

@raysan5 It probably should. I [played around with doing that](https://github.com/gfaster/raylib/commit/33752bea059e3bbfb368aee135ff5d25dbfa5d13), but I had to make changes to `SetupViewport` in `rcore.c` which I don't know how it would affect the other backends. It will probably require changing them to make sure everything stays consistent in addition to revisiting the semantics of `FLAG_HIGHDPI` (even when it's off on X11, I think(?) KWin 6 may still resize the window under you). Personally, I believe it should probably be on by default since MacOS, Wayland, XWayland, and web(?) all have independent render resolution and window size.

 I don't have the knowledge of the other backends, the indented semantics of Raylib, nor the access to other supported platforms to make that change with any confidence, but I'll try to look into it a bit more.

andrewfunston:
Just a note, I'm seeing this PR cause [rendering issues](https://discord.com/channels/426912293134270465/434448301522157570/1412135621933273252) on my Mac build. 

As highlighted there, adding `SetConfigFlags(FLAG_WINDOW_HIGHDPI);` before initializing my window fixes it.

PR #4910, opened 2025-04-26
build.zig: fix building raylib for emscripten

marler8997:
build.zig doesn't support the examples for the emscripten OS but it does support the raylib library. However, the examples panic if emscripten is configured which prevents it from building the library. I've replaced this panic with a fail step to fix this.


raysan5:

PR #3362, opened 2023-09-30
Optimized rtext functions

DissolveDZ:
## Hello, I've discovered several things that can be improved in the raylib text functions which would be more efficient memory and performance-wise. <br>
Defining TextLength as strlen since strlen uses vector operations and <br> rewriting TextFormat to use vsnprintf instead of allocating a static 2-dimensional buffer <br> and just simple syntax improvements for better readability for several functions. <br>I already tested the new functions and they work totally fine with the current raylib examples

PR #4911, opened 2025-04-27
[rtext] Use libc for `TextLength()`, `TextCopy()`, `TextSubtext()` and `TextInsert()`

williewillus:


src/rtext.c
```diff
@@ -1497,20 +1495,13 @@ int TextCopy(char *dst, const char *src)
 {
     int bytes = 0;
 
-    if ((src != NULL) && (dst != NULL))
+    // strcpy is marked restrict, meaning src and dst must not alias.
+    // Attempt to defend against that, but this is not fully robust
+    // as someone could pass in two sub-portions of the same string.
```
williewillus, line 1500: note that this is also already a problem with the preexisting code.

williewillus:
This is a reduced version of #3362 that avoids all the stylistic changes of that PR, and also does not change `TextFormat`. The minimum possible changes are done to achieve the goal.

Test Plan:
Wrote a small test program outside of Raylib with my versions of the functions, and things look as I expect:

```c
#include <string.h>
#include <stdlib.h>
#include <stdio.h>

#define MAX_TEXT_BUFFER_LENGTH              1024

unsigned int TextLength(const char *text)
{
    unsigned int length = 0;

    if (text != NULL)
    {
        length = strlen(text);
    }

    return length;
}

int TextCopy(char *dst, const char *src)
{
    int bytes = 0;

    // strcpy is marked restrict, meaning src and dst must not alias.
    // Attempt to defend against that, but this is not fully robust
    // as someone could pass in two sub-portions of the same string.
    if ((src != NULL) && (dst != NULL) && (src != dst))
    {
        strcpy(dst, src);
        bytes = strlen(src);
    }

    return bytes;
}

const char *TextSubtext(const char *text, int position, int length)
{
    static char buffer[MAX_TEXT_BUFFER_LENGTH] = { 0 };
    memset(buffer, 0, MAX_TEXT_BUFFER_LENGTH);

    int textLength = TextLength(text);

    if (position >= textLength)
    {
        return buffer; //First char is already '\0' by memset
    }

    int maxLength = textLength - position;
    if (length > maxLength) length = maxLength;
    if (length >= MAX_TEXT_BUFFER_LENGTH) length = MAX_TEXT_BUFFER_LENGTH - 1;

    memcpy(buffer, text + position, length);
    buffer[length] = '\0';

    return buffer;
}

char *TextInsert(const char *text, const char *insert, int position)
{
    int textLen = TextLength(text);
    int insertLen = TextLength(insert);

    char *result = (char *)malloc(textLen + insertLen + 1);

    memcpy(result, text, position);
    memcpy(result + position, insert, insertLen);
    memcpy(result + position + insertLen, text + position, textLen - position);

    result[textLen + insertLen] = '\0';     // Make sure text string is valid!

    return result;
}

int main() {
  const char *e = "";
  const char *a = "foo";
  const char *b = "bars";

  printf("%u\n", TextLength(e));
  printf("%u\n", TextLength(a));
  printf("%u\n", TextLength(b));

  char buf[4] = { 0 };
  int ret = TextCopy(buf, a);
  printf("%d %s\n", ret, buf);

  printf("%s\n", TextSubtext(b, 1, 0));
  printf("%s\n", TextSubtext(b, 1, 10));

  char *j = TextInsert(a, b, 2);
  printf("%s\n", j);
  free(j);
  return 0;
}
```

Result:
```
0
3
4
3 foo

ars
fobarso
```

jestarray:
awesome, please merge this for 6.0 <3

src/rtext.c
JayLCypher: The function prototype of TextCopy should probably be marked restrict since strcpy already is. Alternative is to use memmove instead of strcpy to protect against aliasing.

Copy does implicitly signal that two non-overlapping memory regions should be used, however there is risk that some developer has used this function to copy into overlapping memory regions. Strcpy will break such code, whilst memmove will not.

Alternate suggested code:
```c
bytes = strlen(src);
memmove(dst, src, bytes);
```

JayLCypher:
Unless there is any specific reason to have the length be of unsigned int (4 bytes), the entire TextLength function could be deprecated since it's literally just manual strlen, missing out on all vectorization and intrinsic functions.

raysan5:
Those functions were created as custom alternatives to libc implementations, also for educational pourposes. There is no plan to replace them by libc implementations. Users wanting libc alternatives can use libc directly.

jestarray:
> Those functions were created as custom alternatives to libc implementations, also for educational pourposes. There is no plan to replace them by libc implementations. Users wanting libc alternatives can use libc directly.

https://github.com/raysan5/raylib/blob/63b988ade906adc7264a8bec8fccec5f47befb55/src/rtext.c#L1174
https://github.com/raysan5/raylib/blob/63b988ade906adc7264a8bec8fccec5f47befb55/src/rtext.c#L1310

I feel like `MeasreTextEx` and `DrawTextEx` should use the optimized version of `strlen`, because I personally call them in the hot gameloop... not to mention `<string.h>` is already a dependency of `rtext.c`

raysan5:
@jestarray I'm reviewing those ones.

JayLCypher:
> @jestarray I'm reviewing those ones...
> 
> EDIT: Sorry, I'm seeing that it will break consistency, `TextLength()` is actually used all around, no `strlen()`. Let me see if I can address it in some way, maybe with a flag.

You can flag the function as a const function pointer, but you may have to break a little ABI a bit since you used int as return value for measureText. Example: https://godbolt.org/z/jsG4zqs3b

Alternatively you can skirt with UB by casting the pointer to strlen as (int (*)(const char *)&strlen.
I'm not sure if return value is mentioned in the calling of a function by pointer, but the args are the same size in any case.

PR #4914, opened 2025-04-28
[raudio] fix: GetMusicTimePlayed returns incorrect time after restarting a paused track

Servall4:
Pausing the active track, then stopping it and playing it again results in the playback restarting from the beginning, while GetMusicTimePlayed incorrectly returns the previous playback time.


raysan5:
@Servall4 thanks for the review!

PR #4915, opened 2025-04-28
cmake: replace custom variable with built-in one

rael346:
Byproduct of #4913

Since the current version of CMake in `raylib` is 3.25, we can replace the custom `RAYLIB_IS_MAIN` with `PROJECT_IS_TOP_LEVEL` (available since 3.21).


raysan5:
@rael346 thanks for the review! 👍🙂

PR #4916, opened 2025-04-30
[examples] PR for #4904, calculate 'right' Vector3 to correctly perform audio positioning.

### LINKED ISSUE #4904: [examples] audio_sound_positioning example appears to have reversed left/right stereo

### Issue description

I noticed the new audio_sound_positioning demo after a `git pull` so gave it a try, and noticed that when I moved the camera to the center of the examples moving audio source, when it was on the left the audio came out of the right speaker (and visa versa) - I checked my own OS audio config and it is indeed configured correctly, so it would appear that the example has the outputs reversed.

### Environment

x86_64 Linux 6.15.0-desktop-0.rc2.2omv2590 - Not a rendering or GPU issue, though "4.6 (Compatibility Profile) Mesa 25.0." - Raylib example compiled with clang version 19.1.7 and GLFW, OpenGL 3.3 API, X11, and miniaudio backend

### Issue Screenshot

Not a visual issue ;)

### Code Example

Not a code example exactly as one can just run the audio_sound_positioning demo to verify, however I think that `Vector3 right` in the `SetSoundPosition` function is being calculated backwards: it's currently written as `Vector3Normalize(Vector3CrossProduct(forward, listener.up))` and I believe this should be `Vector3Normalize(Vector3CrossProduct(listener.up, forward))`.

daniel-abbott:
Changes line 111 from `Vector3 right = Vector3Normalize(Vector3CrossProduct(forward, listener.up));` to `Vector3 right = Vector3Normalize(Vector3CrossProduct(listener.up, forward));` which will correctly output positional audio via stereo.

Fixes #4904

### DIFF
--- a/examples/audio/audio_sound_positioning.c
+++ b/examples/audio/audio_sound_positioning.c
@@ -108,7 +108,7 @@ static void SetSoundPosition(Camera listener, Sound sound, Vector3 position, flo
     // Calculate normalized vectors for spatial positioning
     Vector3 normalizedDirection = Vector3Normalize(direction);
     Vector3 forward = Vector3Normalize(Vector3Subtract(listener.target, listener.position));
-    Vector3 right = Vector3Normalize(Vector3CrossProduct(forward, listener.up));
+    Vector3 right = Vector3Normalize(Vector3CrossProduct(listener.up, forward));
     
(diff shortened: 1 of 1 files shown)


raysan5:
@daniel-abbott thanks for the review!

PR #4918, opened 2025-05-01
[rText] Fix issues with default font being used before InitWindow in the image API.

JeffM2501:
A few issues were found with how the default font is loaded if it is used before InitWindow, with the Image API. This PR resolves them.

- Don't try to use the texture size when building up the source rects for the default font, it may not be loaded.
- Don't try to reload the default font if it's glyphs have already been loaded, this would leak memory on every call to ImageDrawText as it would reload and reallocate the buffers every time.
- If the font glyhps are loaded but the texture was not, just upload the atlas texture and return early, so we don't again leak memory.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -161,6 +161,10 @@ extern void LoadFontDefault(void)
 {
     #define BIT_CHECK(a,b) ((a) & (1u << (b)))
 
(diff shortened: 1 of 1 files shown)


raysan5:

PR #4921, opened 2025-05-04
[rcore] Some adjustments for `FilePathList`

Bigfoot71:
I'm just proposing a few adjustments to `LoadDirectoryFiles` and `UnloadDirectoryFiles`.

Originally, I only intended to add a null check for `file.paths` in `UnloadDirectoryFiles`.

But I also took the opportunity to remove a redundant call to `memset(path, 0, size);` in the `ScanDirectoryFiles` and `ScanDirectoryFilesRecursively` functions, since the array is already zero-initialized on the stack:

```c
    static char path[MAX_FILEPATH_LENGTH] = { 0 };
    memset(path, 0, MAX_FILEPATH_LENGTH);  //< redundant
```

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -3681,7 +3681,6 @@ void SetupFramebuffer(int width, int height)
 static void ScanDirectoryFiles(const char *basePath, FilePathList *files, const char *filter)
 {
     static char path[MAX_FILEPATH_LENGTH] = { 0 };
-    memset(path, 0, MAX_FILEPATH_LENGTH);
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Wait, I got it wrong, in `ScanDirectoryFilesRecursively`, the array is not static, so the `memset` is indeed unnecessary.
But in `ScanDirectoryFiles`, it *is* static, so the inconsistency is different.
I'll redo the PR accordingly.

PR #4922, opened 2025-05-04
[rcore] Some adjustments for `FilePathList`

Bigfoot71:
This PR adds a null check for `files.paths` before calling `free()` in `UnloadDirectoryFiles`

While at it, I fixed an inconsistency in `ScanDirectoryFilesRecursively` by making `path` static, just like in `ScanDirectoryFiles`

Previously, both functions called `memset(path, 0, size)`, but only `ScanDirectoryFiles` had `path` declared as `static`
In the recursive version, `path` was reallocated on the stack and zero-initialized twice on each call

This change aligns the recursive version with the apparent original intent

___

Review of my previous PR: https://github.com/raysan5/raylib/pull/4921

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2314,9 +2314,12 @@ FilePathList LoadDirectoryFilesEx(const char *basePath, const char *filter, bool
 // WARNING: files.count is not reseted to 0 after unloading
 void UnloadDirectoryFiles(FilePathList files)
 {
-    for (unsigned int i = 0; i < files.capacity; i++) RL_FREE(files.paths[i]);
+    if (files.paths != NULL)
+    {
+        for (unsigned int i = 0; i < files.capacity; i++) RL_FREE(files.paths[i]);
 
-    RL_FREE(files.paths);
+        RL_FREE(files.paths);
+    }
 }
 
 // Create directories (including full path requested), returns 0 on success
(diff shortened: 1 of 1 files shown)


Bigfoot71:
After double-checking, I’ve realized that functions like `UnloadImage` don’t perform NULL-checks either, whereas functions like `UnloadTexture` do a `if (texture.id > 0) rlUnloadTexture(texture.id);`

It’s a bit of a strange difference, was this intentional?

If so, then this PR would ultimately only be limited to the fix that makes `path` static in `ScanDirectoryFilesRecursively`

raysan5:
@Bigfoot71 Afaik, `free()` already takes care of NULL pointers in C, that's the reason I avoid performing those checks.

Bigfoot71:
> @Bigfoot71 `free()` already takes care of NULL pointers in C standard, that's the reason I avoid performing those checks.

@raysan5 Yes, but here we're looping through `file.paths`, which is why I thought it made sense to suggest this check.

Personally, I tend to always anticipate this kind of situation when a structure or an array is passed to a `free` function, just to avoid potential issues.

That said, I didn’t really think about it specifically for `UnloadImage`, it was more of a reflex. In this case is unlikely to ever cause a problem indeed.

Bigfoot71:
> @Bigfoot71 `free()` already takes care of NULL pointers in C standard, that's the reason I avoid performing those checks.

@raysan5 Sorry, that was a fix I noted a long time ago, and I had forgotten the reason behind it

The issue is that if we don’t check `file.paths`, it could lead to dereferencing a null pointer when iterating over it later

As for `UnloadImage`, I’m not sure what I was thinking at the time, I’m not in great shape these days, but yes, it’s safe to pass a null pointer to `free()`

raysan5:
> The issue is that if we don’t check file.paths, it could lead to dereferencing a null pointer when iterating over it later

Oh, I see what you mean, interesting that technically a pointer is the same as an array of pointers!

Thanks for the review!

PR #4923, opened 2025-05-05
[rmodels] Fix GLTF bone weight assignments for meshes that are parented to armature bones.

JeffM2501:
In the current GLTF import code, cgltf does not assign bone weights to meshes that do not have bone weight attributes but are parented to the bone. This case in the GLTF file means that every vertex in the mesh is 100% weighted to the parent bone, but since the GLTF import code leaves the bone weight and id lists empty, raylib will never animate these mehses. This is why in some models, things like hats and weapons will not animate.
You can see an example here, notice that the bow and hood from the model are not attached to any bones.
https://github.com/user-attachments/assets/68cf4192-bb37-44ff-a8fb-e561db3099f8

But if you open the same model in a simple GLTF viewer, those meshes do animate.

https://github.com/user-attachments/assets/10fcf63d-5a71-48d6-b8b2-6395943a43bc

To fix this, we need to detect this case, and add a bone assignment to the parent bone so that raylib can do it's vertex animation on the entire mesh.
This PR adds a check for this case (when we have a skeleton and a mesh has no attributes, and is attached to a bone node in the tree), and then allocates the correct bone ID and weight buffers.

You can see the same model now works in raylib with this fix.

https://github.com/user-attachments/assets/8a037cb4-fa53-4b1a-a8e0-9fc43592982e

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5815,15 +5815,17 @@ static Model LoadGLTF(const char *fileName)
 
             for (unsigned int p = 0; p < mesh->primitives_count; p++)
             {
+                bool hasJoints = false;
+
                 // NOTE: We only support primitives defined by triangles
                 if (mesh->primitives[p].type != cgltf_primitive_type_triangles) continue;
 
                 for (unsigned int j = 0; j < mesh->primitives[p].attributes_count; j++)
                 {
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 great improvement! thanks! 👍😄

PR #4925, opened 2025-05-05
[rmodels] support CPU animation in OpenGL 1.1

JeffM2501:
Currently when drawing meshes with GL 1.1 only the static vertex and normal buffers are used, this means that CPU animation can't work with GL 1.1.
This PR uses the animated vertex and normal buffers if they exist just like GL 2.2+ does, supporting animated models in GL1.1.
It's not going to be fast since it's all done CPU side, but that should be up to the user, not raylib.


raysan5:
@JeffM2501 Nice improvement! Thanks!

PR #4926, opened 2025-05-06
Add normals to DrawSphereEx

karl-zylinski:
DrawCube has normals, DrawSphere doesn't. So I added it to DrawSphere.

Note that DrawCylinder, DrawCapsule, DrawTriangle3D and DrawCircle3D still doesn't have normals. But drawing a sphere is common for testing lighting, so I thought it would be nice. It's cheap to do for a sphere since the non-scaled vertex position is also the vertex normal. It may not be as cheap for those other drawing functions.


raysan5:
@karl-zylinski thanks for the addition!

PR #4927, opened 2025-05-06
build.zig fixes for android targets

lumenkeyes:
Added some fixes to build.zig to correctly detect target abi and generate lib paths that android sdk expects


raysan5:
@lumenkeyes thanks for the improvement!

PR #4928, opened 2025-05-08
[rtext] Properly clean up the default font on unload

JeffM2501:
The default font is unloaded, but its fields are never set to null. This will cause problems if the window is opened again and something tries to reinitialize the default font. This PR resets the fields back to null on unload.


raysan5:
@JeffM2501 Good catch!

PR #4929, opened 2025-05-09
Create compile_flags.txt

Lmenuontop:
Compile_flags.txt for clangd and ccls language servers.


raysan5:
@Lmenuontop I prefer to avoid creating additional files to address languages not used by the repo.

PR #4933, opened 2025-05-10
[rcore] Proposal: Review and extend raylib's random generation utilities

Bigfoot71:
I'd just like to open a discussion about possible improvements to random number generation in raylib, which currently only provides the `GetRandomValue()` function

I've always found this limitation a bit unfortunate, especially considering that rprand is already available and could easily be adapted to efficiently generate floating-point numbers, as demonstrated here.

At the moment, generating a float between 0 and 1, which is a very common need, requires something like:
```c
float value = (float)GetRandomValue(0, 1000) / 1000.0f;
```

This involves integer modulo, floating-point division, a call to `abs`, subtraction, and two additions.

In contrast, we could introduce a `GetRandomFloat()` function that performs just a bit-shift, an OR operation, and a subtraction, resulting in better performance and accuracy.

This is just a suggestion, I don’t necessarily expect the PR to be accepted, but I’d be happy to start a conversation around this topic.

### DIFF
--- a/examples/audio/audio_module_playing.c
+++ b/examples/audio/audio_module_playing.c
@@ -50,11 +50,11 @@ int main(void)
     for (int i = MAX_CIRCLES - 1; i >= 0; i--)
     {
         circles[i].alpha = 0.0f;
-        circles[i].radius = (float)GetRandomValue(10, 40);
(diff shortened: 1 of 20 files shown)


raysan5:
@Bigfoot71 I'm afraid I prefer to keep it simple with the int values and let users manage float values if required.

Bigfoot71:
It's strange to keep it as is, it feels like an incomplete feature, but okay

PR #4936, opened 2025-05-12
build.zig fixes for android targets

lumenkeyes:
EGL was previously not being linked properly for Android targets


raysan5:
@lumenkeyes thanks for the review

PR #4937, opened 2025-05-12
[rmodels] Fix and improve `GenMeshTangents`

Bigfoot71:
I propose a nearly complete review of `GenMeshTangents`, which had several issues

Here is the list of changes:
- Support for indexed meshes (`mesh->indices`)
- Correct accumulation of tangents/bitangents per vertex
- Improved handling of degenerate cases (invalid UVs, null tangents -> fallback to orthogonal)
- More robust orthogonalization (Gram-Schmidt with fallback in case of collinearity)
- Bonus, some comments

The function was tested with R3D, here is a quick example "before / after" with IBL:

screenshot000

screenshot000

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -3608,16 +3608,16 @@ BoundingBox GetMeshBoundingBox(Mesh mesh)
 }
 
 // Compute mesh tangents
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Thanks for the improvement! That function has already gone through several redesigns and the truth is that it was not properly tested... Thanks!

PR #4938, opened 2025-05-13
[utils] fix const warning with SaveFileText callback

JeffM2501:
The SaveFileTextCallback does not take it's second argument as a const char*, this is not the same as the argument the caller passes in in SaveFileText function, this generates a warning in gcc.
```utils.c
../external/raylib-master/src/utils.c: In function 'SaveFileText':
../external/raylib-master/src/utils.c:416:43: warning: passing argument 2 of 'saveFileText' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  416 |             return saveFileText(fileName, text);
      |                                           ^~~~
../external/raylib-master/src/utils.c:416:43: note: expected 'char *' but argument is of type 'const char *'
```

This PR changes the callback signature to match. This should not change binary compatibility since const is not used at runtime. The argument should be const as the callback can not modify the text it is supposed to save.

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -3139,7 +3139,7 @@
           "name": "fileName"
         },
         {
-          "type": "char *",
+          "type": "const char *",
           "name": "text"
         }
       ]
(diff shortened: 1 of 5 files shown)


raysan5:

PR #4939, opened 2025-05-13
[rcore] add `SetEventWaitingTimeout()`

david-vanderson:
Proof of concept for the ability to wait for input events with a timeout.  This is a request from people using [dvui](https://github.com/david-vanderson/dvui) with raylib for variable framerate applications.

Is this something raylib would like?

If so I will work on adding the capability to the other platforms.  Any advice about the API or code is appreciated.

Should the function log something or error if the value is <= 0?  Is there an example this should be added to?


raysan5:
@david-vanderson I prefer to avoid adding this function, it is not available on all supported platforms and it implies additional globals. Same functionality can be managed on user side if really required.

PR #4941, opened 2025-05-16
[build][proposal] Yet another build system

TomasBorquez:
Hey I'm the author of [mate.h](https://github.com/TomasBorquez/mate.h) a build system for C in C, all you need is a compiler and the `mate.h` header library. In the files changed I made a quick example of how it would look like to compile raylib on linux with GLFW to keep it simple, integration with windows and other compilers is also possible but I wanted to keep the most simple example I could for this PR.

And I was wondering if I made a proper PR for this integration would it be accepted? Since it would make raylib way more accessible if no build system would be needed other than `gcc mate.c -o mate`, specially for beginners wanting to compile the project themselves or just to run examples, tell me what you think!

I already added a compilation of raylib on my [test suite](https://github.com/TomasBorquez/mate.h/tree/master/tests/07-raylib-source-code) if you are interested.

### DIFF
--- a/mate.c
+++ b/mate.c
@@ -1,0 +1,33 @@
+#define MATE_IMPLEMENTATION
+#include "mate.h"
+
+i32 main() {
+  StartBuild();
+  {
+    // Asume it will be GLFW and X11, can later get options from `String arg = Argument("target")`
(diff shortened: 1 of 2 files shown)


raysan5:
@TomasBorquez Thanks, looks nice but I prefer to avoid another build system, it implies maintenance cost and potential issues.

PR #4944, opened 2025-05-17
[examples] Basic pbr example fix

Pivok7:
When I was playing around with basic pbr example i noticed that chnging metallness and roughness values at these lines:
    car.materials[0].maps[MATERIAL_MAP_METALNESS].value = 0.0f;
    car.materials[0].maps[MATERIAL_MAP_ROUGHNESS].value = 0.0f;
did nothing so I decided to fix this. Now changing these values has an effect on the final image.

preview


raysan5:
@Pivok7 Nice improvement! Thanks! 👍😄

PR #4947, opened 2025-05-18
[rcore] Fixed bug in hovering and input for android

padmadevd:
raylib treats both touch and hover events as same in android platform. This causes problem with android devices with pen or stylus support.

IsMouseButtonPressed() and IsMouseButtonDown() are triggered just for hovering above the screen with stylus. And IsMouseButtonReleased() is not triggered. meaning, the mouse is always down. Because, raylib not check AMOTION_EVENT_ACTION_HOVER_EXIT to release the input.

So this PR adds code to ignore the Hovering Input completely.

The Code is Tested in my samsung A13 phone. It does not have pen or stylus support. But i used [Genymobile/scrcpy](https://github.com/Genymobile/scrcpy) to emulate hovering input.
The hovering is ignored successfully and in my phone it works perfectly without any side effects to other inputs.
I already tested another version of the code in my own game in a samsung tablet with stylus. But this new version is not tested in real device.

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -70,6 +70,16 @@ typedef struct {
     EGLConfig config;                   // Graphic config
 } PlatformData;
 
+typedef struct {
+    // Store data for both Hover and Touch events
+    // Used to ignore Hover events which are interpreted as Touch events
+    int32_t pointCount;                             // Number of touch points active
+    int32_t pointId[MAX_TOUCH_POINTS];              // Point identifiers
+    Vector2 position[MAX_TOUCH_POINTS];             // Touch position on screen
+
+    int32_t hoverPoints[MAX_TOUCH_POINTS];          // Hover Points
+} TouchRaw;
+
 //----------------------------------------------------------------------------------
 // Global Variables Definition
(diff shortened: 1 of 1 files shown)


raysan5:
@padmadevd Mmmh... I was not aware of that possible input state...

Please, first of all, could you review code conventions: https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md

padmadevd:
i really apologize for the mistake.
Now, i corrected the code to meet the coding conventions.

raysan5:
@padmadevd thanks! It looks good to me. Merging.

padmadevd:
i previously implemented a bool array of size 32 to store the hover state of the pointers. Because, chat GPT told me that the pointer ID returned by AMotionEvent_getPointerId(event, i) will be in the range of 0 to 31. Is it true?
If it is true, then i can directly get hover state using the pointer ID as the index to the hover states array.
This can avoid looping to check the state in the current code (But MAX_TOUCH_POINTS iterations only).
I m not sure about the fact chat GPT told. Because i can't find any official documentation stating that. But the previous code worked correctly. But i changed that for PR.

Issue #4940, opened 2025-05-15
[rcore_web] Mouse position is locked after calling HideCursor in PLATFORM_WEB


proc0:
### Issue description

I'm trying to show and hide the cursor for a pause menu and this works on desktop but when compiling for web if the cursor is hidden the mouse position does not update at all. (Side note: This [example](https://www.raylib.com/examples/core/loader.html?name=core_input_mouse) seems broken (pressing H does not hide cursor?) not sure if that's the same issue or not. I reproduced the issue [here](https://github.com/proc0/raylib-starter/blob/main/source/game.cpp) and it's the same as the example).

I expect the behavior of HideCursor and ShowCursor to be the same as Desktop (whether it locks the mouse or not).

### Environment

Raylib 5.5, PLATFORM_WEB, compiled on Windows11, MinGW and Emscripten

### Issue Screenshot

![Image](https://github.com/user-attachments/assets/7b1ade37-6e0a-4d6d-881f-17a6c6a7ac8e)

### Code Example

```
    if (IsKeyPressed(KEY_H)){
        if (isCursorHidden == 0)
        {
            HideCursor();
            isCursorHidden = 1;
        }
        else
        {
            ShowCursor();
            isCursorHidden = 0;
        }
    }
    
    ballPosition = GetMousePosition(); // gets stuck in same X,Y if HideCursor was called
```

### Workaround and comments

I think this is caused because CORE.Input.Mouse.cursorHidden is being used for both hiding the cursor and locking the cursor. I would expect that the DisableCursor would lock the mouse, and for HideCursor to not lock the mouse. But in rcore_web there is a check for cursorHidden before updating the mouse position.

![Image](https://github.com/user-attachments/assets/ef2448ce-a209-4535-9e59-3503d4d3949d)

I think there could be two flags here, maybe? cursorHidden and cursorLocked. and cursorHidden only gets set and unset from Hide/ShowCursor. I could be missing something as well and maybe this is intended behavior for Web (since hiding cursor automatically locks it?). 

The workaround I found is to use GetMouseDelta, but this only works if I use DisableCursor and EnableCursor instead of Hide and ShowCursor... which then makes me ask the question what exactly is the difference between Hiding and Disabling the Cursor? And also between Hiding and Locking (as used by Emscripten and GLFW).

raysan5:
> I think this is caused because CORE.Input.Mouse.cursorHidden is being used for both hiding the cursor and locking the cursor.

[...]

PR #4955, opened 2025-05-28
[rcore][web] Support mouse position update on `HideCursor()`

Moros1138:
Fixes Issue #4940

The problem is relatively simple, ``CORE.Input.Mouse.cursorHidden`` was being used for both hiding the cursor and locking the cursor. This PR adds a variable to track when the cursor is locked so that the 2 states can be tracked independently of eachother.

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -863,8 +863,11 @@ void EnableCursor(void)
 
     // Set cursor position in the middle
     SetMousePosition(CORE.Window.screen.width/2, CORE.Window.screen.height/2);
+    
+    // Hide cursor
+    ShowCursor();
 
-    // NOTE: isMouseLocked handled by EmscriptenPointerlockCallback()
+    // NOTE: lockedMouseCursor handled by EmscriptenPointerlockCallback()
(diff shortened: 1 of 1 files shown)


Moros1138:
Sorry for the after-editing, I compared the behavior with windows and made further changes to make the way web handles it closer to how it works on windows.. Here's a video

https://github.com/user-attachments/assets/6881cd6b-cae6-459f-8e6f-16d6ebcb8e69

Moros1138:
> @Moros1138 My concern with this change is if it will change examples behaviour between platforms... if that's the case it should be addressed on all platforms.

I was seeking to make the web platform work more as it does on windows. See video in previous post.

Moros1138:
Is there a blocker for this PR that I have missed?

raysan5:
@Moros1138 Just tested it and it breaks `DisableCursor()` behaviour, you can test `core_3d_camera_first_person`.

Please, be careful testing those PRs, there are many platforms and use cases involved and one solution for one of those could break other things.

raysan5:
Just reviewed required code to add a new state.

Issue #4951, opened 2025-05-21
[rcore][web] Pressing F1 opens a help tab in certain chromium based browsers


Moros1138:
### Issue description

In Microsoft Edge and Google Chrome I have noticed pressing F1 opens a help tab.

Repoduce in the affected browsers by following these steps:

1. Visit: https://www.raylib.com/examples/shapes/loader.html?name=shapes_top_down_lights
2. click anywhere in the running example to ensure it has focus
3. Press 'F1' key
4. See the problem.

I assume, given the pattern involves Chromium based browsers that this is true of other similarly based browsers. Firefox was unaffected.

### Environment

Windows 11
Microsoft Edge Version 136.0.3240.76 (Official build) (64-bit)
Google Chrome Version 136.0.7103.114 (Official Build) (64-bit)

### Possible Solution

I fixed this locally in my own build adding the following to my shell:

```
        document.addEventListener('keydown', function(event)
        {
            if (event.key == 'F1' || event.keyCode == 112)
                event.preventDefault();
        });
```

this could easily be integrated into the web platform code like so:

```
EM_ASM({
        document.addEventListener('keydown', function(event)
        {
            if (event.key == 'F1' || event.keyCode == 112)
                event.preventDefault();
        });
});
```

Probably somewhere in InitPlatform.

maiconpintoabreu:
Hi @Moros1138 you can avoid the help page holding Ctrl and pressing F1.

Moros1138:
@maiconpintoabreu thanks for the workaround! unfortunately this isn't desired let's say I've written an app the checks for Ctrl + F1, my web build would require special treatment because I can't count on the keyboard being handled the same way across the platforms.  However, my PR for this issue has been merged so I'm gonna close this issue as is appropriate.

raysan5:
@Moros1138 This PR broke text input on web and was reverted: https://github.com/raysan5/raylib/issues/5107

PR #4956, opened 2025-05-28
[rcore][web] Add `EmscriptenKeyboardCallback()` to consume key events

Moros1138:
There are times when we want to use keys that are otherwise used for browser functions.

F1 -> opens a help tab (in Chromium based browsers, see #4951)
F5 -> refresh the page

There are many more key/key combinations that trigger browser actions. This PR seeks prevent the default browser actions by consuming all keyboard events. This leaves us free to use any keyboard cominations as we would in any other environment.

This PR fixes issue #4951

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -137,6 +137,7 @@ static EM_BOOL EmscriptenFocusCallback(int eventType, const EmscriptenFocusEvent
 static EM_BOOL EmscriptenVisibilityChangeCallback(int eventType, const EmscriptenVisibilityChangeEvent *visibilityChangeEvent, void *userData);
 
 // Emscripten input callback events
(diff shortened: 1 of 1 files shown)


raysan5:
@Moros1138 I'm merging this change but I'm not fully convinced if this is the best/expected behaviour or the fact of consuming keys on raylib side could affect other systems... let's see...

Moros1138:
@raysan5 sorry for the delayed response.

I don't think it's unreasonable for a raylib user to expect their web builds to function the same way as it does on their windows builds.

Google Docs, Google Sheets, Youtube, Vimeo, Tiktok, X, Snapchat, Monaco Editor, Github Workspaces, and CodeMirror just to name a few off the top of my head that all override/consume keyboard events.

I realize "other projects do it" isn't a complete argument. I do hope these points serve to boost your confidence in your decision!

PR #4958, opened 2025-05-28
Remove unused variable

M374LX:
Nothing more than the removal of an unused variable from ``rcore.c``.


raysan5:

PR #4969, opened 2025-05-30
Update comments

M374LX:
This mostly replaces "axis" with "axes" where it makes sense.

There is still the ``GamepadAxis`` enum, which should be renamed to ``GamepadAxes``, but this is better left for another PR because it is not a comment.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1,6 +1,6 @@
 /**********************************************************************************************
 *
-*   rcore_desktop - Functions to manage window, graphics device and inputs
+*   rcore_desktop_glfw - Functions to manage window, graphics device and inputs
 *
 *   PLATFORM: DESKTOP: GLFW
 *       - Windows (Win32, Win64)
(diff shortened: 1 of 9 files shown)


src/platforms/rcore_desktop_glfw.c
```diff
@@ -1,6 +1,6 @@
 /**********************************************************************************************
 *
-*   rcore_desktop - Functions to manage window, graphics device and inputs
```
M374LX, line 3: This is due to this source file being previously named ``rcore_desktop.c`` (without ``_glfw``).

src/raylib.h
```diff
@@ -1192,8 +1192,8 @@ RLAPI bool IsGamepadButtonDown(int gamepad, int button);      // Check if a game
 RLAPI bool IsGamepadButtonReleased(int gamepad, int button);  // Check if a gamepad button has been released once
 RLAPI bool IsGamepadButtonUp(int gamepad, int button);        // Check if a gamepad button is NOT being pressed
 RLAPI int GetGamepadButtonPressed(void);                      // Get the last gamepad button pressed
-RLAPI int GetGamepadAxisCount(int gamepad);                   // Get gamepad axis count for a gamepad
-RLAPI float GetGamepadAxisMovement(int gamepad, int axis);    // Get axis movement value for a gamepad axis
```
M374LX, line 1196: The words ``gamepad`` and ``axis`` were duplicated in these lines.

raysan5:
@M374LX thanks for the further review!

PR #4970, opened 2025-05-30
[utils] Add colors to the trace log

sudoer-zero:


raysan5:
@sudoer-zero I'm afraid this solution is not portable to all supported raylib platforms so I prefer to keep the logs simple.

PR #4971, opened 2025-05-30
Rename GamepadAxis enum to GamepadAxes (breaking change)

M374LX:
A follow-up to the likes of 341bfb2 and 3418172.

This is a breaking change.


raysan5:
Other enums use the singular form, considering it is also abreaking change I prefer to keep it as is.

PR #4974, opened 2025-05-31
RGFW: fix Escape always closing the window

### LINKED ISSUE #4967: [rcore][rgfw] Pressing Escape always closes the window
When using the RGFW backend on the desktop platform, pressing Escape always closes the windows, even if a different key or ``KEY_NULL`` is set with ``SetExitKey()``.

In fact, ``RGFW_window_shouldClose()`` always checks for Escape.

Maybe ``WindowShouldClose()`` should stop relying on ``RGFW_window_shouldClose()``.

M374LX:
Fixes #4967.


raysan5:
@M374LX thanks for the review...

M374LX:
@raysan5 I have just noticed that the ability to change GLFW's exit key was added to a recent development version and is not yet present in 1.7.

We need to either revert this PR or update GLFW to the latest development version.

M374LX:
Do you agree with the idea of updating RGFW to the latest development version, knowing that it fixes some issues present in 1.7, instead of reverting the change?

PR #4980, opened 2025-06-01
[rshapes] Fix incorrect parameter names in DrawRectangleGradientEx

williewillus:
Examining the code shows that the rectangle is drawn winding counterclockwise, starting
with the top left. Therefore the colors used should be in the order: topLeft, bottomLeft,
bottomRight, topRight.

However, the variables actually being used are topLeft, bottomLeft, topRight,
bottomRight. I was confused by this as I was getting striping where I didn't expect any.
Put another way, the last two parameters are misnamed.

This diff swaps the parameter names and their usages. The result is that no runtime
behaviour changes: the same parameter order yields the same visual result both before and
after this change, but the parameter names now correctly reflect what they are actually
used for.

You can actually see this in the implementation of DrawRectangleGradientV, which
(correctly) passes top, bottom, bottom, top to DrawRectangleGradientEx.


raysan5:
@williewillus thanks for the review but note this change could be a breaking change for some users...

williewillus:
@raysan5 can you elaborate on that? My commit summary explains it: the runtime behaviour before and after this PR is exactly the same. The only thing is that the parameters have correct names now that match what they're actually used for.

raysan5:
@williewillus yeah, you are right, testing it with the raylib shapes example and raygui (that uses it internally), all looks good.

PR #4981, opened 2025-06-01
Add LICENSE to build.zig.zon

garrisonhh:
Zig users importing raylib need to have the license too.


raysan5:
@garrisonhh thanks for the review!

PR #4982, opened 2025-06-01
Change `RGFW_window_eventWait` timeout to -1 to fix `EnableEventWaiting`

LainLayer:
there is a mistake in rcore_desktop_rgfw line 992
currently it looks like this:
```c
if ((CORE.Window.eventWaiting) || (IsWindowState(FLAG_WINDOW_MINIMIZED) && !IsWindowState(FLAG_WINDOW_ALWAYS_RUN)))
{
    RGFW_window_eventWait(platform.window, 0); // Wait for input events: keyboard/mouse/window events (callbacks) -> Update keys state
    CORE.Time.previous = GetTime();
}
```

with `RGFW_window_eventWait` set to 0 timeout. this makes `EnableEventWaiting();` not work, since the timeout is always 0. setting it to -1 makes it work correctly.


raysan5:
@LainLayer thanks for the review!

PR #4988, opened 2025-06-05
[rcore][drm] Replace DRM swap buffer implementation with asynchronous page-flipping and triple framebuffer caching

rob-bits:
The original implementation created/destroyed framebuffers (FBs) per-frame, leading to kernel overhead and screen tearing. This commit replaces it with a different approach using:
- Asynchronous `drmModePageFlip()` with vblank sync
- Framebuffer caching to reduce repeated FB creation/removal operations
- Proper resource management through BO callbacks and buffer release synchronization
- Added error handling for busy displays, cache overflows, and flip failures
- Event-driven cleanup via page_flip_handler to prevent GPU/scanout conflicts

Would you please review this code?
In my case it works better, has less framedrop or tearing.
Tested it on a Raspberry Pi Zero 2W.

Maybe this implementation should be handled as a compiler option for the user. To select if he wants the minimal DRM or the page-flipping, cached DRM.

#### **Summary: Key Differences**

| **Aspect**               | Function (`SwapScreenBuffer - Original`)                     | Function (`SwapScreenBuffer new`)                              |
|--------------------------|---------------------------------------------------------------|-----------------------------------------------------------------|
| **Framebuffer Creation**  | Creates/removes an FB per frame (expensive).                 | Caches Fbs for reuse; minimizes kernel operations.               |
| **Flip Synchronization**  | No vblank sync: likely causes tearing.                       | Uses `drmModePageFlip()` with vblank sync to prevent tearing.    |
| **Error Handling**        | Minimal recovery (logs errors, continues).                    | Graceful error handling and tracking via counters/callbacks.     |
| **Resource Management**   | Frees buffers immediately; risks conflicts.                   | Properly releases BOs/Fbs only after flips complete.             |
| **Complexity/Overhead**   | Simple code but performs poorly at high frame rates.         | More complex but optimizes for performance      |

---

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -71,6 +71,7 @@
 #include "EGL/egl.h"        // Native platform windowing system interface
 #include "EGL/eglext.h"     // EGL extensions
 
+#if defined(SUPPORT_DRM_CACHE)
 #include <poll.h>  // for drmHandleEvent poll
 #include <errno.h> //for EBUSY, EAGAIN
 
@@ -85,6 +86,8 @@ static FramebufferCache fbCache[MAX_CACHED_BOS];
 static volatile int fbCacheCount = 0;
 static volatile bool pendingFlip = false;
 static bool crtcSet = false;
+
+#endif //SUPPORT_DRM_CACHE
 
 #ifndef EGL_OPENGL_ES3_BIT
     #define EGL_OPENGL_ES3_BIT  0x40
@@ -232,7 +235,6 @@ static const short linuxToRaylibMap[KEYMAP_SIZE] = {
 //----------------------------------------------------------------------------------
 // Module Internal Functions Declaration
 //----------------------------------------------------------------------------------
-int InitSwapScreenBuffer(void);
 int InitPlatform(void);          // Initialize platform (graphics, inputs and more)
 void ClosePlatform(void);        // Close platform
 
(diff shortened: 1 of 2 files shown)


src/platforms/rcore_drm.c
```diff
@@ -910,7 +1099,8 @@ int InitPlatform(void)
         EGL_BLUE_SIZE, 8,           // BLUE color bit depth (alternative: 5)
         EGL_ALPHA_SIZE, 8,        // ALPHA bit depth (required for transparent framebuffer)
         //EGL_TRANSPARENT_TYPE, EGL_NONE, // Request transparent framebuffer (EGL_TRANSPARENT_RGB does not work on RPI)
-        EGL_DEPTH_SIZE, 24,         // Depth buffer size (Required to use Depth testing!)
+        //ToDo: verify this. In 5.5 it is 16, in master it was 24
```
raysan5: Yes, that's correct, it was increased to address some z-fighting isssues. Afaik, most devies already support 24bit depth buffers.

raysan5:
@rob-bits Thanks for the improvement, it would be nice to keep both methods to be selected with a compilation flag.

Please, could you first review the code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

rob-bits:
> @rob-bits Thanks for the improvement, it would be nie to keep both methods to be selected with a compilation flag.
> 
> Please, could you first review the code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

Yeah, will review it and revisit the code if it brokes any convention. And will add a compiler define to have control which one should be used.

rob-bits:
Added compilation config flag and reimplemented some variable names to be compatible with coding conventions. The default code points to the original implementation. With additional testing, benchmarking later we can move on with this.

raysan5:
@rob-bits This is a great addition! Thanks! Merging and reviewing!

PR #4991, opened 2025-06-12
Remove depth texture from RenderTexture struct

M374LX:
It is created but not used, apparently.


raysan5:
@M374LX Depth texture is required.

PR #4992, opened 2025-06-12
Update RGFW

### LINKED ISSUE #4975: [rcore][rgfw] No key release on focus loss
With the RGFW backend, when the window loses focus while a key is still being held, the program behaves as if the key were still being held, even if it is actually released after the focus loss.

The issue is not present in the GLFW and SDL backends.

M374LX:
Update RGFW to commit 55073bc.

Fixes #4975.


raysan5:
@M374LX thanks for the update!

PR #4995, opened 2025-06-14
Allow passing options to raygui in build.zig

Not-Nik:
This fixes build issues when building raygui on Linux systems without wayland, that are missing the `wayland-scanner` program. More details on the issue can be found at Not-Nik/raylib-zig#243.

A different solution to this problem could be requiring `raylib_dep` to be passed into `addRaygui` directly, but I don't see any semantic difference, as the dependency functions should reuse previously created `std.Build.Dependency` instances.


raysan5:
@Not-Nik thanks for the review

Issue #4996, opened 2025-06-14
[rmodels] LoadModel leaks memory for .glb files with vertex colors


TatuLaras:
### Issue description

I was noticing some memory leaks that happen while loading / unloading models repeatedly.
I decided to investigate this issue and made a minimal reproduction of this bug (seen at the bottom of this issue). Increasing the iteration count in that code example proportionally leads to alarger maximum resident set. I compiled raylib with symbols and ran that example through asan, according to which the leak is from an allocation in line 5737 in `rmodels.c`:

```c
if (attribute->component_type == cgltf_component_type_r_8u)
{
    // Init raylib mesh color to copy glTF attribute data
=>  model.meshes[meshIndex].colors = (unsigned char *)RL_MALLOC(attribute->count*4*sizeof(unsigned char)); // !! OFFENDING LINE !!

    // Load 4 components of unsigned char data type into mesh.colors
    LOAD_ATTRIBUTE(attribute, 4, unsigned char, model.meshes[meshIndex].colors)
}
else if (attribute->component_type == cgltf_component_type_r_16u)
```

The memory leak seems to happen in relation to vertex color data.
Another fact that speaks to this is that the bug is not present when loading a .glb of a plain blender default cube, but memory starts being leaked as soon as it's exported with some vertex colors applied from the vertex paint mode.

`mesh.colors` is unloaded in `UnloadMesh`, commenting out that line leads to even more memory being leaked.
I tried to search for instances where a meshes color array pointer was changed in the course of the functions called in the code example, but couldn't find any.

### Environment

```
OS: Arch Linux
Kernel: 6.14.3-arch1-1
CPU: Intel i5-9600K (6) @ 4.700GHz
GPU: NVIDIA GeForce RTX 2070

OpenGL device information:
> Vendor:   NVIDIA Corporation
> Renderer: NVIDIA GeForce RTX 2070/PCIe/SSE2
> Version:  3.3.0 NVIDIA 570.144
> GLSL:     3.30 NVIDIA via Cg compiler
```

### Issue Screenshot

N/A

### Code Example
```c
#include "raylib.h"

int main(int argc, char **argv) {
    InitWindow(800, 450, "Test");

    Model model;

    for (int i = 0; i < 20; i++) {
        model = LoadModel("model_with_vertex_colors.glb");
        UnloadModel(model);
    }

    CloseWindow();

    return 0;
}
```

TatuLaras:
Narrowed it down a bit more. Changing this:
```c
[...]

PR #4998, opened 2025-06-15
[RMODELS] Ensure that GLTF buffers are deallocated before using them.

JeffM2501:
Check to see if a mesh buffer is allocated before reallocating it so we don't leak that buffer in cases where the gltf may have duplicate attributes at different resolutions.

Should fix issue #4996

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5532,6 +5532,10 @@ static Model LoadGLTF(const char *fileName)
 
                         // WARNING: SPECS: POSITION accessor MUST have its min and max properties defined
 
+                        // make sure we don't leak this buffer if there are multiple attributes of the same type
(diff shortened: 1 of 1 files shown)


raysan5:
I'm afraid I'm not adding all those checks, two reasons: 

1. Users are the owners of data and should be careful with it
2. If the pointers are freed but not zeroed, it will fail

JeffM2501:
we will have to make a different check then. This is a legitimate memory leak.
The other option is to skip loading the attribute if one of the same type has already been loaded.
but if we do not do anything the library will leak memory.
The duplicate attributes are added by the model exporters, and are not controllable by the end user, so raylib must deal with this situation in a valid way.

PR #5000, opened 2025-06-18
[rtext] Add DrawTextGradientV and DrawTextCodepointGradientV for drawing text with a vertical gradient

chrg127:
This pull request adds two functions: `DrawTextGradientV` and `DrawTextCodepointGradientV`. These respectively draw a string of text and a single codepoint, while applying to the text a vertical gradient described by two colors passed as parameters. They were written to be similar to the already present `DrawTextEx` and `DrawTextCodepoint`, with the addition of another `Color` parameter for each one.

Rationale: in my game I wanted to be able to draw text with vertical gradients just like those found in various commercial games. When I tried using raylib for this I found out that it doesn't have any function for this and that you're basically required to modify the library in a not-super-obvious way if you want even the simplest of text effects. I don't think newbies should be required to do that to emulate what a lot of games do...

Notes:
- To add these functions I had to write a new function, `_DrawTextureGradientV`. It is the same as `DrawTexturePro`, except it has a second `Color` parameter and no rotation parameters. This is a private function inside `rtext` now, but could be later put into `rtextures` if the need arises.
- With the parameter count at 7, `DrawTextGradientV` exceeds the limit of 6 parameters. I don't know how could I lower that limit, since I based it on `DrawTextEx`. `DrawTextCodepointGradientV` has a parameter count of 8, but that function can be removed if needed.
- There are no examples, and I didn't modify any exisiting ones.

### DIFF
--- a/parser/output/raylib_api.json
+++ b/parser/output/raylib_api.json
@@ -9288,6 +9288,41 @@
       ]
     },
     {
+      "name": "DrawTextGradientV",
+      "description": "Draw text with vertical gradient",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "Font",
+          "name": "font"
+        },
+        {
+          "type": "const char *",
+          "name": "text"
+        },
+        {
+          "type": "Vector2",
+          "name": "position"
+        },
+        {
+          "type": "float",
+          "name": "fontSize"
+        },
+        {
+          "type": "float",
+          "name": "spacing"
+        },
+        {
+          "type": "Color",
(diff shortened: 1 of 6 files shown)


raysan5:
@chrg127 I'm afraid I'm not adding those function to the API, I think they are higher-level and can be defined in user code. In any case, an example implementing those functions can be added as a reference for users with that specific need.

chrg127:
@raysan5 I understand, I'll try working on an example then.

PR #5002, opened 2025-06-19
[rcore] Fix ScanDirectoryFilesRecursively

mlorenc227:
Fixes a regression since 5.5, where `ScanDirectoryFilesRecursively` no longer does the "recursively" part due to `path` being `static`.

The issue was once already fixed in https://github.com/raysan5/raylib/commit/5530a3ceb88962066affa5db8e13b00b64444b37 but recently made it back it in.


raysan5:
@mlorenc227 Oh, good catch, I'm adding a comment to that line to remember!

PR #5004, opened 2025-06-23
[examples] New example: `core_3d_fps_controller`

nezvers:
FPS camera example when it has to follow a body's position.
Quake-like camera animations and strafe jump movement.
Simplified physics, just 0.f as floor.
Jump sound effect, created by me influenced by original quake sound.
Should be ready for emscriptem compilation.
Tested with Visual Studio from CMake at root folder.

I couldn't figure out how to create Visual Studio project for the example.


nezvers:
Is it OK, or do I need to improve something for it?

raysan5:
@nezvers Thanks for the example, it's actually longer than expected. I'll review it carefully, probably also renaming it.

Issue #4959, opened 2025-05-29
[rcore] Building with SDL fails without rtext


M374LX:
The ``rtext`` module is supposed to be optional. However, when building a program for the desktop platform using the SDL backend, it fails with the following message due to ``rcore_desktop_sdl.c`` requiring the ``GetCodepointNext()`` function:

```
/usr/bin/ld: ../src/libraylib.a(rcore.o): in function `PollInputEvents':
rcore.c:(.text+0x2d212): undefined reference to `GetCodepointNext'
collect2: error: ld returned 1 exit status
```

It fails when building a program, but not when building raylib itself as a static library.

## Steps to reproduce

1. Edit ``config.h`` do disable the ``rtext`` module, like so:
    ```//#define SUPPORT_MODULE_RTEXT            1```
2. Build raylib like so:
    ```make PLATFORM=PLATFORM_DESKTOP_SDL SDL_INCLUDE_PATH=/usr/include/SDL2/ SDL_LIBRARY_PATH=/usr/lib```
3. Finally, build a program using raylib:
    ```gcc test.c -lraylib -lSDL2```

NimComPoo-04:
Would it make sense to pull out the unicode related functions out of the text module and put them in a separate module (maybe inside util), since it is useful outside just text rendering?

by unicode related function I specifically mean the 3 functions `GetCodepoint`, `GetCodepointNext`, `GetCodepointPrevious`. Since they are not dependent on any other part of the rtext module could we enable those functions even if the rtext module is disabled?

PR #5005, opened 2025-06-24
Adding a check in rtext module to enable the GetCodepoint, GetCodepointNext, GetCodepointPrevious incase the backend is SDL

NimComPoo-04:
This pull request is in reference to issue #4959 

I added a `#if` to check if the platform is SDL, if it is we enable the Codepoint functions. This allows the program the build in that case.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1988,6 +1988,10 @@ const char *CodepointToUTF8(int codepoint, int *utf8Size)
     return utf8;
 }
 #endif      // SUPPORT_TEXT_MANIPULATION
+#endif
+
+// The function GetCodepointNext is utilized in rcore_desktop_sdl.c:1610 for textinput
+#if defined(SUPPORT_MODULE_RTEXT) || defined(PLATFORM_DESKTOP_SDL)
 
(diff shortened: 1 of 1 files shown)


raysan5:
@NimComPoo-04 I prefer to look for other solution, I try to keep modules as decoupled as possible, actually, afaik there is no check for `PLATFORM_` in any other module than `rcore` and `rlgl`. I think `rtext` module shouldn't be checking for platform-specific code.

raysan5:
For reference, SDL3 has [`SDL_StepUTF8()`](https://wiki.libsdl.org/SDL3/SDL_StepUTF8) available as a replacement for `GetCodepointNext()`.

raysan5:
Reviewed, added required function to `rcore_desktop_sdl`.

NimComPoo-04:
> @NimComPoo-04 I prefer to look for other solution, I try to keep modules as decoupled as possible, actually, afaik there is no check for `PLATFORM_` in any other module than `rcore` and `rlgl`. I think `rtext` module shouldn't be checking for platform-specific code.

Ok Yes that makes sense, should have looked out for it sorry.

PR #5006, opened 2025-06-24
Fix typo on config.h

ElDigoXD:
Fixes a typo in config.h, from "Selecte" to "Selected", same expression as the comment on line 223.


raysan5:
@ElDigoXD thanks for the review

PR #5011, opened 2025-06-27
[zig] Update emsdk version for zig build to fix the issue with the EM_BOOL

maiconpintoabreu:
In `v3.1.62` of `Emscripten`, they changed their bool type from int to bool, so the size of the `EmscriptenTouch` Structs were different from the current major version of the `Emscripten`, It was provoking memory corruption, getting completely wrong values from `GetTouchPosition`.

My issue was with the TouchScreen but it solves as well any issue that the struct is created on `Emscripten` side.

Issue:
* `Emscripten` trying to create a struct size: 1552
* `C` receiving a struct size: 1696

image

Solved:
* `Emscripten` trying to create a struct size: 1552
* `C` receiving a struct size: 1552

image

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -12,8 +12,8 @@
             .lazy = true,
         },
         .emsdk = .{
-            .url = "git+https://github.com/emscripten-core/emsdk#3.1.50",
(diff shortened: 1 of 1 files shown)


raysan5:
@maiconpintoabreu Oh! Thanks for the update!

PR #5015, opened 2025-07-01
[rmodels] Fixing offset for processing tangents for GLTF/GLB loading

Sir-Irk:
The [GLTF spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#meshes) defines TANGENT for meshes as a vec4. The code is loading this correctly but when transforming the tangents it's using an offset of 3 floats instead of 4.


raysan5:
@Sir-Irk thanks for the review!

PR #5016, opened 2025-07-01
[example] Fix pbr example shaders to use vec4 for vertexTangent

Sir-Irk:
The shaders are currently using a vec3 for vertexTangent despite Raylib uploading vec4's to them. I also multiplied the biNormal by the w component as this example loads .glb files which use the w component for the tangent basis sign.


raysan5:
@Sir-Irk thanks for the review!

PR #5018, opened 2025-07-04
[Zig] Fix issue on zig build Emscripten run if no emsdk installed

maiconpintoabreu:
As the zig build lazy download the `emsdk` while compiling, the user does not need to install `emscripten` on the machine previously.
This PR points the `emrun` and emrun.bat to zig cache folder if no `sysroot`.


raysan5:

PR #5020, opened 2025-07-04
update dr_libs

Emil2010:


Emil2010:
[raudio] because i'm stupid and always forgot to check other pull requests

raysan5:
@Emil2010 thanks for the update! Note it should be also reviewed here: https://github.com/raysan5/raylib/wiki/raylib-dependencies

PR #5023, opened 2025-07-05
Swapped `topRight` and `bottomRight` color parameters on `DrawRectangleGradientEx` as they were inverted.

Paperdomo101:
As the title says, the last two color paramaters on the function `DrawRectangleGradientEx` were the wrong way around.

Thanks to Jairo Corrêa on the discord for noticing the issue.


Paperdomo101:
Already fixed in #4980

PR #5026, opened 2025-07-06
fixed compile error for PLATFORM sdl

jonathandw743:
there is a compile error when PLATFORM=PLATFORM_DESKTOP_SDL, this fixes it


raysan5:
@jonathandw743 Good catch! Thanks!

PR #5031, opened 2025-07-08
Safety Comments for Sound/Texture Update Functions

AmityWilder:
`UpdateSound` neither checks to confirm that `sound.stream.buffer->data` can take `frameCount` frames, nor warns callers that `data` must be smaller than or equal to `sound.stream.buffer->data`. Users who don't check the definition of the function may mistakenly think it can resize `sound` to fit `frameCount` elements if additional space is needed, when in reality it would simply overrun the buffer. Textures perform no such checks either.

Raylib doesn't know the buffer size, but callers should at least be made aware of the preconditions.


AmityWilder:
Sorry for the big merge; I'm not super great at git and didn't realize it could have been a rebase instead

raysan5:
@AmityWilder thanks for the improvement

PR #5032, opened 2025-07-08
[examples] New example: `shaders_normalmap`

Sir-Irk:
This adds a basic example showing how to use a normal map on a plane and single point light. 
image

Edit: The texture files are from https://publicdomaintextures.com/


raysan5:
@Sir-Irk Thank you very much for the new great example! 😄

PR #5033, opened 2025-07-08
Update raudio.c

wwderw:
Allow for raudio to be used as a standalone module.  Including it here, because this fix was original done in the main raudio repo, but during an update to raudio to bring it up to date to the version that is being used here, that fix was regressed back.

Putting this PR here in hoping that the fix would be more persistent and filter to the raudio specific repo as well.  If accepted here, I'll remove the original PR from raudio's repo.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -81,7 +81,7 @@
     #include "utils.h"          // Required for: fopen() Android mapping
 #endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@wwderw thanks for the re-review!

PR #5034, opened 2025-07-09
[rcore] Added function to get previous mouse position.

sumten10:
In multiple projects i had to track the mouse position from previous frames and this is redundant to do manually, considering the previous position variable is already there, there just isn't a function to get the value.


raysan5:
@sumten10 Sorry but I prefer to avoid adding new functions if the functionality can be managed on user side.

Issue #5027, opened 2025-07-06
[rcore] gamepad inputs not working with linux glfw


jonathandw743:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib
 
There is a similar issue #3651 

### Issue description

gamepad inputs do not work on linux desktop glfw
I am using a Xbox Series X controller, wired.
The name and axis count are detected but the inputs are not.
The Xbox Series controller also seems to be here:
https://github.com/raysan5/raylib/blob/master/src/external/glfw/src/mappings.h

### Environment

Fedora Linux 42 (Workstation Edition)
OpenGL version string: 4.6 (Compatibility Profile) Mesa 25.0.7

### Code Example

https://github.com/jonathandw743/rlctt
`gcc main.c -Iinclude -Llibmastersdl -lraylib -lm -lSDL2`
`gcc main.c -Iinclude -Llibmasterglfw -lraylib -lm`
`gcc main.c -Iinclude -Llib550release -lraylib -lm`

The SDL version works but the other 2 do not.

Peter0x44:
Another piece of info you could provide is if it works through the browser.
I have a (third party) switch pro controller with pretty much the same problem. For controller inputs, SDL is king, and you should use it.
I might honestly be in favor of removing glfw's support just so people use something better.

jonathandw743:
> Another piece of info you could provide is if it works through the browser.

It does work on the online example website

jonathandw743:
a weird thing that I noticed was that the name returned:
"Microsoft Xbox Series S|X Controller"
is not in the GLFW mappings.h

JupiterRider:
@jonathandw743 In the meanwhile you can load the mappings manually:

```c
char *mappings = LoadFileText("gamecontrollerdb.txt");
SetGamepadMappings(mappings);


// ...
UnloadFileText(mappings);
```

Source: https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt

sleeptightAnsiC:
> a weird thing that I noticed was that the name returned:
> "Microsoft Xbox Series S|X Controller"
> is not in the GLFW mappings.h

[...]

PR #5036, opened 2025-07-09
[glfw] update mappings.h using GenerateMappings.cmake

sleeptightAnsiC:
I did this with the intention of fixing https://github.com/raysan5/raylib/issues/5027 but it's hard to tell if it will help in that case or not. Anyway, bumping the header with mappings should be done once in a while.

I haven't found any Raylib script for doing this, so here's how I did it step-by-step with CMake rule provided by GLFW (I'm on Linux, but this should be similar on any other POSIX system):
```sh
$ git clone https://github.com/raysan5/raylib.git
$ cd raylib/src/external/glfw/
$ cmake -B ./build -DGLFW_BUILD_EXAMPLES=OFF -DGLFW_BUILD_TESTS=OFF -DGLFW_BUILD_DOCS=OFF -G "Unix Makefiles"
$ make -C ./build update_mappings
$ rm -fr ./build
```
then I checked, if any mappings were deleted (without replacement) with following shell one-liner:
```sh
  # you must be at 'raylib/src/external/glfw/' when executing this
$ for i in $(git diff ./src/mappings.h | sed -n -e 's/^-"\([0-9a-z]*\),.*$/\1/p'); do grep <./src/mappings.h "$i" > /dev/null || echo "$i"; done;
```
which gave me the following IDs:
```txt
0300000066f700000500000000000000
030000000d0f00001600000000007803
03000000efbe0000edfe000000000000
03000000786901006e70000000000000
050000005e040000e002000030110000
03000000790000000600000010010000
030000000d0f00000d00000000010000
030000000d0f0000d800000072056800
03000000c0160000dc27000001010000
030000008916000000fd000024010000
```
I have checked all those IDs with `git log -p` against https://github.com/mdqinc/SDL_GameControllerDB and it seems that all of them were deleted by upstream (I was afraid that maybe some mappings were added manually by a Raylib contributor and then overwritten by my patch, but they weren't; All good!)


sleeptightAnsiC:
Also see https://github.com/raysan5/raylib/issues/5027#issuecomment-3054561168

sleeptightAnsiC:
Also this PR may fix some issues from https://github.com/raysan5/raylib/issues/3651

raysan5:
@sleeptightAnsiC thanks for the update!

PR #5037, opened 2025-07-10
[retextures] fix integer overflow in cast for `ImageBlurGaussian`

garrisonhh:
I had a weird bug where `ImageBlurGaussian` with `blurSize=1` added weird discolored lines:
<div>
lightblub-noblur
lightbulb-blur
</div>

The bug happens in the [final 'reverse premultiply' step](https://github.com/raysan5/raylib/blob/55a567471433dc60d92cb06f530d09ede97190c3/src/rtextures.c#L2219-L2222) where a rounding error can produce RGB color values that are >=256.0. When cast back to unsigned char for storage, this overflows to zero -- producing colors that are way off. This fix rounds these values down to 255 instead.


raysan5:
@garrisonhh Oh! Good catch! Thanks for the review!

PR #5041, opened 2025-07-14
[example][web] Fix core_custom_frame_control.c to properly register inputs on web platform

PanicTitan:
In the online example and when compiling locally, the inputs aren't being registered. While testing, I found that on the web, the input event poll only works when called after the inputs checks.

So I just added:
```
#ifndef PLATFORM_WEB
      PollInputEvents(); 
#endif
// ... inputs checks
```

and 

```
// ... inputs checks
#ifdef PLATFORM_WEB 
      PollInputEvents(); 
#endif
```

I don't know the root cause, I just made it work on web properly.

### DIFF
--- a/examples/core/core_custom_frame_control.c
+++ b/examples/core/core_custom_frame_control.c
@@ -61,7 +61,9 @@ int main(void)
     {
         // Update
         //----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


raysan5:
@PanicTitan I try to keep examples platform-agnostic so I try to avoid that kind of paltform-specific checks... still I'm merging until a better solution could be found.

PR #5043, opened 2025-07-14
Fix warning log macro in rlgl

vinnyhorgan:
Hi ray!

I was using rlgl standalone and bumped into this! I believed you meant to use the RLGL version of the macro.


raysan5:
@vinnyhorgan thanks for the review! Actually, `TRACELOG` should probably be `RL_TRACELOG()` for consistency, I'm reviewing it.

PR #5044, opened 2025-07-15
[rtextures] Fix HalfToFloat Mantissa hex value

ghost:
To match the [reference](https://github.com/raysan5/raylib/blob/f34e2406884b03ded23cd982ae12bcc2a29d13fe/src/rtextures.c#L5401). It used to be `0x03FF` (1023) but 5ddd13b changed it to `0x03cc` (972).


raysan5:
@backspaceoverload oh! good catch! thanks for the fix!

PR #5046, opened 2025-07-15
[raudio] Prevent UpdateMusicStream to run without music playing

maiconpintoabreu:
Save some resource when no music playing by stopping the UpdateMusicStream to process without music.


raysan5:

Issue #5009, opened 2025-06-25
[rtextures] `ColorReplace()` does not replace alpha values for RGB images


jestarray:
## EDIT
In a way this isn't a bug cause RGB images have no alpha values,  but nonetheless

## steps to reproduce:
1) Load in an imageformat that is RGB(use the one here):

![Image](https://github.com/user-attachments/assets/d1cc9646-9f68-4cb9-9193-821e687a1275)

2) Call `ColorReplace()` and you will notice it doesnt replace the alpha values

If we have:
```c
// Imagine the image we're passing in a 1x1 (0, 0, 0, 255) pixel
ColorReplace([0, 0, 0, 255], [0, 0, 0, 0]);
// the 255 does not get replaced with a 0 for RGB image formats
```

https://github.com/raysan5/raylib/blob/master/src/rtextures.c#L2930
The bug occurs because of these lines:
```c

    int format = image->format;
    RL_FREE(image->data);

    image->data = pixels;
    image->format = PIXELFORMAT_UNCOMPRESSED_R8G8B8A8;

    ImageFormat(image, format);
```
We set the new format but ImageFormat will write it back to the OLD format because that's what we pass in. This leads to the `PIXELFORMAT_UNCOMPRESSED_R8G8B8A8` branch not being taken for Images that are RGB

## purposed solutions:
- Don't call `ImageFormat` because we've already replaced the colors above. HOWEVER, `ImageFormat` does also handle mipmap generation, so if `ColorReplace` are to replace the colors in those mipmaps, then this solution is invalid.
- Let `ImageFormat` do the overwriting of `image->format`

A lot of the other image manipulation functions also have this same pattern, so there might be potential bugs there, and they should be reviewed for RGB images

### Environment
Linux, GL 3.3, rx480

This guy had same issue
https://discord.com/channels/426912293134270465/540620507708522517/1387477654768320552

RomainPlmg:
I am working on it

PR #5047, opened 2025-07-15
[rtextures] Fix ImageColorReplace() alpha issue replacement #5009

RomainPlmg:
### Summary

Fix a bug in `ImageColorReplace()` where replacing a color in an image with RGB format would drop the alpha values from the replacement color.

### Problem

Currently, `ImageColorReplace()` always calls `ImageFormat(image, format)` at the end, even if the input image was in RGB format. This causes the newly replaced alpha values to be discarded, since the image is converted back to a format that does not support alpha (`R8G8B8`).

This issue was reported [on Github](https://github.com/raysan5/raylib/issues/5009), and can be reproduced with:

```c
// A pixel originally (0, 0, 0, 255) in RGB image
ImageColorReplace(img, (Color){ 0, 0, 0, 255 }, (Color){ 0, 0, 0, 0 });
// The pixel's alpha stays 255 instead of becoming 0
```

### Solution

Only convert the image back to the original format if it supports an alpha channel.
This avoids losing alpha data introduced during the replacement.

```c
if (!formatHasNoAlpha) ImageFormat(image, format);
```

### Platforms tested

- Windows11
-  Tested with a 16x16 PNG image using `PIXELFORMAT_UNCOMPRESSED_R8G8B8` format: test-alpha

Let me know if you'd prefer a helper function like FormatHasAlpha() to be added.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -2921,13 +2921,27 @@ void ImageColorReplace(Image *image, Color color, Color replace)
         }
     }
 
-    int format = image->format;
+    const int format = image->format;
     RL_FREE(image->data);
 
     image->data = pixels;
     image->format = PIXELFORMAT_UNCOMPRESSED_R8G8B8A8;
 
-    ImageFormat(image, format);
+    const bool formatHasNoAlpha = (format == PIXELFORMAT_UNCOMPRESSED_R8G8B8) ||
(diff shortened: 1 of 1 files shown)


raysan5:
@RomainPlmg thanks for the review!

Issue #5021, opened 2025-07-05
[raudio] Non-looping music stream stops before end


Paperdomo101:
### Issue description

Playing a music stream with `looping` set to false will stop just before the end.

### Environment

```
OS: macOS 15.5
Processor: Apple M4 Pro (arm64)
```

### Issue Screenshot

Tested with `resources/sound.wav` 

![Image](https://github.com/user-attachments/assets/5530b252-af4f-4ebc-aea8-825b17c1d0c8)

### Code Example

In `examples/audio/audio_music_stream.c` 
```
// change
Music music = LoadMusicStream("resources/country.mp3");

// to
Music music = LoadMusicStream("resources/sound.wav"); // just so we don't have to wait as long
music.looping = false;
```

I believe the issue lies here at the end of `UpdateMusicStream` in `raudio.c`:
```
if (framesLeft <= subBufferSizeInFrames)
{
    if (!music.looping)   /*== this is being executed too early ==*/
    {
        ma_mutex_unlock(&AUDIO.System.lock);
        // Streaming is ending, we filled latest frames from input
        StopMusicStream(music);
        return;
    }
}
```

raysan5:
@Paperdomo101 It seems issue could be related to last audio frames not loaded to the streaming buffer or not processed. It requires investigation.

maiconpintoabreu:
Hi @Paperdomo101, looks like this issue started with: https://github.com/raysan5/raylib/pull/4898, can you try without this commit?
Thanks.

JeffM2501:
Try calling SetAudioStreamBufferSizeDefault with a specific size, like 4k, and see if that helps, perhaps the default buffer size computation is not correct.

veins1:
When the music stream ends in UpdateMusicStream, StopMusicStream is called which sets buffer->playing = false;
Then in OnSendAudioDataToDevice it just skips this audio buffer 
```
            // Ignore stopped or paused sounds
            if (!audioBuffer->playing || audioBuffer->paused) continue; 
```
so the audio samples are decoded into a buffer but never actually go out as audio.
#4898 by making buffers bigger it only reveals an issue that was there before. For resources/sound.wav the buffer is so big for me that I get no audio at all.

raysan5:
@Paperdomo101 Please, could you verify the issue haas been fixed with implemented PR?

Paperdomo101:
[...]

PR #5048, opened 2025-07-16
[raudio] Fix: music stopping early. Fix: GetMusicTimePlayed for short music

veins1:
Fix for #5021


raysan5:

PR #5050, opened 2025-07-18
[rtext] Update font load message

katanya04:
When loading a font with the LoadFont function in rtext.c, you get an info message on success that displays the pixel size and number of glyphs it has. But for some reason it doesn't use the actual font size and glyphs, it uses the constants `FONT_TTF_DEFAULT_SIZE` and `FONT_TTF_DEFAULT_NUMCHARS`. It's not a big deal but I spent like 10 minutes looking through the load font funcions to try and see why it "wasn't loading all the glyphs" (it was, it was the info message that was wrong).


raysan5:
@katanya04 Good catch! Thanks for the review!

PR #5053, opened 2025-07-19
Fixes GetWindowHandle returning a bad handle under Linux X11

didas72:
**The problem**: The `GetWindowHandle` function returns invalid handles in Linux with X11
**The solution**: Invoke the correct glfw function to get the window handle

After some experimentation with `GetWindowHandle` in Linux with X11, I realized that the current implementation of the function was returning invalid handles. After reading a [discussion](https://github.com/raysan5/raylib/discussions/2973#discussioncomment-5372414), it was clear what the source of the problem was.

As @raysan5 correctly identified, the naming conflict between the X11 lib and raylib when defining fonts inhibited the use of some X11 functions, resulting in a less-than-ideal implementation for this specific platform.

After some tinkering, I found a workaround and implemented the intended behavior for the function. The implementation has some details that likely need discussing and adjusting so I decided to create the PR.

The main topics that I believe need addressing are:
- Whether the naming conflict workaround is acceptable or ways to improve it
- Narrowing down the code definition to X11 only (and disable the new code on Wayland), as right now it will be included in any Linux targeting build

That being said, the code builds and runs as expected on Linux with X11. I noticed there were build warnings, but they were related to `rmodels.c`. This may be due to me using `gcc 15.1.1`, which may not be available in all platforms.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -74,9 +74,14 @@
 #if defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__)
     #include <sys/time.h>               // Required for: timespec, nanosleep(), select() - POSIX
 
-    //#define GLFW_EXPOSE_NATIVE_X11      // WARNING: Exposing Xlib.h > X.h results in dup symbols for Font type
     //#define GLFW_EXPOSE_NATIVE_WAYLAND
+    #define GLFW_EXPOSE_NATIVE_X11
+    #define Font X11Font                // Hack to fix 'Font' name collision
+                                        // The definition and references to the X11 Font type will be replaced by 'X11Font'
+                                        // Works as long as the current file consistently references any X11 Font as X11Font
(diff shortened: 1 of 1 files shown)


raysan5:
@didas72 Thanks for the review, merging it! Still not fully convinced about the requirement of a static variable roaming around, considering the value is returned every time the function is called, maybe it could be simplified to `return &glfwGetX11Window(platform.handle);` directly... but not sure about the implications...

didas72:
Thank you. I will leave a more detailed explanation of my rationale on the Discord.

creutzerb:
Hello
@raysan5 I have been using GetWindowHandle whenever I wanted to use DearImgui in a project (for the ImGui_ImplGlfw_InitForOpenGL call).

If I'm not mistaken,  that was the only way to retrieve the GLFWwindow* pointer of my raylib window.

Would you suggest to just modify a bit my build and add something like   GLFWwindow *GetGLFWWindow(void) that would return platform.handle ? or maybe that's something worth having in raylib anyways ?

PR #5055, opened 2025-07-20
C++20 formatters

lukasx999:
I know this is mainly a C library, however I think that being able to format some of the raylib types via `std::println` and `std::format` makes for a more seemless C++ debugging experience, and therefore warrants an exception.

I think this will be a great feature for C++ users, in addition to the overloaded math operators in `raymath.h`.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1713,4 +1713,54 @@ RLAPI void DetachAudioMixedProcessor(AudioCallback processor); // Detach audio s
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


orcmid (commented):
There are already C++ bindings that do not end up mingling C++ into the C Language raylib.h.  Please review those bindings for what you are looking for.  There is no need for such mechanisms at the Standard C raylib.h and supporting *.c modules level.

The addition of formatters might be more useful in one of those, or in a separate *.hpp that could be added at the C++ application level.

raysan5:
raylib is a C library, I'm afraid I'm not adding that C++ specific code.

PR #5061, opened 2025-07-23
[build][zig] use `addLibrary()` and set root modules manually

myQwil:
These changes will be required by Zig version 0.15.0.

Zig is phasing out the functions `addSharedLibrary` and `addStaticLibrary` and replacing them with `addLibrary`. Linkage mode is now specified using the `.linkage` option.

`.target` and `.optimize`  are also being removed as library/executable options. They now go directly into the root module.


raysan5:
@myQwil thanks for the review!

PR #5062, opened 2025-07-23
[rlgl] Add rlGetProcAddress

danilwhale:
Adds function to return address of the specified OpenGL procedure using previously specified `loader` from `rlLoadExtensions`.


raysan5:
@danilwhale thanks for the addition!

PR #5063, opened 2025-07-24
Fixed: CMake support for SDL3

kariem2k:
Added CMake support for SDL3.
Now supports including SDL2 or SDL3 as a subdirectory within the project. The system will first check for SDL3, then SDL2. If neither is found, it will attempt find_package(SDL3), followed by find_package(SDL2). If all these checks fail, the process will terminate with an error.


raysan5:
@kariem2k thanks for the review, note that the additional `USING_SDL3_PROJECT` could be renamed.

PR #5066, opened 2025-07-27
[glfw] Update `mappings.h`

lepasona:
i was experiencing #5027 with my Xbox Series X wired controller. these are the only ways i got it to work:
- add it to `mappings.h` and recompile raylib
- switch to the SDL backend [^1]
- target webassembly [^2]

so i added my gamepad to the upstream `gamecontrollerdb.txt` and updated `mappings.h` as in #5036

[^1]: SDL works but triggers behave differently. instead of ranging from -1 to 1 (like with glfw), they start at -1 and when pressed they jump to 0 which becomes their new minimum.
[^2]: webassembly works, i tested on https://www.raylib.com/examples.html, but my X and Y buttons are swapped (idk how to debug that).


sleeptightAnsiC:
Thanks for sending this to the upstream first. This greatly improves gamepad usability for everyone (not only for Raylib users).

> so i added my gamepad to the upstream gamecontrollerdb.txt

Just for cross-reference, the mentioned upstream PR is here: https://github.com/mdqinc/SDL_GameControllerDB/pull/869

> SDL works but triggers behave differently. instead of ranging from -1 to 1 (like with glfw), they start at -1 and when pressed they jump to 0 which becomes their new minimum.

SDL uses the same gamecontrollerdb.txt (with some additional logic) and since you upstreamed the mappings they should be fixed in SDL soon. Wrong axis range can be caused by wrong/missing mappings.

> webassembly works, i tested on https://www.raylib.com/examples.html, but my X and Y buttons are swapped (idk how to debug that).

Depending on the webbrowser you used, it should also has it's own gamepad detection. Chromium-based browsers should be a bit better in this case (better than Firefox/Safari) since Chromium is supported by gamepad vendors. You should probably take a look at the bugtracker of your browser. Here's an example ticket that might be related to this problem: https://issues.chromium.org/issues/415511856


### anyway, LGTM :)

lepasona:
ah thanks, i didn't know that about gamepads in browsers. i tested in chromium and the buttons are correct. it looks like the issue is basically this https://bugzilla.mozilla.org/show_bug.cgi?id=1763931 (untouched for 3 years because firefox ;_;). i'm not rly targeting browsers but thanks for the guidance on that anyway.

raysan5:
@lepasona thank you very much for this improvement!

PR #5068, opened 2025-07-27
[examples] typo: hardcoded gamepad ids in `core_input_gamepad.c`

lepasona:
noticed this while i was porting this example to zig, i think it's a mistake


raysan5:
@lepasona Thanks for the review! Definitely seems an issue!

PR #5069, opened 2025-07-29
Fixes GLFW failing when SetPropW fails silently

Peru-S:
**Issue**

In some cases, on Windows, `SetPropW`  returns `FALSE`.
raylib currently silently ignores the result.
This means that the message pump `windowProc` fails to work (does not receive keyboard events for example). Drawing is fine though as it's a different GLFW mechanism.

I had this failure frequently infrequently and it was frustrating to deal with (just have to rerun the program a few times and hope `SetPropW` doesn't return `FALSE`). So sending this fix in case others are facing the issue on Windows.

**Fix**

On Windows+GLFW, `windowProc` will fallback to looking up the GLFW window list (just like `PollEvents` does).
Also added the error code to `_glfwInputError` message to aid debugging in the future.

My suggestion is to get rid of `SetPropW`/`GetPropW` as it is a string-based Windows API (slow/fragile)- we already have the lightweight linked list, might as well just use it always - also it's usually just the one or two windows we deal with.

- [x] Read the [contributor guidelines](https://github.com/raysan5/raylib/blob/138010c1ebabe184097f3945534dfed1ca19e929/CONTRIBUTING.md)
- [x] Verified on Windows - message pump is stable with this fix (once I was able to repro this issue).

### DIFF
--- a/src/external/glfw/src/win32_window.c
+++ b/src/external/glfw/src/win32_window.c
@@ -551,8 +551,23 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM l
                     EnableNonClientDpiScaling(hWnd);
             }
         }
+        else
+        {
+            // HACK: SetPropW returns FALSE in some instances (GetLastError() = 0x8 Insufficient resources).
+            //       We already have the global list of windows, check against it. It's not many, and we are already
+            //       doing this in the (one of the callers) PollEvents below.
+            window = _glfw.windowListHead;
+            while (window)
+            {
+                if (window->win32.handle == hWnd) { break; }
+                window = window->next;
+            }
+        }
 
+        if (!window)
+        {
         return DefWindowProcW(hWnd, uMsg, wParam, lParam);
+    }
     }
 
     switch (uMsg)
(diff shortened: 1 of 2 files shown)


raysan5:
@Peru-S Note that GLFW is an external library, it must be fixed upstream on GLFW repo, if not, it will be override when updating GLFW on raylib.

src/external/glfw/src/win32_window.c
```diff
@@ -551,8 +551,24 @@ static LRESULT CALLBACK windowProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM l
...
+                window = window->next;
+            }
+        }
 
-        return DefWindowProcW(hWnd, uMsg, wParam, lParam);
+        if (!window)
+        {
+            return DefWindowProcW(hWnd, uMsg, wParam, lParam);
+        }
+    }
```
raysan5, line 571: This bracket seems duplicated, why there is another closing bracked aligned below?

Peru-S:
> @Peru-S Note that GLFW is an external library, it must be fixed upstream on GLFW repo, if not, it will be override when updating GLFW on raylib.

Good point, I will send it to glfw soon, thanks. Should I close this? Can leave this in case someone faces this problem on Windows.

Peru-S:
Moved to [glfw](https://github.com/glfw/glfw/pull/2749). Please close as needed.

raysan5:
This improvement will be hopefully implemented on next glfw library update.

PR #5076, opened 2025-07-31
[zig] Add root examples to be able to run rexm

maiconpintoabreu:
Add root example so it covers any example on the examples folder root.


raysan5:
I'm afraid I can't see the use case for this change.

PR #5077, opened 2025-07-31
Added certain variable assigment to the rename option and fixed some arugment based logs

lpow100:
Added the defining of exName, exCategory, and exRename.
Also fixed something where there was more args needed than said anywhere else


raysan5:
@lpow100 thanks for the improvement!

PR #5079, opened 2025-08-01
[rmodels] Fixing OBJ loading crash when there are no normals present

zet23t:
When loading OBJ files that have no normals in the file, the loader crashes. It does not crash for web builds, but I guess this is due to WASM not crashing for such memory access errors.


raysan5:
Thanks for the review @zet23t!

PR #5080, opened 2025-08-01
[examples][core] Propose to Add simple side to side ui controller example

maiconpintoabreu:
Small example on all to create UI control for for touch screen games.

https://github.com/user-attachments/assets/0456f2c5-0a67-4297-907f-f9f76bd6e599


maiconpintoabreu:
Sorry I did not see, that was a similar example.

Issue #4682, opened 2025-01-12
[examples] List of examples not working properly on Web platform


raysan5:
This issue is intended to get feedback for examples not working properly. Please, answer to this issue with the examples not working properly and the unexpected behaviour detected. Thanks!

michealguy:
![image](https://github.com/user-attachments/assets/6d2672bd-2261-4a48-a4fc-be50281c4f22)
Vertex Displacement doesn't seem to work for me

michealguy:
![image](https://github.com/user-attachments/assets/4b5850e6-ef16-4676-aaa9-e784852aaa65)
Mesh Instancing example doesn't seem to work as well.

nullstare:
Deferred render, Julia set and Texture tiling failed to compile the shader. Hybrid Rendering failed to load the rasterizing shader.

treblig-punisher:
Custom frame control not responding to keyboard input. Pressing Space Bar doesn't pause the animation and using the arrow keys doesn't make it faster or slower. Tested it in chrome, firefox nightly, regular firefox, and also Opera GX.
![image](https://github.com/user-attachments/assets/cb5b844e-b528-4df6-b76a-2323ab082951)

raysan5:
@treblig-punisher thanks for reporting, custom frame control requires a special raylib build (`SUPPORT_CUSTOM_FRAME_CONTROL` flag enabled) so it's expected not working on web. 

`models_skybox` is already on above list.

satchelfrost:
I see that `models_point_rendering` is not working on web. Sorry, I should have tested that. It appears that rlEnablePointMode() doesn't work on web targets. However, I can confirm glDrawArrays with GL_POINTS does work on web. I have an idea of how to fix it. Would you like me to give it a try and make a pull request, or perhaps open an issue to discuss the idea first?

raysan5:
@satchelfrost I think supporting GL_POINTS implies some redesign, global variables and code complexity increment. Am I right?

satchelfrost:
You hit the nail on the head 😅. I was thinking add a Boolean to the RLGL structure that gets toggled with rlEnablePointMode. If the added complexity is not wanted, then I completely understand rolling back the changes on point rendering.

zet23t:
The window flags example on browser is broken when making the window resizable and going into fullscreen mode. The cursor positions will then no longer match the coordinates and when returning from full screen, the canvas has a strange size.

[...]

PR #5081, opened 2025-08-01
[examples] resize fixes using IsWindowResized()

jopadan:
**changes**
- Use `IsWindowResized()` to update screen size as needed.
- Update examples window size dependent HUD and content

**example bug list** 
- #4682
- #3929
- #4908

**To reproduce just maximize the example window**

- **Before**
shaders_postprocessing

- **After**
shaders_postprocessing_new

### DIFF
--- a/examples/models/models_animation.c
+++ b/examples/models/models_animation.c
@@ -30,8 +30,8 @@ int main(void)
 {
     // Initialization
     //--------------------------------------------------------------------------------------
-    const int screenWidth = 800;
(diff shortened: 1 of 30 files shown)


raysan5:
@jopadan Please note that most examples are not intended to be resized, so I prefer to keep simplicity and avoid populating code with those checks. Users with specific resizing needs can implement it on user code. I think there is no need to add that extra code to multiple examples. Thanks for your understanding.

PR #5082, opened 2025-08-02
[CORE} Make GetRenderWidth/Height return the FBO size if one is active

JeffM2501:
This PR makes the GetRender size functions return the size of the active render target, if a render texture is active then the size of that texture will be returned. This gives users a consistent way to get the current target frame buffer size without having to know the current target. This makes creating reusable libraries much easier.


raysan5:

PR #5083, opened 2025-08-02
[rModels] Don't Double Upload Meshes in LoadObj, 

JeffM2501:
Both LoadObj and LoadModel try to upload the meshes for an obj file. This causes a warning when reading OBJ files.
None of the other interchange format readers try to upload meshes, they rely on LoadModel to do that.
This PR removes the UploadMesh calls from LoadObj so that it works like the other functions and removes the warning.


raysan5:
@JeffM2501 wow! good catch!

PR #5084, opened 2025-08-03
Removed verbose logging on SDL3. 

kariem2k:
Logs the property every time the property is read! It should be logged on the user side if needed.


raysan5:
@kariem2k Good catch! Thanks!

PR #5085, opened 2025-08-04
Make culling distances consistent across raylib

Not-Nik:
After updating my project from a six month old version to the current head, I had some pain with culling distances, specifically for my shadow maps. I went through and cleaned up anything not using the new culling distances so we are consistent across the library.


raysan5:
@Not-Nik thanks for the update! Note that those cull distances have been modified many times and they could be highly-dependant on target platform, GPU capabilities and drivers...

PR #5086, opened 2025-08-05
[rexm] Add compiled rexm to gitignore

maiconpintoabreu:


raysan5:
@maiconpintoabreu thanks for the addition!

PR #5087, opened 2025-08-05
add the ability to build rexm with CMake

JohnnyCena123:
now when building raylib as a standalone project it will automatically build rexm too


raysan5:
@JohnnyCena123 I prefer to avoid CMake.

JohnnyCena123:
oh

PR #5092, opened 2025-08-07
[remx] Add warning for missing new name to avoid segmentation fault

maiconpintoabreu:


raysan5:

PR #5093, opened 2025-08-07
[remx] Add check for category on new name

maiconpintoabreu:
This change is to block `./remx rename core_ test_`.
Ignore this PR if it was intended to be able to rename all file categories.


raysan5:
@maiconpintoabreu Nice! Thanks for the improvement!

PR #5103, opened 2025-08-09
Update raylib-zig in BINDINGS.md

Not-Nik:
The raylib-zig repository moved to its own organisation and already runs on a 5.6-dev version :)


raysan5:
@Not-Nik oh! that's great! thanks for the update!

PR #5104, opened 2025-08-09
Fix backspace key in Rpi SSH keyboard

matthijskooijman:
There was a typo in the generated keycode.


raysan5:
@matthijskooijman Good catch! Thanks for the review!

PR #5108, opened 2025-08-10
fix: QOA when using with LoadMusicStreamFromMemory

lassade:
LoadMusicStreamFromMemory isn't working for QOA right now, this PR fixes it, the main issue was that `file_data_offset` should be equal to `first_frame_pos` mimicking FILE behaviour.

I also joined the file data copy into a single malloc and removed the need of having an extra buffer when the data is already available. So overal it should do less malloc's and less memcpy's.


raysan5:
@lassade Thanks for the review!

PR #5111, opened 2025-08-14
Adds more filters for conditional build of GetWindowHandle with GLFW

### LINKED ISSUE #5110: [regression] broken wayland build
Commit https://github.com/raysan5/raylib/commit/36364192d59212dc48102e8ccb6d9162bf7fa78f has broken the Linux wayland build.

```
rcore.c.o: in function `GetWindowHandle':
rcore.c:(.text+0x21e8c): undefined reference to `glfwGetX11Window'
collect2: σφάλμα: η ld επέστρεψε κατάσταση εξόδου 1
```

didas72:
Fixes #5110

Adds more conditions for the macros for conditional compilation of X11 window handle getter.
Now allows code to be built exclusively for Wayland.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -75,13 +75,15 @@
     #include <sys/time.h>               // Required for: timespec, nanosleep(), select() - POSIX
 
     //#define GLFW_EXPOSE_NATIVE_WAYLAND
-    #define GLFW_EXPOSE_NATIVE_X11
-    #define Font X11Font                // Hack to fix 'Font' name collision
-                                        // The definition and references to the X11 Font type will be replaced by 'X11Font'
-                                        // Works as long as the current file consistently references any X11 Font as X11Font
-                                        // Since it is never referenced (as of writting), this does not pose an issue
-    #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window()
-    #undef Font                         // Revert hack and allow normal raylib Font usage
+    #ifdef _GLFW_X11
+        #define GLFW_EXPOSE_NATIVE_X11
+        #define Font X11Font                // Hack to fix 'Font' name collision
+                                            // The definition and references to the X11 Font type will be replaced by 'X11Font'
+                                            // Works as long as the current file consistently references any X11 Font as X11Font
+                                            // Since it is never referenced (as of writting), this does not pose an issue
+        #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window()
+        #undef Font                         // Revert hack and allow normal raylib Font usage
+    #endif
 #endif
 #if defined(__APPLE__)
     #include <unistd.h>                 // Required for: usleep()
@@ -710,7 +712,7 @@ void SetWindowFocused(void)
     glfwFocusWindow(platform.handle);
 }
 
-#if defined(__linux__)
+#if defined(__linux__) && defined(_GLFW_X11)
 // Local storage for the window handle returned by glfwGetX11Window
 // This is needed as X11 handles are integers and may not fit inside a pointer depending on platform
 // Storing the handle locally and returning a pointer in GetWindowHandle allows the code to work regardless of pointer width
@@ -723,7 +725,7 @@ void *GetWindowHandle(void)
     // NOTE: Returned handle is: void *HWND (windows.h)
     return glfwGetWin32Window(platform.handle);
 #endif
-#if defined(__linux__)
+#if defined(__linux__) && defined(_GLFW_X11)
     // Store the window handle localy and return a pointer to the variable instead
     // Reasoning detailed in the declaration of X11WindowHandle
     X11WindowHandle = glfwGetX11Window(platform.handle);


src/platforms/rcore_desktop_glfw.c
```diff
@@ -710,7 +712,7 @@ void SetWindowFocused(void)
     glfwFocusWindow(platform.handle);
 }
 
-#if defined(__linux__)
+#if defined(__linux__) && defined(_GLFW_X11)
```
planetis-m, line 726:
```c
#if defined(__linux__)
    #if defined(_GLFW_WAYLAND)
        #if defined(_GLFW_X11)
            int platformID = glfwGetPlatform();
            if (platformID == GLFW_PLATFORM_WAYLAND) {
                return glfwGetWaylandWindow(platform.handle);
            }
            else {
                return glfwGetX11Window(platform.handle);
            }
        #else
            return glfwGetWaylandWindow(platform.handle);
        #endif
    #elif defined(_GLFW_X11)
        return glfwGetX11Window(platform.handle);
    #endif
#endif
```

src/platforms/rcore_desktop_glfw.c
```diff
@@ -75,13 +75,15 @@
...
-    #undef Font                         // Revert hack and allow normal raylib Font usage
+    #ifdef _GLFW_X11
+        #define GLFW_EXPOSE_NATIVE_X11
+        #define Font X11Font                // Hack to fix 'Font' name collision
+                                            // The definition and references to the X11 Font type will be replaced by 'X11Font'
+                                            // Works as long as the current file consistently references any X11 Font as X11Font
+                                            // Since it is never referenced (as of writting), this does not pose an issue
+        #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window()
+        #undef Font                         // Revert hack and allow normal raylib Font usage
+    #endif
```
planetis-m, line 85:
```c
#if defined(_GLFW_X11) || defined(_GLFW_WAYLAND)
    // Set appropriate expose macros based on available backends
    #if defined(_GLFW_X11)
        #define GLFW_EXPOSE_NATIVE_X11
        #define Font X11Font                // Hack to fix 'Font' name collision
                                            // The definition and references to the X11 Font type will be replaced by 'X11Font'
                                            // Works as long as the current file consistently references any X11 Font as X11Font
                                            // Since it is never referenced (as of writting), this does not pose an issue
    #endif
    
    #if defined(_GLFW_WAYLAND)
        #define GLFW_EXPOSE_NATIVE_WAYLAND
    #endif

    // Include native header only once, regardless of how many backends are defined
    #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window() and glfwGetWaylandWindow()

    // Clean up X11-specific hacks
    #if defined(_GLFW_X11)
        #undef Font                         // Revert hack and allow normal raylib Font usage
    #endif
#endif
```

src/platforms/rcore_desktop_glfw.c
didas72: The suggested code does not work in the X11 variant, which is why my older PR was made. I will adapt the proposed code to work with the temporary handle store for X11.

In terms of what is more formally correct, should I include your email as co-author in the commit message?

planetis-m: Why it doesn't work? Afaik the previous code was just : return platform.handle

> should I include your email as co-author in the commit message?

No need to.

didas72: glfwGetX11Window returns a long unsigned, which might always be castable to void* (and is also not advisable to cast it anyways) so in my last PR i made it store the handle statically and return a pointer to the static location instead (in the case of X11 only).

I'll commit the new changes soon, should have the intended behavior now, but all feedback is welcome.

src/platforms/rcore_desktop_glfw.c
```diff
@@ -724,10 +738,25 @@ void *GetWindowHandle(void)
...
 #endif
 #if defined(__linux__)
-    // Store the window handle localy and return a pointer to the variable instead
-    // Reasoning detailed in the declaration of X11WindowHandle
-    X11WindowHandle = glfwGetX11Window(platform.handle);
-    return &X11WindowHandle;
+    #if defined(_GLFW_WAYLAND)
+        #if defined(_GLFW_X11)
+            int platformID = glfwGetPlatform();
+            if (platformID == GLFW_PLATFORM_WAYLAND) {
```
raysan5: Please, review raylib conventions.

src/platforms/rcore_desktop_glfw.c
```diff
@@ -74,14 +74,28 @@
...
-                                        // The definition and references to the X11 Font type will be replaced by 'X11Font'
-                                        // Works as long as the current file consistently references any X11 Font as X11Font
-                                        // Since it is never referenced (as of writting), this does not pose an issue
-    #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window()
-    #undef Font                         // Revert hack and allow normal raylib Font usage
+#if defined(_GLFW_X11) || defined(_GLFW_WAYLAND)
+    // Set appropriate expose macros based on available backends
+    #if defined(_GLFW_X11)
+        #define GLFW_EXPOSE_NATIVE_X11
+        #define Font X11Font                // Hack to fix 'Font' name collision
```
raysan5: Please, could you align comments with above ones?

src/platforms/rcore_desktop_glfw.c
```diff
@@ -74,14 +74,28 @@
...
+    #if defined(_GLFW_WAYLAND)
+        #define GLFW_EXPOSE_NATIVE_WAYLAND
+    #endif
+
+    // Include native header only once, regardless of how many backends are defined
+    #include "GLFW/glfw3native.h"       // Required for: glfwGetX11Window() and glfwGetWaylandWindow()
+
+    // Clean up X11-specific hacks
+    #if defined(_GLFW_X11)
+        #undef Font                         // Revert hack and allow normal raylib Font usage
```
raysan5: Please, can you align comment?

src/platforms/rcore_desktop_glfw.c
```diff
@@ -724,10 +738,25 @@ void *GetWindowHandle(void)
...
-    // Reasoning detailed in the declaration of X11WindowHandle
-    X11WindowHandle = glfwGetX11Window(platform.handle);
-    return &X11WindowHandle;
+    #if defined(_GLFW_WAYLAND)
+        #if defined(_GLFW_X11)
+            int platformID = glfwGetPlatform();
+            if (platformID == GLFW_PLATFORM_WAYLAND) {
+                return glfwGetWaylandWindow(platform.handle);
+            }
+            else {
```
raysan5: Braces are aligned in raylib

didas72:
My apologies for the formatting. Should have done a better review on formatting before the commit.

raysan5:
@didas72 thanks for the review!

src/platforms/rcore_desktop_glfw.c
planetis-m: Good point, I missed it.

PR #5113, opened 2025-08-14
Specified the pixel sequence order in the code comment

RolandMarchand:
Problem: When using the function `Color *LoadImageColors(Image image)`, I did not know what was the pixel order (left-to-right, top-to-bottom, etc.).

Solution: I read through the source code and identified it was row-major (left-to-right, top-to-bottom). I added this information to the comment of the function for future users.

No logic code was modified, so no need for tests, ports, or examples.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -2943,7 +2943,7 @@ void ImageColorReplace(Image *image, Color color, Color replace)
 }
 #endif      // SUPPORT_IMAGE_MANIPULATION
 
-// Load color data from image as a Color array (RGBA - 32bit)
+// Load color data from image as a Color array (RGBA - 32bit) with row-major order
 // NOTE: Memory allocated should be freed using UnloadImageColors();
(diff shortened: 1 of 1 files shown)


raysan5:
@RolandMarchand I think that comment is not required, afaik, in the graphics programming world and images world, the default order for loading pixel data into memory is always left-right and top-bottom.

Please, could you point me to some popular engine or graphics library with a similar function using a different approach?

lazzy-cipher:
Godot 3 uses bottom-top ordering of viewport images, which was my main experience with graphics programming before Raylib. But I understand if this is niche and think this comment is redundant.

https://docs.godotengine.org/en/3.5/classes/class_viewport.html

> Note: By default, a newly created Viewport in Godot 3.x will appear to be upside down. Enabling [render_target_v_flip](https://docs.godotengine.org/en/3.5/classes/class_viewport.html#class-viewport-property-render-target-v-flip) will display the Viewport with the correct orientation.

PR #5118, opened 2025-08-17
Fix SDL3 build if using package

slendidev:
Before it would try to include "SDL.h" which is incorrect. There's also a mistake when calling SDL_StepUTF8 that resulted in the platform not being built.


raysan5:
@slendidev Thanks for the review!

PR #5119, opened 2025-08-17
Fix SDL3 compilation error with gcc

jan-beukes:
When compiling PLATFORM_DESKTOP_SDL using gcc, it results in an error: -Wincompatible-pointer-types. 
The fix from #5118 worked when using clang as it only gives a warning.


raysan5:
@jan-beukes thanks for the review

PR #5123, opened 2025-08-19
[raudio] Initialize sound alias properties as if it was a new sound

JeffM2501:
The sound alias would copy the volume from the source sound, this made the new sound be in a different state than it would be if it was loaded as a copy.
This PR forces sound alias to the same state as a new sound for consistency.


raysan5:

PR #5124, opened 2025-08-19
[utility] Add helper header to fix compatibility with windows.h

JeffM2501:
This PR adds a utility header to external that undefines the parts of windows that conflict with raylib, allowing raylib functions to run unimpeded


raysan5:

PR #5127, opened 2025-08-20
Add Color to DrawGrid

Xphalnos:
Being able to choose the grid color helps avoid this kind of case with the background color:

We can use custom colors (Or change transparency):

For the examples, I put GRAY because in most cases, RAYWHITE is used for the background


raysan5:
@Xphalnos I'm afraid this change would break many projects out there, it can no be changed at the moment, or at least this way.

If you need a custom grid you can just implement it, it's a couple of lines of code.

Issue #5126, opened 2025-08-20
[*] Update build.zig for 0.15.0


CashWasabi:
# Update build.zig

I updated the `build.zig` and `build.zig.zon` files to be compatible with zig's `0.15.0` release.
It needed some minor changes because the writer interfaces and `ArrayList` changed recently.
These changes are not backwards compatible and would be breaking changes for versions earlier than `0.15.0` though.
I have a working version in a forked repository [here](https://github.com/CashWasabi/raylib/tree/zig-0.15).

If you want I can make a PR.

Hope you have a nice day and thanks for all the work on raylib. 👍

raysan5:
@CashWasabi Feel free to send a PR.

PR #5128, opened 2025-08-20
[build] Update `build.zig` (breaking)

CashWasabi:
# Update build.zig (breaking)

As mentioned [here](https://github.com/raysan5/raylib/issues/5126) I prepared a PR for zig's 0.15.0 release.
This release has breaking changes though.
I would love to have it reviewed from someone that has worked a bit with 0.15.0.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,7 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 /// Minimum supported version of Zig
-const min_ver = "0.13.0";
+const min_ver = "0.15.1";
 
 const emccOutputDir = "zig-out" ++ std.fs.path.sep_str ++ "htmlout" ++ std.fs.path.sep_str;
 const emccOutputFile = "index.html";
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -2,7 +2,7 @@ const std = @import("std");
 const builtin = @import("builtin");
 
 /// Minimum supported version of Zig
-const min_ver = "0.13.0";
+const min_ver = "0.15.0";
```
Not-Nik: This should use 0.15.1 as minimum Zig version, as the 0.15 release is broken

Not-Nik (changes requested):

raysan5:
@Not-Nik do you think it is safe to merge this update? Is Zig community updating to 0.15.1?

build.zig
CashWasabi: Good to know! I just updated and tested it on `0.15.1`.
It's building all examples so I think it's good now.
While reading the release notes I realized that `std,ArrayList` is now `std.array_list.Managed` and changed it accordingly where it matters.
This reduces the lines changed a little bit.

CashWasabi: I read a little further and it seems that the intention is to only keep `std.ArrayList` in the future.
Now I'm thinking it might make sense to just use `std.ArrayList`.

Not-Nik: Yup, from the change log it seems `std.ArrayList` is the way to go, since `std.array_list.Managed` will be removed in the future.

Not-Nik:
> Is Zig community updating to 0.15.1?

Definitely, yes. Many users are on the master branch version anyways (which is now some 0.16-dev version), so I expect the community will transition quickly. Let's just wait until the `std.ArrayList` vs. `std.array_list.Managed` is sorted out here, before this is merged.

build.zig
objarni: Oh, thanks for clarifying the ArrayList release notes to me! I thought ArrayList was being removed in favor of std.array_list.Managed. 

What does the "->" notation in the Release notes mean? I don't get it.

objarni: BTW really appreciate getting build.zig version for zig 0.15.1 for Raylib out the door!

CashWasabi: > Oh, thanks for clarifying the ArrayList release notes to me! I thought ArrayList was being removed in favor of std.array_list.Managed.
> 
> What does the "->" notation in the Release notes mean? I don't get it.

It basically means that the functionality of those structs moved over to other structs.
`std.ArrayList` behavior for example moved to `std.array_list.Managed`.
`std.ArrayList` is now an unmanaged list.

objarni: Thanks!

CashWasabi:
I did the dirty deed and force pushed the changes.
The PR should be ready for merge now.

raysan5:
@CashWasabi @Not-Nik thanks for the review!

PR #5129, opened 2025-08-20
remove nonexistent example

JohnnyCena123:
text_unicode_fonts no longer exists, so its removed
I also ran `rexm update` to sync everything


JohnnyCena123:
I did an oopsie

PR #5132, opened 2025-08-21
[examples][core] refactoring core_input_mouse.c to use IsCursorHidden function for consistency

maiconpintoabreu:
Small refactor to make the example use a Raylib's function instead of local variable by swapping local variable `isCursorHidden` with `IsCursorHidden()`


raysan5:

Issue #5131, opened 2025-08-20
[raudio] UnloadMusicStream fails to close FLAC file


rossberg:
The UnloadMusicStream function, in the case for FLAC files, only invokes drflac_free, but drflac_close is never called by Raylib. Due to that, an opened FLAC file is never closed, and some memory internal to dr_flac is never freed. As a consequence, Raylib leaks both file handles and memory. In addition, on Windows, all played FLAC file(s) stay locked and cannot be moved or deleted until the application exits.

See the implementation of drflac_close [here](https://github.com/raysan5/raylib/blob/99cb82a82ce9973d45e61ad24ad9899e6f1d0d38/src/external/dr_flac.h#L8944).

PR #5133, opened 2025-08-21
[raudio] Properly close FLAC in UnloadMusicStream

rossberg:
Fix  raysan5/raylib#5131 by invoking drflac_close before drflac_free in UnloadMusicStream..


raysan5:
@rossberg Oh! Good catch! Thanks for the review!

PR #5137, opened 2025-08-21
Clamp camera pitch

feive7:
If you look straight down (or up I think) the camera yaw flips 180 degrees. This fixes that

### DIFF
--- a/examples/core/core_3d_fps_controller.c
+++ b/examples/core/core_3d_fps_controller.c
@@ -261,7 +261,9 @@ static void UpdateCameraAngle(Camera *camera)
     Vector3 right = Vector3Normalize(Vector3CrossProduct(yaw, up));
 
     // Rotate view vector around right axis
-    Vector3 pitch = Vector3RotateByAxisAngle(yaw, right, -lookRotation.y - lean.y);
+    float pitchAngle = -player.body.lookRotation.y - player.lean.y;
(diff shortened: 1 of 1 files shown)


examples/core/core_3d_fps_controller.c
```diff
@@ -261,7 +261,9 @@ static void UpdateCameraAngle(Camera *camera)
     Vector3 right = Vector3Normalize(Vector3CrossProduct(yaw, up));
 
     // Rotate view vector around right axis
-    Vector3 pitch = Vector3RotateByAxisAngle(yaw, right, -lookRotation.y - lean.y);
+    float pitchAngle = -player.body.lookRotation.y - player.lean.y;
+    pitchAngle = Clamp(pitchAngle, -PI / 2 + 0.0001f, PI / 2 - 0.0001f); // Clamp angle so it doesn't go past straight up or straight down
+    Vector3 pitch = Vector3RotateByAxisAngle(yaw, right, pitchAngle);
```
TokyoSU, line 266: You can do this to optimize it (well it's just micro-optimization, you can ignore it):
```
// Clamp angle so it doesn't go past straight up or straight down
Vector3 pitch = Vector3RotateByAxisAngle(yaw, right, Clamp(-player.body.lookRotation.y - player.lean.y, -PI / 2 + 0.0001f, PI / 2 - 0.0001f));
```

raysan5: It's a nice optimization but considering it's an example for education, I think current code is a bit more readable.

raysan5:
@feive7 thanks for the improvement!

PR #5139, opened 2025-08-23
use RL_MALLOC in stb_truetype

Siltnamis:
Some allocations that were made by STBTT_malloc are being deallocated with RL_FREE in LoadFontFromMemory. This is an issue if user defines custom RL allocation functions because he is asked to deallocate something he never allocated.


raysan5:
@Siltnamis Good catch! Thanks for the review!

PR #5143, opened 2025-08-24
Add pub to compileRaylib

annaymone:
This is necessary to allow for using the function in dependent `build.zig`s.


raysan5:
@annaymone thanks for the review!

PR #5149, opened 2025-08-28
Add headless mode

JohnnyCena123:
this is the initial implementation.
it might not be the cleanest, barely held together, but i can still polish it more
heres what i tested:
- building works for both make and cmake, examples are disabled entirely in headless mode
- rexm links with the built headless-mode raylib library, and works well
and that's it really.

the main goal of this was to avoid linking graphical/audio/input related modules to cli tools which dont need it (i.e. rexm)


JohnnyCena123:
@raysan5 I know you said you dont need it right now, but it can wait however much you want

raysan5:
@JohnnyCena123 Thanks but there are no plans to add this change for now and keeping open PRs just generates burden to me. I prefer to avoid it if there is no intention to at least review (and potentially merge).

PR #5151, opened 2025-08-28
fix: digital clock 

hmz-rhl:
the minute number position is outside the window, so we can't see the minute (when we change mode to hand free mode after pressing spacebar)
image


raysan5:
@hmz-rhl Good catch! Thanks for the fix!

PR #5152, opened 2025-08-28
fix unnecessary warnings in some files

JohnnyCena123:


raysan5:
@JohnnyCena123 thanks for the review!

PR #5154, opened 2025-08-29
change all raymath functions to be constexpr in C++ by default

JohnnyCena123:


raysan5:
@JohnnyCena123 Sorry, no plans to change this.

JohnnyCena123:
well, I prefer constexpr when possible, but its your choice

PR #5156, opened 2025-08-30
avoid cleaning web builds with plain `make clean` for examples

JohnnyCena123:


raysan5:
@JohnnyCena123 Why it should avoid cleaning web builds?

JohnnyCena123:
because there's a clean for web builds separately
and as of right now, it only cleans the wasm files - not js/html/data - and only on posix, where it deletes files based on them having execute permissions

PR #5157, opened 2025-08-30
[build.zig]: a concrete approach to build for web with zig-build

HaxSam:
For a long time, I didn't like how Zig builds for the web.
The main problem was that there wasn't any helper method and that the build system emsdk was cloned twice—once for the Raylib dependency and once on the user side.
With the helper methods, it will be cloned only once and use the same folder for the operation.

Here is an example of how the build step can look now on the user side:
```zig
const std = @import("std");
const rl = @import("raylib");

pub fn build(b: *std.Build) !void {

    const raylib_dep = b.dependency("raylib", .{
        .target = target,
        .optimize = optimize,
    });
    const raylib_artifact = raylib_dep.artifact("raylib");

    const exe_mod = b.createModule(.{
        .root_source_file = b.path("src/main.zig"),
        .target = target,
        .optimize = optimize,
    });
    exe_mod.linkLibrary(raylib_artifact);

    const run_step = b.step("run", "Run the app");

    //web exports are completely separate
    if (target.query.os_tag == .emscripten) {
        const emsdk = rl.emsdk;
        const wasm = b.addLibrary(.{
            .name = "raylib_testing",
            .root_module = exe_mod,
        });

        const install_dir: std.Build.InstallDir = .{ .custom = "web" };
        const emcc_flags = emsdk.emccDefaultFlags(b.allocator, .{
            .optimize = optimize,
        });
        const emcc_settings = emsdk.emccDefaultSettings(b.allocator, .{
            .optimize = optimize,
        });

        const emcc_step = emsdk.emccStep(b, raylib_artifact, wasm, .{
            .optimize = optimize,
            .flags = emcc_flags,
            .settings = emcc_settings,
            .install_dir = install_dir,
        });
        b.getInstallStep().dependOn(emcc_step);

        const html_filename = try std.fmt.allocPrint(b.allocator, "{s}.html", .{wasm.name});
        const emrun_step = emsdk.emrunStep(
            b,
            b.getInstallPath(install_dir, html_filename),
            &.{},
        );

        emrun_step.dependOn(emcc_step);
        run_step.dependOn(emrun_step);
    } else {
            .name = "raylib_testing",
            .root_module = exe_mod,
        });
        b.installArtifact(exe);

        const run_cmd = b.addRunArtifact(exe);
        run_cmd.step.dependOn(b.getInstallStep());

        run_step.dependOn(&run_cmd.step);
    }
}
```

It relies heavily on the zemscripten library, which provides a solid foundation for using emsdk on different platforms.

I would like to request that @Not-Nik and @maiconpintoabreu review it and share their thoughts, especially from Nik, since he maintains the Zig bindings. This approach also works with the Zig bindings and can be copied and pasted. I already tested it with a local raylib-zig instance, where I replaced the raylib bindings with my own, and it allowed me to remove `emsdk.zig`.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -14,6 +14,72 @@ comptime {
         @compileError("Raylib requires zig version " ++ min_ver);
 }
 
+pub const emsdk = struct {
+    const zemscripten = @import("zemscripten");
+
+    pub fn shell(b: *std.Build) std.Build.LazyPath {
+        return b.dependency("raylib", .{}).path("src/shell.html");
+    }
+
+    pub const FlagsOptions = struct {
+        optimize: std.builtin.OptimizeMode,
+        asyncify: bool = true,
+    };
+
+    pub fn emccDefaultFlags(allocator: std.mem.Allocator, options: FlagsOptions) zemscripten.EmccFlags {
+        var emcc_flags = zemscripten.emccDefaultFlags(allocator, .{
+            .optimize = options.optimize,
+            .fsanitize = true,
+        });
+
+        if (options.asyncify)
+            emcc_flags.put("-sASYNCIFY", {}) catch unreachable;
+
+        return emcc_flags;
+    }
+
+    pub const SettingsOptions = struct {
+        optimize: std.builtin.OptimizeMode,
+        es3: bool = true,
+        emsdk_allocator: zemscripten.EmsdkAllocator = .emmalloc,
+    };
+
+    pub fn emccDefaultSettings(allocator: std.mem.Allocator, options: SettingsOptions) zemscripten.EmccSettings {
+        var emcc_settings = zemscripten.emccDefaultSettings(allocator, .{
+            .optimize = options.optimize,
+            .emsdk_allocator = options.emsdk_allocator,
+        });
+
+        if (options.es3)
+            emcc_settings.put("FULL_ES3", "1") catch unreachable;
+        emcc_settings.put("USE_GLFW", "3") catch unreachable;
+        emcc_settings.put("EXPORTED_RUNTIME_METHODS", "['requestFullscreen']") catch unreachable;
+
+        return emcc_settings;
+    }
+
+    pub fn emccStep(b: *std.Build, raylib: *std.Build.Step.Compile, wasm: *std.Build.Step.Compile, options: zemscripten.StepOptions) *std.Build.Step {
+        const activate_emsdk_step = zemscripten.activateEmsdkStep(b);
+
+        const emsdk_dep = b.dependency("emsdk", .{});
+        raylib.root_module.addIncludePath(emsdk_dep.path("upstream/emscripten/cache/sysroot/include"));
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -364,25 +386,17 @@ pub fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize:
...
+                raylib.root_module.addCMacro("GRAPHICS_API_OPENGL_ES3", "");
             }
         },
         else => {
             @panic("Unsupported OS");
         },
     }
 
-    raylib.root_module.addCSourceFiles(.{
+    raylib.addCSourceFiles(.{
```
Not-Nik: This should stay a call to `root_module.addCSourceFiles`. The `Step.Compile` version is [deprecated in 0.15.1](https://ziglang.org/documentation/0.15.1/std/#std.Build.Step.Compile.addCSourceFile).

HaxSam: Good catch. I think I accidentally merged the wrong snippet when I updated my branch.

HaxSam: I looked into it and there are multiple lines with the same problem.
~I have the idea to change the flow a bit.~
~First just create a module and include everything then create the lib see:~
~https://github.com/raysan5/raylib/blob/bbf884d7b3068089d007ccc9a9eac0e7ea59af2b/build.zig#L162C5-L169C8~

~new approach:~
```zig
const raylib_mod = b.createModule(.{
    .optimize = optimize,
    .target = target,
    .link_libc = true,
}),

// lot of code
   
// at the end 
const raylib = b.addLibrary(.{
     .name = "raylib",
     .linkage = options.linkage,
     .root_module = raylib_mod,
 });
```

---
edit:
it won't work out I just do `root_module` now on every deprecated method

HaxSam:
@Not-Nik Ok fixed it and did the others too

build.zig
```diff
@@ -14,58 +14,80 @@ comptime {
...
+    }
+};
+
+fn setDesktopPlatform(raylib: *std.Build.Module, platform: PlatformBackend) void {
+    switch (platform) {
+        .glfw => raylib.addCMacro("PLATFORM_DESKTOP_GLFW", ""),
+        .rgfw => raylib.addCMacro("PLATFORM_DESKTOP_RGFW", ""),
+        .sdl => raylib.addCMacro("PLATFORM_DESKTOP_SDL", ""),
+        .android => raylib.addCMacro("PLATFORM_ANDROID", ""),
+        else => {},
```
Not-Nik: Keeping this function the way it was (with `raylib: *std.Build.Step.Compile`) is a bit cleaner imo, but it's not that important and definitely not blocking.

Not-Nik (approved):

build.zig
HaxSam: Oh yeh is quick change let me do that

raysan5:
@HaxSam @Not-Nik Thanks for the update and review!

Issue #5065, opened 2025-07-26
[rcore][drm] `FindNearestConnectorMode()` always prefers equal fps


maiphi:
### Issue description

I've got a display that reports as preferred mode `1280x800 @ 59.492904 Hz` and also `1920x1080@60` (among others).

With a target framerate of 60 Hz, this preferred mode will never be selected due to the way `FindNearestConnectorMode()` in `platforms/rcore_drm.c:2186` identifies the nearest mode if there is no exact match:

`if ((widthDiff < nearestWidthDiff) || (heightDiff < nearestHeightDiff) || (fpsDiff < nearestFpsDiff)) nearestIndex = i;`

This will never prefer the correct resultion if there's a mode with matching fps. In the present case we get `1920x1080@60`.

Setting `SetTargetFPS(59)` solves the problem, but will fail the same way as soon as we connect a different display that actually announces 60 Hz.

### Possible solutions

* Allow for a small fps difference when the resolution matches exactly.
* Introduce a distance measure for modes.
* The perfect solution would be to default to the actual current framerate (no idea how to do that).

### Environment

* Raspberry Pi Compute Module 5 with Raspberry Pi OS 32-bit, up to date.
* Display: Waveshare 10.1HP-CAPLCD.
* Raylib `5.5-540-g1fc0b495` (main branch from today).
* Configured with `cmake -DBUILD_SHARED_LIBS=ON -DPLATFORM=DRM`.

```
INFO: Initializing raylib 5.6-dev
INFO: Platform backend: NATIVE DRM
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
INFO:     > raudio:.... loaded (optional)
WARNING: DISPLAY: No graphic card set, trying platform-gpu-card
WARNING: DISPLAY: Failed to open platform-gpu-card, trying card1
INFO: DISPLAY: card1 opened successfully
TRACE: DISPLAY: Connectors found: 2
TRACE: DISPLAY: Connector index 0
TRACE: DISPLAY: Connector modes detected: 10
TRACE: DISPLAY: DRM mode connected
TRACE: DISPLAY: Searching exact connector mode for 1280x800@60, selecting an interlaced mode is allowed: no
[...]

PR #5158, opened 2025-08-30
`FindNearestConnectorMode()`: prefer mode with smallest number of unused pixels

maiphi:
Fixes https://github.com/raysan5/raylib/issues/5065.

Always prefer a mode with the exact resolution, even if the fps doesn't match the requested value. If such a mode doesn't exist, prefer the one with the smallest number of unused pixels. The fps are only used to break ties: prefer smaller fps mismatch in that case.

*Notes:*
* As before, only modes with equal or larger width and height are allowed (one could loosen this restriction if scaling of the output is requested).
* There is a parameter `const drmModeConnector *connector` which is redundant, because inside the function, `&platform.connector` is used instead (I left it that way).
* One could prefer modes with higher fps instead of the smallest difference, if such a mode exists. Do we want that?


raysan5:
@maiphi Thanks for the review! I think this approach is better!

PR #5160, opened 2025-08-31
changed function params formatting

wileyanderssen:


raysan5:
@wileyanderssen Good catch! Thanks for the review!

PR #5163, opened 2025-09-01
removed macro that does nothing

wileyanderssen:


raysan5:
@wileyanderssen Actually, it does... but it's in the wrong place. If PNG is disabled, code will fail on BMP starting with an `else` statement. I'm reviewing it.

PR #5164, opened 2025-09-01
Update Camera3D.fovy description in raylib.h

0stamina:
Previously stated it was the near plane width in orthographic, in actuality it's the height. The wording has been fixed, with added clarity that it's in world units so no conversion is needed.


raysan5:
@0stamina thanks for the review!

PR #5168, opened 2025-09-03
[build.zig] Make X11 the default display backend instead of choosing at runtime

### LINKED ISSUE #5165: [rcore] Setting Window Icon does not work when building with zig toolchain (Zig build should use X11 by default as done in CMake setup)
### Issue description

Setting window/app icon via `SetWindowIcon(icon);` has no effect when building with zig toolchain. Zig toolchain uses X11 and Wayland both as default, seen here:

https://github.com/raysan5/raylib/blob/aa475b10461018f6f2ed02cfe0bc628d6a8b9698/build.zig#L428

On Wayland, app icon doesn't work on my system, and throws these Errors
```
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not provide the window position
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not support setting the window position
WARNING: GLFW: Error: 65548 Description: Wayland: The platform does not support setting the window icon
```

Setting backend to X11, fixes it.

```zig
    const raylib = b.dependency("raylib", .{
        .target = target,
        .optimize = optimize,
        .linux_display_backend = .X11, // <----
    });
```

**In my opinion**, setting app icon on wayland is a well known problem, so building with Zig toolchain should use X11 as default, just like it is done in CMake toolchain.

### Environment

* OS: Manjaro Linux x86_64 
* CPU: AMD Ryzen 5 5600H with Radeon Graphics (12) @ 4.280GHz 
* GPU: AMD ATI Radeon Vega Series / Radeon Vega Mobile Series 
* Kernel: 6.16.0                               

#### With Wayland backend:

Image

#### With X11 backend:

Image

Not-Nik:
In line with Makefiles (#4369) and Make (#4432) this makes X11 the default Linux display backend again, instead of choosing one at runtime, where it would prefer Wayland over X11. The reasoning is explained further here (#4371) and here (#5165)

Fixes #5165


raysan5:
@Not-Nik Thank you very much for the review!

PR #5170, opened 2025-09-05
[rcore] Fixing base64 decoding error when input string is bad

zet23t:
The following code would crash the previous version when calling MemFree:

	// 53 * A
	const char maliciousBase64Input[] = "AAAAAAAAAAAAAAAAAAAAAAAA"
		"AAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
	int decodedSize = 0;
	unsigned char *decodedData = DecodeDataBase64(
		maliciousBase64Input, &decodedSize);
	if (decodedData) {
		MemFree(decodedData);
	}

The reason is a lack of array bound checks in the decoding loop, which corrupted here the heap (though this is platform dependent).

Adding the bound checks here prevents the memory corruption.

Tested with encoding random data of sizes 0-1023 and comparing it with the decoded result.

Note 1: It would be great if the functions like DecodeDataBase64 or DecompressData had "Try..." functions that return 0 or 1 depending on success. I understand that passing pointers is somewhat awkward for less advanced users, but this function already accepts a pointer anyway. My data comes from user inputs, so I have to run some checks on it and it would be a bit easier if there were "Try..." functions in case decoding fails. I currently can't tell if the decoding was successful or not.

Alternatively: In case errors are encountered, the function could also free the memory and return 0. I think this would be better than returning a data block that is containing corrupted data.

Note 2: I created a test file to first verify the crash before fixing and verifying that the code still works as expected. Would it make sense to add a test folder to raylib that contains small test programs that test certain aspects of the rcore library functions? In that case, I would have a first test for that 😅

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2671,13 +2671,24 @@ unsigned char *DecodeDataBase64(const char *text, int *outputSize)
     for (int i = 0; i < dataSize;)
     {
         // Every 4 sixtets must generate 3 octets
+        if (i + 2 >= dataSize)
+        {
+            TRACELOG(LOG_WARNING, "BASE64 decoding error: Input data size is not valid");
+            break;
+        }
+
         unsigned int sixtetA = base64DecodeTable[(unsigned char)text[i]];
         unsigned int sixtetB = base64DecodeTable[(unsigned char)text[i + 1]];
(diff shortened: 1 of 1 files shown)


raysan5:
@zet23t Thanks for the review!

1. Not sure if you refer to try-catch statements but they are not supported in C. In any case, the function could return NULL in case the process fails.
2. I prefer not to add test cases, it will imply adding more and more things to maintain.

PR #5173, opened 2025-09-05
 Add RL_PIXELFORMAT_UNCOMPRESSED_B8G8R8A8 

Fancy2209:
Add BGRA8 as a texture format


raysan5:
@Fancy2209 After some thinking I decided not to add this new format. It implies checking the extension availability and also not fully supported on OpenGL ES 2.0 low-end devices. Adding it would also imply adding `B8G8R8`, adding more code and more maintenance work to address a very specific use case that most users will never need.

Converting the image to one of the supported formats on user side is the easiest solution.

Fancy2209:
> @Fancy2209 After some thinking I decided not to add this new format. It implies checking the extension availability and also not fully supported on OpenGL ES 2.0 low-end devices. Adding it would also imply adding `B8G8R8`, adding more code and more maintenance work to address a very specific use case that most users will never need.
> 
> Converting the image to one of the supported formats on user side is the easiest solution.

I see, understandable!

PR #5174, opened 2025-09-08
[rtext] Add GetDefaultFontSpacing()

jaxvanyang:
Hi, I was trying to draw text in the center of screen, so I need to measure text size. But `MeasureTextEx()` needs text spacing, and there's no way to get the spacing `DrawText()` use except reading the code. This PR just adds a function for this kind of situation. Since it's a rare situation, it's totally fine to close this as not planned. Just open here in case of this situation not considered.

About the new function name, I don't know how to suit Raylib, maybe `GetDrawTextSpacing()` is better?


raysan5:
@jaxvanyang I prefer to avoid adding a new function just for that, default text spacing is a known value.

PR #5177, opened 2025-09-09
[rlgl] Allow tint colors in `GL_LINE` (wires) and `GL_POINT` (points) draw modes on OpenGL 1.1

meisei4:
Decided to make a PR because the fix was a little more involved than i felt i could show on discord, and i wanted to show more cleanly where in the code im observing the fix behavior.

### Issue
In the Opengl 1.1 context the colors array is getting enabled incorrectly during wire and point mode i think

**Desired behavior:** `DrawModelWires(..., Color tint)`, should draw the wire model with the wires being colored by "tint"

**Observed Behavior:** loading an obj file and setting it to draw the model fill with WHITE and wire frame any color other than WHITE fails. 

**Workaround:** if you nullify the `.colors` field in the mesh before drawing with wires or points, you can set the color in the draw function

**this PR's solution:** during `DrawMesh`, find the current `GL_POLYGON_MODE` and from there do not enable the `GL_COLOR_ARRAY` if you are in modes of `GL_LINE` or `GL_POINT`, this allows for the draw functions tint to apply without the workaround.

NOTE: I also added `rlSetPointSize(float size) ` (and getter) to allow for `glPointSize(size)`, because my eyes could not debug point mode and color well without this feature of increasing point size (its also a feature i would like to have for debugging in opengl1.1)

### Concerns:
I am wondering if this code change has any consequences outside of what i believe to be normal behavior matching the other opengl modes now.

Images of current behavior (will only draw white wires, even when undesired -- e.g. on top of an also white model fill)
image

Images of desired behavior (custom point and wire colors over the main model color)
image

Screenshot From 2025-09-10 01-22-25

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -6239,19 +6239,17 @@ static ModelAnimation *LoadModelAnimationsGLTF(const char *fileName, int *animCo
             *animCount = (int)data->animations_count;
             animations = (ModelAnimation *)RL_CALLOC(data->animations_count, sizeof(ModelAnimation));
 
-            Transform worldTransform;
-            {
-                cgltf_float cgltf_worldTransform[16];
-                cgltf_node* node = skin.joints[0];
-                cgltf_node_transform_world(node->parent, cgltf_worldTransform);
-                Matrix worldMatrix = {
-                    cgltf_worldTransform[0], cgltf_worldTransform[4], cgltf_worldTransform[8], cgltf_worldTransform[12],
-                    cgltf_worldTransform[1], cgltf_worldTransform[5], cgltf_worldTransform[9], cgltf_worldTransform[13],
-                    cgltf_worldTransform[2], cgltf_worldTransform[6], cgltf_worldTransform[10], cgltf_worldTransform[14],
-                    cgltf_worldTransform[3], cgltf_worldTransform[7], cgltf_worldTransform[11], cgltf_worldTransform[15]
-                };
-                MatrixDecompose(worldMatrix, &(worldTransform.translation), &(worldTransform.rotation), &(worldTransform.scale));
-            }
+            Transform worldTransform = { 0 };
+            cgltf_float cgltf_worldTransform[16] = { 0 };
+            cgltf_node *node = skin.joints[0];
+            cgltf_node_transform_world(node->parent, cgltf_worldTransform);
+            Matrix worldMatrix = {
+                cgltf_worldTransform[0], cgltf_worldTransform[4], cgltf_worldTransform[8], cgltf_worldTransform[12],
+                cgltf_worldTransform[1], cgltf_worldTransform[5], cgltf_worldTransform[9], cgltf_worldTransform[13],
+                cgltf_worldTransform[2], cgltf_worldTransform[6], cgltf_worldTransform[10], cgltf_worldTransform[14],
+                cgltf_worldTransform[3], cgltf_worldTransform[7], cgltf_worldTransform[11], cgltf_worldTransform[15]
+            };
+            MatrixDecompose(worldMatrix, &worldTransform.translation, &worldTransform.rotation, &worldTransform.scale);
 
             for (unsigned int i = 0; i < data->animations_count; i++)
             {
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -447,8 +447,8 @@ int GetMonitorCount(void)
 int GetCurrentMonitor(void)
 {
     int displayId = -1;
-    JNIEnv* env = NULL;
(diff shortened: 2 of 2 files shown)


src/rmodels.c
```diff
@@ -1436,7 +1439,13 @@ void DrawMesh(Mesh mesh, Material material, Matrix transform)
     if (mesh.animNormals) rlEnableStatePointer(GL_NORMAL_ARRAY, mesh.animNormals);
     else rlEnableStatePointer(GL_NORMAL_ARRAY, mesh.normals);
 
-    rlEnableStatePointer(GL_COLOR_ARRAY, mesh.colors);
+    extern void glGetIntegerv(unsigned int pname, int *data);
```
raysan5: I'm afraid this is an undesired approach...

src/rlgl.h
```diff
@@ -2009,6 +2011,14 @@ float rlGetLineWidth(void)
     return width;
 }
 
+void rlSetPointSize(float size) { glPointSize(size); }
+
+float rlGetPointSize(void) {
```
raysan5: This is not following [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md).

raysan5:
@meisei4 I see the issue but the proposed approach should be reviewed. `rmodels` never calls `gl*()` functions directly

src/rlgl.h
meisei4: updated to have the curly brace conform and added comments the same as Line width (I am basing the change mostly on line width)

meisei4:
updated with a simple null check, even with the wire and points poor solution it didnt solve the fill mode, actual solution is just `if(mesh.colors)`, just like the rest of the mesh fields

meisei4:
final result with the fix:
BLUE fill, RED wires, GREEN points
image

and then with a material example
Screenshot From 2025-09-10 02-46-37

meisei4:
Updating this because actually i falsely assumed i had rebuilt everything correctly. The issue still remains with OpenGL 1.1 even with the `mesh.colors `null check. 

The reason in my case is because `LoadOBJ` will always allocate WHITE/255 into the `colors` field for the `mesh`: 
https://github.com/raysan5/raylib/blob/84e2cbccc4b734939cce87dbd462d5f21d61a540/src/rmodels.c#L4422
https://github.com/raysan5/raylib/blob/84e2cbccc4b734939cce87dbd462d5f21d61a540/src/rmodels.c#L4490

The workaround is to just set the `mesh.colors` to NULL after you load the obj

apologies for the sloppiness, I am still learning, and i imagine there is a more appropriate way than just inserting the opengl  context macro into the `LoadOBJ` (Also I have a feeling this might be an issue elsewhere where for opengl 1.1 the colors can be a bit tricky to deal with smoothly in comparison to other opengl contexts)🙇‍♂️.

src/rlgl.h
```diff
@@ -2010,6 +2012,17 @@ float rlGetLineWidth(void)
     return width;
 }
 
+// Set the point drawing size
+void rlSetPointSize(float size) { glPointSize(size); }
```
raysan5: I'm afraid `glPointSize()` is not available on OpenGL ES 2.0. Please, could we avoid adding this function?

src/rlgl.h
```diff
@@ -2010,6 +2012,17 @@ float rlGetLineWidth(void)
...
+
+// Get the point drawing size
+float rlGetPointSize(void)
+{
+    float size = 0;
+    glGetFloatv(GL_POINT_SIZE, &size);
```
raysan5: I'm afraid `GL_POINT_SIZE` is not available on OpenGL ES 2.0. Please, could we avoid adding this function?

src/rlgl.h
meisei4: I was under the impression that it would be ok, because of the equivalent pattern for `WireMode` (and polygon mode in general according to comments i believe) being gated by context macros: 

https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L1968

Would it be ok to keep it if we add the appropriate version context checks like with `WireMode`?🤔

raysan5: yeah, we can use same approach

meisei4:
I would like to clean this PR up, it seems to have gotten very verbose because of git for some reason bringing in a ton of the upstream commits, even though i tried to rebase so that its only my file changes, i will try to organize that better when i have time, and then i can test the gating feature

raysan5:
@meisei4 I don't know what happened with this PR, please, could you review it to only update the required files?

meisei4:
I messed up the rebasing process, i will fix this now, and force push up ONLY the relevant changes from this PR, otherwise i will close this and then make new PR, will do right nwo

PR #5179, opened 2025-09-12
[build][cmake] export automatically raylib definitions and compile/link options

brccabral:
Installing `raylib-targets.cmake` from CMake sends all PUBLIC definitions and options added from `target_compile_definitions`, `target_compile_options` to consumer.  
It is the raylib library telling consumer how to import itself. Consumer just has to `target_link_library(project PRIVATE raylib)` and the project will be correctly configured.

A user that selects Platform Desktop will get
```cmake
set_target_properties(raylib PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "PLATFORM_DESKTOP;GRAPHICS_API_OPENGL_33"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:>"
)
```
if Web
```cmake
set_target_properties(raylib PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "PLATFORM_WEB;GRAPHICS_API_OPENGL_ES2"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_OPTIONS "-sUSE_GLFW=3;-sEXPORTED_RUNTIME_METHODS=ccall;-sASYNCIFY"
)
```


raysan5:
@brccabral Merging but I do not maintain the CMake build system, so user with more experience should take a look. 

@Not-Nik please, could you take a look?

Not-Nik:
I don't really use CMake's install functionality, but I had a look anyways. FWICT this doesn't break anything (good) and it looks to be a bit more idiomatic (also good). That being said, I don't think it really changes anything for the end user, apart from maybe saving a few linking directives here and there. For example on macOS it includes information to link with OpenGL.framework. Overall a decent update.

raysan5:
@Not-Nik Thank you very much for taking a look, I like PRs that simplify code/building!

PR #5180, opened 2025-09-12
[build][cmake] Added emscripten build options

brccabral:
Missing options for compiling as Web.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -69,7 +69,7 @@ else()
 endif()
 
 if (${PLATFORM} MATCHES "Web")
-    target_link_options(raylib PUBLIC "-sUSE_GLFW=3")
+    target_link_options(raylib PUBLIC "-sUSE_GLFW=3" -sEXPORTED_RUNTIME_METHODS=ccall -sASYNCIFY)
     if(${GRAPHICS} MATCHES "GRAPHICS_API_OPENGL_ES3")
         target_link_options(raylib PUBLIC "-sMIN_WEBGL_VERSION=2")
         target_link_options(raylib PUBLIC "-sMAX_WEBGL_VERSION=2")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ include(CompilerFlags)
 # Registers build options that are exposed to cmake
 include(CMakeOptions.txt)
 
-if (UNIX AND NOT APPLE AND NOT "${PLATFORM}" MATCHES "DRM")
+if (UNIX AND NOT APPLE AND NOT "${PLATFORM}" MATCHES "DRM" AND NOT "${PLATFORM}" MATCHES "Web")
   if (NOT GLFW_BUILD_WAYLAND AND NOT GLFW_BUILD_X11)
     MESSAGE(FATAL_ERROR "Cannot disable both Wayland and X11")
   endif()


src/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ else()
 endif()
 
 if (${PLATFORM} MATCHES "Web")
-    target_link_options(raylib PUBLIC "-sUSE_GLFW=3")
+    target_link_options(raylib PUBLIC "-sUSE_GLFW=3" -sEXPORTED_RUNTIME_METHODS=ccall -sASYNCIFY)
```
raysan5, line 72: Why `-sASYNCIFY` is being added for source code building?

brccabral: Without `-sASYNCIFY` Firefox throws Javascript error and application does not load.  
`-sEXPORTED_RUNTIME_METHODS=ccall` is not need though, I took these options from [examples/build_example_web.bat](https://github.com/raysan5/raylib/blob/master/examples/build_example_web.bat).  Let me know if you want to remove EXPORTED_RUNTIME_METHODS.

image

raysan5: oh, so that cmake is used to build examples? I thought it was only for raylib source code, considering is placed in src directory 🤔

brccabral: In top level CMakeLists.txt it has add_subdirectory(src) which builds the library, but also a condition "if BUILD_EXAMPLES add_subdirectory(examples)" and inside examples folder there is another CMakeLists.txt that gathers all examples and creates targets for them. Then, we can build one example at a time, or build them all in one go. For any selected platform (desktop, sdl, web...)
```cmake
cmake -DBUILD_EXAMPLES=ON -S . -B build
cmake --build build --target all
```
I use CLion, which makes things much easier. There are extensions for VSCode too.

raysan5:
@brccabral Thanks for the review and clarifications!

Peter0x44:
Are you sure this is a good idea? -sASYNCIFY is not needed if you use emscripten_set_main_loop and apparently makes perf worse.

Peter0x44:
IMO this should be reverted.

raysan5:
@Peter0x44 thanks for the review and the comments! As per mi understanding, this CMake is intended only for raylib examples collection, that requires `-sASYNCIFY`... or I missunderstood it? @brccabral please, can you confirm?

And yeah, `-sASYNCIFY` affects performance, it must not be used if using `emscripten_set_main_loop()`.

Peter0x44:
If I'm reading this correctly, it will make sure any time you link with raylib through cmake, `-sASYNCIFY` gets passed. When you might not want it.
It's not specific to the examples.

Peter0x44:
If the intention is that it only passes this to the examples, then IMO, it should be the examples  CMakeLists.txt providing the link options, not raylib.

brccabral:
Well, the option `-sUSE_GLFW=3` was already there, I saw the others in the [examples/build_example_web.bat](https://github.com/raysan5/raylib/blob/master/examples/build_example_web.bat) and thought this was the way Raylib uses Emscripten.  
I am fine with having link options in the examples (SDL does it).  
It would be nice having a specific example, but I am not sure how `emscripten_set_main_loop()` works.  
[SDL examples](https://github.com/libsdl-org/SDL/blob/main/examples/CMakeLists.txt) sets only `-sALLOW_MEMORY_GROWTH=1`, no `-sASYNCIFY `...

Peter0x44:
-sUSE_GLFW=3 is fine because it is required 100% of the time. The other two flags are not, and ASYNCIFY has negative consequences!

It's unacceptable to require those flags for everyone consuming raylib through cmake.

core_basic_window_web.c demonstrates how to use emscripten_set_main_loop.
I'll submit a PR reverting this tomorrow if no one else gets to it first.

brccabral:
This line is in [examples/CMakeLists.txt](https://github.com/raysan5/raylib/blob/master/examples/CMakeLists.txt)
```cmake
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s WASM=1 -s ASYNCIFY -s ALLOW_MEMORY_GROWTH=1 --shell-file ${CMAKE_SOURCE_DIR}/src/shell.html")
```

brccabral:
Where is `core_basic_window_web.c`? https://github.com/raysan5/raylib/tree/master/examples/core ?

Peter0x44:
Ray deleted it in this commit.
https://github.com/raysan5/raylib/commit/c17f82a78cf7cbfa08646271f72f4d209ffb5f9f

brccabral:
Found something, the browser error was caused by `emscripten_sleep(12);` in [src/platforms/rcore_web.c](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_web.c).  
There is no Javascript error if I comment this line.
Edit: after reading the comment, looks like WindowShouldClose() should not be used on web...

```c
bool WindowShouldClose(void)
{
    // Emscripten Asyncify is required to run synchronous code in asynchronous JS
    // REF: https://emscripten.org/docs/porting/asyncify.html

    // WindowShouldClose() is not called on a web-ready raylib application if using emscripten_set_main_loop()
    // and encapsulating one frame execution on a UpdateDrawFrame() function,
    // allowing the browser to manage execution asynchronously

    // Optionally we can manage the time we give-control-back-to-browser if required,
    // but it seems below line could generate stuttering on some browsers
    emscripten_sleep(12);

    return false;
}
```

brccabral:
This file contains a PLATFORM_WEB example [projects/CMake/core_basic_window.c](https://github.com/raysan5/raylib/blob/master/projects/CMake/core_basic_window.c), but its [projects/CMake/CMakeLists.txt](https://github.com/raysan5/raylib/blob/master/projects/CMake/CMakeLists.txt) uses `-s ASYNCIFY`

raysan5:
> Ray deleted it in this commit. [c17f82a](https://github.com/raysan5/raylib/commit/c17f82a78cf7cbfa08646271f72f4d209ffb5f9f)

Yeah, I'm reorganizing examples at the moment and I just deleted it temporally to place it somewhere else. It was just a few days ago, sorry for the confusion.

@brccabral When building for web, it's recommended to use the `emscripten_set_main_loop()` mechanism provided by emscriptem but that's not always possible so ASYNCIFY alternative is available, with some performance hit. Examples structure is common for all platforms so using that ASYNCIFY approach but users creating code from scratch with web in mind (specially if it is a big project) should use the `emscripten_set_main_loop()` approach.

When using `ASYNCIFY`, emscripten must be notified when it has time to manage the loop, so `emscripten_sleep()` is required... but never when using `emscripten_set_main_loop()`.

PR #5184, opened 2025-09-15
[rcore] Don't overwrite old gifs

mfbulut:
Having a gif recorder is a nice feature but when I try to record gifs in seperate sessions it overwrites by old gifs and this is a bit annoying so here is a small fix for that

So it makes

#### Session 1
screenrec0.gif
screenrec1.gif

#### Session 2
screenrec2.gif
screenrec3.gif


raysan5:
@mfbulut I'm afraid this feature is being removed, no plans to update it now.

PR #5192, opened 2025-09-17
[rTextures] Use the render size not the screen size for screenshots

JeffM2501:
The screenshot code is always applying the DPI scale to the image, even if the framebuffer was not created with that scale, thus if you have a DPI scale but are NOT doing DPI auto scaling and using the window at it's native resolution, the image will be too big.

The screenshot code should never use ScreenSize, it should use RenderSize, so that it gets the size of the framebuffer that it's saving. There should not be any need for High DPI logic here since this is just saving an existing framebuffer that has a fixed size.

This PR makes the change so that screenshots work in both modes on all platforms since it's just using the same size that the framebuffer was created with.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -592,11 +592,10 @@ Image LoadImageFromTexture(Texture2D texture)
 // Load image from screen buffer and (screenshot)
 Image LoadImageFromScreen(void)
 {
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 Good catch! Thanks for the review!

PR #5193, opened 2025-09-18
[build][cmake] set lib m as public so it can be linked by consumer

brccabral:
If m is private consumer has to link both raylib and m, as public consumer can link with just raylib and raylibConfig will tell them to link m.


raysan5:
@Peter0x44 please, could you take a look to this PR?

cmake/LibraryConfigurations.cmake
```diff
@@ -38,19 +38,22 @@ if (${PLATFORM} MATCHES "Desktop")
...
     else ()
         find_library(pthread NAMES pthread)
         find_package(OpenGL QUIET)
         if ("${OPENGL_LIBRARIES}" STREQUAL "")
             set(OPENGL_LIBRARIES "GL")
         endif ()
 
-        set(LIBS_PRIVATE m atomic pthread ${OPENGL_LIBRARIES} ${OSS_LIBRARY})
+        set(LIBS_PRIVATE atomic pthread ${OPENGL_LIBRARIES} ${OSS_LIBRARY})
+        set(LIBS_PUBLIC m)
```
Peter0x44, line 51: Why is this here, when later there's  a check for netbsd and openbsd that also adds m. Wouldn't that make this second one redundant? Are you sure it needs to be there?

brccabral: new commit removes redundancy

raysan5:
@brccabral @Peter0x44 thanks for the review.

Issue #5142, opened 2025-08-24
[rlgl] Drawing triangles breaks rendering when using more than one texture


MilesVII:
### Issue description

When drawing textured triangles with `rlTexCoord2f` and `rlVertex2f` functions from `rlgl` as in [textured polygon example](https://github.com/raysan5/raylib/blob/master/examples/textures/textures_polygon.c), after setting the second texture, every following drawing in the frame has texcoords and vertices corrupted, including those made with `rshapes` and `rtextures` rlDraw* functions.

### Environment

Desktop, Win10 22H2, OpenGL 4.6, tested on two devices

### Issue Screenshots

When drawing in TRIANGLES mode (broken)
<img width="364" height="276" alt="Image" src="https://github.com/user-attachments/assets/6e11d8cb-e260-406e-99b2-bbce770fe810" />

When drawing in QUADS mode (as expected)
<img width="418" height="384" alt="Image" src="https://github.com/user-attachments/assets/732a90c6-ef06-4865-8811-4be852a1b8f8" />
_The fox and background are drawn before the tiles. When rendered after the tiles, the fox sprite gets corrupted too. Black tiles are drawn before the red ones, and only the red ones are broken_

Minimal example; triangles vs quads, blue tile is drawn after the red one:
<img width="427" height="250" alt="Image" src="https://github.com/user-attachments/assets/99b22921-992b-42b2-93a5-2ddbe02af179" />

### Code Example

I prepared a minimal example in Odin ([repo](https://github.com/MilesVII/rlgl-issue-demo/blob/main/main.odin)), can port to C if needed
``` odin
package main

import rl "vendor:raylib"
import rlgl "vendor:raylib/rlgl"

SIZE     := [2]f32 {  50, 50 }
ORIGIN_A := [2]f32 {  50, 75 }
ORIGIN_B := [2]f32 { 100, 75 }

main :: proc() {
	rl.SetTraceLogLevel(.WARNING)
	rl.InitWindow(200, 200, "fops")
	defer rl.CloseWindow()

	rl.SetTargetFPS(60)

	texA := rl.LoadTexture("./a.png")
	texB := rl.LoadTexture("./b.png")

	for !rl.WindowShouldClose() {
		rl.BeginDrawing()
		rl.ClearBackground(rl.WHITE)

		// the goal is to draw two textures by specifying their vertices

		buggyTriangles(texA, texB)
		// niceQuads(texA, texB)

		rl.EndDrawing()
	}
}

[...]

PR #5194, opened 2025-09-18
[rlgl] Fix Not default Draw Mode breaking when changing texture

maiconpintoabreu:
This PR is to fix the `rlSetTexture` as when you have different textures it will not override the mode that is default to GL_QUAD.
The fix is to check if the `drawCounter` has more then 1 and copy the previous mode to the new Draw.

Fix for the issue: https://github.com/raysan5/raylib/issues/5142 

image


maiconpintoabreu:
I am closing as it needs more work

PR #5195, opened 2025-09-19
[rlgl] Fix Not default Draw Mode breaking when changing texture

maiconpintoabreu:
This PR is to fix the `rlSetTexture` as when you have different textures it will not override the mode that is default to GL_QUAD.
The fix is to check if the `drawCounter` has more then 1 and copy the previous mode to the new Draw.
If the developer is drawing a mix of Modes The `rlSetTexture` need to be after `rlBegin`.

Fix for the issue: https://github.com/raysan5/raylib/issues/5142 

image

## RenderDoc tests:

image

Screenshot_20250919_095035

Screenshot_20250919_101854

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -1676,13 +1676,13 @@ void rlSetTexture(unsigned int id)
                     RLGL.State.vertexCounter += RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].vertexAlignment;
 
                     RLGL.currentBatch->drawCounter++;
+                    
+                    RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].mode = RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 2].mode;
                 }
             }
 
             if (RLGL.currentBatch->drawCounter >= RL_DEFAULT_BATCH_DRAWCALLS) rlDrawRenderBatch(RLGL.currentBatch);
-            if (RLGL.currentBatch->drawCounter > 1) {
-                RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].mode = RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 2].mode;
-            }
+
             RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].textureId = id;
             RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].vertexCount = 0;
         }
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -3939,8 +3939,13 @@ void DrawBillboardPro(Camera camera, Texture2D texture, Rectangle source, Vector
     texcoords[2] = (Vector2){ (float)(source.x + source.width)/texture.width, (float)source.y/texture.height };
     texcoords[3] = (Vector2){ (float)source.x/texture.width, (float)source.y/texture.height };
 
+#if defined(GRAPHICS_API_OPENGL_11)
     rlSetTexture(texture.id);
     rlBegin(RL_QUADS);
+#else
+    rlBegin(RL_QUADS);
+    rlSetTexture(texture.id);
+#endif
 
         rlColor4ub(tint.r, tint.g, tint.b, tint.a);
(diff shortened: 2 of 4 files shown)


raysan5:
@maiconpintoabreu I can't remember at the moment if `rlSetTexture()` needs to be after `rlBegin()` or before but I wonder if everything still works on OpenGL 1.1, it could be sensible in that case...

This PR changes how it was designed originally... and I'm afraid it was structure that way for some reason... but it was so long ago that I can't remember...

maiconpintoabreu:
@raysan5, this change does not change any OpenGL call, but I will try to test it on 1.1 and post the results later.
I did a diagram explaining the issues involved, maybe makes it more clear.

image

maiconpintoabreu:
> @maiconpintoabreu I can't remember at the moment if `rlSetTexture()` needs to be after `rlBegin()` or before but I wonder if everything still works on OpenGL 1.1, it could be sensible in that case...
> 
> This PR changes how it was designed originally... and I'm afraid it was structure that way for some reason... but it was so long ago that I can't remember...

Based on OpenGL 1.1 `rlSetTexture()` needs to be before `rlBegin()` 
I will check if I can do something different.

maiconpintoabreu:
Hi @raysan5 , is it ok to use `#if defined` to check for opengl 1.1 and swap `rlSetTexture` position?
Like:

```
#if defined(GRAPHICS_API_OPENGL_11)
    rlSetTexture(texture.id);
    rlBegin(RL_QUADS);
#else
    rlBegin(RL_QUADS);
    rlSetTexture(texture.id);
#endif
```

maiconpintoabreu:
Running on Opengl 1.1, The opengl 1.1 rlSetTexture does not need to be after rlBegin as it does not include the code that I am trying to fix.

The Devs that will need to make sure that the rlSetTexture is after rlBegin is the ones doing draw of multiple Modes RL_TRIANGLES and RL_QUADS, if it is only RL_QUARDS, the rlSetTExture can be anywhere.

image

maiconpintoabreu:
Changed the PR to draft as I am trying to test all the cases again

maiconpintoabreu:
Hi @raysan5, 

Tested and working:

`GRAPHICS_API_OPENGL_33` 
* `rlBegin -> rlSetTexture` [Working]

`GRAPHICS_API_OPENGL_11` 
* `rlSetTexture -> rlBegin` [Working]

`GRAPHICS_API_OPENGL_21` 
* `rlBegin -> rlSetTexture` [Working]

`GRAPHICS_API_OPENGL_43` 
* `rlBegin -> rlSetTexture` [Working]

`GRAPHICS_API_OPENGL_ES2` 
* `rlBegin -> rlSetTexture` [Working]

`rlBegin -> rlSetTexture` only needed when using a mode like RL_TRIANGLES and change to another mode like RL_QUADS or RL_LINES.

raysan5:
@maiconpintoabreu I'm afraid proposed solution is not ideal. raylib is intended to work the same way independently of the OpenGL version (despite feature limitations). And requesting different approaches depending on the version will confuse rlgl users. I prefer to keep the library with the current limitation.

casperbear:
"raylib is intended to work the same way independently of the OpenGL version" This seems to be exactly the purpose of this PR. I'm no specialist but if OpenGL 3 behavior is purposefully different from OpenGL 1 (batches vs simple state machine) then rlgl, being a wrapper layer dealing with implementation details, is the perfect place to abstract this away.

casperbear:
@raysan5 @maiconpintoabreu 
rlSetTexture (OpenGL 3.3 implementation) sets
RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].textureId = id;
and then rlBegin (OpenGL 3.3 implementation) sets
RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].textureId = RLGL.State.defaultTextureId;
I've no idea what defaultTextureId is in this context, but here is a PR that dealt with seemingly similar issue
https://github.com/raysan5/raylib/pull/4364/files

maiconpintoabreu:
Hi @casperbear, I am working on another PR to solve this issue without changing the order of settexture and begin.

https://github.com/raysan5/raylib/pull/5200

PR #5196, opened 2025-09-19
Use lowercase "message" to remove warning spam when building

Auios:
CMake is adding this in my terminal output when I build: "It is recommended to use lowercase CMake commands"
Making "MESSAGE" lowercase will help remove just a little bit more warning spam that we don't need so we can see actual warnings.

(Also it's messing with my OCD...)


raysan5:
@Auios thanks for the review!

PR #5181, opened 2025-09-13
[build][cmake] Review web compilation system

brccabral:
Changes following discussion from #5180  .  
Moved file `projects/CMake/core_basic_window.c` to `examples/others/web_basic_window.c`.  
Removed target options from library to examples.

PR #5198, opened 2025-09-21
[build][cmake] fix issue introduced by #5181

JohnnyCena123:
see [here](https://github.com/JohnnyCena123/raylib-projects/actions/runs/17729061729/job/50376069171#step:4:72) for real time demonstration of the issue


raysan5:
@Peter0x44 please, could you take a look? Not sure if ASYNCIFY should be forced, unless it is exclusively target for raylib examples

Also, it's quite confusing to me having examples specific build config inside raylib/src 🤔

Peter0x44:
this PR isn't forcing ASYNCIFY. the point of the public link options is so that anyone consuming this library from cmake automatically gets those flags applied, so it's okay for any flag that are absolutely REQUIRED 100% of the time to be there, but stuff like ASYNCIFY shouldn't be. I think this PR is correct.

The only thing I'm not completely sure about is if -sEXPORTED_RUNTIME_METHODS=ccall is necessary. I think older emscripten might need it? If so, I'm fine with that.

raysan5:
@JohnnyCena123 @Peter0x44 thanks you very much for the improvement and the review. Afair, `-sEXPORTED_RUNTIME_METHODS=ccall` was required for some specific cases but maybe emscripten addressed it and it is not required any more...

PR #5201, opened 2025-09-23
Fix: MatrixDecompose removes shear

ArmanOmmid:
Matrix Decompose now correctly removes shear. Prior implementation does not correctly orthogonalize the basis vectors to produce a proper rotation.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -2552,65 +2552,91 @@ RMAPI int QuaternionEquals(Quaternion p, Quaternion q)
     return result;
 }
 
-// Decompose a transformation matrix into its rotational, translational and scaling components
+// Decompose a transformation matrix into its rotational, translational and scaling components and remove shear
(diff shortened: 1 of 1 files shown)


ArmanOmmid:
I'm currently working on addressing some other bugs in the GLTF loading code. 
I have those fixes tested and working and am now working on integrating the fixes gracefully.
This fix was more isolated so I thought I'd drop it in early :)

raysan5:
@ArmanOmmid Thanks for the review, I'm not an expert in math so I can not easely verify the new implementation is right and works for all use cases... but I'm merging it.

raysan5:
@ArmanOmmid I imagine you tried the raylib gltf animation examples and they still work as expected...

ArmanOmmid:
> @ArmanOmmid I imagine you tried the raylib gltf animation examples and they still work as expected...

Yes both the gltf animations (robot.glb and greenman.glb) still work as expected :) Thanks!

PR #5202, opened 2025-09-24
[examples] Fix textures polygon drawing example for opengl 1.1

maiconpintoabreu:
Issue was only happening on Opengl 1.1

Before reorder:

image

After reorder:

image


raysan5:

PR #5204, opened 2025-09-24
[rcore_android] implement GetCurrentMonitor()

keks137:
literally just calls https://developer.android.com/reference/android/view/Display#getDisplayId() using the jni
also not sure if LOG_ERROR is overkill, but the windows implementation does the same


raysan5:
@keks137 thanks! looks good!

PR #5205, opened 2025-09-25
Update palette_switch.fs

dog6:
Removed unnecessary `a` in palette_switch.fs `note` comment


raysan5:
@dog6 thanks for the review!

PR #5206, opened 2025-09-25
[rmodels] GLTF Anim Correctly Inherits World Transform

ArmanOmmid:
I'm only looking at GLTF at the moment, but possible that the IQM logic would also appreciate a similar change given that it uses the same `BuildPoseFromParentJoints`.

Meanwhile, consider the following helper functions below (and exposing them with `RLAPI`)
I think these would be useful tools for animators to be able to use. I am currently using these for machine learning for animation on my personal fork.

Just let me know where you'd like me to put them (and their header declarations) if you agree :)

```
// Compose Model Space Transform
Transform ComposeTransform(Transform transform, Transform parentTransform)
{
    transform.rotation = QuaternionMultiply(parentTransform.rotation, transform.rotation);
    transform.scale = Vector3Multiply(transform.scale, parentTransform.scale);
    transform.translation = Vector3Multiply(transform.translation, parentTransform.scale);
    transform.translation = Vector3RotateByQuaternion(transform.translation, parentTransform.rotation);
    transform.translation = Vector3Add(transform.translation, parentTransform.translation);
    return transform;
}

// Extract Local Space Transform
Transform ComposeTransformInverse(Transform transform, Transform parentTransform)
{
    transform.rotation = QuaternionMultiply(QuaternionInvert(parentTransform.rotation), transform.rotation);
    transform.scale = Vector3Divide(transform.scale, parentTransform.scale);
    transform.translation = Vector3Subtract(transform.translation, parentTransform.translation);
    transform.translation = Vector3RotateByQuaternion(transform.translation, QuaternionInvert(parentTransform.rotation));
    transform.translation = Vector3Divide(transform.translation, parentTransform.scale);
    return transform;
}
```

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4261,9 +4261,10 @@ static void BuildPoseFromParentJoints(BoneInfo *bones, int boneCount, Transform 
                 continue;
             }
             transforms[i].rotation = QuaternionMultiply(transforms[bones[i].parent].rotation, transforms[i].rotation);
+            transforms[i].scale = Vector3Multiply(transforms[i].scale, transforms[bones[i].parent].scale);
+            transforms[i].translation = Vector3Multiply(transforms[i].translation, transforms[bones[i].parent].scale);
             transforms[i].translation = Vector3RotateByQuaternion(transforms[i].translation, transforms[bones[i].parent].rotation);
             transforms[i].translation = Vector3Add(transforms[i].translation, transforms[bones[i].parent].translation);
-            transforms[i].scale = Vector3Multiply(transforms[i].scale, transforms[bones[i].parent].scale);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ArmanOmmid:
Also yes, these were tested on Raylib GLTFs and others including some examples from Khronos Group.
The fix is apparent when using models whose world transform is not identity, for example:
[KhronosCesiumMan](https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/CesiumMan/glTF-Binary/CesiumMan.glb)
[KhronosRiggedFigure](https://github.com/KhronosGroup/glTF-Sample-Models/blob/main/2.0/RiggedFigure/glTF-Binary/RiggedFigure.glb)

raysan5:
@ArmanOmmid At the moment I prefer to keep the API simple and expose those functions... but you can add them to `rmodels.c` as a reference and maybe for the future.

ArmanOmmid:
> @ArmanOmmid At the moment I prefer to keep the API simple and expose those functions... but you can add them to `rmodels.c` as a reference and maybe for the future.

@raysan5 
I can add them to `rmodels.c` for now, can still be useful for internal anim function (i.e the ones modified in this PR).

Do you have a preference for where you'd like them to be in the file? Or just use my judgement on what location makes sense.

PR #5207, opened 2025-09-26
[rlgl] Allow tint colors in GL_LINE (wires) and GL_POINT (points) draw modes on OpenGL 1.1 clean

meisei4:
clean up 5177 PR after failed rebase

Original PR: https://github.com/raysan5/raylib/pull/5177

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -686,6 +686,10 @@ RLAPI void rlDisableScissorTest(void);                  // Disable scissor test
 RLAPI void rlScissor(int x, int y, int width, int height); // Scissor test
 RLAPI void rlEnablePointMode(void);                     // Enable point mode
 RLAPI void rlDisablePointMode(void);                    // Disable point mode
+#if defined(GRAPHICS_API_OPENGL_11)
+RLAPI void rlSetPointSize(float size);                  // Set the point drawing size
+RLAPI float rlGetPointSize(void);                       // Get the point drawing size
+#endif
 RLAPI void rlEnableWireMode(void);                      // Enable wire mode
 RLAPI void rlDisableWireMode(void);                     // Disable wire mode
 RLAPI void rlSetLineWidth(float width);                 // Set the line drawing width
@@ -2016,6 +2020,18 @@ float rlGetLineWidth(void)
     return width;
 }
 
+#if defined(GRAPHICS_API_OPENGL_11)
+// Set the point drawing size
+void rlSetPointSize(float size) { glPointSize(size); }
+
+// Get the point drawing size
+float rlGetPointSize(void)
+{
+    float size = 0;
+    glGetFloatv(GL_POINT_SIZE, &size);
+    return size;
+}
+#endif
 // Enable line aliasing
 void rlEnableSmoothLines(void)
 {
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1436,7 +1436,7 @@ void DrawMesh(Mesh mesh, Material material, Matrix transform)
     if (mesh.animNormals) rlEnableStatePointer(GL_NORMAL_ARRAY, mesh.animNormals);
     else rlEnableStatePointer(GL_NORMAL_ARRAY, mesh.normals);
 
-    rlEnableStatePointer(GL_COLOR_ARRAY, mesh.colors);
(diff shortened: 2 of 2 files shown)


meisei4:
**add `rlSetPointSize/rlGetPointSize`.**
- OPENGL_11 point size API (`rlgl.h`):  enables fixed-function point sizing via `glPointSize`, mirroring raylib’s existing `rlSetLineWidth` and matching `rlEnablePointMode()` expectations on OPENGL_11

---
**`UploadMesh`: make texcoords optional (with "safe" defaults)**
  - If `mesh->texcoords` exists, upload/enable. otherwise set default `(0,0)` and disable the attribute (Same pattern taken from existing missing normals/tangents behavior)
---
**`DrawMesh` add better null checks, improve color default behavior**
- (OPENGL_11): only enable `GL_TEXTURE_COORD_ARRAY`, `GL_NORMAL_ARRAY`, and `GL_COLOR_ARRAY` when data is provided/exists. (Same pattern taken from existing vertices, animated or not, etc)
- (OPENGL_33/OPENGL_ES_2) When the shader doesn’t expose `SHADER_LOC_COLOR_DIFFUSE`, set a default vertex color from `material.maps[DIFFUSE].color` and disable the color attrib. (Same pattern taken from existing attribute default handling behavior, just moved)
---
**`LoadOBJ`: make texcoords optional with defaults, and remove color allocation (fallback on tint for all OpenGL contexts)**
 -  Read and flip `texcoords` only if `vt`'s exists in the OBJ file, otherwise default to `(0,0)` (Same pattern taken from existing normals behavior), allows for `texcoords` to be truly optional.
 -  remove forced per-vertex color allocation, let raylib fall back to `material` color when `mesh.colors == NULL`. Same behavior, but now allows for expected fixed-function color path on OPENGL_11 (Color array comes from tint during DrawMesh, after begin loaded from OBJ.

---
TODO: 
* **`DrawModelPoints`: clarify more opengl context limitations with proper guard flags to reduce confusion**

I would like to go through and perhaps add better gating to the feature opengl1.1 features set distinguishing API prototypes/definitions, it seems very helpful when teaching about opengl compatability, and opengl1.1 is a great place to start with this i believe

src/rlgl.h
```diff
@@ -686,6 +686,10 @@ RLAPI void rlDisableScissorTest(void);                  // Disable scissor test
 RLAPI void rlScissor(int x, int y, int width, int height); // Scissor test
 RLAPI void rlEnablePointMode(void);                     // Enable point mode
 RLAPI void rlDisablePointMode(void);                    // Disable point mode
+#if defined(GRAPHICS_API_OPENGL_11)
```
raysan5: Expected approach is to provide a common API for all OpenGL versions and check the version flags inside the function body, please, review it.

meisei4: Alright, thanks for clarification, I will also update that for `rlEnableStatePointer` and `rlDisableStatePointer` (where i saw it first) https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L647

raysan5:
@meisei4 Is this PR reviewed and tested, ready to merge?

meisei4:
It has not been reviewed but I have tested it locally for the following main cases:
opengl11: tint is now applied to each draw model mode (wires, vs points, vs fill)
opengl3.3~es2: tint works the exact same as before (white is default filled initially for loadOBJ -- something i missed initially but fixed in the last commit)

 I am unsure how rigorously testing with raylib is involved, but I can say that `./models_loading`, ` ./models_mesh_generation`, .`/models_textured_cube` are all working
 
But so i am rather confident it does not change any existing behavior other than:
 1. fixing the opengl11 tint color issue
 2. adding the point mode for opengl11 
 3. `LoadOBJ`  is no longer requires texcoords to work (most all tests involve OBJ having texcoords as expected but i have simple obj files i test with that do not involve texcoords, only bare minimum vertices <- and they work as i am expecting with just zeroed texcoords that can be updated later through buffer updates)
 
So while I believe that it is ok for merging, i should also mention that am still going to be working with this closely when dealing with comparing opengl11 behavior to PS2GL behavior so i will also be able to continue to test with opengl3.3 as I continue to do so.

raysan5:
@meisei4 thanks for the details and the review!

PR #5208, opened 2025-09-28
Update miniaudio to v0.11.23

M374LX:


raysan5:
@M374LX thanks for the update! I imagine you verified all raylib audio examples keep working, right?

M374LX:
I did not actually check all audio examples, as the likelihood of any of them to stop working is very low.

**Update:** It looks like the examples that load an audio file actually stopped working, but I am checking.

**Update 2:** Everything seems to be working again after running ``make clean`` and rebuilding everything. Maybe I accidentally used an outdated raylib version.

**Update 3:** Confirmed. All audio examples work as expected.

PR #5213, opened 2025-09-29
[examples] Add ASCII post processing effect

maiconpintoabreu:
Adding an example shader using character bitmask by color to draw using ascii symbols.

Added for: https://github.com/raysan5/raylib/discussions/5209

Tested on Desktop:

image

Tested on Web:

image


raysan5:
@maiconpintoabreu thanks for the new example! 😄

raysan5:
@maiconpintoabreu I'm checking the example and I don't understand how it works, where is the font used? how it is selected? It would be nice, considering it is a learning example, to have more details explained in the shader and not just a copy from somewhere else.

Note the ASCII rendering looks a bit off, characters are not clear...

maiconpintoabreu:
Hi @raysan5 I will add more comments so, the bitmaps I got from https://www.shadertoy.com/view/lssGDj, he developed a bitmap maker: https://thrill-project.com/archiv/coding/bitmap/, about how it work is simple, based on the gray level it will select one of the characters and the color is based on the cell size and the color of the cell.
```
    vec2 cellUV = floor(fragTexCoord / uvCellSize) * uvCellSize;

    vec3 cellColor = texture2D(texture0, cellUV).rgb;
```

About the ASCII a big off is because I set the fontsize too low, it should be a min of 5, I was testing with different numbers and I forgot to set it back.

raysan5:
@maiconpintoabreu that's the kind of information that should be explained in the shader (also reviewed shader formatting for glsl330 version), you can also include the links, so other users trying to learn how it works will have that info available. raylib examples are intended to be educational, for users without much experience on the graphics/shader fields.

maiconpintoabreu:
Fixed a bug in the process but this is the expected result:

image

PR #5214, opened 2025-09-29
[rtextures] fix align order initialization in Image

psxdev:
We have detected some issues with old console platforms when initialization for Image structure is not done in Image structure definition order.

Using defined order fix the issue.


raysan5:
@psxdev thanks for the review! 👍😄

PR #5217, opened 2025-09-30
[examples] Update `shapes_bouncing_ball` with gravity

Jopestpe:
Adds a new example based on the classic bouncing ball, modified to include gravity and floor damping. 

Added for: #5209

Tested on Desktop (Linux).

screenshot


raysan5:
@Jopestpe It would be nice if gravity could be enabled/disabled

raysan5:
@Jopestpe thanks for the improvement!

PR #5218, opened 2025-09-30
ADDED: example: `shapes_bullet_hell`

zerohorsepower:
# Description

This PR adds an example that creates a basic bullet hell circular pattern. It aims to demonstrate how to move objects using `direction` and `speed` (trigonometry) and it exemplify how basic bullet hell patterns can be made with loops, rows and timers.

It also allows the user to manipulate many variables in runtime to demonstrate that is possible to create many different basic patterns by only changing values.

image

# Changelog

- Added `examples/shapes/shapes_bullet_hell.c` - Main example code
- Added `examples/shapes/shapes_bullet_hell.png` - Screenshot

Tested and it compiles successfully on Linux. The performance on web build is very close to bunnymark example.


raysan5:
@zerohorsepower thanks! nice addition!

PR #5220, opened 2025-09-30
[examples] Corrected typo in core_input_mouse

RobinsAviary:
I noticed while porting the example to Odin that the boolean check of whether the mouse is visible is actually the opposite of what it should be.


raysan5:
@RobinsAviary Good catch! Thanks for the review!

PR #5223, opened 2025-09-30
[examples] REVIEW: `core_3d_camera_free` uses `IsKeyPressed()` incorrectly

RobinsAviary:
Hello! I noticed core_3d_camera_free uses a char in place of the constants that it is supposed to be using. I don't see this repeated in other examples, so I think it would be better to replace it with the constant to prevent future confusion.

### DIFF
--- a/examples/core/core_3d_camera_free.c
+++ b/examples/core/core_3d_camera_free.c
@@ -49,7 +49,7 @@ int main(void)
         //----------------------------------------------------------------------------------
         UpdateCamera(&camera, CAMERA_FREE);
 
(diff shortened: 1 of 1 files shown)


raysan5:
@RobinsAviary Good catch! Actually, it also used to work this way in the past, long time ago...

RobinsAviary:
> @RobinsAviary Good catch! Actually, it also used to work this way in the past, long time ago...

The irony is it still technically works, I got it to compile and it still technically functioned the same on my pc LOL. I'm assuming the ascii codes of each char just happen to line up with the constants tho

PR #5224, opened 2025-10-01
FIXED: rexm's makefile and default paths

Teeto44:
Fixed a windows only backslash in
rexm's makefile causing build errors on macOS and linux

Added unix paths to rexm for better compatibility

### DIFF
--- a/tools/rexm/rexm.c
+++ b/tools/rexm/rexm.c
@@ -205,12 +205,22 @@ int main(int argc, char *argv[])
     exCollectionFilePath = getenv("REXM_EXAMPLES_COLLECTION_FILE_PATH");
     exVSProjectSolutionFile = getenv("REXM_EXAMPLES_VS2022_SLN_FILE");
 
+#if defined(_WIN32)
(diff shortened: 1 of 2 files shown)


tools/rexm/rexm.c
```diff
@@ -436,7 +446,7 @@ int main(int argc, char *argv[])
             SaveFileText(TextFormat("%s/%s/%s.c", exBasePath, exCategory, exName), exTextUpdated[1]);
             for (int i = 0; i < 6; i++) { MemFree(exTextUpdated[i]); exTextUpdated[i] = NULL; }
             UnloadFileText(exText);
-        }
+        } break;
```
raysan5: This break shouldn't be here, when `create` a new example it should also be `added`

Teeto44:
Ok just pushed a new version, I haven't been able to test this on other platforms nor with raylib.com present but I don't think either of those contexts should be broken.

raysan5:
@Teeto44 thanks for the update!

PR #5225, opened 2025-10-01
[examples] REVIEW: `shapes_collision_area.c` has an uneccessary include.

RobinsAviary:
Hello! I noticed while porting `shapes_collision_area.c` to Odin that it doesn't actually use the include that it claims it needs.


raysan5:
@RobinsAviary thanks for the review!

PR #5226, opened 2025-10-02
swap TraceLog to TRACELOG

JohnnyCena123:


raysan5:
@JohnnyCena123 Good catch!

PR #5232, opened 2025-10-03
C++ compiler support

sakgoyal:
Allow C++ compilers to compile this code without separately linking as a C library.

Changes:
* Backport RGFW unicode fix
  * https://github.com/ColleagueRiley/RGFW/commit/0ebeb37b7fc311a5ed5ed9a28404bfddffde43f3
* Add C-style casting
* Fix a initializer list being in the wrong order (C++ requires declaration order)
* GLFW currently has issues with cross compiling so I am using RGFW for now 
  * https://github.com/glfw/glfw/pull/2777
* Change C Aggregate initializer to C++ default initializer. 
* Remove duplicate key in `keyMappingRGFW`

```powershell
clang++ -static --std=c++23 -D PLATFORM_DESKTOP_RGFW -D _CRT_SECURE_NO_WARNINGS \
.\src\main.cpp .\vendor\raylib\src\raudio.c .\vendor\raylib\src\rcore.c ... \
-lgdi32 -lkernel32 -lmsvcrt -luser32 -lopengl32 -Xlinker /NODEFAULTLIB:libcmt
```

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -519,8 +519,14 @@ static void RecordAutomationEvent(void); // Record frame events (to internal eve
 #endif
 
 #if defined(_WIN32) && !defined(PLATFORM_DESKTOP_RGFW)
+#ifdef __cplusplus
+extern "C" {
+#endif
 // NOTE: We declare Sleep() function symbol to avoid including windows.h (kernel32.lib linkage required)
 __declspec(dllimport) void __stdcall Sleep(unsigned long msTimeout); // Required for: WaitTime()
+#ifdef __cplusplus
+}
+#endif
 #endif
 
 #if !defined(SUPPORT_MODULE_RTEXT)
@@ -2522,7 +2528,7 @@ int MakeDirectory(const char *dirPath)
     // Create final directory
     if (!DirectoryExists(pathcpy)) MKDIR(pathcpy);
     RL_FREE(pathcpy);
-    
+
     // In case something failed and requested directory
     // was not successfully created, return -1
     if (!DirectoryExists(dirPath)) return -1;
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ packages/
 *.so
 *.so.*
 *.dll
+*.obj
 
 # Emscripten
 emsdk
--- a/src/external/win32_clipboard.h
+++ b/src/external/win32_clipboard.h
@@ -13,7 +13,7 @@ unsigned char* Win32GetClipboardImageData(int* width, int* height, unsigned long
 #include <stdlib.h>
 #include <assert.h>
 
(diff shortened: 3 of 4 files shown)


sakgoyal:
I would appreciate it if somebody could review this and suggest how I can improve this. I dont think my solutions for aggregate initializers is good, but I am not sure how to improve it.

raysan5:
@sakgoyal Thanks for the review, I see multiple improvements that can be implemented but I'm afraid external libraries should not be modified on raylib side because once updated to latest versions, changes will be lost. Still, some of the them are managed by me so only libraries managed by other developers will suffer that fate.

I also see some specific changes that I don't understand why are required. Added some notes.

Also, about coding conventions, raylib uses a space before asterix on casting, please, could you review that? (i.e. `(float*)`--> `(float *)`.

src/raudio.c
```diff
@@ -1062,7 +1062,12 @@ bool ExportWave(Wave wave, const char *fileName)
     else if (IsFileExtension(fileName, ".wav"))
     {
         drwav wav = { 0 };
+        #ifdef __cplusplus
+        drwav_data_format format = {};
```
raysan5, line 1066: Why this is required? I prefer to avoid it.

src/rcore.c
```diff
@@ -166,11 +166,17 @@
         #define MAX_PATH 1025
     #endif
 struct HINSTANCE__;
+#if defined(__cplusplus)
+extern "C" {
```
raysan5, line 170: Mmmh... why is this required? Because name mangling?

src/rmodels.c
```diff
@@ -5208,7 +5208,11 @@ static Image LoadImageFromCgltfImage(cgltf_image *cgltfImage, const char *texPat
                 int outSize = numberOfEncodedBits/8 ;                           // Actual encoded bytes
                 void *data = NULL;
 
+                #ifdef __cplusplus
+                cgltf_options options = {};
```
raysan5, line 5212: I prefer to avoid the __cplusplus check only for that.

sakgoyal:
I will revert GLFW and qoa. 

is `jar_*` managed by you? I couldn't find it anywhere else.

my fix for `tinyobj_loader` was actually fixed 3 years ago. seems like that update was not pulled in. 

RGFW seems to be introducing some breaking changes for v1.8 (removing gamepad support, and more). so that will require a rework of the RGFW backend in raylib. for now, I think backporting this change is useful for now while 1.8 is being worked on

raysan5:
@sakgoyal Yees, `jar_*` is mostly managed by me at this moment... or afaik, it is not available in an other repo. The tinyobj_loader can be updated.

sakgoyal:
Oh man I messed up the force push. Let me try fixing this. I am going to close this an open a new PR. Not sure how to resolve this.

PR #5235, opened 2025-10-05
[rcore] Check if mode is valid on `GetMonitorRefreshRate` to avoid Segfault when removing Monitor

maiconpintoabreu:
This PR is to fix the issue when removing a Monitor while getting information about it.
The issue was confirmed on Linux and Windows.

image

image

Renamed `vidmode` to `mode` to match with methods like `GetMonitorHeight` and `GetMonitorWidth`.


raysan5:
@maiconpintoabreu Nice! Thanks for the review!

PR #5241, opened 2025-10-08
include `raylib.h` and `rcore.c` in platform files

JohnnyCena123:
this has already been done in `rcore_desktop_rgfw.c`, but not perfectly as the comments state, its there for LSPs


raysan5:
@JohnnyCena123 that's not needed for the other plaftorms

JohnnyCena123:
> @JohnnyCena123 that's not needed for the other plaftorms

why exactly?

PR #5242, opened 2025-10-08
remove duplicate entry from `keyMappingRGFW`

JohnnyCena123:


raysan5:
@JohnnyCena123 thanks for the review

PR #5244, opened 2025-10-08
[rmodels] `UpdateModelAnimation()` reviewed matrix-transpose, only once per bone

Arrangemonk:
i updated UpdateModelAnimation  to calculate the inverted bone matrix only once per bone instead of for every vertex

i measured time and its only twice as fast as before, so its not much of an improvement but still better than nothing

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -2352,6 +2352,8 @@ void UpdateModelAnimation(Model model, ModelAnimation anim, int frame)
         Mesh mesh = model.meshes[m];
         Vector3 animVertex = { 0 };
         Vector3 animNormal = { 0 };
+        Matrix boneMatrix = { 0 };
+        Matrix InverseBoneMatrix = { 0 };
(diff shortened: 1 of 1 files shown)


raysan5:
@Arrangemonk thanks for the review!

JeffM2501:
Was this PR tested on any kind of complex model?
I understand what it's trying to solve, but the bone weight check is in the wrong place and will early out incorrect.
This looks like it only iterates 4 bones per model, not 4 bones per vertex.

I would remove this PR until the questions are answered or the code is fixed.

raysan5:
@Arrangemonk Please, could you check the questions from Jeff? We have some concerns about the correctness of this optimization...

Arrangemonk:
@JeffM2501 @raysan5 , the tests with the robot model were misleading, im sorry for causing inconvenience, in the future im going to test my ideas thoroughly, before creating a pull request

PR #5245, opened 2025-10-09
[examples] Update `core_monitor_change` to remove warnings

maiconpintoabreu:
Set name as const as it will not change and avoids the warning of  `initialization discards ‘const’ qualifier` and removed not in use variable `currentMonitor`.


raysan5:
@maiconpintoabreu thansk for the review!

Issue #4842, opened 2025-03-15
[rcore][drm] `GetTouchPointCount()` returns 1 when screen is not touched


larry104:
Just did a git pull to get the latest; compiled on a RaspberryPi 3 with make PLATFORM=PLATFORM_DRM

To reproduce just add to the core_input_multitouch.c a statement to print tCount and compile and run it.

When touching the screen first with your index finger when with your thumb  and then releasing the index finger and after that the thumb the tCount will still show 1 even though nothing touches the screen anymore.

maiconpintoabreu:
Hi @larry104 looks like this behaviour is provoked by 

`PollInputEvents(void)`
...
`CORE.Input.Touch.position[0] = CORE.Input.Mouse.currentPosition;`
And `rcore_drm` is checking if position is different than 0 instead of keeping the `CORE.Input.Touch.pointCount` based on events
```
        // Update touch point count
        CORE.Input.Touch.pointCount = 0;
        for (int i = 0; i < MAX_TOUCH_POINTS; i++)
        {
            if (CORE.Input.Touch.position[i].x >= 0) CORE.Input.Touch.pointCount++;
        }
```

I do not have how to test it so I cannot try to fix it.

MUL7ii:
@raysan5 I have made a PR #5279 could you look into it , if there are any changes let me know.

CrackedPixel:
@larry104 @raysan5 can this be closed? the PR was merged

larry104:
I don't think these issues are related nor does the  CORE.Input.Touch.pointCount = 0 fix the reported issue. You can check with the multitouch example core_input_multitouch.c. 

1) When you touch the screen with one finger and stop moving the finger around GetTouchPointCount() get's reset to 0 but the moment you start moving your finger again it gets back to 1.

2) If you touch the screen with first one finger then add the second. Now lift of the first and GetTouchPointCount()  will be still 2 (of course when moving the finger around)

It looks to me touchpoints do not get tracked properly in general especially when you don't have any movement.

Let me know if you need more explanation or want me to do tests or a take video  for print any debug info while touching.

MUL7ii:
@larry104 I made changes could you check the pr merged and test it?

larry104:
I only see the master branch which was updated 7 hours ago as of this writing .. are you talking about that? if yes, no changes in there.

MUL7ii:
@larry104 check this #5279 PR

larry104:
[...]

PR #5247, opened 2025-10-09
Update rcore_drm.c

aSh-dev-io:
Title: Fix: [rcore][drm] GetTouchPointCount() returns 1 when screen is not touched #4842

Description:
This pull request resolves issue #4842. The bug was causing GetTouchPointCount() to incorrectly return 1 on the DRM platform, even when no touch input was present.

The fix is to add CORE.Input.Touch.pointCount = 0; at the beginning of the PollInputEvents() function. This ensures that the touch point count is correctly reset at the start of every frame, preventing the use of stale or incorrect data.

Testing and Validation
I have personally tested this fix on a Raspberry Pi 4 with an official 7" touchscreen display, which is a standard DRM platform environment.

The test was successful:

Without touching the screen, the GetTouchPointCount() function now correctly and consistently returns 0.

When touching the screen with one finger, the function immediately returns 1.

The fix works as expected and completely resolves the bug. The code is now ready to be merged.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -1072,15 +1072,7 @@ const char *GetKeyName(int key)
     TRACELOG(LOG_WARNING, "GetKeyName() not implemented on target platform");
     return "";
 }
-
-// Register all input events
-void PollInputEvents(void)
-{
-    CORE.Input.Touch.pointCount = 0;   // <-- ADD THIS LINE HERE
-    
-    // ... the rest of the original code follows
-}
-
+CORE.Input.Touch.pointCount = 0;
 {
 #if defined(SUPPORT_GESTURES_SYSTEM)
     // NOTE: Gestures update must be called every frame to reset gestures correctly


src/platforms/rcore_drm.c
```diff
@@ -1075,6 +1075,12 @@ const char *GetKeyName(int key)
 
 // Register all input events
 void PollInputEvents(void)
+{
```
raysan5: Please, could you remove the open-close brackets? the function already contains them

src/platforms/rcore_drm.c
```diff
@@ -1075,6 +1075,12 @@ const char *GetKeyName(int key)
...
+    CORE.Input.Touch.pointCount = 0;   // <-- ADD THIS LINE HERE
```
raysan5: Please, could you remove uneedded comments?

src/platforms/rcore_drm.c
```diff
@@ -1124,7 +1130,7 @@ void PollInputEvents(void)
         CORE.Input.Touch.currentTouchState[i] = platform.currentButtonStateEvdev[i];
     }
 
-    // Register gamepads buttons events
+    // Register gamepads buttons events in
```
raysan5, line 1125: That `in` at the end seems confusing, please, remove it.

src/platforms/rcore_drm.c
aSh-dev-io: ok

aSh-dev-io:
i  hope ive completed could you please review it approve my contribution

raysan5:
I'm afraid the change broke the library

aSh-dev-io:
> I'm afraid the change broke the library

could u  please merge it..!

PR #5255, opened 2025-10-10
[examples] Added `shapes_starfield`

themushroompirates:


raysan5:
@themushroompirates thanks for the new example

PR #5260, opened 2025-10-12
[examples] Added: `shapes_particles`

JordSant:
Added new example shapes_particles


raysan5:
@JordSant thanks for the new example!

PR #5265, opened 2025-10-13
change free to RL_FREE

CrackedPixel:
oops :sweat_smile:


raysan5:
@CrackedPixel good catch! thanks!

PR #5268, opened 2025-10-14
Added colored logging using ANSI escape code coloring

Robert-DeForrest-Reynolds:
I added colored logging using ANSI escape codes. I've tested it on Windows 11 (cmd, powershell, and msys), as well as the common Linux distros (Ubuntu, Debian, Arch), and it seems to work perfectly fine. Have not tested on mobile devices.
I changed the example for custom logging to demonstrate how to use the ANSI colors.
The buffer takes in a few less characters, as the "LOG_TYPE: " is simply just printf'd, and I think that's okay. I can't see any negative implications there. Hope all is well, thank you.

**Short review:**
_In `src/raylib.h`:_
Added ANSI colors

_In `src/utils.c`:_
`strcpy(buffer, "LOG_TYPE: ")` -> `printf(ANSI_CODE "LOGTYPE: ")`

_In `examples/core_custom_logging.c`:_
Change the example to demonstrate using ANSI colors for custom logger

**PS**: this is my first ever pull request, so don't hurt me. I tried to make it small, but "meaningful".

Screenshot 2025-10-14 055004

### DIFF
--- a/examples/core/core_custom_logging.c
+++ b/examples/core/core_custom_logging.c
@@ -57,7 +57,7 @@ int main(void)
     const int screenHeight = 450;
 
     // Set custom logger
-    // SetTraceLogCallback(CustomLog);
+    SetTraceLogCallback(CustomLog);
 
     InitWindow(screenWidth, screenHeight, "raylib [core] example - custom logging");
 


raysan5:
@Robert-DeForrest-Reynolds Thanks but I'm afraid this change is not portable, not all terminals support it, already been proposed a couple of times in the past. I prefer to avoid it in raylib main library but maybe it can be added to custom logging example?

sleeptightAnsiC:
> I've tested it on Windows 11 (cmd, powershell, and msys), as well as the common Linux distros (Ubuntu, Debian, Arch), and it seems to work perfectly fine.

This would not work on older Windows versions and with some terminals, see: https://stackoverflow.com/a/63913131/19562526

You're also not checking if stdout stream is connected to terminal. You need to use `isatty` for that (this function is slightly different on Win32 and POSIX systems).

Anyway, it's better to provide your own logging with [SetTraceLogCallback](https://github.com/raysan5/raylib/blob/1b5a14e5167755d54e8b4e9851b9167302641d2f/src/raylib.h#L1111-L1113) instead of changing anything in Raylib itself.

PR #5269, opened 2025-10-14
[RCORE] Fix windows forward declarations so all windows platforms work

JeffM2501:
The windows function forward declarations were not being done for RGFW and GCC (only MSVC), they need to be there for any compiler for RGFW to work. This PR enables them for all windows platforms.


raysan5:

PR #5272, opened 2025-10-15
Added `void rlClearDepthBuffer(void)` to clear the depth buffer. 

8bitprodigy:
Very useful for making FPS games.

Beats having to `#import <GL/gl.h>` just to call one function.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -2108,6 +2108,13 @@ void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsigned ch
     glClearColor(cr, cg, cb, ca);
 }
 
+// Clears the depth buffer. Useful for drawing first person weapons over the
+// rendered scene without them clipping through world geometry.
+void rlClearDepthBuffer(void)
+{
+    glClear(GL_DEPTH_BUFFER_BIT); // Beats having to #import <GL/gl.h> just to call this one function.
+}
+
 // Clear used screen buffers (color and depth)
 void rlClearScreenBuffers(void)
 {


raysan5:
glClearScreenBuffers() already does it, I think an additional function is not needed

8bitprodigy:
> glClearScreenBuffers() already does it, I think an additional function is not needed

That clears the colors, too, though, so what'd even be the point of that?...

This function clears *JUST* the depth buffer which is handy for ANYONE making an FPS and needs a way to draw the HUD weapon on top of the scene such that it won't clip through scene geometry.

Did you even read the submitted code?

raysan5:
Yes, I read the code and noone in 12 years has needed this function because there are other mechanisms to accomplish the same result. Still if you need it for your specific use case, you can call OpenGL directly.

raysan5:
Please, note that rlgl does not pretend to be acomplete replacement for OpenGL.

src/rlgl.h
```diff
@@ -2108,6 +2108,13 @@ void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsigned ch
     glClearColor(cr, cg, cb, ca);
 }
 
+// Clears the depth buffer. Useful for drawing first person weapons over the
+// rendered scene without them clipping through world geometry.
+void rlClearDepthBuffer(void)
+{
+    glClear(GL_DEPTH_BUFFER_BIT); // Beats having to #import <GL/gl.h> just to call this one function.
```
CrackedPixel, line 2115: i would remove this comment from the PR

8bitprodigy: Fair enough.

PR #5273, opened 2025-10-15
[rlgl] ADDED: `rlClearScreenBuffersEx()` to clear the depth buffer (and clear other buffers good, too)

8bitprodigy:
Very useful for making FPS games, as it allows you to draw the HUD weapon over top the rendered scene without it clipping through world geometry.

`rlClearScreenBuffers()` clears ***BOTH*** the color and depth buffers, so if you were to call that, you'd wipe away the entire frame you just rendered just to draw the weapon on top of it, which would defeat the point entirely.

This single function solves not having to import `<GL/gl.h>` just to call one function AND allows for rendering 3D stuff over top the scene, such as HUD weapons in FPS games or the floating arrow in Crazy Taxi  on systems which only support OpenGL 1.1, as they can't make use of RenderTextures.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -2126,7 +2126,7 @@ void rlClearScreenBuffersEx(rlBufferBitFlags flags)
 // Clear used screen buffers (color and depth)
 void rlClearScreenBuffers(void)
 {
-    rlClearScreenBuffersEX(RL_COLOR_BUFFER_BIT | RL_DEPTH_BUFFER_BIT);     // Clear used buffers: Color and Depth (Depth is used for 3D)
+    rlClearScreenBuffersEx(RL_COLOR_BUFFER_BIT | RL_DEPTH_BUFFER_BIT);     // Clear used buffers: Color and Depth (Depth is used for 3D)
     //glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);     // Stencil buffer not used...
 }
 


JeffM2501:
Why not make it rlClearScreenBuffersEX and have it take bitmasks for color and depth (and stencil)? then it could be use to clear just the color or just the depth, or both. That makes it useable in more situations

8bitprodigy:
> Why not make it rlClearScreenBuffersEX and have it take bitmasks for color and depth (and stencil)? then it could be use to clear just the color or just the depth, or both. That makes it useable in more situations

That's an even better idea! I could do that if Ray says I ought to do it that way.

8bitprodigy:
> Why not make it rlClearScreenBuffersEX and have it take bitmasks for color and depth (and stencil)? then it could be use to clear just the color or just the depth, or both. That makes it useable in more situations

Updated to reflect this superior idea: https://github.com/raysan5/raylib/pull/5273/files

src/rlgl.h
```diff
@@ -705,6 +712,7 @@ RLAPI bool rlIsStereoRenderEnabled(void);               // Check if stereo rende
 
 RLAPI void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsigned char a); // Clear color buffer with color
 RLAPI void rlClearScreenBuffers(void);                  // Clear used screen buffers (color and depth)
+RLAPI void rlClearScreenBuffersEX(rlBufferBitFlags flags); // Clear used screen buffers (selectable with '|'-separated flags)
```
CrackedPixel: @8bitprodigy 
Can you change this `EX` to `Ex`? Thats the style raylib uses in other places (I know jeff wrote it as `EX` but i believe that was a typo)

CrackedPixel: Sample: https://github.com/raysan5/raylib/blob/master/src/raylib.h#L1263

8bitprodigy: Can do!

8bitprodigy: Done!

src/rlgl.h
```diff
@@ -2108,10 +2116,17 @@ void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsigned ch
...
+void rlClearScreenBuffersEx(rlBufferBitFlags flags)
+{
+    glClear(flags);
+}
+
 // Clear used screen buffers (color and depth)
 void rlClearScreenBuffers(void)
 {
-    glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);     // Clear used buffers: Color and Depth (Depth is used for 3D)
+    rlClearScreenBuffersEX(RL_COLOR_BUFFER_BIT | RL_DEPTH_BUFFER_BIT);     // Clear used buffers: Color and Depth (Depth is used for 3D)
```
CrackedPixel: @8bitprodigy missed a spot :stuck_out_tongue: 

make sure you test compile the code whenever you have a PR -- ray has been burned on this in the past :rofl:

8bitprodigy: AAAAAAAAAAAAAAAAAAAAAAAA

8bitprodigy: Done.

src/rlgl.h
```diff
@@ -705,6 +712,7 @@ RLAPI bool rlIsStereoRenderEnabled(void);               // Check if stereo rende
 
 RLAPI void rlClearColor(unsigned char r, unsigned char g, unsigned char b, unsigned char a); // Clear color buffer with color
 RLAPI void rlClearScreenBuffers(void);                  // Clear used screen buffers (color and depth)
+RLAPI void rlClearScreenBuffersEx(rlBufferBitFlags flags); // Clear used screen buffers (selectable with '|'-separated flags)
```
CrackedPixel, line 715: minor: instead of specifying `|` separated flags it may be better to just call it a bitmask or something

raysan5:
@8bitprodigy rlgl API does not implement any *Ex() function, no plans for it.

src/rlgl.h
8bitprodigy: I mean, I'm indifferent on the subject and I think you'd know better than me what to put there, so whatever you think would be the best version of the comment, I'll put there.

raysan5:
I decided not to add it.

PR #5252, opened 2025-10-09
C++ compiler support v2

sakgoyal:
Redoing #5232 because I messed up the force push.

Allow C++ compilers to compile this code without separately linking as a C library.

Changes:
* Backport RGFW unicode fix
  * https://github.com/ColleagueRiley/RGFW/commit/0ebeb37b7fc311a5ed5ed9a28404bfddffde43f3
* Add C-style casting where necessary
* Fix a initializer list being in the wrong order (C++ requires declaration order)
* GLFW currently has issues with cross compiling so I am using RGFW for now 
  * https://github.com/glfw/glfw/pull/2777
* Change C Aggregate initializer to C++ default initializer. 

## TODO:
* Do not modify external libraries. update them instead
  * `jar_*` is ok to update. Managed by raysan5
* C++ ` = {0}`
  * maybe consider `-ftrivial-auto-var-init=zero` or something instead

PR #5274, opened 2025-10-15
[build] Code conventions auto-formatting with `.clang-format` -WIP-

sakgoyal:
Fixes https://github.com/raysan5/raylib/pull/5252#issuecomment-3403917757

## TODO: 
* Finalize the format. The current version is not perfect and is just here to provide an example on how to set it up. using this causes many files to change dramatically so it needs to be adjusted to better reflect the actual formatting used.
* can it be placed in a different location? 
  * I don't want *another* file in the root directory that clutters it up more.


raysan5:
@sakgoyal It would be great to have this option but I'm afraid this is not an easy task, still, let's see if it can be accomplished...

raysan5:
I'm closing this PR, nobody is interested in working on it.

PR #5275, opened 2025-10-15
Fix DrawBillboardPro so that flipped images work correctly.

aixiansheng:
DrawBillboardPro was working correctly with flipped x/y axes for textures whose entire image was sampled by the source rectangle, but for textures drawn with a source rectangle that was smaller than the texture, the sampling was incorrect.  The attached source file and texture file can be used to demonstrate the issue before and after the fix.

[donutz.c](https://github.com/user-attachments/files/22935493/donutz.c)

test


aixiansheng:
Sorry, Added more tests, need to update this code

PR #5276, opened 2025-10-16
Issue 5259: Fix DrawBillboardPro so that flipped textures draw correctly when sampling parts of a texture

aixiansheng:
Previously, DrawBillboardPro would not sample a texture from the correct UV coordinates if the size had negative values and the whole texture was not being used (the issue would not be seen if the whole texture was used and flipped).

This PR corrects that issue.

The test code has comments that explain what parts were broken before and will be fixed afterward.  It also shows that flipping of whole textures worked before and after the fix.

htest
vtest

[donutz.txt](https://github.com/user-attachments/files/22939132/donutz.txt)

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -3904,14 +3904,14 @@ void DrawBillboardPro(Camera camera, Texture2D texture, Rectangle source, Vector
     // Flip the content of the billboard while maintaining the counterclockwise edge rendering order
     if (size.x < 0.0f)
     {
-        source.x += size.x;
+        source.x -= size.x;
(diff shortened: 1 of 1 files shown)


CrackedPixel:
~~did you check it with existing examples and make sure nothing gets broken by this? if people are relying on its current implementation and it switches `+` to `-` that could be bad for peoples projects.~~

~~would be good to see a before vs after comparing different situations to verify easier~~

Update: i see the `.txt` file now

raysan5:
@aixiansheng Looks good! Thanks for the nice test example!

PR #5277, opened 2025-10-16
[examples] Added donut pie charts to `shapes_pie_chart`

GideonSerf:
Added an option to turn the pie chart into a donut.

shapes_pie_chart

---

Fixed formatting issue with `shapes_mouse_trail`'s header.


raysan5:
@GideonSerf Thank you very much! This example looks great! Very complete and illustrative! All new examples should be in this same line!

PR #5279, opened 2025-10-16
[rcore][drm] Disable touch position simulation from mouse movement for touchscreen devices #4842

MULTidll:
## Summary

* Touch positions are now reset to an invalid state (`{-1, -1}`) instead of zero in fuction `PollInputEvents` from  `src/platforms/rcore_drm.c` file, which helps avoid unintended touch detections on DRM touchscreen devices.
* The mapping of mouse position to touch position is disabled for DRM touchscreen devices.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -1130,11 +1130,12 @@ void PollInputEvents(void)
     // Register previous touch states
     for (int i = 0; i < MAX_TOUCH_POINTS; i++) CORE.Input.Touch.previousTouchState[i] = CORE.Input.Touch.currentTouchState[i];
 
-    // Reset touch positions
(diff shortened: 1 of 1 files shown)


raysan5:
@MULTidll Merging the change because I understand you are working on DRM with a touch display at the moment but I'm not sure about the -1 convention, default state is 0, and if finally adopted, it requires to be reviewed in all platforms, so this change conditions multiple other changes for platform consistency.

MUL7ii:
@raysan5 sure. yes indeed i understannd the usage of -1 convention would affect for other platforms I will try something if the issue can be rectified using the default 0 state ,  I would like to know if you have any suggestions to handle the issue in other way so that the default state is kept 0?

MUL7ii:
@raysan5 So I guess if (0, 0) is the default, making (0, 0) not to be counted as a valid touch unless a real touch event sets it to (0,0), but this assumes that a real touch at (0,0) is very rare to happen so we would be needing to handle that special case.

PR #5280, opened 2025-10-16
Fix triangle strip array size and simplify loop

Jopestpe:
# Description

😅 Oops… the array couldn’t hold all points with `segments = 60`

* Increased points array size to close the triangle strip loop  
* Loop simplified for clarity

image


raysan5:
@Jopestpe thanks for the review!

PR #5285, opened 2025-10-19
[rcore] ADD IsCursorDisabled

CrackedPixel:
https://github.com/raysan5/raylib/pull/4955 this PR was to fix the issue on https://github.com/raysan5/raylib/issues/4940

There was never a `IsCursorDisabled` added to go with the change

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -3591,6 +3591,11 @@
       "returnType": "bool"
     },
     {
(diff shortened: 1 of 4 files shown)


raysan5:
@CrackedPixel I'm afraid I'm not adding this new function. Trying to think about a real use case that can not be addressed otherwise but I can't find one. User is in charge of enabling/disabling (locking) mouse so state can be tracked at user side and there is no need of an extra function for that.

PR #5286, opened 2025-10-20
[examples] Fixed `shaders_mandelbrot_set` for WebGL

JordSant:
Fix some limitations on the WebGL glsl shader implementation: no while loops and for loop with a const iteration limit.
Also set a lower number of iterations for the web platform.


raysan5:
@JordSant thanks for the review! yeah, glsl100 is quite limited...

JordSant:
Yes, sorry for that.
I tested it only on desktop, so when it was not working on the raylib examples web page I investigated a bit more and found these issues.
I hope it runs now.

JordSant:
Hi!
I see it does not work on the web page [raylib.com/examples.html](https://www.raylib.com/examples.html).
Can you recompile it with the new glsl100 shader?
Thanks

PR #5289, opened 2025-10-21
implemented monitor profile functions

AzuxirenLeadGuy:
This commit provides function to obtain all fullscreen resolutions supported in desktop. This makes use of GLFW functions for the same. 

Here is a example for its use

```c
        // Get supported modes for monitor_id = 0
	MonitorProfile* profile = GetSupportedModes(0);

	if (profile != NULL) {
                // The order of modes is ascending, so we take the highest mode as such
		const MonitorResolution mode = profile->resolutions[profile->resolution_count - 1];
		SetWindowSize(
			(int)mode.width,
			(int)mode.height
		);
		ToggleFullscreen();
	}


        // Later, when not in use
	ClearMonitorProfile(profile);
```

PS: My first commit here. I want to express my gratitude for the incredible work that all the maintainers and contributors put into raylib. Tinkering with games using this library has been an absolute joy for me. I sincerely wish everyone continued success in maintaining and enhancing this amazing project!

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -956,6 +956,17 @@ typedef bool (*SaveFileDataCallback)(const char *fileName, void *data, int dataS
 typedef char *(*LoadFileTextCallback)(const char *fileName);            // FileIO: Load text data
 typedef bool (*SaveFileTextCallback)(const char *fileName, const char *text); // FileIO: Save text data
 
+typedef struct{
+    unsigned int width;
+    unsigned int height;
+} MonitorResolution;
+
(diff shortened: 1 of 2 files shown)


raysan5:
@AzuxirenLeadGuy Sorry but I'm afraid I prefer to keep the API as minimal as posdible and only add new structures and functions when strictly needed.

casperbear:
@AzuxirenLeadGuy thanks! I was wondering how to get the list of resolutions

Issue #5256, opened 2025-10-10
[rcore][windows] Failed to find selected monitor on SDL platform


NoNameAuthenticated:
### I have checked that:
 - [ X] I tested it on latest raylib version from master branch
 - [ X] I checked there is no similar issue already reported
 - [ X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [X ] My code has no errors or misuse of raylib

### Issue description

I have compiled raylib on the SDL platform (for touch screen compatibility) with SDL3 and also on the GLFW platform for comparison purposes and I observe different behavior on my system with these two platforms with respect to the GetMonitorWidth(GetCurrentMonitor()) and GetMonitorHeight(GetCurrentMonitor()) raylib functions.  

### Environment 
Platform: SDL (SDL3) vs GLFW , 
Operating System: Windows 11
OpenGL version: 3.3.0
GPU: Intel(R) UHD Graphics 620

### Code Example

```c++
#include <raylib.h>
#include <iostream>


int main()
{
    InitWindow(0, 0, "");
    SetWindowPosition(0, 0);

    int currentMonitorNumber{GetCurrentMonitor()};
    std::cout << "Current monitor number: " << currentMonitorNumber << '\n';
    std::cout << "Monitor count: " << GetMonitorCount() <<  '\n';
    std::cout << "Current monitor width: " << GetMonitorWidth(currentMonitorNumber) << '\n';
    std::cout << "Current monitor height: " << GetMonitorHeight(currentMonitorNumber) << '\n';

    CloseWindow();
    return 0;
}
``` 

I get the following output on the following platforms, both of which were compiled with the latest raylib master (raylib 5.6-dev):

**SDL (SDL3):**
Current monitor number: 1
Monitor count: 1
Current monitor width: WARNING: SDL: Failed to find selected monitor 0
Current monitor height: WARNING: SDL: Failed to find selected monitor 0

Here is what is displayed in the PowerShell:
DISPLAY: Device initialized successfully
    > Display size: 1920 x 1080
    > Screen size: 0 x 0
    > Render size: 0 x 0
    > Viewport offsets: 0, 0
PLATFORM: DESKTOP (SDL3): Initialized successfully

**GLFW:**
Current monitor number: 0
Monitor count: 1
Current monitor width: 1920
Current monitor height: 1080

Here is what is displayed in the PowerShell:
DISPLAY: Device initialized successfully
    > Display size: 1920 x 1080
[...]

PR #5290, opened 2025-10-21
[rcore_desktop_sdl] fix: handle monitor ID correctly on SDL3

sleeptightAnsiC:
SDL3 uses ID when dealing with monitors, unlike SDL2 which uses Index for the same thing. This problem was already fixed in multiple places by use of preprocessor branches, so I did the very same thing elsewhere.

Please, notice that this is a pretty bad solution to this problem, and I only did it to keep it consistent with the rest of the code. The more about why it's not correct is mentioned here https://github.com/raysan5/raylib/issues/5256#issuecomment-3428782642 and here https://github.com/raysan5/raylib/issues/5256#issuecomment-3429156919 . Hopefully, someone will refactor it someday :)

Fixes: https://github.com/raysan5/raylib/issues/5256


raysan5:

PR #5292, opened 2025-10-22
[examples] Fix warnings in more examples.

JeffM2501:
Fixes typecast warnings in MSVC for more examples.

### DIFF
--- a/examples/shapes/shapes_clock_of_clocks.c
+++ b/examples/shapes/shapes_clock_of_clocks.c
@@ -14,6 +14,9 @@
 *   Copyright (c) 2025 JP Mortiboys (@themushroompirates)
 *
 ********************************************************************************************/
+#if defined(WIN32)
+#define _CRT_SECURE_NO_WARNINGS
+#endif
 
 #include "raylib.h"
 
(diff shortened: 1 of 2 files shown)


examples/shapes/shapes_clock_of_clocks.c
```diff
@@ -14,6 +14,9 @@
 *   Copyright (c) 2025 JP Mortiboys (@themushroompirates)
 *
 ********************************************************************************************/
+#if defined(WIN32)
+#define _CRT_SECURE_NO_WARNINGS
```
raysan5, line 18: Why is this required? It is not required for any example... 

I prefer to avoid/minimize platform-specific code in examples...

JeffM2501: Because this example calls a time function that windows thinks is insecure. Other examples do not.

raysan5: Mmmmh... ok

raysan5:
@JeffM2501 thanks for the reviews, note that most of the new examples still require formating and structure review, I'll do that as I got time for it...

PR #5295, opened 2025-10-23
Fix typo in RPI cross-compiler variable name

diogohartuiqdebarba:
It's called USE_RPI_CROSSCOMPILER in other places and only here as USE_RPI_CROSS_COMPILER. Which was causing me to compile incorrectly with gcc instead of RPI_TOOLCHAIN.


raysan5:
@diogohartuiqdebarba Oh! Good catch! I imagine that was a flag not used a lot... 😄

PR #5299, opened 2025-10-25
[Win32 Platform] Fix some issues building in mingw

JeffM2501:
The Win32 platform undefines too many versions of LoadImage, but then also tries to use them, this will fail to compile with some compilers (mingw).
This PR removes the unneeded redefines so that the code will work in mingw.


raysan5:

PR #5300, opened 2025-10-25
[rlsw] Subpixel correction

Bigfoot71:
I finally managed to fix the triangle cracking and jittering issues!

The problem was actually pretty silly, the difference between the triangle edge and the first pixel of each line wasn't accounted for in the interpolation. Very, very silly...

While I was at it, I also made a bunch of other logical adjustments and renamed several variables to be more "standard" with what you usually see in papers.

I also fixed the same subpixel correction issue for quads and lines.

Additionally, I replaced the Bresenham line algorithm with a DDA. The new version has the advantage of being much more temporally stable during motion. It uses more floating-point calculations than Bresenham (which previously used some fixed point), but it's also much more vectorizable.

And for the most important and visible part, the triangles, here's a before/after comparison.

_Note: The small flickering that remains in the second version is due to z-fighting with the lines under the mesh_

## Before

https://github.com/user-attachments/assets/3656fffb-f4ee-4621-8b0e-d502c4b9f87d

## Now

https://github.com/user-attachments/assets/5b454273-646a-45b5-bd8a-406603f84da4

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -3097,34 +3097,34 @@ static inline void FUNC_NAME(const sw_vertex_t *v0, const sw_vertex_t *v1) \
     {                                                                   \
         steps = fabsf(dx);                                              \
         if (steps < 1.0f) return;                                       \
-        substep = (dx >= 0.0f) ? (1.0f - sw_fract(x0)) : sw_fract(x0);  \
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Oh! Nice improvement! I hardly noticed previous issue! Thanks!

PR #5304, opened 2025-10-25
[rlsw] Completeness of `glDraw` functions

Bigfoot71:
The `glDrawElements` implementation was missing (no idea how that got overlooked), and I also fixed the logic. I remember quickly getting `swDrawArrays` working with the immediate functions and that I needed to rewrite it up properly, but I totally forgot...

Done!

https://github.com/user-attachments/assets/f3aa8d62-9846-4160-bd7b-8a45446145d2

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -675,6 +675,7 @@ SWAPI void swBindTexture(uint32_t id);
 // Defines and Macros
 //----------------------------------------------------------------------------------
 #define SW_PI       3.14159265358979323846f
+#define SW_INV_255  0.00392156862745098f
 #define SW_DEG2RAD  (SW_PI/180.0f)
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 Oh! Nice addition! 👍😄

I see one operation has been moved to defines, wondering if precalculating some values to minimize divisions and prioritize multiplications could suppose some performance improvement... or maybe compiler already takes care of those?

Bigfoot71:
> I see one operation has been moved to defines, wondering if precalculating some values to minimize divisions and prioritize multiplications could suppose some performance improvement... or maybe compiler already takes care of those?

Edit: Sorry, I'm tired, I just placed a `const float inv255 = 1.0f / 255` in swDrawArrays at some point, and I just thought it was silly that it wasn't being reused, that's all

PR #5307, opened 2025-10-26
[examples] Added `shaders_color_correction`

JordSant:
Added a new example to shaders: basic color correction.
It has controls to modify brightness, contrast and saturation.
To select one image from four it uses a GuiToggleGroup
Tested on Windows with Visual Studio and on web on Chrome.


raysan5:
@JordSant Very nice addition! Thanks!

PR #5310, opened 2025-10-26
[examples] Fixed spaces `shaders_mandelbrot_set`

JordSant:
Fixed spaces in c file: some tabs replaced by spaces.


raysan5:
@JordSant Good catch! Thanks for the review!

PR #5314, opened 2025-10-28
[rlsw] Fix axis aligned quad detection

Bigfoot71:
### **TL;DR** The PR was mainly aimed at fixing the axis-aligned quad detection function, but also contains the fix for DRM and a review for the alignments.

___

The issue for the axis-aligned quad detection was that it should have relied on transformed screen coordinates. This could even cause crashes in some cases, I spotted it while trying to transform text with `rlRotatef`

I also made it a bit more readable and documented.

___

I also fixed the DRM and `swGetColorBuffer` issue. The function wasn't actually needed even to get the internal buffer dimensions, `swBlitFramebuffer` should handle everything automatically, and it now does.

___

Regarding `alignas` the idea was to align both size and alignment (eg, 4-byte values aligned to 4). This allows, in theory for ARM, to use optimal load/store instructions instead of risking unaligned accesses that take extra cycles or require software emulation in worst case (old CPUs). On x86-64, it doesn't change much, but on ARM and RISC-V it should ensure predictable and optimal performance. Benchmarks are obviously needed to measure the real gain.

That said, the `alignas` on `color` was just one of my tests. I got confused with my branches and noticed it only after making my 'typo' fix. I figured, well let's leave it and I'll revisit it in the next PR. Sorry about that, but it reminded me that it's C11!

___

So, last PR ended up being a bit of a "mixed bag". From now on I'll try to make PRs more focused on specific issues.

Next on my list are handling cull face properly (culling is still happening even when disabled) and adding depth mask, since depth is currently always written.

I'm also working on a 3D example/mini game that runs quite well. It will help isolate real world issues, so expect more fixes coming from that.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -622,6 +622,18 @@ SWAPI void swBindTexture(uint32_t id);
     #define SW_ALIGN(x) // Do nothing if not available
 #endif
 
+#if defined(_M_X64) || defined(__x86_64__)
+    #define SW_ARCH_X86_64
+#elif defined(_M_IX86) || defined(__i386__)
+    #define SW_ARCH_X86
+#elif defined(_M_ARM) || defined(__arm__)
+    #define SW_ARCH_ARM32
+#elif defined(_M_ARM64) || defined(__aarch64__)
+    #define SW_ARCH_ARM64
+#elif defined(__riscv)
+    #define SW_ARCH_RISCV
+#endif
+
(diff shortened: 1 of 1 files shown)


Bigfoot71:
Oh, and I forgot, for the alignment question, I added definitions by architecture, not used for the moment, but could be useful if we want to make specific ARM/RISC-V optimizations later

raysan5:
@Bigfoot71 thanks for the super-fast review! 👍😄

PR #5315, opened 2025-10-29
[rcore] Add function SetMediaPath() for modifying the output directory for screenshots/gifs

Yonokid:
I have a particular problem of not being able to access my screenshots/gifs due to the working directory that my bindings use for raylib. I added a mediaPath to the Storage struct in rcore that defaults to the same as basePath (I can't be absolutely sure basePath isn't used for other things or may be used for other things in the future so I added a new attribute). Calling the function with a valid path will save all screenshots/gifs there.


raysan5:
@Yonokid I'm afraid I'm not adding a new function just for that, use TakeScreenshot() at user code instead.

Yonokid:
That's ok, but there still isn't a way to modify the path of a gif recording...

raysan5:
@Yonokid Gif recording is being removed for next release.

Issue #5309, opened 2025-10-26
[rlsw] Undefined behaviour reported by Sanitizer


raysan5:
One user on Discord reported the folllowing Undefined Behaviour:

```
../raylib/src\external/rlsw.h:1357:29: runtime error: -65535 is outside the range of representable values of type 'unsigned short'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../raylib/src\external/rlsw.h:1357:29
```

Full description (from Dicord user):

 > Converting a float outside of the range of an int to an int is difficult in C without triggering UB. (if you care it is in the depth buffer code) of course you cast to long long first or maybe it is a problem with the range of values in the depth buffer (idk what it should be).

 > I can do: 

```c
    #define DEPTH_TYPE uint16_t
    #define SIGNED_DEPTH_TYPE int32_t
    #define DEPTH_IS_PACKED 1
    #define DEPTH_MAX UINT16_MAX
    #define DEPTH_SCALE (1.0f/UINT16_MAX)
    #define PACK_DEPTH(d) ((DEPTH_TYPE)((intmax_t)((d)*DEPTH_MAX)))
    #define UNPACK_DEPTH(p) (p)
```
 > to make it shut up, as intmax_t is large enough, but idk if that would be masking a bug if the depth calc is wrong.

 > Anyway, even with that the software rasterizer works great.

PR #5317, opened 2025-10-29
[rlsw] Review depth formats and fix depth writing

Bigfoot71:
I reviewed the depth formats and removed the 8-bit one. It could have been marginally useful for orthographic projection, but without heavier adjustments it would never really be practical.

I also added a 32-bit floating-point depth format, which avoids conversions, though after testing it turned out to be slightly slower than 16-bit with conversion. So 16-bit remains the default.

I also added saturation when writing depth values, since in some cases, when the camera was very close to a clipped triangle, the depth value could exceed the limit and cause an overflow.

This might require reviewing some details in the clipping, I find it strange at first glance...

This issue seems to mention the related UB problem: https://github.com/raysan5/raylib/issues/5309

I compiled in debug mode with UBSan and ran several examples, didn't notice any issues on my side with this update.

The 24-bit format has also been quickly reviewed and tested, everything seems perfect!

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -1346,7 +1346,7 @@ static inline void sw_framebuffer_write_color(sw_pixel_t *dst, const float src[4
 
 static inline void sw_framebuffer_write_depth(sw_pixel_t *dst, float depth)
 {
-    depth = sw_saturate(depth);
+    depth = sw_saturate(depth); // REVIEW: An overflow can occur in certain circumstances with clipping, and needs to be reviewed...
 
 #if SW_DEPTH_IS_PACKED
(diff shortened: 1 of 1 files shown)


Bigfoot71:
After some more testing, the results are mixed about 32-bit format.

I found it strange that the 32-bit format was slower, but it seems to vary depending on the example, sometimes it's very slightly slower or basically identical, but in the `first_person_maze` example it actually gains just under 10 FPS.

Well, it would probably need more specific profiling.

raysan5:
@Bigfoot71 thanks for the review! Definitely the 8-bit depth buffer use is very limited.

PR #5318, opened 2025-10-30
[rcore] Add non-va_list custom logging callback

eutro:
This PR adds a version of `TraceLogCallback`, called `TraceLogCallback2` (pending a better name), which is like the original but instead of receiving a format string and a `va_list` of arguments, receives a pre-formatted string of the arguments only.

```c
typedef void (*TraceLogCallback)(int logLevel, const char *text, va_list args);  // Logging: Redirect trace log messages
typedef void (*TraceLogCallback2)(int logLevel, const char *text);  // Logging: Redirect trace log messages 2
```

## Why?

The default logger logs to standard output, but can be redirected using `SetTraceLogCallback()`. This works for users of Raylib interfacing using C, but cannot be used with some other programming languages' bindings (e.g. Racket) because the foreign function interface (FFI) used to call Raylib does not support `va_list`. This is unfortunate, because redirecting Raylib logging to the language's own logging facilities seems like an excellent use-case for `TraceLogCallback`, but one for which it cannot be used.

This PR solves the problem by formatting the log message ahead of time, before it is passed to the callback, where it is then much easier for non-C languages to use. I considered other designs, such as using a `union` to pass out all the arguments in an FFI-compatible way, but this seems the simplest.

## What is this PR?

I am mainly just throwing this PR out here as a possible implementation to fix the issue, but I'm not sure if I will have the time to see it through.

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -3085,6 +3085,21 @@
       ]
     },
     {
+      "name": "TraceLogCallback2",
+      "description": "Logging: Redirect trace log messages 2",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int",
+          "name": "logLevel"
+        },
+        {
+          "type": "const char *",
+          "name": "text"
+        }
+      ]
+    },
+    {
       "name": "LoadFileDataCallback",
       "description": "FileIO: Load binary data",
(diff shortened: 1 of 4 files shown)


raysan5:
@eutro I'm afraid I'm not adding that new function. I prefer to avoid adding function targeting specific use-cases for language bindings.

PR #5319, opened 2025-10-30
Adding SwiftForth language binding

cthulhuology:
I've created a mostly functional Forth wrapper that works with Forth Inc's SwiftForth.  Currently, I am testing on Windows, but will begin Linux and MacOS testing in the next couple weeks.  Right now it supports all of the functions in raylib.h, and will be adding the additional header files next week.

SwiftForth is a cross platform Forth implementation, that is written in Forth, and has its own assembler, compiler, ide, and runtime environment.  Forth Inc. was originally founded in 1971 by Charles Moore, who invented Forth, and SwiftForth https://www.forth.com/product/swiftforth/ is a for-profit environment which comes with all of it's source code!


raysan5:
@cthulhuology thanks! great addition! 😄

PR #5325, opened 2025-11-03
[rcore] Fix broken SDL build

komunre:
I don't know what happened there, but `rcore_desktop_sdl.c` on master seems to have an unclosed if statement and a line with an undefined variable, so this PR is to just fix these.


raysan5:
@komunre Oh! Thanks for the review! It was me and a following PR!

PR #5327, opened 2025-11-03
[rcore] Determine whether or not an SDL3 monitor ID is valid using SDL_GetDisplayUsableBounds()

NoNameAuthenticated:
On the SDL3 backend, when a laptop is connected to an external monitor and then the laptop lid is closed, the original SDL3 monitor ID for the laptop's monitor is invalidated and replaced with a monitor ID that exceeds the total amount of monitors upon reopening the laptop lid. Many of Raylib's functions in the file "rcore_desktop_sdl.c" do not properly handle such a situation, as they return an error message and do not actually run when the following condition is met:

#if defined(USING_VERSION_SDL3)
    if((monitor > 0) && (monitor <= monitorCount))

I think the monitor number ("monitor") should be validated not against the monitor count, but rather the code should attempt to call the SDL3 function "SDL_GetDisplayUsableBounds(monitor, &displayBoundsRectangle)", passing in the monitor number ("monitor") and the "SDL_Rect" object "displayBoundsRectangle" by address. This function returns "true" on success, and tells you that the requested monitor is valid.

It worked with the following example (I was able to toggle the borderless windowed mode, even though the monitor number was now 4 upon reopening the laptop lid):

```c++
#include <raylib.h>
#include <iostream>


int main()
{
    InitWindow(800, 600, "");
    SetWindowPosition(100, 100);
    SetTargetFPS(60);

    int currentMonitor{GetCurrentMonitor()};
    
    std::cout << "GetCurrentMonitor(): " << GetCurrentMonitor() << '\n';
    std::cout << "GetMonitorCount(): " << GetMonitorCount() << '\n';

    while (!WindowShouldClose())
    {
        int newMonitor{GetCurrentMonitor()};
        if (newMonitor != currentMonitor)
        {
            currentMonitor = newMonitor;
            std::cout << "GetCurrentMonitor(): " << GetCurrentMonitor() << '\n';
            std::cout << "GetMonitorCount(): " << GetMonitorCount() << '\n';
        }

        if (IsKeyPressed(KEY_F11)
        {
            ToggleBorderlessWindowed();
        }

        BeginDrawing();
        ClearBackground(BLACK);
        EndDrawing();
    }

    CloseWindow();
    return 0;
}


```

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -463,7 +463,8 @@ void ToggleFullscreen(void)
     const int monitorCount = SDL_GetNumVideoDisplays();
 
 #if defined(USING_VERSION_SDL3) // SDL3 Migration: Monitor is an id instead of index now, returns 0 on failure
-    if ((monitor > 0) && (monitor <= monitorCount))
+    SDL_Rect displayBoundsRect; // Different SDL_Rect variable from displayBounds to avoid overwriting it
+    if (SDL_GetDisplayUsableBounds(monitor,  &displayBoundsRect)) // Returns true on success, which confirms that the monitor is valid
 #else
     if ((monitor >= 0) && (monitor < monitorCount))
 #endif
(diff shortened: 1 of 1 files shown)


NoNameAuthenticated:
Instead of "SDL_GetDisplayUsableBounds(monitor, &displayBoundsRectangle)", which requires to initialize an "SDL_Rect" object "displayBoundsRectangle" and pass it by address, another SDL3 function could be used, which does not require to pass in other arguments than the monitor ID: 

SDL_PropertiesID SDL_GetDisplayProperties(SDL_DisplayID displayID);

It returns a valid property ID on success or 0 on failure, so you could use the following conditional statements instead:

```c++
if (SDL_GetDisplayProperties(monitor) != 0) // returns 0 on failure, so a value other than zero indicates that the monitor number is valid
``` 

see this link for more details: [https://wiki.libsdl.org/SDL3/SDL_GetDisplayProperties](url)

PR #5169, opened 2025-09-03
use FLAG_* macros where possible

JohnnyCena123:

PR #5333, opened 2025-11-04
[rcore][glfw] fix `IsWindowFocused()` inverted logic introduced by #5169

JohnnyCena123:
see https://github.com/raysan5/raylib/pull/5169#issuecomment-3482234403


raysan5:
@JohnnyCena123 thanks for the review!

PR #5340, opened 2025-11-07
Added Vector2DistanceAngle function

RobinsAviary:
This one is pretty tiny, but it's a function that I always missed from GameMaker, now bundled into raylib's wonderful Vector2 type! I find myself reaching for this calculation pretty often when rendering graphics, so I figured it might be useful to have in the base library.

I set it up to be as easy to use as possible, so it accepts radians for angle. I also applied an offset of `-(.5f * PI)` to make the function output a line pointing straight up if the user passes in `0.0f` for angle. If this is too much, please feel free to let me know, and I can remove it.

Thank you again for the awesome library!

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -320,6 +320,15 @@ RMAPI float Vector2Distance(Vector2 v1, Vector2 v2)
     return result;
 }
 
+// Calculate a vector with a distance and angle
+// Angle is in radians (full rotation is 2.0f * PI), starts up and goes clockwise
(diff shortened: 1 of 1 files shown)


RobinsAviary:
Oh, also, I can probably make an example for this for the collection if you'd think it'd be useful.

RobinsAviary:
I see now there is actually a recent example that demonstrates this concept, this function would allow that example to be simplified, if that's something you're interested in.

raysan5:
@RobinsAviary I prefer to minimize function addition if same result could be accomplished with existing ones, but thanks anyway.

Issue #5326, opened 2025-11-03
[rlsw] Quad drawer crashes when a rectangle clipped at windows border.


AdjWang:
Hi, 

### Issue description

I'm using rlsw on win32 with raygui, the program crashes in `rlsw.h` when I moving a sub-window outside the main window.

At a glance of the code, it turns out to be `sw_quad_clip_and_project()` producing a zero area quad, then `sw_quad_sort_cw()` outputs 4 vertices with 2 of them are NULL.

### Environment

I'm testing on windows 10, compiles raylib with `OPENGL_VERSION=Software`.

> I didn't find a way to get rid of glfw, so I hard coded `PLATFORM_DESKTOP_WIN32` in `rcore.c` to force using `rcore_desktop_win32.c`. I don't think it would affect this issue.

### Issue Screenshot

<img width="1089" height="441" alt="Image" src="https://github.com/user-attachments/assets/e63a8f2c-b476-4be7-adf5-ea5871b872e1" />

### Code Example

```
#include "raylib.h"

int main() {
  int screenWidth = 1280;
  int screenHeight = 800;
  InitWindow(screenWidth, screenHeight, "Raylib");
  SetTargetFPS(60);

  while (!WindowShouldClose()) {
    BeginDrawing();
    ClearBackground(DARKGRAY);
    DrawRectangle(0, 800, 10, 10, RAYWHITE);
    EndDrawing();
  }

  CloseWindow();
  return 0;
}
```

The rectangle is outside the main window with only the top border overlapping with the main window bottom border.
The program crashes inside `sw_quad_raster_axis_aligned_TEX_BLEND()`.

raysan5:
@Bigfoot71 Please, could you take a look to this issue?

Bigfoot71:
@raysan5 Sorry for the delay, I couldn't get to it earlier

Thank you @AdjWang for the debugging and the example, I've just proposed a fix, your example now works as expected

___

As a quick explanation, the problem simply stemmed from the clipping, the vertices shouldn't have been clipped at 800, but at 799, and therefore the quad should simply have been considered off-screen.

raysan5:
@Bigfoot71 Thanks for the review! No worries, really, take your time!

@AdjWang Please, let me know if issue is fixed, closing the issue in the meantime.

AdjWang:
Nice and quick fix, thanks!

PR #5342, opened 2025-11-07
[rlsw] Fix clipping boundary overflow

Bigfoot71:
Fixes the buffer overrun issue reported here: https://github.com/raysan5/raylib/issues/5326#issuecomment-3487022305

After investigating the problem, the most reasonable approach seems to be using exclusive tests for the X/Y planes during clipping while keeping inclusive tests for Z.

The current clipping implementation is based on multiple rewrites of the approach originally described in this blog:
https://fabiensanglard.net/polygon_codec/

Our original implementation followed the same conditional logic as in the blog post, but this behavior might also be related to the projection [NDC -> screen] convention we follow _(and which didn't seem to be a problem anyway in the context of the blog post)_, so clipping with exclusive X/Y borders seems more consistent in our case, while keeping Z inclusive avoids losing fragments.

For reference, here are the original papers that describes the logic:
* https://fabiensanglard.net/polygon_codec/clippingdocument/p245-blinn.pdf
* https://fabiensanglard.net/polygon_codec/clippingdocument/Clipping.pdf

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -2118,12 +2118,12 @@ static inline int sw_clip_##name(                                               
 // Frustum cliping functions
 //-------------------------------------------------------------------------------------------
 #define IS_INSIDE_PLANE_W(h) ((h)[3] >= SW_CLIP_EPSILON)
-#define IS_INSIDE_PLANE_X_POS(h) ((h)[0] <= (h)[3])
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 thanks for the review and the detailed explanation and references!

PR #5343, opened 2025-11-07
[examples] `core_directory_files` fixes

hugoarnal:
Fixed the following issues:
- When clicking on a folder, the index can be superior to the count of the new entered folder which instantly crashes the example
- Char devices (such as the ones in `/dev`) can be clicked even though they are not directories.


hugoarnal:
For some reason CI absolutely wants to push `rlparser` even though I'm based on master which should ignore it :thinking:

raysan5:
@hugoarnal Please, could you remove `rlparser` binary?

EDIT: Just saw it's automatically added by the CI Workflow but  it's strange because that workflow should only run on `raylib.h` updates...

raysan5:
@hugoarnal Just reviewed it myself and merged

PR #5344, opened 2025-11-08
[examples] Added wabbit original author

giuliano-macedo:
After an email exchange with @iainlobb he wrote me who the original author is, and that he has the copyright.


raysan5:
@giuliano-macedo @iainlobb Thanks for the update, I couldn't find it on its day! What is the usage license for this image? Is it creative commons? raylib is an open-source project with a very permissive license and I try to only use assets with permissive licenses on its examples. 

Also note that the image is not exactly the same as original, I modified it to fit into a 32x32 pixels square POT image, required by some limited GPUs.

raysan5:
As the sprite license was not clear, just replaced the wabbit sprite by a new sprite created specifically for the example.

PR #5352, opened 2025-11-13
Refactor shapes generation from Par Shapes mesh

und3f:
Add internal method BuildMeshFromParShapesMesh and apply it where Par Shapes mesh converted to Raylib mesh


raysan5:
@und3f Thanks but I prefer to minimize in raylib function-calling-function

PR #5357, opened 2025-11-15
[examples] Added: `textures_screen_buffer`

nezvers:
Example with retro flame simulation to showcase how to do software render on an image that is updated every frame.


raysan5:
@nezvers thanks for the addition! Note that I'm probably changing the name...

PR #5358, opened 2025-11-18
Fix IsGamepadAvailable() returns false for an available controller immediately after window initialization

und3f:
Fixing issue of `IsGamepadAvailable()` returning `false` for available controllers right after window initialization despite `GetGamepadName()` contains the name.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1749,7 +1749,12 @@ int InitPlatform(void)
     {
         // WARNING: If glfwGetJoystickName() is longer than MAX_GAMEPAD_NAME_LENGTH,
         // we can get a not-NULL terminated string, so, we only copy up to (MAX_GAMEPAD_NAME_LENGTH - 1)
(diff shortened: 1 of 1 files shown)


src/platforms/rcore_desktop_glfw.c
```diff
@@ -1749,7 +1749,12 @@ int InitPlatform(void)
     {
         // WARNING: If glfwGetJoystickName() is longer than MAX_GAMEPAD_NAME_LENGTH,
         // we can get a not-NULL terminated string, so, we only copy up to (MAX_GAMEPAD_NAME_LENGTH - 1)
-        if (glfwJoystickPresent(i)) strncpy(CORE.Input.Gamepad.name[i], glfwGetJoystickName(i), MAX_GAMEPAD_NAME_LENGTH - 1);
+        if (glfwJoystickPresent(i)) {
```
raysan5: Please, align the brackets to follow raylib cod conventions.

raysan5:
@und3f thanks for the review!

PR #5359, opened 2025-11-18
rcore_desktop_sdl.c, fix crash when strncpy tries copy using NULL pointer

MikiZX1:
When SDL_GameControllerNameForIndex returns null, the app crashes. This was addressed earlier in PR#4859 though the fix submitted on PR #4859 was only fixing the crashing and not addressing the root cause.


raysan5:
@MikiZX1 thanks for the review

PR #5370, opened 2025-11-27
feat: add elle bindings

acquitelol:
At the moment, the bindings are included in the standard library. Eventually, bindings for common libraries will be moved to a seperate `vendor` folder which will not be part of the standard library. When that time comes, I will open a follow-up PR.


raysan5:
@acquitelol thanks for the addition! feel free to send an update when required

PR #5373, opened 2025-11-29
[examples] text_inline_styling: make inline text and background colors respect base alpha

johnmichaeljimenez:
### Summary
This makes `DrawTextStyled()` respect the alpha channel of the base `color` parameter when inline colors are used.
I am currently using this function in my game project, and having automatic alpha multiplier helps with text effects like fade animations.

### Changes
- Inline foreground `[cRRGGBBAA]` and background `[bRRGGBBAA]` alpha values are now **multiplied** by the base color's alpha. This allows global transparency while keeping per-section styling.
- Updated the example with a new demo line clearly showing the effect:
  ```c
  DrawTextStyled(..., "This changes the [c00ff00ff]alpha[r] relative [cffffffff][b000000ff]from source[r] [cff000088]color[r]!!!",
      ..., (Color){0, 0, 0, 100});

<img width="800" height="450" alt="image" src="https://github.com/user-attachments/assets/a49084c4-24dd-4017-b988-7873f54976a9" />

### DIFF
--- a/examples/text/text_inline_styling.c
+++ b/examples/text/text_inline_styling.c
@@ -70,6 +70,8 @@ int main(void)
             // - Define foreground color:      [cRRGGBBAA]
             // - Define background color:      [bRRGGBBAA]
             // - Reset formating:              [r]
+            // Colors defined with [cRRGGBBAA] or [bRRGGBBAA] are multiplied by the base color alpha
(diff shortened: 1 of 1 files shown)


raysan5:
@johnmichaeljimenez Thanks! Nice improvement!

PR #5376, opened 2025-11-30
[examples] Added: `shapes_penrose_tile`

davidbuzatto:
I've created an Penrose Tile example as requested here: https://github.com/raysan5/raylib/discussions/5209
The video below, captured at 30 frames per second, shows the example in action.
[penrose-tile.webm](https://github.com/user-attachments/assets/e29827ea-0104-4b25-88b4-0489cb424875)


raysan5:
@davidbuzatto Note that screenshot is not following expected format, it should be a 800x450 image with no window decoration.

You can capture a screenshot just pressing F12 while the program is running, it is saved in same directory as executable, just need to rename it.

raysan5:
@davidbuzatto thanks for the review and this new example!

PR #5379, opened 2025-12-02
[examples] Added: `text_strings_management`

davidbuzatto:
I've created an example using the majority of the strings management funcions:

- int TextCopy(char *dst, const char *src);
- unsigned int TextLength(const char *text);
- const char *TextFormat(const char *text, ...);
- const char *TextSubtext(const char *text, int position, int length);
- const char **TextSplit(const char *text, char delimiter, int *count);
- const char *TextToUpper(const char *text);
- const char *TextToLower(const char *text);
- const char *TextToPascal(const char *text);
- const char *TextToSnake(const char *text);
- const char *TextToCamel(const char *text);

[strings-management.webm](https://github.com/user-attachments/assets/a72f211d-e8d6-4cbd-8fd4-3c9d62617cc3)

### DIFF
--- a/examples/text/text_strings_management.c
+++ b/examples/text/text_strings_management.c
@@ -19,10 +19,13 @@
 
 #include <stdlib.h>
 
-#define MAX_TEXT_LENGTH 100     // Maximum text particle text length
-#define MAX_TEXT_PARTICLES 100 // Maximum quantity of text particles
-#define FONT_SIZE 30
+#define MAX_TEXT_LENGTH      100
+#define MAX_TEXT_PARTICLES   100
+#define FONT_SIZE             30
 
+//----------------------------------------------------------------------------------
+// Types and Structures Definition
+//----------------------------------------------------------------------------------
 typedef struct TextParticle {
     char text[MAX_TEXT_LENGTH];
     Rectangle rect;    // Boundary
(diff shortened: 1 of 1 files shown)


raysan5:
@davidbuzatto Example is a bit longer than expected but it looks nice. Please, could you review code structure to follow same structure as reviewed in the preevious one? --> https://github.com/raysan5/raylib/commit/d3addad9a7ee1538552f8ca401eb6f84d2019c42

davidbuzatto:
> @davidbuzatto Example is a bit longer than expected but it looks nice. Please, could you review code structure to follow same structure as reviewed in the preevious one? --> [d3addad](https://github.com/raysan5/raylib/commit/d3addad9a7ee1538552f8ca401eb6f84d2019c42)

Done! I fixed some bugs and made some improvements. Yes, unfortunately the example ended up being quite large 😢

raysan5:
@davidbuzatto thanks for the review! merged!

I'm reviewing `shapes_penrose_tile` example and it's very difficult to understand, specially for users with not much coding experience. It would be nice to include some comments and explanations about how things work, that's the pourpose of the examples, to be like small tutorials, otherwise, many users will just look at its beautiful output but will not spend time trying to understand the code. I usually add a comment for every function added, beside the descriptive naming, and also comment the fields of added structurs. I try to avoid shortening variable names like `pos` or `vel`, in raylib I prefer `position` and `speed`/`velocity`.

davidbuzatto:
> @davidbuzatto thanks for the review! merged!
> 
> I'm reviewing `shapes_penrose_tile` example and it's very difficult to understand, specially for users with not much coding experience. It would be nice to include some comments and explanations about how things work, that's the pourpose of the examples, to be like small tutorials, otherwise, many users will just look at its beautiful output but will not spend time trying to understand the code. I usually add a comment for every function added, beside the descriptive naming, and also comment the fields of added structurs. I try to avoid shortening variable names like `pos` or `vel`, in raylib I prefer `position` and `speed`/`velocity`.

If you want, I can do that, you can count on me. The problem was that I was afraid of adding too much text in the form of comments in the code and exceeding the recommended number of lines.

PR #5382, opened 2025-12-02
Replace sprintf() with snprintf() for buffer safety

Auios:
Replace `sprintf()` with `snprintf()` in two locations to prevent potential buffer overflows.
Both calls now explicitly limit the buffer size to 4 bytes.


raysan5:
@Auios Looks good, thanks for the review

PR #5383, opened 2025-12-02
Fix ToggleBorderlessFullscreen() Not Hiding Taskbar

XenoMustache:
Replace the `glfwSetWindowPos` and `glfwSetWindowSize` functions with `glfwSetWindowMonitor` to properly set the window to borderless fullscreen mode.

I did not change any of the other window size or position editing functions to keep this resolution self-contained. This can be amended in a future commit.

Intended to be tested with the following:
```c
#include <raylib.h>

int main(void) {
	InitWindow(800, 600, "Raylib Game");

		if (IsKeyPressed(KEY_F1)) {
			ToggleBorderlessWindowed();
		}

		BeginDrawing();
		ClearBackground(RED);
		EndDrawing();
	}

	CloseWindow();

	return 0;
}
```

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -265,18 +265,15 @@ void ToggleBorderlessWindowed(void)
                 const int monitorHeight = mode->height;
 
                 // Set screen position and size
-            	glfwSetWindowMonitor(
-					platform.handle,
-					monitors[monitor],
-					monitorPosX,
-					monitorPosY,
-					monitorWidth,
-					monitorHeight,
-					mode->refreshRate
-				);
-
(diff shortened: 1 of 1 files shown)


XenoMustache:
I found an additional issue caused by this fix, toggling out of borderless fullscreen causes the window not to be reset properly. I will add an additional commit to this to fix that issue.

XenoMustache:
Additional commit includes replacing the case where toggling out of fullscreen using the GLFW Size and Pos functions doesn't properly reset the window, achieved by using the  `glfwSetWindowPos` function and setting the monitor to `NULL`.

JeffM2501 (approved):
looks like a better way to handle it.

raysan5:
@XenoMustache Thanks for the review! Looks good!

PR #5386, opened 2025-12-03
[cmake] Expose RGFW to cmake

Sethbones:
Added checks for RGFW for cmake

I was trying to automate raylib setup for a simple script, and i ran into compile errors, and out of boredom i was like "yeah why not".

I was contemplating on having an external check like with GLFW, but i needed to sound out the idea first before committing to something stupid.


raysan5:
@Sethbones thanks for the improvement!

PR #5388, opened 2025-12-05
[rModels] Support 16 bit vec3 values in gltf reader

JeffM2501:
There are applications that output size optimized GLTF files. These pack mesh buffer data down into 16 bit formats and use the matrix transform to scale them back to the correct float values.

Raylib's gltf reader only reads float values, meaning that gltf files that are saved in these optimized formats will not read into raylib.

This PR adds support for 16 bit position and both 16 bit and 8 bit normal buffers, letting those files at least read in basic geometry. This change should not effect any end user APIs as the result is still a buffer of floats.

Tested using an optimized gltf file [optimzed.zip](https://github.com/user-attachments/files/23972323/optimzed.zip)

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -5680,6 +5680,54 @@ static Model LoadGLTF(const char *fileName)
                                     normals[3 * k + 2] = nt.z;
                                 }
                             }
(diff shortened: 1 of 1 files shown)


raysan5:
@JeffM2501 thanks for the improvement! there are so many vertex format possibilities for glTF! 😅

PR #5389, opened 2025-12-06
[build][android] fix newer NDK version compile errors

Sethbones:
-target already gets assigned by the clang macro it points to, overwriting it causes it to target linux instead of android, making it check for usr directories instead of the NDK's directories
this makes raylib error during compile in versions of the NDK higher than r23b citing usr/include as the problem when it shouldn't even be there
```
/usr/include/bits/libc-header-start.h:74:5: error: function-like macro '__GLIBC_USE' is not defined
   74 | #if __GLIBC_USE (IEC_60559_BFP_EXT) || __GLIBC_USE (ISOC23)
      |     ^
```
case in point:
image
get overwritten by
image
specifically -target aarch64 overwrites --target=aarch64-linux-android34 from the macro

aarch64 is a valid target by clang but it also points to linux aarch64 and not android aarch64

now the piece of mystery in all of this is, why does r23b work with no errors? and to be honest i've been trying to figure that out for 2 days straight and the only conclusion i could come up to is that it shouldn't work, yet it does.
as if the files it needed just so happened to be similar enough to the native linux libraries that it compiled just fine with no errors.
if anyone knows more please chime in, because by all accounts it doesn't make sense.

### DIFF
--- a/examples/Makefile.Android
+++ b/examples/Makefile.Android
@@ -130,7 +130,7 @@ ifeq ($(ANDROID_ARCH),ARM)
     CFLAGS = -std=c99 -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16
 endif
 ifeq ($(ANDROID_ARCH),ARM64)
-    CFLAGS = -std=c99 -target aarch64 -mfix-cortex-a53-835769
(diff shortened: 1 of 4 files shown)


raysan5:
@Sethbones mmmh... interesting, thanks for the detailed explanation and the fix!

PR #5390, opened 2025-12-06
Fix build.zig typos

spineda2019:
# Small typos
* Fixes small typos in the build.zig build script
* Found when experimenting with zig+cpp+raylib for first time, tried to make sure I didn't miss other typos in the script
* Doesn't change executable code, so shouldn't affect build behavior


raysan5:
@spineda2019 thanks for the review!

PR #5391, opened 2025-12-07
Extending utils

Matorio:
New default color options.
Direct shortcut to center the screen.
Extended scissors tool.
Two utilities for uniform scaling of rectangles.


raysan5:
@Matorio I'm afraid I'm not merging those changes, they can be implemented on user side. I try to minimize raylib API.

PR #5392, opened 2025-12-07
[rcore] Fix modulo bias in `GetRandomValue()`

### LINKED ISSUE #5374: [core] Uniform Random Generators are not Uniform
---

 - [ ] I tested it on latest raylib version from master branch
 - [ ] My code has no errors or misuse of raylib

### Issue description

In `rcore.c`  when `SUPPORT_RPRAND_GENERATOR` is not defined, the supplied code for computing uniform random values in `GetRandomValue(min, max)` and `LoadRandomSequence(count, min, max)` does not provide uniformly random results.

That is because 

```
value = (rand()%(abs(max - min) + 1) + min)
```

will not be uniform  even though `rand()` is.   According to Knuth, the bias is small when `abs(max-min)` is small, but the larger it is, the greater is the bias.  I am looking at Knuth's explanation and solution in *The Stanford GraphBase* Code for  GB_FLIP section 12. Uniform Integers.

Also, since `GetRandomValue(min, max)` already interchanges the values so that `max > min`, the abs(max-min) is unnecessary.  The cautions about overflowing either `INT_MAX` and `RAND_MAX` remain important.

### Environment

This applies to all platforms where `rand()` is supplied by the C Language `<stdlib.h>`.

### Issue Screenshot

The issue is visible in the code of `rcore.c`.

### Code Example

I am not going to the effort to demonstrate the bias.  It occurs when `RANDMAX` is not a multiple of `max-min+1`, so some remainders occur more than others.

### Proposed Solution

First, `LoadRandomSequence` should use `GetRandomValue(min, max)` and that will also protect against deviant parameter cases. 
If that is undesirable, unwind the key part of updated `GetRandomValue(min, max)` there. 
**UPDATE** `LoadRandomSequence` is currently inefficient and switching to use of a corrected `GetRandomValue` is useful but still inefficient..  A Separate Issue will be raised on that account.

Secondly, in `GetRandomValue(int min, int max)`, replace the line

```
value = (rand()%(abs(max - min) + 1) + min;
```

with the following code 

```C
const long m = (long) max - (long) min + 1;    // # values in range 0 to m-1
const long c = (long) RAND_MAX+1;               // # values in range 0 to RAND_MAX
const long t = c - c % m;                                  // largest multiple of m not greater than c
long r;

do r = (long) rand();
     while (t <= r)                                               //  reject r's not inside the uniform t box
value = min + (int)(r % m);
```

I have not tested this code.  I have probably failed the raylib code-formatting requirements.
This is, by the way, demonstration of when `do ... while` is perfect.   

KEY POINTS: The use of `long` is to deal with cases where `RAND_MAX` is `INT_MAX`.  An useful case to desk-check is with `min = 0` and `max = RAND_MAX`.  It will just fall through.   This code will fail if `m` (that is, `max-min+1`) is greater than `RAND_MAX`, even if `max - min +1` does not overflow.  The precautions at the beginning of `GetRandomValue` are important.

Marcos-D:
Fixes #5374

This PR fixes a subtle modulo bias in GetRandomValue() when the fallback
rand()-based implementation is used (SUPPORT_RPRAND_GENERATOR disabled).

Let c = RAND_MAX + 1 and m = (max - min + 1). If we write c = q·m + r with 
0 ≤ r < m, then for outputs k < r:

    P(rand() % m == k) = (q + 1) / c

and for k ≥ r:

    P(rand() % m == k) = q / c

This produces a non-uniform distribution whenever r ≠ 0. (Knuth, TAOCP –
Uniform Random Integers.)

To verify this behavior, I wrote a Python script that computes theoretical 
probabilities and compares them with empirical samples from:

 • the current modulo-based implementation
 • a corrected rejection-sampling implementation

The script is available here: https://gist.github.com/Marcos-D/9ee6830a2c3d3e165d50e5ab19e52afc

Results:
 • For RAND_MAX = 32767 and m = 10, theory predicts two distinct probabilities 
   (≈0.1000061 for 0–7, ≈0.0999756 for 8–9). Empirical data matches.
 • Using RAND_MAX = 14 and m = 10 shows a dramatic 0.1333 vs 0.0667 split.
 • The corrected rejection sampler produces uniform frequencies ≈ 1/m.

The fix replaces rand() % m with a rejection-sampling step that preserves 
uniformity across all valid ranges.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -1758,11 +1758,12 @@ int GetRandomValue(int min, int max)
         unsigned long t = c - (c % m);                    // largest multiple of m <= c
         unsigned long r;
 
-        do
+        for (;;)
         {
             r = (unsigned long)rand();
+            if (r < t) break;   // Only accept values within the fair region
         }
-        while (r >= t);
+
 
         value = min + (int)(r % m);
     }


orcmid (approved):
I always have to remind myself that the accepted uniform-sampled range is `0` to `t-1` and the **while** condition is correct.

src/rcore.c
```diff
@@ -1743,7 +1743,29 @@ int GetRandomValue(int min, int max)
...
+    }
+    else
+    {
+        // Rejection sampling to get a uniform integer in [min, max]
+        unsigned long c = (unsigned long)RAND_MAX + 1UL;  // number of possible rand() results
+        unsigned long m = (unsigned long)range;           // size of the target interval
+        unsigned long t = c - (c % m);                    // largest multiple of m <= c
+        unsigned long r;
+
+        do
```
raysan5: in raylib I try to avoid `do { } while` loops, please, could you review it?

orcmid (approved):

raysan5:
@Marcos-D @orcmid thanks  for the review!

raysan5:
@Marcos-D @orcmid I'm wondering... will this change break all the applications/projects using previous implementation with a predefined seed?

Marcos-D:
In the old implementation, GetRandomValue() always consumed exactly one rand() call.
In the new rejection-sampling implementation, it may consume multiple rand() calls whenever a draw falls in the rejection region.

Because the RNG state now advances by a different number of steps, the sequence of values produced from the same seed (e.g., SetRandomSeed(2)) will not match the sequence produced by previous raylib versions.

This change only affects deterministic behavior across versions. Existing binaries are unaffected, but projects that rely on fixed seeds for reproducible layouts, enemy placement, world generation, or tests will observe different results after upgrading raylib.

A useful analogy is Minecraft: using the same world seed after an update, but generating a different world because the underlying RNG algorithm changed.

raysan5:
> projects that rely on fixed seeds for reproducible layouts, enemy placement, world generation, or tests will observe different results after upgrading raylib.

Mmmh... that represents a dilema, should in this case the algorithm be changed? is it really that big the improvement to potentially break some projects? also, now the function cost (and complexity) have increased, worths it? I'll keep this change for now but I have some doubts about the overall benefits of this kind of improvements...

orcmid:
@raysan5 
> Mmmh... that represents a dilemma, should in this case the algorithm be changed? is it really that big the improvement to potentially break some projects? also, now the function cost (and complexity) have increased, worths it? I'll keep this change for now but I have some doubts about the overall benefits of this kind of improvements...

A better way to look at this is that it is a long-standing bug that is finally fixed.  Also, the greater the value of `max-min+1` is, the more the results are distorted.  The bias is essentially (max-min+1) % RAND_MAX.  Also, reductions are infrequent.  According to Knuth (or someone), around 1.06 rands get generated per call.   That's because the largest multiple of (max-min+1) not greater than RAND_MAX is usually a sizeable number and rejection becomes the exception.

With regard to reproducibility, that depends a lot on rand( ) too.  Someone who is depending on reproducibility might want to look at an entirely different random number generation scheme and is likely not depending on raylib for it.  Not fixing this bug runs up against the whole idea of **uniform** *random*.  

I agree that rand( ) sequences in the standard C library are required to be reproducible.  But there already are differences between Linux and other builds because of the alternative use of `SUPPORT_RPRAND_GENERATOR`.  Also, the repaired generation of uniform randoms is also reproducible, just different.

I stumbled over this because I have a project where I am looking into (uniform) random number generators and porting DieHard tests of randomness along with that.  I will have to run the DieHard checks on the old and proposed new versions as tests of my work.  I will do timing comparisons too: DieHard tests require generation of gigawords of sample values.  It will be a while, but my focus is on this for now.

ocornut:
Any software relying on deterministic seed based random would likely want to use their own random implementation anyhow.

orcmid:
@ocornut
> Any software relying on deterministic seed based random would likely want to use their own random implementation anyhow.

I don't want to make any assumptions about the level of sophistication demanded of those using `GetRandomValue`.  I presume most simply trust in the raylib definition of the function.  

If correcting the bias is considered a breaking change, it might be identified as such in the next official stable release of raylib.  It is now corrected in the current build, so attentive users of raylib have an opportunity to see its impact, if any.

raysan5:
In any case, the old implementation has been left commented explaining the issue and why the change. Yes, it will be stated on next release. 

Also note that this is not the default PRNG in raylib, it was already updated to the new [`rprand.h`](https://github.com/raysan5/raylib/blob/master/src/external/rprand.h) some years ago, based on `Xoshiro128**` and `SplitMix64` algorithms.

orcmid:
@raysan5 
> Also note that this is not the default PRNG in raylib, it was already updated to the new [`rprand.h`](https://github.com/raysan5/raylib/blob/master/src/external/rprand.h) some years ago, based on `Xoshiro128**` and `SplitMix64`

Are you saying that `SUPPORT_RPRAND_GENERATOR` is set by default?

raysan5:
>  Are you saying that `SUPPORT_RPRAND_GENERATOR` is set by default?

Yes, it has been since October 2023, it was changed for `raylib 5.0`.

orcmid:
@raysan5 @Marcos-D 
> > Are you saying that `SUPPORT_RPRAND_GENERATOR` is set by default? 
> Yes, it has been since October 2023, it was changed for `raylib 5.0`.

I couldn't find where that is set but I did look at `external/rprand.h`.  There  `rprand_load_sequence` has exactly the same defect as with `rcore.h` except that it uses `rprand_xoshiro( )` instead of `rand( )`.  The choices are still not uniform.

`rprand_get_value(min, max)` is also not uniform in exactly the same way the `rcore.c` alternative was not.

I don't question the quality of Xoshiro128 and SplitMix64.  But the uniform-over-range solutions have the same defects that are observed and now repaired for  `rcore.c`.  I guess we should check with the author :) .

PR #5394, opened 2025-12-07
[examples] Add `shaders_game_of_life`

JordSant:
Example to show an implementation of Conway's Game of Life.

The Game of Life follows a simple set of rules on a 2D grid with cells in two possible states.
Interesting patterns emerge from the interaction of the previous patterns and the simple rules.

In this implementation some basic starting patterns are used in pre-stored pngs, that the user can select from a list.

The user also can draw its own patterns with the mouse.
For performance, the next generation is computed with a fragment shader on the world of size 2048 x 2048 cells.

### DIFF
--- a/examples/shaders/shaders_game_of_life.c
+++ b/examples/shaders/shaders_game_of_life.c
@@ -92,7 +92,7 @@ int main(void)
 
     int preset = -1;            // No button pressed for preset 
     int mode = MODE_RUN;        // Starting mode: running
-	bool buttonZoomIn = false;  // Button states: false not pressed
+    bool buttonZoomIn = false;  // Button states: false not pressed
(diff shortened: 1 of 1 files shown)


raysan5:
@JordSant thanks for the addition! this example is a bit bigger than usual but still veery nice! Thanks!

JordSant:
Thanks! It's a pleasure working on raylib, and the game of life is something I wanted to do for some time.
Sorry if it's a bit big, but I couldn't add all the functionality in less lines.
I created another PR to fix the loading of the resources. When approved and merged, please recompile it and upload it to the web.
Please recompile also the `shaders_mandelbrot_set` example, because it's not working on the raylib web.
Thanks!

PR #5395, opened 2025-12-07
[examples] Add `textures_cellular_automata`

JordSant:
Added example for Wolfram's one dimensional elementary cellular automata.

When the rule is changed, both by selecting a preset rule or by changing one bit of the rule, the new automaton is computed line by line.

The new lines are computed on an image with `GetImageColor` and `ImageDrawPixel`, then the texture is updated with `UpdateTexture`.

 The input is done via simple rectangle drawing.


raysan5:
@JordSant Another great example! Thanks!

PR #5399, opened 2025-12-09
[examples] Fixed `shaders_game_of_life` for web

JordSant:
Fixed pre-loading of resources for web compilig of the example `shaders_game_of_life`.

### DIFF
--- a/.github/workflows/update_examples.yml
+++ b/.github/workflows/update_examples.yml
@@ -29,8 +29,9 @@ jobs:
       shell: bash
 
     - name: Build and run rexm tool (requires GNU Makefile)
-      # "rexm update" validates and updates all required examples in raylib and even raylib.com repos,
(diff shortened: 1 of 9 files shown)


raysan5:
@JordSant this is a bit tricky, this change will be override on next `rexm update` because the tool scans the file looking for `Load*()`  functions and `"filename_to_load"` to automatically detect required resources and add them to build system.

Would it be possible to refactor the code to load all images at initialization and just use rquired one on pattern change to avoid the `pattern = LoadImage(TextFormat("resources/game_of_life/%s.png", presetPatterns[preset].fileName));`?

JordSant:
Yes, later today i will do it

raysan5:
@JordSant thanks, no hurries!

raysan5:
@JordSant Great! Thanks for the review!

PR #5401, opened 2025-12-10
[core] Camera2d comment updates

JeffM2501:
This PR makes the comments on the camera 2d fields more clear about what space each one is in.


raysan5:

PR #5409, opened 2025-12-14
Fix misleading example text.

Kaluub:
The "Core 2D camera" example has a panel stating "Right/Left to move Offset". This used to refer to the camera offset, but this is no longer the case, it only moves the player now. The text was misleading due to this (may have been cause of some confusion in help-general on raylib Discord server). This change just makes it say "player" instead to avoid this confusion.

The image is also recreated to reflect the new text.


raysan5:
@Kaluub thanks for the review!

Issue #5405, opened 2025-12-12
[build] Building with Zig fails with undeclared identifier 'MAX_FILEPATH_LENGTH'


cryptocode:
### Issue description

I'm trying to use raylib master branch as a dependency in a Zig project, but running into an error.

To repro, clone master branch and build directly with Zig 0.15.2 (my repro was done on macOS):

`zig build`

Output:

```bash
raylib/src/rmodels.c:4313:21: error: use of undeclared identifier 'MAX_FILEPATH_LENGTH'
    char currentDir[MAX_FILEPATH_LENGTH] = { 0 };
                    ^
raylib/src/rmodels.c:4314:48: error: use of undeclared identifier 'MAX_FILEPATH_LENGTH'
    strncpy(currentDir, GetWorkingDirectory(), MAX_FILEPATH_LENGTH - 1); // Save current working directory
```

### Environment

macOS aarch64, Zig 0.15.2

raysan5:
@cryptocode Are you defining `EXTERNAL_CONFIG_FLAGS`? `MAX_FILEPATH_LENGTH` is defined by `config.h` that is included by `rmodels.c` at the beginning:
```c
#if !defined(EXTERNAL_CONFIG_FLAGS)
    #include "config.h"     // Defines module configuration flags
#endif
```

cryptocode:
I am not defining EXTERNAL_CONFIG_FLAGS. Just tried on Linux as well - all I did to repro was clone this repo and run zig build.

I bisected this to 2a566544d4dabdd29628747e0ea2f18de063b69a - building works fine on commits before that. I guess it's because that commit introduces a usage of `MAX_FILEPATH_LENGTH`

raysan5:
@cryptocode yes, I added it yesterday but it works for all platforms in C, there should be some issue with Zig build system...

olaron:
EXTERNAL_CONFIG_FLAGS seems to always be defined when building with zig because of this change : [fixed platform change for zig build file #4531](https://github.com/raysan5/raylib/pull/4531/commits/9fe728b80223df96b85922e11a6fb5b9ffbb8b3b#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR97)
This was probably a mistake that had no consequence at the time ? Or was this by design in some way ?
[...]

PR #5410, opened 2025-12-14
Fix EXTERNAL_CONFIG_FLAGS being defined even when no custom config is used when building with zig

olaron:
This fixes issue #5405


raysan5:
@olaron thanks for the review

PR #5411, opened 2025-12-15
Modify text_words_alignment.c

Luca-coder07:
There are errors when pressing up/left touch twice so this is the change that I manage these erroes


raysan5:
@Luca-coder07 thanks for the review

PR #5414, opened 2025-12-16
[rcore] Implement `FLAG_WINDOW_ALWAYS_RUN` on Android

### LINKED ISSUE #5412: [rcore][android] Inhibit event waiting on Android focus loss
### Request for Enhancement

Hello, I'm working on an OpenXR interop module called [rlxr](https://github.com/caszuu/rlxr/tree/main) and so far raylib has been amazing. I was able to get away with using upstream raylib until now but with Android i hit a minor issue.

When raylib receives the `APP_CMD_LOST_FOCUS` command from the system, It pauses the main game loop by blocking in `PollInputEvents()` until a `APP_CMD_GAINED_FOCUS` command is received again. This works nicely for flatscreen android apps but with XR this might block even when the app is still visible and needs to output frames. (see bellow)

So It would be nice to prevent `PollInputEvents()` from blocking to some capacity if the app requests it or not reacting to `_FOCUS` commands preferring `_PAUSE` and `_RESUME` commands instead.

### Environment

Issue is Android specific and possibly vendor specific. Confirmed to be an issue on Meta headsets.

### Issue Screenshot

https://github.com/user-attachments/assets/f8c86fd0-e9b1-4803-912a-c622d3baa963

### Code Example

One possible way to implement this would be to make use of the `FLAG_WINDOW_ALWAYS_RUN` flag which would prevent `rcore_android.c` from blocking for events if set. OpenXR provides its own focus mechanism, so this would be ideal in my case, though I'm not sure if it would be an abuse of the raylib API.

caszuu:
Allow the app to keep running while not in foreground. 

I'm not entirely sure if this is the right way to do it as it allows the app to completely side-step the Android process lifecycle. With this change the app can for example run full-speed without VSync (due to no surface being present) while android already sent `APP_CMD_TERM_WINDOW` and `APP_CMD_STOP` to the app.

A check was added to `SwapScreenBuffers()` to at least not submit frames while a surface is not present, but not sure if this is ideal for battery related reasons.

Closes #5412.

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -357,13 +357,125 @@ void RestoreWindow(void)
 // Set window configuration state using flags
 void SetWindowState(unsigned int flags)
 {
-    TRACELOG(LOG_WARNING, "SetWindowState() not available on target platform");
+    if (!CORE.Window.ready) TRACELOG(LOG_WARNING, "WINDOW: SetWindowState does nothing before window initialization, Use \"SetConfigFlags\" instead");
+
+    // State change: FLAG_WINDOW_ALWAYS_RUN
+    if (FLAG_IS_SET(flags, FLAG_WINDOW_ALWAYS_RUN))
+    {
+        FLAG_SET(CORE.Window.flags, FLAG_WINDOW_ALWAYS_RUN);
+    }
+
+    // Setting other window flags is not supported on android
+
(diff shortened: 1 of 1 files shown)


raysan5:
@caszuu thanks for the improvement, I'm reviewing the PR to minimize warning logs, specially on potential loop called functions, to avoid spamming the log.

PR #5415, opened 2025-12-16
[build] fixed build errors with zig. now compatible with zig master 0.16.0-dev.1593+c13857e50. still backwards compatible with 0.15.1

dtasada:
building with zig was broken due to custom config flags being set by default.


raysan5:
@dtasada thanks for the review!

PR #5419, opened 2025-12-20
[rcore] Add IsCursorEnabled

squidink7:
Adds a simple function to query if the cursor has been captured.


raysan5:
@squidink7 I think there is no need for this function, it can be tracked at user level.

PR #5424, opened 2025-12-23
[rcore] Fix stray negation in Android

caszuu:
This is a minor oversight from updating #5414. Works as intended after fix.


raysan5:
@caszuu thanks for the review

Issue #5316, opened 2025-10-29
[raymath] Considering adding some SIMD optimizations


raysan5:
Recently a new software renderer backend ([rlsw](https://github.com/raysan5/raylib/blob/master/src/external/rlsw.h)) was added to raylib and some performance considerations arised to improve speed. 

`rlsw` added support for [some SIMD optimizations](https://github.com/raysan5/raylib/blob/master/src/external/rlsw.h#L1113) and I'm considering adding something similar to `raymath`, at least for the most critical math functions.

As [explained in a previous PR](https://github.com/raysan5/raylib/pull/4599#issuecomment-2543202239) (#4599), using SIMD has some implications, target devices hardware MUST support the required SIMD vector instructions (SSE, AVX, RVV...) and only users aware of their target systems should use those vector extensions... but I think it could be really useful for the special cases where software rendering is required.

It could really improve performance considerably.

orcmid:
The TL;DR: Any SIMD dependence in raymath should be with alternative, similarly-named, functions, not by replacing existing ones.  There should also be a function, probably platform-dependent, for detecting whether a processor has SIMD functions or not.  This approach may need to be extended to other functions of raylib.

  That makes it possible for developers to craft agnostic releases if that is important for the intended distribution of a raymath-dependent product.

 - - - - - - - - - My analysis - - - - - - - -

There is a worrisome pattern here, and a potentially promising, but careful, solution.

The limitation of Windows 11 to computers that have SIMD extensions has created a considerable problem.  At least there was a way to learn whether a computer was eligible for upgrade or not.  The end of Windows 10 support is not going well, however.

Not so up-front were two difficulties I observed.  Ableton Live 12 (and the Live Lite 12 version) all require SIMD extensions.  But they would install on eligible systems (that is, of people who already owned the version 11 product) but failed to run.  They finally fixed things so 12 would not install.  But the explanation to ordinary users was and is not helpful.  Apparently, no one wants to simply say that Windows 11 (eligibility) is required.

The same happened with the social world browser, Firestorm, for Second Life.  That approach was to issue two versions of recent updates.  One that used the extensions, and reported their absence, and one that continues to not use them.  The explanation to users is also inept, but installing the wrong one will lead to a report.  The problem is users needed to know more about the processor on their computer in terms that most users will not understand (the usual problem of developers thinking users should be like themselves).

[...]

PR #5427, opened 2025-12-23
[raymath] Added SSE to `MatrixMultiply()`

mcdubhghlas:
I saw your LLM challenge on Twitter and decided to try it. I had to edit it a bit to make it work, but hey. It runs and it does what you need it to. I saw https://github.com/raysan5/raylib/issues/5316 and was like "oh yeah."

I am getting roughly half a second per 200000000 ops, with a range of 2.5ns - 3.2 ns per op. The after picture is roughly one second per 200000000 ops, with a range of 4.6ns - 5.1ns per ops.

I used an LLM to generate the test: https://gist.github.com/mcdubhghlas/53f780c35482aa87713f34ded941a67d

It was compiled using this: `cc benchmark.c src/libraylib.a -Isrc -O3 -march=native -lm -lpthread -ldl -lrt -lX11 -o bench`

Let me know if that is insufficient, I genuinely was too busy to read over documentation to see if you had a test suite -- That said, I love you what you do Ray. Keep up the good work man :)

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -170,6 +170,11 @@ typedef struct float16 {
 
 #include <math.h>       // Required for: sinf(), cosf(), tan(), atan2f(), sqrtf(), floor(), fminf(), fmaxf(), fabsf()
 
+#if defined(__SSE__) || defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 1)
+    #include <xmmintrin.h>
+    #define RAYMATH_SSE_ENABLED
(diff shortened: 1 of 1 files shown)


raysan5:
@mcdubhghlas did you try this change with some raylib examples, to verify that nothing is broken? Specially with 3d examples...

raysan5:
@mcdubhghlas thanks for the improvement, I'm reviewing it to allow users select the SIMD option is desired

Issue #5429, opened 2025-12-23
[rtext] LoadCodepoints is broken, returns NULL


KiviTK:
I tried using the `int *LoadCodepoints(const char *text, int *count)` function from rtext today and discovered a bug causing the function to correctly parse the codepoints but still return `NULL`. 

The assignment on line 2106 is done on an inner scope variable that shadows the outer scope variable that is meant to be set.
I will submit a PR for this, but I suspect there are more errors with loading individual utf8 codepoints with `LoadFontsEx`, I will try to look into it further, I just needed this functionality and fixed it in my version so I will push a fix.

```
// Load all codepoints from a UTF-8 text string, codepoints count returned by parameter
int *LoadCodepoints(const char *text, int *count)
{
    int *codepoints = NULL;
    int codepointCount = 0;
    
    if (text != NULL)
    {
        int textLength = TextLength(text);

        // Allocate a big enough buffer to store as many codepoints as text bytes
        int *codepoints = (int *)RL_CALLOC(textLength, sizeof(int));

        int codepointSize = 0;
        for (int i = 0; i < textLength; codepointCount++)
        {
            codepoints[codepointCount] = GetCodepointNext(text + i, &codepointSize);
            i += codepointSize;
        }

        // Create second buffer and copy data manually to it
        int *temp = (int *)RL_CALLOC(codepointCount, sizeof(int));
        for (int i = 0; i < codepointCount; i++) temp[i] = codepoints[i];
        RL_FREE(codepoints);
        codepoints = temp; // codepoints is shadowed, only overwrite local
    }

    *count = codepointCount;
    return codepoints;
}
```

PR #5430, opened 2025-12-23
[rtext] Fixed LoadCodepoints returning NULL

KiviTK:
Due to a bug in LoadCodepoints the function shadowed an outer scope variable causing the function to return NULL in all cases.
See #5429


raysan5:
@KiviTK Good catch! Thanks for the review!

PR #5431, opened 2025-12-24
[rcore][rgfw] RGFW also requires RGBA8 images as window icons, as raylib already reports

Crisspl:
Raylib already documents this requirement in its own API in raylib.h:
```
RLAPI void SetWindowIcon(Image image);                            // Set icon for window (single image, RGBA 32bit)
RLAPI void SetWindowIcons(Image *images, int count);              // Set icon for window (multiple images, RGBA 32bit)
```


Crisspl:
It didnt work for RGBA8 as well though, since `RGFW_formatToChannels` reported it as 2-channel format. In this commit it requires RGBA8 and passes constant 4 as number of channels.

raysan5:
@Crisspl thanks for the review!

PR #5437, opened 2025-12-24
DudydeeCreate SECURITY.md for security policy

Noprompramzeez:
Add a security policy document outlining supported versions and vulnerability reporting.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.


raysan5:
@Noprompramzeez That document looks like a template, could you please fill it?

raysan5:
Document is an empty template, closing.

orcmid:
@raysan5 @Noprompramzeez @oneafter

Yes, the only part of the template that was filled in had to do with selected versions that are supported or not with respect to security defects and mitigations.

The idea is to state a security policy.  The form for reporting a suspected/confirmed vulnerability is different.

You can see a policy statement at <https://github.com/orcmid/nfoTools/security>.  (NOTE: If you own the repository, it looks different when you use the security counterpart and have created a policy.  But then `SECURITY.md` will be in the repository root, and can be reviewed and edited there.)  My `nfoTools` one is similar to what I would suspect is applicable for raylib.  If there were a security policy for raylib, I presume @oneafter would have relied on it.

I have also provided a "sandbox" version to help readers of my security policies (one per repository) see what is involved in making a vulnerability report.  One can even make a practice report and see how that works.  The sandbox version is at
<https://orcmid.github.io/SECURITY.html>.  Cybersecurity experts will understand more of the fields.  The sandbox tutorial assumes non-expert reporting and shows how to ignore the finer details.

**NOTE**: It can be useful to create a Security Policy that says you don't want the mechanism to be used and that you want reports to be via repository Issues.  The problem with this, for some projects, is that revealing an exploitable vulnerability in public is not a great idea, and the discovery of a vulnerability does not mean the discoverer is in a position to submit a fix/mitigation.  There are code-in-the-open projects (e.g., LibreOffice) where security patches are snuck in as part of other fixes so that malicious observers don't get a hint about unpatched security flaws in existing releases before the fix (and security warnings) ship.  There remains the problem of existing users staying informed and receiving/heeding the advice to update.  If there is an automatic update scheme, that's a big help.

raysan5:
@orcmid thanks for the feedback, I created the document just in case but I don't know how I should proceed from here neither if I have the resources to do it.

orcmid:
@raysan5 
I think that statement is fine.  With regard to the TODO there, you might want to provide advice for the limited case of an *exploitable* security defect and how that could be communicated privately.   The "report a vulnerability" option does take care of that.  In general, if someone detects an exploitable vulnerability, they probably know how to use that report.  

I wouldn't offer anything else in your policy.  The rest can be worked out once private communication is established via the GitHub security-handling mechanism.

Side comment: In some respect, raylib is part of someone's tool chain.  Exploitation, if any, lurks in the downstream application of raylib and those projects need to be the vigilant ones, depending on who the end-users are.  I don't foresee any good way to *securely* coordinate at that level.  Quietly referring someone to the source of an external becomes nice, though.

orcmid:
> @raysan5 
Big PS: If you enable vulnerability reporting, that will give cybersecurity and other folks a private, secure connection to you, and you can reply via that secure arrangement.  Even if you don't find it worth pursuing, it gives you a private way of responding.  Also, that's a good private way to refer someone to an upstream source, such as for an external.

Also: cybersecurity professionals will generally ask for a private way of communicating, especially if they have discovered an exploitable vulnerability.  Having vulnerability reporting enabled allows them to do that privately, whether they should be redirected elsewhere or not.

PR #5439, opened 2025-12-24
#5387 - Fix keyboard input detected as gamepad on some Android devices

### LINKED ISSUE #5387: [core] Android keyboard input detected as gamepad input
### Issue description

On my device (Motorola Razr 2024), keyboard inputs have both the `AINPUT_SOURCE_KEYBOARD` and `AINPUT_SOURCE_GAMEPAD` flag set, resulting in the keypress being registered as a gamepad input event. This might be because they both share a common bit (`AINPUT_SOURCE_CLASS_BUTTON`).

### Environment

Android 34, Motorola Razr 2024, OpenGL ES 2

kellemar:
**Fixes #5387**

_Decided to take the challenge of using an LLM to fix a bug. I have experiences building iOS and Android games with Cocos2D back in the day, so this looks like a straightforward fix (LGTM). Though I could be wrong._ 

Some Android devices report both AINPUT_SOURCE_KEYBOARD and AINPUT_SOURCE_GAMEPAD flags for keyboard events. This causes keyboard input to incorrectly trigger gamepad detection, making IsGamepadAvailable() return true and registering phantom gamepad button presses.

The fix adds a check to skip gamepad handling when AINPUT_SOURCE_KEYBOARD is set, since a real gamepad wouldn't have the keyboard flag.

Also added a test example (core_input_keyboard_gamepad_test.c) that visualizes both keyboard and gamepad state side-by-side - useful for verifying the fix on affected devices.

**Testing:**
- Build for Android and test on a device that has this issue
- Press keyboard/arrow keys - gamepad should stay "NO" and only keyboard input should register

### DIFF
--- a/examples/core/core_input_keyboard_gamepad_test.c
+++ b/examples/core/core_input_keyboard_gamepad_test.c
@@ -1,0 +1,173 @@
+/*******************************************************************************************
+*
+*   raylib [core] example - Keyboard vs Gamepad Input Test
+*
+*   Example complexity rating: [★☆☆☆] 1/4
+*
+*   This example is a diagnostic tool to verify that keyboard input is not
+*   incorrectly detected as gamepad input on Android devices.
+*
+*   Issue reference: https://github.com/raysan5/raylib/issues/5387
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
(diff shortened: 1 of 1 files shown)


raysan5:
@kellemar thank you very much for working on this issue, reported the fix to user for further testing before merging. Thanks!

raysan5:
@kellemar I'm merging the fix, note that I'm not sure if the example will be kept.

PR #5441, opened 2025-12-24
Fixed FLAG_IS_SET to check if all bits in the flag are set in the value

vushu:
The problem affected Android volume buttons because:

Volume buttons are AINPUT_SOURCE_KEYBOARD (0x00000101)
This shares bit 0 with AINPUT_SOURCE_GAMEPAD (0x00000401)
With the broken macro, volume button events were incorrectly detected as gamepad input and consumed by the gamepad handler instead of being processed normally


raysan5:
@vushu Oh! Very good catch! Thanks for reviewing it!

Issue #5443, opened 2025-12-25
[rcore][glfw] AutomationEvent records gamepad axis motions for triggers while no gamepad is connected


LeapersEdge:
Please, before submitting a new issue verify and check:

- [x] I tested it on latest raylib version from master branch
- [x] I checked there is no similar issue already reported
- [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
- [x] My code has no errors or misuse of raylib

## Issue description

[`RecordAutomationEvent()`](https://github.com/raysan5/raylib/blob/master/src/rcore.c#L4047) inside `rcore.c` will always log axis motion for the axes `GAMEPAD_AXIS_LEFT_TRIGGER` and `GAMEPAD_AXIS_RIGHT_TRIGGER` while gamepad is disconnected, on `GLFW` platform backend.

This was tested with [automation events example](https://github.com/raysan5/raylib/blob/master/examples/core/core_automation_events.c) on `GLFW`, `SDL` and `RGFW` platform backends, and the problem was only present on `GLFW`.

This is because the criteria for recording the gamepad axis AutomationEvent is the following:
```c
float defaultMovement = (axis == GAMEPAD_AXIS_LEFT_TRIGGER || axis == GAMEPAD_AXIS_RIGHT_TRIGGER)? -1.0f : 0.0f;
if (GetGamepadAxisMovement(gamepad, axis) != defaultMovement)
```
It assumes the default triggers are `-1.0f`.
In `rcore_desktop_glfw.c`, inside `PollInputEvents()`,  `glfwGetGamepadState` is used to aquire gamepad state, but [glfwGetGamepadState sets axes to 0.0 if gamepad isn’t connected.](https://www.glfw.org/docs/latest/group__input.html#gadccddea8bce6113fa459de379ddaf051)

## Resulting automation.rae

All disconnected gamepads record AutomationEvent for both triggers:
[automation.rae.txt](https://github.com/user-attachments/files/24341075/automation.rae.txt)

LeapersEdge:
I have not reviewed any other platform backends since I don't have the devices (or Windows) to test them on.

I have opened [pull request](https://github.com/raysan5/raylib/pull/5444) with a proposed fix, tho I am not sure if there is any other part of the code that expects a default value when the gamepad isn't connected, like `RecordAutomationEvent()` does.

I'm very sorry if this issue and pull request are not done well. This is my first time doing this. Thank you for your time and hard work!

PR #5444, opened 2025-12-25
fix: set correct default axes for gamepads that are not connected (inside rcore_desktop_glfw.c)

LeapersEdge:
## Description

`glfwGetGamepadState` will set all gamepad state variables to 0.0 if requested gamepad is not connected, but `RecordAutomationEvent()` inside rcore.c expects trigger axes to be -1.0f when gamepad is not connected. Since SDL and RGFW return -1.0f in such case, this change is aligning it with them

### Related Issues
Fixes [#5443 ](https://github.com/raysan5/raylib/issues/5443)


raysan5:
@LeapersEdge thanks for the review, it looks ok. Just note that the return value is the result of the call, not only if the gamepad is connected or not. I'm reviewing that name.

PR #5453, opened 2025-12-29
Add GetCollisionV

danemadsen:
This is a function to get a vector for a collision which corresponds to how much the rectangles collide in each axis (relative to rec1). 

For example:
Given a Rectangle a = { 10, 10, 20, 20 } 
and Rectangle b = { 20, 10, 20, 20 }.

The resulting output of this function would be { 10, 0 }. 
Which means there is a collision of 10 units in the x axis.

If you flipped the order of the rectangles so rec1 = b and rec2 = a the resulting output would be { -10, 0 }.

This is handy for collisions in 2D games where you may want a character to be blocked from moving in one axis but not in another without doing two collision checks.

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -2470,28 +2470,23 @@ Vector2 GetCollisionV(Rectangle rec1, Rectangle rec2)
     result.x = 0.0f;
     result.y = 0.0f;
 
-    // Centers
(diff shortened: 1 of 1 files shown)


raysan5:
@danemadsen I think there is no need for an additional function, `GetCollisionRec()` can be used for that and managed in user side. Note that raylib priority is minimize API.

PR #5456, opened 2025-12-30
Remove stdio.h unused header

Luca-coder07:


raysan5:
@Luca-coder07 thanks for the review!

PR #5457, opened 2025-12-31
Fix window width calculation by adding wOffset

CosmosShell:
when asking for window size (GetScreenWidth ,GetScreenHeight) and also when we're creating a window it's not obvious wether we're specifying full window size (with all decorations, title bar, scroll bars, borders), or "client size" (using Windows' terminology) that should just match backbuffer's size.
This results in backbuffer being too small in X dimension by 16 pixels. RGFW currently used by raylib is taking into account title bar size (hOffset) when handling WM_SIZE but not X-dimension decorations.


raysan5:
@CosmosShell thanks for the review! raylib considers requested screen size on InitWindow() as the desired "client size" but note that depending on display ppi and if explicitly requested with `FLAG_WINDOW_HIGHDPI`, internal framebuffer render size could be bigger than screen size.

PR #5458, opened 2025-12-31
[rcCamera] Fix CameraMove forward and right with Z up cameras

JeffM2501:
CameraMoveForward and CameraMoveRight make the assumption that the camera up is Y when MoveInWorldPlane is true. This does not work if the camera up is not Y.

This PR forces the axis component that the camera up vector is in to 0 so that the world plane is normal to the up vector regardless of what axis is up.


raysan5:

PR #5459, opened 2025-12-31
[rCamera] Add a tolerance to camera movement plane detection

JeffM2501:
This PR adds a fix to PR #5458, a non zero value in an up vector component is not enough to determine what axis is up, some up vectors may be non-orthogonal. This PR adds a tolerance to see if the up axis is mostly pointing along an axis, then picks a plane from that. This should just prevent edge cases from giving people odd behaviors, and always ensure we fallback to Y up if we can't figure it out for legacy compatibility.

### DIFF
--- a/src/rcamera.h
+++ b/src/rcamera.h
@@ -255,8 +255,8 @@ void CameraMoveForward(Camera *camera, float distance, bool moveInWorldPlane)
     if (moveInWorldPlane)
     {
         // Project vector onto world plane (the plane defined by the up vector)
-        if (fabsf(camera->up.z) > 0) forward.z = 0;
-        else if (fabsf(camera->up.x) > 0) forward.x = 0;
+        if (fabsf(camera->up.z) > 0.7071f) forward.z = 0;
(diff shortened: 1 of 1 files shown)


raysan5:

casperbear:
@JeffM2501 Since you recently looked into camera functions could you please check if it's better to add early return in the beginning of CameraMoveToTarget. I have strong suspicion that, even though camera->position shouldn't change mathematically speaking, float operations accumulate small error and camera->position does change, which can trigger collision depending on how sensitive it is.
```
void CameraMoveToTarget(Camera *camera, float delta)
{
    if (delta == 0.0f) return;

    float distance = Vector3Distance(camera->position, camera->target);

    // Apply delta
    distance += delta;
    // Distance must be greater than 0
    if (distance <= 0) distance = 0.001f;
    // Set new distance by moving the position along the forward vector
    Vector3 forward = GetCameraForward(camera);
    camera->position = Vector3Add(camera->target, Vector3Scale(forward, -distance));
}
```

PR #5463, opened 2026-01-02
[example] Small update to textured curve example.

JeffM2501:
This PR updates the textured curve example to correctly use the texture size when computing the texture UV coordinates, so that repeating textures look good. The current code only looks good because the road texture can repeat on one pixel, if it was a pattern it would not look good. This change fixes that.


raysan5:

Issue #5460, opened 2025-12-31
[build] [web] [examples] `!=` not supported on default macos `make`


pentallelogram:
`!=` makefile sytax is not supported until GNU Make 4.0. OSX ships GNU Make 3.81. This breaks web builds of the examples outside of windows platforms

```make
ifeq ($(TARGET_PLATFORM),$(filter $(TARGET_PLATFORM),PLATFORM_WEB PLATFORM_WEB_RGFW))
    ifeq ($(OS),Windows_NT)
        MAKE = mingw32-make
    else
        EMMAKE != type emmake # <----
        ifneq (, $(EMMAKE))
            MAKE = emmake make
        else
            MAKE = mingw32-make
        endif
    endif
endif
```

I fixed by replacing the offending line with `EMMAKE := $(shell type emmake)`

Peter0x44:
I feel like trying to support 20yo gnu make versions is not really worth it...
Can't you get a more modern one from brew?

sleeptightAnsiC:
> I feel like trying to support 20yo gnu make versions is not really worth it...
> Can't you get a more modern one from brew?

Keeping the support for 3.81 is common. It's the last one being GPLv2 and the most useful functionality, like `$(call ...)` or `$(eval ...)`, is already there.

> != makefile sytax is not supported until GNU Make 4.0. 

I'm quite familiar with GNU Make and yet this `!=` assignment is new to me. It does not appear anywhere in https://www.gnu.org/software/make/manual/html_node/Quick-Reference.html and I was just able to find it in https://www.gnu.org/software/make/manual/html_node/Setting.html It's hard to believe that someone would use it in this case.

Also, the `EMMAKE != type emmake` seems very ***very*** strange here:
The PR that added this was https://github.com/raysan5/raylib/pull/4434
Why does it use `type` instead of `command -v` ?
and why `!= ...` instead of `= $(shell ...)` ?
Is this some kind of Windows CMD.EXE trick?

Peter0x44:
> It's the last one being GPLv2

Yes. That's why apple ship it...


> Is this some kind of Windows CMD.EXE trick?

i doubt it... the windows cmd equivalent of that is `where.exe`. Neither of `which` `command -v` or `type` would work.

If you want to know how to detect the shell used by the Makeflie reliably, I wrote a blog post about it:
https://peter0x44.github.io/posts/make_shell_cmd/

raysan5:
[...]

PR #5464, opened 2026-01-02
fix(build): do not use != assignment in Makefiles

sleeptightAnsiC:
For more information see the commit message and the linked issue.

I was unable to test it on MacOSX because I don't have a setup with working XCode, and Emscripten requires XCode even when installing from Brew. Hopefully, a CI or some other contributor will check it, but this _should just work_ ™

Fixes: https://github.com/raysan5/raylib/issues/5460


sleeptightAnsiC:
> Hopefully, a CI [...] will check it

ahh, we don't have CI that would catch this exact case, but maybe we should have one (?)

raysan5:

Issue #4551, opened 2024-11-28
[module] Short description of the issue/bug/feature


konsumer:
Please, before submitting a new issue verify and check:

 - [X] I tested it on latest raylib version from master branch
 - [X] I checked there is no similar issue already reported
 - [X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [X] My code has no errors or misuse of raylib

### Issue description

As-per discussion [here](https://github.com/raysan5/raylib/discussions/1631) the filesystem callbacks are useful and work great, when trying to implement a virtual filesystem (I am using [physfs](https://icculus.org/physfs/)), but there are still some functions without hooks or macros that need to some way to be overwritten:

```
FileExists
DirectoryExists
GetFileLength
ChangeDirectory
IsPathFile
LoadDirectoryFiles
LoadDirectoryFilesEx
GetFileModTime
```

I made a little library [here](https://github.com/konsumer/raylib-wasmcart/blob/main/host/src/raylib-physfs.h) and just added `FS` to the end. Also @RobLoach made [raylib-physfs](https://github.com/robloach/raylib-physfs), which doesn;t use callbacks, but takes a similar naming-approach.

I think either a macro or the callback-system are fine, I just need a way to overwrite these, so when the user calls `FileExists` it calls `PHYSFS_stat`. I am happy to PR for a macro or callback, whatever everyone thinks is appropriate.

### Code Example

It might look like either of these:

```c

bool FileExistsPhysFS(const char *fileName) {
    PHYSFS_Stat stat;
    if (!PHYSFS_stat(fileName, &stat)) return false;
    return stat.filetype == PHYSFS_FILETYPE_REGULAR;
}

// later
SetFileExistsCallback(FileExistsPhysFS);

// OR

#define RL_FS_FILE_EXISTS FileExistsPhysFS
```

raysan5:
@konsumer Just realized that I did not answer to this improvement proposal, I'm still considering it. 

Just for completion, here it is the full list of functions accessing the file system:

- `LoadFileData()` - Implemented `SetLoadFileDataCallback()` for customization
- `SaveFileData()` - Implemented `SetSaveFileDataCallback()` for customization
- `LoadFileText()` - Implemented `SetLoadFileTextCallback()` for customization
- `SaveFileText()` - Implemented `SetSaveFileTextCallback()` for customization
- `FileExists()`
- `FileRemove()`
- `FileRename()`
- `FileCopy()`
[...]

PR #5465, opened 2026-01-02
[utils] Remove `utils` module, move FS functionality to `rcore` module

konsumer:
This is related to #4551

Basically it does this:

- Remove utils (moved to rcore, including android-specific stuff in `rcore_android`)
- Remove utils from build-systems
- use `RL_FS` macros so it can all be overwritten (before `#include "raylib.h"`)
- simple example that overrides macros with very chatty macros (no-dep tester.) I am not totally sure I am doing this right, but made it work sort of like `rlgl_standalone`

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -529,6 +529,7 @@ CORE = \
     core/core_clipboard_text \
     core/core_compute_hash \
     core/core_custom_frame_control \
+    core/core_custom_filesystem \
     core/core_custom_logging \
     core/core_delta_time \
     core/core_directory_files \
--- a/examples/core/core_custom_filesystem.c
+++ b/examples/core/core_custom_filesystem.c
@@ -1,0 +1,335 @@
+/*******************************************************************************************
+*
+*   raylib [core] example - Custom file system with RL_FS_* macros
+*
+*   This example demonstrates how to override raylib's file system operations using
+*   the RL_FS_* macros. Each file operation is wrapped to add detailed logging,
+*   showing when and how file operations occur.
+*
+*   NOTE: This example implements file I/O functions locally to demonstrate the macros.
+*         In a real project, you'd define these macros before compiling raylib itself.
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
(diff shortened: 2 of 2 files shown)


raysan5:
@konsumer This is a big change and I haven't decided yet if it worths it or the best way to implement it but there is no plats to populate `raylib.h` with that amount of macros. If implemented, and considering it is a compile-time option (only required by a VERY minimal number of users), macros will be added to `rcore.c` or `config.h`, not exposed on `raylib.h`.

Also, I prefer to use separate PRs for `utils` removal and macros definition (when decided if implemented).

konsumer:
> This is a big change and I haven't decided yet if it worths it or the best way to implement it 

Ok, that makes sense. Let me know what you decide, and I am happy to help.

> there is no plats to populate `raylib.h` with that amount of macros. If implemented, and considering it is a compile-time option (only required by a VERY minimal number of users), macros will be added to `rcore.c` or `config.h`, not exposed on `raylib.h`.

Yes, it is a lot of macros, for a limited use-case (very few people will likely want to do this sort of thing.) I put it in `raylib.h` so the macros would be available all over (since it's used in a bunch of places) but I could put it in `config.h` instead.

> Also, I prefer to use separate PRs for utils removal and macros definition (when decided if implemented).

I hear that. I did it all in 1 PR because I needed to implement some of the macros to move things out of utils, but it could be done more atomically.

It sounds like this idea needs more planning, and maybe will not even be included. I will close this PR, and if you want help on it later, let me know (on #4551)

raysan5:
@konsumer Actually, this this PR could have been reviewed to just include the `utils` module removal, that part looks great. Did you test if after that change Android build still work?

konsumer:
> @konsumer Actually, this this PR could have been reviewed to just include the `utils` module removal, that part looks great. Did you test if after that change Android build still work?

I don't mind making a cleaner PR with just that 1 change. Part of the move used some of the new macros, so it all seemed like it should go together, but I understand it's huge. I will make another PR that is just the move, then we can build the macros (or not) on top of that.

I wasn't sure how to test on Android. I think it should work, but it's totally untested.

raysan5:
@konsumer Ok, thanks, let's start with that PR, removing utils module. Note that `TRACELOGD()` macro has been removed, in the process of removing `utils` I'll be reviewing and redesigning some parts... Just sync to latest master branch.

PR #5466, opened 2026-01-03
[rcore] Fix typecast warnings in rcore

JeffM2501:
multiply assignment of an int by a float is a float, thus it will generate a typecast warning, This PR casts the value correctly


raysan5:

PR #5467, opened 2026-01-03
[examples] Cleanup warnings (part 1)

JeffM2501:
This PR fixes warnings in examples as reported by MSVC. There are a lot of them (over 150), so this PR is just one part so that it's not too much to review at once.


raysan5:
@JeffM2501 thanks for the review! Note that most of those new examples have not been properly reviewed yet, I'd like to improve some of them...

PR #5468, opened 2026-01-03
[examples] Added `textures_frame_buffer_rendering`

jackboakes:
The subject camera is a orbital camera that is filming the scene. This also has a cropped texture2D of the green section on the left-hand side.

The observer camera is a third person camera that looks at the subject camera and its prism.

The cameras prism is created by inverting the view and projection matrices. Instead of the prism showing the near and far planes, instead it's sliced at the target, this plane then connects to the cameras location.

I'm happy to make any changes, simplify or scrap it entirely if this isn't what you envisioned.

textures_frame_buffer_rendering


raysan5:
@jackboakes Very nice example! Thank you!

PR #5469, opened 2026-01-03
[rmodels] Fix `DrawMeshInstanced()` breaking if `instanceTransform` is unused

al13n321:
`DrawMeshInstanced` does `rlSetVertexAttribute(material.shader.locs[SHADER_LOC_VERTEX_INSTANCE_TX] + i, ...)` for i=0..3. But `material.shader.locs[SHADER_LOC_VERTEX_INSTANCE_TX]` can be -1 if the vertex shader doesn't use the vertex attribute `instanceTransform` (even if it's declared in the shader code). In such case these calls overwrite unrelated vertex attributes 0..2. In particular, 0 is usually `vertexPosition`, so all vertices end up having the same position, so nothing gets rendered, and no error is generated, and the user is frustrated.

Program reproducing the issue:
```c++
#include "raylib.h"

const char *vertex_shader = R"(
#version 330

in vec3 vertexPosition;
in mat4 instanceTransform;

out vec3 fragPosition;

uniform float zero = 0.;

void main()
{
    vec4 p1 = instanceTransform*vec4(vertexPosition, 1.0);
    vec4 p2 = vec4(vertexPosition, 1.0);
    gl_Position = p1*zero + p2;
}
)";

int main(void)
{
    InitWindow(600, 600, "oh no");
    Mesh cube = GenMeshCube(1.0f, 1.0f, 1.0f);
    Material matInstances = LoadMaterialDefault();
    matInstances.shader = LoadShaderFromMemory(vertex_shader, 0);
    while (!WindowShouldClose())
    {
        BeginDrawing();
        ClearBackground(RAYWHITE);
        Matrix transforms[2] = {0};
        DrawMeshInstanced(cube, matInstances, transforms, 2);
        EndDrawing();
    }
}
```
On my machine this renders white screen, with no cube. But if you change `float zero = 0.;` to `uniform float zero = 0.;` (so that the shader compiler can't optimize out the usage of `instanceTransform`), the cube shows up (black rectangle at the center of the screen).

(Why would a shader ever not use `instanceTransform`? In my case, I initially had a shader that uses it, but nothing was rendered, and as part of investigating it I did `gl_Position = vec4(vertexPosition, 1.0);` to check whether the problem is with my matrices or something else. The problem was with matrices, but this raylib bug misled me into thinking it was something else. Another use case I can imagine is passing custom vertex attributes using rlgl directly and using them instead of the default `instanceTransform`.)

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1762,11 +1762,14 @@ void DrawMeshInstanced(Mesh mesh, Material material, const Matrix *transforms, i
     instancesVboId = rlLoadVertexBuffer(instanceTransforms, instances*sizeof(float16), false);
 
     // Instances transformation matrices are sent to shader attribute location: SHADER_LOC_VERTEX_INSTANCE_TX
-    for (unsigned int i = 0; i < 4; i++)
+    if (material.shader.locs[SHADER_LOC_VERTEX_INSTANCE_TX] != -1)
     {
-        rlEnableVertexAttribute(material.shader.locs[SHADER_LOC_VERTEX_INSTANCE_TX] + i);
-        rlSetVertexAttribute(material.shader.locs[SHADER_LOC_VERTEX_INSTANCE_TX] + i, 4, RL_FLOAT, 0, sizeof(Matrix), i*sizeof(Vector4));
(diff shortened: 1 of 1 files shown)


raysan5:
@al13n321 thanks for the fix!

PR #5470, opened 2026-01-03
[rcore] Fix touch position automation event handling

ssszcmawo:
this PR fixes INPUT_TOUCH_POSITION automation event handling by properly using the touch id and position data. 
The event is now recorded only when the touch position changes.


raysan5:
@ssszcmawo thanks for the review!

PR #5471, opened 2026-01-04
[rcore] Allow opening window on selected surface

### LINKED ISSUE #4009: [rcore] Allow opening wayland layer surfaces with raylib
### Issue description

I'm developing a project called [woomer](https://github.com/coffeeispower/woomer) which is a clone of [tsoding's boomer](https://github.com/tsoding/boomer) but for wayland using raylib, it allows you to zoom into your desktop using your mouse wheel and move around by dragging.

However i ran into a bit of an issue in coffeeispower/woomer#5 , which is that as it says in the issue, if you open woomer while having a fullscreen application, it makes the other application exit fullscreen mode because woomer is just a fullscreen window opened with raylib.

The solution to this is using [wayland layer surfaces](https://wayland.app/protocols/wlr-layer-shell-unstable-v1) which are special windows that applications can open to make themselves part of the desktop like for example bars, menus, application launchers and etc, they're not tiled by tiling window managers, you can make them always appear on top of all applications and you can make them fullscreen without excluding normal windows.

I think we can probably use a config flag in SetConfigFlag, something like `SetConfigFlag(FLAG_OPEN_WAYLAND_LAYER)`

nouritsu:
Tested on NixOS 25.11 running Hyprland (commit: e5d20b56bcad78df42c9060a5d330274b0a6e510).
A demo can be found [on discord](https://cdn.discordapp.com/attachments/426912531429457930/1457157865671364628/demo.mp4?ex=695afb83&is=6959aa03&hm=9721112135c7c837612eb46b750a68ce49652dfdc23e76a9962e6f11eef955c9&).

Drafting this pull request because this will be my first contribution to a non-rust project, it likely needs a lot of changes :)

fix #4009

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -554,11 +554,10 @@ typedef enum {
     FLAG_BORDERLESS_WINDOWED_MODE = 0x00008000, // Set to run program in borderless windowed mode
     FLAG_MSAA_4X_HINT       = 0x00000020,   // Set to try enabling MSAA 4X
     FLAG_INTERLACED_HINT    = 0x00010000,    // Set to try enabling interlaced video format (for V3D)
-
-    FLAG_WAYLAND_LAYER_BACKGROUND = 0x00020000,  // Layer 0
-    FLAG_WAYLAND_LAYER_BOTTOM     = 0x00040000,  // Layer 1
-    FLAG_WAYLAND_LAYER_TOP        = 0x00080000,  // Layer 2
-    FLAG_WAYLAND_LAYER_OVERLAY    = 0x00100000,  // Layer 3
+    FLAG_WAYLAND_LAYER_BACKGROUND = 0x00020000,  // Layer level 0: background (for wallpapers)
+    FLAG_WAYLAND_LAYER_BOTTOM     = 0x00040000,  // Layer level 1: bottom (for normal applications)
+    FLAG_WAYLAND_LAYER_TOP        = 0x00080000,  // Layer level 2: top (for panels or popups)
+    FLAG_WAYLAND_LAYER_OVERLAY    = 0x00100000,  // Layer level 3: overlay (for lock screens or on screen keybords)
 } ConfigFlags;
 
 // Trace log level


nouritsu:
An application of this change is demonstrated by [this demo](https://www.youtube.com/watch?v=7RNJsmSBj_4). Woomer is able to spawn itself on the "overlay" layer, allowing it to work as an overlay rather than a forced fullscreen window, this fixes a bug highlighted [here](https://github.com/coffeeispower/woomer/issues/5).

This change would enable developers to create wallpaper engines, application launchers, logout menus, lockscreens and more using only raylib (on Wayland, anyways)

It may be possible to extend the layering functionality to other compositors, for example
- x11 with `_NET_WM_STATE_ABOVE` and `_NET_WM_STATE_BELOW` (https://specifications.freedesktop.org/wm/latest/ar01s05.html#id-1.6.8)
- windows with `SetWindowPos`'s `hWndInsertAfter` (https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowpos)
- macOS with `NSWindow.Level` (https://developer.apple.com/documentation/appkit/nswindow/level-swift.struct)

Note that these are not consistent with Wayland or with each-other.

I think it is possible to design non platform specific flags which rely on common functionality, like (names could be better)
- `FLAG_WINDOW_Z_INF` which will render the window above every other window 
- `FLAG_WINDOW_Z_ZERO` which will render the window below every other window

But then we end up losing *some* functionality for Wayland since there is no way to spawn a window on `Layer level 1 (bottom)`, so raylib applications like status bars may not be on the conventionally correct layer.

raysan5:
@nouritsu I'm afriad this change involves lots of changes on GLFW side and should be implemented in GLFW project to avoid overriding it when a new release is issued (if that ever happens...). Also note that `raylib.h` does not expose platform-specific flags, those flags should be defined in some other way (maybe in `config.h`).

Also note that now raylib on Wayland has several issues related to windowing system. A future project would be adding a custom `rcore_desktop_wayland` platform layer to remove GLFW dependency and have more control over this kind of changes but that represents a considerable amount of work...

I'm closing this PR but it can be a reference for anyone with those specific requirements willing to implement it on their forks. Thanks for the reference implementation!

nouritsu:
@raysan5 That's fair enough. I'm not sure if I currently have the skills or understanding to be able to make the changes you highlighted. However, I'll keep learning and if the opportunity ever presents, contribute to raylib!

PR #5473, opened 2026-01-05
[rcore] implemented saving for INPUT_GAMEPAD_CONNECT and INPUT_GAMEPAD_DISCONNECT

ssszcmawo:
Here is the implementation for saving gamepad connection and disconnection events. I added a helper variable, `prevReady`, to the `Gamepad` structure to keep track of its previous ready state. Now it is initialized along with `ready` in the `PlayAutomationEvent` function during INPUT_GAMEPAD_CONNECT. `prevReady` doesn't need to be set to `false` or `true` in `INPUT_GAMEPAD_DISCONNECT` because when disconnection happens it will be automatically updated at the end of the loop (line 4167)

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -362,6 +362,7 @@ typedef struct CoreData {
             int lastButtonPressed;          // Register last gamepad button pressed
             int axisCount[MAX_GAMEPADS];    // Register number of available gamepad axes
             bool ready[MAX_GAMEPADS];       // Flag to know if gamepad is ready
+            bool prevReady[MAX_GAMEPADS];   // Flag to know previous ready state
             char name[MAX_GAMEPADS][MAX_GAMEPAD_NAME_LENGTH];               // Gamepad name holder
             char currentButtonState[MAX_GAMEPADS][MAX_GAMEPAD_BUTTONS];     // Current gamepad buttons state
             char previousButtonState[MAX_GAMEPADS][MAX_GAMEPAD_BUTTONS];    // Previous gamepad buttons state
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -362,6 +362,7 @@ typedef struct CoreData {
             int lastButtonPressed;          // Register last gamepad button pressed
             int axisCount[MAX_GAMEPADS];    // Register number of available gamepad axes
             bool ready[MAX_GAMEPADS];       // Flag to know if gamepad is ready
+            bool prevReady[MAX_GAMEPADS];   // Flag to know previous ready state
```
raysan5, line 365: I'm afraid this variable is not desired because it implies an extra variable that must be managed in all supported platforms and some of them can not require it. Did you check how `Gamepad.ready` is filled in every platform? Maybe it can just be updated in the PollEvents() loop so `Gamepad.ready` always reflects current state and there is no need to check between current and previous.

EDIT: I missunderstood the issue, you want to register the change event, not the current state. Still, what is the use case for this feature?

ssszcmawo: You mean there’s no need to check the previous state? I wanted to ask yesterday on Discord what’s the best way to handle this, since these are my first commits and I might not fully understand how certain things work

raysan5: @ssszcmawo If you want to register the change event, then yes, you need to register the current and previous gamepad state but it implies adding that extra variable only for what it seems a very specific use case that probably most users won't need.

ssszcmawo: maybe i misunderstood the point and made it too complicated, i could rewrite so that it will only check the `ready` state

> I'm afraid this variable is not desired because it implies an extra variable that must be managed in all supported platforms and some of them can not require it. Did you check how `Gamepad.ready` is filled in every platform? Maybe it can just be updated in the PollEvents() loop so `Gamepad.ready` always reflects current state and there is no need to check between current and previous.
> 
> EDIT: I missunderstood the issue, you want to register the change event, not the current state. Still, what is the use case for this feature?

maybe i misunderstood the point and made it too complicated, i could rewrite so that it will only check `ready` state ,without `prevReady`

ssszcmawo: > @ssszcmawo If you want to register the change event, then yes, you need to register the current and previous gamepad state but it implies adding that extra variable only for what it seems a very specific use case that probably most users won't need.

when i opened this piece of code, in the commented part there was a check of the current  state and previous state, and I thought I needed to implement that previous state because it wasn’t in the structure, my bad if i misunderstood

raysan5: @ssszcmawo To register the event, you are right, previous and current states are required. But it adds an extra variable that maybe can be avoided just not implementing this feature. Why do you need it? Are you using it in some project?

ssszcmawo: > @ssszcmawo Why do you need it? Are you using it in some project?                                                                                                  

No i don't actually

ssszcmawo: so,can i just rewrite it checking only `ready` state and delete this extra variable?

raysan5: > No i don't actually

Then I prefer to adding an extra variable to all platforms to add a feature that is not required.

 > so.can i just rewrite it checking only ready state and delete this extra variable?

No, actually your implementation was correct. It will be a reference for users checking for it (if ever needed).

ssszcmawo: > No, actually your implementation was correct. It will be a reference for users checking for it (if ever needed).

but the problem still remains unsolved, i'd like to rework it in appropriate way, without `prevReady` and create another PR, if needed

raysan5: Most users won't need it ever. Users only need to know if gamepad is ready. No worries.

PR #5475, opened 2026-01-06
[rModel] Don't require a M3d animation only file to have a mesh. 

JeffM2501:
The current M3d anim file loader checks to ensure that the file has at least one skinned mesh. There are valid use cases for animation only files that can be applied to N other meshes, and this check prevents that from happening.

This PR removes the skin check from animation loading so that users can load a file that has animations only, and then apply it to many different models (this is common in games with different player models, but the same animations for all).

Tested with LoadM3d example and it works fine. This should not change any existing behavors.


raysan5:

PR #5476, opened 2026-01-06
[rcore] Fixed ExportAutomationEventList

ssszcmawo:
added saving to memory buffer and `SaveFileData` for binary files


raysan5:
@ssszcmawo thanks for the improvement

PR #5477, opened 2026-01-06
[rcore_android] Fix Android build 

maiconpintoabreu:
Adding missing fields for egl window system.

Before fix:

image

After fix:

image


raysan5:
@maiconpintoabreu Actually, `gbmDevice` is only for Linux DRM, so that change was wrong on Android. I prefer to just revert failing code instead of adding more compleexity.

PR #5478, opened 2026-01-07
examples/models: optimize collision detection in first person maze example

pauldahacker:
Replaces the full cubicmap collision scan with a check limited to the player’s surrounding cells, as suggested by the existing TODO comment.

This improves performance while preserving the example’s behavior.


raysan5:
@pauldahacker thanks for the improvement!

PR #5481, opened 2026-01-08
[raudio] Fix freeing the wrong memory

MarcosTypeAP:


raysan5:
@MarcosTypeAP Oh! Good catch! Thanks for the fix!

MarcosTypeAP:
@raysan5 Thank you for your amazing work!

PR #5483, opened 2026-01-09
fix(examples): tell user how to run example

vegerot:


raysan5:
@vegerot User should be aware, no intention to add this kind of comments

PR #5484, opened 2026-01-10
fix: android now can read actual path by adding fread fallback

lucas150670:
on android, formerly raylib can only access internal file by
```cpp
return fopen(TextFormat("%s/%s", internalDataPath, fileName), mode);
```

add a fallback when this failed it will try to read file directly from the path.
```cpp
if(fopen(TextFormat("%s/%s", internalDataPath, fileName), mode) == NULL) {
      return fopen(fileName, mode);
}
```
suitable for app guaranteed with these premissions:
```xml
    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
    <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE"/>
```
and try to directly read & write external storage.


raysan5:
@lucas150670 thanks for the review

PR #5486, opened 2026-01-11
fix rglfw.c path in build.zig for bsd platforms

jamesmintram:


raysan5:
@jamesmintram thanks for the review!

PR #5487, opened 2026-01-11
Update build.zig to work with 0.16

mdm5995:
Just a few small changes to update the build.zig to work with the latest breaking changes for zig v16

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -522,12 +522,13 @@ fn addExamples(
     raylib: *std.Build.Step.Compile,
 ) !*std.Build.Step {
     const all = b.step(module, "All " ++ module ++ " examples");
+    const io = all.owner.graph.io;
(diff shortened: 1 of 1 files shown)


raysan5:
@mdm5995 thanks for the update... but would it be fine for any Zig user?

ToastyGoblin:
Well, I can only speak for myself, but the new update, as far as I know, made it incompatible with the latest stable release for Zig 0.15.2; it wouldn’t build until I changed to the beta dev build 1.16. So it would be a good idea to update build.zig `const min_ver = 0.15.1` to `const min_ver = 0.16` inside the build.zig script to let Zig users know they have to change from the latest stable release build to the beta dev build.


> @mdm5995 thanks for the update... but would it be fine for any Zig user?

PR #5490, opened 2026-01-12
fix: zig build accessing env var

### LINKED ISSUE #5489: [build] Build fails with zig 0.16.0-dev.2145+ec25b1384

 - [ ] ~~My code has no errors or misuse of raylib~~ N/A

### Issue description

Running `zig build` on master results in the following error:

```
build.zig:451:113: error: root source file struct 'process' has no member named 'getEnvVarOwned'
            .android_ndk = b.option([]const u8, "android_ndk", "specify path to android ndk") orelse std.process.getEnvVarOwned(b.allocator, "ANDROID_NDK_HOME") catch "",
                                                                                                     ~~~~~~~~~~~^~~~~~~~~~~~~~~
/Users/iisakki/zig/zig-aarch64-macos-0.16.0-dev.2145+ec25b1384/lib/std/process.zig:1:1: note: struct declared here
const builtin = @import("builtin");
^~~~~
referenced by:
    build: build.zig:497:74
    runBuild__anon_27490: /Users/iisakki/zig/zig-aarch64-macos-0.16.0-dev.2145+ec25b1384/lib/std/Build.zig:2259:44
    5 reference(s) hidden; use '-freference-trace=7' to see all references
```

As far as I can see, this is due to this change in zig: https://codeberg.org/ziglang/zig/pulls/30644

As stated in the PR,

> The new argument to pub fn main may be one of three things:
> 
> - missing. empty main parameter list is still legal, however it now means you can't access CLI arguments or environment variables.
> - ...

I'm not experienced enough with Zig to know if there is another way of accessing env vars, but it sounds to me like there isn't.

iisakkirotko:
Fixes #5489.

Accessing env vars in this fashion was removed in https://codeberg.org/ziglang/zig/pulls/30644.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -448,7 +448,7 @@ pub const Options = struct {
             .linux_display_backend = b.option(LinuxDisplayBackend, "linux_display_backend", "Linux display backend to use") orelse defaults.linux_display_backend,
             .opengl_version = b.option(OpenglVersion, "opengl_version", "OpenGL version to use") orelse defaults.opengl_version,
             .config = b.option([]const u8, "config", "Compile with custom define macros overriding config.h") orelse &.{},
-            .android_ndk = b.option([]const u8, "android_ndk", "specify path to android ndk") orelse std.process.getEnvVarOwned(b.allocator, "ANDROID_NDK_HOME") catch "",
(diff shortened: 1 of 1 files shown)


raysan5:
@iisakkirotko thanks for the review

PR #5492, opened 2026-01-13
make glfw use absolute paths so no include path has to be supplied

keks137:
One thing I really like about raylib is how stupidly easy it is to compile from source, even without a build system.
But the thing that was always kind of annyoing about it is that you have to supply the include directory for GLFW.
I'm lazy, so I decided to make them absolute, which is a bit uglier, but it's only used in 5 places anyways.
This way compiling becomes even simpler (linux example)
```bash
cc  src/*.c -DPLATFORM_DESKTOP  -D_GLFW_X11
```
is all you need

I know this isn't that important, but it makes me happy.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -48,7 +48,7 @@
 
 #define GLFW_INCLUDE_NONE       // Disable the standard OpenGL header inclusion on GLFW3
                                 // NOTE: Already provided by rlgl implementation (on glad.h)
-#include "GLFW/glfw3.h"         // GLFW3 library: Windows, OpenGL context and Input management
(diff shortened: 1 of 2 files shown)


raysan5:
@keks137 This is not desired, code should be as much decoupled as possible from build system.

jgabaut:
I don't think this is doable since you lose the ability to easily switch the glwf include.
Any user that has an include substitution would need to change its current setup.

keks137:
> I don't think this is doable since you lose the ability to easily switch the glwf include. Any user that has an include substitution would need to change its current setup.

didn't know that's a thing people do
I blindly assumed everyone just uses the glfw that comes with raylib, but yea, you could just swap it with another one, which in the case of glfw you might want to do, if you link it dynamically

didn't think of that since I never do anything like that
that's what maintainers are for I guess

PR #5493, opened 2026-01-14
removed unnecessary memset calls; also cleaned up DrawText

mikedesu:
I decided to take my first stab at contributing to raylib's development today.

While using `aider (deepseek)` to evaluate `rtext.c`, I discovered several unnecessary calls to `memset`:

In C, initializing a buffer with `{0}`, when compiling with `gcc`, sets every cell in the buffer to `0`, making the calls to `memset` redundant. 

I have not tested this on Windows, but the code builds successfully and integrates with my game.

In addition, I did some light cleanup to `DrawText`:

1. I introduced an early exit during the check if the default font is loaded to reduce if-statement nesting.
2. I made several variables `const`.

-----

This commit isn't too special but I want to find small performance and style enhancements to the codebase like these to begin with.

Thank you for making raylib! I hope these changes are seen as improvements. I was also careful to not modify the formatting of the file itself.

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1208,16 +1208,14 @@ void DrawFPS(int posX, int posY)
 void DrawText(const char *text, int posX, int posY, int fontSize, Color color)
 {
     // Check if default font has been loaded
-    if (GetFontDefault().texture.id != 0)
-    {
-        Vector2 position = { (float)posX, (float)posY };
-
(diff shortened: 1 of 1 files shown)


raysan5:
@mikedesu Those buffers are static so they are reeused on every function call, so, they need to be cleared on each call.

mikedesu:
Oh you're right...never mind, my bad!

PR #5496, opened 2026-01-15
[rcore] `LoadDirectoryFilesEx()`, count files if not recursive

katanya04:
The `LoadDirectoryFilesEx` contained a comment saying: `// NOTE: On recursive loading we do not pre-scan for file count, we use MAX_FILEPATH_CAPACITY`. But if you take a look at the original function, you will see that it uses `MAX_FILEPATH_CAPACITY` on both cases.

I added the (I suppose?) expected behaviour: count files if not loading recursively, and also taking into account the filter passed to the function. I tested it on Linux and it works, the file count is the same but the capacity is equal to the file count instead of `MAX_FILEPATH_CAPACITY`, although further testing would be nice I guess, you know on Windows for example.

If it doesn't follow the raylib coding conventions, please do tell me and I'll fix it.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -511,7 +511,7 @@ extern void ClosePlatform(void);        // Close platform
 static void InitTimer(void);                                // Initialize timer, hi-resolution if available (required by InitPlatform())
 static void SetupViewport(int width, int height);           // Set viewport for a provided width and height
 
-static void ScanDirectoryFiles(const char *basePath, FilePathList *list, const char *filter, unsigned int fileCount, bool scanSubdirs); // Scan all files and directories in a base path
+static void ScanDirectoryFiles(const char *basePath, FilePathList *list, const char *filter, unsigned int expectedFileCount, bool scanSubdirs); // Scan all files and directories in a base path
 
 #if defined(SUPPORT_AUTOMATION_EVENTS)
 static void RecordAutomationEvent(void); // Record frame events (to internal events array)
@@ -4270,7 +4270,7 @@ void SetupViewport(int width, int height)
 // Scan all files and directories in a base path
 // WARNING: files.paths[] must be previously allocated and
 // contain enough space to store all required paths
-static void ScanDirectoryFiles(const char *basePath, FilePathList *files, const char *filter, unsigned int fileCount, bool scanSubdirs)
+static void ScanDirectoryFiles(const char *basePath, FilePathList *files, const char *filter, unsigned int expectedFileCount, bool scanSubdirs)
 {
     // WARNING: Path can not be static or it will be reused between recursive function calls!
     char path[MAX_FILEPATH_LENGTH] = { 0 };
@@ -4281,7 +4281,7 @@ static void ScanDirectoryFiles(const char *basePath, FilePathList *files, const 
 
     if (dir != NULL)
     {
-        while (((dp = readdir(dir)) != NULL) && (files->count < fileCount))
+        while (((dp = readdir(dir)) != NULL) && (files->count < expectedFileCount))
         {
             if ((strcmp(dp->d_name, ".") != 0) && (strcmp(dp->d_name, "..") != 0))
             {
@@ -4313,7 +4313,7 @@ static void ScanDirectoryFiles(const char *basePath, FilePathList *files, const 
                         files->count++;
                     }
 
-                    if (scanSubdirs) ScanDirectoryFiles(path, files, filter, fileCount, scanSubdirs);
+                    if (scanSubdirs) ScanDirectoryFiles(path, files, filter, expectedFileCount, scanSubdirs);
                 }
             }
         }


raysan5:
@katanya04 This is an interesting redesign, it "potentially" increases cost because a pre-scan is required for directories but it adds more flexibility and allows removing that `capacity` field in `FilePathList`. Could you verify if it's possible to remove that field?

EDIT: Also, we can consider adding `GetDirectoryFileCount()`/`GetDirectoryFileCountEx()`, despite, afaik, there is barely any API out there with that function.

katanya04:
Well sure at first glance I guess it could be removed since `capacity` is either equals to the number of files/directories in the path, or to `MAX_FILEPATH_CAPACITY`.

There's this one check that is done on `LoadDirectoryFiles`:
```c
// Security check: read files.count should match fileCounter
if (files.count != files.capacity) TRACELOG(LOG_WARNING, "FILEIO: Read files count do not match capacity allocated");
```
I can't think of a scenario where that would happen (maybe delete or add files into the directory to be scanned while the scan is being done... but yeah, that would be rare).

On `UnloadDirectoryFiles`, memory is deallocated on a for loop, from 0 to `files.capacity`. If it was from 0 to `files.count`, it would be the same, except if that previous error case happened where both fields didn't match, or on a more common scenario: if you used `LoadDirectoryFilesEx` with recursive scanning, because remember, it uses `MAX_FILEPATH_CAPACITY` as the `capacity` to allocate, so it needs to deallocate `MAX_FILEPATH_CAPACITY` once again, not `files.count`, a memory leak would happen otherwise. So, if you were to remove `capacity` from the `FilePathList` struct, you would need to also count the file number on recursive scanning.

`field.capacity` is also used on `ScanDirectoryFilesRecursively`, to check if it can continue adding paths to the path array. If the file count is correct, this should not be a big deal... You would have to use a temp variable to store the current file count to be able to compare it to the total file count, or skip the security comparison entirely and just suppose we have not reached the maximum allocation space calculated from our first file count (which on a normal case it should not happen).

So as a summary, yeah it can be removed, but files should also be counted on recursive scanning. I am unsure which way would be more performant, on the one hand you can choose a faster initial scan but allocate `MAX_FILEPATH_CAPACITY`, and on the other hand you only allocate the real number of files but do a slower initial scan. How much faster/slower it would be, I don't know. Still, I can update the pr to remove the field if you want. Probably using those functions you said on your edit to do the initial scan.

raysan5:
@katanya04 Feel free to update the PR and remove `capacity` field, I think it actually makes not sense to keep it considering the array is designed to be immutable after loading.

katanya04:
Well I think that should be it, I tested it for a bit and it works. A few notes:
- I added `GetDirectoryFileCount()` and `GetDirectoryFileCountEx()` as static functions in rcore.c. There's no real reason for these functions to be hidden from the programmer, but I also felt that they may be very specific to offer on `raylib.h`. But I don't know, change them to be public if you want.
- `LoadDirectoryFiles()` and `LoadDirectoryFilesEx()` are almost identical functions, I made different implementations for each, but you could just simply implement `LoadDirectoryFiles()` in 1 line as a call to `LoadDirectoryFilesEx()` with the `filter` set to `DIRECTORY_FILTER_TAG` and `scanSubdirs` to `false`. The same could be said about `GetDirectoryFileCount()` and `GetDirectoryFileCountEx()`. But since on the current implementation they were both implemented as different functions that don't call each other, I suppose you prefer it this way.
- File count is passed to the `ScanDirectoryFiles()` and `ScanDirectoryFilesEx()` functions, just to check on the while loop that it doesn't load more paths than allocated. Again, this is unlikely to happen, but I think it's better to make a safe function rather than an unsafe one even if the condition for a buffer overflow is never met.
- Just like there is an upper bounds check (that previous point), there's a lower bounds check: if the number of paths loaded is less than the files counted, a warning will get logged and `files.count` will be set to `fileCount`, since to unload the `FilePathList` struct we deallocate `files.count` times, if `files.count` was less than allocated (`fileCount`) a memory leak would happen.
- I found it quite interesting that on the current implementation of `LoadDirectoryFilesEx()`, if you pass `NULL` as the `filter` parameter, it will count all the files... but not the directories. I did not modify this behaviour, I'm just commenting that it feels a bit odd. I would have expected to get "everything" (files + directories) counted if I don't use a filter. But maybe that's just me.

raysan5:
@katanya04 thanks! this is a great improvement! Here some comments to review:

- I think `GetDirectoryFileCount()` and `GetDirectoryFileCountEx()` can be useful for programmers, better expose them. **Please, could you review it?**
- If the functions can be simplified calling the `*Ex()` versions, I prefer to do so in this case, for simplification and minimize FileSystem calls locality. **Please, could you review it?**
- For now it's ok to add an extra parameter to `ScanDirectoryFiles()` because it's an internal function but I always prioritize the minimum required number of parameters...
- Those bounds checks are nice
- I don't remember why but I probably need some way to accomplish that and it was the simpler/faster option... probably a `PATH_FILTER_FILE_ONLY` can be added for consistency with `DIRECTORY_FILTER_TAG`... maybe better names should be also considered... **Ideas?**

katanya04:
`GetDirectoryFileCount()` and `GetDirectoryFileCountEx()` have been made public. And as for rewritting the functions to just call the `*Ex()` versions, I just realized I need to pass a filter that would capture all the file types and directories, because the non `*Ex()` versions count every file and directory inside. If I pass `NULL` as the `filter`, it would count all files, but not the directories. If I pass `DIRECTORY_FILTER_TAG` as the `filter`, it would now count the directories... but not the files. As I said, I think it would make sense that if you pass `NULL` as the `filter`, it would count everything. And there should probably be a filter for all file types, like `*` for example. So if you want to just count directories, you pass `DIRECTORY_FILTER_TAG`, if you want to count only files you pass a constant for example `FILE_FILTER_TAG` whose value could be `*`, and if you want to count everything you pass `NULL`. But that's just an idea.

raysan5:
@katanya04 thanks for the review! Definitely the filter tags should be reviewed.

 > As I said, I think it would make sense that if you pass `NULL` as the filter, it would count everything. 

This would be a breaking change and I'm afraid there could be some code bases already relying on that behaviour, including some of my tools. But, `NULL` neither seems much descriptive for an expected behaviour, I think new tags can be defined to make it more explicit:
```c
#define FILE_FILTER_TAG_ALL          "*.*"
#define FILE_FILTER_TAG_FILE_ONLY    "FILES*"
#define FILE_FILTER_TAG_DIR_ONLY     "DIR*"
```
And let `NULL` to keep current behaviour of (`FILE_FILTER_TAG_FILE_ONLY`).

Also, I got my doubts about the extra parameter `fileCount` on `ScanDirectoryFiles*()`, it does not seem intuitive to me looking at it, is there any specific use case where it could fail? It seems the chances are really low. Maybe that parameter can be avoided and just do a check on the `LoadDirectoryFiles*()` against `GetDirectoryFiles*()` result?

katanya04:
> Also, I got my doubts about the extra parameter fileCount on ScanDirectoryFiles*(), it does not seem intuitive to me looking at it, is there any specific use case where it could fail? It seems the chances are really low. Maybe that parameter can be avoided and just do a check on the LoadDirectoryFiles*() against GetDirectoryFiles*() result?

Definitely, it's a low chance, but a buffer overflow could occur if the number of files in the directory increases after the number of files has been count (so between scan #1 and scan #2), it's a race condition ([more specifically this type of race condition](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use)). Without that parameter it's like the unsafe standard libc functions that don't check for the amount of bytes to copy/paste/manipulate (like `sprintf()`) vs the safe ones that you have to pass the max amount of bytes to manipulate (like `snprintf()`). Maybe the name of the parameter can be changed to something more intuitive, and/or document it with a comment. Or I can remove it. The choice is yours.

And as for the tags, I'll look into implementing them now.

katanya04:
> #define FILE_FILTER_TAG_DIR_ONLY     "DIR*"

Would this tag behave differently from `DIRECTORY_FILTER_TAG`? Would it replace that one or should both coexist?

raysan5:
> Would this tag behave differently from `DIRECTORY_FILTER_TAG`? Would it replace that one or should both coexist?

`FILE_FILTER_TAG_DIR_ONLY` would be a replacement, more consistent with the new ones.

About the `fileCount` parameter, we can keep it for now, considering the scan functions are internal. But it can be renamed to `expectedFileCount` for more clarity.

katanya04:
I have also noticed, that `ScanDirectoryFiles()` and `ScanDirectoryFilesEx()` will have practically the same body (the only difference is the recursive call in `ScanDirectoryFilesEx()`, literally 1 line of code). They could be simplified into just 1 function, where you pass as the 5th argument a bool to indicate whether to do recursive scanning or not. It feels natural to do so. The only problem is that if 4 arguments alerady felt like a lot, it would be 5 now... Yeah. You tell me.

raysan5:
@katanya04 Ok, let's add that 5th parameter to reduce the functions count.

katanya04:
Should `IsFileExtension()` check if the extension passed as parameter is `FILE_FILTER_TAG_ALL` or `FILE_FILTER_TAG_FILE_ONLY` and return true in those cases, or should these tags only be used on the `ScanDirectoryFiles()` function? It felt logic to add that behaviour to `IsFileExtension()`, but as of now, `IsFileExtension()` works even if the file passed doesn't exist in the file structure of your system, it just does some string comparisons on the passed `fileName`, but if I were to add this check against those tags, I should first test if it is a file and not a directory... A file without extension on its name should still return true if the extension passed is one of those tags. So yeah, I'm not sure whether to add that behavior to the function or not.

katanya04:
Everything should be done (except what I just commented on my last comment, `IsFileExtension()`). I tested it before the changes, after the changes, with filters and no filters, and the file counts were the same, although a bit more of testing could be nice (once again, in Windows for example).
There's a small detail... if you call `LoadDirectoryFilesEx()` with the `scanSubdirs` bool set to `true` on a path with protected directories, it will throw a warning for each protected directory saying that it cannot be opened. That's normal. But it will log that warning twice, first during the count files scan, and then on the load paths scan, both warnings being exactly the same string (`WARNING: FILEIO: Directory cannot be opened ([the path here]`). That's a bit... well, "not ideal" I guess. I don't know if you feel like this should be fixed or not.

raysan5:
@katanya04 thanks for the further review! It looks great and it simplifies internal code! Thanks!

About `IsFileExtension()`, I think we can keep as it is for now, in any case, that improvement could be added in the future.

Merging!

PR #5499, opened 2026-01-15
[build] Add QNX EGL2.0 Library configuration

jscaff:
Adds library configuration for the QNX developer desktop which only has OpenGL ES2.0 support.

The configuration is based off of the DRM configuration.


raysan5:
@jscaff Oh! Did you try raylib on QNX? How does it work???

jscaff:
> @jscaff Oh! Did you try raylib on QNX? How does it work???

The few examples I tried out worked great.

raysan5:
@jscaff Just out of curiosity, did raylib compile and run on QNX on DRM backend with no changes or did it require some custom backend for that platform? 🤔

jscaff:
It didn't require that many changes, just setting a few compiler flags to reflect what is and isn't offered on the platform. 
I took inspiration from the DRM section since it was the only platform that also targets OpenGL ES2.0, however we are compiling it for a general purpose desktop platform.

If you're curious, you can find our APKBUILD file here: [https://github.com/qnx-ports/aports/blob/803/extra/raylib/APKBUILD](https://github.com/qnx-ports/aports/blob/803/extra/raylib/APKBUILD) (do note, it has to run on a QNX system to compile)

PR #5500, opened 2026-01-18
[rlgl] Fix condition for WebGL depth texture support

SlateyDev:
Condition looks to be incorrect. If texDepthWebGL is available, it should be using GL_DEPTH_COMPONENT16/24/32 but currently in WebGL always uses GL_DEPTH_COMPONENT which can fail especially in WebGL2.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -3429,7 +3429,11 @@ unsigned int rlLoadTextureDepth(int width, int height, bool useRenderBuffer)
     {
         if (RLGL.ExtSupported.maxDepthBits == 32) glInternalFormat = GL_DEPTH_COMPONENT32_OES;
         else if (RLGL.ExtSupported.maxDepthBits == 24) glInternalFormat = GL_DEPTH_COMPONENT24_OES;
+#if !defined(GRAPHICS_API_OPENGL_ES3)
+        else if (useRenderBuffer) glInternalFormat = GL_DEPTH_COMPONENT16;
+#else
         else glInternalFormat = GL_DEPTH_COMPONENT16;
+#endif
     }
 #endif
 


SlateyDev:
This does not work for ES2 as WEBGL_depth_texture extension may exist, but it does not support the formats GL_DEPTH_COMPONENT16/24/32. These formats are supported for renderbuffer but not for texImage2D in ES2 and you need to use GL_DEPTH_COMPONENT for a 16-bit depth buffer.

raysan5:
@SlateyDev I'm reviewing this issue but I think I'm not understanding it... As per [WebGL 1.0 specifications](https://registry.khronos.org/webgl/specs/latest/1.0/#5.14), `glTexImage2D()` only supports `GL_DEPTH_COMPONENT` for internal pixel format, so, the sized pixel formats are only assigned if `!RLGL.ExtSupported.texDepthWebGL`.

Checking the official specs for [`WebGL_depth_texture` extension](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_depth_texture), no sized constant is added, actually it lists `GL_DEPTH_COMPONENT` as the required internal format. Also, it states:

> Note: This extension is only available to [WebGL1](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext) contexts. In [WebGL2](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext), the functionality of this extension is available on the WebGL2 context by default. The constant in WebGL2 is gl.UNSIGNED_INT_24_8.

Checking  WebGL 2 specs, the sized components are supported but the provided information [is missleading in reference to WebGL 1](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/texImage2D).

image

So, I'm confused, probably the code should be reviewed to consider `OpenGL ES 2.0 (WebGL 1.0)` and `OpenGL ES 3.0 (WebGL 2.0)` separately. Did you test this functionality on the different versions?  `OpenGL ES 2.0`, `WebGL 1.0` and `WebGL 2.0`.

SlateyDev:
The problem is when WebGL2 is in use, RLGL.ExtSupported.texDepthWebGL is true and so it does not use sized pixel formats and fails because it does not work with GL_DEPTH_COMPONENT as it did in WebGL1.

I haven't yet been able to test under all cases to come up with a better design that works with all of ES2, WebGL1 and WebGL2 as I also haven't played with render buffers in any of my testing yet.

raysan5:
> The problem is when WebGL2 is in use, RLGL.ExtSupported.texDepthWebGL is true and so it does not use sized pixel formats and fails because it does not work with GL_DEPTH_COMPONENT as it did in WebGL1.

Ok, now I see the issue.

 > I haven't yet been able to test under all cases to come up with a better design that works with all of ES2, WebGL1 and WebGL2 as I also haven't played with render buffers in any of my testing yet.

Ok, I will wait for those tests, feel free to redesign that approach if required.

raysan5:
It should be fixed with my commit

PR #5502, opened 2026-01-18
Added DrawMeshInstancedColumnMajor

KenthJohan:
Addition of `DrawMeshInstancedColumnMajor`, which allows passing transformation matrices as column-major, and updating `DrawMeshInstanced` to use this new function internally.

Keep in mind that 
```c
void DrawMeshInstancedColumnMajor(Mesh mesh, Material material, const void * transforms, int instances)
```
uses a void pointer, which avoids misleading users into thinking that a Matrix type row order is directly compatible with DrawMeshInstancedColumnMajor.


raysan5:
@KenthJohan Thanks but I'm afraid I'm not adding a new function for this use-case

KenthJohan:
There are some math libraries like BLAS or ATLAS that uses column major. It just gives the user the option to choose. But I understand that more functions may complicate the API unnecessary. Thanks for feedback.

PR #5505, opened 2026-01-20
[rlgl] Clean up Matrix handling and label implicit transposes in rlgl.h

noinodev:
raylib convention uses m0..m15 named fields for Matrix operations: 
- Use named fields in rlTranslatef, rlScalef and rlMatrixIdentity instead of array literals 

Additionally, the memory order of struct Matrix is row-major, therefore some functions implicitly convert to column-major and vice-versa: 
- Label implicit row-major to column-major transpose in MatrixToFloatV 
- Label implicit column-major to row-major transpose in rlMultMatrixf 
- Update rlSetUniformMatrix to use rlMatrixToFloat convention

Aimed for readability and maintainability, does not change behaviour.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -1270,12 +1270,12 @@ void rlLoadIdentity(void)
 // Multiply the current matrix by a translation matrix
 void rlTranslatef(float x, float y, float z)
 {
-    Matrix matTranslation = {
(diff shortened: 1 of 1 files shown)


raysan5:
@noinodev thanks for the review!

PR #5507, opened 2026-01-23
[web] Fix Emscripten's Closure compiler error: undeclared canvas variable

vdemcak:
The original implementation relied on a global `canvas` variable, which is typically defined in Emscripten's default shell. When using Emscripten's Closure compiler, this file is not visible to the compiler, causing minification to fail with
`ERROR - [JSC_UNDEFINED_VARIABLE] variable canvas is undeclared`.

This pull request resolves the issue by no longer relying on the implicit global and instead explicitly querying the document for the canvas element by its ID.

One potential (currently unverified) side effect is that projects using a custom Emscripten shell without a canvas element tagged with `id="canvas"` may need to update their HTML accordingly. However, the default Emscripten shell already queries the canvas using `id="canvas"`, so existing working setups should remain unaffected.

### DIFF
--- a/src/platforms/rcore_web.c
+++ b/src/platforms/rcore_web.c
@@ -316,15 +316,16 @@ void ToggleBorderlessWindowed(void)
         // 2. The style unset handles the possibility of a width="value%" like on the default shell.html file
         EM_ASM
         (
+            const canvasId = UTF8ToString($0);
             setTimeout(function()
             {
                 Module.requestFullscreen(false, true);
                 setTimeout(function()
                 {
-                    document.getElementById("canvas").style.width="unset";
+                    document.querySelector(canvasId).style.width="unset";
                 }, 100);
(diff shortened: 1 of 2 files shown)


raysan5:
@vdemcak Note that there could be multiple canvas running different raylib instances on a single webpage. Recently it was implemented `SetCanvasIdJs()` that gets current canvas id registered on `platform.canvasId`. Maybe it should use that variable instead?

vdemcak:
Thanks for the heads up. I've implemented the changes, though they may need a quick glance-over since I'm not deeply familiar with the codebase yet.

I also noticed a hardcoded canvas ID outside the scope of this PR and fixed that while I was there (`src/platforms/rcore_web.c`, line `1240`)

raysan5:
@vdemcak Looks good! Thanks for the review!

PR #5511, opened 2026-01-24
[rtext] Fix and enhance `TextReplace()` function 

ssszcmawo:
add check for replacement,replace strcpy,strncpy with memcpy

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1734,7 +1734,7 @@ char *TextReplace(const char *text, const char *search, const char *replacement)
 
     if ((text != NULL) && (search != NULL))
     {   
-        if(replacement == NULL) replacement = "";
+        if (replacement == NULL) replacement = "";
 
         char *insertPoint = NULL;   // Next insert point
         char *temp = NULL;          // Temp pointer
@@ -1769,12 +1769,12 @@ char *TextReplace(const char *text, const char *search, const char *replacement)
             insertPoint = (char *)strstr(text, search);
             lastReplacePos = (int)(insertPoint - text);
             
-            memcpy(temp,text,lastReplacePos);
+            memcpy(temp, text, lastReplacePos);
             temp += lastReplacePos;
             
-            if(replaceLen > 0)
+            if (replaceLen > 0)
             {
-                memcpy(temp,replacement,replaceLen);
+                memcpy(temp, replacement, replaceLen);
                 temp += replaceLen; 
             }
 


raysan5:
@ssszcmawo Please, could you review code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

ssszcmawo:
> @ssszcmawo Please, could you review code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

added 4 spaces in `if` statement, if this wasn’t the only problem, could you please specify what exactly is still wrong?

src/rtext.c
```diff
@@ -1727,14 +1727,15 @@ char *GetTextBetween(const char *text, const char *begin, const char *end)
...
-// TODO: If (replacement == "") remove "search" text
 // WARNING: Allocated memory must be manually freed
 char *TextReplace(const char *text, const char *search, const char *replacement)
 {
     char *result = NULL;
 
     if ((text != NULL) && (search != NULL))
-    {
+    {   
+        if(replacement == NULL) replacement = "";
```
raysan5: space after `if`

src/rtext.c
```diff
@@ -1767,16 +1768,15 @@ char *TextReplace(const char *text, const char *search, const char *replacement)
...
-            // OK - Those lines work
-            temp = strncpy(temp, text, lastReplacePos) + lastReplacePos;
-            temp = strcpy(temp, replacement) + replaceLen;
-            // WRONG - But not those ones
-            //temp = strncpy(temp, text, tempLen - 1) + lastReplacePos;
-            //tempLen -= lastReplacePos;
-            //temp = strncpy(temp, replacement, tempLen - 1) + replaceLen;
-            //tempLen -= replaceLen;
+            
+            memcpy(temp,text,lastReplacePos);
```
raysan5: space after `,`

src/rtext.c
```diff
@@ -1767,16 +1768,15 @@ char *TextReplace(const char *text, const char *search, const char *replacement)
...
+            temp += lastReplacePos;
+            
+            if(replaceLen > 0)
```
raysan5: space after `if`

src/rtext.c
```diff
@@ -1767,16 +1768,15 @@ char *TextReplace(const char *text, const char *search, const char *replacement)
...
+            {
+                memcpy(temp,replacement,replaceLen);
```
raysan5: space after `,`

ssszcmawo: > space after `,`

done

src/rtext.c
```diff
@@ -1727,14 +1727,15 @@ char *GetTextBetween(const char *text, const char *begin, const char *end)
...
-// TODO: If (replacement == "") remove "search" text
 // WARNING: Allocated memory must be manually freed
 char *TextReplace(const char *text, const char *search, const char *replacement)
 {
     char *result = NULL;
 
     if ((text != NULL) && (search != NULL))
-    {
+    {   
+        if (replacement == NULL) replacement = "";
```
raysan5, line 1737: Is this correct? if `replacement == NULL`, there is no space allocated to store anything on it...

Did you test this code works as expected?

ssszcmawo: > Is this correct? if `replacement == NULL`, there is no space allocated to store anything on it...
> 
> Did you test this code works as expected?

Yes, that’s fine. Assigning  ""  when `replacement` is `NULL` is safe, it points to an empty string, so no extra allocation is needed and string functions will work as expected.

raysan5: As per my understading, this is not correct, a NULL pointer has no space allocated while an empty string actually contains '\0' character. I understand it could work but I don't think it's correct that assumption. Is it documented somewheree? Trying to find some official docs...

EDIT: I could be wrong, I never seen that approach before, just using my programmer logic to try to understand it...

ssszcmawo: > As per my understading, this is not correct, a NULL pointer has no space allocated while an empty string actually contains '\0' character. I understand it could work but I don't think it's correct that assumption. Is it documented somewheree? Trying to find some official docs...
> 
> EDIT: I could be wrong, I never seen that approach before, just using my programmer logic to try to understand it...

assigning `replacement` = "" when it’s `NULL` makes it point to a valid empty string, so string operations like `strlen` and `strcpy` will work safely. It also allows the function to remove the target substring when no replacement is provided.

I’ve also just tested it, and it works as intended.

vdemcak: Hey, just wanted to chime in. @ssszcmawo's logic should be correct.

If `NULL` gets passed in for the replacement, the `if (replacement == NULL) replacement = "";` case correctly redirects the pointer from `NULL` towards a static string literal `""`. This should provide a valid memory address containing a null terminator, so subsequent reads won't crash. The assignment would only be an issue if the pointer was dereferenced and you were assigning a value to it (which you don't).

raysan5: @ssszcmawo @vdemcak thanks for the further explanation, I understand it now.

raysan5:
@ssszcmawo thanks for the review

PR #5512, opened 2026-01-26
[CI] Removing double zip and misleading zip type

maiconpintoabreu:
At the moment the Artifact is adding a zip type file inside another zip file.

Example:

image

Here you can see that all files are zip, even if you name it tar.gz and inside this zip you have the tar.gz or zip file based on the platform.

Upload-artifact limitation is `When an Artifact is uploaded, all the files are assembled into an immutable Zip archive. There is currently no way to download artifacts in a format other than a Zip or to download individual artifact contents.`

https://github.com/actions/upload-artifact/tree/v4/?tab=readme-ov-file#zip-archives

Solution:

```
        name: ${{ env.RELEASE_NAME }}
        path: |
          ./build/${{ env.RELEASE_NAME }}
          !./build/${{ env.RELEASE_NAME }}.tar.gz <- ignore the manually created zip but keeping it for release phase
```

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -138,7 +138,7 @@ jobs:
         7z a ./${{ env.RELEASE_NAME }}.zip ./${{ env.RELEASE_NAME }}
         dir
       shell: cmd
-        
+
     - name: Upload Artifacts
       uses: actions/upload-artifact@v4
       with:


raysan5:
@maiconpintoabreu Nice improvement! I didn't know it was possible to do that! Thanks!

PR #5515, opened 2026-01-27
fixed typos preventing launch of native win32 backend

mikeemm:
The native win32 backend currently does not work: trying to run a simple example fails due to a few typos during the window creation process. This PR fixes these so the backend can run if chosen instead of GLFW.

### DIFF
--- a/src/platforms/rcore_desktop_win32.c
+++ b/src/platforms/rcore_desktop_win32.c
@@ -1227,7 +1227,7 @@ void SwapScreenBuffer(void)
 // Get elapsed time measure in seconds
 double GetTime(void)
 {
-    LARGE_INTEGER now = 0;
+    LARGE_INTEGER now = { 0 };
     QueryPerformanceCounter(&now);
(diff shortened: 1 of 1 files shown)


raysan5:
@mikeemm thanks for the review, the Win32 backend is not completely ready yet, some parts need more testing

mikeemm:
@raysan5 No problem, do you know anything else in particular from that backend that could use a review? I noticed that _UpdateFlags()_ has some issues too so I'll look into that next if that's OK.

raysan5:
@mikeemm this backend was contributed but I rewrote most of it for simplification and to support modern OpenGL as well as the new software renderer. Some of the parts that require further review:

 - Windowing flags, some could not work, some use a delay mechanism, I think it can be simplified
 - HighDPI/Content scaling, I remember commenting that part, it should be reenabled and tested
 - Inputs should mostly work but I think they need more testing, recently [created this example](https://github.com/raysan5/raylib/blob/master/examples/core/core_keyboard_testbed.c) for that pourpose

PR #5517, opened 2026-01-28
[refactor] Refactoring based on Coding Style Conventions

maiconpintoabreu:
Refactoring based on:
`Pointers	MyType *pointer	Texture2D *array = NULL;`
and
`Operators	value1*value2	int product = value*6;`


raysan5:
@maiconpintoabreu thank you very much! I love those small tweaks for consistency!

PR #5518, opened 2026-01-28
[rmodels] Added a check in `isMaterialValid()` function

ssszcmawo:
Added a check to ensure the material has a valid texture , if i understand correctly this function must return `true` when both `Shader` and `Texture` are valid so i slightly refactored the logic and switched from two `if` statements to two `bool` flags

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -2228,10 +2228,14 @@ bool IsMaterialValid(Material material)
 {
     bool result = false;
 
-    if ((material.maps != NULL) &&      // Validate material contain some map
-        (material.shader.id > 0)) result = true; // Validate material shader is valid
+    bool hasValidShader = ((material.maps != NULL) &&   // Validate material contain some maps
+        (material.shader.id > 0));  // Validate material shader is valid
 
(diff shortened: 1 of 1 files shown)


src/rmodels.c
```diff
@@ -2228,10 +2228,14 @@ bool IsMaterialValid(Material material)
...
     bool result = false;
 
-    if ((material.maps != NULL) &&      // Validate material contain some map
-        (material.shader.id > 0)) result = true; // Validate material shader is valid
+    bool hasValidShader = ((material.maps != NULL) &&   // Validate material contain some maps
+        (material.shader.id > 0));  // Validate material shader is valid
 
-    // TODO: Check if available maps contain loaded textures
+    bool hasValidTexture = ((material.maps != NULL) && // Validate material contain some maps
+        (material.maps->texture.id > 0));   // Validate material texture is valid
```
raysan5: This check is not correct, `material.maps[]` is actually and array of `MaterialMap`, not a single map. Also, there could be materials not requiring a texture, just defining a plain color, and still be valid Materials.

PR #5521, opened 2026-01-29
[rcore] Fixed win32 vsync flag not being applied

mikeemm:
CORE.Window.flags is supposed to reflect the window's current state, but it is currently being used to determine whether VSync needs to be turned on or off, so it will always remain off. This PR fixes this so VSync can be turned on or off. I also simplified the call to wglSwapIntervalEXT for easier reading.


raysan5:
@mikeemm Nice, thanks for the review!

PR #5523, opened 2026-01-30
trivial: Correct typo in log message.

eloj:


raysan5:
@eloj Good catch! Thanks!

PR #5525, opened 2026-02-02
Added documentation comments for `LoadDirectoryFilesEx`

XenoMustache:
Added the lines 
```
// Use 'DIR*' to include directories on directory scan
// Use '*.*' to include all file types and directories on directory scan
```
To reflect new `FILE_FILTER_TAG_ALL` and `FILE_FILTER_TAG_DIR_ONLY` macros. These resolved to `*.*` and `DIR*` respectively.


raysan5:
@XenoMustache thanks for the improvement, note that you can also use the macro names, they are there for convenience, and they can also be customized if desired

PR #5526, opened 2026-02-03
[Backend/RGFW] fixed an issue when using an empty window title

CrackedPixel:
Same thing GLFW does


raysan5:
@CrackedPixel thanks for the update

PR #5529, opened 2026-02-04
Close opened directory in `GetDirectoryFileCountEx()`

alexf91:
The directory opened in `GetDirectoryFileCountEx()` was never closed, introducing a memory leak.


raysan5:
@alexf91 Good catch! That function was redesigned a couple of weeks ago! Thank for the review!

PR #5530, opened 2026-02-04
[Refactor] Refactor accumulators to follow the conventions

maiconpintoabreu:
Tested on C and GSLS and changing pre-incrementation to post-incrementation inside a for does not change the functionality.


raysan5:
@maiconpintoabreu Thank you very much! Nice consistency improvement! I frequently check code to detect those small details and I still miss some! Thanks!

Issue #5510, opened 2026-01-24
[rshapes] `CheckCollisionLines()` does not work properly for almost-horizontal line


bielern:
### Issue description

I tried to detect a collision between two lines. One perfectly horizontal and the other one almost perfectly vertical. If the almost vertical line was shifted a bit (either exactly vertical or more tilted), the collision was detected correctly.

I have created a reference implementation that was able to capture the collision. (Although I am not an expert here).

I have tried to also debug the code from the [repo](https://github.com/raysan5/raylib/blob/a33ae4a8ef8809cc7f3c078651bd91e6e56591f4/src/rshapes.c#L2366). The condition on line 2379 is `true` (because of `true && (false || true)`) and because `xi = 209.352905`.

### Environment

MacOS, M1, raylib 5.5

### Program Output

```
No raylib COLLISION
MyCheck COLLISION @ (209.352890, 160.000000)
MyCheckNoVec COLLISION @ (209.352890, 160.000000)
```

### Code Example

```
#include <stdio.h>
#include <stdlib.h>
#include <math.h>
#include <float.h>      // Required for: FLT_EPSILON

#include "raylib.h"
#include "raymath.h"


float Vector2Cross(Vector2 A, Vector2 B) {
    return A.x * B.y - A.y * B.x;
}

bool MyCheck(Vector2 A, Vector2 B, Vector2 C, Vector2 D, Vector2 *collision) {
    Vector2 r = Vector2Subtract(B, A);
    Vector2 s = Vector2Subtract(D, C);

    float div = Vector2Cross(r, s);
    
    if (fabs(div) < FLT_EPSILON) {
        return false;
    }

    Vector2 AC = Vector2Subtract(C, A);

    float t = Vector2Cross(AC, s) / div;
    float u = Vector2Cross(AC, r) / div;

    if (0.0f <= t && t <= 1.0 && 0.0f <= u && u <= 1.0f) {
        collision->x = A.x + t * r.x;
        collision->y = A.y + t * r.y;
        return true;
    }
    return false;
}

bool MyCheckNoVec(Vector2 A, Vector2 B, Vector2 C, Vector2 D, Vector2 *collision) {
    float rx = B.x - A.x;
    float ry = B.y - A.y;
    float sx = D.x - C.x;
    float sy = D.y - C.y;

    float div = rx * sy - ry * sx;
    
    if (fabs(div) < FLT_EPSILON) {
        return false;
    }

    float ACx = C.x - A.x;
[...]

PR #5531, opened 2026-02-04
Fix: Detect collision if one line is almost vertical (#5510)

bielern:


raysan5:
@bielern thanks for the improvement!

bielern:
You are welcome! Thank you for the nice library!

PR #5534, opened 2026-02-05
Fix out of bound Memory read in Material.maps

Sumethh:
Currently when rendering a mesh through the DrawMeshInstanced function, if the shader has a specular shaderloc defined it will read out of bounds as it uses `SHADER_LOC_COLOR_SPECULAR` as the indexing value into the `material.maps` array which is by default 12 long. `SHADER_LOC_COLOR_SPECULAR` is of value 13, this value should be using the `MATERIAL_MAP_SPECULAR` define which translates to `MATERIAL_MAP_METALNESS`.


raysan5:
@Sumethh Good catch! Thanks for the review!

PR #5536, opened 2026-02-07
[rcore] Implemented SetWindowMaxSize, SetWindowMinSize, and SetWindowSize

mikeemm:
This PR implements the functions required to limit the window's size, as well as to directly set it (respecting said limits), for the native win32 backend. It behaves like the GLFW implementation (automatic resize upon setting limits that contradict the current window, warning and prevention when setting incompatible limits). It currently does not try to take DPI into account, as DPI doesn't work yet and that part will probably have to be reworked when DPI is tested regardless.

I would've preferred to call CalcWindowSize in SetWindowMax/MinSize so as to not repeat those operations on receiving WM_GETMINMAXINFO, but I didn't figure out a clean way to do it without adding extra variables to CORE.Window. I also set screenMax.width/height to 9999 by default in InitPlatform, as I don't think there's any downsides to doing so since screenMax doesn't seem to be used or retrieved anywhere else and it simplifies the constraint checks significantly; if it's preferable to keep it as 0 by default, I'll change it no problem.

I also removed the TODO inside HandleWindowResize, as I believe the function already modifies the framebuffer's size as requested correctly, unless I'm possibly missing something? Same with WM_APP_UPDATE_WINDOW_SIZE, I don't think this is being used nor is it useful for anything currently, but I could restore it if needed of course.

### DIFF
--- a/src/platforms/rcore_desktop_win32.c
+++ b/src/platforms/rcore_desktop_win32.c
@@ -972,6 +972,12 @@ void SetWindowMonitor(int monitor)
 // Set window minimum dimensions (FLAG_WINDOW_RESIZABLE)
 void SetWindowMinSize(int width, int height)
 {
+    if ((width > CORE.Window.screenMax.width) || (height > CORE.Window.screenMax.height))
+    {
+        TRACELOG(LOG_WARNING, "WIN32: WINDOW: Cannot set minimum screen size higher than the maximum");
+        return;
+    }
+
     CORE.Window.screenMin.width = width;
     CORE.Window.screenMin.height = height;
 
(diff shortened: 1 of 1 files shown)


raysan5:
@mikeemm thank you very much for the review, it looks great to me. I always preoritize simpler approaches.

About HighDPI/ContentScaling support, it requires a separate review. I worked on that for glfw implementation last holidays, I'll try to jump into it at some moment.

mikeemm:
@raysan5 I'm glad it was acceptable, I redid the whole thing a few times over as I wasn't sure which approach to take.

I don't have access to a pre-Win10 PC at the moment, but I could review the DPI stuff for Win10+ sometime soon if that's okay?

raysan5:
> but I could review the DPI stuff for Win10+ sometime soon if that's okay?

Feel free. Note that I reviewed it for GLFW last holidays for Windows/Linux/macOS. I created [`core_highdpi_testbed`](https://github.com/raysan5/raylib/blob/master/examples/core/core_highdpi_testbed.c) for testing. You can test it changing Content Scaling options on Windows (100%, 125%, 150%, 200%), I also tested with a second 4K monitor, moving the window from one to the other with/without the HIGHDPI flag.

EDIT: I'd love to make this backend the Windows default, replacing GLFW if possible.

PR #5540, opened 2026-02-09
raymath: wrap float3 and float16 for consistency with other types

n-s-kiselev:
Adds conditional type guards for float3 and float16 in raymath using dedicated macros. This prevents redefinition conflicts when these types are provided by external libraries, while preserving the existing layout and behavior.


raysan5:
@n-s-kiselev thanks for the review!

PR #5541, opened 2026-02-09
[raudio] add AudioCallbackEx with context ptr

fmixolydian:
A significant issue with raylib's direct audio implementation is that it's impossible to pass a context pointer to callback functions.

This PR aims to achieve the same goal of #3700, although introducing a different, non-breaking API in the form of SetAudioStreamCallbackEx and friends.

Internal data structures (`rAudioProcessor`, `rAudioBuffer`) have been extended with an AudioCallbackEx pointer and context voidptr.

Audio stream attach/detach functions have been added and modified accordingly (with a minor refactor of the linked-list implementation of AudioProcessors to reduce repetition).

Also added some documentation to linked list append/remove funcs, since it was quite lacking in the original code.

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -1155,6 +1155,11 @@
           "description": "Pointer to internal data used by the audio system"
         },
         {
+          "type": "rAudioProcessor *",
+          "name": "processor",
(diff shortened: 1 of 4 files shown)


raysan5:
@fmixolydian Thanks but I prefer to avoid this new function approach, it would require redesigning other callbacks for consistency.

PR #5543, opened 2026-02-10
[examples] Add bone blending animation example

dmitrii-brand:
- Demonstrates per-bone animation blending for smooth transitions
- Supports upper/lower body selective blending (walk + attack)
- Includes uniform blending mode for comparison
- Uses GPU skinning for performance
- Follows raylib example conventions

### DIFF
--- a/examples/models/models_animation_bone_blending.c
+++ b/examples/models/models_animation_bone_blending.c
@@ -1,0 +1,291 @@
+/*******************************************************************************************
+*
+*   raylib [models] example - animation bone blending
+*
+*   Example complexity rating: [★★★★] 4/4
+*
+*   Example originally created with raylib 5.5, last time updated with raylib 5.5
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
+*   This example demonstrates per-bone animation blending, allowing smooth transitions
+*   between two animations by interpolating bone transforms. This is useful for:
+*   - Blending movement animations (walk/run) with action animations (jump/attack)
+*   - Creating smooth animation transitions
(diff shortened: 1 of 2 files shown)


raysan5:
@dmitrii-brand Note that there is another similar example here: https://github.com/raysan5/raylib/pull/4578, how does it compare?

dmitrii-brand:
> @dmitrii-brand Note that there is another similar example here: #4578, how does it compare?

This is just an example, it does not make any changes to lib itself.

examples/models/models_animation_bone_blending.c
```diff
@@ -0,0 +1,291 @@
...
+*   BSD-like license that allows static linking with closed source software
+*
+*   This example demonstrates per-bone animation blending, allowing smooth transitions
+*   between two animations by interpolating bone transforms. This is useful for:
+*   - Blending movement animations (walk/run) with action animations (jump/attack)
+*   - Creating smooth animation transitions
+*   - Layering animations (e.g., upper body attack while lower body walks)
+*
+*   Note: Due to limitations in the Apple OpenGL driver, GPU skinning does not work on MacOS
+*
```
raysan5, line 19: Note that the header section does not follow required structure, there is no Copyright line, it should be added or it will fail on processing.

raysan5:
@dmitrii-brand Please, could you rename the example to just `models_animation_blending`? Also a screenshot needs to be provided, you can capture automatically it from running example with F12.

Also added a comment for review about the header information, copyright data is required, check other examples for reference.

dmitrii-brand:
Thanks will do

raysan5:
@dmitrii-brand I'm merging the example, I will review the missing bits. Thanks for the addition.

PR #5547, opened 2026-02-11
fix(examples): don't bleed fog when on edge

vegerot:
Steps to reproduce:
1. play textures_fog_of_war example
2. Move player to edge of screen
3. Note the light bleeds to the other side of the screen

Before: 
image

After:
image


raysan5:
@vegerot thanks for the review!

PR #5548, opened 2026-02-12
[rcore][SDL] fix SDL version SetGamepadMappings

arlez80:
The `SetGamepadMappings` in the SDL version doesn't support multiple lines.
This is because `SDL_GameControllerAddMapping` only accepts one set of data. unlike `glfwUpdateGamepadMappings`.
also fixed the return value being different from the GLFW version.

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1287,7 +1287,25 @@ void OpenURL(const char *url)
 // Set internal gamepad mappings
 int SetGamepadMappings(const char *mappings)
 {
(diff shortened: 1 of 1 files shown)


raysan5:
@arlez80 Please, could you review code to follow [raylib coding conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

Also, I prefer to avoid `size_t` usage and `do`-`while` style loops.

arlez80:
@raysan5 I have made the changes in following coding conventions.
and I avoids `size_t` and `do`-`while` style loop.

raysan5:
@arlez80 thanks for the review!

PR #5554, opened 2026-02-14
[rcamera] update camera pan speed

CrackedPixel:
It was brought up before that when it updated to use `GetFrameTime()` with the camera speeds that the pan speed was too slow, but no one updated it yet

This feels way more comfortable and can be easily tested in `core/core_3d_camera_free`


raysan5:
@CrackedPixel I'm merging it but it can be a breaking change for some users...

PR #5557, opened 2026-02-14
[examples] change d-pad text to shapes

CrackedPixel:
It says D-Pad but shows what looks like "face buttons", so i thought it would maybe be better to change it to shapes. I don't think it really matters, though, so you can close this if you want

before:
image

after:
image


raysan5:
@CrackedPixel thanks for the review of all  those examples. Actually, I think this one  could be furtheer improved.

PR #5558, opened 2026-02-14
[examples] added missing builds from makefile

CrackedPixel:
the `others` folder was being skipped from building


raysan5:
@CrackedPixel that was on pourpose, some of those examples have specific requirements. Actually, I'm planning to remove that category in a close future.

PR #5561, opened 2026-02-15
Fixed doesn't property load when 1 frame animation.

arlez80:
`LoadModelAnimationsGLTF` doesn't properly loads 1 frame animation.
This is because `GetPoseAtTimeGLTF` will be returns interpolated value to `void *data`.
but It doesn't set any values for no need interpolate animations.


raysan5:
@arlez80 thanks for the review

PR #5563, opened 2026-02-15
[rcore] Add SetScreenshotKey to allow users to override the default F12 keybinding

JeffM2501:
Raylib has SetExitKey that allows the user to change the key used for closing the window (defualt escape)
but does not offer the same functionality to set the screenshot key (F12).
This PR adds SetScreenshotKey to let users set the key in the same way for consistency in the API.

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -5059,6 +5059,17 @@
       ]
     },
     {
+      "name": "SetScreenshotKey",
+      "description": "Set a custom key for taking a screenshot (default is F12)",
+      "returnType": "void",
+      "params": [
+        {
+          "type": "int",
+          "name": "key"
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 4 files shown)


raysan5:
@JeffM2501 Thanks but I have no plans to add a specific function for that. Exit key is critical for many projects so a function to allow disabling it is important to be available. Screenshot is just a convenience functionality I need for my students, users needing to bind it to another key can just disable it with available flag and use the following one-line code for a custom implementation:

```c
if (IsKeyPressed(KEY_SPACE)) TakeScreenshot("screenshot_name.png");
```

JeffM2501:
It's more about disabling it without needing to add an entirely new config.h

raysan5:
> It's more about disabling it without needing to add an entirely new config.h

I see, the `config.h` system redesign should allow it. Does F12 interfere with some keys needed on a specific use case?

PR #5564, opened 2026-02-15
[rcore][glfw] Fix window scaling on Wayland with GLFW 3.4+

### LINKED ISSUE #5504: [rcore][glfw] Window scaling in Linux Wayland is broken
https://github.com/user-attachments/assets/e08c33b6-b827-48ff-aefd-dcbe44c01457

---

 - [X] I checked there is no similar issue already reported
 - [X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)

### Issue description

*Briefly describe the issue you are experiencing (or the feature you want to see added to raylib). Tell us what you were trying to do and what happened instead. Remember, this is not the best place to ask questions. For questions, go to [raylib Discord server](https://discord.gg/raylib).*

### Environment

# System Details Report

Screen scaling is 125%

- **Date generated:**                              2026-01-19 20:12:50

## Hardware Information:
- **Processor:**                                   Intel® Core™ Ultra 5 225H × 14
- **Graphics:**                                    Intel® Graphics (ARL)

## Software Information:
- **OS Name:**                                     Fedora Linux 43 (Workstation Edition)
- **OS Type:**                                     64-bit
- **GNOME Version:**                               49
- **Windowing System:**                            Wayland
- **Kernel Version:**                              Linux 6.18.5-200.fc43.x86_64

### Code Example

Running the example: core_highdpi_testbed.c

0xPD33:
## Summary

GLFW 3.4 changed `GLFW_SCALE_FRAMEBUFFER` to default `TRUE` on all platforms. Previously only macOS set this hint explicitly. On Wayland with display scaling (1.5x, 2x, etc.), this causes:

- Content rendering in a subset of the window (typically bottom-left)
- Incorrect mouse coordinates
- Wrong `GetScreenWidth`/`GetScreenHeight` values

### Changes

- **Non-HiDPI path:** Disable `GLFW_SCALE_FRAMEBUFFER` on Wayland when `FLAG_WINDOW_HIGHDPI` is not set, restoring 1:1 window-to-framebuffer mapping
- **HiDPI init:** On Wayland, read actual framebuffer size from GLFW via `glfwGetFramebufferSize()` instead of calling `glfwSetWindowSize()` with scaled dimensions (which double-scales, since `GLFW_SCALE_TO_MONITOR` has no effect on Wayland but `GLFW_SCALE_FRAMEBUFFER` scales independently)
- **Mouse scaling:** Skip `SetMouseScale()` on Wayland since GLFW already reports mouse coordinates in logical (window) space — unlike X11/Windows where `GLFW_SCALE_TO_MONITOR` resizes the window and mouse coords are in pixel space

All changes are guarded by `glfwGetPlatform() == GLFW_PLATFORM_WAYLAND` to avoid affecting existing X11/Windows/macOS behavior.

## Test plan

- [x] Tested at 1x, 1.5x, and 2x display scaling on Niri (Wayland compositor)
- [x] Standard mode (no `FLAG_WINDOW_HIGHDPI`): grid fills window, corner rectangles touch edges, mouse tracks correctly
- [x] HiDPI mode (`FLAG_WINDOW_HIGHDPI`): same visual checks, `GetScreenWidth` returns logical size, `GetRenderWidth` returns physical size
- [ ] X11/Windows — existing behavior unchanged (no Wayland platform detected, original code paths execute)

**Environment:** NixOS, Niri compositor, GLFW 3.4, NVIDIA RTX 5060 Ti (driver 580.126.09)

Fixes #5504

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1469,10 +1469,8 @@ int InitPlatform(void)
 #if defined(__APPLE__)
         glfwWindowHint(GLFW_SCALE_FRAMEBUFFER, GLFW_FALSE);
 #endif
-        // GLFW 3.4+ defaults GLFW_SCALE_FRAMEBUFFER to TRUE, causing framebuffer/window
-        // size mismatch on Wayland with display scaling. Disable without FLAG_WINDOW_HIGHDPI.
-        if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND)
-            glfwWindowHint(GLFW_SCALE_FRAMEBUFFER, GLFW_FALSE);
+        // GLFW 3.4+ defaults GLFW_SCALE_FRAMEBUFFER to TRUE, causing framebuffer/window size mismatch on Wayland with display scaling
+        if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND) glfwWindowHint(GLFW_SCALE_FRAMEBUFFER, GLFW_FALSE);
     }
 
     // Mouse passthrough
@@ -1671,14 +1669,10 @@ int InitPlatform(void)
             // NOTE: On APPLE platforms system manage window and input scaling
             // Framebuffer scaling is activated with: glfwWindowHint(GLFW_SCALE_FRAMEBUFFER, GLFW_TRUE);
 
-            // Screen scaling matrix is required in case desired screen area is different from display area
-            CORE.Window.screenScale = MatrixScale(scaleDpi.x, scaleDpi.y, 1.0f);
-
 #if !defined(__APPLE__)
             if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND)
             {
-                // On Wayland, GLFW_SCALE_FRAMEBUFFER handles scaling; read actual framebuffer
-                // size instead of resizing the window (which would double-scale).
+                // On Wayland, GLFW_SCALE_FRAMEBUFFER handles scaling; read actual framebuffer size instead of resizing the window (which would double-scale)
                 int fbWidth, fbHeight;
                 glfwGetFramebufferSize(platform.handle, &fbWidth, &fbHeight);
                 CORE.Window.render.width = fbWidth;
@@ -1876,8 +1870,7 @@ static void FramebufferSizeCallback(GLFWwindow *window, int width, int height)
         CORE.Window.screenScale = MatrixScale(1.0f, 1.0f, 1.0f);
         SetMouseScale(1.0f, 1.0f);
 
-        // On Wayland with GLFW_SCALE_FRAMEBUFFER, the framebuffer is still scaled
-        // in fullscreen. Use logical window size as screen and apply screenScale.
+        // On Wayland with GLFW_SCALE_FRAMEBUFFER, the framebuffer is still scaled in fullscreen, use logical window size as screen and apply screenScale
         if ((glfwGetPlatform() == GLFW_PLATFORM_WAYLAND) &&
             FLAG_IS_SET(CORE.Window.flags, FLAG_WINDOW_HIGHDPI))
         {
@@ -1905,8 +1898,7 @@ static void FramebufferSizeCallback(GLFWwindow *window, int width, int height)
             CORE.Window.screenScale = MatrixScale(scaleDpi.x, scaleDpi.y, 1.0f);
 #if !defined(__APPLE__)
             // On Wayland, mouse coords are already in logical space
-            if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND)
-                SetMouseScale(1.0f/scaleDpi.x, 1.0f/scaleDpi.y);
+            if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND) SetMouseScale(1.0f/scaleDpi.x, 1.0f/scaleDpi.y);
 #endif
         }
         else
@@ -1936,8 +1928,7 @@ static void WindowContentScaleCallback(GLFWwindow *window, float scalex, float s
 
 #if !defined(__APPLE__)
     // On Wayland, mouse coords are already in logical space
-    if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND)
-        SetMouseScale(1.0f/scalex, 1.0f/scaley);
+    if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND) SetMouseScale(1.0f/scalex, 1.0f/scaley);
 #endif
 }
 


kevin-kmetz:
I noted in the relevant open issue that this pull request when checked out resolved the issue on my system (Fedora + Wayland + 4 monitors + one scaled up to 225%).

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1667,11 +1675,23 @@ int InitPlatform(void)
             CORE.Window.screenScale = MatrixScale(scaleDpi.x, scaleDpi.y, 1.0f);
```
CrackedPixel: we should also get rid of these 2 lines. its duplicate from 1660 (1668 in yours), presumably by mistake

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1465,6 +1469,10 @@ int InitPlatform(void)
 #if defined(__APPLE__)
         glfwWindowHint(GLFW_SCALE_FRAMEBUFFER, GLFW_FALSE);
 #endif
+        // GLFW 3.4+ defaults GLFW_SCALE_FRAMEBUFFER to TRUE, causing framebuffer/window
+        // size mismatch on Wayland with display scaling. Disable without FLAG_WINDOW_HIGHDPI.
```
CrackedPixel: the period on this comment should be removed per the guidelines/conventions

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1465,6 +1469,10 @@ int InitPlatform(void)
...
+        if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND)
```
CrackedPixel: this should be 1 line per the guidelines/conventions

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1867,8 +1904,9 @@ static void FramebufferSizeCallback(GLFWwindow *window, int width, int height)
             CORE.Window.screen.height = (int)((float)height/scaleDpi.y);
             CORE.Window.screenScale = MatrixScale(scaleDpi.x, scaleDpi.y, 1.0f);
 #if !defined(__APPLE__)
-            // Mouse input scaling for the new screen size
-            SetMouseScale(1.0f/scaleDpi.x, 1.0f/scaleDpi.y);
+            // On Wayland, mouse coords are already in logical space
+            if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND)
+                SetMouseScale(1.0f/scaleDpi.x, 1.0f/scaleDpi.y);
```
CrackedPixel: this should be 1 line per the guidelines/conventions

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1897,8 +1935,9 @@ static void WindowContentScaleCallback(GLFWwindow *window, float scalex, float s
     CORE.Window.screenScale = MatrixScale(scalex, scaley, 1.0f);
 
 #if !defined(__APPLE__)
-    // Mouse input scaling for the new screen size
-    SetMouseScale(1.0f/scalex, 1.0f/scaley);
+    // On Wayland, mouse coords are already in logical space
+    if (glfwGetPlatform() != GLFW_PLATFORM_WAYLAND)
+        SetMouseScale(1.0f/scalex, 1.0f/scaley);
```
CrackedPixel: this should be 1 line per the guidelines/conventions

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1667,11 +1675,23 @@ int InitPlatform(void)
             CORE.Window.screenScale = MatrixScale(scaleDpi.x, scaleDpi.y, 1.0f);
 
 #if !defined(__APPLE__)
-            // Mouse input scaling for the new screen size
-            SetMouseScale(1.0f/scaleDpi.x, 1.0f/scaleDpi.y);
+            if (glfwGetPlatform() == GLFW_PLATFORM_WAYLAND)
+            {
+                // On Wayland, GLFW_SCALE_FRAMEBUFFER handles scaling; read actual framebuffer
+                // size instead of resizing the window (which would double-scale).
```
CrackedPixel: the period on this comment should be removed per the guidelines/conventions

src/platforms/rcore_desktop_glfw.c
```diff
@@ -1855,6 +1875,23 @@ static void FramebufferSizeCallback(GLFWwindow *window, int width, int height)
         CORE.Window.screen.height = height;
         CORE.Window.screenScale = MatrixScale(1.0f, 1.0f, 1.0f);
         SetMouseScale(1.0f, 1.0f);
+
+        // On Wayland with GLFW_SCALE_FRAMEBUFFER, the framebuffer is still scaled
+        // in fullscreen. Use logical window size as screen and apply screenScale.
```
CrackedPixel: the period on this comment should be removed per the guidelines/conventions

CrackedPixel:
You beat me to it and did it better :rofl:  nice job :+1:

0xPD33:
Will fix the formatting now, thanks for the review!

raysan5:
@0xPD33 thank you very much for the fix and @kevin-kmetz @CrackedPixel thanks for reviewing it.

I tested it on Windows 10 and macOS, with multi-monitor and multi-dpi, with HighDPI enabled and disabled. All seems to work as expected.

My only concern, there is no need to include platform-specific code that won't be executed so, maybe `#if defined(_GLFW_WAYLAND)` can be used instead of checking current platform through glfw? Afair, it doesn't need additional changes on build systems...

planetis-m:
@raysan5 isn't figuring out which platform to use (X11/Wayland) negotiated at runtime? Seems like a downgrade to use defines.

raysan5:
@planetis-m no, platform to use always has been defined by user on Makefile, at least for GLFW backend

planetis-m:
@raysan5 was this PR reverted? https://github.com/raysan5/raylib/pull/3863

raysan5:
@planetis-m I don't know, you can probably check That PR was modifying an external library that potentially could be updated so changes reverted. I don't know at the moment if it was the case.

In any case I think it's a better approach let the dev choose the desired windowing system backend instead of making assumptions depending on the windowing system where the application is compiled.

planetis-m:
> @planetis-m I don't know, you can probably check That PR was modifying an external library that potentially could be updated so changes reverted. I don't know at the moment if it was the case.
> 
> In any case I think it's a better approach let the dev choose the desired windowing system backend instead of making assumptions depending on the windowing system where the application is compiled.

Ray, I may be misinterpreting your message, but I think it’s worth reconsidering what you’re asking for. Linux already has runtime platform detection since the PR I referenced. For that reason, using a C preprocessor definition would not be advisable, as it would break that behavior.

Additionally, it is generally uncommon in software development to close an issue before the PR that addresses it has been merged.

raysan5:
@planetis-m I prefer current approach, thanks for your understanding.

0xPD33:
> @0xPD33 thank you very much for the fix and @kevin-kmetz @CrackedPixel thanks for reviewing it.
> 
> I tested it on Windows 10 and macOS, with multi-monitor and multi-dpi, with HighDPI enabled and disabled. All seems to work as expected.
> 
> My only concern, there is no need to include platform-specific code that won't be executed so, maybe `#if defined(_GLFW_WAYLAND)` can be used instead of checking current platform through glfw? Afair, it doesn't need additional changes on build systems...

Okay so I can give this a shot, it does make sense to me. Would be great if we can agree on it first. Also unsure if that flag is correct or if it should be `#if defined(GLFW_PLATFORM_WAYLAND)`. The one you mentioned might be for internal use only during GLFW's own build, the one I just mentioned is the public equivalent from GLFW 3.4 headers. Please correct me if I'm wrong.

raysan5:
> Okay so I can give this a shot, it does make sense to me. Would be great if we can agree on it first. Also unsure if that flag is correct or if it should be #if defined(GLFW_PLATFORM_WAYLAND). The one you mentioned might be for internal use only during GLFW's own build, the one I just mentioned is the public equivalent from GLFW 3.4 headers. Please correct me if I'm wrong.

Reviewing the code, it seems `GLFW_PLATFORM_WAYLAND` is only intended to be used with `glfwGetPlatform()` at runtime, the compile-time flags to be checked are `_GLFW_WAYLAND` and `_GLFW_X11` [as shown here](https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_glfw.c#L770).

Further investigating the code, it seems `_GLFW_WAYLAND` is [enabled through CMake](https://github.com/raysan5/raylib/blob/master/src/external/glfw/src/CMakeLists.txt#L54) when `GLFW_BUILD_WAYLAND` is defined, but raylib does not use GLFW CMake build system, and expects to build for specific platform [using `_GLFW_WAYLAND` flag](https://github.com/raysan5/raylib/blob/master/src/rglfw.c#L106C17-L106C30), that It's managed by [raylib Makefile](https://github.com/raysan5/raylib/blob/master/src/Makefile#L451).

So, `_GLFW_WAYLAND` and `_GLFW_X11` must be checked in code for compile time. The only concern is that it seems both flags can be enabled, if both backends are supported (X11 + Wayland).

raysan5:
@0xPD33 I'm checking the code and I think it can be tricky in some cases to consider the flags, so I'm merging and reviewing it myself when required. Thank you very much for the improvement.

planetis-m:
@raysan5  Please don’t break Linux again. Take a close look at the previous PR I linked and what it’s supposed to do. If you want a compile-time define, use a standard #if defined(__linux__) or BSD check, and keep runtime detection for the windowing backend (X11 vs. Wayland). Thank you for being reasonable.

raysan5:
@planetis-m I don't understand what you mean, but feel free to send a PR for review with your proposed improvements. Thanks.

 > Please don’t break Linux again.

And please, moderate your tone. This kind of messages definitely do not help the project.

Issue #3743, opened 2024-01-20
[raudio] Crash when changing playback device


thabetx:
### Issue description

Example (audio_sound_loading.c) crashes when changing playback device,.

### Environment

This happens on Windows.

<details><summary>Initialization Logs</summary>

INFO: Initializing raylib 5.1-dev
INFO: Platform backend: DESKTOP (GLFW)
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
INFO:     > rlgl:...... loaded (mandatory)
INFO:     > rshapes:... loaded (optional)
INFO:     > rtextures:. loaded (optional)
INFO:     > rtext:..... loaded (optional)
INFO:     > rmodels:... loaded (optional)
INFO:     > raudio:.... loaded (optional)
INFO: DISPLAY: Device initialized successfully
INFO:     > Display size: 1600 x 900
INFO:     > Screen size:  800 x 450
INFO:     > Render size:  800 x 450
INFO:     > Viewport offsets: 0, 0
INFO: GLAD: OpenGL extensions loaded successfully
INFO: GL: Supported extensions count: 230
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Intel
INFO:     > Renderer: Intel(R) UHD Graphics
INFO:     > Version:  3.3.0 - Build 27.20.100.8190
INFO:     > GLSL:     3.30 - Build 27.20.100.8190
INFO: GL: VAO extension detected, VAO functions loaded successfully
INFO: GL: NPOT textures extension detected, full NPOT textures supported
INFO: GL: DXT compressed textures supported
INFO: GL: ETC2/EAC compressed textures supported
INFO: GL: ASTC compressed textures supported
INFO: PLATFORM: DESKTOP (GLFW): Initialized successfully
INFO: TEXTURE: [ID 1] Texture loaded successfully (1x1 | R8G8B8A8 | 1 mipmaps)
INFO: TEXTURE: [ID 1] Default texture loaded successfully
INFO: SHADER: [ID 1] Vertex shader compiled successfully
INFO: SHADER: [ID 2] Fragment shader compiled successfully
INFO: SHADER: [ID 3] Program shader loaded successfully
INFO: SHADER: [ID 3] Default shader loaded successfully
INFO: RLGL: Render batch vertex buffers loaded successfully in RAM (CPU)
INFO: RLGL: Render batch vertex buffers loaded successfully in VRAM (GPU)
[...]

PR #5568, opened 2026-02-17
[raudio] Remove usage of `ma_data_converter_get_required_input_frame_count()`

mackron:
The function `ma_data_converter_get_required_input_frame_count()` is being removed from miniaudio. This PR is prep work to ease the transition for when the time comes to update to version miniaudio v0.12.

To make this work with the current architecture of raylib it requires the use of a cache. It should technically be possible to avoid the cache for a static AudioBuffer, but it would be unpractical to avoid it with streams and callback-based AudioBuffers so I've decided to keep it simple and use a generic solution. This might be a worthwhile optimization to keep in mind for later.

The size of the cache can be configured with a macro called `AUDIO_BUFFER_CONVERSION_CACHE_SIZE`. Higher values use more memory, but have less overhead. There's diminishing returns the higher you go.

I've tried to keep the coding style consistent. Happy to fix anything I've let slip.

I've run all audio_* examples on Linux. Could be worth someone having a second look at it however. In testing I did actually find a few issues with some examples, but these were also present prior to this PRs changes and are unrelated. I'll leave those issues for a separate discussion.


src/raudio.c
```diff
@@ -337,6 +341,9 @@ typedef enum {
 // Audio buffer struct
 struct rAudioBuffer {
     ma_data_converter converter;    // Audio data converter
+    unsigned char* converterCache;  // Cached input samples for use by the converter when resampling is required
+    unsigned int converterCacheCap; // The capacity of the converter cache in frames
```
raysan5: If the `converterCacheCap` is always `AUDIO_BUFFER_CONVERSION_CACHE_SIZE` and there is no intention of dynamically change this array, probably this variable can be avoided and just use `AUDIO_BUFFER_CONVERSION_CACHE_SIZE`

src/raudio.c
```diff
@@ -337,6 +341,9 @@ typedef enum {
 // Audio buffer struct
 struct rAudioBuffer {
     ma_data_converter converter;    // Audio data converter
+    unsigned char* converterCache;  // Cached input samples for use by the converter when resampling is required
```
raysan5: Is this the only option? All audio elements will now add that extra cache data? It increases memory requirements considerably for each audio...

src/raudio.c
```diff
@@ -337,6 +341,9 @@ typedef enum {
...
+    unsigned int converterCacheCap; // The capacity of the converter cache in frames
+    unsigned int converterCacheLen; // The number of valid frames sitting in the converter cache
```
raysan5: I understand this is the remaining length of data after every iteration, right?

raysan5:
@mackron Thanks for the update! I added some questions, not sure if I'm understanding the need for the data cache...

As per my understanding it stores the stream data converted, ready to be played, on a user side buffer. Was it previously done inside `miniaudio`? Why is it required to cache that data? To improve performance? The conversion is not fast enough on-the-go?

I also understand this improvement is mostly for streamed data and static data won't need to initialize that cache array, could it be just skipped if `usage == AUDIO_BUFFER_USAGE_STATIC`?

src/raudio.c
```diff
@@ -586,6 +593,16 @@ AudioBuffer *LoadAudioBuffer(ma_format format, ma_uint32 channels, ma_uint32 sam
...
 
+    // A cache for use by the converter is necessary when resampling because
+    // when generating output frames a different number of input frames will
+    // be consumed. Any residual input frames need to be kept track of to
+    // ensure there are no discontinuities. Since raylib supports pitch
+    // shifting, which is done through resampling, a cache will always be
+    // required. This will be kept relatively small to avoid too much wastage.
+    audioBuffer->converterCacheLen = 0;
+    audioBuffer->converterCacheCap = AUDIO_BUFFER_CONVERSION_CACHE_SIZE;
+    audioBuffer->converterCache = (unsigned char*)RL_CALLOC(audioBuffer->converterCacheCap*ma_get_bytes_per_frame(format, channels), 1);
```
raysan5: Could it just be skipped `if (usage == AUDIO_BUFFER_USAGE_STATIC)`?

psxdev: Hello,  for future 0.12 update remove usage of ma_get_backend_name and ma_get_format_name because wiil not be available

mackron:
The cache is not for performance, it's to avoid glitching. So when you are resampling, you consume input frames and from that generate output frames. The complication is that the number of output frames you generate is different from the number of input frames you'll consume. Say, for example, you need to output 2 output frames, you will *not* be consuming exactly 2 input frames. Imagine you are going from 48000 to 24000. That's a ratio of 2:1. If you output 2 frames, you'll be consuming 4 input frames.

The way audio works in raylib is the audio device requests a number of frames from a callback, and raylib's job is to fill the output buffer with audio data. If the audio device asks for 512 frames, which is the output frames, how does raylib know how many input frames will be required to fill that? It doesn't, so it continuously reads from the `AudioBuffer` until the output buffer is full. But in order to generate those output frames, it needs input data, and that input data needs to be pulled from the `AudioBuffer`. But since you don't know how much input data is required, you need to take a best guess, and sometimes that guess is wrong. The cache is required for storing the residual input frames for when the guess goes wrong. The next time audio is processed, this residual input will be processed from the cache, and the processing continues. If you were to just discard that residual you would have a discontinuity in the input stream which would result in glitching.

Previously, the function `ma_data_converter_get_required_input_frame_count()` would be used to predict how many input frames would be required, but that has proven to be unreliable and difficult to maintain, so I've decided to remove it hence the reason for this PR.

I'll change the `converterCacheCap` thing as you suggested. That was just my personal style leaking through.

I had anticipated you might have some reservations about the overhead of a cache which is why I mentioned the `STATIC` case. But having thought about this a bit more, I *think* I might be able to rework this so it uses maybe only 16 frames of cache or so, dependant on the maximum pitch shifting range. Will play around with some ideas and report back.

src/raudio.c
mackron: `ma_get_format_name()` still exists though? But it's no worries, I'm already working on getting raylib working with miniaudio 0.12 for the purpose of testing miniaudio's new Dreamcast stuff. I won't be submitting a PR for that until way later though.

psxdev: i am working right now to add your miniaudio backend 0.12 to raylib for dreamcast, i have some issues with raudio and samples. Initialization of miniaudio in raudio is done and get some sound but stop to play a few second later in raylib for dreamcast sample.

where would you like to speak about dreamcast port? discord channel raylib-consoles-homebrew or simulant raylib4dc channel?

Also i am thinking to add support to other sony consoles

mackron: Yeah I actually found my own bug, not related to Dreamcast specifically. I was going to get my raylib fork working with miniaudio 0.12 on my end first, and then do a post in the raylib Discord server, probably the homebrew channel. Feel free to ping me on Discord - I'm in the raylib server.

raysan5:
@mackron thank you very much for the detailed explanation, I understand now. I'll follow the progress.

mackron:
I've gone ahead and optimized the code to use a different approach and have got the cache from 256 down to 8 frames. For a stereo sound that equates to 64 bytes. I have not yet done the `STATIC` specialization that we talked about. Before I did that, I wanted to ask if you thought it was still worthwhile. The advantage to that is a `STATIC` AudioBuffer will save 64 bytes for stereo, and 32 bytes for mono, but the downside is that it'll require two separate code paths which will add code complexity and maintenance costs. I wasn't sure what your preference was here so I thought it would be best to ask first.

I did some renaming of some variables to hopefully make it a bit clearer as to the purpose of the buffer. I renamed it from "cache" to "residual" which I think should be a bit clearer? Either way, happy to make any style changes you don't like.

raysan5:
> Before I did that, I wanted to ask if you thought it was still worthwhile.

It's ok, it's small enough and I beter to avoid the code complexity of two different code paths.

I also think the renaming is good. Thank you very much for all the improvements! Merging!

casperbear:
@mackron @raysan5 It may be a coincidence but I started to hear audio artifacts after updating RayLib from master today. I believe this was the only PR that touched audio, but I may be wrong. The artifacts sound like 1-2 short clicks after the sound ends with delay approximately 0.1-0.3 seconds. The OS is Windows 7. The used method is PlaySound. It happens for many sounds but I attached an example in case you need to check wav properties.
[MenuChoose.wav](https://github.com/user-attachments/files/25455932/MenuChoose.wav)

mackron:
I hear it. Investigating.

mackron:
This should be fixed here: https://github.com/raysan5/raylib/pull/5578

Would you be able to try that branch? Thanks for reporting this.

casperbear:
@mackron I cannot test the branch, but incorporating changes into my branch fixes the issue. Thank you.

btw, it's not related to this issues, but maybe you'll be interested in working on this in the future. This PR https://github.com/raysan5/raylib/issues/3743 fixed the crash at the moment of changing default audio device, but the sound isn't restored. Switching back to the first audio device doesn't return sound either. Also if default audio device was changed, CloseAudioDevice() hangs inside ma_device_uninit(&AUDIO.System.device); It's not a big issue though because CloseAudioDevice() is only needed on exit, and if it's not called, app termination succeeds (doesn't hang or anything) even if default audio device was changed. I can easily imagine that this is specific to my outdated setup (Win 7) so I didn't create a ticket. But since you yourself is here, I'm reporting it just in case you are interested.

mackron:
@casperbear I *think* that might be addressed in the next version of miniaudio (0.12 - current version is 0.11.x). I have changed the way WASAPI works, and that COM error should not be relevant anymore. In any case, the code has changed so much now that I don't particularly have any interest in fixing that in the 0.11.x cycle. If it still happens when miniaudio 0.12 is released (any raylib is updated) I'd be very interested to hear about that.

PR #5570, opened 2026-02-17
[rshape] added checkcollisiontriangles function

emrullahcmn:
I have added `CheckCollisionTriangles()` function. That function first checks if one of the vertices of the triangles are contained by each other. If not, then, it secondly checks if the edges of the triangles are colliding.


raysan5:
@emrullahcmn I'm afraid I prefer to avoid that function, it requires lots of input parameters and the use case seems very specific, I think most users rarely need it and even in that case, the logic can be easely implemented at user-side.

emrullahcmn:
Thank you for reviewing and taking your time!

PR #5574, opened 2026-02-20
Fix text input on SDL3

TheKodeToad:
Tested with https://github.com/raysan5/raylib/blob/master/examples/text/text_input_box.c. Before this change you can't type anything - that is because SDL3 requires text input to explicitly be enabled: https://wiki.libsdl.org/SDL3/README-migration#sdl_keyboardh.

Unfortunately this has its issues such as https://github.com/libsdl-org/SDL/issues/9268... perhaps there's a workaround?

Should there be something in the license header? :P


raysan5:
@TheKodeToad oh! thanks for the review!

PR #5575, opened 2026-02-20
[examples] fix example compile error

CrackedPixel:
```
core/core_directory_files.c:82:17: error: passing argument 2 of ‘GuiListViewEx’ 
from incompatible pointer type [-Wincompatible-pointer-types]
```

raylib.h defines FilePathList.paths as `char**` (not const)
raygui `GuiListViewEx` requires `const char**`


raysan5:

PR #5576, opened 2026-02-21
[examples] Change make `clean` (testing requested)

CrackedPixel:
The previous version was tested on Linux+Mac and will wipe ALL files in the example folder. This is because it was only checking that they have the "executable" flag on them, but many files have that flag including the Makefile itself

I tested this new command on Linux+Mac but if possible can someone else check and see if they have the same behavior on these platforms?

DRM and BSD are untested by me but since their commands are similar to unix it should also work


CrackedPixel:
The `execute` permissions were being added from me copying the files to my macs. This doesnt happen with a proper git pull. Whoops! False alarm :sweat:

PR #5577, opened 2026-02-21
Audio: Improvements to device configuration

mackron:
This is a small PR with improvements to the audio device configuration. I've broken it down into multiple commits with rationale for the changes to make it easier for review.

Summary below:

  * `config.capture.*` options have been removed. There is no need to set these for a playback device.
  * Added `config.noPreSilencedOutputBuffer = true;`
  * Added `config.noFixedSizedCallback = true;`
  * Added a new config called `AUDIO_DEVICE_PERIOD_SIZE_IN_FRAMES` for configuring the period size (latency)
  * Fixed a documentation error for `pan`.

All audio examples have been tested on Linux.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -352,7 +352,7 @@ struct rAudioBuffer {
 
     float volume;                   // Audio buffer volume
     float pitch;                    // Audio buffer pitch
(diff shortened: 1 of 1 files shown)


raysan5:
@mackron Looks good. Thanks for the improvement.

PR #5578, opened 2026-02-21
Audio: Fix a glitch at the end of a sound.

mackron:
This addresses the regression reported here: https://github.com/raysan5/raylib/pull/5568#issuecomment-3938211023

The error is happening because the mixing function is not terminating when the sound reaches the end.


raysan5:
@mackron thanks for the fix, I'm testing it on Windows

PR #5579, opened 2026-02-21
Can we expose PLATFORM_WEB_RGFW to cmake?

Crisspl:
Hey,

Can we expose PLATFORM_WEB_RGFW to cmake?
Just those few lines is enough for me. Is there anything more to be done for it to be ready?

Btw: PLATFORM_WEB_RGFW is the only backend that works well for me on web (emscripten/wasm, browser). Current PLATFORM_WEB seem to have tragic input handling (a lot of mouse clicks and/or keyboard presses seem to be never recognized), while PLATFORM_WEB_RGFW works smoothly.


raysan5:
@Crisspl It can be added. But what problems do you have with `PLATFORM_WEB`? All examples are compiled for web with that backend and work well, afaik. Maybe you refer to `PLATFORM_WEB_EMSCRIPTEN`? That platform was added recently and I'm aware there are issues with the inputs.

Crisspl:
I'll try to build and check some examples to web platform then. Maybe emcc linker/compiler flags im using broke something. 
From what i've seen so far: a lot of the input events seems to never come through. Mouse position is always up to date, but mouse clicks and keyboard presses doesnt work most of the time. I've also noticed there are some problems with checking if window (canvas in web case?) is focused. It seemed even like it returned the opposite which was making rlimgui unusable because of this check: https://github.com/raylib-extras/rlImGui/blob/main/rlImGui.cpp#L820
When i get rid of this check in rlimgui (assume always true), rlimgui becomes reactive to input, but still the problem with missing most of clicks/presses is present.

Maybe focus/unfocus recognition is the root of all the problems, because RGFW backend does not use window state UNFOCUS flag at all (it's always "focused"), and GLFW (which web backend is based on) does it.

PR #5580, opened 2026-02-22
[Examples] Update cmakelists

CrackedPixel:
Some linker flags were missing for RGFW when using cmake on the examples


raysan5:

PR #5583, opened 2026-02-23
[Examples] glfw makefile consistency

CrackedPixel:
In the `src` Makefile we have the following available+used by glfw:
```make
# Enable support for X11 by default on Linux when using GLFW
# NOTE: Wayland is disabled by default, only enable if you are sure
GLFW_LINUX_ENABLE_WAYLAND  ?= FALSE
GLFW_LINUX_ENABLE_X11      ?= TRUE
```

however, this is not used in the `example` Makefile, leading to inconsistent builds and possibly misconfiguration. it should also be possible to allow both, for x11 fallback options, which the current does not allow

the current `USE_WAYLAND_DISPLAY` is only used by SDL in the `src` Makefile

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -90,6 +90,11 @@ BUILD_MODE            ?= RELEASE
 # Use external GLFW library instead of rglfw module
 USE_EXTERNAL_GLFW     ?= FALSE
 
+# Enable support for X11 by default on Linux when using GLFW
+# NOTE: Wayland is disabled by default, only enable if you are sure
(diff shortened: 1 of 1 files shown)


examples/Makefile
raysan5, line 106: Probably this line should be removed....

CrackedPixel: it is still used for the SDL, do you want it renamed? `SDL_ENABLE_WAYLAND` or something?

raysan5:
@CrackedPixel ok, looks good, thanks!

PR #5584, opened 2026-02-23
[Backend/GLFW] Remove wayland files

CrackedPixel:
This requires that the user has `glfw` package installed on their system. I think that is reasonable for linux users using glfw

NOTE: x11 builds (default) are unaffected

```
cd src
make GLFW_LINUX_ENABLE_WAYLAND=TRUE

cd examples
make GLFW_LINUX_ENABLE_WAYLAND=TRUE
```
it will compile successfully, and you can detect x11/wayland in code if needed:

```cpp
    #include <GLFW/glfw3.h>

    // after InitWindow()
    int platform = glfwGetPlatform();
    if (platform == GLFW_PLATFORM_WAYLAND) {
        printf("App using native Wayland backend\n");
    } else if (platform == GLFW_PLATFORM_X11) {
        printf("App using X11/XWayland backend\n");
    } else {
        printf("Unknown platform: %d\n", platform);
    }
   ```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -872,7 +872,7 @@ clean:	clean_shell_$(PLATFORM_SHELL)
 	@echo "removed all generated files!"
 
 clean_shell_sh:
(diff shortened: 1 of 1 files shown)


raysan5:
It has been discussed on Discord, we can't assume glfw will be available on target platform, better keepraylib as much self-contained and portable as possible.

PR #5586, opened 2026-02-23
win32 clipboard: fix for BI_ALPHABITFIELDS narrow support

Crisspl:
It doesn't compile for me (`BI_ALPHABITFIELDS` is not defined anywhere, especially not in GDI headers). Did some quick research and according to [wikipedia](https://en.wikipedia.org/wiki/BMP_file_format#Compression) it is supported only on Windows CE (4.0 and above).

### DIFF
--- a/src/external/win32_clipboard.h
+++ b/src/external/win32_clipboard.h
@@ -178,7 +178,9 @@ typedef struct tagRGBQUAD {
 #define BI_CMYK      0x000B
 #define BI_CMYKRLE8  0x000C
 #define BI_CMYKRLE4  0x000D
+#endif
 
+#ifndef BI_ALPHABITFIELDS
 // Bitmap not compressed and that the color table consists of four DWORD color masks, 
 // that specify the red, green, blue, and alpha components of each pixel
 #define BI_ALPHABITFIELDS 0x0006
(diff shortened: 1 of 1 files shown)


Crisspl:
Alternatively we could do something like

```c++
#if defined(_WIN32_WCE) && _WIN32_WCE >= 0x0400
    // CE .NET 4.0+: Headers should already define it, but guard anyway
    #ifndef BI_ALPHABITFIELDS
    #define BI_ALPHABITFIELDS 6
    #endif
#else
    // Desktop: Define if needed for CE BMP compatibility
    #define BI_ALPHABITFIELDS 6
#endif
```

but not sure if this is safe (if gdi on target platform will actually support it)

maiconpintoabreu:
If the compression is only for windows, does not need another if.

src/external/win32_clipboard.h
raysan5, line 186: Note that the value is defined here... but maybe some condition shadows it...

raysan5:
@Crisspl Thanks for reporting, the value is actually defined but maybe some condition is not correct...

Crisspl:
well, this `#if` condition https://github.com/raysan5/raylib/blob/master/src/external/win32_clipboard.h#L121  (`#if !defined(_WINBASE_) || !defined(WINBASE_ALREADY_INCLUDED)`) checks if gdi was already included, and `#define BI_ALPHABITFIELDS 0x0006` is inside this #if.
So effectively BI_ALPHABITFIELDS is not getting defined if gdi was included, whereas gdi headers does not always define BI_ALPHABITFIELDS (maybe never? or maybe only on Windows CE? i dont know frankly speaking)

maiconpintoabreu:
Sorry I got confused by the indentation of the #ifs.
`#define BI_ALPHABITFIELDS 0x0006` is inside the `#if !defined(_WINBASE_) || !defined(WINBASE_ALREADY_INCLUDED)`, I think it could be moved out as it is not defined in the `wingdi.h`

raysan5:
@maiconpintoabreu @Crisspl Please, could you move the define out to merge this PR?

Crisspl:
Sure, please take a look now. I just moved it out of #if <gdi headers included>, and and put within #ifndef BI_ALPHABITFIELDS block in case it actually was defined in gdi on this OS version

raysan5:
@Crisspl Looks good! Thanks for the update!

Crisspl:
Sure, glad to contribute:)

PR #5587, opened 2026-02-24
[examples] Add `magnifying_glass`

BadRAM:
This example demonstrates multiple practical uses of blend modes, including blend_custom_separate for the purpose of masking a render texture to a custom shape. A detailed description is given for how the specific blend factors used achieve this effect. It also demonstrates a fairly standard use of a render texture and camera2D.

Demo video:

https://github.com/user-attachments/assets/103b5065-be2d-4c28-8ac3-215bfadc50b5


raysan5:
@BadRAM Very nice example! Thanks for the addition!

PR #5589, opened 2026-02-24
ANDROID: Fix broken LoadMusicStream due to missing fopen redirect in raudio.c

ghera:
### Problem
`LoadMusicStream` fails on Android for all file formats (OGG, WAV, MP3).

This is a regression introduced when `utils.h` was removed. In raylib 5.5, `utils.h` contained the `fopen` macro override for Android and was included by both `rcore.c` and `raudio.c`. After its removal, the override was moved into `rcore_android.c`,
but that file is only pulled into `rcore.c` via `#include`.

`LoadMusicStream` calls the file-based decoder init functions directly (`stb_vorbis_open_filename`, `drwav_init_file`, `drmp3_init_file`), which internally call the standard `fopen` — unable to open assets from the APK.

`LoadWave`/`LoadSound` are not affected because they go through raylib's public `LoadFileData` (defined in `rcore.c`), where the redirect is already active.

### Fix
I placed the override in `raudio.c` as a minimal fix, but happy to move it to a more appropriate location if you prefer.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -181,6 +181,7 @@ typedef struct tagBITMAPINFOHEADER {
 #include <string.h>                     // Required for: strcmp() [Used in IsFileExtension(), LoadWaveFromMemory(), LoadMusicStreamFromMemory()]
 
 #if defined(PLATFORM_ANDROID)
+    FILE *android_fopen(const char *fileName, const char *mode);
     #define fopen(name, mode) android_fopen(name, mode)
 #endif
 


maiconpintoabreu:
Hi @ghera did you test this fix?

ghera:
Hi @maiconpintoabreu, I tested it on a dev branch of mine and it worked, but I didn't test it properly after cherry-picking the fix — it needs a forward declaration in order to work on master. Sorry about that.

raysan5:
@ghera Mmmh... I see, does it work for other types of assets (Texture2D, Font...)?

If I understand it correctly, the issue is only when the file access is not going through raylib provided `LoadFile*()` function, right?

I'd like to keep the `raudio` module as much platform-independant as possible but not sure how this issue can be addressed otherwise...

maiconpintoabreu:
> @ghera Mmmh... I see, does it work for other types of assets (Texture2D, Font...)?
> 
> If I understand it correctly, the issue is only when the file access is not going through raylib provided `LoadFile*()` function, right?
> 
> I'd like to keep the `raudio` module as much platform-independant as possible but not sure how this issue can be addressed otherwise...

To keep it out of `raudio` or `rmodel` we could explain on the Wiki that it can be done using wrap here is a cmake example:

image

implementation:

image

Logs:

image

ghera:
@raysan5 As far as I can see the only load method which doesn't use `LoadFile*` is `LoadMusicStream`. 
That said, I agree duplicating the `#define` in `raudio.c` is not the best approach. Having a shared internal header (like `utils.h`) was probably a cleaner solution, but keeping it just for platform-dependent exceptions was probably the reason it was removed in the first place.

@maiconpintoabreu The linker wrapping approach is clever and it would allow removing the `fopen` define entirely. However, I'd personally prefer Android to work out of the box, without a toolchain-dependent workaround.
There is also a potential infinite loop: the `fopen` call inside `android_fopen` after `#undef fopen` would still be redirected by the linker, causing recursion (though I think it's easily fixable using `__real_fopen`).

maiconpintoabreu:
@ghera `tinyobj_loader_c.h` uses `fopen` as well, I did not check others.

raysan5:
@maiconpintoabreu Afaik, only `tinyobj_loader_c.h` requires `fopen()` for material loading, all the other data loaded by raylib goes through memory and `LoadFileData()`/`LoadFileText()` functions, that map `fopen()` to `android_fopen()`.

@ghera I don't like this platform-specific approach but I can't think of a better solution at the moment, so I'm merging this patch... hopefully a temporal patch...

ghera:
@raysan5 I think there are two ways to solve the problem in the long run:

1 - reintroduce an internal header for these platform-dependent defines, with a name like `rplatform.h` that clearly describes its intent

2 - linker wrapping as suggested by @maiconpintoabreu. I actually wrote a patch with this approach and it's simpler than anticipated. It needs 2 modifications: 

I - adding
```
if (${PLATFORM} MATCHES "Android")
  target_link_options(raylib INTERFACE -Wl,--wrap=fopen)
endif()
```
in `src/CMakeLists.txt` 

II - adding  `__real_fopen` forward declaration + renaming `android_fopen` to `__wrap_fopen` in `src/platforms/rcore_android.c`.

As a plus, we could remove the `fopen` `#define` entirely. The only drawback is that it requires a linker option, but I don't think that's a real issue.

If you like option 2, I could submit another PR.

PR #5590, opened 2026-02-25
Fix reversed window blur/focus logic for web

aceiii:
On master with a web build, mouse input didn't seem like it was working. I was able to trace it to the blur/focus logic which seems to be reversed.

The `rcore_web_emscripten.c` file doesn't seem to be used, but contains the similarly reversed logic so I fixed that as well.


raysan5:
@aceiii Good catch! Thanks for the review!

PR #5592, opened 2026-02-25
[raylib.h] renamed SHADER_LOC_VERTEX_INSTANCETRANSFORMS

CrackedPixel:
This seems to have been renamed `SHADER_LOC_VERTEX_INSTANCETRANSFORMS` (plural) to `SHADER_LOC_VERTEX_INSTANCETRANSFORM` (singular)?

https://github.com/raysan5/raylib/blob/master/src/raylib.h#L817 (plural, error?)
https://github.com/raysan5/raylib/blob/master/src/rcore.c#L1308 (singular, new?)

if you want it the other way around i can change it and rename them all, too

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -2556,7 +2556,7 @@
           "description": "Shader location: matrix attribute: bone transforms (animation)"
         },
         {
-          "name": "SHADER_LOC_VERTEX_INSTANCETRANSFORMS",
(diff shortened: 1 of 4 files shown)


raysan5:

casperbear:
@CrackedPixel then 's' should've been removed in the comment too. There is only one transform per instance, unless it's meant that translate/rotate/scale are different transforms
SHADER_LOC_VERTEX_INSTANCETRANSFORM // Shader location: vertex attribute: instance transformS

PR #5594, opened 2026-02-26
[raudio] Fix a 64 bit to 32 bit int cast warning.

JeffM2501:
There is a 64 bit int that is assigned to a 32 bit int, and that generates a warning in MSVC. This PR adds a cast to fix it.


raysan5:

JeffM2501:
> It looks like `(size_t)` would be safer, based on the cast in the previous line, and assuming `bpf` is a relatively-small integer.   Perhaps the problem is with the size of the destination of that assignment?

Ray does not like to use size_t

orcmid:
I mentioned `size_t` because it is already used in a cast earlier in the very-same code.

I understand a warning is being avoided.  My only concern was whether or not the warning represented a serious case here, and that over-riding it is not solving the problem.

PR #5596, opened 2026-02-26
[rcore] Removing `#pragma message` and adding `TRACELOG` inside the function

maiconpintoabreu:
Making Warning message consistent and avoiding Warning every build by default.
Extra: Cosmetic refactor in `raymath`


maiconpintoabreu:
Left it as Draft as I will test again on Windows.

dreais:
Just pointing out the irony after this post: https://x.com/raysan5/status/2026929451938808072

Good luck!

raysan5:
@maiconpintoabreu thanks for the review! Maybe I'm further reviewing it to simplify it a bit... too many flags involved...

maiconpintoabreu:
@raysan5 maybe we could combine the warnings later and just say what it needs to run instead of a validation by "field".

raysan5:
> @raysan5 maybe we could combine the warnings later and just say what it needs to run instead of a validation by "field".

Yeah, that was my thought, it will be easier to read with same result, do you want to send another PR?

maiconpintoabreu:
@raysan5 sure I will do it.

PR #5598, opened 2026-02-26
[rcore] fix memory leak found by maiconpintoabreu

CrackedPixel:
move the calloc to inside the `#if`

Tested the `core/core_automation_events` with and without the flag enabled

EDIT: original discovery by  `QuantumKarl`


raysan5:

PR #5600, opened 2026-02-27
[rcore] Remove some tags to make it more readable 

maiconpintoabreu:
Removed check for `SUPPORT_FILEFORMAT_BMP` on Windows devices as it is guarantied.
Combining `SUPPORT_CLIPBOARD_IMAGE` and `SUPPORT_MODULE_RTEXTURES` as it is a hard requirement.


raysan5:
@maiconpintoabreu thanks! looks great!

PR #5601, opened 2026-02-27
[build] Review fullscreen toggle on `shell.html` on chromium based browsers

moaz-karam:
This PR allows the full-screen button to open on chromium based browsers

### DIFF
--- a/src/shell.html
+++ b/src/shell.html
@@ -179,7 +179,7 @@ jwE50AGjLCVuS8Yt4H7OgZLKK5EKOsLviEWJSL/+0uMi7gLUSBseYwqEbXvSHCec1CJvZPyHCmYQffaB
 
         <span id='controls'>
           <span><input type="button" value="📜 SOURCE CODE" onclick="location.href='https://github.com/raysan5/raylib';"/></span>
-          <span><input type="button" value="🖵 FULLSCREEN" onclick="Module.canvas.requestFullscreen(false, false)"></span>
+          <span><input type="button" value="🖵 FULLSCREEN" onclick="Module.canvas.requestFullscreen()"></span>
           <span><input type="button" id="btn-audio" value="🔇 MUTE" onclick="toggleAudio()"></span>
         </span>
 


maiconpintoabreu:
@moaz-karam `requestFullScreen` receives 2 args, you can see here: https://github.com/emscripten-core/emscripten/blob/af01558779231dcf3524438e904b688a5576432c/src/lib/libbrowser.js#L218
Maybe you are using a old version of `emscripten`.

moaz-karam:
> @moaz-karam `requestFullScreen` receives 2 args, you can see here: https://github.com/emscripten-core/emscripten/blob/af01558779231dcf3524438e904b688a5576432c/src/lib/libbrowser.js#L218 Maybe you are using a old version of `emscripten`.

I am using the latest version of the [emsdk](https://emscripten.org/docs/getting_started/downloads.html) and `requestFullscreen(false, false)` works well on Firefox, but fails on Chrome unless the args are removed.

maiconpintoabreu:
@moaz-karam I can replicate the issue as well now, never saw it before.

aceiii:
The current documentation shows that it's using `Module.requestFullscreen()` and NOT `Module.canvas.requestFullscreen()`: https://github.com/emscripten-core/emscripten/blob/main/html/shell_minimal.html

What you see [here](https://github.com/emscripten-core/emscripten/blob/af01558779231dcf3524438e904b688a5576432c/src/lib/libbrowser.js#L218) gets exported when you add `requestFullscreen` to the `EXPORTED_RUNTIME_METHODS` link options. I've tested this out on my own code and it works with both Firefox and Chrome when you call this instead one from Module instead of the one from Module.canvas.

Edit: Also noticed the hide/lock pointer didn't actually work before on Firefox but this fixes it.

raysan5:
Tested on Windows with Edge and Chrome browsers, the other options do not work:

 - `Module.canvas.requestFullscreen(false, false)` - FAIL
 - `Module.requestFullscreen(false, false)` - FAIL
 - `Module.requestFullscreen()` - FAIL

Tested with latest `Emscripten/emsdk 5.0.2`

PR #5602, opened 2026-02-27
[rtext] fix memory corruption and invalid size calculation in GenImageFontAtlas

konakona418:
i was just using `LoadFontEx` for font loading. and the program exited with segmentation fault. the font i'm using is `NotoSansJP`, a font for Japanese text, and what i'm trying to display is `1234567`

the backtrace tells that the `GenImageFontAtlas` caused this. when the font atlas exceeds its initial estimated size, the reallocation logic was using the old height to calculate the new buffer size and failing to update the atlas.data pointer after freeing it

so i wrote this patch which hopefully can fix the problem


raysan5:
@konakona418 oh! thanks for the review!

PR #5605, opened 2026-02-28
[rcore][android] Replace `android_fopen()` with linker `--wrap=fopen`

ghera:
### Context

This is a follow-up to #5589, which fixed `LoadMusicStream` failing on Android for OGG, WAV, and MP3 formats. That fix was accepted by @raysan5 as a **temporary patch** — adding a `#define fopen android_fopen` override directly inside `raudio.c` — with the hope of finding a cleaner solution.

As discussed in the comments of #5589, @maiconpintoabreu proposed this alternative approach: using the linker `--wrap=fopen` option to redirect all `fopen` calls at link time. This PR implements that approach.

### What changed

- **`src/CMakeLists.txt`**: Added `target_link_options(raylib INTERFACE -Wl,--wrap=fopen)` for Android builds, so the linker redirects all `fopen` calls to `__wrap_fopen` transparently.
- **`src/Makefile`**: Added `LDFLAGS += -Wl,--wrap=fopen`.
- **`src/platforms/rcore_android.c`**: Renamed `android_fopen` → `__wrap_fopen` and added a `__real_fopen` declaration. Removed all `#undef fopen` inside the function body. Also fixes a missing fallback in write mode (`'w'`): the old code only tried `internalDataPath/fileName`; now falls back to the literal path if that fails, allowing absolute paths to work correctly.
- **`src/raudio.c`**: Removed the platform-specific `#define fopen android_fopen` block added in the previous patch.

### Why this is better than the temporary patch

The macro approach only covered the source files that explicitly included the override. The linker `--wrap=fopen` covers **all** source files and third-party libraries simultaneously:

- `LoadMusicStream()` for OGG/WAV/MP3 — already fixed in via macro; now covered cleanly
- `LoadOBJ()` and `LoadMaterials()` for OBJ/MTL files — when loading the companion `.mtl` file, `tinyobj_loader_c.h` calls `fopen` directly (`tinyobj_parse_and_index_mtl_file`) and was **not covered** by the previous patch
- Any future file-loading code (new formats, new decoders) — automatically covered with no extra changes needed
- `__wrap_fopen` is cleaner than the old android_fopen without the `#undef` / `#define` trick.

### Testing

Tested on Android with:
- `LoadMusicStream` (read) for OGG, WAV, and MP3 — loading from APK assets via relative path
- C++ `iostream` (write) to external storage via absolute path — which in the Android NDK internally uses `fopen` — confirming third-party code is correctly intercepted

Absolute path write support also benefits from the new fallback added in `__wrap_fopen`: when `internalDataPath/fileName` fails, the function now retries with the path as-is, which was previously missing.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -607,6 +607,9 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_ANDROID)
     LDFLAGS += -Lsrc
     # Avoid unresolved symbol pointing to external main()
     LDFLAGS += -Wl,-undefined,dynamic_lookup
+    # Wrap fopen at link time so all code (including third-party libs) goes
+    # through __wrap_fopen, which handles Android APK asset loading
+    LDFLAGS += -Wl,--wrap=fopen
 endif
 
 # Define libraries required on linking: LDLIBS


maiconpintoabreu:
Hi @ghera, do not forget the Makefile if ray accepts it.

raysan5:
@ghera @maiconpintoabreu thanks for the improvement and review, I like the code simplification but I'm a but scary about moving that feature control to the build system instead of keeping it inside code, what about `build.zig`? what about other custom build systems on user side...

I'm merging it and let's see if it generates further issues on that regards... I also think it should be better documented in the code with some `// WARNING: ...` messages.

PR #5609, opened 2026-03-01
[rlgl] fix vertex calculation precision in vertex shader for depth buffer

arlez80:
Insufficient precision in vertex calculations causes depth calculation issues when 3D rendering, resulting in Z-fighting or clipping error.
I fixed specify highp to ensure sufficient precision.

The screen flickers with every camera movement, I'm took two screenshots of the issues occurring.
environment: RG40XXH (Allwinner h700; Quad-core ARM Cortex-A53; G31 MP2 GPU)

error-screen-2

the ground is clipped.

error-screen-3

the character's some polygon is z-fighting. the ground depth is weird also.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -4994,7 +4994,7 @@ static void rlLoadShaderDefault(void)
 #if defined(GRAPHICS_API_OPENGL_ES3)
     "#version 300 es                    \n"
     "precision mediump float;           \n"     // Precision required for OpenGL ES3 (WebGL 2) (on some browsers)
-    "in vec3 vertexPosition;            \n"
+    "in highp vec3 vertexPosition;      \n"
     "in vec2 vertexTexCoord;            \n"
     "in vec4 vertexColor;               \n"
     "out vec2 fragTexCoord;             \n"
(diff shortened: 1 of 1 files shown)


maiconpintoabreu:
Hi @arlez80 the default is already highp no?
Reference: https://www.khronos.org/files/opengles20-reference-card.pdf

arlez80:
@maiconpintoabreu yes the glsl default is highp.
but the default precision set mediump at  [here](https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L4993) or [here](https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L5001).
therefore, I have set highp for the vertex variable and model-view-projection matrix.

raysan5:
@arlez80 Afair, not all platforms support `highp`, specially for OpenGL ES 2.0, also specifying the precision at beginning should be enough. I prefer to avoid this change.

arlez80:
@raysan5 
I have a proposal, how about separate them by macro?

like:
```C

// ...
#if defined(GRAPHICS_API_USE_HIGHP_VERTEX)
"in highp vec3 vertexPosition;      \n"
#else
"in vec3 vertexPosition;            \n"
#endif
// ...
```

PR #5611, opened 2026-03-02
[arm64ec] Support for arm64EC on windows.

aalekhm:
Added changes for build for arm64EC.


raysan5:
@aalekhm I'm afraid those changes should be sent to original repos for those libraries, otherwise they will be ovewrite when updating the libraries.

PR #5613, opened 2026-03-02
[raymath] Add C++ `MatrixUnit` constant and `MatrixMultiplyValue` function

lamweilun:


lamweilun:
@raysan5 
Thank you for the feedback previously. This PR is just to add in an identity matrix constant for C++ and a matrix scalar multiplication function.

I did want to call it `MatrixIdentity` but there's already a function with that name so I just use `MatrixUnit` instead. Same for `MatrixMultiplyValue`, in theory it should have been `MatrixScale` since it will follow the same convention like `Vector3Scale` for example. But of course there's already a `MatrixScale` function that returns a scale matrix, so I just called it `MatrixMultiplyValue`.

Let me know if there's any issues or if any of the naming should change. Thanks!

raysan5:
@lamweilun thanks for the addition, it looks good.

Issue #5612, opened 2026-03-02
[rexm] Not able to run `rexm validate`


maiconpintoabreu:
When I try to run `rexm validate`, it breaks the `example_list.txt`

[Action with the issue](https://github.com/raysan5/raylib/actions/runs/22574159087/job/65389161525)
```
gcc -o ./rexm  rexm.o -std=c99 -Wall -Wno-missing-braces -Wno-unused-value -Wno-pointer-sign -D_DEFAULT_SOURCE  -s -O2 -I. -Iexternal -I../../src -L. -L../../src -lraylib -lGL -lm -lpthread -ldl -lrt -lX11 -DPLATFORM_DESKTOP
make[1]: Leaving directory '/home/runner/work/raylib/raylib/tools/rexm'
*** buffer overflow detected ***: terminated
/home/runner/work/_temp/6585a7a7-94b1-4128-858b-9d49e3ad3a3d.sh: line 16:  3154 Aborted                 (core dumped) ./rexm validate
``` 

[Screencast from 2026-02-27 22-46-37.webm](https://github.com/user-attachments/assets/53004142-9e91-4269-ae66-7c9df84c57c0)

raysan5:
@maiconpintoabreu I saw this issue, no idea why does it happen, definitely something to investigate

UPDATE: I don't get this issue on Windows...

maiconpintoabreu:
> [@maiconpintoabreu](https://github.com/maiconpintoabreu) I saw this issue, no idea why does it happen, definitely something to investigate
> 
> UPDATE: I don't get this issue on Windows...

Looks like the issue is on this line: https://github.com/raysan5/raylib/blob/e1113c88337b46cb01f702ff7b0b2794261ebf8e/tools/rexm/rexm.c#L1043C60-L1043C120

It looks for "text"(next category of texture) and gets the index of 
`core;core_render_(text)(Adding here the new texture example)ure;★☆☆☆;5.6-dev;5.6-dev;2025;2025;"Ramon Santamaria";@raysan5`

I don't know how it works on Windows.

raysan5:
@maiconpintoabreu thanks for the further investigation! I will review it!

PR #5616, opened 2026-03-03
[tools/rexm] Update nextCatIndex

CrackedPixel:
taken from https://github.com/raysan5/raylib/issues/5612

~~i am less familiar with this tool so please advise :pray: i don't know what the purpose of the `+1` was~~

```
category list:
  [0]: core
  [1]: shapes
  [2]: textures  <-- nextCatIndex
  [3]: text      <-- +1
  [4]: models
  [5]: shaders
```

when searching for "`textures`" it was adding +1 and trying to affect "`text`", which messed up the generated output

it also crashes when running `rexm validate` (tested on linux) with current `master` with a segfault

with the PR changes it detects all categories perfectly and the crash doesn't happen anymore

### DIFF
--- a/tools/rexm/rexm.c
+++ b/tools/rexm/rexm.c
@@ -1034,13 +1034,13 @@ int main(int argc, char *argv[])
 
                     if (catIndex > -1)
                     {
-                        int nextCatIndex = catIndex;
+                        int nextCatIndex = catIndex + 1;
                         if (nextCatIndex > (REXM_MAX_EXAMPLE_CATEGORIES - 1)) nextCatIndex = -1; // EOF
 
                         // Find position to add new example on list, just before the following category
                         // Category order: core, shapes, textures, text, models, shaders, audio
                         int exListNextCatIndex = -1;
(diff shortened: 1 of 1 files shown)


maiconpintoabreu:
Hi @CrackedPixel, the propose of the +1 is to find the next category and add the new example on the last position of the category pushing the next one down 1.

Ps. This fix will work untill you need to add a text example

tools/rexm/rexm.c
maiconpintoabreu:
```suggestion
                        if (nextCatIndex != -1) exListNextCatIndex = TextFindIndex(exList, TextFormat("\n%s", exCategories[nextCatIndex])) + 1;
```

Check if the category is on the beginning of the line and adding one to fix the index.

CrackedPixel: this seems to also have worked and added the clipboard example correctly!

raysan5:
@CrackedPixel @maiconpintoabreu thanks for finding and fixing this issue!

PR #5617, opened 2026-03-03
Typo in find_dependency for SDL2

m039:
I tried to build raylib for SDL2 and found this typo.


raysan5:
@m039 thanks for the review!

PR #5620, opened 2026-03-04
Fix VSCode template Makefile for PLATFORM_WEB

dodome2k6:
This fixes the VSCode template Makefile for PLATFORM_WEB when using the Windows Installer package.

The installer provides libraylib.web.a for web builds, but the template was linking to libraylib.a, causing a link error.

Tested locally with the Windows Installer package.


raysan5:
@takenoko-pm thanks for the review!

Issue #4954, opened 2025-05-27
[rcore][glfw] Crash when plugging in a controller


CowCowFuture:
### Issue description
If I start a raylib window using the GLFW backend with no gamepads plugged in, then plug in a gamepad, I get a segfault.
Using raylib-quickstart found in the wiki it will crash no matter what controller I plug in, and using raylib examples compiled from the source in Github it will freeze if I plug in a DualSense controller and uncap the FPS when I plug in a Nintendo Switch Pro controller. This issue does not occur using the latest release v5.5, only on the latest from Github.

### Environment
`uname -a`
```
Linux magnesium 6.14.6-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Fri, 09 May 2025 17:35:59 +0000 x86_64 GNU/Linux
```
`glxinfo | grep "OpenGL version"`
```
OpenGL version string: 4.6.0 NVIDIA 570.144
```
### Issue Backtrace
```txt

➜  raylib-quickstart git:(main) ✗ gdb ./bin/Debug/raylib-quickstart
GNU gdb (GDB) 16.3
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./bin/Debug/raylib-quickstart...
(gdb) r
Starting program: /tmp/raylib-quickstart/bin/Debug/raylib-quickstart

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.archlinux.org>
Enable debuginfod for this session? (y or [n]) n
Debuginfod has been disabled.
To make this setting permanent, add 'set debuginfod enabled off' to .gdbinit.
Function(s) ^std::(move|forward|as_const|(__)?addressof) will be skipped when stepping.
Function(s) ^std::(shared|unique)_ptr<.*>::(get|operator) will be skipped when stepping.
[...]

PR #5621, opened 2026-03-04
[backend/GLFW] Added bounds check

CrackedPixel:
I ran into an issue where my apps were crashing if you connect a gamepad during runtime. Only on linux. 

I found this with the same issue, and it showed a commit where it "started": https://github.com/raysan5/raylib/issues/4954

in that issue, [this commit](https://github.com/raysan5/raylib/commit/74256943a41166b7f9b811727eb9d9fb03d83e25) was noted as the culprit.. and if you set the `MAX_GAMEPAD_NAME_LENGTH` back to 64 it masks the issue

If you add a debug log to this function, you will see the `jid` is beyond the array index. im not sure how glfw decides the IDs but i believe its some platform specific code they have that makes it not happen on windows? or maybe with the alignment its just been lucky. 

This must be why it has different behavior on different versions of raylib or different runs (RLGL has null context, fps cap changes, etc), because **its overwriting random memory with the overflow**

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -2178,23 +2178,19 @@ static void CursorEnterCallback(GLFWwindow *window, int enter)
 // GLFW3: Joystick connected/disconnected callback
 static void JoystickCallback(int jid, int event)
 {
-    if (jid >= MAX_GAMEPADS)
+    if (jid < MAX_GAMEPADS)
     {
-        // WARNING: If jid is higher than maximum supported joysticks, just return. This
-        // prevents an out-of-bounds crash on linux when connecting a gamepad to a running app
-        return;
-    }
-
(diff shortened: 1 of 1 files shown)


src/platforms/rcore_desktop_glfw.c
```diff
@@ -2178,6 +2178,12 @@ static void CursorEnterCallback(GLFWwindow *window, int enter)
 // GLFW3: Joystick connected/disconnected callback
 static void JoystickCallback(int jid, int event)
 {
+    if (jid >= MAX_GAMEPADS) {
```
JeffM2501: I recommend you use the same brace structure as the rest of raylib.

CrackedPixel: done, good catch

raysan5:
@CrackedPixel Nice! Thanks for the review!

PR #5623, opened 2026-03-04
[rtext] Add `MeasureTextCodepoints()` for direct measurement of codepoints

konakona418:
currently, raylib has `DrawTextCodepoints` function for rendering a sequence of codepoints but lacks a corresponding measurement function.

i'm working on a program which needs to handle some CJK text. first i converted the UTF8 sequence to unicode codepoints by `LoadCodepoints` function, and freed the original buffer. However later in that program i have to measure the length of a certain slice of the text. so i resorted to converting it back to UTF8 and use `MeasureTextEx` function, which does not require too much extra work but is less efficient.

so i wrote these two functions, which are basically analogous to `MeasureText` and `MeasureTextEx` but takes unicode codepoints as arguments.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1509,8 +1509,7 @@ RLAPI void DrawTextCodepoints(Font font, const int *codepoints, int codepointCou
 RLAPI void SetTextLineSpacing(int spacing);                                                 // Set vertical line spacing when drawing with line-breaks
 RLAPI int MeasureText(const char *text, int fontSize);                                      // Measure string width for default font
 RLAPI Vector2 MeasureTextEx(Font font, const char *text, float fontSize, float spacing);    // Measure string size for Font
-RLAPI int MeasureCodepoints(const int *codepoints, int length, int fontSize);               // Measure string width for an existing array of codepoints for default font
-RLAPI Vector2 MeasureCodepointsEx(Font font, const int *codepoints, int length, float fontSize, float spacing); // Measure string size for an existing array of codepoints for Font
+RLAPI Vector2 MeasureTextCodepoints(Font font, const int *codepoints, int length, float fontSize, float spacing); // Measure string size for an existing array of codepoints for Font
 RLAPI int GetGlyphIndex(Font font, int codepoint);                                          // Get glyph index position in font for a codepoint (unicode character), fallback to '?' if not found
 RLAPI GlyphInfo GetGlyphInfo(Font font, int codepoint);                                     // Get glyph font info data for a codepoint (unicode character), fallback to '?' if not found
 RLAPI Rectangle GetGlyphAtlasRec(Font font, int codepoint);                                 // Get glyph rectangle in font atlas for a codepoint (unicode character), fallback to '?' if not found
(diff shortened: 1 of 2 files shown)


src/raylib.h
```diff
@@ -1509,6 +1509,8 @@ RLAPI void DrawTextCodepoints(Font font, const int *codepoints, int codepointCou
 RLAPI void SetTextLineSpacing(int spacing);                                                 // Set vertical line spacing when drawing with line-breaks
 RLAPI int MeasureText(const char *text, int fontSize);                                      // Measure string width for default font
 RLAPI Vector2 MeasureTextEx(Font font, const char *text, float fontSize, float spacing);    // Measure string size for Font
+RLAPI int MeasureCodepoints(const int *codepoints, int length, int fontSize);               // Measure string width for an existing array of codepoints for default font
```
raysan5: Codepoints is intended for advance users, I don't think a default-font version is needed, I prefer to avoid it.

src/raylib.h
```diff
@@ -1509,6 +1509,8 @@ RLAPI void DrawTextCodepoints(Font font, const int *codepoints, int codepointCou
...
+RLAPI Vector2 MeasureCodepointsEx(Font font, const int *codepoints, int length, float fontSize, float spacing); // Measure string size for an existing array of codepoints for Font
```
raysan5: Please, can you rename it to `MeasureTextCodepoints()` for consistency with `DrawTextCodepoints()`

src/rtext.c
```diff
@@ -1388,6 +1388,75 @@ Vector2 MeasureTextEx(Font font, const char *text, float fontSize, float spacing
     return textSize;
 }
 
+// Measure string width for an existing array of codepoints for default font
+int MeasureCodepoints(const int *codepoints, int length, int fontSize) {
```
raysan5: I prefer to avoid adding this one

src/rtext.c
```diff
@@ -1388,6 +1388,75 @@ Vector2 MeasureTextEx(Font font, const char *text, float fontSize, float spacing
...
+        int spacing = fontSize/defaultFontSize;
+
+        textSize = MeasureCodepointsEx(GetFontDefault(), codepoints, length, fontSize, spacing);
+    }
+
+    return (int)textSize.x;
+}
+
+// Measure string size for an existing array of codepoints for Font
+Vector2 MeasureCodepointsEx(Font font, const int *codepoints, int length, float fontSize, float spacing) {
```
raysan5: Please, could you review formating to follow [raylib code conventions](https://github.com/raysan5/raylib/blob/master/CONVENTIONS.md)?

konakona418:
@raysan5 thank you for the quick and detailed code review!

i just removed the function `MeasureCodepoints()` which uses the default font implicitly, renamed `MeasureCodepointsEx()` to `MeasureTextCodepoints()` and applied the formatting requirements.

please let me know if any further changes needs to be made.

raysan5:
@konakona418 Looks good, thanks for the addition

PR #5624, opened 2026-03-04
[rcore][android] Document --wrap=fopen linker flag requirements per build system

ghera:
Follow-up to #5605, addressing @raysan5's suggestion for `// WARNING:` comments in code.

Adds a comment block before `__real_fopen`/`__wrap_fopen` explaining how and where `-Wl,--wrap=fopen` must be applied for each supported build system, including the behavioral difference between CMake (flag propagates to final app link) and Make (SHARED) (wraps libraylib.so only).

Tested on CMake and Make (both SHARED and STATIC). For build.zig, several approaches were attempted but none are available in the std.Build API required by raylib (0.16.0-dev.2349+); this will probably change in the near future, but for now the flag must be passed externally in the final link step.

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -275,6 +275,22 @@ static int android_write(void *cookie, const char *buf, int size);
 static fpos_t android_seek(void *cookie, fpos_t offset, int whence);
 static int android_close(void *cookie);
 
+// WARNING: fopen() calls are intercepted via linker flag -Wl,--wrap=fopen: the linker renames
+// the original fopen -> __real_fopen and redirects all call sites to __wrap_fopen
+// The flag MUST be applied at every final link step that needs wrapping,
+// it has no effect when only building a static archive (.a)
+//
+//         CMake: no action required, raylib's CMakeLists.txt already sets 
+//                target_link_options(raylib INTERFACE -Wl,--wrap=fopen) which propagates to
+//                the final app link, wrapping app code and all static (.a) dependencies too
+// Make (SHARED): no action required for raylib itself, src/Makefile already sets 
+//                LDFLAGS += -Wl,--wrap=fopen wrapping fopen inside libraylib.so only;
+//                app code and static (.a) dependencies are NOT wrapped unless -Wl,--wrap=fopen
+//                is also added to the final app link step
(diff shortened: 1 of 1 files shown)


maiconpintoabreu:
Hi @ghera for zig, can you not just add on the list of flags?
```
    try raylib_flags_arr.appendSlice(
        b.allocator,
        &[_][]const u8{
            "-std=gnu99",
            "-D_GNU_SOURCE",
                       <--------------------------------------------------------- Here
            "-DGL_SILENCE_DEPRECATION=199309L",
            "-fno-sanitize=undefined", // https://github.com/raysan5/raylib/issues/3674
        },
    );
```
I do not have the v16 yet so I cannot test it sorry.

ghera:
@maiconpintoabreu No worries. I hadn’t tried that part earlier because I assumed those were C compiler flags rather than linker flags. I tried using `.addLinkerArgs` /  `.addLinkerArg`, but none of them seem to work with Zig 0.16.

When I get a chance, I’ll try your suggestion and put together a more detailed report covering the different build scenarios.

raysan5:
@ghera thanks for the improvement! Definitely helpful for users dealing with custom build systems!

ghera:
Hi @maiconpintoabreu, I ran the full Android checks for build.zig, and adding `-Wl,--wrap=fopen` to `raylib_flags_arr` gives this for `libraylib.so`:

```
T __wrap_fopen
U __real_fopen
U fopen@LIBC
```

while the expected pattern should be `T __wrap_fopen` + `U fopen@LIBC` (without `U __real_fopen`).

So, based on the symbol evidence, it looks like `--wrap=fopen` is not being applied in Zig, even though it appears in the command line. Practically, it behaves as if that option is ineffective (or applied in the wrong phase) for this workflow.

For static, it remains true that wrapping must be activated at the consumer final link step.

Because of that, the Zig warning seems correct: `--wrap` should currently be treated as a final-link-step requirement (until `std.Build` provides a more suitable linker hook / API).

Thanks to your suggestion and the resulting tests, I also found a small issue in the standalone CMake shared build, and I’ll address it in a separate PR. Building `libraylib.so` standalone is a relatively uncommon workflow for most Android users though.

PR #5625, opened 2026-03-04
[cmake][X11] Add x11 libraries for raylib as glfw make private

maiconpintoabreu:
Initial log: INFO: PLATFORM: DESKTOP (GLFW - X11): Initialized successfully
Quick Check
```
#if defined(_GLFW_X11)
    TraceLog(LOG_INFO, "BACKEND: X11");
#elif defined(_GLFW_WAYLAND)
    TraceLog(LOG_INFO, "BACKEND: Wayland");
#else
    TraceLog(LOG_WARNING, "BACKEND: Unknown");
#endif
#if defined(__linux__)
    TRACELOG(LOG_WARNING, "On LINUX");
#endif
```

CMAKE  Result
```
WARNING: BACKEND: Unknown
WARNING: On LINUX
```

Make Result

```
INFO: BACKEND: X11
WARNING: On LINUX
 ```
 
 The issue is because on Cmake the `_GLFW_X11` is defined as PRIVATE and not readeable outside of the glfw library

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,6 +117,21 @@ target_include_directories(raylib
 
 # Copy the header files to the build directory for convenience
 file(COPY ${raylib_public_headers} DESTINATION "include")
+if (PLATFORM STREQUAL "Desktop")
+    if (UNIX AND NOT APPLE)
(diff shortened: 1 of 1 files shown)


raysan5:

PR #5628, opened 2026-03-07
[rlgl] Added rlLoadTextureCubemapDepth

R2turnTrue:
Adds a helper function to create depth cubemap textures in rlgl.

This allows easier setup of cubemap depth textures for use cases such as point light shadow mapping. The function allocates a cubemap with depth format and configures the required texture parameters.

Supported across OpenGL 3.3, OpenGL ES 3.0 and OpenGL ES 2.0 (with GL_OES_depth_texture).


raysan5:
@R2turnTrue thanks but I prefer to avoid adding this new function. rlgl does not pretend to be a complete wrapper of OpenGL and this function seems quite specific and advance for a certain functionality that most users don't need. It adds more code to maintain that I prefer to avoid.

R2turnTrue:
> @R2turnTrue thanks but I prefer to avoid adding this new function. rlgl does not pretend to be a complete wrapper of OpenGL and this function seems quite specific and advance for a certain functionality that most users don't need. It adds more code to maintain that I prefer to avoid.

@raysan5 Then how about adding `RL_PIXELFORMAT_UNCOMPRESSED_DEPTH` to enum `rlPixelFormat`?

PR #5629, opened 2026-03-07
Fix memory leak in LoadGLTF when model has no bones

victorberdugo1:
Fix memory leak in LoadGLTF for models without skeletal animation

`currentPose` and `boneMatrices` are allocated but never freed when
boneCount == 0, causing an 18-byte leak detected with LeakSanitizer on Linux.

Also fixes unreachable code and missing cleanup in the Draco early-exit path.


raysan5:
@victorberdugo1 thanks for the review

PR #5630, opened 2026-03-07
[rlgl] add `RL_PIXELFORMAT_UNCOMPRESSED_DEPTH` to enum `rlPixelFormat`

R2turnTrue:
This PR adds `RL_PIXELFORMAT_UNCOMPRESSED_DEPTH` pixel format to rlgl.

Depth textures were already supported through `rlLoadTextureDepth`, but there was no explicit pixel format defined in `rlPixelFormat` for them.
This addition allows depth textures (including depth cubemaps) to be created through `rlLoadTexture` or `rlLoadTextureCubemap` instead of requiring manual OpenGL setup.

This is really useful for features such as point light shadow maps using depth cubemaps, depth-based post processing effects.

The internal format is selected automatically based on the supported depth precision and graphics API.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -461,7 +461,7 @@ typedef enum {
     RL_PIXELFORMAT_UNCOMPRESSED_R16,               // 16 bpp (1 channel - half float)
     RL_PIXELFORMAT_UNCOMPRESSED_R16G16B16,         // 16*3 bpp (3 channels - half float)
     RL_PIXELFORMAT_UNCOMPRESSED_R16G16B16A16,      // 16*4 bpp (4 channels - half float)
-    RL_PIXELFORMAT_UNCOMPRESSED_DEPTH,             // Depth component (24 or 16 bit, depending on OpenGL version)
+    RL_PIXELFORMAT_UNCOMPRESSED_DEPTH,             // Depth component (32 or 24 or 16 bit, depending on OpenGL version)
     RL_PIXELFORMAT_COMPRESSED_DXT1_RGB,            // 4 bpp (no alpha)
     RL_PIXELFORMAT_COMPRESSED_DXT1_RGBA,           // 4 bpp (1 bit alpha)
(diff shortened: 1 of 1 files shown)


src/rlgl.h
```diff
@@ -3630,6 +3631,36 @@ void rlGetGlTextureFormats(int format, unsigned int *glInternalFormat, unsigned
         #endif
         #endif
         #endif
+        case RL_PIXELFORMAT_UNCOMPRESSED_DEPTH:
+            if (RLGL.ExtSupported.texDepth)
+            {
+                *glFormat = GL_DEPTH_COMPONENT;
+
+                if (RLGL.ExtSupported.texDepthWebGL)
```
raysan5, line 3639: Logic here could not be correct, considering nuances when defining `internalFormat` for the [different OpenGL ES versions](https://github.com/raysan5/raylib/blob/master/src/rlgl.h#L3428).

src/rlgl.h
```diff
@@ -461,6 +461,7 @@ typedef enum {
     RL_PIXELFORMAT_UNCOMPRESSED_R16,               // 16 bpp (1 channel - half float)
     RL_PIXELFORMAT_UNCOMPRESSED_R16G16B16,         // 16*3 bpp (3 channels - half float)
     RL_PIXELFORMAT_UNCOMPRESSED_R16G16B16A16,      // 16*4 bpp (4 channels - half float)
+    RL_PIXELFORMAT_UNCOMPRESSED_DEPTH,             // Depth component (32 or 24 or 16 bit, depending on OpenGL version)
```
raysan5, line 464: This enum is aligned with raylib textures [`PixelFormat`](https://github.com/raysan5/raylib/blob/master/src/raylib.h#L850), adding this value here could generate missalignments issues.

raysan5:
I see the use case but I think the addition cost is higher than the real usefulness for most users of this addition. It woudl also imply a redesign (and further testing and review) of `PixelFormat` structure to accomodate this new flag... I'm afraid I'm prefer to avoid this change for now.

PR #5633, opened 2026-03-10
[raylib.h] fix audio pan comment

CrackedPixel:
It incorrectly says 0.5f is the center, but the range is -1.0f to 1.0 with 0.0 being center. 

found by user `a.b.c.d.a.b.c.d` on discord


raysan5:

PR #5634, opened 2026-03-10
[rmodels] Implement loading of mesh names from glTF/IQM/OBJ.

arlez80:
This PR aims to implement loading of mesh names from glTF/IQM/OBJ and add a `name` member to the `Model` structure.

In glTF, node names are used as mesh names because Blender object names are written to the nodes.

For IQM, the mesh name is copied from the already loaded mesh name.

For OBJ, the mesh name is copied from `objShapes[*].name`.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -341,6 +341,7 @@ typedef struct Camera2D {
 
 // Mesh, vertex data and vao/vbo
 typedef struct Mesh {
+    char name[32];          // Mesh name
     int vertexCount;        // Number of vertices stored in arrays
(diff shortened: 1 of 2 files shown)


arlez80:
These image aren't part of this PR currently, would it be acceptable to add a sample for displaying mesh names to `models_loading.c`?
The reason I asked was that I was wondering if it would be better to put the sample in a separate file.

image
image

raysan5:
@arlez80 I'm afraid it has already been proposed in the past and there is no intention for now to add those 32bytes of additional information to Mesh structure. Most users do not really need that information.

PR #5635, opened 2026-03-10
[examples][models] Change `GetRenderWidth()` to `GetScreenWidth()` in models_basic_voxel.c

SardineMilk:
I'm pretty sure that `GetRenderWidth()` is the wrong function for this situation.
`GetScreenHeight()` is used in the same line, so I changed it to `GetScreenWidth()` for consistency.

### DIFF
--- a/examples/models/models_basic_voxel.c
+++ b/examples/models/models_basic_voxel.c
@@ -137,7 +137,7 @@ int main(void)
             EndMode3D();
             
             // Draw reference point for raycasting to delete blocks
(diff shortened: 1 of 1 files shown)


raysan5:
@SardineMilk Thanks for the review! Actually I think there was some issue with raycasting... not sure now...

SardineMilk:
> @SardineMilk Thanks for the review! Actually I think there was some issue with raycasting... not sure now...

I took another look at it, and there is an error. I'll make an issue for it.

PR #5637, opened 2026-03-11
[examples] Update audio_raw_stream.c

dan-hoang:
One of the first examples I played with when I found out about Raylib was audio_raw_stream.c.

I love how easy it is to synthesize stuff in Raylib, but the example did puzzle me for a bit. The big issues:

1. There are two ways to upload data, and both are put in this one example. The reader must sort out what code is necessary to do things the way they want.
2. The mouse controls are unintuitive.

To address (1), I propose to split this example in two: `audio_raw_stream.c` and `audio_raw_stream_callback.c`. To address (2), I propose to use the keyboard instead of the mouse.

This commit updates `audio_raw_stream.c` to

1. Only use the old API (i.e., `IsAudioStreamProcessed` and `UpdateAudioStream`). I feel that the old API is more in the spirit of Raylib and so should be the one that takes a plainer name.
2. Use the keyboard instead of the mouse.
3. Remove the use of `malloc`, `free`, and `memcpy`. This is to reduce the number of headers included.
4. Show a more dynamic visualization.

### DIFF
--- a/examples/audio/audio_raw_stream.c
+++ b/examples/audio/audio_raw_stream.c
@@ -32,7 +32,6 @@ int main(void)
     const int screenHeight = 450;
 
     InitWindow(screenWidth, screenHeight, "raylib [audio] example - raw stream");
-    SetTargetFPS(30);
 
     InitAudioDevice();
 
@@ -51,6 +50,7 @@ int main(void)
     int sineIndex = 0;
     double sineStartTime = 0.0;
 
+    SetTargetFPS(30);
     //--------------------------------------------------------------------------------------
 
     // Main game loop


examples/audio/audio_raw_stream.c
```diff
@@ -1,56 +1,25 @@
...
-*
-*   Example created by Ramon Santamaria (@raysan5) and reviewed by James Hofmann (@triplefox)
-*
-*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
-*   BSD-like license that allows static linking with closed source software
-*
-*   Copyright (c) 2015-2025 Ramon Santamaria (@raysan5) and James Hofmann (@triplefox)
-*
-********************************************************************************************/
+ *
```
raysan5: Why there is a space in from ot the comment, please remove it

examples/audio_raw_stream.c
```diff
@@ -0,0 +1,146 @@
+/*******************************************************************************************
```
raysan5: It seems this file shouldn't be here.

examples/audio/audio_raw_stream.c
```diff
@@ -63,147 +32,110 @@ int main(void)
     const int screenHeight = 450;
 
     InitWindow(screenWidth, screenHeight, "raylib [audio] example - raw stream");
+    SetTargetFPS(30);
```
CrackedPixel: iirc ray likes this to be where it was, just before `while (!WindowShouldClose())` (above the `//------` line)

examples/audio_raw_stream.c
dan-hoang: Sorry, I thought I had already deleted it. Will do.

examples/audio/audio_raw_stream.c
dan-hoang: Sorry, I didn't even notice. It's because I pasted it in my editor, which automatically formatted it. I deleted the spaces.

examples/audio/audio_raw_stream.c
dan-hoang: Gotcha, I put it back.

raysan5:
@dan-hoang thanks for the simplification and @CrackedPixel thanks for the review!

PR #5638, opened 2026-03-11
[examples] Add `audio_stream_callback`

dan-hoang:
This is a sister pull request to [https://github.com/raysan5/raylib/pull/5637#issue-4054960372]().

This commit adds `audio_raw_stream_callback.c`, which is the same as `audio_raw_stream.c` in the pull request above, but uses the new API (i.e., `SetAudioStreamCallback`) instead of the old API (i.e., `IsAudioStreamProcessed` and `UpdateAudioStream`).

### DIFF
--- a/examples/audio/audio_stream_callback.c
+++ b/examples/audio/audio_stream_callback.c
@@ -182,50 +182,20 @@ int main(void)
 
         if (IsKeyPressed(KEY_LEFT))
         {
-            switch (waveType)
-            {
-                case SINE:
-                {
-                    waveType = SAWTOOTH;
-                } break;
-                case SQUARE:
-                {
-                    waveType = SINE;
-                } break;
-                case TRIANGLE:
-                {
-                    waveType = SQUARE;
-                } break;
-                case SAWTOOTH:
-                {
-                    waveType = TRIANGLE;
-                } break;
-            }
+            if (waveType == SINE) waveType = SAWTOOTH;
+            else if (waveType == SQUARE) waveType = SINE;
+            else if (waveType == TRIANGLE) waveType = SQUARE;
+            else waveType = TRIANGLE;
 
             SetAudioStreamCallback(stream, waveCallbacks[waveType]);
         }
 
         if (IsKeyPressed(KEY_RIGHT))
         {
-            switch (waveType)
-            {
-                case SINE:
-                {
-                    waveType = SQUARE;
-                } break;
-                case SQUARE:
-                {
-                    waveType = TRIANGLE;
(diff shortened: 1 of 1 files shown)


examples/audio/audio_raw_stream_callback.c
```diff
@@ -0,0 +1,148 @@
...
+ *
+ *   Example created by Ramon Santamaria (@raysan5) and reviewed by James Hofmann (@triplefox)
+ *
+ *   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+ *   BSD-like license that allows static linking with closed source software
+ *
+ *   Copyright (c) 2015-2025 Ramon Santamaria (@raysan5) and James Hofmann (@triplefox)
+ *
+ ********************************************************************************************/
+
```
CrackedPixel: This one also has spaces before the `*` lines that ray mentioned he would like removed (in the other PR)

examples/audio/audio_raw_stream_callback.c
```diff
@@ -0,0 +1,148 @@
...
+//------------------------------------------------------------------------------------
+int main(void)
+{
+    // Initialization
+    //--------------------------------------------------------------------------------------
+    const int screenWidth = 800;
+    const int screenHeight = 450;
+
+    InitWindow(screenWidth, screenHeight, "raylib [audio] example - raw stream callback");
+    SetTargetFPS(30);
```
CrackedPixel: iirc ray likes this to be just before `while (!WindowShouldClose())` (above the `//------` line)

raysan5:
@dan-hoang @CrackedPixel I think this example can be redesigned to avoid being mostly a copy of the other one, maybe also renamed to `audio_stream_callback` and do some processing in the callback? Ideas for a more interesting and appealing example to illustrate callback functionality?

dan-hoang:
> @dan-hoang @CrackedPixel I think this example can be redesigned to avoid being mostly a copy of the other one, maybe also renamed to `audio_stream_callback` and do some processing in the callback? Ideas for a more interesting and appealing example to illustrate callback functionality?

I'm not an audio guy, so I don't have any good ideas. For now, I just let the user choose between four different waves (sine, square, triangle, and sawtooth), with each wave generated by a different callback. When the user changes the wave they want, the stream's callback is changed appropriately.

I'm thinking that you would do something like this if you have a pipeline of effects set up (like in the stream effects example) and want to swap out the base signal.

raysan5:
@dan-hoang example looks good, thanks for the improvement

PR #5640, opened 2026-03-11
[examples/makefile] remove duplicate line in makefile

CrackedPixel:
`RAYLIB_RELEASE_PATH` is already set to `RAYLIB_PATH/src` so this just adds a duplicate entry. im guessing this was done at some point when they were different? or maybe it was renamed etc?


raysan5:
@CrackedPixel thanks for the review, actually I think there is room for improvement on Makefiles, they have changed a lot in the last few years...

PR #5641, opened 2026-03-11
[examples/makefile] remove path override/change for linux

CrackedPixel:
I noticed this `TODO`. on linux, it converts a relative path to an absolute path. both are valid when running from the `examples` folder, so i guess it being "needed" just depends if you want to guarantee absolute paths for symlink build scripts or ones that `cd` around? maybe its fine to keep it? :man_shrugging: 

EDIT: 
actually, this fixed an issue for me... i had a space in a filename and it adds the path without quotes, breaking the command. relative path does not break. so thats a win :joy_cat:

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -155,15 +155,6 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_DRM)
     endif
 endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@CrackedPixel thanks for the review! I like simplification!

PR #5643, opened 2026-03-11
[examples] Fix raycasting logic in models_basic_voxel.c

SardineMilk:
The original logic iterated through the world and broke at the first found voxel that intersected the ray. This broke in some cases, removing a voxel out of view. I changed the algorithm to track the closest found voxel, and remove it at the end of the loop if one was found.

### DIFF
--- a/examples/models/models_basic_voxel.c
+++ b/examples/models/models_basic_voxel.c
@@ -72,6 +72,7 @@ int main(void)
         UpdateCamera(&camera, CAMERA_FIRST_PERSON);
 
         // Handle voxel removal with mouse click
+        // This method is quite inefficient. Ray marching through the voxel grid using DDA would be faster, but more complex.
         if (IsMouseButtonPressed(MOUSE_LEFT_BUTTON))
         {
             // Cast a ray from the screen center (where crosshair would be)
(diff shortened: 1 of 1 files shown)


examples/models/models_basic_voxel.c
```diff
@@ -72,19 +72,22 @@ int main(void)
...
         if (IsMouseButtonPressed(MOUSE_LEFT_BUTTON))
         {
             // Cast a ray from the screen center (where crosshair would be)
             Vector2 screenCenter = { GetScreenWidth()/2.0f, GetScreenHeight()/2.0f };
             Ray ray = GetMouseRay(screenCenter, camera);
 
             // Check ray collision with all voxels
-            bool voxelRemoved = false;
-            for (int x = 0; (x < WORLD_SIZE) && !voxelRemoved; x++)
+            float closestDistance = 99999.0f;
```
CrackedPixel, line 83: since `raymath.h` imports `math.h` you could use `INFINITY` if you wanted 😉 not necessary, though

examples/models/models_basic_voxel.c
```diff
@@ -72,19 +72,22 @@ int main(void)
...
+            Vector3 closestVoxelPosition = { -1, -1, -1 };
+            bool voxelFound = false;
+            for (int x = 0; x < WORLD_SIZE; x++)
             {
-                for (int y = 0; (y < WORLD_SIZE) && !voxelRemoved; y++)
+                for (int y = 0; y < WORLD_SIZE; y++)
                 {
-                    for (int z = 0; (z < WORLD_SIZE) && !voxelRemoved; z++)
+                    for (int z = 0; z < WORLD_SIZE; z++)
```
CrackedPixel, line 90: it sucks that you have to iterate over the entire world each time, but i dont have a suggestion

CrackedPixel:
(im not ray and this is just my opinion as a supporter/contributor)

Because this is an example, i dont think the performance hit tradeoff and extra complexity is worth the bug fix for this. Taking what ray said in the issue (`just a comment can be added explaining the issue and how it can be addressed`) I think it may be enough to add a comment saying _"in some cases the hit may remove cubes out of view of the camera. this can be fixed in a real project by doing XYZ"_ etc etc

SardineMilk:
I agree, it is a very inefficient and naive way to do it. It's ~8x slower than the previous algorithm. Implementing DDA would be the same or fewer lines of code, but it's quite a dense algorithm and I think it would be too complicated for the example.
However, I think it's better to have a naive but working solution, rather than an incorrect one.
I'm not too attached to this though, and am fine with closing the pr if that's what is wanted.

raysan5:
@SardineMilk I know is not the most optimal solution but I think it's good enough as an example

Issue #5642, opened 2026-03-11
[rtext] Semantics of TextInsert


chrg127:
`TextInsert` appears to be buggy to me.

Here's some example code:

```
#include <raylib.h>
#include <stdio.h>

int main()
{
    const char *first_part = "hello";
    const char *second_part = " world";
    char *text = TextInsert(first_part, second_part, strlen(first_part));
    printf("%s\n", text);
    free(text);
    return 0;
}
```

When trying to run it, I get `hellod` instead of `hello world`, which is what I would expect the output to be.

Looking inside `TextInsert`, it appears the problem is line 1878:

```
        for (int i = position; i < insertLen + position; i++) result[i] = insert[i];
```

The function copies part of `insert` starting from index `position`; to get my desired output, I'd have to change `insert[i]` to `insert[i - position]`. That said, since this function has been like this for a long time, I'm wondering if I'm misusing it? Please let me know.

CrackedPixel:
This is definitely a bug. Good find. I think it just went so long without being noticed due to limited use?

PR #5644, opened 2026-03-11
[rtext] fix error in textinsert, found by github user chrg127

CrackedPixel:
Fixes the bug found in https://github.com/raysan5/raylib/issues/5642

```cpp
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

#include "raylib.h"

int main()
{
    const char *first_part = "hello";
    const char *second_part = " world";
    char *text = TextInsert(first_part, second_part, strlen(first_part));
    printf("%s\n", text);
    free(text);
    return 0;
}
```

expected: `hello world`
actual: `hellod`

(since it grabs out of bounds, the `actual` may vary)

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1875,7 +1875,7 @@ char *TextInsert(const char *text, const char *insert, int position)
         result = (char *)RL_MALLOC(textLen + insertLen + 1);
 
         for (int i = 0; i < position; i++) result[i] = text[i];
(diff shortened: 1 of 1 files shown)


raysan5:
@CrackedPixel Good catch! Thanks for the fix! Just checked why it was not detected before and I found that I never used that function... wondering why it was added... most of the functions added are usually used by me in some of my projects...

PR #5645, opened 2026-03-12
Adapt build.zig to redesigned raylib build features config system

Wertual08:
Hello, It`s my first time submitting the PR, so, sorry to waste you time if it does not worth to be merged...

**Problem:**
Since commit 5361265a7d5d1ef662077350ef3db9747333b95b raylib config system was redesigned, and current build.zig stopped working at all by providing empty defines. Furthermore, I believe that with new config system there is no need to parse config.h to extract default defines. 

**Solution:**
I removed default config.h parsing and replaced passing -D flags with appropriate addCMacro build call.
Overriding default values can now be done by users as in example below instead of passing options.config which seemed to me like a crutch O_o

```zig
const raylib_library = raylib_dependency.artifact("raylib");
raylib_library.root_module.sanitize_c = .off;
raylib_library.root_module.sanitize_thread = false;
raylib_library.root_module.addCMacro("SUPPORT_MOUSE_GESTURES", "0");
```

I`ve tested building of my projects on Windows 11 and latest Linux arch.

Thanks you very much in advance for your time. Hope this will be helpful...

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -90,41 +90,7 @@ fn setDesktopPlatform(raylib: *std.Build.Step.Compile, platform: PlatformBackend
     }
 }
 
-/// A list of all flags from `src/config.h` that one may override
-const config_h_flags = outer: {
-    // Set this value higher if compile errors happen as `src/config.h` gets larger
-    @setEvalBranchQuota(1 << 20);
-
-    const config_h = @embedFile("src/config.h");
-    var flags: [std.mem.count(u8, config_h, "\n") + 1][]const u8 = undefined;
-
-    var i = 0;
-    var lines = std.mem.tokenizeScalar(u8, config_h, '\n');
-    while (lines.next()) |line| {
-        if (!std.mem.containsAtLeast(u8, line, 1, "SUPPORT")) continue;
-        if (std.mem.containsAtLeast(u8, line, 1, "MODULE")) continue;
-        if (std.mem.startsWith(u8, line, "//")) continue;
-        if (std.mem.startsWith(u8, line, "#if")) continue;
-
-        var flag = std.mem.trimStart(u8, line, " \t"); // Trim whitespace
-        flag = flag["#define ".len - 1 ..]; // Remove #define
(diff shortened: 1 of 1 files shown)


felipeasimos:
Hi @Wertual08, I ended up also opening a [PR](https://github.com/raysan5/raylib/pull/5650) that ends up tackling the same issue in parsing `config.h`. I ended up taking another approach though, so I guess is up to raysan to choose which way we should go. Either way, I just want to let you know that this solution seems to result in a black screen when running examples, using the latest zig version (0.16.0-dev.2860+9c5460316).

felipeasimos:
I was taking a look at it, and the black screen in zig master is probably because not all macros are included and also don't have their default values included in the try raylib_flags_arr.append(b.allocator, "-DMACRO"); statements.

As far as I understand, not using a value was compatible with the previous config system where only defined(MACRO) was used, so now we need to also pass the value, correct? (ex.: -DMACRO=1)

Wertual08:
Yeah, but new config.h already provides meaningfull default values, so why parse it to override them with the same ones?

felipeasimos:
I am not entirely familiar with the build process, so I could be missing something obvious, but I feel like like the default values are not being propagated the way you think they are? For example, I managed to find exactly what makes `zig build core_basic_window` result in a black screen. Just add this to line 225:

`try raylib_flags_arr.append(b.allocator, "-DSUPPORT_CUSTOM_FRAME_CONTROL=0");`

If you add this, the example now works! Using `=1` or leaving it empty doesn't.

I think your approach is good, I just think we need to validate with zig master.

Wertual08:
Thanks, I`ll check this out today.

felipeasimos:
Actually, I am sorry, I ended up using the `master` branch from your fork. The example works correctly! I'll close my PR

Wertual08:
Oh, no problem! Thanks you very much for testing my PR. I was surprised, because I tried it on two systems, but now we came to agreement. Thanks for your time!

raysan5:
@Wertual08 @felipeasimos thanks for the improvement and further reviews!

PR #5648, opened 2026-03-13
Update BINDINGS.md Factor bindings up to 5.5

ArnautDaniel:
Factor bindings are up to 5.5.


raysan5:
@ArnautDaniel thanks for the update!

PR #5650, opened 2026-03-13
fix black screen when using zig master: config.h macro extraction in build.zig

felipeasimos:
**Problem**:
On linux X11, using the latest zig master (0.16.0-dev.2860+9c5460316), every example was resulting in a non-interactive black screen.

**Solution**:
Using git bisect I tracked the [bad commit 5361265](https://github.com/raysan5/raylib/commit/5361265a7d5d1ef662077350ef3db9747333b95b). The new config system introduced in that commit uses values instead of just checking `defined(MACRO)`, but the config macro parsing in `build.zig` wasn't updated to take that into account.

I just fixed the parsing, so now it works. Also added `zig-pkg/` to .gitignore, since it is created when building with the latest zig version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,6 +116,7 @@ GRTAGS
 GTAGS
 
 # Zig programming language
+zig-pkg/
 .zig-cache/
 zig-cache/
 zig-out/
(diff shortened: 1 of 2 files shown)


Wertual08:
Hey, can you please check my similar PR https://github.com/raysan5/raylib/pull/5645 which also addresses simplification made possible by refactored configuration system.

felipeasimos:
sorry, I moved the comment to your PR, I think it is more organized if we keep the discussion about your PR there

felipeasimos:
closing it since @Wertual08 PR already solve this.

PR #5653, opened 2026-03-14
[rcore][sdl] REVIEWED: `GetTime()`, improve resolution

somamizobuchi:
Problem
-------
`GetTime()` in the SDL backend uses `SDL_GetTicks()`, which has only
millisecond resolution. This causes frame timing set with `SetTargetFPS()`
to be inaccurate at high frame rates. For example, targeting 240 Hz results
in an actual cap of ~200 Hz, because the inter-frame sleep is rounded up to
the nearest millisecond. This is observable in the `core_delta_time` example.

Cause
-----
`SDL_GetTicks()` has millisecond resolution, making it insufficient for
sub-millisecond frame timing.

Solution
--------
Calculate time via:

    (double)SDL_GetPerformanceCounter() / (double)SDL_GetPerformanceFrequency()

This uses an OS-specific high-resolution counter (typically micro- or
nanosecond precision). Note: unlike `SDL_GetTicks()`, the absolute value
returned is not relative to SDL initialization — its origin is OS-defined.
However, since `GetTime()` is used only to compute deltas between two calls,
this has no practical effect on behavior.

Tested
------
- MacBook Air M3
- macOS 26.3 (Tahoe) + SDL2 / SDL3
- examples/core/core_delta_time.c

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1276,9 +1276,7 @@ void SwapScreenBuffer(void)
 // Get elapsed time measure in seconds
 double GetTime(void)
 {
-    unsigned int ms = SDL_GetTicks();    // Elapsed time in milliseconds since SDL_Init()
-    double time = (double)ms/1000;
-    return time;
+    return (double)SDL_GetPerformanceCounter() / SDL_GetPerformanceFrequency();
 }
 
 // Open URL with default system browser (if available)


slendidev (changes requested):
nit: Why have the intermediate `time` variable? You can return the new calculated time directly.

slendidev (changes requested):
Another nit, I don't think the first cast to double would be required, since it should implicitly convert to it because of the RHS being a double. Also should probably squash those changes into one commit, it is a tiny change and having 3 commits for a 4 line change is a bit much imho.

raysan5:
@somamizobuchi @slendidev I'm merging changes but note that code will be reviewed to be aligned with other platforms `GetTime()`, considering `CORE.Time.base` for window initialization time.

PR #5654, opened 2026-03-14
 LibraryConfigurations.cmake: exchanged MATCHES -> STREQUAL in platform choosing if-statements

Crisspl:
To be honest I dont know how it didnt break for me before.
However, it became apparent for me that `if (${PLATFORM} MATCHES "Web")` with `PLATFORM=WebRGFW` resolves to true, as partial matches as apparently accepted as well (?) by cmake. So it was effectively impossible to choose WebRGFW platform.


raysan5:
@Crisspl Thanks for the review, notice that `WebRGFW` was added recently.

PR #5656, opened 2026-03-15
Add SetMouseCursorImage() and ClearMouseCursorImage() for GLFW platform

UniquePython:
Adds two new API functions for setting a custom image as the mouse cursor on the GLFW desktop platform:
```c
    void SetMouseCursorImage(Image image, int hotX, int hotY);
    void ClearMouseCursorImage(void);
```
Implementation details:
- Converts the input Image to RGBA format before passing to GLFW (required by glfwCreateCursor)
- Stores the GLFWcursor handle in PlatformData to allow proper cleanup
- Previous custom cursor is destroyed before creating a new one
- SetMouseCursor() calls ClearMouseCursorImage() first to avoid leaking handles when switching back to a standard cursor
- ClosePlatform() destroys the custom cursor handle before glfwDestroyWindow()

Other platforms (Android, DRM, web, etc.) are not affected.

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -727,6 +727,17 @@ void SetMouseCursor(int cursor)
     TRACELOG(LOG_WARNING, "SetMouseCursor() not implemented on target platform");
 }
 
+void SetMouseCursorImage(Image image, int hotX, int hotY)
+{
+    TRACELOG(LOG_WARNING, "SetMouseCursorImage() not implemented on target platform");
+}
+
+void ClearMouseCursorImage(void)
+{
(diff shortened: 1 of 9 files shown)


raysan5:
@UniquePython I prefer to avoid those new functions, custom cursor images can be set at user side using images, otherwise that's a platform-specific feature that many platforms do not support. In raylib I prioritize fixes and internal improvements over additions.

UniquePython:
> @UniquePython I prefer to avoid those new functions, custom cursor images can be set at user side using images, otherwise that's a platform-specific feature that many platforms do not support. In raylib I prioritize fixes and internal improvements over additions.

Thanks for the feedback, understood. I'll focus on bug fixes and internal improvements going forward.

PR #5657, opened 2026-03-15
Add GetMonitorVideoModes() for querying supported monitor resolutions (GLFW platform)

UniquePython:
Adds a new API function and supporting struct to query all supported video modes for a given monitor:
```c
    VideoMode *GetMonitorVideoModes(int monitor, int *count);
```
New struct:
```c
    typedef struct VideoMode {
        int width;
        int height;
        int refreshRate;
    } VideoMode;
```
Implementation details:
- Uses glfwGetVideoModes() internally to enumerate all supported resolutions
- Allocates and returns a caller-owned array — caller must free with RL_FREE()
- Writes the number of modes into the count out-parameter
- All non-GLFW platforms return NULL with count = 0 and a LOG_WARNING stub

Use case: allows users to enumerate valid fullscreen resolutions before calling
SetWindowSize() or toggling fullscreen, which was previously impossible without
going outside the raylib API.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1001,6 +1001,40 @@ int GetMonitorRefreshRate(int monitor)
     return refresh;
 }
 
+// Get available video modes for the specified monitor
+VideoMode *GetMonitorVideoModes(int monitor, int *count)
+{
+    VideoMode *modes = NULL;
+    *count = 0;
+
+    int monitorCount = 0;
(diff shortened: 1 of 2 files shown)


raysan5:
@UniquePython No plans to add those functions, I try to keep raylib API and structs minimal

UniquePython:
> @UniquePython No plans to add those functions, I try to keep raylib API and structs minimal

Got it. Will keep that in mind next time

PR #5669, opened 2026-03-17
[rlsw] Fix typos, update build scripts, and align style

Bigfoot71:
There were a few small typos introduced during the reformatting.
I also fixed the build scripts to account for the renaming of `GRAPHICS_API_OPENGL_SOFTWARE`
Finally, I made one last consistency adjustment based on your changes.


raysan5:
@Bigfoot71 Ouch! Good catches! Thanks for the further review! (that's what happens when you use two different IDEs at the same time, Notepad++ and VS2022)

PR #5671, opened 2026-03-18
Fix Uninstall Process to Remove rcamera Header File in Makefile

0x00650a:
Addressing _(which i think)_ an oversight in the uninstall functionality within the Makefile, which may lead to leftover files after uninstall.
### Changes Made

- Updated the uninstall section of the Makefile to include the command to remove `rcamera.h` header file.


raysan5:
@0x00650a thanks for the update

PR #5672, opened 2026-03-18
Make fogColor in the fog tutorial a parameter

devel60-alt:
Pretty self explanatory, now the fogColor vec4 in fog.fs can be edited from code. I wasn't sure though if I should have converted the output of ColorNormalize to a float* like ambient used to be (most likely with another function, but I don't think it would have been necessary), so I decided to make them both Vector4's for consistency


raysan5:
@devel60-alt thanks for  the revieew

PR #5674, opened 2026-03-19
Required change for ESP-IDF 6

georgik:
I would like to open a conversation, how to get support for embedded platforms in Raylib 6.

Basically I found two changes that were necessary to make for Raylib 6 to get it working with ESP32: https://github.com/georgik/esp-idf-component-raylib

Note: Raylib 5.6 already works with ESP-IDF.

Suggested changes:
- swGetColorBuffer - expose pixel buffer, so that it could be sent to SPI display directly using DMA transactions.
- in case of `CORE.Window.render.width` == 0 set it to `CORE.Window.screen.width`, since embedded platforms does not have window manager. I can imagine moving this part to compat layer. I value opinions.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -689,6 +689,7 @@ SWAPI void swClear(uint32_t bitmask);
 SWAPI void swBlendFunc(SWfactor sfactor, SWfactor dfactor);
 SWAPI void swPolygonMode(SWpoly mode);
 SWAPI void swCullFace(SWface face);
+SWAPI void *swGetColorBuffer(int *width, int *height); // Restored for ESP-IDF compatibility
 
 SWAPI void swPointSize(float size);
 SWAPI void swLineWidth(float width);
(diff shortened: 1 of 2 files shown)


src/external/rlsw.h
```diff
@@ -4200,6 +4201,15 @@ void swCullFace(SWface face)
...
 }
 
+// Get direct pointer to the default framebuffer's pixel data
+// Restored for ESP-IDF compatibility - removed in Raylib 6.0 PR #5655
+void *swGetColorBuffer(int *width, int *height)
+{
+    if (width) *width = RLSW.colorBuffer->width;
+    if (height) *height = RLSW.colorBuffer->height;
+    return RLSW.colorBuffer->pixels;
+}
```
Bigfoot71, line 4211: If you want to get the dimensions/pointer of the default framebuffer, you should go through `RLSW.framebuffer.color`, because `RLSW.colorBuffer` corresponds to a pointer of the currently bound framebuffer, not necessarily the default one.

raysan5: Thanks! I'm reviewing it after merge!

raysan5:
@georgik Thanks for the update! I'm reviewing the further required changes!

PR #5677, opened 2026-03-20
[rlsw] Remove position attribute from `sw_vertex_t`

Bigfoot71:
Just replace  `float coord[4]` with `float position[4]` instead.

Until now the world position was being preserved, but the only real point would be if we ever add lighting as defined in OpenGL 1.1, and therefore also at the rlgl level, but that doesn't seem to be planned.

By the way, positions were interpolated during clipping, so this should produce a marginal gain at this stage.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -4371,7 +4371,7 @@ void swPopMatrix(void)
                 return;
             }
 
-            RLSW.currentMatrix = &RLSW.stackProjection[--RLSW.stackProjectionCounter];
+            RLSW.currentMatrix = &RLSW.stackProjection[(--RLSW.stackProjectionCounter) - 1];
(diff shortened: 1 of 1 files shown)


Bigfoot71:
I also just fixed a bug with `swPopMatrix()` along the way, I had noticed it with `models_rlgl_solar_system`

_By the way, the push/pop functions could be easily simplified, but hey_

raysan5:

 > By the way, the push/pop functions could be easily simplified, but hey

I imagine you mean on rlgl side, that system was implemented ~10 years ago and probably it can be improved

Bigfoot71:
> I imagine you mean on rlgl side, that system was implemented ~10 years ago and probably it can be improved

 No I did mean `rlsw`, there's some duplication, but it works for now, so I'll avoid over-reviewing it.

PR #5680, opened 2026-03-21
[Examples] fixed error in readme, found by discord user the.enemy

CrackedPixel:
`rlgl.c` does not exist, it's `rlgl.h`

found by discord user `the.enemy`


raysan5:
@CrackedPixel Good catch! Thanks for the fix!

PR #5681, opened 2026-03-22
Add support for some integer texture formats

callidus:
Adding support for RGBA8888 integer and unsigned integer texture formats.

I needed this for a project I'm building using raylib, so I added it. I think this is all correct and it works fine for me so contributing it back up in the hope that it is useful.


raysan5:
@callidus Thanks for the proposed addition but I prefer to avoid it for now. It seems intended for very specific use cases and it adds more maintenance work.

callidus:
fair enough, thanks for your efforts on raylib and have a good day.

PR #5682, opened 2026-03-22
[rcore] refactor possible data loss on FileMove()

lvntky:
The FireRemove() function was called without any checks. If FileCopy encountered an error for any reason, FileRemove() would irreversibly delete the copied data.

Also `result` was initialized to `0` (success) making the return 
value meaningless when `FileCopy()` or `FileRemove()` failed.

Fix by checking `FileCopy()` return value before removing the source 
file, and initializing `result` to `-1`.


raysan5:
@lvntky thanks for  the review!

lvntky:
> @lvntky thanks for the review!

Thank you @raysan5 ! this was my first commit on raylib !!

PR #5683, opened 2026-03-22
[examples] Fix memory out of bounds in [shapes] example - ball physics

Jopestpe:
## Description

* Fixes memory out of bounds in [shapes] example - ball physics by using malloc/free

## Changelog

* Modified examples/shapes/shapes_ball_physics.c - code
* Updated examples/shapes/shapes_bouncing_ball.png – refreshed screenshot for accuracy


raysan5:
@Jopestpe Thanks for the review but... did you try it on `PLATFORM_WEB`?

Because this change is probably not related to the listed issue...

Jopestpe:
> @Jopestpe Thanks for the review but... did you try it on `PLATFORM_WEB`?
> 
> Because this change is probably not related to the listed issue...

My bad, I didn't mention it earlier, I tested it on both Linux and the web. Initially, reducing MAX_BALLS to 1300 or less didn't cause the memory out of bounds since as it fits within [Emscripten's STACK_SIZE (64*1024)](https://emscripten.org/docs/tools_reference/settings_reference.html#stack-size:~:text=Default%20value%3A%2064%2A1024), but I preferred to move it to the heap while keeping the 5000 from the original script.

PR #5684, opened 2026-03-23
[examples][models] Adding `TEXTURE_WRAP_REPEAT` in `models_yaw_pitch_roll`

maiconpintoabreu:
Adding `TEXTURE_WRAP_REPEAT` force the repeat to happen on Web version and works as normal on the other ones.

Web:

image

Desktop:

image


raysan5:
@maiconpintoabreu Thanks for the review... Note that it only works if NPOT textures are supported... and depends on OpenGL ES 2.0 / WebGL 1.0 implementation, depending on the device it could not be supported... but I imagine current WebGL 1.0 shipped with current browsers should address it...

PR #5688, opened 2026-03-23
[examples] Fix UI Cellular Automata

JordSant:
Fix bug in computing the red rectangle to draw when the mouse is hovering over: to compute correctly the X coordinate it must be an integer division (i/2) not a floating point division ((float)i/2). Now fixed


raysan5:
@JordSant Good catch! Thanks for the review!

PR #5691, opened 2026-03-25
[rtext] Add security check for `LoadCodepoints()`

konakona418:
When the `count` parameter is specified as a null value, `LoadCodepoints()` function can cause segmentation fault. By checking whether the param is null can improve the robustness of the function.


raysan5:
@konakona418 Thanks but I'm not merging this change for consistency, raylib has +30 functions that return a *count parameter by reference and expect user to provide a valid variable for those cases. It's up to the user to do so.

PR #5692, opened 2026-03-25
[rmodel] fix for devices without VAO support

### LINKED ISSUE #5690: [rmodel] Issue with es 2.0 device with no vao support
introduced return at line https://github.com/raysan5/raylib/blob/5dd4036ed0065e8e89a1ee84d89d3a18bb7a3d59/src/rmodels.c#L1284
when vao is 0 in es 2.0 devices without vao extesion support( es 2.0 has not vao support, only by extension , it depends of implementation in the device) will go out without uploading data mesh so it will not draw model finally when we can follow using vbo instead of vao.

One solution can be quit that line because we will use vbo instead vao in DrawMesh but we need mesh data uploaded.

Tested without that line in a console device with es 2.0 implementation without vao extesion or not available. 

What do you think?

psxdev:
Resolve #5690 

Avoid return for devices without VAO support, because they will use VBO.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -1281,7 +1281,6 @@ void UploadMesh(Mesh *mesh, bool dynamic)
 
 #if defined(GRAPHICS_API_OPENGL_33) || defined(GRAPHICS_API_OPENGL_ES2)
     mesh->vaoId = rlLoadVertexArray();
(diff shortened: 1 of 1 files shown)


raysan5:
@psxdev thanks for the review!

PR #5698, opened 2026-03-27
fix: remove dead variable in GetRender and change an unimplementation…

Monjaris:
… log's level from INFO to WARNING

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -819,23 +819,15 @@ int GetScreenHeight(void)
 // Get current render width which is equal to screen width*dpi scale
 int GetRenderWidth(void)
 {
-    int width = 0;
-
     if (CORE.Window.usingFbo) return CORE.Window.currentFbo.width;
-    else width = CORE.Window.render.width;
-
-    return width;
+    else return CORE.Window.render.width;
 }
 
 // Get current screen height which is equal to screen height*dpi scale
 int GetRenderHeight(void)
 {
-    int height = 0;
-
     if (CORE.Window.usingFbo) return CORE.Window.currentFbo.height;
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -819,23 +819,15 @@ int GetScreenHeight(void)
 // Get current render width which is equal to screen width*dpi scale
 int GetRenderWidth(void)
 {
-    int width = 0;
```
raysan5, line 822: raylib prioritizes usage of intermidiate variables to avoid early returns, please, could you review this code?

src/rcore.c
```diff
@@ -2281,7 +2273,7 @@ int FileMove(const char *srcPath, const char *dstPath)
 		if (FileCopy(srcPath, dstPath) == 0) result = FileRemove(srcPath);
         else TRACELOG(LOG_WARNING, "FILEIO: [%s] Failed to copy file to [%s]", srcPath, dstPath);
     }
-	else TRACELOG(LOG_WARNING, "FILEIO: [%s] Source file does not exist", srcPath);
+    else TRACELOG(LOG_WARNING, "FILEIO: [%s] Source file does not exist", srcPath);
```
raysan5, line 2276: what is the changee in this line?

monjaris: thought indentation was different from the others

src/rcore.c
monjaris: I removed the `width` variable because it wasn’t reused and the expression is short enough to be readable inline. Keeping it didn’t improve clarity and just added an extra variable.
If you think it improves readability in this context, I can revert it.

raysan5: yes, please, revert it

PR #5701, opened 2026-03-28
Add CheckCollisionPointEllipse and CheckCollisionEllipses functions

Monjaris:
#### This time i wrote fairly amount of usable two functions.
I was writing `struct Bullet` for my toy game and wanted bullet to  be ellipse instead of a circle.
Therefore, I decided to contribute my function by converting it to Raylib's "style/conventions"


raysan5:
@Monjaris I prefer to avoid those functions, add maintenance cost and most users could not need it. In any case, they can be added in a separate example.

PR #5703, opened 2026-03-28
[rlgl] Remove forced normalization of normal vectors in rlNormal3f

JeffM2501:
Remove forced normalization of normal vectors in rlNormal3f, assume the user has set the correct input for the shader. This allows users to use the batch system normals for special effects and custom lighting that may need to use a non unit magnitude.


raysan5:

PR #5704, opened 2026-03-28
[CORE] Unify raymath.h and raylib.h structure checks

JeffM2501:
This PR makes raylib.h check to see if math types like Vector2/3 and matrix are defined before defining them, in the same way that raymath.h does. Previously raylib.h assumed it was always setting the structures, and this forces raylib.h to be included before raymath.h.
This PR makes the include order not matter, making it much easier to include external libs that may use raymath or raylib in any order.
This should not break any compatibly, the same flags are set, just in a slightly different order during compilation.

### DIFF
--- a/tools/rlparser/output/raylib_api.json
+++ b/tools/rlparser/output/raylib_api.json
@@ -103,36 +103,6 @@
       "description": ""
     },
     {
-      "name": "RL_VECTOR2_TYPE",
-      "type": "GUARD",
-      "value": "",
-      "description": ""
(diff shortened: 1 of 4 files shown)


raysan5:
@JeffM2501 I prefer to avoid that change, one type should not be used if the other types are not defined

JeffM2501:
That's what this does, ensure that once one type is used, the other is not, it just makes raylib behave that same way as raymath.h.....

Issue #5658, opened 2026-03-15
[rcore] Discussion: window resize detection on web platforms


Crisspl:
Currently WebRGFW platform discards any resize events completely*: https://github.com/raysan5/raylib/blob/faf42366ec554c929118b39b1b612f9925f50a02/src/platforms/rcore_desktop_rgfw.c#L1334

I understand it's because our "window" is canvas, and browsers are only required to fire resize events on Window (which is not what we render to in case of web platforms).
However, wouldnt it be better if we report window resize event, but without doing any resize? This way raylib users would be able to use raylib API (`IsWindowResized()`) to check if window was resized and do some actions based just the info that the resize happened. For example: use emscripten API to query <body> element current size and resize canvas adequately.

What this proposal means in code boils down to replacing `return;` line (see link above) with `CORE.Window.resizedLastFrame = true; break;`. At least in case of WebRGFW, not sure about rcore_web and rcore_web_emscripten.

\* - actually i think the `return;` currently present in code is a bug, since it returns from whole `PollInputEvents()` call. Shouldnt it be `break;` or just nothing?

CrackedPixel:
I agree with everything said here. PR made: https://github.com/raysan5/raylib/pull/5662

PR #5709, opened 2026-03-31
[rcore_rgfw] Updating render resolution as well on SetWindowSize

Crisspl:
Shouldnt we update `CORE.Window.render` resolution here as well? Otherwise it is never updated correctly, unless handling of window resize event updates it. Which is never on WebRGFW unfortunately since resize event handling  intentionally (related to https://github.com/raysan5/raylib/issues/5658 ) doesn't update any sizes.
So: if window (canvas in case of web) is only ever resized with `SetWindowSize()` calls, then `CORE.Window.render` size will always be equal to initial window size.

### Explanation
In my case it was breaking scissor calls on web, after exiting "texture mode" (i.e. rendering to texture):

rcore.c:EndTextureMode:
```
// Reset current fbo to screen size
CORE.Window.currentFbo.width = CORE.Window.render.width;
CORE.Window.currentFbo.height = CORE.Window.render.height;
CORE.Window.usingFbo = false;
```

rcore.c:BeginScissorMode:
```
rlScissor(x, CORE.Window.currentFbo.height - (y + height), width, height);
```

`EndTextureMode` was "resetting" `currentFbo` size to initial window's size, so subseqent `BeginScissorMode` scissor would pass incorrect value as 2nd parameter to rlScissor (sometimes even negative if scissor's y+height was bigger than initial window's size).

### DIFF
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -823,8 +823,7 @@ void SetWindowSize(int width, int height)
 
     if (!CORE.Window.usingFbo)
     {
-        CORE.Window.render.width = CORE.Window.screen.width;
-        CORE.Window.render.height = CORE.Window.screen.height;
+        SetupViewport(CORE.Window.screen.width, CORE.Window.screen.height);
     }
 
     RGFW_window_resize(platform.window, CORE.Window.screen.width, CORE.Window.screen.height);


raysan5:
@Crisspl reading the code, it make sense but note that I do not maintain the RGFW backend, pinging @CrackedPixel for review.

PR #5711, opened 2026-04-01
[examples][shaders_texture_tiling][WEB] explicitly state TEXTURE_WRAP_REPEAT for web build

r3g492:
issue: 
- WEB example does not fallback to TEXTURE_WRAP_REPEAT default

RLGL.ExtSupported.texNPOT is 1 on desktop,
but 0 for the web.
this is environment capability, so one cannot simply change this.

the problem i believe is that 
inside rlLoadTexture, if texNPOT is false, it fallbacks to CLAMP_TO_EDGE, even if the example png is a POT.

the code is not checking actual POT at the moment.

we should add that, 
or explicitly call SetTextureWrap(texture, TEXTURE_WRAP_REPEAT); for web.

as a newcomer, i limited myself to example change here.
thank you,


raysan5:
@r3g492 thanks for the review!

PR #5717, opened 2026-04-04
[rcore][SDL] Reviewed monitor indexing for SDL3

somamizobuchi:
## Problem
SDL3 changed monitor indexing from ordered to logical IDs which breaks many functions that requires monitor index as input. 
## Solution
Similar to how gamepads are currently stored, keep an array of `displayID`s that are queried and set once during `PlatformInit()` and indexed during function calls that require them. The event polling does not currently handle display add/remove events so this LUT is only valid if no monitors are added/removed.  

## Tests
MacOS Tahoe 26.3
MacBook Air M3
SDL 3.4.2
Tested using `example/core/monitor_detector` with two monitors

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -91,6 +91,10 @@
 
 #define SCANCODE_MAPPED_NUM     232
 
+#if defined(USING_VERSION_SDL3)
+    #define MAX_MONITORS 8
+#endif
+
 //----------------------------------------------------------------------------------
 // Types and Structures Definition
 //----------------------------------------------------------------------------------
@@ -100,6 +104,9 @@ typedef struct {
 
     SDL_GameController *gamepad[MAX_GAMEPADS];
     SDL_JoystickID gamepadId[MAX_GAMEPADS]; // Joystick instance ids, they do not start from 0
+#if defined(USING_VERSION_SDL3)
+    SDL_DisplayID displayId[MAX_MONITORS];
+#endif
     SDL_Cursor *cursor;
 } PlatformData;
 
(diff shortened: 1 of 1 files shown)


somamizobuchi:
Hi @raysan5 I know this is not the priority for 6.0 but is there a timeline for review on this? Or anything I can do in the meantime?

raysan5:
@somamizobuchi Just wait a bit more, `raylib 6.0` is being released tomorrow and I'm working on it at the moment, I will review it in the following weeks...

src/platforms/rcore_desktop_sdl.c
```diff
@@ -2091,6 +2122,28 @@ int InitPlatform(void)
...
 
+#if defined(USING_VERSION_SDL3)
+    // Init monitors
+    for (int i = 0; i < MAX_MONITORS; i++) 
+    {
+        platform.displayId[i] = -1;
+    }
+
+    int numMonitors = 0;
+    SDL_DisplayID *displays = SDL_GetDisplays(&numMonitors);
```
raysan5, line 2133: I think an `SDL_free(displays)` is missing at the end

src/platforms/rcore_desktop_sdl.c
```diff
@@ -2091,6 +2122,28 @@ int InitPlatform(void)
...
+    if (numMonitors < 1) 
+    {
+        TRACELOG(LOG_FATAL, "PLATFORM: Failed to find monitors");
+        SDL_free(displays);
+        return -1;
```
raysan5, line 2138: I prefer to avoid ending the program if no monitor connected, maybe `CORE.Window.ready = false`?

raysan5:
No answer from contributor. closing.

PR #5719, opened 2026-04-04
[rlsw] Some platform fixes

Bigfoot71:
- Fixes compilation errors for DRM.
- Handles resizing for SDL and WIN32.

---

A little validation for WIN32 would be preferable; I don't have a Windows machine to test on, but I simply added this to `HandleWindowResize`:
```c
#if defined(GRAPHICS_API_OPENGL_SOFTWARE)
swResize(clientSize.cx, clientSize.cy);
#endif
```

Also, I had added it for RGFW, but when I tried to test it I realized that software rendering is not supported, black screen, so I didn't do anything.


Bigfoot71:
Oops, branch issue

PR #5724, opened 2026-04-05
[rcore_rgfw] Icon color format fix

Crisspl:
After RGFW update to 2.0.0-dev in fbd83cafc7c51ecc38be7f7b19c185a42f333cb0, RGFW_window_setIcon api has changed -- not it takes pixel format enum instead of number of channels. On raylib side it was still passing 4 (number of channels in rgba) which is enum value for BGRA8. Instead we have to pass 2 now (RGFW_formatRGBA8 = 2).


raysan5:
@Crisspl thanks for the review

PR #5725, opened 2026-04-05
Specified default for visible and enabled cursor, and event waiting

RolandMarchand:
The reference documentation for the following on/off settings didn't specify which one was set by default. This PR fixes that. No logic code was modified.

```c
void EnableEventWaiting(void);
void DisableEventWaiting(void);
void ShowCursor(void);
void HideCursor(void);
void EnableCursor(void);
void DisableCursor(void);
```


raysan5:
@RolandMarchand thanks but I consider those anotations are not required

Issue #5726, opened 2026-04-05
[rlgl] duplicate RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES define blocks


casperbear:
In rlgl.h there are duplicate RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES define blocks with different comments

```
#ifndef RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES
    #define RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES "boneMatrices"   // Bound by default to shader location: RL_DEFAULT_SHADER_ATTRIB_LOCATION_BONEMATRICES
#endif

#ifndef RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES
    #define RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES "boneMatrices"     // bone matrices (required for GPU skinning)
#endif
```

PR #5727, opened 2026-04-05
Issue #5726 ( RL_DEFAULT_SHADER_UNIFORM_NAME_BONEMATRICES ) 

angshumankishore:
The first definition (line 1018) has a comment referencing RL_DEFAULT_SHADER_ATTRIB_LOCATION_BONEMATRICES (which seems like a copy-paste mistake, since this is a uniform, not an attribute).
The second definition (line 1053) has a clearer comment: "bone matrices (required for GPU skinning)".


raysan5:
@angshumankishore Good catch! thanks for the review!

PR #5729, opened 2026-04-06
[raymath] Fix: ``QuaternionMultiply``

0stamina:
the multiplication algorithm was slightly off
previously when combining quaternion rotations objects would shrink now they rotate as expected

trying to get around the issue by using ``QuaternionToMatrix`` then ``QuaternionTransform`` doesn't work, and I'm not sure why

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -4384,7 +4384,7 @@ static void BuildPoseFromParentJoints(BoneInfo *bones, int boneCount, Transform 
                 TRACELOG(LOG_WARNING, "Skipping bone not topologically sorted: Bone %d has parent %d", i, bones[i].parent);
                 continue;
             }
-            transforms[i].rotation = QuaternionMultiply(transforms[bones[i].parent].rotation, transforms[i].rotation);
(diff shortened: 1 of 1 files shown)


raysan5:
@0stamina This is a sensible change, used by raylib animation system, did you verify it does not break animations rotations?

0stamina:
turns out the multiplications for animations were swapped to account for the mistake in the formula, I can't find any other instance of using quaternion multiplication to check

raysan5:
@0stamina I'm afraid I need some time to review this sensible change. Sending it without checking the implications in raylib from your side is very concerning and makes me doubt.

casperbear:
AI doesn't recommend to merge the fix and suggests to check for problems on user side, in particular 1) whether a Left-Handed coordinate system is used in a Right-Handed library (RayLib/OpenGL). 2) whether arguments are passed in incorrect order to QuaternionMultiply - new/local rotation should be passed first, base/world rotation second

raysan5:
I decided not to merge this change for now.

PR #5730, opened 2026-04-06
Add GenMeshPlaneData() for CPU-side mesh generation without GPU upload

coochewgames:
Currently GenMeshPlane() generates the mesh and immediately uploads it to the GPU via UploadMesh(). This makes it impossible to manipulate vertex data on the CPU (e.g. procedural height deformation) before the GPU upload, without resorting to UpdateMeshBuffer() after the fact.  I have been kind of stuck using a modified 4.2 version of Raylib for the last few years and would like to get this into the main branch for future continuity with this PR; it introduces GenMeshPlaneData() which contains all the existing mesh generation logic but returns the mesh without uploading it to the GPU.
GenMeshPlane() is updated to delegate to GenMeshPlaneData() then call UploadMesh(), so existing behaviour is completely unchanged for all current callers.

Use case — procedural terrain generation:
```
// Generate the mesh on the CPU without uploading
Mesh mesh = GenMeshPlaneData(width, length, resX, resZ);

// Deform vertex Y values using noise
for (int i = 0; i < mesh.vertexCount; i++)
{
    mesh.vertices[i*3 + 1] = SampleHeight(mesh.vertices[i*3], mesh.vertices[i*3 + 2]);                                                      
}

// Recalculate normals, then upload once
RecalcNormals(&mesh);
UploadMesh(&mesh, false);
```

Changes:
- src/raylib.h — added GenMeshPlaneData() declaration
- src/rmodels.c — added GenMeshPlaneData() implementation; GenMeshPlane() now wraps it

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1617,7 +1617,8 @@ RLAPI bool ExportMeshAsCode(Mesh mesh, const char *fileName);                   
 
 // Mesh generation functions
 RLAPI Mesh GenMeshPoly(int sides, float radius);                                            // Generate polygonal mesh
-RLAPI Mesh GenMeshPlane(float width, float length, int resX, int resZ);                     // Generate plane mesh (with subdivisions)
+RLAPI Mesh GenMeshPlane(float width, float length, int resX, int resZ);                     // Generate plane mesh (with subdivisions) and upload to GPU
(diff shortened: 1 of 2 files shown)


raysan5:
@coochewgames thanks but I'm not adding this function, it would break consistency with other functions and it can be implemented in user side.

PR #5732, opened 2026-04-06
raymath: add ZERO_INITIALIZE macro to support C++ compilers

ruuzia:
Thank you for this wonderful standalone library!

I know nobody likes adding more defines in header files. Sorry about that, but I think this one is necessary and hardly that bad. In C++, struct zero-initialization is `{}` and in C it is `{0}`.

It's fine not to support C++ compilers, but this lovely header file already contains 400 lines of C++-specific code (operator definitions). So if C++ is not supported, we should not include C++-only features. If we're including C++ features, we probably want to support C++ compilers :p

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -178,6 +178,12 @@ typedef struct float16 {
 #define RL_FLOAT16_TYPE
 #endif
 
+#if defined(__cplusplus)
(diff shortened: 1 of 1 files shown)


raysan5:
@ruuzia Thanks but I prefer to avoid this macro.

PR #5733, opened 2026-04-06
Maked meson.build and meson_options.txt for Meson build system.

pliniopvv:
Maked meson.build and meson_options.txt for Meson build system.

image


raysan5:
@pliniopvv Thanks but I prefer not to add more build systems, they require lot of maintenance work and even current ones need improvements.

PR #5734, opened 2026-04-08
Add comment for player rectangle initialization

Bleawu:
Hadn't come across Rectangle struct till now thought this would help other people too.


raysan5:
@Bleawu this comment is not required here

Issue #5745, opened 2026-04-09
Crash when calling GenTextureMipmaps before InitWindow


EmmanuelMess:
Calling:

https://github.com/raysan5/raylib/blob/fbcc8cdb554574aa2abbafecb348c32928fd713e/src/rtextures.c#L4364-L4369

Before `InitWindow` crashes with SEGFAULT, but `LoadFontEx` doesn't, maybe a check and an error should be added.

```c
int main(void)
{
	SetTraceLogLevel(LOG_DEBUG);
	SetConfigFlags(FLAG_MSAA_4X_HINT);
	const int WIDTH = 800;
	const int HEIGHT = 450;

	Font ubuntu_font = LoadFontEx("resources/Ubuntu-M.ttf", 48, NULL, 0);
	GenTextureMipmaps(&ubuntu_font.texture);
	InitWindow(WIDTH, HEIGHT, "raylib example - basic window");
}
```

Where [Ubuntu-M.zip](https://github.com/user-attachments/files/26603843/Ubuntu-M.zip) is the ttf.

PR #5747, opened 2026-04-10
[rlgl] added gpuready check to gentexturemipmaps

CrackedPixel:
found by https://github.com/raysan5/raylib/issues/5745


raysan5:

Issue #5731, opened 2026-04-06
[build] `PLATFORM_MEMORY` not supported by CMake


nagolove:
On revision c5fc7716229cef1727e7baf325a695a0ac00cf27 I could not compile library with `Memory` backend for testing purposes.
```
raylib ❯ cmake -DPLATFORM=Memory .                                                             git:(master) ✗  
-- Testing if -Werror=pointer-arith can be used -- compiles
-- Testing if -Werror=implicit-function-declaration can be used -- compiles
-- Testing if -fno-strict-aliasing can be used -- compiles
CMake Error at cmake/EnumOption.cmake:7 (message):
  Unknown value Memory.  Only
  -DPLATFORM=Desktop;Web;WebRGFW;Android;Raspberry Pi;DRM;SDL;RGFW allowed.
Call Stack (most recent call first):
  CMakeOptions.txt:9 (enum_option)
  CMakeLists.txt:30 (include)
```
I think `Memory` should be in PLATFORM enum with additional flags like ` -DGRAPHICS=GRAPHICS_API_OPENGL_SOFTWARE ` and  `-DBUILD_EXAMPLES=OFF`

### Environment

Arch Linux distribution

raysan5:
Pinging the CMake maintainers for future review: @Peter0x44 @manuel5975p @VictorHM @JohnnyCena123

PR #5750, opened 2026-04-10
[makefile] update makefile for PLATFORM_MEMORY

CrackedPixel:
found while looking at https://github.com/raysan5/raylib/issues/5731

the default `make PLATFORM=PLATFORM_MEMORY` command will not compile due to the `-D{GRAPHICS}` since `graphics` is empty for memory platform


raysan5:

PR #5751, opened 2026-04-11
[text_words_alignment] - Windows (GCC)

jochemherz:
At line 53, the example uses:

`char **words = TextSplit(...);`

However, `TextSplit(...)` returns a `const char **`.

This causes a compilation error with strict GCC due to incompatible pointer types.

Suggested fix: 

`const char **words = TextSplit(...);`

This change helps avoid confusion for beginners and ensures compatibility with strict GCC/Clang builds.

### DIFF
--- a/examples/text/text_words_alignment.c
+++ b/examples/text/text_words_alignment.c
@@ -50,7 +50,7 @@ int main(void)
     // Define the text we're going to draw in the rectangle
     int wordIndex = 0;
     int wordCount = 0;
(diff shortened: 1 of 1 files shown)


maiconpintoabreu:
Every time I see branch called patch-1(any number) I think it is AI generated.

raysan5:
@maiconpintoabreu it probably is, it actually fixes it for 5.5 but not for 6.0

PR #5748, opened 2026-04-10
[raylib.rc] update raylib.rc to 6.0 (2026)

CrackedPixel:
taken from https://github.com/raysan5/raylib/issues/5746

PR #5754, opened 2026-04-12
Update 2025->2026, 5.5->6.0

CrackedPixel:
Idea by @maiconpintoabreu in https://github.com/raysan5/raylib/pull/5748#issuecomment-4222530730

This PR does not include examples because im not 100% confident they should really be updated

## Note: if this is merged before official release of 6.0 it will cause some things to break or not make sense. Example, the cmake will search for "6.0" release etc etc


raysan5:
@CrackedPixel Thanks for the update but note that current release is not released yet, so I didn't update that part in CHANGELOG... I think it's ok to leave it like this...

PR #5755, opened 2026-04-12
Update BINDINGS.md to add new FreeBASIC port

mudhairless:
Added the raylib4fb port


raysan5:
@mudhairless thanks for the addition!

PR #5758, opened 2026-04-14
[zig] Adjusting build.zig to run on 0.16.0

maiconpintoabreu:


HaxSam:
Looks good

build.zig.zon
```diff
@@ -1,7 +1,7 @@
 .{
     .name = .raylib,
     .version = "5.6.0-dev",
```
raysan5: probably this should be updated to raylib 6.0

HaxSam: Oh yeh for sure would have done it in my pr too

maiconpintoabreu: True, I will change and test it again just in case

maiconpintoabreu: Done

raysan5:
@maiconpintoabreu @HaxSam thanks for the review

PR #5760, opened 2026-04-15
update defines for new format etc

CrackedPixel:
i saw a comment suggesting to "undefine", but this no longer works. 

[raymath.h was updated here](https://github.com/raysan5/raylib/blob/master/src/raymath.h#L33) to show 0/1 sort of thing so i tried to follow that while also mentioning to define as 0/1 in the file OR the build system (since that is supported now)

all added values are set to match the default values in config.h

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -252,6 +252,9 @@
 #ifndef SUPPORT_FILEFORMAT_PIC
     #define SUPPORT_FILEFORMAT_PIC      0       // Disabled by default
 #endif
+#ifdef SUPPORT_FILEFORMAT_PNM
+    #define SUPPORT_FILEFORMAT_PNM      0       // Disabled by default
+#endif
 #ifndef SUPPORT_FILEFORMAT_KTX
     #define SUPPORT_FILEFORMAT_KTX      0       // Disabled by default
 #endif
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -3,22 +3,22 @@
 *   rtext - Basic functions to load fonts and draw text
 *
 *   CONFIGURATION:
-*       #define SUPPORT_MODULE_RTEXT
+*       #define SUPPORT_MODULE_RTEXT 1
 *           rtext module is included in the build
 *
-*       #define SUPPORT_FILEFORMAT_FNT
-*       #define SUPPORT_FILEFORMAT_TTF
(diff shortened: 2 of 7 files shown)


src/config.h
```diff
@@ -252,6 +252,9 @@
 #ifndef SUPPORT_FILEFORMAT_PIC
     #define SUPPORT_FILEFORMAT_PIC      0       // Disabled by default
 #endif
+#ifdef SUPPORT_FILEFORMAT_PNM
+    #define SUPPORT_FILEFORMAT_PNM      0       // Disabled by default
+#endif
 #ifndef SUPPORT_FILEFORMAT_KTX
```
CrackedPixel, line 258: this was missing but im not sure if you want it added 🤷‍♂️

src/rtext.c
```diff
@@ -3,22 +3,22 @@
...
-*       #define SUPPORT_FILEFORMAT_TTF
-*       #define SUPPORT_FILEFORMAT_BDF
+*       #define SUPPORT_FILEFORMAT_FNT 1
+*       #define SUPPORT_FILEFORMAT_TTF 1
+*       #define SUPPORT_FILEFORMAT_BDF 0
 *           Selected desired fileformats to be supported for loading. Some of those formats are
-*           supported by default, to remove support, comment unrequired #define in this module
+*           supported by default, to remove support, #define as 0 in this module or your build system
 *
-*       #define TEXTSPLIT_MAX_TEXT_BUFFER_LENGTH
```
CrackedPixel, line 15: this was incorrectly named here. if you look at the function it is `MAX_TEXT_BUFFER_LENGTH` without the `TEXTSPLIT_` prefix

raysan5:
@CrackedPixel I'm merging this change but note that those comments were added when configuring those features was dependant only on the define, I think the comments should be reworded in a diffrent way to better explain those configuration options. I take note for the future.

src/config.h
```diff
@@ -252,6 +252,9 @@
 #ifndef SUPPORT_FILEFORMAT_PIC
     #define SUPPORT_FILEFORMAT_PIC      0       // Disabled by default
 #endif
+#ifdef SUPPORT_FILEFORMAT_PNM
```
steampuker, line 255: Shouldn't this be #ifndef?

PR #5763, opened 2026-04-15
 Remap stb_vorbis malloc/free calls to RL_MALLOC/RL_FREE

nadav78:
Implements the TODO comment in `src/raudio.c`. `stb_vorbis` was the only audio format not respecting raylib's custom
  allocator hooks, making it consistent with `dr_wav`, `dr_mp3`, and `qoa`.

PR #5765, opened 2026-04-15
Remap stb_vorbis malloc/free calls to RL_MALLOC/RL_FREE

nadav78:
Implements the TODO in `src/raudio.c`. Adds STB_VORBIS_MALLOC and STB_VORBIS_FREE hooks to `stb_vorbis.c`, consistent with the named macro pattern used by dr_wav, dr_mp3, and qoa. Previous PR #5763 used a raw `#define malloc` approach which conflicted with stb_vorbis's internal stdlib.h include.

### DIFF
--- a/src/external/stb_vorbis.c
+++ b/src/external/stb_vorbis.c
@@ -957,13 +957,21 @@ static void *setup_malloc(vorb *f, int sz)
       f->setup_offset += sz;
       return p;
    }
+   #ifdef STB_VORBIS_MALLOC
+   return sz ? STB_VORBIS_MALLOC(sz) : NULL;
+   #else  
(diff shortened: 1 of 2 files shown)


raysan5:
@nadav78 I'm afraid `stb_vorbis.c` is an external library and it shouldn't be modified here but in original repo.

nadav78:
My bad! Is there a preferred approach here? It seems like this would indeed need to be fixed upstream since stb_vorbis doesn't expose allocator hooks like the other libraries.

raysan5:
@nadav78 I'm afraid there are not many options without modifying `stb_vorbis` library, your first approach looked good but it generated some issues...

PR #5766, opened 2026-04-15
[raymath] Refactor `QuaternionFromAxisAngle`

nilsojunior:
Checking if lenght equals 0 inside the if statement is not necessary.


raysan5:
@nilsojunior thanks for the review

PR #5767, opened 2026-04-16
[Examples/Makefile] fix examples makefile define

CrackedPixel:
All of raylib uses `GRAPHICS_API_OPENGL_SOFTWARE` but the example makefile has `GRAPHICS_API_OPENGL_11_SOFTWARE`


raysan5:
@CrackedPixel Good catch! It was changed recently!

PR #5769, opened 2026-04-16
[build-windows.bat] Update build-windows script

areynaldo:
### Problem:
Windows batch build script still referenced utils.c, which no longer exists in src
Script did not detect VS2019 BuildTools vcvarsall path

### Changes:
Added VS2019 BuildTools vcvarsall.bat detection
Removed stale utils.c from RAYLIB_C_FILES

### Impact:
Keeps Windows script aligned with current src layout
Improves MSVC environment detection on systems with VS2019 BuildTools

### Testing:
- Verified diff scope is limited to `build-windows.bat`
- Example builds correctly for release and `-DEBUG`

### DIFF
--- a/projects/scripts/build-windows.bat
+++ b/projects/scripts/build-windows.bat
@@ -26,15 +26,13 @@ REM Checks if cl is available and skips to the argument loop if it is
 REM (Prevents calling vcvarsall every time you run this script)
 WHERE cl >nul 2>nul
 IF %ERRORLEVEL% == 0 goto READ_ARGS
+
 REM Activate the msvc build environment if cl isn't available yet
-IF EXIST "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" (
-  set VC_INIT="C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat"
-) ELSE IF EXIST "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" (
-  set VC_INIT="C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat"
-) ELSE IF EXIST "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" (
-  set VC_INIT="C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat"
-) ELSE IF EXIST "C:\Program Files (x86)\Microsoft Visual C++ Build Tools\vcbuildtools.bat" (
-  set VC_INIT="C:\Program Files (x86)\Microsoft Visual C++ Build Tools\vcbuildtools.bat"
+for /f "tokens=*" %%i in (
+  '"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath 2^>nul'
+) do set VS_PATH=%%i
+IF defined VS_PATH (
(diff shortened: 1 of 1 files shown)


orcmid:
This approach to detection of Visual Studio's developer's CMD environment setup is unreliable and also out of date, considering that we're now at VS 2026.

The appropriate way to accomplish this is to use one of Visual Studio's Developer command prompts for the platform build desired, then run any batch script from that CMD window.  For example, "x64 Native Tools Command Prompt for VS" (depending on the version of VS you've installed - I use the freely-available Native Tools).

One can adapt shortcuts that do this and also start the CMD at a preferred location other than the location of VS Studio code itself.

I made an adaptation of `build-windows.bat` that operates that way, but it does other things that you may not need.  You can find out more at [VCrayApp tool](https://orcmid.github.io/nfoTools/dev/D211101/).   The [VCrayApp.bat script](https://github.com/orcmid/nfoTools/blob/master/devKits/WinDev/VCrayApp/dev/VCrayApp.bat) can be examined directly on GitHub.

This project of mine needs more work to reach version 1.0.0.  I have neglected it while working on a more-important project for me.

areynaldo:
I've updated the script to use `vswhere.exe` instead of hardcoded paths, which handles any VS version automatically (not the ones previous to 2017 afaik).

I think the point of this script is still to auto-initialize the environment so you don't need to start from a Developer Command Prompt or configure any shortcuts. Just copy, run, and it works.

orcmid:
That's a nice fix.  If you look at the script, or try to use it, you'll still be using a command prompt.  The script takes parameters and assumes some things about where things (such as `raylib/src/`) are located.  It also does command-line compiles and linking using the VC/C++ CL.exe command-line compiler. 

Compilation messages and error/warning messages all come to the cmd.exe terminal window.

Side Note: The Visual Studio IDE does this sort of thing by creating batch files that are run "under the covers."  You can see it in the IDE compilation output window that shows build progress.  Of course, in that case, compiler error/warning messages are handled in a nice way, showing you where in the source code they apply and you can click on them to see the code and what the message is about there.

I think you can get the IDE to use the raylib script, just like you can get it to us a `make.`  For myself I simply use the "developer command prompts" of my Visual Studio Build Tools and I use Visual Studio Code for text editing and its support of Intellisense.  That's all free.  But I'm an old-timer and I understand how these tools evolved and what problems they solve.

raysan5:
@areynaldo @orcmid thanks for the improvement and review

orcmid:
@areynaldo @raysan5 That fix depends on the user not having installed VS on a different drive/location, something I did when my C: was very tiny.  That was fixed when I finally upgraded my PC to one with a 2 TB C: drive last month.

 Using the VS-provided shortcuts for developer command prompts avoids a lot, including choice of what kind of build (x64, x86, etc.).  The script requires a lot of customization and that may be beyond a lot of folks who are just starting out.

I don't think there is anything to be done about this under `projects/scripts`.  I suspect most VS users will go the VS Projects route and not know what to do with a CMD script.

areynaldo:
@orcmid 

As I see it and as a user of the script myself, it's more aimed at people who are already comfortable working from the command line and prefer a minimal setup: just a `.bat` file and the compiler, no IDE or other build system involved. That kind of workflow is relatively common in circles with a Handmade Hero / Network mindset, and it has its audience. Such as unity builds or header only libraries. It's just style, has its pros and its cons.

If you're a newcomer or just prefer using the other project options, you'll almost certainly default to a VS or VS Code or a Cmake project setup, and that's totally fine. You probably won't even open that `.bat` file if you are not looking for it. But some do like the simplicity of the one script. It's just another option available.

orcmid:
@areynaldo 
> As I see it and as a user of the script myself, it's more aimed at people who are already comfortable working from the command line and prefer a minimal setup: just a `.bat` file and the compiler, no IDE or other build system involved. 

I am one of those.  I just use the free Visual Studio Build Tools (no IDE), along with Visual Studio Code as my editor of choice.  I use the model behind `build-windows.bat` for my own version that does not require customization and I am now adapting the same technique for non-raylib projects of mine.

But then I'm an old grey-beard and I was around when compiling separate modules that could be linked and made libraries was invented.

PR #5772, opened 2026-04-17
[Platform/RGFW] fix issue with gettime

CrackedPixel:
`GetTime()` was previously broken


raysan5:
@CrackedPixel thanks for the review, I think that `Time.base` variable is confusing because it refers to different values depending on the platform, it could be ticks or nanoseconds. Something to review in the future...

PR #5774, opened 2026-04-17
[Platform/SDL] Add a fallback to SDL joystick for controllers that do not have an SDL mapping

Yonokid:
Some third party controllers do not have recognizable SDL mappings and fail to initialize when plugged in; Opening the controller as a joystick allows for the input to be registered


raysan5:
@Yonokid Sorry but I'm afraid I prefer to avoid adding extra variables and logic to maintain only to support those side cases.

Yonokid:
Ok, thank you for reviewing this

PR #5775, opened 2026-04-17
[Platform/RGFW] fix range in minigamepad.h

CrackedPixel:
The zig build system found this issue


CrackedPixel:
riley said this fix is incorrect, will come back to it later

PR #5779, opened 2026-04-19
Mouse delta calculation with scaling

### LINKED ISSUE #5778: GetMouseDelta not scaled by SetMouseScale
When rendering to different resolutions,  it is suggested that the functions SetMouseScale and SetMouseOffset can be used to correct for differences to actual window size. GetMouseDelta is not taking into account the scaling meaning this has to be done externally. I ran into this while implementing some GUI elements for a project a while ago and continue to bodge around it by storing my own values for scale and creating a GetMouseDeltaScaled function.
It could be solved very simply by modifying GetMouseDelta to:
```
Vector2 GetMouseDelta(void)
{
    Vector2 delta = { 0 };

    delta.x = (CORE.Input.Mouse.currentPosition.x - CORE.Input.Mouse.previousPosition.x)*CORE.Input.Mouse.scale.x;
    delta.y = (CORE.Input.Mouse.currentPosition.y - CORE.Input.Mouse.previousPosition.y)*CORE.Input.Mouse.scale.y;

    return delta;
}
```
Maybe there's a reason it isn't implemented like this already but I couldn't find any previous issues concerning this.

Itwernme:
fixes #5778 
GetMouseDelta scaling issue by adding scaling from SetMouseScale.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -4125,8 +4125,8 @@ Vector2 GetMouseDelta(void)
 {
     Vector2 delta = { 0 };
 
-    delta.x = CORE.Input.Mouse.currentPosition.x - CORE.Input.Mouse.previousPosition.x;
-    delta.y = CORE.Input.Mouse.currentPosition.y - CORE.Input.Mouse.previousPosition.y;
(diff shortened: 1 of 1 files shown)


Itwernme:
Sorry for linking to the wrong issue. Should be correct now

raysan5:
@Itwernme thanks! looks good!

PR #5781, opened 2026-04-22
[dr_mp3] Update to reduce stack size requirements

CrackedPixel:
With the latest drmp3, the stack size no longer needs changed for the audio spectrum visualizer example


raysan5:
@CrackedPixel this change should be applied in dr_mp3 side and then updated the library here, is it the case?

raysan5:
`dr_mp3` was already updated to latest version of official repo.

PR #5782, opened 2026-04-22
[Examples] removed web_basic_window mentions

CrackedPixel:
The examples were failing to build in `cmake` since the example `others/web_basic_window` was removed

reproduce steps:
```bash
# (root folder)
mkdir build
cd build
cmake .. -DPLATFORM=Web
```
```
CMake Error at examples/CMakeLists.txt:157 (add_executable):
  Cannot find source file:

    others/web_basic_window.c


CMake Error at examples/CMakeLists.txt:157 (add_executable):
  No SOURCES given to target: web_basic_window

```


raysan5:
@CrackedPixel thanks for the update! that `raylib_opengl_interop` will also disappear soon...

PR #5785, opened 2026-04-23
BINDINGS: Update raylib-cpp version to 6.0

RobLoach:
The C++ wrapper, [raylib-cpp](https://github.com/RobLoach/raylib-cpp), now works with raylib 6.0: https://github.com/RobLoach/raylib-cpp/releases/tag/v6.0.0


raysan5:
@RobLoach thanks for the update!

PR #5787, opened 2026-04-23
[raudio] add missing UnloadFileData()

### LINKED ISSUE #5786: [raudio] standalone lib specific code missing UnloadFileData()
When RAUDIO_STANDALONE is defined, _raudio.c_ expects _raudio.h_ instead of _raylib.h_. In order to use file handling specific functions without _raylib.h_, it has sections of code dedicated to implement those, inside of a `#if defined (RAUDIO_STANDALONE)` clause. The function `LoadWave()`, implemented in the module itself, allocates some data to be used by `LoadWaveFromMemory()`, and then attempts to free it by calling `UnloadFileData()`, which doesn't exist if you're using it as a standalone library. This results in an error when the module is compiled without the rest of raylib. In contrast, raudio's own repository doesn't have this issue, since it just uses `RL_FREE()`. This is a very small piece of code, and it might not even affect most people, but whoever uses the module separate from raylib will have to fix it themselves as is.

rigidatoms:
Adds missing function necessary for using the module as a standalone library. Fixes #5786.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -434,6 +434,7 @@ static const char *GetFileName(const char *filePath);               // Get point
 static const char *GetFileNameWithoutExt(const char *filePath);     // Get filename string without extension (uses static string)
 
 static unsigned char *LoadFileData(const char *fileName, int *dataSize);    // Load file data as byte array (read)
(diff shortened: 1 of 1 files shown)


raysan5:
@rigidatoms I'm wondering why that function was not defined... but thanks for the review

PR #5789, opened 2026-04-24
Update BINDINGS.md to add vala-vapi version update

edwood-grant:
This is a simple change.

The vala language has updated VAPI bindings to the recently released version 6.0.

Just wanted to update the BINDINGS.md file to reflect that update.

Thanks a a lot,


raysan5:
@edwood-grant thanks for the update!

PR #5791, opened 2026-04-24
[examples] Move window for ball shake

Sebbl0508:
Adds the ability to shake the balls in the `shapes_ball_physics` example via moving/shaking the window.  
I though this is a fun feature for this example. If you want, i can auto-disable this and let it be toggled via a keyboard input.  
  
I guess this may not work for all platforms though.

https://github.com/user-attachments/assets/71fc3ea0-e1de-4e52-9893-4d0157c7657a


raysan5:
@Sebbl0508 Hahaha... this is a nice addition! 😄

PR #5792, opened 2026-04-24
Add moving circle example demonstrating basic animation

thealokverse:
Added a simple shapes example demonstrating basic animation using a moving circle.

This example shows how to:

* Move an object using GetFrameTime()
* Implement basic frame-based animation
* Create smooth motion independent of FPS

The example follows the existing raylib structure and is intended to help beginners understand animation fundamentals.


raysan5:
@thealokverse thanks but there is already a very similar example, I prefer to minimize duplicates.

thealokverse:
got it, thanks for the clarification.

PR #5794, opened 2026-04-24
[tools] various rlparser fixes

gigamonster256:
fixes the gitignores

syncs default path with the Makefile (and fixes broken link in README)

makes selecting CODE mode also compare the null terminator


raysan5:
@gigamonster256 thanks for the review!

PR #5796, opened 2026-04-24
[NppExec] convert all hardcoded paths to dynamic

fgmenth:
This converts all hardcoded paths in NppExec scripts to dynamic to make the installation truly portable and greatly improve the out-of-box experience for new or inexperienced users. This allows running Raylib's npp from any random path or usb drive and all compilation scripts work without needing editing every time.

In order to account for spaces in folder names, some paths needed to be enclosed in quotation marks.

I only tested on windows so far but it seems to work wonderfully:
image


raysan5:
@fgmenth those files are just for reference, no plans to change that for now.

PR #5798, opened 2026-04-25
fix: buffer over-read, integer overflow, NULL deref, and command injection

srpatcha:
### 1. Fix buffer over-read in `TextRemoveSpaces()` (`src/rtext.c`)
The loop checked `text[j]` instead of `text[i]` for the null terminator. When input starts with spaces, `j` stays at 0 while `i` advances, causing `text[j]` to always check `text[0]` instead of the current position — leading to a buffer over-read.

### 2. Add integer overflow guard in `TextToInteger()` (`src/rtext.c`)
Added overflow check before `value = value*10 + digit` to prevent integer overflow when parsing large numeric strings.

### 3. Add NULL check after `RL_MALLOC()` in `WaveCrop()` and `LoadWaveSamples()` (`src/raudio.c`)
Both functions call `RL_MALLOC()` without checking the return value before using the pointer, causing NULL pointer dereference if allocation fails.

### 4. Harden `OpenURL()` against command injection (`src/platforms/rcore_desktop_glfw.c`)
The existing check only rejects single-quote characters. Added checks for double-quote, `&`, `|`, `;`, backtick, `$`, and backslash to prevent command injection via `system()` on all platforms.

### DIFF
--- a/examples/audio/audio_spectrum.c
+++ b/examples/audio/audio_spectrum.c
@@ -1,0 +1,126 @@
+/*******************************************************************************************
+*
+*   raylib [audio] example - Audio Spectrum Analyzer
+*
+*   Example originally created with raylib, using raudio_analyzer module
+*
+*   Example licensed under an unmodified zlib/libpng license, which is an OSI-certified,
+*   BSD-like license that allows static linking with closed source software
+*
+*   Copyright (c) 2024 raylib contributors
+*
+********************************************************************************************/
+
+#include "raylib.h"
+#include "../src/raudio_analyzer.h"
+
+#define SCREEN_WIDTH    800
+#define SCREEN_HEIGHT   450
+#define NUM_BARS         64     // Number of frequency bars to draw
+#define FFT_SIZE        512     // FFT bin count (must be power of 2)
+
+int main(void)
+{
+    // Initialization
+    //--------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Hex4dec:
> The existing check only rejects single-quote characters. Added checks for double-quote, &, |, ;, backtick, $, and backslash to prevent command injection via system() on all platforms.

Can't URLs legitimately contain `&`?

srpatcha:
Great point @Hex4dec! You're right — URLs can legitimately contain \&\ for query parameters (e.g., \?key=val&key2=val2\). Blocking \&\ would break valid URL inputs.

A better approach would be to either:
1. **Escape/quote** the URL argument when passing to \system()\ (e.g., wrap in single quotes on Unix)
2. **Only block characters that start new commands** (\;\, \|\, backtick, \$(...)\) while allowing \&\ in the URL string
3. Use \xecvp()\/\CreateProcess()\ instead of \system()\ to avoid shell interpretation entirely

I'll update the PR to allow \&\ in URLs while still preventing command injection. Thank you for the catch!

srpatcha:
Fixed in commit 674e4c86 - removed & from the rejected characters list. URLs with query parameters (e.g., ?key=val&key2=val2) will now work correctly. The remaining metacharacter checks still prevent command injection. Thank you for catching this!

raysan5:
@srpatcha please, split those changes into separate PRs.

Issue #5793, opened 2026-04-24
[rshapes] Changes to CheckCollisionLines() with raylib 6.0 cause segfaults in previously-working code


luphi:
### Issue description

CheckCollisionLines() has a parameter that outputs the point the lines intersect at:
```
bool CheckCollisionLines(Vector2 startPos1, Vector2 endPos1, Vector2 startPos2, Vector2 endPos2, Vector2 *collisionPoint);
```

This parameter has been optional in the past meaning NULL could be passed if that point wasn't wanted. CheckCollisionLines() previously had a null check:
```
if (collision && (collisionPoint != 0))
{
    collisionPoint->x = xi;
    collisionPoint->y = yi;
}
```

Currently, the first use of 'collisionPoint' is:
```
collisionPoint->x = startPos1.x + t*rx;
collisionPoint->y = startPos1.y + t*ry;
```
The null check is no longer there.

This caused a segfault in code that had previously been working with raylib 5.5 which passed NULL for that parameter.

The descriptions of the function in rshapes and the cheatsheet have not changed and the CHANGELOG includes a couple lines about CheckCollisionLines() but nothing about this parameter. As a result, it was a complete surprise.

Personally, I wish the parameter had remained optional (i.e. NULL could be passed). That's a fairly common pattern, at least in the .NET world I work in. It's probably not as common with C. In lieu of that, I think the descriptions should be updated to mention the parameter cannot be NULL.

### Environment

Observed on Windows 11 25H2 (26200.8246) with a raylib 6.0 build downloaded from itch on 2026/04/23.

### Code Example

I have a project called [raytmx](https://github.com/luphi/raytmx) that uses CheckCollisionLines() to detect collisions in Tiled maps. The change to avoid the segfault is in [commit d76aa7b](https://github.com/luphi/raytmx/commit/d76aa7b3862d417b0de778f482a7e968a2bd260f).

The previously-working line that now causes a segfault:
```
if (CheckCollisionLines(startPos, endPos, currentPoint, nextPoint, NULL)) return true;
```

The new lines that prevent segfaults:
```
Vector2 collisionPoint = { 0 };
if (CheckCollisionLines(startPos, endPos, currentPoint, nextPoint, &collisionPoint)) return true;
```

raysan5:
@bielern it seems your PR broke this functionality, please, could you review it?

bielern:
Hi
Yes, will have a look asap.

raysan5:
[...]

PR #5802, opened 2026-04-25
Fix: Add NULL check to CheckCollisionLines

bielern:
Fixes the segfault issue when passing in a NULL for the collisionPoint

https://github.com/raysan5/raylib/issues/5793


raysan5:
@bielern thanks for the review

bielern:
np - thank you!

PR #5803, opened 2026-04-25
[examples] Improve `core_smooth_pixelperfect`

rose-mtz:
Added overscan and smoothing toggle, so that the user can quickly and easily see the problem and solution of pixel-perfect rendering.


raysan5:
@rose-mtz thanks for the improvement, why the info about world and render size was removed from UI? I think it was useful.

rose-mtz:
> @rose-mtz thanks for the improvement, why the info about world and render size was removed from UI? I think it was useful.

I think, its easy to see from code what the render size is, and it doesn't change through out program. But, I could add it back if you'd like.

raysan5:
@rose-mtz yes, please, add it back

raysan5:
@rose-mtz Nice! Thanks for the update!

PR #5804, opened 2026-04-25
[build] Shift `__cplusplus` check to a higher level for `bool` type define

lamweilun:
The main reason for this change is that, C++ compilers, or at least g++ and clang++ from my own testing, defines __STDC__ but not __STDC_VERSION__. And when compiling with -Wundef, it would trigger an error when its checking for __STDC_VERSION__.

This change basically says, check if you're a C++ compiler first before even entering any of the bool type definition since C++ has bool as a built-in type.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -203,11 +203,13 @@
 // Types and Structures Definition
 //----------------------------------------------------------------------------------
 // Boolean type
-#if (defined(__STDC__) && (defined(__STDC_VERSION__) && __STDC_VERSION__) >= 199901L) || (defined(_MSC_VER) && _MSC_VER >= 1800)
+#if !defined(__cplusplus)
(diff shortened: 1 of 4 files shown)


lamweilun:
The main reason for this change is that, C++ compilers, or at least g++ and clang++ from my own testing, defines `__STDC__` but not `__STDC_VERSION__`. And when compiling with `-Wundef`, it would trigger an error when its checking for `__STDC_VERSION__`.

This change basically says, check if you're a C++ compiler first before even entering any of the bool type definition since C++ has bool as a built-in type.

raysan5:
@lamweilun Please,  could you split those changes into two separate PRs, also, raymath convention is avoiding functions calling other functions, please, review that, thanks!

lamweilun:
@raysan5 Thanks for the review. I have kept this to just the shifting of the C++ check

raysan5:
@lamweilun thanks for the review... wondering why it was not reported before...

PR #5808, opened 2026-04-26
[RAUDIO] Update to latest DR libs to fix some warnings in MSVC

JeffM2501:
This PR updates the 3 external DR libs to the current version from https://github.com/mackron/dr_libs where an upstream fix has been applied to remove a warning in the dr_wav.h code.
Tested the audio examples and they work just fine.


raysan5:
@JeffM2501 Nice! Thanks for the update!

PR #5810, opened 2026-04-26
[rcore][Android] Allow setting window size on Android.

b1bradders:
# My situation

1. User rotates device while app is open.
2. Android sends `APP_CMD_CONFIG_CHANGED`.
3. Resizing the framebuffer and setting screen sizes to reflect the new 'window size' causes stretching due to Android-reported width and height still **(usually - see below)** reporting with values from before the orientation change.

## Why not just add this in `AndroidCommandCallback`'s `APP_CMD_CONFIG_CHANGED` case?
Android does not send the correct width and height in time, therefore leading to you having to manually switch the height and width. This is fine, until you notice that Android has a somewhat-common chance to send the correct width and height and then everything gets stretched because you swapped the width to be the height to account for the wrong timing.

You could make a variable to mark the window as dirty then in a place that is triggered every frame, if that variable is true and the Android-reported size != current window size, set the variable to false and then update to the new size.

## The solution for devs using raylib
1. Allow `SetWindowSize` to be usable, via this patch, on android.
2. Devs can now use the NDK themselves to check every frame if a resize occured:
```c
extern struct android_app* GetAndroidApp();

static int ndk_window_width() {
	struct android_app* app = GetAndroidApp();
	if (!app || !app->window) return 0;

    return ANativeWindow_getWidth(app->window);
}
static int ndk_window_height() {
    struct android_app* app = GetAndroidApp();
    if (!app || !app->window) return 0;

    return ANativeWindow_getHeight(app->window);
}
```
3. Devs call these every frame, and do manual resize checking. If a resize is detected, call `SetWindowSize(int width, int height)`.

## Additional comments
Let me know if this is okay - personally I wouldn't accept this PR as I'm sure there is a solution that allows putting this code in `APP_CMD_CONFIG_CHANGED` while receiving correct sizes from Android (see my 'Why not ...' above for why we can't do this already).

This also is a better solution as it also correctly sets the size when INIT_WINDOW happens. So, if the app is in landscape and the users turns off the phone screen then turns it back on with the orientation in portrait, then that is also handled as the size change is detected.

## Demos
Fix from this PR:

https://github.com/user-attachments/assets/a118dbe4-5482-489b-a3fc-1e41d63ce05b

Default behaviour:

https://github.com/user-attachments/assets/66a6fc0d-cff7-4c6f-8032-63f9519a248b

### DIFF
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -465,13 +465,10 @@ void SetWindowSize(int width, int height)
     CORE.Window.screen.width = width;
     CORE.Window.screen.height = height;
 
-    EGLint displayFormat = 0;
-    eglGetConfigAttrib(platform.device, platform.config, EGL_NATIVE_VISUAL_ID, &displayFormat);
-
     ANativeWindow_setBuffersGeometry(platform.app->window,
         CORE.Window.render.width + CORE.Window.renderOffset.x,
         CORE.Window.render.height + CORE.Window.renderOffset.y,
-        displayFormat);
+        0);
 }
 
 // Set window opacity, value opacity is between 0.0 and 1.0
--- a/src/platforms/rcore_android.c
+++ b/src/platforms/rcore_android.c
@@ -455,7 +455,23 @@ void SetWindowMaxSize(int width, int height)
 // Set window dimensions
 void SetWindowSize(int width, int height)
 {
-    TRACELOG(LOG_WARNING, "SetWindowSize() not available on target platform");
+	CORE.Window.display.width = width;
+    CORE.Window.display.height = height;
+
+    SetupViewport(width, height);
+    CORE.Window.currentFbo.width = width;
+    CORE.Window.currentFbo.height = height;
+
+    CORE.Window.screen.width = width;
+    CORE.Window.screen.height = height;
+
+    EGLint displayFormat = 0;
+    eglGetConfigAttrib(platform.device, platform.config, EGL_NATIVE_VISUAL_ID, &displayFormat);
+
+    ANativeWindow_setBuffersGeometry(platform.app->window,
+        CORE.Window.render.width + CORE.Window.renderOffset.x,
+        CORE.Window.render.height + CORE.Window.renderOffset.y,
+        displayFormat);
 }
 
 // Set window opacity, value opacity is between 0.0 and 1.0


raysan5:
@Bigfoot71 since you have more experience with Android, please, could you take a quick look to this PR?

src/platforms/rcore_android.c
```diff
@@ -455,7 +455,23 @@ void SetWindowMaxSize(int width, int height)
...
+
+    SetupViewport(width, height);
+    CORE.Window.currentFbo.width = width;
+    CORE.Window.currentFbo.height = height;
+
+    CORE.Window.screen.width = width;
+    CORE.Window.screen.height = height;
+
+    EGLint displayFormat = 0;
+    eglGetConfigAttrib(platform.device, platform.config, EGL_NATIVE_VISUAL_ID, &displayFormat);
```
Bigfoot71: Normally there is no need to retrieve `displayFormat`, if you specify 0 in `format` to `ANativeWindow_setBuffersGeometry()` it is the already defined one that remains.

See: https://developer.android.com/ndk/reference/group/a-native-window#group___a_native_window_1ga155e5218b14386baed6cb833e5842108

Bigfoot71:
@raysan5 I'm not aware of the problem described, I'd have to dust off raymob and test that _(I will do it soon, given the release of 6.0)_.

But as it stands, it seems a bit tricky to me. I will give better feedback as soon as I can.

src/platforms/rcore_android.c
b1bradders: Ah thanks, I did look at the function on the NDK docs but I guess I just skimmed over that part without realising I could do that. 😅

b1bradders:
It'd be great if you could test for 6.0 - I was testing this using [raylib-go](https://github.com/gen2brain/raylib-go) which is still on the last released 5.x version :)

<sub>_Also: sorry for updating after 4 days - I don't have the GitHub app or GH notification emails enabled, so I only see notifications when I next come onto the website lol_</sub>

b1bradders:
@Bigfoot71 @raysan5 Hey guys, any updates? 😁

raysan5:
@Bigfoot71 any update on this issue?

jn-jairo:
I'm not an expert in android but the [`documentation`](https://developer.android.com/reference/games/game-activity/group/android-native-app-glue#group__android__native__app__glue_1ggab1238b3017ceefa7fdfbcc79eee6c289a448514d4220320ec7a7e7324e019f42c) has the `APP_CMD_WINDOW_RESIZED` event but you tried the `APP_CMD_CONFIG_CHANGED`.

As I understand in this case the `APP_CMD_CONFIG_CHANGED` is sent when the orientation changes, what implies a window size change, but the actual window resize may not be done yet, and when it is done, the `APP_CMD_WINDOW_RESIZED` is sent.

What you described sounds like a racing condition for not using the proper event.

b1bradders:
Oh interesting! I believe I did try `APP_CMD_WINDOW_RESIZED` (i'm 60% sure i didn't _not_ try it) although I must've still had issues. When I get the chance to though, I'll 100% try it.

b1bradders:
Before I test, is this really the best way to resize the raylib "window"/screen?

```c++
CORE.Window.display.width = width;
CORE.Window.display.height = height;

SetupViewport(width, height);
CORE.Window.currentFbo.width = width;
CORE.Window.currentFbo.height = height;

CORE.Window.screen.width = width;
```
(where `width` and `height` is the new width and height)

I only ask because that's a lot of different variables (`.display`, `.currentFbo`, `.screen`) that I'm changing the width and height of and, tbh, I have no clue what `currentFbo`'s dimens. are affecting and I don't really know the difference between changing `.screen`'s w&h and `.display`'s w&h (if that makes sense?)

raysan5:
I'm not sure about the approach of this PR so  I'm closing it.

b1bradders:
@raysan5 Yeah, that's fair tbh...

If I ever get around to testing this again, and manage to make a decent solution for this issue, I'll make a (hopefully better) PR. 👍🏻

PR #5811, opened 2026-04-27
[examples] Update build.zig to support wasm examples

shadeops:
Prior this PR, while the provided `build.zig` compiles the examples without errors, the resources for each example are not provided to Emscripten. This leaves the examples in a compiled but semi broken state when actually run due to missing textures and shaders.

This PR introduces a `examples/example_resources.zon` file which contains a list of resources that is needed for each example. The list of resources was generated from the existing `examples/Makefile.Web`.

During the `build.zig` execution the `example_resources.zon` is imported and if the current example has a resource dependency, they are passed to the Emscripten compile step as preloads.

Tested on:

- Fedora 43 with Zig 0.16.0
- Windows 10 with Zig 0.16.0

`zig build -Dtarget=wasm32-emscripten textures_polygon_drawing`

Master branch / 6.0 tag:
master

PR:
pr


raysan5:
@michaelfiber, @Not-Nik, @HaxSam please, could you take a look to this PR?

build.zig
```diff
@@ -706,10 +706,16 @@ fn addExamples(
             const emcc_flags = emsdk.emccDefaultFlags(b.allocator, .{ .optimize = optimize });
             const emcc_settings = emsdk.emccDefaultSettings(b.allocator, .{ .optimize = optimize });
 
+            const EmccExamplesPreloadMap = std.static_string_map.StaticStringMap([]const emsdk.zemscripten.EmccFilePath);
+            const EmccExamplesPreloadKV = struct { []const u8, []const emsdk.zemscripten.EmccFilePath };
+            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @import("examples/example_resources.zon");
+            const emcc_examples_preloads_map = EmccExamplesPreloadMap.initComptime(emcc_examples_preloads);
+
```
Not-Nik, line 713: I'd move this outside of the loop.

Not-Nik (approved):
Didn't do any testing if this actually works, but given the benefit of the doubt; r+

build.zig
shadeops: Thanks for flagging!

Since the StaticStringMap creation is fully comptime the cost of creating the map is only paid for once. So I opted to have the code block creating the map close to the actual usage site.

That said, since I was curious, I tested on [Godbolt](https://godbolt.org/z/33x1fTf99) and having those 4 lines does result in 2 mov instructions per loop. So cheap, but not completely free.

raysan5:
@shadeops thanks for the improvement, I'm merging, feel free to send another PR if it requires some review.

HaxSam:
Looks good! thx for the help

PR #5812, opened 2026-04-27
fix(rcore): correct string boundary handling

zhaob1n:
I compiled the latest git and gcc15 -O3 gives me:
```
rcore.c: In function ‘IsFileExtension’:
rcore.c:2380:9: warning: ‘strncpy’ output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
 2380 |         strncpy(extList, ext, extLength);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
rcore.c:2377:30: note: length computed here
 2377 |         int extLength = (int)strlen(ext);
      |                              ^~~~~~~~~~~
rcore.c: In function ‘ScanDirectoryFiles’:
rcore.c:4303:25: warning: ‘strncpy’ output may be truncated copying 4095 bytes from a string of length 4095 [-Wstringop-truncation]
 4303 |                         strncpy(files->paths[files->count], path, MAX_FILEPATH_LENGTH - 1);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
rcore.c:4311:25: warning: ‘strncpy’ output may be truncated copying 4095 bytes from a string of length 4095 [-Wstringop-truncation]
 4311 |                         strncpy(files->paths[files->count], path, MAX_FILEPATH_LENGTH - 1);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
rcore.c:4303:25: warning: ‘strncpy’ output may be truncated copying 4095 bytes from a string of length 4095 [-Wstringop-truncation]
 4303 |                         strncpy(files->paths[files->count], path, MAX_FILEPATH_LENGTH - 1);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This pr fixes them and other issues I found:
1. misunderstanding about the size parameter of `strncpy`, `snprintf`, `getcwd`
2. fileExtLower may not end with '\n' without this pr.
3. seemingly wrong trace log

IMO, there's a LOT to improve in terms of the safety and performance of the string handling in `rcore.c`, this pr just contains some quick fixes.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -2004,7 +2004,7 @@ unsigned char *LoadFileData(const char *fileName, int *dataSize)
                     {
                         *dataSize = (int)count;
 
-                        if ((*dataSize) != size) TRACELOG(LOG_WARNING, "FILEIO: [%s] File partially loaded (%i bytes out of %i)", fileName, dataSize, count);
+                        if ((*dataSize) != size) TRACELOG(LOG_WARNING, "FILEIO: [%s] File partially loaded (%i bytes out of %i)", fileName, *dataSize, size);
                         else TRACELOG(LOG_INFO, "FILEIO: [%s] File loaded successfully", fileName);
                     }
                 }
@@ -2365,8 +2365,7 @@ bool IsFileExtension(const char *fileName, const char *ext)
     {
         int fileExtLength = (int)strlen(fileExt);
         char fileExtLower[16] = { 0 };
-        char *fileExtLowerPtr = fileExtLower;
-        for (int i = 0; (i < fileExtLength) && (i < 16); i++)
+        for (int i = 0; (i < fileExtLength) && (i < 15); i++)
         {
             // Copy and convert to lower-case
             if ((fileExt[i] >= 'A') && (fileExt[i] <= 'Z')) fileExtLower[i] =  fileExt[i] + 32;
@@ -2377,7 +2376,7 @@ bool IsFileExtension(const char *fileName, const char *ext)
         int extLength = (int)strlen(ext);
         char *extList = (char *)RL_CALLOC(extLength + 1, 1);
         char *extListPtrs[MAX_FILE_EXTENSIONS] = { 0 };
-        strncpy(extList, ext, extLength);
+        memcpy(extList, ext, extLength);
         extListPtrs[0] = extList;
 
         for (int i = 0; i < extLength; i++)
@@ -2386,11 +2385,15 @@ bool IsFileExtension(const char *fileName, const char *ext)
             if ((extList[i] >= 'A') && (extList[i] <= 'Z')) extList[i] += 32;
 
             // Get pointer to next extension and add null-terminator
-            if ((extList[i] == ';') && (extCount < (MAX_FILE_EXTENSIONS - 1)))
+            if (extList[i] == ';')
             {
                 extList[i] = '\0';
-                extListPtrs[extCount] = extList + i + 1;
-                extCount++;
+
+                if (extCount < MAX_FILE_EXTENSIONS)
+                {
+                    extListPtrs[extCount] = extList + i + 1;
+                    extCount++;
+                }
             }
         }
 
@@ -2398,7 +2401,7 @@ bool IsFileExtension(const char *fileName, const char *ext)
         {
             // Consider the case where extension provided
             // does not start with the '.'
-            fileExtLowerPtr = fileExtLower;
+            char *fileExtLowerPtr = fileExtLower;
             if (extListPtrs[i][0] != '.') fileExtLowerPtr++;
 
             if (strcmp(fileExtLowerPtr, extListPtrs[i]) == 0)
@@ -2611,7 +2614,7 @@ const char *GetWorkingDirectory(void)
     static char currentDir[MAX_FILEPATH_LENGTH] = { 0 };
     memset(currentDir, 0, MAX_FILEPATH_LENGTH);
 
-    char *path = GETCWD(currentDir, MAX_FILEPATH_LENGTH - 1);
+    char *path = GETCWD(currentDir, MAX_FILEPATH_LENGTH);
 
     return path;
 }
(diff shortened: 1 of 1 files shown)


src/rcore.c
```diff
@@ -2958,9 +2961,9 @@ unsigned int GetDirectoryFileCountEx(const char *basePath, const char *filter, b
             {
                 // Construct new path from our base path
                 #if defined(_WIN32)
-                    int pathLength = snprintf(path, MAX_FILEPATH_LENGTH - 1, "%s\\%s", basePath, entity->d_name);
+                    int pathLength = snprintf(path, MAX_FILEPATH_LENGTH, "%s\\%s", basePath, entity->d_name);
```
raysan5, line 2964: I don't understand this change, it could return a non-null terminated string, please, can you confirm?

zhaob1n: the `size` parameter of snprintf includes(guarantees) the terminating null byte. See [cppreference](https://cppreference.com/c/io/fprintf):
> 4) Writes the results to a character string buffer. At most bufsz - 1 characters are written. The resulting character string will be terminated with a null character, unless bufsz is zero. If bufsz is zero, nothing is written and buffer may be a null pointer, however the return value (number of bytes that would be written not including the null terminator) is still calculated and returned.

actually, if you look at the signature of the function from man:
```c
     int snprintf(size_t size;
                 char str[restrict size], size_t size,
                 const char *restrict format, ...);
```
`char str[restrict size]` means the buffer's size should be at least `size`

The same goes for strncpy:
```
     char *strncpy(size_t dsize;
                   char dst[restrict dsize], const char *restrict src,
                   size_t dsize);
```


that's why the warning shows
```
rcore.c:4303:25: warning: ‘strncpy’ output may be truncated copying 4095 bytes from a string of length 4095 [-Wstringop-truncation]
 4303 |                         strncpy(files->paths[files->count], path, MAX_FILEPATH_LENGTH - 1);
```
the correct way should be `strncpy(files->paths[files->count], path, MAX_FILEPATH_LENGTH);`
But ofc memcpy is more efficient.

Also by the way, `strncpy` is a design flaw, linux kernel removed all the use of `strncpy` recently, because:
1. it doesn't guarantee null termination without care, unlike snprintf
2. it pads null bytes when `size` is not reached, which is inefficient.

That's why glibc recently introduced bsd-style `strlcpy` and `strlcat`.

orcmid: @zhaob1n 
I don't understand the ";" inside those prototypes, along with duplication of `size_t dsize`).

This is the prototype in the ISO Standard C11 Synopsis for `strncpy`:

```C
     #include <string.h>
     char *strncpy(char * restrict s1,
          const char * restrict s2,
          size_t n);
```
It is correct that there is no guarantee of a null terminator.  A practical solution is to have an array `char s1[n+1]` and ensure that `s1[n]=='\0'` (usually accomplished by initialization of `s1[]`).

The inefficiency of '\0'-padding is a context-less concern to weigh against the actual situation, especially for code to be maintained/site-checked/understood by others, including novices.  The best way to learn defensive coding is *first*.

raysan5: @zhaob1n thanks for the details, I carefully tested `strncpy()` and `sprintf()` and saw the potential issues. Quite worrying because it is used in multiple places in raylib code, afaik.

raysan5:
@zhaob1n thank you very much for the review and the detailed explanations on potential issues with `strncpy()`.

Just a side note, I detected different behaviour with `snprintf()` using GCC vs MSVC, the function is supposed to return the number of bytes to be potentially written, so, if copied string has no null terminator, in case of MSVC returned value is its length (> MAX_STR_LEN) while on GCC it returns MAX_STR_LEN.

orcmid:
@raysan5 @zhaob1n 
> Just a side note, I detected different behaviour with `snprintf()` using GCC vs MSVC, the function is supposed to return the number of bytes to be potentially written, so, if copied string has no null terminator, in case of MSVC returned value is its length (> MAX_STR_LEN) while on GCC it returns MAX_STR_LEN.

In the ISO C11 specification, this is all that is said about the value returned from `snprintf`:

>The **snprintf** function returns the number of characters that would have been written
had **n** been sufficiently large, not counting the terminating null character, or a negative value if an encoding error occurred. Thus, the null-terminated output has been completely written if and only if the returned value is nonnegative and less than **n**.

Note that it is `size_t n` but the return from snprint is an `int` (and negative values are possible).  It appears that an `\0` is only written if the sufficient length is less than **n** chars as reflected above.  There is no mention of `MAX_STR_LEN` anywhere in the ISO C 11 specification that I am looking at.

PS: These functions are not in C98.

PPS: My previous precautionary suggestion about having the output array being (at least) one larger than **n** and having one (or more) protective `'\0'` seems extremely applicable here, avoiding a hard to debug self-exploit.

zhaob1n:
@raysan5 @orcmid 
> @zhaob1n thank you very much for the review and the detailed explanations on potential issues with `strncpy()`.
> 
> Just a side note, I detected different behaviour with `snprintf()` using GCC vs MSVC, the function is supposed to return the number of bytes to be potentially written, so, if copied string has no null terminator, in case of MSVC returned value is its length (> MAX_STR_LEN) while on GCC it returns MAX_STR_LEN.

No, they have the same behavior, see https://godbolt.org/z/7e5o5Wqve
image

I have also commented on your tweets :), basically your code contains UB so everything can happen.

> @raysan5 @zhaob1n
> 
> > Just a side note, I detected different behaviour with `snprintf()` using GCC vs MSVC, the function is supposed to return the number of bytes to be potentially written, so, if copied string has no null terminator, in case of MSVC returned value is its length (> MAX_STR_LEN) while on GCC it returns MAX_STR_LEN.
> 
> In the ISO C11 specification, this is all that is said about the value returned from `snprintf`:
> 
> > The **snprintf** function returns the number of characters that would have been written
> > had **n** been sufficiently large, not counting the terminating null character, or a negative value if an encoding error occurred. Thus, the null-terminated output has been completely written if and only if the returned value is nonnegative and less than **n**.
> 
> Note that it is `size_t n` but the return from snprint is an `int` (and negative values are possible). It appears that an `\0` is only written if the sufficient length is less than **n** chars as reflected above. There is no mention of `MAX_STR_LEN` anywhere in the ISO C 11 specification that I am looking at.
> 
> PS: These functions are not in C98.


No, snprintf guarantees the terminating '\n' as long as the `n` is not zero, here is the document from C99 standard.

image

here is C23:
image

> 
> PPS: My previous precautionary suggestion about having the output array being (at least) one larger than **n** and having one (or more) protective `'\0'` seems extremely applicable here, avoiding a hard to debug self-exploit.

That only confuses things. The C rule is really simple: a buffer and the size of the buffer are given together, and the function only uses buffer[0] to buffer[size-1].

orcmid:
@raysan5 @zhaob1n  
> No, snprintf guarantees the terminating '\n' as long as the `n` is not zero, here is the document from C99 standard.
[ ... ]
> That only confuses things. The C rule is really simple: a buffer and the size of the buffer are given together, and the function only uses buffer[0] to buffer[size-1].

You're absolutely correct.  If I read the preceding paragraph and not just the one about the return value, I would have seen that in the C11 specification also.

From that specification, there is no need for some guard `'\0'` chars.

PR #5814, opened 2026-04-28
[rtext] Remove `strcmp()` dependency

nilsojunior:


raysan5:
@nilsojunior no need for this change, thanks

PR #5815, opened 2026-04-28
[raymath] Add missing float*Vector multiplication C++ operator overloads

alphahex99:
Is there a reason **raymath.h** only supports `Vector` * `float` C++ math operators, **but `float` * `Vector` isn't supported**?

I'm talking about just swapping the vector and scalar multiplication order like so:

```cpp
#include "raymath.h"

  Vector2 velocity;

  Vector2 delta = ... ;
  float overlapLen = ... ;

  // This works as intended
  velocity += delta * overlapLen * 0.5f;

  // THIS DOESN'T COMPILE
  velocity += 0.5f * delta * overlapLen;
```

```
Example1.cpp: error: no match for 'operator*' (operand types are 'float' and 'Vector2')
   69 |                                       velocity += 0.5f * delta * overlapLen;
      |                                                   ~~~~ ^ ~~~~~
      |                                                   |      |
      |                                                   float  Vector2
```

I kept finding myself constantly writing `float` constants first before vectors and wondering why a build failed before remembering *Oh yeah, raymath.h doesn't allow that for some reason...* over and over

Extra annoying if it's not just some simple constant like the 0.5 above, but a `float` variable (in my examples the "overlapLen" variable):

```cpp
  // This works as intended
  velocity += delta * overlapLen * 0.5f;

  // THIS DOESN'T COMPILE
  velocity += overlapLen * delta * 0.5f;
```

```
Example2.cpp: error: no match for 'operator*' (operand types are 'float' and 'Vector2')
   69 |                                       velocity += overlapLen * delta * 0.5f;
      |                                                   ~~~~~~~~~~ ^ ~~~~~
      |                                                   |            |
      |                                                   float        Vector2
```

Constantly keeping an eye on this weird arbitrary order of operations issue is unnecessary.

In math, **multiplying a vector by scalars is commutative** and there is **nothing wrong with writing the scalar first**.

In fact, **it is the more common way of doing it**, even for matrices:

image

image

source: https://en.wikipedia.org/wiki/Scalar_multiplication

---

Is there like a philosophy or formatting reason for not doing this? Or to keep a strict rule of `Vector` or `Matrix` **always** being the first parameter in the `VectorNDoSomething(VectorN ...)` and `MatrixDoSomething(Matrix ...)` pure C functions?

Feel free to close the Pull Request if there is some reason against this.

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -2802,6 +2802,11 @@ inline const Vector2& operator -= (Vector2& lhs, const Vector2& rhs)
     return lhs;
 }
 
+inline Vector2 operator * (const float& lhs, const Vector2& rhs)
+{
+    return Vector2Scale(rhs, lhs);
+}
+
 inline Vector2 operator * (const Vector2& lhs, const float& rhs)
 {
     return Vector2Scale(lhs, rhs);
@@ -2894,6 +2899,11 @@ inline const Vector3& operator -= (Vector3& lhs, const Vector3& rhs)
 {
     lhs = Vector3Subtract(lhs, rhs);
     return lhs;
+}
+
+inline Vector3 operator * (const float& lhs, const Vector3& rhs)
+{
+    return Vector3Scale(rhs, lhs);
 }
 
 inline Vector3 operator * (const Vector3& lhs, const float& rhs)
@@ -2989,6 +2999,11 @@ inline const Vector4& operator -= (Vector4& lhs, const Vector4& rhs)
 {
     lhs = Vector4Subtract(lhs, rhs);
     return lhs;
+}
+
+inline Vector4 operator * (const float& lhs, const Vector4& rhs)
+{
+    return Vector4Scale(rhs, lhs);
 }
 
 inline Vector4 operator * (const Vector4& lhs, const float& rhs)


raysan5:
@JeffM2501 please, can you review this addition? do you think it is really needed? I only see more code to maintain...

JeffM2501:
> @JeffM2501 please, can you review this addition? do you think it is really needed? I only see more code to maintain...

I will take a look, I understand the general issue.

JeffM2501:
So yeah, I specifically didn't do this ordering when doing the C++ operators.
My thought process was.
1) match the order of the C API
2) the vector is the more important thing, so it should be first, since the result is a vector, it should be closest to the =, not a float

That said, I can see there may be a few users/cases where doing it backwards could read better. C++ will not reverse the order of operations for you.

Maintenance wise, this still just calls the raymath functions, so I don't expect there to be a huge maintenance burden, we won't be going in and changing this wrapper code ever again.

I'm on the fence, I'd personally never put the vector first, but that may just be my personal preference for readability, so I leave it up to you. I have no issues with this PR code, it's really just a question of API design, if you want to allow something like 

```
Vector2 pos = 10.0f * GetObstaclePos();
```
To me this is more prone to mistakes, if you remove the end part, it's now invalid.
Where
```
Vector2 pos = GetObstaclePos() * 10.0f;
```
Doesn't have that issue, and matches the order of the C version.
```
Vector2 pos = Vector2Scale(GetObstaclePos(), 10.0f);
```

alphahex99:
If this is added, its usage is entirely voluntary and nothing new to maintain since it's so simple, but a nice QoL feature for people stealing all the vector/matrix math from the "raymath.h" header in their projects like myself.

It just feels weird and wrong for someone with a math background to write some scalar constant you're multiplying vectors by at the very end of a physics equation 🥲 

---

Any reason there's no float addition and multiplication operators for matrices? I could add those C functions and C++ operators too if you want. I assume there's not much use for it in the usual rendering transforms and rotations?

image

raysan5:
@JeffM2501 thanks for the review, it does not seem to be a dramatic change, let's add it...

@alphahex99 thanks for the addition.

PR #5817, opened 2026-04-28
[raymath] Adds missing macro define for `RAYMATH_USE_SIMD_INTRINSICS`

lamweilun:
Missing `RAYMATH_USE_SIMD_INTRINSICS` define causes compiler errors when it needs to be checked later on with a `#if`

### DIFF
--- a/src/raymath.h
+++ b/src/raymath.h
@@ -84,6 +84,11 @@
     #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


raysan5:
@lamweilun Afaik, any undefined macro is considered as 0, so, why this addition is required?

raysan5:
It seems those lines should not be required.

PR #5827, opened 2026-04-30
[rlsw] ESP32 optimizations

jensroth-git:
# [rlsw] ESP32 / Xtensa hot-path optimizations (sw_rcp + ESP-DSP + opt-in POT wrap)

## Summary

Three small, isolated, opt-out-safe optimizations to `rlsw.h` that together
roughly **halve frame time** on ESP32-S3 for textured 3D rendering, while
staying bit-for-bit identical on existing desktop / RISC-V / non-ESP32 builds.

These were extracted from a larger ESP32 port experiment after a code review
with @Bigfoot71, who suggested upstreaming the parts related to `sw_rcp` and
`dspm_mult_4x4x?_f32` while leaving the rest (clear, conversions, sampler
refactor, async double-buffer) for a separate effort.

The PR is split into three logically-isolated commits so each can be reviewed
or reverted independently.

---

### 1. `sw_rcp` — Xtensa `recip0.s` fast reciprocal *(commit 1)*

Float division on Xtensa LX6 / LX7 (ESP32, ESP32-S3) compiles to a software
`__divsf3` call. The rasterizer performs multiple `1.0f/x` operations per
triangle setup, per 16-pixel affine block, and per vertex transform.

A new `sw_rcp(x)` helper emits the hardware `recip0.s` seed plus two
Newton-Raphson refinement steps — 1-ULP accurate in ~7 instructions, all in
FPU registers. On every other target it expands to plain `1.0f/x`, so
generated code is **byte-identical to before** for non-Xtensa builds.

Applied only to documented hot-path reciprocals:
- perspective divide (`1/w`) in clip-and-project (PCT and PC paths)
- line/point clip-and-project NDC conversion
- triangle span: `dxRcp`, `blockLenRcp`, `wRcpA`, `wRcpB`
- triangle scanline: `h02Rcp`, `h01Rcp`, `h12Rcp`
- axis-aligned quad: `wRcp`, `hRcp`
- line rasterizer: `stepRcp`

Other `1.0f/x` uses (`sw_matrix_translate` length, texture init `tx`/`ty`)
are not on the per-pixel hot path and are left untouched.

### 2. ESP-DSP matrix kernels *(commit 2)*

ESP-DSP is ESP-IDF's official optimized math library and ships hand-vectorized
kernels for the matrix sizes rlsw uses. Two integration points:

- `sw_matrix_mul_rst` → `dspm_mult_4x4x4_f32` — used for MVP build, lookat,
  push/multiply, etc. The flat-buffer call still produces the correct
  column-major product (transpose-of-transposes equivalence; see comment).

- `sw_immediate_push_vertex` → `dspm_mult_4x4x1_f32` — the per-vertex clip
  transform. ESP-DSP wants a row-major matrix here, so a `matMVP_rm[16]`
  row-major copy is maintained alongside `matMVP` and refreshed once per
  `isDirtyMVP` rebuild in `sw_immediate_begin`.

Detection is **opt-in** via `SW_USE_ESP_DSP` so existing ESP-IDF projects
that don't depend on the `esp-dsp` component keep building unchanged.
A user enables it from CMakeLists.txt (or anywhere before including rlgl.h):

```cmake
target_compile_definitions(${COMPONENT_LIB} PRIVATE SW_USE_ESP_DSP=1)
```

and adds the dependency to `idf_component.yml`:

```yaml
dependencies:
  espressif/esp-dsp: "^1.4.0"
```

### 3. `SW_TEXTURE_REPEAT_POT_FAST` — opt-in POT bitmask wrap *(commit 3)*

Addresses the long-standing `// NOTE: If the textures are POT, avoid the
division for SW_REPEAT` TODO in `sw_texture_sample_linear`.

When defined, textures whose width/height are powers of two use a bitmask
wrap (`x & (size-1)`) instead of `floorf`-based fractional wrap (nearest)
or the signed `(x % w + w) % w` chain (linear). NPOT textures keep using
the original paths via a runtime `(size & (size-1)) == 0` check, so
SW_REPEAT remains correct for them.

**Off by default**: for POT textures sampled with negative UV coordinates,
bitmask wrap (two's complement) can differ from `sw_fract` wrap by one
texel at the boundary. Imperceptible at typical resolutions but technically
a behavior change, so existing users get bit-for-bit identical output.

---

## Numbers

Measured on ESP32-S3 @ 240 MHz, 240×240 R5G6B5 framebuffer, textured 3D model
with depth testing, all three optimizations enabled:

| Phase | Before | After (this PR — render only) |
|-------|-------:|--:|
| Frame `model` time | ~44 ms | ~39 ms |

(Larger frame-time numbers in the original write-up include a separate
async-double-buffer change in the platform layer that's not part of this PR.)

## Risk / compatibility

- **Non-ESP32 builds**: no behavior change. All new code paths sit behind
  `#if defined(__XTENSA__)`, `#ifdef SW_HAS_ESP_DSP`, or
  `#ifdef SW_TEXTURE_REPEAT_POT_FAST`.
- **ESP32 build without flags**: only `sw_rcp` activates (auto-on via
  `__XTENSA__`). Should be a strict perf improvement; `recip0.s` + 2 N-R
  steps gives 1-ULP accuracy, equivalent to `1.0f/x`.
- **ESP32 build with `SW_USE_ESP_DSP`**: requires `espressif/esp-dsp` in
  the project's component manifest. Without it, the include of `dspm_mult.h`
  fails fast at compile time.
- **`SW_TEXTURE_REPEAT_POT_FAST`**: NPOT textures are unaffected (runtime
  branch). POT + negative UVs differ by one texel from `sw_fract` wrap.

## Testing

- **Static check (Windows x86_64, clang)**: `-fsyntax-only` clean for the full
  `rlsw.h` (and via `rlgl.h` with `GRAPHICS_API_OPENGL_SOFTWARE`), with and
  without `SW_TEXTURE_REPEAT_POT_FAST`.
- **End-to-end on x86_64 (Windows, mingw-w64 / gcc 14.2)**: built raylib with
  `PLATFORM=Memory` + `OPENGL_VERSION=Software` (which routes through rlsw),
  ran the `shapes_basic_shapes` example, captured the rlsw framebuffer via
  `TakeScreenshot`. The `sw_rcp` scalar fallback produces a correctly
  rasterized scene (shapes/text/lines all rendered as expected). Verified
  the build is clean both with and without `-DSW_TEXTURE_REPEAT_POT_FAST`.
  *(Note: the resulting PNG is upside-down due to an unrelated upstream
  bug in `rlReadScreenPixels` -- it unconditionally vertical-flips assuming
  real-`glReadPixels` bottom-left origin, but `swReadPixels` already returns
  top-down. Visible content is correct; can be fixed in a separate PR.)*
- **ESP-DSP path** (`SW_USE_ESP_DSP`): cannot be exercised on x86 since the
  kernels don't exist outside ESP-IDF. Sits behind a compile-time guard so
  builds without the flag are byte-identical to the scalar version. Validated
  end-to-end on real hardware (next bullet).
- **ESP32-S3**: end-to-end tested in a private fork of raylib-on-IDF with
  the full optimization stack (`__XTENSA__` `sw_rcp` + `SW_USE_ESP_DSP` +
  `SW_TEXTURE_REPEAT_POT_FAST`). Rendering is correct and the perf delta
  matches the table above.

## Credits

- Original `rlsw` design and review: @Bigfoot71
- ESP32 perf investigation and changes in this PR: @jensroth-git
- Kept narrow per @Bigfoot71's review feedback ("you could open a PR for the
  parts related to `sw_rcp` and `dspm_mult_4x4x4/1_f32`").

Created in Cursor

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -164,6 +164,19 @@
     #endif
 #endif
 
+// Fast power-of-two texture wrap (SW_REPEAT mode only)
+// When defined, textures whose width/height are powers of two use a bitmask
+// wrap (`x & (size-1)`) instead of `floorf`-based fractional wrap or the
+// signed `%` chain in the linear sampler. Saves a software divide on Xtensa
+// and a few instructions everywhere. NPOT textures keep using the original
+// path via a runtime `(size & (size-1)) == 0` check, so SW_REPEAT remains
+// correct for them. The only observable behavior change is for POT textures
+// sampled with negative UV coordinates: bitmask wrap (two's complement) can
+// differ from `sw_fract` by one texel. Off by default to keep bit-for-bit
+// behavior; opt in if you control your asset UVs.
+//
+// #define SW_TEXTURE_REPEAT_POT_FAST
+
 //----------------------------------------------------------------------------------
 // OpenGL Compatibility Types
 //----------------------------------------------------------------------------------
@@ -2456,11 +2469,31 @@ static inline void sw_texture_free(sw_texture_t *texture)
 
 static inline void sw_texture_sample_nearest(float *SW_RESTRICT color, const sw_texture_t *SW_RESTRICT tex, float u, float v)
 {
-    u = (tex->sWrap == SW_REPEAT)? sw_fract(u) : sw_saturate(u);
-    v = (tex->tWrap == SW_REPEAT)? sw_fract(v) : sw_saturate(v);
+    int x, y;
 
-    int x = u*tex->width;
-    int y = v*tex->height;
+#ifdef SW_TEXTURE_REPEAT_POT_FAST
+    if ((tex->sWrap == SW_REPEAT) && ((tex->width & tex->wMinus1) == 0))
+    {
+        x = (int)(u*tex->width) & tex->wMinus1;
+    }
+    else
+#endif
+    {
+        u = (tex->sWrap == SW_REPEAT)? sw_fract(u) : sw_saturate(u);
+        x = (int)(u*tex->width);
+    }
+
+#ifdef SW_TEXTURE_REPEAT_POT_FAST
+    if ((tex->tWrap == SW_REPEAT) && ((tex->height & tex->hMinus1) == 0))
+    {
+        y = (int)(v*tex->height) & tex->hMinus1;
+    }
+    else
+#endif
+    {
+        v = (tex->tWrap == SW_REPEAT)? sw_fract(v) : sw_saturate(v);
+        y = (int)(v*tex->height);
+    }
 
     tex->readColor(color, tex->pixels, y*tex->width + x);
 }
@@ -2482,13 +2515,19 @@ static inline void sw_texture_sample_linear(float *SW_RESTRICT color, const sw_t
     int x1 = x0 + 1;
     int y1 = y0 + 1;
 
-    // NOTE: If the textures are POT, avoid the division for SW_REPEAT
-
     if (tex->sWrap == SW_CLAMP)
     {
         x0 = (x0 > tex->wMinus1)? tex->wMinus1 : x0;
         x1 = (x1 > tex->wMinus1)? tex->wMinus1 : x1;
     }
+#ifdef SW_TEXTURE_REPEAT_POT_FAST
+    else if ((tex->width & tex->wMinus1) == 0)
+    {
+        // POT fast path: bitmask wrap covers negative ints via two's complement
+        x0 = x0 & tex->wMinus1;
+        x1 = x1 & tex->wMinus1;
+    }
+#endif
     else
     {
         x0 = (x0%tex->width + tex->width)%tex->width;
@@ -2500,6 +2539,13 @@ static inline void sw_texture_sample_linear(float *SW_RESTRICT color, const sw_t
         y0 = (y0 > tex->hMinus1)? tex->hMinus1 : y0;
         y1 = (y1 > tex->hMinus1)? tex->hMinus1 : y1;
     }
+#ifdef SW_TEXTURE_REPEAT_POT_FAST
+    else if ((tex->height & tex->hMinus1) == 0)
+    {
+        y0 = y0 & tex->hMinus1;
+        y1 = y1 & tex->hMinus1;
+    }
+#endif
     else
     {
         y0 = (y0%tex->height + tex->height)%tex->height;
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -844,6 +844,17 @@ SWAPI void swGetFramebufferAttachmentParameteriv(SWattachment attachment, SWatta
     #endif
 #endif
 
+// ESP-DSP acceleration: ESP-IDF ships an optimized math library that includes
+// `dspm_mult_4x4x4_f32` (4x4 matrix multiply) and `dspm_mult_4x4x1_f32`
+// (matrix * vector). These are S3-tuned hand-vectorized kernels that beat the
+// scalar versions for both throughput and code-size. Detection is opt-in to
+// keep the dependency optional: define SW_USE_ESP_DSP from your build system
+// (or rely on the `idf_component.yml` example shown in the rlsw docs).
+#if defined(ESP_PLATFORM) && defined(SW_USE_ESP_DSP)
+    #define SW_HAS_ESP_DSP
+    #include "dspm_mult.h"
+#endif
+
 #ifdef __cplusplus
     #define SW_CURLY_INIT(name) name
 #else
@@ -1038,6 +1049,9 @@ typedef struct {
     SWmatrix currentMatrixMode;                                 // Current matrix mode (e.g., sw_MODELVIEW, sw_PROJECTION)
     sw_matrix_t *currentMatrix;                                 // Pointer to the currently used matrix according to the mode
     sw_matrix_t matMVP;                                         // Model view projection matrix, calculated and used internally
+#ifdef SW_HAS_ESP_DSP
+    float matMVP_rm[16];                                        // Row-major MVP, kept in sync for esp-dsp dspm_mult_4x4x1_f32 vertex transform
+#endif
     bool isDirtyMVP;                                            // Indicates if the MVP matrix should be rebuilt
 
     sw_handle_t boundFramebufferId;                             // Framebuffer currently bound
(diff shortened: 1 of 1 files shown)


src/external/rlsw.h
```diff
@@ -2406,11 +2469,31 @@ static inline void sw_texture_free(sw_texture_t *texture)
 
 static inline void sw_texture_sample_nearest(float *SW_RESTRICT color, const sw_texture_t *SW_RESTRICT tex, float u, float v)
 {
-    u = (tex->sWrap == SW_REPEAT)? sw_fract(u) : sw_saturate(u);
-    v = (tex->tWrap == SW_REPEAT)? sw_fract(v) : sw_saturate(v);
+    int x, y;
+
+#ifdef SW_TEXTURE_REPEAT_POT_FAST
+    if ((tex->sWrap == SW_REPEAT) && ((tex->width & tex->wMinus1) == 0))
```
Bigfoot71, line 2475: I think it's fine that when `SW_TEXTURE_REPEAT_POT_FAST` is defined, we restrict `SW_REPEAT` to POT textures only, this would remove all extra branches here.

This would require updating `swTexParameteri` so `SW_REPEAT` is applied only if the texture is POT, otherwise leave it unchanged and set `RLSW.errCode = SW_INVALID_OPERATION`.

We could also store a `bool isPOT;` in `sw_texture_t` determined during `sw_texture_alloc`.

We should also explicitly set default parameters in `sw_texture_alloc`, like `SW_CLAMP` for wrapping.

src/external/rlsw.h
```diff
@@ -164,6 +164,19 @@
...
+// wrap (`x & (size-1)`) instead of `floorf`-based fractional wrap or the
+// signed `%` chain in the linear sampler. Saves a software divide on Xtensa
+// and a few instructions everywhere. NPOT textures keep using the original
+// path via a runtime `(size & (size-1)) == 0` check, so SW_REPEAT remains
+// correct for them. The only observable behavior change is for POT textures
+// sampled with negative UV coordinates: bitmask wrap (two's complement) can
+// differ from `sw_fract` by one texel. Off by default to keep bit-for-bit
+// behavior; opt in if you control your asset UVs.
+//
+// #define SW_TEXTURE_REPEAT_POT_FAST
```
Bigfoot71, line 178: In addition to the suggestion about repeated checks during sampling, I think we could generalize this with a `SW_SUPPORT_NPOT_TEXTURE` define, enabled by default.
Setting it to 0 would enable optimizations for non-POT textures.
The description could also be significantly simplified.

Bigfoot71:
> (Larger frame-time numbers in the original write-up include a separate
> async-double-buffer change in the platform layer that's not part of this PR.)

Even though ESP32 is not "natively" supported by raylib, I think it would be interesting to consider how to implement the necessary components for this in rlsw.

jensroth-git:
> > (Larger frame-time numbers in the original write-up include a separate
> > async-double-buffer change in the platform layer that's not part of this PR.)
> 
> Even though ESP32 is not "natively" supported by raylib, I think it would be interesting to consider how to implement the necessary components for this in rlsw.

This is how we solved it in the [esp32 repo](https://github.com/jensroth-git/raylib-esp32).

Double-buffer the color framebuffer and run the display transfer on a dedicated FreeRTOS task pinned to core 0, overlapping it with rendering on core 1.

**rlsw.h changes:**

`sw_default_framebuffer_t` now holds `sw_texture_t color[2] + int colorIndex`
`sw_default_framebuffer_alloc/free` manage both buffers
New public function `swSwapColorBuffer()` toggles the render target and returns a pointer to the just-rendered buffer

**Port layer changes:**
- `raylib_port_flush_async(src, w, h)` — stores buffer pointer, notifies
  flush task, returns immediately
- `raylib_port_wait_flush()` — blocks until the previous transfer completes
- Flush task: persistent FreeRTOS task on core 0 (4 KB stack), sleeps via
  `ulTaskNotifyTake`, runs vflip + byte-swap + DMA, signals completion

**SwapScreenBuffer flow:**
```
wait_flush()          // ensure previous transfer finished
rendered = swSwapColorBuffer()  // swap render target, get finished buffer
flush_async(rendered) // kick off transfer on core 0
```

**Memory cost:** ~115 KB PSRAM (second color buffer) + 4 KB task stack.

Since render time (~42 ms) > transfer time (~15 ms), the wait never blocks in
steady state. Effective swap cost drops to ~30 µs.

src/external/rlsw.h
raysan5: Agree.

src/external/rlsw.h
raysan5: I usually try to minimize to store information that can be calculated when needed but I understand that `isPOT` would be useful for every texture access. I forget that we are operating one level below OpenGL...

src/external/rlsw.h
```diff
@@ -3366,7 +3462,7 @@ static void sw_triangle_clip_and_project(void)
 
             // Calculation of the reciprocal of W for normalization
             // as well as perspective-correct attributes
-            const float wRcp = 1.0f/v->position[3];
+            const float wRcp = sw_rcp(v->position[3]);
```
raysan5, line 3465: This is a great improvement, would it be possible to minimize/avoid all similar required divisions in `rlsw` module?

raysan5:
> Even though ESP32 is not "natively" supported by raylib, I think it would be interesting to consider how to implement the necessary components for this in rlsw.

Agree. I think ESP32 is a great testbed for further performance improvements that can be extrapolated to other microcontrollers, I'd love to see raylib running on RPI Pico 2 in the future! 😄

src/external/rlsw.h
Bigfoot71: This is already handled, perspective correction divisions are reduced using affine interpolation per blocks (16 pixels by default). You can increase the block size if you want fewer divisions, but it comes at the cost of precision:
https://github.com/raysan5/raylib/blob/cf9f27db545dd686b03f9f773cfbdb8a8dd9f72f/src/external/rlsw.h#L5320-L5329

For projection, it's per-vertex anyway, so there's not really much more we can do without causing issues.

Bigfoot71: I'd add that for `1.0f/x` itself, on modern x86 and ARM manual implementations are rarely worth it in terms of performance/precision. If you really want that tradeoff, compiling with `-ffast-math` is usually the better option.

raysan5:
@jensroth-git @Bigfoot71 thanks for the improvement and the review! I'm merging the changes! Feel free to send other PRs for further revisions and improvements! 😄

PR #5828, opened 2026-05-01
Add extended versions of DrawGrid and DrawPlane functions

invin1x:
Add `DrawPlaneEx` and `DrawGridEx` functions to allow drawing planes and grids at custom angles.


raysan5:
@invin1x Thanks but I prefer to minimize additions and I think those users can be very easely implemented at user side for specific project needs.

PR #5830, opened 2026-05-01
SDL2: Use precise mouse wheel values

Menduist:
https://wiki.libsdl.org/SDL2/SDL_MouseWheelEvent

```c
    Sint32 x;           /**< The amount scrolled horizontally, positive to the right and negative to the left */
    Sint32 y;           /**< The amount scrolled vertically, positive away from the user and negative toward the user */
[...]
    float preciseX;     /**< The amount scrolled horizontally, positive to the right and negative to the left, with float precision (added in 2.0.18) */
    float preciseY;     /**< The amount scrolled vertically, positive away from the user and negative toward the user, with float precision (added in 2.0.18) */
```

SDL 2.0.18 was released in [2021](https://discourse.libsdl.org/t/sdl-2-0-18-released/33768)

### DIFF
--- a/src/platforms/rcore_desktop_sdl.c
+++ b/src/platforms/rcore_desktop_sdl.c
@@ -1675,8 +1675,13 @@ void PollInputEvents(void)
             } break;
             case SDL_MOUSEWHEEL:
             {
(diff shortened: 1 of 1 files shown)


raysan5:
@Menduist thanks for the update

PR #5831, opened 2026-05-02
Add "UpdateModelPose" to API for manual posing support

fyl2xp1:
I was trying to work on procedural animations and noticed there was no interface to update the matrices of a skinned mesh based on manually posed bone transforms (currentPose).
Ended up writing this simple patch with the intention of allowing procedural animations, custom ragdoll-driven skinning or any sort of posing that doesn't work by traditional animations, while trying to minimize the amount of code that had to be added to the project.
The added function allows for a clean way of manually posing the bones of a skinned mesh, without changing the existing animation API.

I have confirmed it works and builds correctly on Windows using MSVC.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1642,6 +1642,7 @@ RLAPI void SetModelMeshMaterial(Model *model, int meshId, int materialId);      
 
 // Model animations loading/unloading functions
 RLAPI ModelAnimation *LoadModelAnimations(const char *fileName, int *animCount);            // Load model animations from file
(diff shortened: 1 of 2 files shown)


raysan5:
@fyl2xp1 I prefer to avoid that addition, I think it could feel confusing for users of the basic API. I recommend implementing it on user side when specifically needed

casperbear:
@raysan5 it is confusing if the name is kept UpdateModelPose, especially with other functions in the group. However if it's called ApplyModelPose it will reflect what it does and be in line with 3D software terminology

PR #5832, opened 2026-05-02
[rmodels] Use inverse bind matrices when possible for bind pose in gltf loader

fyl2xp1:
Trying to fix an edge case I've encountered when loading gltf models that provide inverseBindMatrices. Right now raylib ignores them and derives bindPose from the skeleton node hierarchy, which differs in this case.
This patch derives bindPose from inverseBindMatrices when possible, and falls back to the existing path if they aren't provided by the model being loaded.

Tested on Windows using MSVC. Loads both a previously improperly handled model, and a previously working model.

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -6140,17 +6140,36 @@ static Model LoadGLTF(const char *fileName)
 
             for (int i = 0; i < model.skeleton.boneCount; i++)
             {
-                cgltf_node *node = skin.joints[i];
-                cgltf_float worldTransform[16];
-                cgltf_node_transform_world(node, worldTransform);
-                Matrix worldMatrix = {
-                    worldTransform[0], worldTransform[4], worldTransform[8], worldTransform[12],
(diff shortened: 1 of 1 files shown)


raysan5:
@fyl2xp1 Please, could you provide some model for testing those changes? Thanks!

raysan5:
@JeffM2501 What do you think about this change?

JeffM2501:
> @JeffM2501 What do you think about this change?

Seems like a good change to me.

fyl2xp1:
[ctm_sas.zip](https://github.com/user-attachments/files/27327160/ctm_sas.zip)
Here is the model that was broken

fyl2xp1:
Bump

kimkulling:
I will have a look.

raysan5:
@kimkulling thank you very much Kim!

kimkulling:
Looks fine for me.

raysan5:
@fyl2xp1 thanks for the improvement and thanks @JeffM2501 @kimkulling for the further review

PR #5833, opened 2026-05-02
FileTextReplace Return Value Proposal

hamuQonda:
This is a suggestion regarding the return value of FileTextReplace().
I've added the suggestion as a comment in src/raylib.h, so please check it.


raysan5:
@hamuQonda return vales for `File*()` should be unified, I prefer to avoid following different conventions.

hamuQonda:
> @hamuQonda return vales for `File*()` should be unified, I prefer to avoid following different conventions.

Yes, I understand. It was my first pull request, so it wasn't very meaningful. I hope to learn more about raylib and contribute someday. Thank you.

PR #5839, opened 2026-05-06
[rlsw][rcore_drm] Silence warnings when using PLATFORM_DRM and GRAPHICS_API_OPENGL_SOFTWARE

NighthowlerStudios:
The following commit silences three sets of warnings.

1. During SW_RASTER_TRIANGLE_SPAN, SW_RASTER_QUAD, and SW_RASTER_LINE, the inner-span for loop uses a goto to escape if z shows a tri/quad is to be discarded.  However, if SW_ENABLE_DEPTH_TEST is omitted (because of self inclusion not needing it in that cycle), the goto's label sits there unused.  

This commit just ifdef-guards those `discard` labels.

2. in SW_RASTER_TRIANGLE, both the x coordinates and the y coordinates of all vertices were cached.  But the x variables were never used.  

The definition lines were separated into separate ones for the x, and the y.  then the x ones are commented out.  If we ever need x in the future, they can be commented back in.

3. On the DRM platform C file, three specific functions are used to find the DRMModeConnector when OpenGL is hardware accelerated.  However, their callers located elsewhere are surrounded by #if !defined(GRAPHICS_API_OPENGL_SOFTWARE), leaving them unused if software rendering is enabled.

This commit adds more #if !defined checks to also leave out the definitions of those functions themselves, as they are not needed in the software renderer anyway.

Feel free to reject as you please.  I will be using this on my side for sanity sake regardless but figured you'd like to know.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -5564,9 +5564,13 @@ static void SW_RASTER_TRIANGLE(const sw_vertex_t *v0, const sw_vertex_t *v1, con
     if (v0->position[1] > v1->position[1]) { const sw_vertex_t *tmp = v0; v0 = v1; v1 = tmp; }
 
     // Extracting coordinates from the sorted vertices
-    float x0 = v0->position[0], y0 = v0->position[1];
-    float x1 = v1->position[0], y1 = v1->position[1];
-    float x2 = v2->position[0], y2 = v2->position[1];
+    // Put x away for safe keeping.  Only y is used right now.  Silences warnings.
+    //float x0 = v0->position[0];
+    float y0 = v0->position[1];
+    //float x1 = v1->position[0];
+    float y1 = v1->position[1];
+    //float x2 = v2->position[0];
+    float y2 = v2->position[1];
 
     // Compute height differences
     float h02 = y2 - y0;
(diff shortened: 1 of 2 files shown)


raysan5:
@Bigfoot71 please, could you take a look to this small review?

src/external/rlsw.h
```diff
@@ -5562,9 +5564,13 @@ static void SW_RASTER_TRIANGLE(const sw_vertex_t *v0, const sw_vertex_t *v1, con
...
-    float x0 = v0->position[0], y0 = v0->position[1];
-    float x1 = v1->position[0], y1 = v1->position[1];
-    float x2 = v2->position[0], y2 = v2->position[1];
+    // Put x away for safe keeping.  Only y is used right now.  Silences warnings.
+    //float x0 = v0->position[0];
+    float y0 = v0->position[1];
+    //float x1 = v1->position[0];
+    float y1 = v1->position[1];
+    //float x2 = v2->position[0];
+    float y2 = v2->position[1];
```
Bigfoot71, line 5573: It's okay, you can remove the comment and the commented lines.

Bigfoot71 (changes requested):

Bigfoot71:
@raysan5 Looks good to me, the commented lines can also be removed

raysan5:
@Bigfoot71 thanks for the review! I'm removing those commented lines!

@NighthowlerStudios thanks for the improvements

PR #5840, opened 2026-05-07
Check fread return value in jar_mod_load_file

fffaraz:
Silences `-Wunused-result` warning by handling fread() failure in jar_mod.h, returning 0 when the read does not complete.


raysan5:
@fffaraz thanks for the review

PR #5844, opened 2026-05-07
[cmake] Fix config.h parsing into cmake options

Peter0x44:
The restructuring of config.h broke the fragile regex that cmake was using to parse config.h for its own configurable options. It assumed the presence of an uncommented line was enabled, and ignored parsing 0/1.

This caused major problems, like:
-DSUPPORT_BUSY_WAIT_LOOP=ON
-DSUPPORT_CUSTOM_FRAME_CONTROL=ON

When configuring using cmake by default, with no options explicitly passed.

@raysan5 I would honestly do another release with this fix. It's really bad...


raysan5:
@Peter0x44 thanks for the fix!

raylib is going at the moment through a private security audit, once critical points get fixed I'll do another release.

Peter0x44:
Do you know when this audit will be finished?

raysan5:
@Peter0x44 I don't know, it started April 23rd, there are about 20 working items open at the moment, 5 potential vulnerabilities found but only 1 critical (`OpenURL()`). When they finish the review and generate the report I should go through the vulnerabilities to see what can be addressed. I'm having a call with them next week for more details.

PR #5847, opened 2026-05-08
[CMake] Add configure-time test for libatomic requirement

### LINKED ISSUE #5836: [build] CMake has Atomic library missing on 32-bit ARM
### Issue description
Previously it was discussed that 32-bit ARM needs the -latomic flag passed manually where as 64-bit does not.

However, later after that fix, which was patched in whilst PLATFORM_RPI was still a thing...
https://github.com/raysan5/raylib/commit/1f806b555dd1ecb7e77797ed0de57d62a6bdd6f0

PLATFORM_RPI was removed in 6.0, so then we point to PLATFORM_DESKTOP as the wiki for Raspberry Pi setup states https://github.com/raysan5/raylib/wiki/Working-on-Raspberry-Pi

This sadly leads me nowhere when looking at https://github.com/raysan5/raylib/issues/2452 

However, this is ONLY happening on CMake.  Using make with 
```
cd src
make PLATFORM=PLATFORM_DESKTOP GRAPHICS=GRAPHICS_API_OPENGL_ES2
```
works fine.

All I know is that raudio, and raudio alone, it preventing me from building absolutely anything on a Raspberry Pi 2 Model B but only if I try using CMake, with the exact same error set that issue 2452 was getting.

### Environment

Raspberry Pi 2 Model B, running Raspbian Trixie (Debian 13)
CMake 4.3.2

### Issue Screenshot

See issue 2452 for the list of errors; it is exactly the same.

### Code Example

Simply doing in the command line
```
mkdir build
cd build 
cmake .. -DPLATFORM=Desktop
make
```

Peter0x44:
CMake now checks if -latomic is required for 64-bit atomics, and links it if it's required. Miniaudio is the only thing in raylib that needs it, so it's put behind SUPPORT_MODULE_RAUDIO.

Fixes #5836

### DIFF
--- a/cmake/LibraryConfigurations.cmake
+++ b/cmake/LibraryConfigurations.cmake
@@ -7,6 +7,30 @@ if(POLICY CMP0072)
   cmake_policy(SET CMP0072 NEW)
 endif()
 
+include(CheckCSourceCompiles)
+include(CMakePushCheckState)
+
+function(raylib_check_libatomic_required result)
+    set(_atomic_test_source "
+int main(void)
+{
+    volatile long long value = 0;
+    return (int)__atomic_fetch_add(&value, 1, __ATOMIC_SEQ_CST);
+}")
+
+    check_c_source_compiles("${_atomic_test_source}" RAYLIB_ATOMICS_WITHOUT_LIBATOMIC)
+
+    if (RAYLIB_ATOMICS_WITHOUT_LIBATOMIC)
+        set(${result} FALSE PARENT_SCOPE)
+    else ()
+        cmake_push_check_state()
+        list(APPEND CMAKE_REQUIRED_LIBRARIES atomic)
+        check_c_source_compiles("${_atomic_test_source}" RAYLIB_ATOMICS_WITH_LIBATOMIC)
+        cmake_pop_check_state()
+        set(${result} ${RAYLIB_ATOMICS_WITH_LIBATOMIC} PARENT_SCOPE)
+    endif ()
+endfunction()
+
 set(RAYLIB_DEPENDENCIES "include(CMakeFindDependencyMacro)")
 
 if (${PLATFORM} STREQUAL "Desktop")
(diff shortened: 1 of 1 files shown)


orcmid:
Side Question: Are "atomics" and "intrinsics" the same thing?  Involving compilers making direct use of now x64 chip-provided functions in-line?  Things like sin(x), various matrix functions, yada yada?

Or by "atomics" is meant some (intrinsic) operations that are not interruptible, such as compare-and-swap, others including increment/decrement operations?

Peter0x44:
> Or by "atomics" is meant some (intrinsic) operations that are not interruptible, such as compare-and-swap, others including increment/decrement operations?

It means the latter. Some chips (e.g. in this case 32-bit arm) don't have instructions to do some atomic operations for 64-bit integers, so routines that do this have to be provided separately in a library, distributed as part of the toolchain. That's why raspberry pi 2 needs to link libatomic, but normal x64 linux desktops don't.

orcmid:
@Peter0x44 > > Or by "atomics" is meant some (intrinsic) operations that are not interruptible, such as compare-and-swap, others including increment/decrement operations?
> 
> It means the latter. Some chips (e.g. in this case 32-bit arm) don't have instructions to do some atomic operations for 64-bit integers, so routines that do this have to be provided separately in a library, distributed as part of the toolchain.

So those library routines have some way to prevent any process from interfering with the operations that don't have hardware intrinsics but must not be interrupted nor interfered with at the memory/cache level?

Peter0x44:
> So those library routines have some way to prevent any process from interfering with the operations that don't have hardware intrinsics but must not be interrupted nor interfered with at the memory/cache level?

Yes, that's my understanding. I don't know the implementation details, but you can read the code here if you would like to find out:

https://github.com/gnutools/gcc/tree/master/libatomic

Peter0x44:
You can count them as "intrinsic" because the compiler itself can emit calls to them (if necessary). I think.

raysan5:
@Peter0x44 thanks for the improvement

PR #5849, opened 2026-05-08
[rlsw][SEGFAULT] Fix triangle and quad spans applying pixels out of bounds

### LINKED ISSUE #5838: [rlsw] Segfaulting some example and resolution combos on RPI with PLATFORM_DRM and OPENGL_SOFTWARE

NighthowlerStudios:
Fixes https://github.com/raysan5/raylib/issues/5838

The code in this commit and text in this PR are entirely human written.  

It has been tested on a Raspberry Pi 2, the same board the original issue was found on.  

Then it was tested on a Raspberry Pi 5 to verify texture and triangle warping wasn't present, using `models_first_person_maze`

# Fixes RLSW Rastering buffer boundary out of bounds

In the raster routine `SW_RASTER_QUAD`, there was the following problem spotted by @ZX41R

> In the quad path, `SW_RASTER_QUAD` derives `xMin/yMin/xMax/yMax` directly from the projected corners and then computes:
> 
> baseOffset = y * stride + xMin;
> cPtr = cPixels + baseOffset * SW_FRAMEBUFFER_COLOR_SIZE;
> without the clamp that the clear paths use. A negative `xMin`/`yMin`, or a projected quad extending outside the target, can therefore move `cPtr` before the color buffer before the inner loop even starts. On 32-bit ARM that underflow can plausibly land on a mapped ELF page, which matches `dst=... "\177ELF"` in the trace.

Additionally, in `TRIANGLE_SPAN`, the issue is much simpler.  It only has an `xMin` that calculates into `baseOffset`, there is no `yMin` involved on that method.  However it is the same boundary crossing bug.

This commit applies some fixes they suggested by ensuring to clamp the loop min and max to inside of `RLSW.colorBuffer`.

Then it moves the start of each interpolation for spans, textures, depths and vertex colours, to compensate for the new clamps so that only the correct colours are grabbed.

This fixes the segfault discovered on the issue and avoids the triangle warping that was previously discussed in the deleted fork on  https://github.com/raysan5/raylib/pull/5843

It should be much safer than the previous PR I self-rejected.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -5389,6 +5389,11 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t 
     int xEnd = (int)end->position[0];
     if (xStart == xEnd) return;
 
+    // Get the current row and skip if outside the framebuffer.
+    // Maybe this check is better suited elsewhere? 
+    int y = (int)start->position[1];
+    if (y < 0 || y >= RLSW.colorBuffer->height) return;
+
     // Compute the inverse horizontal distance along the X axis
     float dxRcp = sw_rcp(end->position[0] - start->position[0]);
 
@@ -5411,25 +5416,30 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t 
     // Compute the subpixel distance to traverse before the first pixel
     float xSubstep = 1.0f - sw_fract(start->position[0]);
 
-    // Initializing the interpolation starting values
-    float w = start->position[3] + dWdx*xSubstep;
+    // Intercept the span bounds to ensure we don't write before the framebuffer.
+    int xLoopStart = sw_clamp_int(xStart, 0, RLSW.colorBuffer->width - 1);
+    int dxStart = xLoopStart - xStart;
+
+    // Initializing the interpolation starting values.
+    // Also step further into them to move away from the colorbuffer edge.
+    float w = start->position[3] + dWdx*xSubstep + dWdx*dxStart;
     float color[4] = {
-        start->color[0] + dCdx[0]*xSubstep,
-        start->color[1] + dCdx[1]*xSubstep,
-        start->color[2] + dCdx[2]*xSubstep,
-        start->color[3] + dCdx[3]*xSubstep
+        start->color[0] + dCdx[0]*xSubstep + dCdx[0]*dxStart,
+        start->color[1] + dCdx[1]*xSubstep + dCdx[1]*dxStart,
+        start->color[2] + dCdx[2]*xSubstep + dCdx[2]*dxStart,
+        start->color[3] + dCdx[3]*xSubstep + dCdx[3]*dxStart
     };
 #ifdef SW_ENABLE_DEPTH_TEST
-    float z = start->position[2] + dZdx*xSubstep;
+    float z = start->position[2] + dZdx*xSubstep + dZdx*dxStart;
 #endif
 #ifdef SW_ENABLE_TEXTURE
-    float u = start->texcoord[0] + dUdx*xSubstep;
-    float v = start->texcoord[1] + dVdx*xSubstep;
+    float u = start->texcoord[0] + dUdx*xSubstep + dUdx*dxStart;
+    float v = start->texcoord[1] + dVdx*xSubstep + dVdx*dxStart;
 #endif
 
     // Pre-calculate the starting pointers for the framebuffer row
-    int y = (int)start->position[1];
-    int baseOffset = y*RLSW.colorBuffer->width + xStart;
+    // Don't allow a y value outside the buffer.
+    int baseOffset = y*RLSW.colorBuffer->width + xLoopStart;
     uint8_t *cPtr = (uint8_t *)(RLSW.colorBuffer->pixels) + baseOffset*SW_FRAMEBUFFER_COLOR_SIZE;
 #ifdef SW_ENABLE_DEPTH_TEST
     uint8_t *dPtr = (uint8_t *)(RLSW.depthBuffer->pixels) + baseOffset*SW_FRAMEBUFFER_DEPTH_SIZE;
@@ -5437,12 +5447,14 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t 
 
 #define SW_AFFINE_BLOCK 16
 
-    int x = xStart;
-    while (x < xEnd)
+    int x = xLoopStart;
+    // Prevent pixels from beyond the buffer from processing.
+    int xLoopEnd = sw_clamp_int(xEnd, 0, RLSW.colorBuffer->width - 1);
+    while (x < xLoopEnd)
     {
         // Clamp last block to remaining pixels
         int blockEnd = x + SW_AFFINE_BLOCK;
-        if (blockEnd > xEnd) blockEnd = xEnd;
+        if (blockEnd > xLoopEnd) blockEnd = xLoopEnd;
         float blockLenF = (float)(blockEnd - x);
         float blockLenRcp = sw_rcp(blockLenF);
 
@@ -5730,21 +5742,38 @@ static void SW_RASTER_QUAD(const sw_vertex_t *a, const sw_vertex_t *b,
     uint8_t *dPixels = RLSW.depthBuffer->pixels;
 #endif
 
-    for (int y = yMin; y < yMax; y++)
+    // Intercept the boundaries to stay within the framebuffer.
+    int yLoopMin = sw_clamp_int(yMin, 0, RLSW.colorBuffer->height - 1);
+    int yLoopMax = sw_clamp_int(yMax, 0, RLSW.colorBuffer->height - 1);
+    int xLoopMin = sw_clamp_int(xMin, 0, RLSW.colorBuffer->width - 1);
+    int xLoopMax = sw_clamp_int(xMax, 0, RLSW.colorBuffer->width - 1);
+    int dxMin = xLoopMin - xMin;
+    #if defined(SW_ENABLE_DEPTH_TEST) || defined(SW_ENABLE_TEXTURE)
+    int dyMin = yLoopMin - yMin;
+    #endif
+
+    for (int y = yLoopMin; y < yLoopMax; y++)
     {
-        int baseOffset = y*stride + xMin;
+        int baseOffset = y*stride + xLoopMin;
         uint8_t *cPtr = cPixels + baseOffset*SW_FRAMEBUFFER_COLOR_SIZE;
     #ifdef SW_ENABLE_DEPTH_TEST
         uint8_t *dPtr = dPixels + baseOffset*SW_FRAMEBUFFER_DEPTH_SIZE;
-        float z = zRow;
+        // Correct our start by how far we clipped outside the framebuffer.
+        float z = zRow + dZdy*dyMin;
     #endif
     #ifdef SW_ENABLE_TEXTURE
-        float u = uRow;
-        float v = vRow;
+        // Correct our start by how far we clipped outside the framebuffer.
+        float u = uRow + dUdy*dyMin;
+        float v = vRow + dVdy*dyMin;
     #endif
-        float color[4] = { cRow[0], cRow[1], cRow[2], cRow[3] };
+        float color[4] = {
+            cRow[0] + dCdx[0]*dxMin,
+            cRow[1] + dCdx[1]*dxMin,
+            cRow[2] + dCdx[2]*dxMin,
+            cRow[3] + dCdx[3]*dxMin
+        };
 
-        for (int x = xMin; x < xMax; x++)
+        for (int x = xLoopMin; x < xLoopMax; x++)
         {
             float srcColor[4] = { color[0], color[1], color[2], color[3] };
 
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -5752,25 +5752,36 @@ static void SW_RASTER_QUAD(const sw_vertex_t *a, const sw_vertex_t *b,
     int dyMin = yLoopMin - yMin;
     #endif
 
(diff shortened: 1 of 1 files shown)


ZX41R:
Thanks for retesting on the RPi 2 + 5. Clipping interpolants by `dxMin`/`dyMin` is the right approach — but two issues in the bound math, both off-by-one on the *upper* clamp:

```c
int yLoopMax = sw_clamp_int(yMax, 0, RLSW.colorBuffer->height - 1);
int xLoopMax = sw_clamp_int(xMax, 0, RLSW.colorBuffer->width  - 1);
// ...
for (int y = yLoopMin; y < yLoopMax; y++) { ... }
for (int x = xLoopMin; x < xLoopMax; x++) { ... }
```

`yMax`/`xMax` are *exclusive* bounds in these loops (the original code is `y < yMax`). Clamping the exclusive upper to `height - 1` means a fully-on-screen quad ending at the bottom row never writes that row. Same in `TRIANGLE_SPAN` for `xLoopEnd`. Correct shape:

```c
int yLoopMax = sw_clamp_int(yMax, 0, RLSW.colorBuffer->height); // not height-1
int xLoopMax = sw_clamp_int(xMax, 0, RLSW.colorBuffer->width);  // not width-1
```

Inclusive lower (`xLoopMin`/`yLoopMin`) is correctly clamped to `width - 1`/`height - 1`, since those *do* index a pixel.

Second concern — the per-row interpolant compensation in `SW_RASTER_QUAD`:

```c
for (int y = yLoopMin; y < yLoopMax; y++) {
    ...
    float z = zRow + dZdy*dyMin;
    float u = uRow + dUdy*dyMin;
    float v = vRow + dVdy*dyMin;
```

If `zRow`/`uRow`/`vRow` are advanced once per row in the surrounding code (the typical pattern), then `dZdy*dyMin` is a constant offset that should be applied *once* before the y loop, not on every iteration. Otherwise every row reads at `(y, dyMin)` rather than `(y, 0)`. If on the other hand `zRow` is held constant and the outer caller doesn't advance it, then this is correct but `z = zRow + dZdy*(y - yMin)` would be the more honest expression. Worth a quick read of the surrounding code to pick which one applies — I'd bet on the former from the variable naming (`zRow` suggests "row's starting z").

Once those two are sorted, the structure is good. RPi 2 retest after fixing the upper clamp will tell you immediately whether the bottom row was missing.

raysan5:
@Bigfoot71 please, could you take a look to this PR when possible? No hurries.

NighthowlerStudios:
> `yMax`/`xMax` are _exclusive_ bounds in these loops (the original code is `y < yMax`). Clamping the exclusive upper to `height - 1` means a fully-on-screen quad ending at the bottom row never writes that row. Same in `TRIANGLE_SPAN` for `xLoopEnd`. Correct shape:

Great catch!  My bad.  That makes sense because of the < check.  

After some research I can agree that a -1 on LoopStart is still safe because these raster routines are reached after culling of triangles entirely outside anyway.  Our issue is only with triangles overlapping the border.

> Worth a quick read of the surrounding code to pick which one applies — I'd bet on the former from the variable naming (`zRow` suggests "row's starting z").

Yes, and I believe that I've discovered that those row vars don't just denote the start of the row, but actually behave as 2-dimensional "cursors" (or turtles) for the interpolation.  At the end of the for loop, they get incremented by `d(*)dy`, which means their positions go down a new row.

So we need *bilinear interpolation offsets*.  Instead of *just* moving them across the X, we need to move them across *both* X and Y for the correct start position.  And yes, only do it once, outside of the loop.  The incrementor for X gets put into a separate variable and the for loop controls how far that will sweep to the right and bottom anyway, and the incrementor at the end of the for loop adding a Y step (just adding row) would keep the x offset, so it's safe to do this only once.

This has to be in, because the for loop is just iterating, so yStart wouldn't have affected where interpolation started.  `cRow`, `zRow`, `uRow` and `vRow` still had their values set at the top left corner of the quad.  My changes moved that top left to inside the colorBuffer.

Tested our changes on both RPI2 and RPI5.

NighthowlerStudios:
Yes, by fixing the off by one, a single pixel wide bar of pixels on the right hand and bottom end of the screen is now being filled by colour information correctly.  The first commit in the PR had them white only, which was definitely something I should have spotted.

Bunnymark shows no issues with colour information after the movement of interpolation starts on raster_quad.  Maybe there's a better example I could look toward that could reliably produce artifacts if our code is wrong here?

src/external/rlsw.h
```diff
@@ -5411,38 +5416,45 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t
     // Compute the subpixel distance to traverse before the first pixel
     float xSubstep = 1.0f - sw_fract(start->position[0]);
 
-    // Initializing the interpolation starting values
-    float w = start->position[3] + dWdx*xSubstep;
+    // Intercept the span bounds to ensure we don't write before the framebuffer.
+    int xLoopStart = sw_clamp_int(xStart, 0, RLSW.colorBuffer->width - 1);
```
Bigfoot71: We can replace the clamps (including `xLoopEnd` below) with this:
```c
int xLoopStart = (xStart >= 0)? xStart : 0;
int xLoopEnd = (xEnd <= RLSW.colorBuffer->width)? xEnd : RLSW.colorBuffer->width;
if (xLoopStart >= xLoopEnd) return;
```

This keeps `xLoopEnd` as an exclusive bound and avoids drawing a ghost pixel when the span is fully outside the framebuffer.

We can then do the span clipping at the beginning of the function, which gives us this:
```c
static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t *end, float dUdy, float dVdy)
{
    // Get the current row and skip if outside the framebuffer.
    // Maybe this check is better suited elsewhere? 
    int y = (int)start->position[1];
    if (y < 0 || y >= RLSW.colorBuffer->height) return;

    // Gets the start/end coordinates and skip empty lines
    int xStart = (int)start->position[0];
    int xEnd = (int)end->position[0];
    if (xStart == xEnd) return;

    // Intercept the span bounds to ensure we don't write before the framebuffer.
    int xLoopStart = (xStart >= 0)? xStart : 0;
    int xLoopEnd = (xEnd <= RLSW.colorBuffer->width)? xEnd : RLSW.colorBuffer->width;
    if (xLoopStart >= xLoopEnd) return;

    // ...
```

src/external/rlsw.h
```diff
@@ -5411,38 +5416,45 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t
...
-    float u = start->texcoord[0] + dUdx*xSubstep;
-    float v = start->texcoord[1] + dVdx*xSubstep;
+    float u = start->texcoord[0] + dUdx*xSubstep + dUdx*dxStart;
+    float v = start->texcoord[1] + dVdx*xSubstep + dVdx*dxStart;
 #endif
 
     // Pre-calculate the starting pointers for the framebuffer row
-    int y = (int)start->position[1];
-    int baseOffset = y*RLSW.colorBuffer->width + xStart;
+    // Don't allow a y value outside the buffer.
```
Bigfoot71: Perhaps a ghost comment to remove here

src/external/rlsw.h
```diff
@@ -5730,21 +5742,48 @@ static void SW_RASTER_QUAD(const sw_vertex_t *a, const sw_vertex_t *b,
...
 #endif
 
-    for (int y = yMin; y < yMax; y++)
+    // Intercept the boundaries to stay within the framebuffer.
+    int yLoopMin = sw_clamp_int(yMin, 0, RLSW.colorBuffer->height - 1);
+    int yLoopMax = sw_clamp_int(yMax, 0, RLSW.colorBuffer->height);
+    int xLoopMin = sw_clamp_int(xMin, 0, RLSW.colorBuffer->width - 1);
+    int xLoopMax = sw_clamp_int(xMax, 0, RLSW.colorBuffer->width);
+    int dxMin = xLoopMin - xMin;
+    int dyMin = yLoopMin - yMin;
```
Bigfoot71: Here too, an explicit intersection might be preferable to clamps, as proposed above:
```c
int yLoopMin = (yMin >= 0)? yMin : 0;
int xLoopMin = (xMin >= 0)? xMin : 0;
int yLoopMax = (yMax <= RLSW.colorBuffer->height)? yMax : RLSW.colorBuffer->height;
int xLoopMax = (xMax <= RLSW.colorBuffer->width)?  xMax : RLSW.colorBuffer->width;

if (yLoopMin >= yLoopMax || xLoopMin >= xLoopMax) return;
```

I also suggest explicitly casting `dxMin` / `dyMin` to float:
```c
float dxMin = (float)(xLoopMin - xMin);
float dyMin = (float)(yLoopMin - yMin);
```

src/external/rlsw.h
```diff
@@ -5411,38 +5416,45 @@ static void SW_RASTER_TRIANGLE_SPAN(const sw_vertex_t *start, const sw_vertex_t
...
 #ifdef SW_ENABLE_DEPTH_TEST
-    float z = start->position[2] + dZdx*xSubstep;
+    float z = start->position[2] + dZdx*xSubstep + dZdx*dxStart;
 #endif
 #ifdef SW_ENABLE_TEXTURE
-    float u = start->texcoord[0] + dUdx*xSubstep;
-    float v = start->texcoord[1] + dVdx*xSubstep;
+    float u = start->texcoord[0] + dUdx*xSubstep + dUdx*dxStart;
+    float v = start->texcoord[1] + dVdx*xSubstep + dVdx*dxStart;
 #endif
```
Bigfoot71: We can also factor the initial interpolation offset:
```c
    // Compute the subpixel distance to traverse before the first pixel
    // Also step further into them to move away from the colorbuffer edge.
    float xSubstep = 1.0f - sw_fract(start->position[0]);
    float dxStart = (float)(xLoopStart - xStart);
    float xOffset = xSubstep + dxStart;

    // Initializing the interpolation starting values.
    float w = start->position[3] + dWdx*xOffset;
    float color[4] = {
        start->color[0] + dCdx[0]*xOffset,
        start->color[1] + dCdx[1]*xOffset,
        start->color[2] + dCdx[2]*xOffset,
        start->color[3] + dCdx[3]*xOffset
    };
#ifdef SW_ENABLE_DEPTH_TEST
    float z = start->position[2] + dZdx*xOffset;
#endif
#ifdef SW_ENABLE_TEXTURE
    float u = start->texcoord[0] + dUdx*xOffset;
    float v = start->texcoord[1] + dVdx*xOffset;
#endif
```

Bigfoot71 (approved):
Approved from my side. I tested the current fix and the suggested changes locally, the logic is preserved, with earlier returns for fully clipped spans/quads. Please still retest on the RPi hardware to confirm there is no platform specific regression.

src/external/rlsw.h
NighthowlerStudios: Definitely.  Whoops.

src/external/rlsw.h
NighthowlerStudios: applied with https://github.com/raysan5/raylib/commit/4a207151e2e5bcd263926d2a723e3a091cf501ac
and followed up with a forgotten line removal (compilation error on my branch) with https://github.com/raysan5/raylib/commit/e192cece3970c244077765379c40b7eb490e713a

src/external/rlsw.h
NighthowlerStudios: resolved with https://github.com/raysan5/raylib/commit/5c9e1ff312e0ad7f8681f478a5480c8238099627

src/external/rlsw.h
NighthowlerStudios: applied with https://github.com/raysan5/raylib/commit/2cd1746465af060bf2396a25ce507261bc64ea25

src/external/rlsw.h
NighthowlerStudios: applied with https://github.com/raysan5/raylib/commit/6768b73adc8604345422a04beb24e3c322f7fa2a

NighthowlerStudios:
Tested @Bigfoot71 's suggestions on my RPi2.  Solid results, no issues anymore on my end.

Merged remote main into my branch to test for conflicts, none found, so all good there.

I think we're ready for this to merge.

raysan5:
@NighthowlerStudios @Bigfoot71 thanks for the improvement and further review and testing

PR #5852, opened 2026-05-09
[Platform/WASIGL] new platform WIP

konsumer:
This is related to [this discussion](https://github.com/raysan5/raylib/discussions/3626).

It adds a new platform that just imports the functions it needs (GLFW and GL) from the wasm-host.

A user should be able to use it on the web with [easywasm/gl](https://github.com/easywasm/gl/), but also can expose the same functions on native wasm hosts, just using GLFW directly.

*why is this better than emscripten?* All functions that are needed are imported from host, so instead of bloating it up with JS and extra code to bind the JS to wasm, it makes it way lighter. The imported functions are mandatory, but can be implemented however the host wants to.

The change itself is fairly trivial. it's essentially just glfw, then `-Wl,--allow-undefined` is added to import anything that is not linked. Additionally, it should be built with wasi-sdk, which presets some flags, and has many stdlib headers implemented to use WASI.

I'm not totally sure if I modified enough files. I messed with build.zig and Makefile. Are there more?

Also, not really sure what the `ubsan_handle` functions are, but it might be worth tracking down why they are not defined (and being imported.)

for illustration core_basic_window.wasm is all wasi and glfw/gl:

```
Import[82]:
 - func[0] sig=0 <__ubsan_handle_nonnull_arg> <- env.__ubsan_handle_nonnull_arg
 - func[1] sig=1 <__ubsan_handle_out_of_bounds> <- env.__ubsan_handle_out_of_bounds
 - func[2] sig=2 <__ubsan_handle_pointer_overflow> <- env.__ubsan_handle_pointer_overflow
 - func[3] sig=3 <fd_close|wasi_snapshot_preview1> <- wasi_snapshot_preview1.fd_close
 - func[4] sig=4 <__imported_wasi_snapshot_preview1_fd_write> <- wasi_snapshot_preview1.fd_write
 - func[5] sig=5 <__imported_wasi_snapshot_preview1_fd_fdstat_get> <- wasi_snapshot_preview1.fd_fdstat_get
 - func[6] sig=6 <__imported_wasi_snapshot_preview1_fd_seek> <- wasi_snapshot_preview1.fd_seek
 - func[7] sig=0 <__imported_wasi_snapshot_preview1_proc_exit> <- wasi_snapshot_preview1.proc_exit
 - func[8] sig=0 <__ubsan_handle_builtin_unreachable> <- env.__ubsan_handle_builtin_unreachable
 - func[9] sig=1 <__ubsan_handle_type_mismatch_v1> <- env.__ubsan_handle_type_mismatch_v1
 - func[10] sig=3 <glfwSetErrorCallback> <- env.glfwSetErrorCallback
 - func[11] sig=0 <glfwInitAllocator> <- env.glfwInitAllocator
 - func[12] sig=7 <glfwInit> <- env.glfwInit
 - func[13] sig=8 <glfwDefaultWindowHints> <- env.glfwDefaultWindowHints
 - func[14] sig=1 <glfwWindowHint> <- env.glfwWindowHint
 - func[15] sig=3 <glfwSetJoystickCallback> <- env.glfwSetJoystickCallback
 - func[16] sig=7 <glfwGetPrimaryMonitor> <- env.glfwGetPrimaryMonitor
 - func[17] sig=3 <glfwGetVideoMode> <- env.glfwGetVideoMode
 - func[18] sig=2 <__ubsan_handle_divrem_overflow> <- env.__ubsan_handle_divrem_overflow
 - func[19] sig=9 <glfwCreateWindow> <- env.glfwCreateWindow
 - func[20] sig=8 <glfwTerminate> <- env.glfwTerminate
 - func[21] sig=3 <glfwGetMonitors> <- env.glfwGetMonitors
 - func[22] sig=3 <glfwGetWindowMonitor> <- env.glfwGetWindowMonitor
 - func[23] sig=2 <__ubsan_handle_add_overflow> <- env.__ubsan_handle_add_overflow
 - func[24] sig=2 <glfwGetWindowPos> <- env.glfwGetWindowPos
 - func[25] sig=2 <glfwGetMonitorPos> <- env.glfwGetMonitorPos
 - func[26] sig=2 <__ubsan_handle_sub_overflow> <- env.__ubsan_handle_sub_overflow
 - func[27] sig=2 <__ubsan_handle_mul_overflow> <- env.__ubsan_handle_mul_overflow
 - func[28] sig=2 <glfwSetWindowSize> <- env.glfwSetWindowSize
 - func[29] sig=0 <glfwMakeContextCurrent> <- env.glfwMakeContextCurrent
 - func[30] sig=3 <glfwGetError> <- env.glfwGetError
 - func[31] sig=1 <__ubsan_handle_load_invalid_value> <- env.__ubsan_handle_load_invalid_value
 - func[32] sig=0 <glfwSwapInterval> <- env.glfwSwapInterval
 - func[33] sig=2 <glfwGetWindowContentScale> <- env.glfwGetWindowContentScale
 - func[34] sig=1 <__ubsan_handle_float_cast_overflow> <- env.__ubsan_handle_float_cast_overflow
 - func[35] sig=7 <glfwGetPlatform> <- env.glfwGetPlatform
 - func[36] sig=2 <glfwGetFramebufferSize> <- env.glfwGetFramebufferSize
 - func[37] sig=10 <glfwGetMonitorWorkarea> <- env.glfwGetMonitorWorkarea
 - func[38] sig=2 <glfwSetWindowPos> <- env.glfwSetWindowPos
 - func[39] sig=0 <glfwIconifyWindow> <- env.glfwIconifyWindow
 - func[40] sig=11 <glfwSetWindowMonitor> <- env.glfwSetWindowMonitor
 - func[41] sig=2 <glfwSetWindowAttrib> <- env.glfwSetWindowAttrib
 - func[42] sig=0 <glfwFocusWindow> <- env.glfwFocusWindow
 - func[43] sig=0 <glfwHideWindow> <- env.glfwHideWindow
 - func[44] sig=5 <glfwGetWindowAttrib> <- env.glfwGetWindowAttrib
 - func[45] sig=0 <glfwMaximizeWindow> <- env.glfwMaximizeWindow
 - func[46] sig=3 <glfwGetProcAddress> <- env.glfwGetProcAddress
 - func[47] sig=5 <glfwSetWindowSizeCallback> <- env.glfwSetWindowSizeCallback
 - func[48] sig=5 <glfwSetFramebufferSizeCallback> <- env.glfwSetFramebufferSizeCallback
 - func[49] sig=5 <glfwSetWindowPosCallback> <- env.glfwSetWindowPosCallback
 - func[50] sig=5 <glfwSetWindowMaximizeCallback> <- env.glfwSetWindowMaximizeCallback
 - func[51] sig=5 <glfwSetWindowIconifyCallback> <- env.glfwSetWindowIconifyCallback
 - func[52] sig=5 <glfwSetWindowFocusCallback> <- env.glfwSetWindowFocusCallback
 - func[53] sig=5 <glfwSetDropCallback> <- env.glfwSetDropCallback
 - func[54] sig=5 <glfwSetWindowContentScaleCallback> <- env.glfwSetWindowContentScaleCallback
 - func[55] sig=1 <glfwSetWindowShouldClose> <- env.glfwSetWindowShouldClose
 - func[56] sig=5 <glfwSetKeyCallback> <- env.glfwSetKeyCallback
 - func[57] sig=5 <glfwSetCharCallback> <- env.glfwSetCharCallback
 - func[58] sig=12 <glfwGetTime> <- env.glfwGetTime
 - func[59] sig=5 <glfwSetMouseButtonCallback> <- env.glfwSetMouseButtonCallback
 - func[60] sig=5 <glfwSetCursorPosCallback> <- env.glfwSetCursorPosCallback
 - func[61] sig=5 <glfwSetScrollCallback> <- env.glfwSetScrollCallback
 - func[62] sig=5 <glfwSetCursorEnterCallback> <- env.glfwSetCursorEnterCallback
 - func[63] sig=3 <glfwGetJoystickName> <- env.glfwGetJoystickName
 - func[64] sig=2 <glfwSetInputMode> <- env.glfwSetInputMode
 - func[65] sig=3 <glfwJoystickPresent> <- env.glfwJoystickPresent
 - func[66] sig=2 <__ubsan_handle_shift_out_of_bounds> <- env.__ubsan_handle_shift_out_of_bounds
 - func[67] sig=13 <__imported_wasi_snapshot_preview1_clock_time_get> <- wasi_snapshot_preview1.clock_time_get
 - func[68] sig=0 <glfwSwapBuffers> <- env.glfwSwapBuffers
 - func[69] sig=5 <glfwGetGamepadState> <- env.glfwGetGamepadState
 - func[70] sig=8 <glfwWaitEvents> <- env.glfwWaitEvents
 - func[71] sig=8 <glfwPollEvents> <- env.glfwPollEvents
 - func[72] sig=3 <glfwWindowShouldClose> <- env.glfwWindowShouldClose
 - func[73] sig=1 <__ubsan_handle_negate_overflow> <- env.__ubsan_handle_negate_overflow
 - func[74] sig=5 <__imported_wasi_snapshot_preview1_fd_prestat_get> <- wasi_snapshot_preview1.fd_prestat_get
 - func[75] sig=14 <__imported_wasi_snapshot_preview1_fd_prestat_dir_name> <- wasi_snapshot_preview1.fd_prestat_dir_name
 - func[76] sig=15 <__imported_wasi_snapshot_preview1_path_open> <- wasi_snapshot_preview1.path_open
 - func[77] sig=5 <__imported_wasi_snapshot_preview1_fd_fdstat_set_flags> <- wasi_snapshot_preview1.fd_fdstat_set_flags
 - func[78] sig=4 <__imported_wasi_snapshot_preview1_fd_read> <- wasi_snapshot_preview1.fd_read
 - func[79] sig=9 <__imported_wasi_snapshot_preview1_path_filestat_get> <- wasi_snapshot_preview1.path_filestat_get
 - func[80] sig=0 <glfwDestroyWindow> <- env.glfwDestroyWindow
 - func[81] sig=5 <__imported_wasi_snapshot_preview1_random_get> <- wasi_snapshot_preview1.random_get
```

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -205,8 +205,9 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
     if (options.rmodels) {
         try c_source_files.append("src/rmodels.c");
     }
-    raylib_mod.addCMacro("SUPPORT_MODULE_RAUDIO", &.{@as(u8, @intFromBool(options.raudio)) + 0x30});
-    if (options.raudio) {
+    const raudio_enabled = options.raudio and options.platform != .wasigl;
+    raylib_mod.addCMacro("SUPPORT_MODULE_RAUDIO", &.{@as(u8, @intFromBool(raudio_enabled)) + 0x30});
+    if (raudio_enabled) {
         try c_source_files.append("src/raudio.c");
     }
 
@@ -360,6 +361,12 @@ fn compileRaylib(b: *std.Build, target: std.Build.ResolvedTarget, optimize: std.
             }
             raylib_mod.addCMacro(OpenglVersion.gl_soft.toCMacroStr(), "");
             raylib_mod.addCMacro("PLATFORM_MEMORY", "");
+        },
+        .wasigl => {
+            raylib_mod.addCMacro("PLATFORM_WASIGL", "");
+            raylib_mod.addCMacro("GRAPHICS_API_OPENGL_ES2", "");
+            raylib_mod.addIncludePath(b.path("src/external/glfw/include"));
+            raylib.import_symbols = true;
         },
         .win32 => {
             if (target.result.os.tag != .windows) {
@@ -573,17 +580,7 @@ pub const LinuxDisplayBackend = enum {
     Both,
 };
 
-pub const PlatformBackend = enum {
-    glfw,
-    rgfw,
-    sdl,
-    sdl2,
-    sdl3,
-    memory,
-    win32,
-    drm,
-    android,
-};
+pub const PlatformBackend = enum { glfw, rgfw, sdl, sdl2, sdl3, memory, win32, drm, android, wasigl };
 
 fn translateCMod(
     comptime header: []const u8,
@@ -687,6 +684,7 @@ fn addExamples(
         }
         if (std.mem.eql(u8, filename, "raylib_opengl_interop")) {
             if (platform == .drm) continue;
+            if (platform == .wasigl) continue;
             if (target.result.os.tag == .macos) continue;
             exe_mod.addIncludePath(b.path("src/external"));
         }
@@ -794,4 +792,3 @@ fn waylandGenerate(
         }
     }
 }
-
--- a/src/Makefile
+++ b/src/Makefile
@@ -578,6 +578,14 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_ANDROID)
     endif
 endif
 
+ifeq ($(PLATFORM),PLATFORM_WASIGL)
+    WASI_SDK ?= /opt/wasi-sdk
+    CC = $(WASI_SDK)/bin/clang
+    AR = $(WASI_SDK)/bin/llvm-ar
+    LDFLAGS += -Wl,--allow-undefined
+    CFLAGS += -DPLATFORM_WASIGL -DGRAPHICS_API_OPENGL_ES2 
+endif
+
 # Define library paths containing required libs: LDFLAGS
 # NOTE: This is only required for dynamic library generation
 #------------------------------------------------------------------------------------------------
--- a/src/external/wasigl_gles2.h
+++ b/src/external/wasigl_gles2.h
@@ -1,0 +1,307 @@
+// GLES2 header for PLATFORM_WASIGL.
+// GL functions are imported directly from the host (via @easywasm/gl), so GLAD's
+// runtime proc-address loading cannot work (imported functions have no stable WASM
+// table indices). This header reuses glad_gles2.h for types/constants, then undefines
+// the function-pointer macros and redeclares functions as plain extern so every GL
+// call compiles to a direct WASM `call` instruction against the host import.
+
+#ifndef WASIGL_GLES2_H
+#define WASIGL_GLES2_H
+
+// Pull in all types, constants, PFNGL typedefs, and extern glad_glXxx declarations
+// (no implementation — the variables are extern-declared but never referenced after
+// the macros below are replaced with direct calls).
+#include "glad_gles2.h"
+
+// Remove the function-pointer redirects added by glad_gles2.h so subsequent
+// glXxx() calls reach the extern functions declared below instead.
+#undef glActiveTexture
+#undef glAttachShader
+#undef glBindAttribLocation
+#undef glBindBuffer
+#undef glBindFramebuffer
+#undef glBindRenderbuffer
+#undef glBindTexture
+#undef glBlendColor
+#undef glBlendEquation
+#undef glBlendEquationSeparate
+#undef glBlendFunc
+#undef glBlendFuncSeparate
+#undef glBufferData
+#undef glBufferSubData
+#undef glCheckFramebufferStatus
+#undef glClear
+#undef glClearColor
+#undef glClearDepthf
+#undef glClearStencil
+#undef glColorMask
+#undef glCompileShader
+#undef glCompressedTexImage2D
+#undef glCompressedTexSubImage2D
+#undef glCopyTexImage2D
+#undef glCopyTexSubImage2D
+#undef glCreateProgram
+#undef glCreateShader
+#undef glCullFace
+#undef glDeleteBuffers
+#undef glDeleteFramebuffers
+#undef glDeleteProgram
+#undef glDeleteRenderbuffers
+#undef glDeleteShader
+#undef glDeleteTextures
+#undef glDepthFunc
+#undef glDepthMask
+#undef glDepthRangef
+#undef glDetachShader
+#undef glDisable
+#undef glDisableVertexAttribArray
+#undef glDrawArrays
+#undef glDrawElements
+#undef glEnable
+#undef glEnableVertexAttribArray
+#undef glFinish
+#undef glFlush
+#undef glFramebufferRenderbuffer
+#undef glFramebufferTexture2D
+#undef glFrontFace
+#undef glGenBuffers
+#undef glGenFramebuffers
+#undef glGenRenderbuffers
+#undef glGenTextures
+#undef glGenerateMipmap
+#undef glGetActiveAttrib
+#undef glGetActiveUniform
+#undef glGetAttachedShaders
+#undef glGetAttribLocation
+#undef glGetBooleanv
+#undef glGetBufferParameteriv
+#undef glGetError
+#undef glGetFloatv
+#undef glGetFramebufferAttachmentParameteriv
+#undef glGetIntegerv
+#undef glGetProgramInfoLog
+#undef glGetProgramiv
+#undef glGetRenderbufferParameteriv
+#undef glGetShaderInfoLog
+#undef glGetShaderPrecisionFormat
+#undef glGetShaderSource
+#undef glGetShaderiv
+#undef glGetString
+#undef glGetTexParameterfv
+#undef glGetTexParameteriv
+#undef glGetUniformLocation
+#undef glGetUniformfv
+#undef glGetUniformiv
+#undef glGetVertexAttribPointerv
+#undef glGetVertexAttribfv
+#undef glGetVertexAttribiv
+#undef glHint
+#undef glIsBuffer
+#undef glIsEnabled
+#undef glIsFramebuffer
+#undef glIsProgram
+#undef glIsRenderbuffer
+#undef glIsShader
+#undef glIsTexture
+#undef glLineWidth
+#undef glLinkProgram
+#undef glPixelStorei
+#undef glPolygonOffset
+#undef glReadPixels
+#undef glReleaseShaderCompiler
+#undef glRenderbufferStorage
+#undef glSampleCoverage
+#undef glScissor
+#undef glShaderBinary
+#undef glShaderSource
+#undef glStencilFunc
+#undef glStencilFuncSeparate
+#undef glStencilMask
+#undef glStencilMaskSeparate
+#undef glStencilOp
+#undef glStencilOpSeparate
+#undef glTexImage2D
+#undef glTexParameterf
+#undef glTexParameterfv
+#undef glTexParameteri
+#undef glTexParameteriv
+#undef glTexSubImage2D
+#undef glUniform1f
+#undef glUniform1fv
+#undef glUniform1i
+#undef glUniform1iv
+#undef glUniform2f
+#undef glUniform2fv
+#undef glUniform2i
+#undef glUniform2iv
+#undef glUniform3f
+#undef glUniform3fv
+#undef glUniform3i
+#undef glUniform3iv
+#undef glUniform4f
+#undef glUniform4fv
+#undef glUniform4i
+#undef glUniform4iv
+#undef glUniformMatrix2fv
+#undef glUniformMatrix3fv
+#undef glUniformMatrix4fv
+#undef glUseProgram
+#undef glValidateProgram
+#undef glVertexAttrib1f
+#undef glVertexAttrib1fv
+#undef glVertexAttrib2f
+#undef glVertexAttrib2fv
+#undef glVertexAttrib3f
+#undef glVertexAttrib3fv
+#undef glVertexAttrib4f
+#undef glVertexAttrib4fv
+#undef glVertexAttribPointer
+#undef glViewport
+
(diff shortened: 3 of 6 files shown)


konsumer:
This definitely needs more testing, but I wanted to get it up here for discussion.

I made [this tester](https://github.com/konsumer/example-5852) to show how it works. You can see it running [here](https://konsumer.js.org/example-5852/)

Essentially, just add gl/glfw/wasi imports, and stub the `ubsan` stuff.

build.zig
```diff
@@ -573,17 +580,7 @@ pub const LinuxDisplayBackend = enum {
...
-    rgfw,
-    sdl,
-    sdl2,
-    sdl3,
-    memory,
-    win32,
-    drm,
-    android,
-};
+pub const PlatformBackend = enum { glfw, rgfw, sdl, sdl2, sdl3, memory, win32, drm, android, wasigl };
```
CrackedPixel, line 583: nitpick: imo this is better like the original with separate lines

build.zig
```diff
@@ -794,4 +792,3 @@ fn waylandGenerate(
         }
     }
 }
-
```
CrackedPixel, line 797: this should probably be added back

src/Makefile
```diff
@@ -578,6 +578,14 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_ANDROID)
...
 endif
 
+ifeq ($(PLATFORM),PLATFORM_WASIGL)
+    WASI_SDK ?= /opt/wasi-sdk
+    CC = $(WASI_SDK)/bin/clang
+    AR = $(WASI_SDK)/bin/llvm-ar
+    LDFLAGS += -Wl,--allow-undefined
+    CFLAGS += -DPLATFORM_WASIGL -DGRAPHICS_API_OPENGL_ES2 
+endif
+
```
CrackedPixel, line 588: i think this would be better if it was separated out like the other platforms. so like the `CFLAGS` go with the other `CFLAGS`, the CC with the other CC, etc. instead of all together in one block

CrackedPixel:
in the example you have `-Wl,--export=update` and the "game loop" function is named `update`. where is that configured? is it something that can be set at runtime, similar to `emscripten_set_main_loop`?

have you tried other examples to document if/where it fails? audio not being compatible seems fine, but what about the other stuff that is expected to work on web?

CrackedPixel:
> I'm not totally sure if I modified enough files. I messed with build.zig and Makefile. Are there more?

for it to be totally complete there is also the cmake stuff and the examples. but imo its better to show it working in practice and get rays approval before adding it to that

CrackedPixel:
emscripten offers a few other options like copying a "resources" folder, having a single html page (no web server needed), custom html shell, etc. does wasigl offer any of these options?

CrackedPixel:
also suggestion: rename the PR something like `[Platform/WASIGL] new platform WIP` and flag as draft so its more clear that its for discussion and not ready to merge

konsumer:
> for it to be totally complete there is also the cmake stuff and the examples. but imo its better to show it working in practice and get rays approval before adding it to that

yep, agreed. I just wanted to show the concept before putting too much more work into it. Doing this is cmake is pretty easy too, but the whole build system is pretty complex (it took me a while to work out the build.zig, for example.) cmake has a very central way of building, like it just sort of expects you to already have your build-tools setup as defults. `emcmake` gets around this by setting it all up. cmake has pre-made build-target things that can do this too, like one comes with wasi-sdk, so I think it should just be a matter of a couple defines (not all the other extra flags and things in make/zig.)

> emscripten offers a few other options like copying a "resources" folder, having a single html page (no web server needed), custom html shell, etc. does wasigl offer any of these options?

No, this is another thing. I dislike emscripten's options for the most part, and find that you trade a great deal of complexity and bloat for something I don't really need (loading wasm in your own js/html is super-easy, and I wrote a library that mounts wasi filesystem you can implement yourself easily.) The fact that it works on web/js is secondary. it's a general wasm file, which can run on anything (native host, no js, etc.) Emscripten wasms don't run at all on wasmer, wamr, wasm3, etc. it's extremely locked in to a single target.

I even think this is out of scope for wasm, in general. like "does cmake automatically bundle assets into the exe?" isn't really something anyone would expect. You could write a script, or make it work more like that, but it only works for emscritpen because they essentially have 1 usecase.

I have made a few engines that use totally different systems (generally loading a zip for wasm+assets) and it all works easily with WASI, but it's not in scope for how wasm works. emscripten adding all this stuff is what locks it in to a specific target (js, web, etc.)

For context/examples, these things I made load assets from a zip file, but from the perspective of the code it's just normal filesystem (using WASI):

- [nullo](https://github.com/notnullgames/null0)
- [raycart](https://github.com/notnullgames/raycart)

I use [this library I made](https://github.com/easywasm/wasi) to do it on web, but I also implemented the same thing in native WASI, for null0 (so I can run the same zip files without a browser.)

> in the example you have `-Wl,--export=update` and the "game loop" function is named `update`. where is that configured? is it something that can be set at runtime, similar to `emscripten_set_main_loop`?

A plain export-function can serve a similar purpose. It's not really "configured" you just export whatever you want, and then call that, in the host. [Here](https://github.com/konsumer/example-5852/blob/main/docs/index.js#L41-L46) is me calling [`update` in the example](https://github.com/konsumer/example-5852/blob/main/example.c#L3-L9). In my [raycart](https://github.com/notnullgames/raycart), I use `CartUpdate`, since it kind of looks like raylib, so it seemed like a good standard export, but it's the same idea.  A key difference is that there is absolutely no API lock-in, outside of standard wasm exports/imports, so it's up to the user to set it up how they want (which is sort of the goal.) Using this trick, for example, I can setup wasi imports that use a zip file instead of a regular (or virtual) filesystem, or even preload some async-fetched files. Following a standard (WASI, glfw API, etc) really helps make reusable interfaces.

> have you tried other examples to document if/where it fails? audio not being compatible seems fine, but what about the other stuff that is expected to work on web?

I'm sure there are a lot of areas it will fail, but the platform itself doesn't need to change much. Like it may ask for more imports from host, for things that are missing, but the actual raylib part doesn't change. I will need to port more examples to find more things I need to implement in host, but the PR is more about the game-side (wasm) API which is just "ask for whatever you ned from host".

One trick I used to make my web wasi lib, initially, was to just build all my examples as if it's setup in host (using wasi-sdk) then analyze wasm directly to figure out what needs to be implemented. it's maybe a bit counterintuitive, but we know that the wasm-side is correct, mostly, so it can be scripted pretty easily. My [build script](https://github.com/notnullgames/raycart/blob/main/scripts/build_cart.js) for raycart kinda does this, and it made it easier to see how things work.

Here is an example:

```
Building cart: tetris
  carts/tetris/main.c -> /Users/konsumer/Desktop/raycart/host/web/carts/tetris.wasm
  Imports
    raylib.IsKeyPressed(i32) => i32
    raylib.GetRandomValue(i32, i32) => i32
    raylib.IsKeyDown(i32) => i32
    raylib.ClearBackground(i32) => null
    raylib.DrawLine(i32, i32, i32, i32, i32) => null
    raylib.DrawRectangle(i32, i32, i32, i32, i32) => null
    raylib.DrawText(i32, i32, i32, i32, i32) => null
    raylib.MeasureText(i32, i32) => i32
    raylib.GetScreenWidth() => i32
    raylib.GetScreenHeight() => i32
    raylib.InitWindow(i32, i32, i32) => null
    wasi_snapshot_preview1.fd_close(i32) => i32
    wasi_snapshot_preview1.fd_seek(i32, i64, i32, i32) => i32
    wasi_snapshot_preview1.fd_write(i32, i32, i32, i32) => i32
    wasi_snapshot_preview1.proc_exit(i32) => null
  Exports
    _start() => null
    CartInit() => null
    CartUpdate() => null
    malloc(i32) => i32
    free(i32) => null
```
Raycart works a bit different, since the ABI is wasi+raylib (not wasi+gl+glfw) but it's the same idea.

ABI is maybe the wrong acronym, since the ABI is just standard WASM import/exports, but like the shape of available imports.

konsumer:
More thoughts on "resources" and emscripten: It's a common problem I have with emscripten. it creates specific js bindings (`emscripten_set_main_loop` reaches into js host and sets up a rAF, for example) and js bindings to it's own virtual filesystem.

This is great, if you just want an existing native thing to run in js (build injects the files the game needs, and it runs on web, using JS shims for glfw, sdl, etc.) it breaks down when you want to use it anywhere else. WASI bindings are pretty simple for web or native (for null0, I just took an existing WASI layer for files, and switched it to use a zip.) It also breaks down if you want to interop with other wasm-hosts (wamr, wasm3, wasmer, etc, which all do not use js.)

For example, users can write null0 games in a bunch of languages, since the wasm import/exports are the same, and no change to the (wamr/web) hosts is needed. In my raycart example, any language can be used with the same idea of "the host will give me raylib API" and you can make your game in any language you like. It's not really a "language binding", more like a new platform that enables using lots of languages.

raysan5:
@konsumer This is a very interesting platform addition but I'm afraid I prefer to avoid it at the moment, WASI is still not stable enough, neither is a popular platform that many users could need; it would add extra maintenance cost prematurely to raylib that I prefer to avoid. Still, we can keep an eye on it for the future.

In any case, if it is added in the future, I'd avoid using GLFW dependency. Lately `rcore_web_emscripten` was added (still uncomplete in some aspects) just to avoid the GLFW dependency and make it as much self-contained as possible, despite the dependency on Emscripten SDK, that I also agree is quite bloated.

konsumer:
Understood. I think for my own stuff, I like the raycart-way better anyway, and I went back to emscripten for the web host part (so user code just imports raylib API, which is either emscritpen or native connect to raylib.) I also do not want to maintain a whole other thing, and I agree it would probably not be widely used. Emscripten has a lot of engineering behind it, and will do what most people need.

Currently, I can run "carts" (zip file of wasm + assets) that target raylib API, and are built with wasi-sdk (no emscripten for users) so the actual interface ends up nicer than glfw, anyway (it's raylib, over wasm-barrier). You can see some [examples](https://notnullgames.github.io/raycart). I think this style will make it easier to port it to more wasm-languages, much easier (targets raylib API, not glw + wasm-side raylib-wrapper.) The emscripten wasm and user-code shares memory, so there is very little copying back-and-forth, and it seems to perform pretty nicely.

PR #5855, opened 2026-05-10
[cmake] added `platform/win32` to cmake

CrackedPixel:
I'm not a cmake user nor a windows user, so it would be great if other people could also test this

I did the following 2 tests:
```bash
mkdir build
cd build
cmake .. -DPLATFORM=Win32
# open raylib.sln
# compile core_2d_camera
# verify win32 + opengl in logs
```
```bash
mkdir build
cd build
cmake .. -DPLATFORM=Win32 -DOPENGL_VERSION=Software
# open raylib.sln
# compile core_2d_camera
# verify win32 + rlsw in logs
```

### DIFF
--- a/cmake/LibraryConfigurations.cmake
+++ b/cmake/LibraryConfigurations.cmake
@@ -67,6 +67,21 @@ if (${PLATFORM} STREQUAL "Desktop")
         endif ()
     endif ()
 
+elseif (${PLATFORM} STREQUAL "Win32")
+    if ((NOT WIN32) AND (NOT CMAKE_C_COMPILER MATCHES "mingw|mingw32|mingw64"))
+        message(FATAL_ERROR "Win32 platform requires Windows or a cross compiler.")
+    endif ()
+
+    set(PLATFORM_CPP "PLATFORM_DESKTOP_WIN32")
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+
+    if (${OPENGL_VERSION} MATCHES "Software")
+        set(GRAPHICS "GRAPHICS_API_OPENGL_SOFTWARE")
+    endif ()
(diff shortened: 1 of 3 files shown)


examples/Makefile
```diff
@@ -794,6 +796,23 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW)
...
+    endif
+    ifeq ($(PLATFORM_OS),LINUX)
+        find . -type f -executable -delete
+        rm -fv *.o
+    endif
+    ifeq ($(PLATFORM_OS),OSX)
+        find . -type f -perm +ugo+x -delete
+        rm -f *.o
+    endif
+endif
```
CrackedPixel, line 815: this is just taken from `PLATFORM_DESKTOP_GLFW` above

cmake/LibraryConfigurations.cmake
```diff
@@ -91,6 +67,21 @@ if (${PLATFORM} STREQUAL "Desktop")
         endif ()
     endif ()
 
+elseif (${PLATFORM} STREQUAL "Win32")
+    if ((NOT WIN32) AND (NOT CMAKE_C_COMPILER MATCHES "mingw|mingw32|mingw64"))
+        message(FATAL_ERROR "Win32 platform requires Windows or a cross compiler.")
```
CrackedPixel, line 72: i havent been able to get these to cross compile, but with the right settings it should work 🤷‍♂️ may as well make it an option

raysan5: ok

raysan5:
@CrackedPixel It looks good to me but something with more CMake experience should review it. In any case, thanks for the addition.

PR #5857, opened 2026-05-10
[raudio]: free converterResidual in UnloadSoundAlias to prevent leak

Peter0x44:
The example audio_sound_multi was leaking memory every single time the spacebar was pressed.

```
Direct leak of 576 byte(s) in 9 object(s) allocated from:
    #0 0x758a41019447 in calloc (/usr/lib/liblsan.so.0+0x19447) (BuildId: 8ee115309adc591d231c961c43d245cfa68d9aa7)
    #1 0x562dfbd2c4f3 in LoadAudioBuffer (/home/peter/raylib/examples/audio/audio_sound_multi+0xfa4f3) (BuildId: ea2a6f45d724abeccf904143a32012266f259f93)
```

This patch fixes that leak.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -1048,6 +1048,7 @@ void UnloadSoundAlias(Sound alias)
     {
         UntrackAudioBuffer(alias.stream.buffer);
         ma_data_converter_uninit(&alias.stream.buffer->converter, NULL);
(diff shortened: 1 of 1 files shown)


raysan5:
@Peter0x44 thanks for the improvement

Issue #1989, opened 2021-09-13
Create Multiple Rendering Windows.


sparkskapil:
Is it possible to create multiple windows and render different scenes on those windows?

DeveloperViraj:
Yes. Of course.

sparkskapil:
Can you please explain how can I achieve this?

MaldoPay:
Have anyone achieved that ? Is there any plan to include such thing in future releases @raysan5  ?

JaydenSu:
> Have anyone achieved that ? Is there any plan to include such thing in future releases @raysan5 ?

+1
I really wish raylib had a feature to create more than one window.

orcmid:
Ramon has answered this above.  There are ways to try it on your own.

You might take a look at RayGui also.

Architecturally, the `InitWindow( )` operates in a gobal static data structure and all of the functions rely directly on the existence of that single global structure and things glued to it from time to time.

Currently, there is no way to speak of "which" window.

That could be handled with a different model using objects, so that a counterpart of InitWindow produced an object and the raylib API involved methods on a given window, as appropriate.  There would probably be other spin-off objects.  That's what GUI systems of various vintages accomplish by passing handles around and needing to be given handles to know what to operate on.

That's a dramatic architectural change and it would really be a *different* product.  rayWin, say, instead of rayLib.  (Probably not "Win" because it implies a platform, but you get the idea.)

It would also change the programming model considerably and callbacks would have to be changed when an event is specific to a particular window and/or focus.  Platform integration might be trickier too.

This would also have a significant impact on binding support, although it would be a better fit with C++ and Java and probably even relying on COM, including flavors found on Linux and JNI.  Not so much Python, but possible.  You'd end up wanting responsive windows for working on different device capabilities including tablets and phones.  That can go on forever.  

All in all, this might be better simply using platform native GUI and windowing provisions.  There are, of course, a variety of these already.

It would get rid of name conflicts with libraries though :).

Probably not your grandfather's raylib :).

jthomazini:
Hello beautiful people, I am relatively new to the Raylib world. I am finishing a product using Raylib and it would be immensely better to have multi-window capability. Since it is not backed in the default library I have analyzed the piece that Derpyzza put together here: https://github.com/raysan5/raylib/wiki/Use-multiple-windows, although I believe that his instructions were for an older version of Raylib than 4.5 and they didn't work quite well, however, he pointed me to the right direction. Any way thank you very much Derpyzza.

[...]

PR #5859, opened 2026-05-10
[rcore][RGFW] Add fullscreen behaviour on size 0 window creation

keks137:
rcore_desktop_glfw has a nice feature of automatically creating a fullscreen window if you InitWindow(0,0,"title")
but on rgfw it instead just got stuck saying "failed to create drawable"
so I mostly just copied the logic, including the comments, from the glfw backend to handle this case
it does mean we now need to call RGFW_init ahead of time, to be able to get the primary monitor in the case of a size zero window

### DIFF
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -1674,8 +1674,12 @@ void PollInputEvents(void)
 // Initialize platform: graphics, inputs and more
 int InitPlatform(void)
 {
+    int result = RGFW_init();
+    if (result != 0) { TRACELOG(LOG_WARNING, "RGFW: Failed to initialize RGFW"); return -1; }
     // Initialize RGFW internal global state, only required systems
     unsigned int flags = RGFW_windowCenter | RGFW_windowAllowDND;
+
+    if ((CORE.Window.screen.width == 0) || (CORE.Window.screen.height == 0)) FLAG_SET(CORE.Window.flags, FLAG_FULLSCREEN_MODE);
 
     // Check window creation flags
     if (FLAG_IS_SET(CORE.Window.flags, FLAG_FULLSCREEN_MODE))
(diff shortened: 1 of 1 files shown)


raysan5:
@CrackedPixel Please, could you take a look to this PR? It surprises me that `RGFW_init()` was not called before... maybe it's not being tested against right RGFW version?

keks137:
> @CrackedPixel Please, could you take a look to this PR? It surprises me that `RGFW_init()` was not called before... maybe it's not being tested against right RGFW version?

it usually automatically does it internally on createwindow for convenience
but we can't rely on that if we want the display size before that, it wouldn't be initialized yet and would segfault
so we need the explicit call

CrackedPixel:
I don't really like a lot of this PR but i will look at it today and compare to glfw and other platforms to see about a good solution

CrackedPixel:
@keks137 i made a PR for suggestions to your branch: https://github.com/keks137/raylib/pull/1

keks137:
the changes seemed reasonable
I also just noticed something, that's what cbc5a1f is for
I missed that I was doing the FLAG_FULLSCREEN_MODE check twice, because I just copied glfw's version and worked from there, but the rgfw path already had a check above
they definitely should just be one check

CrackedPixel:
LGTM! nice work :sunglasses:

keks137:
actually one thing to note is that the return value of RGFW_init is rather weird:

from RGFW's source code:
/**!
 * @brief Initializes the RGFW library.
 * @return 0 on success, or a negative error code on failure.
 * @note This is automatically called when the first window is created.
*/

but if you look at the internal code, when you call it but it is already initialized it returns 1 - which is kind of weird but technically still correct within the api

where it gets truly weird is that it actually *doesn't* always return a negative number on error
for example in the wayland init there's this bit of code
```c
if (_RGFW->compositor == NULL) {
		RGFW_sendDebugInfo(RGFW_typeError, RGFW_errWayland, "Can't find compositor.");
		return 1;
	}
```

this means we can't just check < 0, unlike the comment implies

for the purposes of raylib, breaking on multiple windows is technically not an issue, since that's not supported anyways, right?
https://github.com/raysan5/raylib/issues/1989#issuecomment-918256167

this might however still be an issue on 
InitWindow
CloseWindow
and then
InitWindow
again
I'm not sure if that is something raylib supports
(one thing of note, this currently works in glfw, but segfaults in rgfw even without this specific issue, meaning I also can't check if this specifically actually doesn't work, but I'll assume it doesn't)

so maybe it would be better to ignore the return value entirely? that's what rgfw does whenever calling it internally
this return value is definitely a janky part of rgfw, maybe I should *also* open an issue there

raysan5:
@keks137 thanks for the improvement and thanks @CrackedPixel for the review

keks137:
it seems we will be able to remove the RGFW_init call entirely in a future update of RGFW, the fact that we need it right now is a bug https://github.com/ColleagueRiley/RGFW/issues/523#issuecomment-4425687138

Issue #5761, opened 2026-04-15
Feature request: support offscreen MSAA render targets


IcyVizi:
## Feature request

Please consider adding first-class support for **offscreen MSAA render targets** in `rlgl` / `raylib`.

Today `rlgl` already exposes enough pieces to attach and resolve framebuffers:

- `rlLoadFramebuffer()`
- `rlFramebufferAttach()`
- `rlFramebufferComplete()`
- `rlBindFramebuffer()`
- `rlBlitFramebuffer()`

It also supports attaching a renderbuffer via `RL_ATTACHMENT_RENDERBUFFER`.

However, there is currently no public helper to **create and allocate a multisampled color/depth renderbuffer**, so an offscreen MSAA pipeline still requires dropping down to raw backend GL calls such as:

- `glGenRenderbuffers`
- `glBindRenderbuffer`
- `glRenderbufferStorageMultisample`
- `glDeleteRenderbuffers`

On Windows this also means manual proc loading, even though the rest of the render path can stay inside `rlgl`.

## Why this would be useful

Default-window MSAA is not enough when rendering goes through an intermediate render target.

Common cases:

- UI frameworks rendering to an offscreen texture before compositing
- post-process pipelines that need a clean pre-resolve edge result
- editor/viewport panels rendered into textures
- situations where FXAA/TAA are undesirable because they blur text or UI glyphs

In these cases, offscreen MSAA is the correct solution, but it is hard to implement cleanly with the current public API.

## Possible API directions

Either a low-level `rlgl` API or a higher-level `raylib` API would help.

Low-level option:

- `rlLoadRenderbuffer(...)`
- `rlLoadRenderbufferMultisample(...)`
- `rlUnloadRenderbuffer(...)`

High-level option:

- `LoadRenderTextureMultisample(width, height, samples)`
- plus a helper to resolve it into a normal texture/render target

Even just the low-level renderbuffer helpers would already remove the need for raw GL access in user code.

## Expected behavior

When the backend supports it, users should be able to:

1. Create a multisampled offscreen color/depth target
2. Render into it normally
3. Resolve/blit it into a non-MSAA texture for sampling/compositing

If backend support is limited, it could still be exposed only for the relevant OpenGL backends.

Thanks.

raysan5:
@IcyVizi Supporting multi-sample fbo implies a considerabke amount of code to maintain and not that many users need it. I'm afraid I prefer to keep it out of scope of raylib for now.

Issue #65, opened 2015-12-13
Store multisampling in FBO


victorfisac:
When using post-production shaders, the multisampling is not stored in FBO so the multisampling hint "SetConfigFlags()" doesn't make any difference...

I found some threads by the Internet about this issue:

- https://www.opengl.org/discussion_boards/showthread.php/167446-Multisample-FBO-and-render-to-depth-texture

- http://stackoverflow.com/questions/20482851/how-to-use-multisampling-with-opengl-fbos

I tryed making some lines of code but anything worked (any difference or "frame buffer object couldn't be created"...).

The same when compiling for web.

raysan5:
Hello @victorfisac ,

`SetConfigFlags(FLAG_MSAA_4X_HINT)` tryes to setup multisampling on main framebuffer, it depends on OpenGL version (> 3.0) and hardware support ([ARB_multisample extension](https://www.opengl.org/registry/specs/ARB/multisample.txt)). Setting multisampling on FBO requires manual setup; the default FBO used for post-processing effects has no multisampling enabled, it will be consume too much resources, specially for embeded and low-end devices.

If you desire to implement an antialiasing algorythm to your post-produced image, my recomendation is using [FXAA antialiasing](https://en.wikipedia.org/wiki/Fast_approximate_anti-aliasing). It can be directly implemented on the fragment shader and it's very fast. You can find more information here:

http://www.kotaku.com.au/2011/12/what-is-fxaa/
http://www.geeks3d.com/20110405/fxaa-fast-approximate-anti-aliasing-demo-glsl-opengl-test-radeon-geforce/

victorfisac:
Found a working FXAA fragment shader! Here is an adapted to raylib version: http://pastebin.com/ZE6w1LKi (changed to version 330 and changed var names to fragTexCoord and fragColor. The shader just needs the screen resolution to work.

Source: https://www.opengl.org/discussion_boards/showthread.php/184192-GLSL-FXAA-rendering-off-screen

I think maybe it can be optimized a little bit... not sure. Tested with shaders_custom_uniform.c and dwarf model, nice result! :D

raysan5:
Thanks @victorfisac! FXAA fragment shader will be added to raylib shaders collection on next version!

PR #5860, opened 2026-05-10
[rlgl] Add multisampled attachments to framebuffer.

steampuker:
Offscreen multisampling is a widely requested feature, see [here](https://github.com/raysan5/raylib/issues/65) and [here](https://github.com/raysan5/raylib/issues/5761).

This PR adds functions to load multisampled textures/renderbuffers plus a function to copy the contents of an MSAA framebuffer to non-MSAA framebuffer. Also changed rlFramebufferAttach to support them.


raysan5:
@steampuker Thanks but there is no plans to implement this feature in raylib, already proposed in the past. Most users don't need it and it adds more code to maintain.

steampuker:
Okay then, just thought this would be useful for those who use rlgl standalone and don't want to use OpenGL loaders.

With rlGetProcAddress, it should be possible to implement it in a particular project, maybe this PR will give pointers to those who want to.

Issue #5856, opened 2026-05-10
[rtext] Font files `.fnt` can only be loaded by `LoadFont()`, not with `LoadFontEx()` and `LoadFontFromMemory()`


Farbs:
### Issue description

I was trying to load BMFont / Angelcode .fnt files using LoadFontEx, however it silently failed and returned the default font. Looking at the code for LoadFontEx, it seems that neither it nor LoadFontFromMemory have a pathway through to LoadBMFont, and that only LoadFont supports it.

I've added a note to the Using BMFonts page in the wiki in case it saves anyone else a bit of time.

### Environment

Desktop, Windows 10, OpenGL 3.3, GTX1060

### Code Example

Font doesNotWork = LoadFontEx("example.fnt", 32, NULL, 0);
Font works = LoadFont("example.fnt");

raysan5:
@Farbs Good catch! Thanks for reporting! I imagine I just ignored that format because it is not the most common use case and `LoadFontEx()` is intended for a codepoint list request, while .fnt has already a codepoint list defined that user can not change... Still, it can be improved...

CrackedPixel:
@raysan5 i was going to take a crack at this, but `fnt` files have separate files for resources. `N` amount, so its not a simple solution and raylib doesnt currently have a pattern for reading multiple files from memory

i made a PR to [add a warning](https://github.com/raysan5/raylib/pull/5864) but i think this cant be pursued further until there is a standard pattern in raylib

apilypas:
I have noticed that `LoadFontFromMemory` does not work with `.fnt` fonts too. It does not give any warning or error - just uses default font. I am trying to implement resource loading from separate package (zip file) and I don't see easy way to move these fonts there, while moving textures or audio files are super easy.

raysan5:
Reviewed this issue, effectively .fnt can not work with `LoadFontFromMemory()` (called by `LoadFontEx()`) because it requires additional external resources (one or more images). Also, font size and required codepoints can not be configured dynamically because are already fixed by the .fnt file. Consequently `LoadFont()` is the only sane option for .fnt.

A warning has been added to let users know.

PR #5864, opened 2026-05-12
[rtext] add warning for loadfontex/loadfontfrommemory

CrackedPixel:
Added warning to `LoadFontEx`/`LoadFontFromMemory`, brought up from https://github.com/raysan5/raylib/issues/5856 where some font formats are unavailable to load from memory

the easiest way to see this is to make a small change to `text/text_unicode_emojis`
```c
Font fontDefault = LoadFont("resources/dejavu.fnt");
// =>
Font fontDefault = LoadFontEx("resources/dejavu.fnt", 32, NULL, 95);
```
```
INFO: FILEIO: [resources/dejavu.fnt] File loaded successfully
WARNING: FONT: Font is not supported by LoadFontEx/LoadFontFromMemory or no glyphs found, reverted to default font
```

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -580,8 +580,13 @@ Font LoadFontFromMemory(const char *fileType, const unsigned char *fileData, int
 
         TRACELOG(LOG_INFO, "FONT: Data loaded successfully (%i pixel size | %i glyphs)", font.baseSize, font.glyphCount);
     }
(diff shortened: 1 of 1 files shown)


raysan5:
@CrackedPixel thanks, looks good as a quick solution but it requires some further review...

PR #5865, opened 2026-05-13
[cmake]: remove GLFW_BUILD warnings for when platform is SDL

jestarray:
This warning is for people who have `GLFW_BUILD_WAYLAND` and `GLFW_BUILD_X11` disabled but this only matters when you're compiling for GLFW platform


Peter0x44:
Honestly, I'd just delete this whole if (). It doesn't need to exist.

PR #5866, opened 2026-05-14
BINDINGS.md: update golang binding to v6.0

JupiterRider:
Hello,

raylib-go now supports raylib 6.0 ( see https://github.com/gen2brain/raylib-go/pull/544 ).

It is not tagged yet, but ready use.


raysan5:
@JupiterRider thanks for the update

Issue #5854, opened 2026-05-09
[rcore] Unable to alt-tab/switch windows when `fullscreen` or `borderless-windowed` modes are enabled


Ethanthegrand:
### Issue description

Pretty simple brief, I am unable to alt-tab or switch windows when my raylib app is in either **Fullscreen** or **BorderlessWindowed** mode when using either of the two functions: 
- ToggleFullscreen()
- ToggleBorderlessWindowed()

This appears to be an issue since raylib 5.x, up to 6.0 (the latest version) on Windows 11. This does not occur on Linux

Tried to record an example video with OBS, however the output video was black whenever i was fullscreen/borderless, despite using BorderlessWindowedMode with Display Capture, which is also an issue.

both fullscreen and borderless also appear to be behaving exactly the same too on my machine, though im unsure if this is expected behaviour or not.

### Environment

- Windows 11 64-bit
- OpenGL 3.3.0
- NVIDIA GeForce RTX 4070 

### Code Example

I just used the window flags example file provided by raylib.

```
if (IsKeyPressed(KEY_B)) ToggleBorderlessWindowed();
```

Also thanks to @Peter0x44 for helping me discover this bug

casperbear:
I also had an issue with alt-tab-ing from fullscreen window on Windows and solved it this way in the main event handling loop
```
    while (!WindowShouldClose() && !windowShouldClose)
    {
        if (!IsWindowFocused() && IsWindowFullscreen())
        {
            MinimizeWindow();
            BeginDrawing();
            EndDrawing();
            continue;
        }
      // other code
   }
```

Ethanthegrand:
When can we expect this to get fixed? 🙏

raysan5:
@Ethanthegrand this is an open source project, it could take days, weeks, months, years or never.

SoloByte:
@Ethanthegrand 

My solution to this problem on my end is to watch for focus changes of the window. Basically if the window is unfocused and fullscreen is currently enabled, I disable fullscreen and I clear the window topmost flag. I remember that I canceled fullscreen and once the window is focused again I apply both the fullscreen and topmost flag again. This works really well. 

The only problem with this is, that it does not work for borderless fullscreen mode. Currently in the 6.0 version of raylib, when you disable borderless fullscreen on an unfocused window, it automatically focuses the window again...

tillua467:
[...]

PR #5867, opened 2026-05-14
[rcore][GLFW] Prevent GLFW fullscreen from always being on top

auwsmit:
fixes this issue https://github.com/raysan5/raylib/issues/5854

Seems to be as simple as changing a flag from "TOPMOST" to "TOP" :)


raysan5:
@auwsmit I'm afraid this issue belongs to an external library and should be addressed on its own repo, not here. Or it will be overriden on  library update.

PR #5868, opened 2026-05-14
[rcore][GLFW] Fix `borderless-windowed` mode being always on top

auwsmit:
Don't pass a monitor to `glfwSetWindowMonitor()`. This avoids setting the flag internal to glfw which keeps the window always on top.

BorderlessWindowed still works fine, and is properly detected on my two monitors at different resolutions.

Addresses this issue: https://github.com/raysan5/raylib/issues/5854, but only for borderless FS.

A workaround for regular aka "exclusive" Fullscreen is mentioned in that thread, just add this to raylib's loop:

    if (!IsWindowFocused() && IsWindowFullscreen())
    {
        MinimizeWindow();
    }

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -287,7 +287,7 @@ void ToggleBorderlessWindowed(void)
                 CORE.Window.screen.height = mode->height;
 
                 // Set screen position and size
-                glfwSetWindowMonitor(platform.handle, monitors[monitor], CORE.Window.position.x, CORE.Window.position.y,
+                glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y,
                     CORE.Window.screen.width, CORE.Window.screen.height, mode->refreshRate);
 
                 // Refocus window
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -287,8 +287,14 @@ void ToggleBorderlessWindowed(void)
                 CORE.Window.screen.height = mode->height;
 
                 // Set screen position and size
+            #if defined(_WIN32)
+                // NOTE: To prevent the fullscreen window from always staying on top, don't pass a monitor at this point.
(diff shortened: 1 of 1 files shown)


raysan5:
@auwsmit This is a sensible change, have you tested it on Windows, Linux, macOS and with multiple monitors and multiple DPIs?

auwsmit:
@raysan5 Testing on Linux, it does create a problem where the taskbar stays on top of the borderless window (at least for my laptop with wayland and KDE).

So is the best approach to do an `#if windows` so this change only affects Windows? Is something like this code snippet the proper way to do it?:

```
                // Set screen position and size
            #if defined(_WIN32)
                // NOTE: To prevent the window from always staying on top, don't pass a monitor at this point.
                glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y,
                    CORE.Window.screen.width, CORE.Window.screen.height, mode->refreshRate);
            #else
                glfwSetWindowMonitor(platform.handle, monitors[monitor], CORE.Window.position.x, CORE.Window.position.y,
                    CORE.Window.screen.width, CORE.Window.screen.height, mode->refreshRate);
            #endif
```

Also sorry about closing the other issue and making this one. I realize in retrospect I probably should've just modified the old PR with a new commit and didnt have to close it and make this new one, but now I know not to that in the future.

raysan5:
As I commented, this is a sensible change, it should be tested on a multi-monitor configuration and with different DPI monitors, I think the change can impact current behaviour, for example have you tried opening the window from a different working monitor instead of primary monitor? because I think that NULL always uses primary instead of current...

auwsmit:
@raysan5 Did thorough testing on Windows with two monitors on two different PCs, at different resolutions and refresh rates. Everything seems to behave the same as before, except you can alt tab properly. 👍

I believe it is still able to get the correct monitor size and position from the [earlier call to `glfwGetMonitorPos()`](https://github.com/auwsmit/raylib/blob/0f8c40a2064ab3c42447e831e507db76f3c3fdd5/src/platforms/rcore_desktop_glfw.c#L285) which remains unchanged.

raysan5:
@auwsmit thanks for the improvement, tested and working as expected

PR #5869, opened 2026-05-15
Fix build.zig building examples all the time

jn-jairo:
The `installArtifact` was adding the examples to be built and installed when building the raylib even when not asking to build the examples.

The `addInstallArtifact` already adds it to be built and installed in the correct folder when asked to build the examples.


raysan5:
@jn-jairo thanks for the review

PR #5875, opened 2026-05-19
[rshapes] Fix `DrawRectangleRounded` automatic segments count to be consistent with `DrawRectangleRoundedLinesEx`

anstropleuton:
The `DrawRectangleRounded` when set segments count < 4 automatically calculates number of segments needed to draw smoother corners, but it divides the count by 4 instead of 2 which `DrawRectangleRoundedLinesEx` does. This fix makes `DrawRectangleRounded` use the same math as `DrawRectangleRoundedLinesEx`.

Before fix:
image

Notice fewer segments for the `DrawRectangleRounded` (gray colored box) compared to the `DrawRectangleRoundedLinesEx` (black colored outline) with same rectangle.

After fix:
image


raysan5:
@anstropleuton thanks for the review!

PR #5881, opened 2026-05-23
[rcore][iOS] New platform support (device/simulator)

vsaint1:
## [RFC] iOS Platform Support

This PR adds initial iOS platform support to raylib, covering the full integration needed to run the default examples on device and simulator.

### What's included

- **Window / display creation** — UIKit-backed window setup with EAGL framebuffer and retina/native scale support
- **Touch input** — touch  mapped to raylib's existing touch and mouse-compatible input API
- **File system / bundle paths** — `GetApplicationDirectory()` and asset loading resolve correctly against the `.app` bundle
- **Build system / Xcode integration** — Xcode project structure and build settings to compile raylib as a static lib for iOS targets (device + simulator)
- **No** Extra code paths/pre-processor dependencies needed

All default raylib examples build and run under this implementation. 

### Known limitation

**Audio is currently non-functional.** miniaudio requires compilation as Objective-C (`.m`) rather than C (`.c`) on iOS due to AVFoundation/CoreAudio backend dependencies. 

### Not yet covered

- Audio (see above)
- Keyboard

### Testing

Tested on:
- [x] iOS Simulator (iPhone 17 PRO)
- [x] Physical device (iPhone 15)

https://youtu.be/ccIeg2oWwdA POC video

### Questions for reviewers

- Any existing platform abstraction conventions I should align with from the Android or web backends?

---

Happy to iterate on any of this


ghera:
Hi, thanks for working on this.

My main concern with this approach is that it relies directly on EAGLContext / CAEAGLLayer / OpenGLES, which Apple deprecated starting with iOS 12.

If you have a chance, please take a look at my fork (https://github.com/ghera/raylib-iOS), which builds on work from a previously closed PR. My fork explores a different iOS backend based on EGL/ANGLE, which uses Metal under the hood, and it already handles things like audio, rotation/resizing, and other platform integration details. I actually shipped an Android/iOS game with it, so it has seen some real-world use.

The original PR was closed because of the main loop approach, so it would be interesting to understand whether your version is closer to what would fit raylib upstream. Speaking of that, I also still have a few performance concerns: the classic while (!WindowShouldClose()) loop is running on a worker thread and is paced with SetTargetFPS() / sleep-based timing, rather than CADisplayLink. That may still work fine, but it could also help explain the CPU spikes / less stable pacing I’m seeing on some screens in your example, for example around 50-60% CPU on the game over screen in the simulator.

In any case, I’d be more than happy to compare approaches or combine efforts if that would be useful. I guess we should also wait for @raysan5’s feedback on the overall direction here.

src/raylib.h
```diff
@@ -113,6 +113,11 @@
     #define RLAPI       // Functions defined as 'extern' by default (implicit specifiers)
 #endif
 
+// iOS owns the real main() via UIApplicationMain.
```
raysan5: I prefer to avoid this kind of platform-specific hacks in the raylib header

src/platforms/rcore_ios.c
```diff
@@ -0,0 +1,553 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#if defined(__APPLE__)
+    #include <unistd.h>
+#endif
+
+// UIKit/OpenGL bridge functions implemented in rcore_ios_main.m
+RLAPI bool ios_initialize_window(int requestedWidth, int requestedHeight, int *screenWidth, int *screenHeight, int *renderWidth, int *renderHeight, float *scaleX, float *scaleY);
```
raysan5, line 34: Why are those functions needed? Probably some of them can b directly implemented inside the available ones. Also, please review raylib coding conventions, those functions do not follow the conventions.

src/platforms/rcore_ios.c
```diff
@@ -0,0 +1,553 @@
...
+RLAPI void ios_make_current_context(void);
+RLAPI void ios_shutdown_window(void);
```
raysan5: For example this one, just add code inside ClosePlatform()

src/platforms/rcore_ios.c
```diff
@@ -0,0 +1,553 @@
...
+
+//----------------------------------------------------------------------------------
+// Types and Structures Definition
+//----------------------------------------------------------------------------------
+typedef struct {
+    bool initialized;
+} PlatformData;
+
+typedef struct {
+    int32_t pointCount;                     // Number of active touch points
```
raysan5: PlatformData structure is intended for specific platform data, no need to create additional structures

src/CMakeLists.txt
```diff
@@ -38,6 +38,12 @@ set(raylib_sources
     rtextures.c
     )
 
+if (${PLATFORM} STREQUAL "iOS")
+    list(APPEND raylib_sources
+        rcore_ios_main.m
```
raysan5: I still have doubts that Objective-C is really needed, afaik, objc functions can be called from C getting the function pointers.

src/rlgl.h
```diff
@@ -876,9 +876,16 @@ RLAPI void rlLoadDrawQuad(void);     // Load and draw a quad
 #endif
 
 #if defined(GRAPHICS_API_OPENGL_ES3)
-    #include <GLES3/gl3.h>              // OpenGL ES 3.0 library
-    #define GL_GLEXT_PROTOTYPES
-    #include <GLES2/gl2ext.h>           // OpenGL ES 2.0 extensions library
+    #if defined(PLATFORM_IOS)
```
raysan5, line 879: `rlgl` should not check for any `PLATFORM_`-specific. Maybe there is some other compile time flag for IOS?

src/platforms/rcore_ios.c
maiconpintoabreu: Of it need to be exported maybe it could be implemented in the dev side.

src/rlgl.h
vsaint1: sadly this was the only option that came to mind, also i dont know if theres other solution to this.

src/raylib.h
vsaint1: This can be moved to `cmake` without any problem

src/platforms/rcore_ios.c
raysan5: do you mean in user side? but then maybe some platform-dependant hooks will need to be provided by raylib...

src/rlgl.h
raysan5: what about `#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED)`?

ghera: you could replace the explicit PLATFORM_IOS check with `__has_include`, but `rlgl.h` already contains other `PLATFORM_*` checks, so that would mostly be a cosmetic change.

The broader issue is that this code path still depends on OpenGLES, which has been deprecated on Apple platforms since 2018. In practice, the most workable solution I found was using ANGLE (I know it's a significant dependency), which provides a Metal-backed implementation.

src/platforms/rcore_ios.c
ghera: the C main wrapper and all of rcore_ios_main.m could also be folded into rcore_ios.c (see https://github.com/ghera/raylib-iOS/blob/master/src/platforms/rcore_ios.c), but then rcore.c must be compiled as Objective-C on iOS.

src/CMakeLists.txt
ghera: Technically, we could avoid direct ObjC syntax by using objc_getClass and similar runtime wrappers, but it comes with significant downsides: much worse readability, very fragile objc_msgSend casts and harder class/delegate/view-controller setup. A particularly important limitation is blocks: many modern UIKit/Foundation APIs use Objective-C blocks, and those are really awkward to handle cleanly from plain C.

ghera:
Hi, I made a more thorough architectural analysis of the iOS platform code in `src/platforms/rcore_ios.c` and `src/rcore_ios_main.m`. Below I summarize the issues that, in my opinion, should be addressed before merge (aside from the aforementioned OpenGL ES deprecation).

### Issue 1: Unreferenced Background Thread + Generic Race Conditions

The game loop runs on a thread pulled from `dispatch_get_global_queue` — a **concurrent** queue shared with system frameworks and other libraries.

**What happens:**

```objc
dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
    raylib_main(ios.argc, ios.argv);
});
```

- The thread has **no identity** — invisible in Xcode debugger, no name in crash logs
- The queue cannot be referenced for synchronization (`dispatch_sync` to a concurrent queue runs inline on the caller via GCD fast-path)
- Multiple entries in the same function or external libraries may share the same pool
- Shared state (`CORE.Input.Touch`, `CORE.Window.flags`, etc.) is accessed from two threads (main + bg) with **no explicit synchronization** — memory ordering depends on compiler luck

**Proposed fix:** Replace the global concurrent queue with a **dedicated serial queue** for the game loop.

```objc
// Instead of dispatch_get_global_queue:
raylibQueue = dispatch_queue_create("com.raylib.engine", NULL);
dispatch_async(raylibQueue, ^{
    raylib_main(ios.argc, ios.argv);
});
```

This gives the thread an identity (visible in Xcode debugger, Instruments, crash logs) and makes it referenceable for future synchronization needs.

**Consequence: shutdown dispatch must change too.** With a serial queue, `dispatch_sync(main_queue, shutdownBlock)` in `ios_shutdown_window()` can deadlock if the main thread is blocked waiting for `raylibQueue`. With the original concurrent queue this was hidden by GCD's fast-path inline optimization — with a serial queue it's deterministic.

Fix: use `dispatch_async` instead — non-blocking:

```objc
dispatch_async(dispatch_get_main_queue(), shutdownBlock);
```

`ios.shuttingDown` stays `true` to prevent re-entry until the async block executes.

**Possible race conditions (low severity in practice):**
- `CORE.Input.Touch` — main thread writes (touchesBegan:) vs bg thread reads (PollInputEvents:), no explicit barrier. On Apple Silicon, aligned 8-byte loads/stores are atomic natively, so `Vector2` tearing should not occur.
- `CORE.Window.flags` — `ios_set_window_focused()` from main thread vs `IsWindowFocused()` from bg thread. Focus detection may be 1 frame delayed, not a correctness issue.

### Issue 2: No Display Frame Sync (Busy-Loop)

**What happens:** `ios_present_frame()` returns immediately. No `CADisplayLink`, no semaphore, no `CFRunLoopRunInMode`. The `raylib_main` while-loop spins at maximum CPU speed.

**Impact:** CPU spikes on static scenes → battery drain. Screen tearing (buffer swap not aligned to vsync). Inconsistent framerate (device overheats, throttles, fps drops).

**Proposed fix:** Add `CADisplayLink` + `dispatch_semaphore` for frame pacing.

A `vsyncCount` is incremented each display frame. `ios_present_frame()` waits for the next vsync before returning:

```objc
// Display link callback (main thread run loop)
+ (void)onDisplayFrame:(CADisplayLink *)link {
    vsyncCount++;
    dispatch_semaphore_signal(frameSem);
}

// In ios_present_frame (bg thread)
void ios_present_frame(void) {
    [ios.context presentRenderbuffer:GL_RENDERBUFFER];

    // Wait for next vsync — CPU sleeps here
    static uint64_t lastVsync = 0;
    while (vsyncCount - lastVsync < 1)
        dispatch_semaphore_wait(frameSem, DISPATCH_TIME_FOREVER);
    lastVsync = vsyncCount;
}
```

---

@vsaint1, the solutions above are just suggestions — feel free to take, adapt, or ignore anything :)

---

That said, I want to be honest: these patches are improvements, but they treat symptoms. The background thread exists solely to preserve the desktop while-loop API on a platform that was not designed for it. iOS uses a run-loop-driven model (`UIApplicationMain` + callbacks), and fighting that adds complexity.

For me, the cleanest solution remains the #3880 approach: everything on the main thread, driven by `CADisplayLink`, with `ios_ready()` / `ios_update(bool)` / `ios_destroy()` callbacks.

| Aspect | Background thread + patches | #3880 (main thread) |
|---|---|---|
| Data race | touch + window flags, but very unlikely in practice | Never existed |
| Frame sync | Semaphore + display link (IPC) | Direct display link callback |
| Shutdown | Async cleanup, fragile timing | None needed |
| Complexity | Serial queue, semaphore, display link, async — 4 parts | One display link |
| API change | None — same while-loop API | Requires changing `main()` see https://github.com/ghera/raylib-iOS/blob/master/README.md |

@raysan5, let me know if preserving desktop-style `main()` is a hard requirement and callbacks approach is totally off the table, still happy to contribute to either direction.

vsaint1:
> @vsaint1, the solutions above are just suggestions — feel free to take, adapt, or ignore anything :)


Thanks for your help with this PR, i think preserving "desktop-style" is a very good point if done correctly, the OpenGLES -> Angle is out of scope here since it can be later easily swapped to angle on Raylib codebase and all platforms benefits from this upgrade. Im taking notes and asking here/discord before pushing code to this, if you have more suggestions/advice feel free to tag me at Raylib discord or any media.

Sorry for any mistakes, english isn't my native language.

> Hi, I made a more thorough architectural analysis of the iOS platform code in `src/platforms/rcore_ios.c` and `src/rcore_ios_main.m`. Below I summarize the issues that, in my opinion, should be addressed before merge (aside from the aforementioned OpenGL ES deprecation).
> ### Issue 1: Unreferenced Background Thread + Generic Race Conditions
> 
> The game loop runs on a thread pulled from `dispatch_get_global_queue` — a **concurrent** queue shared with system frameworks and other libraries.
> 
> **What happens:**
> 
> ```objc
> dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
>     raylib_main(ios.argc, ios.argv);
> });
> ```
> 
>     * The thread has **no identity** — invisible in Xcode debugger, no name in crash logs
> 
>     * The queue cannot be referenced for synchronization (`dispatch_sync` to a concurrent queue runs inline on the caller via GCD fast-path)
> 
>     * Multiple entries in the same function or external libraries may share the same pool
> 
>     * Shared state (`CORE.Input.Touch`, `CORE.Window.flags`, etc.) is accessed from two threads (main + bg) with **no explicit synchronization** — memory ordering depends on compiler luck
> 
> 
> **Proposed fix:** Replace the global concurrent queue with a **dedicated serial queue** for the game loop.
> 
> ```objc
> // Instead of dispatch_get_global_queue:
> raylibQueue = dispatch_queue_create("com.raylib.engine", NULL);
> dispatch_async(raylibQueue, ^{
>     raylib_main(ios.argc, ios.argv);
> });
> ```
> 
> This gives the thread an identity (visible in Xcode debugger, Instruments, crash logs) and makes it referenceable for future synchronization needs.
> 
> **Consequence: shutdown dispatch must change too.** With a serial queue, `dispatch_sync(main_queue, shutdownBlock)` in `ios_shutdown_window()` can deadlock if the main thread is blocked waiting for `raylibQueue`. With the original concurrent queue this was hidden by GCD's fast-path inline optimization — with a serial queue it's deterministic.
> 
> Fix: use `dispatch_async` instead — non-blocking:
> 
> ```objc
> dispatch_async(dispatch_get_main_queue(), shutdownBlock);
> ```
> 
> `ios.shuttingDown` stays `true` to prevent re-entry until the async block executes.
> 
> **Possible race conditions (low severity in practice):**
> 
>     * `CORE.Input.Touch` — main thread writes (touchesBegan:) vs bg thread reads (PollInputEvents:), no explicit barrier. On Apple Silicon, aligned 8-byte loads/stores are atomic natively, so `Vector2` tearing should not occur.
> 
>     * `CORE.Window.flags` — `ios_set_window_focused()` from main thread vs `IsWindowFocused()` from bg thread. Focus detection may be 1 frame delayed, not a correctness issue.
> 
> 
> ### Issue 2: No Display Frame Sync (Busy-Loop)
> 
> **What happens:** `ios_present_frame()` returns immediately. No `CADisplayLink`, no semaphore, no `CFRunLoopRunInMode`. The `raylib_main` while-loop spins at maximum CPU speed.
> 
> **Impact:** CPU spikes on static scenes → battery drain. Screen tearing (buffer swap not aligned to vsync). Inconsistent framerate (device overheats, throttles, fps drops).
> 
> **Proposed fix:** Add `CADisplayLink` + `dispatch_semaphore` for frame pacing.
> 
> A `vsyncCount` is incremented each display frame. `ios_present_frame()` waits for the next vsync before returning:
> 
> ```objc
> // Display link callback (main thread run loop)
> + (void)onDisplayFrame:(CADisplayLink *)link {
>     vsyncCount++;
>     dispatch_semaphore_signal(frameSem);
> }
> 
> // In ios_present_frame (bg thread)
> void ios_present_frame(void) {
>     [ios.context presentRenderbuffer:GL_RENDERBUFFER];
> 
>     // Wait for next vsync — CPU sleeps here
>     static uint64_t lastVsync = 0;
>     while (vsyncCount - lastVsync < 1)
>         dispatch_semaphore_wait(frameSem, DISPATCH_TIME_FOREVER);
>     lastVsync = vsyncCount;
> }
> ```
> 
> @vsaint1, the solutions above are just suggestions — feel free to take, adapt, or ignore anything :)
> 
> That said, I want to be honest: these patches are improvements, but they treat symptoms. The background thread exists solely to preserve the desktop while-loop API on a platform that was not designed for it. iOS uses a run-loop-driven model (`UIApplicationMain` + callbacks), and fighting that adds complexity.
> 
> For me, the cleanest solution remains the #3880 approach: everything on the main thread, driven by `CADisplayLink`, with `ios_ready()` / `ios_update(bool)` / `ios_destroy()` callbacks.
> Aspect 	Background thread + patches 	#3880 (main thread)
> Data race 	touch + window flags, but very unlikely in practice 	Never existed
> Frame sync 	Semaphore + display link (IPC) 	Direct display link callback
> Shutdown 	Async cleanup, fragile timing 	None needed
> Complexity 	Serial queue, semaphore, display link, async — 4 parts 	One display link
> API change 	None — same while-loop API 	Requires changing `main()` see https://github.com/ghera/raylib-iOS/blob/master/README.md
> 
> @raysan5, let me know if preserving desktop-style `main()` is a hard requirement and callbacks approach is totally off the table, still happy to contribute to either direction.

raysan5:
@ghera thanks for the review and all the improvements proposed, definitely should be reviewed

 > @raysan5, let me know if preserving desktop-style main() is a hard requirement and callbacks approach is totally off the table, still happy to contribute to either direction.

Mostly. One of the key design constraints of raylib was allowing same code-base to be compiled and run on all supported platforms, **with no changes**. Actually, Android, afaik, follows the same pattern as the proposed on this PR for iOS. Said that, this same constrain was also a reason to not add iOS support to raylib and keep it as a separate library, as you did with `raylib-ios`... but currently my main concern about this addition is **maintenance**, @vsaint1 are you ready to maintain this platform backend for a couple of years? Note that all related issues/PRs will be pointed to you and users could be quite demanding, this is a key factor.

ghera:
@raysan5 understood, no more objections from my side.

Just a clarification: the Android NDK's android_native_app_glue gives you full control — it lets you handle android_main() directly on the looper thread, no framework taking over. iOS has no equivalent: UIApplicationMain owns the main thread and forces UIKit callbacks onto it, so a background worker thread is needed.

vsaint1:
> Mostly. One of the key design constraints of raylib was allowing same code-base to be compiled and run on all supported platforms, **with no changes**. Actually, Android, afaik, follows the same pattern as the proposed on this PR for iOS. Said that, this same constrain was also a reason to not add iOS support to raylib and keep it as a separate library, as you did with `raylib-ios`... but currently my main concern about this addition is **maintenance**, @vsaint1 are you ready to maintain this platform backend for a couple of years? Note that all related issues/PRs will be pointed to you and users could be quite demanding, this is a key factor.

yes no problem, also ill remove the iOS example and keep using existing example/side-projects as test cases

src/rcore_ios_main.m
```diff
@@ -0,0 +1,546 @@
+#import <Foundation/Foundation.h>
+#import <OpenGLES/EAGL.h>
+#import <OpenGLES/ES3/gl.h>
+#import <QuartzCore/CADisplayLink.h>
+#include <QuartzCore/CAFrameRateRange.h>
```
v-belyaev: Minor: was there any particular reason for including this header with `#include`?
Maybe for consistency we could import it with `#import`, like the other Apple headers?

src/platforms/rcore_ios_main.m
```diff
@@ -0,0 +1,546 @@
...
+extern void ios_request_close(void);
+extern void ios_set_window_focused(bool focused);
+
+typedef struct IOSBridgeState {
+  int argc;
+  char **argv;
+  UIWindow *window;
+  UIViewController *viewController;
+  UIView *glView;
+  EAGLContext *context;
```
v-belyaev, line 50: I suggest explicitly adding the `-fobjc-arc` compilation flag to the CMake configuration.
Maybe we could also explicitly specify ARC lifetime qualifiers for the ObjC properties?

src/rcore_ios_main.m
```diff
@@ -0,0 +1,546 @@
...
+      *scaleY = ios.scaleY;
+}
+
+void *ios_get_proc_address(const char *name) {
+  return dlsym(RTLD_DEFAULT, name);
+}
+
+void ios_open_url(const char *url) {
+  NSString *nsurlString = [NSString stringWithUTF8String:url];
+  NSURL *nsurl = [NSURL URLWithString:nsurlString];
```
v-belyaev: According to the [documentation](https://developer.apple.com/documentation/foundation/nsurl/urlwithstring:?language=objc), this method may behave differently across iOS versions.
To keep the behavior consistent, I suggest switching to URL parsing via the `NSURLComponents` API.

vsaint1: I noted with some issues that i need to refactor, thanks for the help and review.

src/rcore_ios_main.m
vsaint1: Fixed thanks

src/platforms/rcore_ios_main.m
vsaint1: Yeah, but it's already using ARC... also I can make some improvements using class.

vsaint1:
# Raylib iOS

Guide how to compile and run a raylib project on iOS using CMake and Xcode.

## Requirements

- macOS with Xcode installed
- CMake 3.25 or newer
- iOS Simulator support from Xcode, or a connected iPhone/iPad for device testing
- Minimun iOS API 13
- An Apple developer signing team if you want to run on a real device (Optional)

## CMake configuration

In your `CMakeLists.txt`, you must add the following to rename `main` to `raylib_main`, since raylib's iOS backend owns the real `main` entry point and calls your code as `raylib_main`:

```cmake
set_source_files_properties(
    ${CMAKE_CURRENT_LIST_DIR}/main.c
    PROPERTIES COMPILE_DEFINITIONS "main=raylib_main"
)
```

>  Now this is required because the code snippet below was removed from`raylib.h`, which isn't  automatically handling the `main` -> `raylib_main` renaming for iOS anymore.

```c
#ifdef PLATFORM_IOS
    #define main raylib_main
#endif
```



## Generate the Xcode project

From your project folder, run:

```sh
cmake -S . -B build -G Xcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_SYSROOT=iphonesimulator
```

This generates `build/<your_project>.xcodeproj`.

## Run in the iOS Simulator

1. Open the generated `.xcodeproj` in Xcode.
2. Select an iPhone Simulator destination.
3. Build and run your scheme.

## Run on a real iPhone or iPad

1. Regenerate the project targeting the device SDK:

```sh
cmake -S . -B build -G Xcode -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_SYSROOT=iphoneos
```

2. Open the generated Xcode project.
3. Select your connected device.
4. Set a valid signing team under **Signing & Capabilities** in Xcode if needed.
5. Build and run your scheme.

raysan5 (commented):
Already addeed a PR...

vsaint1:
> Already addeed a PR...

Sorry for the delay, I tested and its working as expected 
image

vsaint1:
@raysan5 do you want me to add some feature or test something specific on the code?

PR #5884, opened 2026-05-24
Raylib-lua version to 5.5

keyle:
minor. raylib-lua from 5.0 to 5.5 in bindings.


raysan5:
@keyle thanks for the update!

PR #5888, opened 2026-05-26
[makefile] ANDROID_ARCH check

Fipaan:
While I was trying to setup Android build pipeline, I encountered a problem, that Makefile didn't print any useful error message on unknown ANDROID_ARCH. It was using default `gcc` for building and printed many errors that weren't informative. I added simple check during `ANDROID_COMPILER_ARCH` identification, because it's probably doesn't make sense to allow arbitrary `ANDROID_ARCH`  when `TARGET_PLATFORM=PLATFORM_ANDROID`


raysan5:
@Fipaan thanks for the review, looks good

PR #5890, opened 2026-05-26
[CMake] Reduce minimum CMake version from 3.25 to 3.22

RobLoach:
There isn't anything in 3.25 that we actually need, so we can reduce the minimum requirement. 3.22 is commonly available across many package managers, while it seems like 3.25 isn't quite there.

For a list of the changes in `FetchContent`, see:
https://cmake.org/cmake/help/latest/module/FetchContent.html

cc @elite0OG


raysan5:
@RobLoach Thanks for the review!

PR #5893, opened 2026-05-27
Support jpg fileformat by default

Tech-Tac:
Re-adds support for the most used image format in the world by default out of the box.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -232,7 +232,7 @@
     #define SUPPORT_FILEFORMAT_TGA      0       // Disabled by default
 #endif
 #ifndef SUPPORT_FILEFORMAT_JPG
-    #define SUPPORT_FILEFORMAT_JPG      0       // Disabled by default
+    #define SUPPORT_FILEFORMAT_JPG      1
 #endif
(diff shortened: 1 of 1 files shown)


raysan5:
@Tech-Tac I prefer to keep current default

Tech-Tac:
Why? What is your reasoning?

raysan5:
@Tech-Tac JPG is an asset format hardly ever used in videogame industry, most raylib users do not need it

Tech-Tac:
Sure but raylib is not used only for games, it is very often used for tools and visualizations that need to display arbitrary image formats including the popular JPG

raysan5:
@Tech-Tac in those specific use cases, you can enable the flag and recompile

Tech-Tac:
In that case, how???? I'm using cmake and couldn't get it right (that's what pushed me to open this pr)

raysan5:
@Tech-Tac I don't use CMake but I think you can change it in `raylib/src/config.h` and recompile raylib or with Makefile you can pass `-DSUPPORT_FILEFORMAT_JPG=1`. You can also join raylib Discord and ask in help channels if it does work.

PR #5896, opened 2026-05-29
[rtextures] ImageDrawLine() rounding and length fix

Kaggen67:
Changed ImageDrawLine() so it draws the correct number of pixels (not one pixel short as before) and more accurate rounding when drawing the pixels.

I've tested it on Windows 11 and on a Raspberry Pi 4.


raysan5:
@Kaggen67 thanks for the review!

PR #5898, opened 2026-05-29
Fix warning about unused variables

M374LX:
This fixes warnings about unused variables when raylib is built with the rtext module disabled.


raysan5:
@M374LX thanks for the review

PR #5901, opened 2026-05-30
[rlsw] Replace `SW_TEXTURE_REPEAT_POT_FAST` with `SW_SUPPORT_NPOT_TEXTURE`

Bigfoot71:
Following the discussion in #5827, the opt-in `SW_TEXTURE_REPEAT_POT_FAST` (runtime POT check per sample) is replaced by a compile-time `SW_SUPPORT_NPOT_TEXTURE` (on by default).

When disabled, POT is assumed globally, no `isPOT` field and no extra branches, and `swTexParameteri` emits `SW_INVALID_OPERATION` if `SW_REPEAT` is set on an NPOT axis.

`SW_CLAMP` remains supported per-axis regardless.

_I also removed what seemed to be an off-topic AI-generated comment which appeared in a previous PR._

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -165,9 +165,8 @@
 #endif
 
 // Full NPOT texture support (enabled by default)
-// When disabled, POT dimensions are assumed, enabling fast bitmask wrapping
-// (x & (size-1)) for SW_REPEAT; SW_CLAMP remains supported for any dimension
-// Setting SW_REPEAT on an NPOT texture will trigger SW_INVALID_OPERATION
+// When disabled, SW_REPEAT requires POT on its axis (fast bitmask wrap)
+// SW_CLAMP remains supported for any dimension, per-axis
 #define SW_SUPPORT_NPOT_TEXTURE true
 
 //----------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/external/rlsw.h
```diff
@@ -5187,11 +5170,25 @@ void swTexParameteri(int param, int value)
         case SW_TEXTURE_WRAP_S:
         {
             if (!sw_is_texture_wrap_valid(value)) { RLSW.errCode = SW_INVALID_ENUM; return; }
+        #if !SW_SUPPORT_NPOT_TEXTURE
+            if (value == SW_REPEAT && (RLSW.boundTexture->width & RLSW.boundTexture->wMinus1) != 0)
```
raysan5, line 5174: I don't fully understand why it's not allowed to set SW_REPEAT wrap mode on a NPOT texture when SW_SUPPORT_NPOT_TEXTURE is not enabled, is it because the fast path optimization on texel/pixel fetching? 🤔

raysan5:
@Bigfoot71 thanks for the improvement! Just added a small comment about a part I don't understand...

src/external/rlsw.h
Bigfoot71: Yes, it's related to the bit masking used to wrap the coordinates, which means the axis has to be POT.

For now, this is the only optimization, so I kept the "not supported" behavior fairly loose, you can still create NPOT textures and use clamp wrapping on the NPOT axes.

More things may be possible in the future, such as computing addresses with shifts, simplifying bilinear filtering, and other more specific specializations, depending on how strict we want to be about that.

So if you'd rather simply prevent NPOT texture creation in anticipation of that, we can do that.

PR #5902, opened 2026-06-01
[rmodels] Enable GPU skinning for GL 3.3+,

JeffM2501:
This PR enables GPU skinning for OpenGL 3.3 and 4.3.
It also ensures that we only do GPU animation if the model shader supports bones. Otherwise it defaults to CPU.
This allows someone to mix and GPU and CPU skinning if needed, and gives modern desktops use of GPU skinning without needing to change the default build.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -322,7 +322,11 @@
 #endif
 #ifndef SUPPORT_GPU_SKINNING
     // GPU skinning disabled by default, some GPUs do not support more than 8 VBOs
-    #define SUPPORT_GPU_SKINNING        0
+    #if defined (GRAPHICS_API_OPENGL_33) || defined (GRAPHICS_API_OPENGL_43)
+        #define SUPPORT_GPU_SKINNING        1
+    #else
+        #define SUPPORT_GPU_SKINNING        0
+    #endif
 #endif
 
 //------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


raysan5:
@JeffM2501 Many users, specially students, probably just want to load a model and play animation, without touching any shader. If GPU-skinning is enabled by default on OpenGL 3.3/4.3 and no shader is provided, does it  fallback to CPU?

JeffM2501:
Yes, the change I made in the mesh update has it check the shader, it will now just work.

Current raylib will fail at cpu animation if you buld with the gpu flag, this change will give you the correct fallback.

raysan5:

raysan5:
@JeffM2501 I'm afraid `models_animation_blending` example does not work anymore...

JeffM2501:
Ok, I can take a look

raysan5:
@JeffM2501 I'm afraid all animation examples stop working... 😢

raysan5:
Actually it makes sense, it's not a fallback mechanism, CPU vs GPU skinning are exclusive, if CPU is selected, required buffers are loaded and updated, if GPU is selected those buffers are omitted and a valid shader is expected. I'm reverting this PR for now.

JeffM2501:
> Actually it makes sense, it's not a fallback mechanism, CPU vs GPU skinning are exclusive, if CPU is selected, required buffers are loaded and updated, if GPU is selected those buffers are omitted and a valid shader is expected. I'm reverting this PR for now.

yeah I see that now, I don't think they are conceptually mutually exclusive, just the way it's written right now is. I have some ideas that I will work on.

PR #5903, opened 2026-06-03
[rcore][RGFW] Fix `mg_gamepads_free()` iterating freed list nodes

robknopf:
The `for` loop in `mg_gamepads_free()` captured `cur->prev` before calling `mg_gamepad_release()`. If the release mutated or freed the node, the next iteration would follow a stale pointer. Changed to a `while` loop that reads `gamepads->list.cur` fresh each iteration.


raysan5:
@robknopf I'm afraid this reeview should be implemented in RGFW repo or it will be override when updating the library

ColleagueRiley:
True, and to be clear this is related to minigamepad which is an entirely separate project to RGFW

PR #5911, opened 2026-06-06
Use CORE.Input.Keyboard.exitKey instead of KEY_ESCAPE in rcore_desktop_win32 when handling window closing

aceinetx:


raysan5:
@aceinetx Good catch! Thanks!

PR #5912, opened 2026-06-06
Update ocaml bindings version

tjammer:


raysan5:
@tjammer Great! Thanks for the update!

PR #5913, opened 2026-06-08
[raudio] Support basic capture streaming to speakers

Q-Wa-Ha:
This commit introduces basic streaming from a capture source (e.g. a microphone) to the device speakers. The playback format was changed from ma_format_f32 to ma_format_s32 to support capturing audio.

Hopefully this commit can be a basis for future capture-related functionality

Note: I have tested this on Windows, macOS, and Linux.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -282,7 +282,7 @@ typedef struct tagBITMAPINFOHEADER {
 // Defines and Macros
 //----------------------------------------------------------------------------------
 #ifndef AUDIO_DEVICE_FORMAT
-    #define AUDIO_DEVICE_FORMAT    ma_format_f32    // Device output format (float-32bit)
+    #define AUDIO_DEVICE_FORMAT    ma_format_s32    // Device output format (signed-32bit)
 #endif
 #ifndef AUDIO_DEVICE_CHANNELS
     #define AUDIO_DEVICE_CHANNELS              2    // Device output channels: stereo
@@ -395,6 +395,9 @@ typedef struct AudioData {
         AudioBuffer *last;          // Pointer to last AudioBuffer in the list
         int defaultSize;            // Default audio buffer size for audio streams
     } Buffer;
+    struct {
+        bool isStreaming;
+    } Capture;
     rAudioProcessor *mixedProcessor;
 } AudioData;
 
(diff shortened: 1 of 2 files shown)


src/raudio.c
```diff
@@ -282,7 +282,7 @@ typedef struct tagBITMAPINFOHEADER {
 // Defines and Macros
 //----------------------------------------------------------------------------------
 #ifndef AUDIO_DEVICE_FORMAT
-    #define AUDIO_DEVICE_FORMAT    ma_format_f32    // Device output format (float-32bit)
+    #define AUDIO_DEVICE_FORMAT    ma_format_s32    // Device output format (signed-32bit)
```
raysan5: This seems a potentially breaking change for many raylib users, did you consider that?

Q-Wa-Ha:
I played around with the original f32 format some more and I think the issue I was originally having that caused me to change formats was with my own hardware. I was able to change the format back and leave the rest of the code alone, so hopefully there should be no more breaking changes.

Note: There does seem to be an issue where the microphone output is being panned to one side only (and this has nothing to do with formats). I can try to fix this in a future PR, unless you would prefer it done now.

src/raudio.c
Q-Wa-Ha: I was able to revert the format back to normal. Please see below for more details.

jn-jairo:
Hi @Q-Wa-Ha I like the initiative to add audio recording in raylib, I did it before in my old binding for elixir [zexray](https://github.com/jn-jairo/zexray)

I don't use this binding anymore, raylib in elixir is cool but not practical for real apps, now I use raylib directly in zig apps.

In this old binding I have the code for recording, I didn't port this code yet because right now I don't need audio recording, but I'll eventually port it to a zig library, so if @raysan5 is ok to have audio recording in raylib I can try to add it in raylib C code.

If I remember correctly my old implementation has options for streaming with callbacks, and fixed size rotating buffer that can be retrieved by the user.

I used a separate device instead of duplex because I couldn't edit the raylib code, it could be a duplex in the raylib code but I want to check the pros/cons of it, having it as duplex will activate the microfone even if the data is being discarded, separated playback/capture devices let the codes that don't need recording to work the same.

In my old implementation I added separated functions to init/close the capture device, and to manage the recording so you can reset/start/stop the recording without closing the device. I also let the user pass the sample_rate, sample_size, channels, buffer_size when initializing the device for capture, but these could be locked as defines in C if needed, or having the defines as a default.

So, what do you people think about it?

raysan5:
> I used a separate device instead of duplex because I couldn't edit the raylib code, it could be a duplex in the raylib code but I want to check the pros/cons of it, having it as duplex will activate the microfone even if the data is being discarded, separated playback/capture devices let the codes that don't need recording to work the same.

Definitely that was a concern with this PR.

 > So, what do you people think about it?

Took a quick look to your implementation and saw lots of functions exposed, I think it can be simplified but I will look at it moree carefully later, quite busy lately.

raysan5:
Closing this PR, current approach is not the desired one, considering https://github.com/raysan5/raylib/pull/5941

PR #5914, opened 2026-06-08
[rlgl] configure default shader float precision with defines

ssleert:
A similar problem has already been discussed, but with the increasing popularity of raylib, I see more and more projects using ANGLE as a backend for OpenGL.

The main reason for rejecting past PR was that it was an overkill to use highp for everything on a mobile device, but on a pc hardware we can afford it for the sake of convenience.

By default, ANGLE uses highp for VULKAN and Metal backends and there are no visible artifacts.

But there is a problem that we are getting into undocumented behavior and it would be nice to be able to expose this directly)


raysan5:
@ssleert I think most raylib users do not need it for now and it adds maintenance work; highp can be used on regular OpenGL on desktop.

PR #5916, opened 2026-06-09
[raudio] remove extra drflac_free

CrackedPixel:
a user made a bug report while using flac files. @mackron said this free should be removed, as it is handled in `drflac_close` now. this was discussed in discord: https://discord.com/channels/426912293134270465/1513529965214437616

to reproduce you just load any flac file and close the window (with flac enabled)


raysan5:
@CrackedPixel thanks for the update!

Issue #5872, opened 2026-05-16
[rcore][RGFW] Abort when using `emscripten_set_main_loop()` instead of `ASYNCIFY`


preda:
- [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [?] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [?] My code has no errors or misuse of raylib

### Issue description

I compile for the Web. I do not want to use ASYNCIFY, so I use emscripten_set_main_loop() instead. I compare Web vs. WebRGFW. Web works but WebRGFW aborts with this stack trace:

```
abort http://localhost:8000/build-WebRGFW-Debug/hypertime.js:547
    _emscripten_sleep http://localhost:8000/build-WebRGFW-Debug/hypertime.js:7502
    RGFW_pollEvents http://localhost:8000/build-WebRGFW-Debug/hypertime.wasm:61424
    RGFW_window_checkEvent http://localhost:8000/build-WebRGFW-Debug/hypertime.wasm:61858
    PollInputEvents http://localhost:8000/build-WebRGFW-Debug/hypertime.wasm:69057
    EndDrawing http://localhost:8000/build-WebRGFW-Debug/hypertime.wasm:77383
    hypertime.wasm.doOneFrame() http://localhost:8000/build-WebRGFW-Debug/hypertime.wasm:9993
    callUserCallback http://localhost:8000/build-WebRGFW-Debug/hypertime.js:6520
    runIter http://localhost:8000/build-WebRGFW-Debug/hypertime.js:6640
    MainLoop_runner http://localhost:8000/build-WebRGFW-Debug/hypertime.js:6739
```

The source code I use is:
```
if __EMSCRIPTEN__
#include <emscripten.h>
#endif

#include "raylib.h"

void doOneFrame() {
    BeginDrawing();
    EndDrawing();
}

int main() {
    InitWindow(800, 600, "Hyper");

#if __EMSCRIPTEN__
    emscripten_set_main_loop(doOneFrame, 0, true);
#else
    /*
    while (!WindowShouldClose()) {
        doOneFrame();
    }
    */
#endif

    CloseWindow();
    return 0;
}

```

The output is:
```
INFO: Initializing raylib 6.1-dev
INFO: Platform backend: WEB (RGFW) (HTML5)
INFO: Supported raylib modules:
INFO:     > rcore:..... loaded (mandatory)
[...]

PR #5919, opened 2026-06-10
[platform/RGFW] update rgfw

CrackedPixel:
Riley has made a lot of good changes and fixes to RGFW since this was last updated

This solves https://github.com/raysan5/raylib/issues/5872 with the upstream fix

### DIFF
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -76,7 +76,7 @@
 
 void ShowCursor(void);
 void CloseWindow(void);
-double get_time_seconds(void);
+double GetTimeSeconds(void);
 
 #if defined(__unix__) || defined(__linux__)
     #define _XTYPEDEF_FONT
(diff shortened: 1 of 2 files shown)


raysan5:
@CrackedPixel Wow! A quite big update! I see RGFW now uses a callbacks approach, similar to GLFW, nice.

src/platforms/rcore_desktop_rgfw.c
```diff
@@ -1886,6 +1886,25 @@ int InitPlatform(void)
         }
     #endif
 
+    // Set callbacks
+    RGFW_setEventCallback(RGFW_mouseEnter, RGFW_cb_mousenotifyfunc);
+    RGFW_setEventCallback(RGFW_mouseLeave, RGFW_cb_mousenotifyfunc);
+    // RGFW_setEventCallback(RGFW_windowClose, RGFW_cb_windowclosefunc); // do not close here. let raylib handle it
```
ColleagueRiley, line 1892: RGFW does not close the window on windowClose events. Raylib still may want to process this event, although it might not need to if it is using `RGFW_windowShouldClose`

ColleagueRiley:
> @CrackedPixel Wow! A quite big update! I see RGFW now uses a callbacks approach, similar to GLFW, nice.

RGFW has actually supported callbacks for a while. I prefer to have a flexable API, especially when it comes to handling events so the user can use RGFW in whatever way fits their project best.

PR #5920, opened 2026-06-12
[rcore] Fix ``FileCopy`` doesn't copy file when directory exists

InviseDivine:
First of all, thanks for this amazing library! I don't know how i do games if raylib doesn't exists. About bug:

Currently ``FileCopy`` doesn't copy file when directory exists because we don't do anything when directory exists, only when it doesnt exist.


raysan5:
Good catch! thanks!

PR #5921, opened 2026-06-13
[rmodels] Smarter use of GPU and CPU skinning

JeffM2501:
This PR enables both GPU and CPU skinning to be used at the same time in the same build of raylib.
It uses the state of the shader instead of the build flag at load time to know if it should use the CPU animation buffers or not.
The update functions now lazy allocate the CPU side animation buffers only when they are needed (this has a side effect of saving memory on animatable models that are never drawn animated).
The update code also checks if the shader is going to use the bones, and if so, skips the CPU side buffers.
I also split the update functions into two parts, one that does just the bone math and one that does it all. This allows the user to update just the bones if they want, and not even try to do the CPU buffers.

This all allows animation to always 'just work' with a correct fallback to CPU is GPU is not supported, and only allocating the CPU side buffers if they are needed.


JeffM2501:
This is a refactor of PR #5902 made after feedback.

src/rmodels.c
```diff
@@ -2295,11 +2310,9 @@
     return animations;
 }
 
-// Update model animation data (vertex buffers / bone matrices) for a specific pose
+// Update model animation data (bone matrices) for a specific pose
 // NOTE 1: Request frame could be fractional, using a lerp interpolation between two frames
-// NOTE 2: Updated vertex animation data is uploaded to GPU in case of CPU skinning,
-// for GPU skinning, bone matrices are uploaded to shader on DrawModelEx()
-void UpdateModelAnimation(Model model, ModelAnimation anim, float frame)
+void UpdateModelAnimationBones(Model model, ModelAnimation anim, float frame)
```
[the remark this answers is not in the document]
JeffM2501: Oh Mr. Robot, you don't know how much I'd love to pass this as a reference, but alas that is not in the cards for us :)

src/config.h
```diff
@@ -322,7 +322,7 @@
 #endif
 #ifndef SUPPORT_GPU_SKINNING
     // GPU skinning disabled by default, some GPUs do not support more than 8 VBOs
```
CrackedPixel: this comment should probably be updated, too

JeffM2501: Updated

src/config.h
```diff
@@ -328,8 +328,9 @@
     #define SUPPORT_MESH_GENERATION     1
 #endif
 #ifndef SUPPORT_GPU_SKINNING
-    // GPU skinning disabled by default, some GPUs do not support more than 8 VBOs
-    #define SUPPORT_GPU_SKINNING        0
+    // GPU skinning enabled by default for GL3.3+, 
+    //  Some GPUs do not support more than 8 VBOs so you may need to disable it if you have issues loading animated models
+    #define SUPPORT_GPU_SKINNING        1
```
raysan5, line 333: I prefer to avoid enabling this by default. CPU solution is widely supported but GPU can fail in some platforms.

src/raylib.h
```diff
@@ -1655,6 +1655,8 @@ RLAPI void SetModelMeshMaterial(Model *model, int meshId, int materialId);
 RLAPI ModelAnimation *LoadModelAnimations(const char *fileName, int *animCount);            // Load model animations from file
 RLAPI void UpdateModelAnimation(Model model, ModelAnimation anim, float frame);             // Update model animation pose (vertex buffers and bone matrices)
 RLAPI void UpdateModelAnimationEx(Model model, ModelAnimation animA, float frameA, ModelAnimation animB, float frameB, float blend); // Update model animation pose, blending two animations
+RLAPI void UpdateModelAnimationBones(Model model, ModelAnimation anim, float frame);        // Update model animation pose (bones only)
```
raysan5, line 1658: Could those functions be kept internal to rmodels, I prefer to expose the minimum required functions for users.

src/rmodels.c
```diff
@@ -2324,8 +2337,8 @@ void UpdateModelAnimation(Model model, ModelAnimation anim, float frame)
         int nextFrame = currentFrame + 1;
         float blend = frame - currentFrame;
         blend = Clamp(blend, 0.0f, 1.0f);
-        if (currentFrame >= anim.keyframeCount) currentFrame = currentFrame%anim.keyframeCount;
-        if (nextFrame >= anim.keyframeCount) nextFrame = nextFrame%anim.keyframeCount;
+        if (currentFrame >= anim.keyframeCount) currentFrame = currentFrame % anim.keyframeCount;
```
raysan5, line 2340: Please, could you review formatting to follow raylib conventions.

src/rmodels.c
```diff
@@ -2347,13 +2360,13 @@ void UpdateModelAnimation(Model model, ModelAnimation anim, float frame)
 
             // Compute runtime bone matrix from model current pose
             //-----------------------------------------------------------------------------------
-            Transform *bindPoseTransform = &model.skeleton.bindPose[boneIndex];
+            Transform* bindPoseTransform = &model.skeleton.bindPose[boneIndex];
```
raysan5, line 2363: Please, could you review formatting to follow raylib conventions.

src/rmodels.c
```diff
@@ -182,6 +182,21 @@ static void ProcessMaterialsOBJ(Material *rayMaterials, tinyobj_material_t *mate
 // Update model vertex data (positions and normals)
 static void UpdateModelAnimationVertexBuffers(Model model);
 
+// Lazy allocation of CPU animation buffers for vertex positions and normals, used for software skinning
+static void AllocateMeshCPUAnimBuffers(Mesh* mesh)
```
raysan5, line 186: As far it is only used once, I prefert to avoid an extra function, code can be used inside function.

src/config.h
JeffM2501: This does not disable CPU skinning. Would you be ok with default enable for GL 3.3+?

src/raylib.h
JeffM2501: It is very useful for people who want to do GPU skinning only. 
This was a function in 5.5, but it was removed for 6.0 for no clear reason (incorrectly in my option).
I am simply restoring it due to it's utility.

src/rmodels.c
JeffM2501: Yes I can make that change. I just didn't want to make the outer function much larger.

src/config.h
raysan5: If GPU skinning can not be enabled on target platform, does it fallback to CPU skinning? I'd like to keep the conditionals at minimum in config.h.

src/raylib.h
raysan5: I removed them because I considered most users don't need them, that was the reason to add the `models_animation_blend_custom` example, for users that need that level of control.

src/config.h
JeffM2501: I have a plan for how to do this. I will test it in a fork. 
Is there any hard info on devices that that the smaller number of vertex attributes? or is it just an ambiguous "some" ?

JeffM2501: My research on the OpenGL spec.
OpenGL 3.0 - 4.6: Requires a minimum of 16 attributes

raysan5: > Is there any hard info on devices that that the smaller number of vertex attributes?

I think that comment was probably related to OpenGL ES 2.0...

JeffM2501: So if we defaulted GPU skinning to OpenGL 3.3+ that should not be an issue.

PR #5924, opened 2026-06-17
Remove commit message from CMakeLists.txt

Wolf-SO:
Remove ambiguous footnote that would be better suited for the commit log.

This footnote describes a change that cannot be deduced by looking at the code.


raysan5:
@Wolf-SO thanks for the review!

PR #5925, opened 2026-06-19
[rlgl] Use defines in default shaders

floopfloopfloopfloopfloop:
The `#ifndef RL_DEFAULT_SHADER_ATTRIB_NAME_POSITION` et all in [rlgl.h](https://github.com/raysan5/raylib/blob/aaacda6e147031f2af0cfb6c1fd7e64d761ddb1f/src/rlgl.h#L996) imply the names can be redefined, but the [default shader sources](https://github.com/raysan5/raylib/blob/aaacda6e147031f2af0cfb6c1fd7e64d761ddb1f/src/rlgl.h#L5020) use hardcoded names, resulting in name mismatches and nothing drawing if one attempts. This PR changes the shaders to use the defines, which seems to fix the issue.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -5020,16 +5020,16 @@ static void rlLoadShaderDefault(void)
     const char *defaultVShaderCode =
 #if defined(GRAPHICS_API_OPENGL_21)
     "#version 120                       \n"
-    "attribute vec3 vertexPosition;     \n"
-    "attribute vec2 vertexTexCoord;     \n"
-    "attribute vec4 vertexColor;        \n"
+    "attribute vec3 "RL_DEFAULT_SHADER_ATTRIB_NAME_POSITION";       \n"
+    "attribute vec2 "RL_DEFAULT_SHADER_ATTRIB_NAME_TEXCOORD";       \n"
+    "attribute vec4 "RL_DEFAULT_SHADER_ATTRIB_NAME_COLOR";          \n"
     "varying vec2 fragTexCoord;         \n"
     "varying vec4 fragColor;            \n"
 #elif defined(GRAPHICS_API_OPENGL_33)
     "#version 330                       \n"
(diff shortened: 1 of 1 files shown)


floopfloopfloopfloopfloop:
image

~~what in tarnation~~ Apparently you can unwittingly use other people's emails in git commits, and github will just say it was made by them. Amazing.

raysan5:
@floopfloopfloopfloopfloop this change reduces default shaders readability and actually makes me question if shader names should be allowed to be changed, nobody requested it in the last 12 years. Most probably not merging it but probably making some changes in consequence.

floopfloopfloopfloopfloop:
12 years is a long time, I'm surprised. Readability could be easily fixed with temporarily redefining the attribute names? 

```C
#define RL_VERTEX_POSITION RL_DEFAULT_SHADER_ATTRIB_NAME_POSITION
//...
"attribute vec3 "RL_VERTEX_POSITION";     \n"
//...
#undef RL_VERTEX_POSITION
```

Being able to avoid mismatched variable name style conventions without forking is admittedly an unusual and rather aesthetic feature I was quite happy to see.

LuisR385:
I only started learning shaders recently,
 but one thing I really appreciate about raylib is how easy the default shaders are to read.

Coming from [ShaderToy](https://www.shadertoy.com/?__cf_chl_f_tk=6VgduAZuvFDx7F1j2qlLrsHIZK.7FtigAy9cK.1VsE8-1782840540-1.0.1.1-Z_tThxRnwem5Tf_Rhx2cdlEjuNBzEDCnqPwpbtqQVjI), being able to immediately see names like vertexPosition and vertexTexCoord without following additional macros makes the code easier to understand.

Raylib's readability was one of the reasons I could understand what was happening in the default shaders.

The proposed change is technically valid, but I personally value the readability of the default shaders because they also serve as learning material for newer users.

PR #5928, opened 2026-06-19
Update supported version for Jai to 6.0

ahmedqarmout2:
Update supported version for Jai to 6.0


raysan5:
@ahmedqarmout2 thanks for the update!

PR #5929, opened 2026-06-20
Added JS output format to RlParser

jiferreras:
I slightly modified the LUA export format for JS/TS bindings, and added the output file to the outputs folder.

### DIFF
--- a/tools/rlparser/output/raylib_api.js
+++ b/tools/rlparser/output/raylib_api.js
@@ -1,0 +1,8525 @@
+export default {
+  defines: [
+    {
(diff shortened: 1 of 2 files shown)


raysan5:
@jiferreras thanks, it looks good

raysan5:
@jiferreras I'm afraid I meerged this PR too quickly, I'm checking it more carefully and I can't see the point of generating a .js file when a .json should be more than enough for the same pourpose. rlparser is not intended to generate bindings but intended to generate intermediate files for bindings generation, and json can be perfectly used for js binding. I'm reverting this change.

PR #5932, opened 2026-06-22
[rcore][GLFW] Update window position in `WindowMaximizeCallback()`

### LINKED ISSUE #5931: [rcore] `GetWindowPosition()` incorrect after `MaximizeWindow()`
When calling GetWindowPosition() after MaximizeWindow(), it no longer reports the correct position, but the one from before maximizing. (Using GLFW on MacOS)

This used to work correctly in Raylib 5.5. I suspect it broke with https://github.com/raysan5/raylib/commit/890ca8d6870585f9e3499eb0e7a327af2cabaa1b, which changed GetWindowPosition to return the buffered size, but without adapting Maximize/Minimize/RestoreWindow to update that size.

rossberg:
Update window position in WindowMaximizeCallback, since maximization does not appear to trigger WindowPosCallback on some platforms, such as MacOS.

Fixes #5931.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -2012,6 +2012,8 @@ static void WindowMaximizeCallback(GLFWwindow *window, int maximized)
 {
     if (maximized) FLAG_SET(CORE.Window.flags, FLAG_WINDOW_MAXIMIZED);  // The window was maximized
     else FLAG_CLEAR(CORE.Window.flags, FLAG_WINDOW_MAXIMIZED);          // The window was restored
+    // Update current window position 
+    glfwGetWindowPos(platform.handle, &CORE.Window.position.x, &CORE.Window.position.y);
 }
 
 // GLFW3: Window focus callback, runs when window get/lose focus


raysan5:
@rossberg what about other platforms? this change can break other platforms

rossberg:
@raysan5, how can it break anything? I assume that the intended invariant of `CORE.Window.position` is that it is always in sync with the actual window position. So in the worst case, it's a redundant call on platforms where Maximize already triggers the Pos callback (such as Win32).

(The minor cost of that redundancy seems fine, given that maximising/restoring a window is highly infrequent, and slow anyway.)

raysan5:
@rossberg did you test it in other platforms? I've been in this situation too many times...

rossberg:
It seems to work on MacOS and Windows. I can't test other platforms, unfortunately.

raysan5:
@rossberg I will test it on Windows, Linux, macOS as soon as possible (it could be several months)

rossberg:
Thanks. The other possible fix for #5931 (GetWindowPosition being broken right now) would be to revert https://github.com/raysan5/raylib/commit/890ca8d6870585f9e3499eb0e7a327af2cabaa1b.

tjammer:
I can confirm that it works on linux

raysan5:
Applied the second proposed fix

PR #5933, opened 2026-06-23
[rcore][DRM] Fix Tegra compatibility (NVIDIA Jetson Orin)

danimdj22-sys:
There were part of the code that was failing when trying to use DRM platform in NVIDIA Orin.

There were 2 main failures:
- No GBM surface was being created.
- When calling the eglChooseConfig the first time to get number of available configs, the program was crashing.

And also a minor failure when the program was closing:
- Freeing the EGL libraries also freed the pointers in the GBM. (This is not throughly tested).

I've added (in the code only, not in the CMAKE) the possibility to define TEGRA_COMPATIBILITY in order to avoid the parts of the code that crashed. I've also changed the line to use eglGetConfigs instead of eglChooseConfig.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -1373,8 +1373,14 @@ int InitPlatform(void)
         return -1;
     }
 
+#if !defined(TEGRA_COMPATIBILITY)
     platform.gbmSurface = gbm_surface_create(platform.gbmDevice, platform.connector->modes[platform.modeIndex].hdisplay,
         platform.connector->modes[platform.modeIndex].vdisplay, GBM_FORMAT_ARGB8888, GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING);
+#else
+    uint64_t mod = DRM_FORMAT_MOD_LINEAR;
+    platform.gbmSurface = gbm_surface_create(platform.gbmDevice, platform.connector->modes[platform.modeIndex].hdisplay,
+        platform.connector->modes[platform.modeIndex].vdisplay, GBM_FORMAT_ARGB8888, &mod, 1);
+#endif
     if (!platform.gbmSurface)
     {
         TRACELOG(LOG_WARNING, "DISPLAY: Failed to create GBM surface");
@@ -1446,7 +1452,7 @@ int InitPlatform(void)
         return -1;
     }
 
-    if (!eglChooseConfig(platform.device, NULL, NULL, 0, &numConfigs))
+    if (!eglGetConfigs(platform.device, NULL, 0, &numConfigs))
     {
         TRACELOG(LOG_WARNING, "DISPLAY: Failed to get EGL config count: 0x%x", eglGetError());
         return -1;
@@ -1650,6 +1656,7 @@ void ClosePlatform(void)
     }
 
 #if !defined(GRAPHICS_API_OPENGL_SOFTWARE)
+#if !defined(TEGRA_COMPATIBILITY)
     if (platform.prevBO)
     {
         gbm_surface_release_buffer(platform.gbmSurface, platform.prevBO);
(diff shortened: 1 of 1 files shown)


raysan5:
@danimdj22-sys did you test your changes in other DRM platforms?

danimdj22-sys:
> @danimdj22-sys did you test your changes in other DRM platforms?

Not really, I only have access right now to the NVIDIA Orin as a DRM platform.
That's mainly the reason why I made the define.

The solution was found when looking at the kmscube repository, and why their example worked while Raylib was failing.

src/platforms/rcore_drm.c
```diff
@@ -1373,8 +1373,14 @@ int InitPlatform(void)
         return -1;
     }
 
+#if !defined(TEGRA_COMPATIBILITY)
     platform.gbmSurface = gbm_surface_create(platform.gbmDevice, platform.connector->modes[platform.modeIndex].hdisplay,
         platform.connector->modes[platform.modeIndex].vdisplay, GBM_FORMAT_ARGB8888, GBM_BO_USE_SCANOUT | GBM_BO_USE_RENDERING);
+#else
+    uint64_t mod = DRM_FORMAT_MOD_LINEAR;
+    platform.gbmSurface = gbm_surface_create(platform.gbmDevice, platform.connector->modes[platform.modeIndex].hdisplay,
+        platform.connector->modes[platform.modeIndex].vdisplay, GBM_FORMAT_ARGB8888, &mod, 1);
```
raysan5, line 1382: Is this correct? is `gbm_surface_create()` signature different for Tegra, allowing 6 function parameters instead of 5??? That seems quite weird to me... maybe you were mean to use `gbm_surface_create_with_modifiers()`?

Also that `mod` should be defined as an array of modifiers for clarity, matching that count = 1.

src/platforms/rcore_drm.c
```diff
@@ -1446,7 +1452,7 @@ int InitPlatform(void)
         return -1;
     }
 
-    if (!eglChooseConfig(platform.device, NULL, NULL, 0, &numConfigs))
+    if (!eglGetConfigs(platform.device, NULL, 0, &numConfigs))
```
raysan5, line 1455: This change can actually break DRM for several platforms... if not imply a performance hit... why using `eglGetConfigs()` instead of `eglChooseConfig()`?

src/platforms/rcore_drm.c
```diff
@@ -1650,6 +1656,7 @@ void ClosePlatform(void)
     }
 
 #if !defined(GRAPHICS_API_OPENGL_SOFTWARE)
+#if !defined(TEGRA_COMPATIBILITY)
```
raysan5, line 1659: This define is blocking the device closing, are you sure that's correct?

raysan5:
I'm afraid I'm not merging this change, it seems it is not properly tested neither understood.

danimdj22-sys:
You're right part of the changes were erased when I was cleaning up.

Sorry for wasting your time in it. I'll update the fork i made for the pull and see if I can test it elsewhere.

src/platforms/rcore_drm.c
danimdj22-sys: For some reason, the tegra platform, when using eglChooseConfig with null arguments, was crashing my program. I think it's more of an issue on their side, but i really dont know.

src/platforms/rcore_drm.c
danimdj22-sys: This also didn't make sense for me. As I mentioned previously, I think that when closing the EGL the GBM pointers were being freed.
It was also strange because it doesn't matter if the GBM were freed prior to the EGL or afterwards, it lead to a crash, either when freeing the GBM surface or when closing the EGL device. Depending on the order of execution.

As I noted in the first comment, it's not throughly tested why this happens.
When I set the environment variable for GBM_BACKEND=dri (using softpipe instead of HW acceleration). The crash wasn't there.

src/platforms/rcore_drm.c
danimdj22-sys: Yes, the function name was a mistake I introduced in the clean up prior to making the pull request.

I added a lot of logs in order to find where it was crashing.

PR #5934, opened 2026-06-23
[rmodels] Add comments for GenMeshTorus

EugenioBarbieriViale:
I found it difficult to get the different radii of the torus with the information given by the documentation and the code. To make it easier to understand for dummies like my, I added some comments. Full reference [here](https://www.reddit.com/r/raylib/comments/1uc14ze/help_understanding_genmeshtorus/).


raysan5:
@EugenioBarbieriViale thanks for the update

PR #5935, opened 2026-06-24
[rlgl] Fix matrix stack overflow in rlPushMatrix()

szarta:
### Problem

`rlPushMatrix()` checks for stack overflow but does not stop on it:

```c
void rlPushMatrix(void)
{
    if (RLGL.State.stackCounter >= RL_MAX_MATRIX_STACK_SIZE) TRACELOG(RL_LOG_ERROR, "RLGL: Matrix stack overflow (RL_MAX_MATRIX_STACK_SIZE)");
    ...
    RLGL.State.stack[RLGL.State.stackCounter] = *RLGL.State.currentMatrix;
    RLGL.State.stackCounter++;
}
```

The check logs an error but has no `return`, so the assignment still runs when the stack is full. With `RLGL.State.stackCounter == RL_MAX_MATRIX_STACK_SIZE`, this writes `RLGL.State.stack[RL_MAX_MATRIX_STACK_SIZE]` — one element past `Matrix stack[RL_MAX_MATRIX_STACK_SIZE]` — corrupting the adjacent `RLGL.State` members (`stackCounter`, `currentTextureId`, …), and it grows with each further unbalanced push.

`rlPopMatrix()` already guards the symmetric underflow case (`if (RLGL.State.stackCounter > 0)`), so this is just a missing early return on the push side.

### Fix

```c
if (RLGL.State.stackCounter >= RL_MAX_MATRIX_STACK_SIZE)
{
    TRACELOG(RL_LOG_ERROR, "RLGL: Matrix stack overflow (RL_MAX_MATRIX_STACK_SIZE)");
    return;
}
```

### Reproduction

A small program mirroring the `RLGL.State` field order (`Matrix stack[RL_MAX_MATRIX_STACK_SIZE]; int stackCounter; unsigned int currentTextureId; …`) and doing 33 unbalanced `rlPushMatrix()` calls:

```
after 32 pushes: stackCounter=32, currentTextureId=0xABCD1234 (intact)
RLGL: Matrix stack overflow            <- guard fires but does NOT prevent the write
after 33rd push: stackCounter=1088421889, currentTextureId=0x40E00000   <- both clobbered
```

`0x40E00000` is the bit pattern of `7.0f`, the matrix payload — i.e. the `Matrix` bytes spilled into the neighbouring `State` members. After the fix the 33rd push is rejected and the neighbouring members stay intact. (It's an intra-object overflow, so default ASan/Valgrind do not flag it; the repro shows the member corruption directly.)

Verified to build cleanly (`make PLATFORM=PLATFORM_DESKTOP`).

---
*Disclosure: this issue was identified by a static-analysis audit and the fix was prepared with AI assistance (Anthropic Claude), then reviewed and submitted by @szarta.*

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -1237,7 +1237,11 @@ void rlMatrixMode(int mode)
 // Push the current matrix into RLGL.State.stack
 void rlPushMatrix(void)
 {
-    if (RLGL.State.stackCounter >= RL_MAX_MATRIX_STACK_SIZE) TRACELOG(RL_LOG_ERROR, "RLGL: Matrix stack overflow (RL_MAX_MATRIX_STACK_SIZE)");
+    if (RLGL.State.stackCounter >= RL_MAX_MATRIX_STACK_SIZE)
+    {
+        TRACELOG(RL_LOG_ERROR, "RLGL: Matrix stack overflow (RL_MAX_MATRIX_STACK_SIZE)");
+        return;
+    }
 
     if (RLGL.State.currentMatrixMode == RL_MODELVIEW)
     {


raysan5:
@szarta thanks for the improvement but I have a couple of concerns: What happens if the limit is reached and it actually returns with no action, what are the implications for the user and results on the render?

Also note that raylib tries to minimize the early-returns as much as possible, maybe that code is refactored in a future.

szarta:
To answer your render implications question: when the guard fires, the behavior is worse than a simple no-op.

The early return exits before the `RL_MODELVIEW` branch that sets `currentMatrix = &transform`, so that pointer isn't redirected. Subsequent transforms (translate, rotate, scale) in the caller's scope write directly into whatever `currentMatrix` already pointed to — the parent scope's working matrix — with no saved copy to isolate them. The matched `rlPopMatrix` still executes (its `stackCounter > 0` guard passes since the counter stays at 32), so it restores `stack[31]` and decrements the counter, consuming one legitimate saved frame. Net effect: the transforms between the overflow push and its matched pop leak into the parent scope, and the parent loses its saved matrix. Repeated overflows cascade — each one drains one real frame from the top of the stack.

`RL_LOG_ERROR` severity reflects this accurately: the render output will be wrong, not just degraded.

Noted on the early-return style. Happy to refactor to the single-return pattern used by the other gamepad functions if you'd like a follow-up PR, or leave it for a future cleanup pass — whichever you prefer.

raysan5:
@szarta Are you copy pasting your AI answers? Because I would really appreciate a minimum of humanity in those answers, like I do, spending my time answering to you. Sincerely, I didn't understand that block of text you pasted, kind of disgusting for me... If you expect to be sending patches and answering this way, better don't do it again. Thanks.

szarta:
Ha, yes, I am a human.  I've been writing a CERT-C static code analysis tool and needed to find good codebases as benchmarks and needed a good C99 reference and this codebase came up - I tried to keep it relevant to issues that were both high enough priority/impact and provide as much due diligence as possible myself before logging it - I didn't want to either waste primary contributor time with non-issues and wanted to provide clear direct evidence of problems with their solutions.  Since finding/reporting issues is a secondary objective to building my tool, I try to keep it as simple/straightfoward as I can, sorry if I come off sounding like a robot as a result. :-D

raysan5:
@szarta thanks for your answer

PR #5937, opened 2026-06-24
[rcore] Bounds-check gamepad index in GetGamepadAxisCount() and GetGamepadName()

szarta:
### Problem

`GetGamepadAxisCount()` and `GetGamepadName()` index the `CORE.Input.Gamepad` arrays with an unvalidated `gamepad` argument:

```c
const char *GetGamepadName(int gamepad)  { return CORE.Input.Gamepad.name[gamepad]; }
int GetGamepadAxisCount(int gamepad)     { return CORE.Input.Gamepad.axisCount[gamepad]; }
```

`axisCount` and `name` are sized `[MAX_GAMEPADS]`, so `gamepad < 0` or `gamepad >= MAX_GAMEPADS` is an out-of-bounds read. Every other gamepad accessor (`IsGamepadAvailable`, `IsGamepadButtonPressed/Down/Released/Up`, `GetGamepadAxisMovement`) already guards with `gamepad < MAX_GAMEPADS`; these two getters are the only ones missing it.

### Fix

```c
const char *GetGamepadName(int gamepad)
{
    if ((gamepad < 0) || (gamepad >= MAX_GAMEPADS)) return NULL;
    return CORE.Input.Gamepad.name[gamepad];
}

int GetGamepadAxisCount(int gamepad)
{
    if ((gamepad < 0) || (gamepad >= MAX_GAMEPADS)) return 0;
    return CORE.Input.Gamepad.axisCount[gamepad];
}
```

(The siblings check only `gamepad < MAX_GAMEPADS`; I added the lower bound too — happy to match the sibling style exactly if you prefer.)

### Reproduction

AddressSanitizer on `GetGamepadAxisCount(MAX_GAMEPADS)`:

```
==ERROR: AddressSanitizer: global-buffer-overflow ... READ of size 4
  #0 GetGamepadAxisCount
  ... 0 bytes to the right of global variable 'axisCount' ... of size 16
```

After the fix, out-of-range indices return a safe default (`0` / `NULL`); valid indices are unchanged. Verified to build cleanly (`make PLATFORM=PLATFORM_DESKTOP`).

---
*Disclosure: this issue was identified by a static-analysis audit and the fix was prepared with AI assistance (Anthropic Claude), then reviewed and submitted by @szarta.*

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -3934,9 +3934,11 @@ bool IsGamepadAvailable(int gamepad)
 // Get gamepad internal name id
 const char *GetGamepadName(int gamepad)
 {
-    if ((gamepad < 0) || (gamepad >= MAX_GAMEPADS)) return NULL;
+    const char *name = NULL;
 
-    return CORE.Input.Gamepad.name[gamepad];
+    if ((gamepad >= 0) && (gamepad < MAX_GAMEPADS)) name = CORE.Input.Gamepad.name[gamepad];
+
+    return name;
 }
 
 // Check if gamepad button has been pressed once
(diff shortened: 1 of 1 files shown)


raysan5:
@szarta Please, can you refactor code to avoid multiple returns and follow similar pattern to the other gamepad functions?

szarta:
Refactored both functions to the single-return pattern used by the other gamepad functions — result/name variable initialized to the safe default, conditionally set inside the bounds check, single return at the end. Updated in bd7d910.

raysan5:
@szarta thanks for the review and improvements!

PR #5938, opened 2026-06-26
[rcore] Add gamepad >= 0 lower-bound guard to six sibling functions (#5937 follow-up)

mahirhir:
## Problem

Commit 3edfe194 (#5937) added a `gamepad >= 0` lower-bound check to `GetGamepadAxisCount()` and `GetGamepadName()`, but the commit message incorrectly stated that sibling functions *already* guarded the index. Six siblings were left with only the upper-bound check (`gamepad < MAX_GAMEPADS`).

A negative signed `int` passes that upper check and causes out-of-bounds array access — undefined behaviour in C:

```c
// gamepad = -1 passes this check (-(1) < MAX_GAMEPADS is true)
if ((gamepad < MAX_GAMEPADS) && CORE.Input.Gamepad.ready[gamepad] ...)
//                                                             ^^^^ OOB read
```

Affected functions:
- `IsGamepadAvailable`
- `IsGamepadButtonPressed`
- `IsGamepadButtonDown`
- `IsGamepadButtonReleased`
- `IsGamepadButtonUp`
- `GetGamepadAxisMovement`

## Fix

Prepend `(gamepad >= 0) &&` to the bounds condition in each of the six functions, exactly matching the pattern already applied in 3edfe194:

```c
if ((gamepad >= 0) && (gamepad < MAX_GAMEPADS) && CORE.Input.Gamepad.ready[gamepad] ...)
```

The change is 6 lines, one per function, no logic change beyond closing the bounds gap.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -3926,7 +3926,7 @@ bool IsGamepadAvailable(int gamepad)
 {
     bool result = false;
 
-    if ((gamepad < MAX_GAMEPADS) && CORE.Input.Gamepad.ready[gamepad]) result = true;
+    if ((gamepad >= 0) && (gamepad < MAX_GAMEPADS) && CORE.Input.Gamepad.ready[gamepad]) result = true;
 
     return result;
 }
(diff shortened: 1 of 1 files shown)


raysan5:
@greymoth-jp thanks for the review

PR #5943, opened 2026-06-28
[tools] Add S-Expression Output Type to `rlparser`

VisenDev:
[S-Expressions are a simple file format that has been around since the 1950s.](https://en.wikipedia.org/wiki/S-expression) They consist of nested lists surrounded by parenthesis. The source code of LISP family languages like Common Lisp, Scheme, Clojure, Racket, or Emacs Lisp are represented using S-Expressions. 

They are a good candidate for a raylib api output format because they are extremely simple to write a parser for, and all lisp family languages have built in functionality for parsing and manipulating S-Expressions, so they are a good choice for a data exchange format.

You could also change the file extensions and flags to `LISP` instead of `SEXPR`, but I wanted to convey that the output is just data, not lisp code.

I chose to only use the most basic of S-Expression syntax so that the output would be mutually parsable by different lisps. (ie no `(dotted . lists)` no common lisp `:keywords`)

Screenshot_20260628_143713


raysan5:
@VisenDev Thanks! It looks like a useful addition

Issue #5946, opened 2026-07-01
[rcore] Outdated joystick API's for SDL3 backend


0xFishbyte:
**WARNING: Please, read this note carefully before submitting a new issue:**

It is important to realise that **this is NOT A SUPPORT FORUM**, this is for reproducible BUGS with raylib ONLY.

There are lots of generous and helpful people ready to help you out on [raylib Discord forum](https://discord.gg/raylib) or [raylib reddit](https://www.reddit.com/r/raylib/).

Remember that asking for support questions here actively takes developer time away from improving raylib.

---

Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

**RCORE** is using old SDL2 API's for joystick initialization stuff even when building for SDL3. This already caused issues with gamepad detection at least on my machine:

```
WARNING: PLATFORM: Unable to open game controller [ERROR: Parameter 'joystick' is invalid]
INFO: PLATFORM: DESKTOP (SDL3): Initialized successfully
```
The culprit is:
https://github.com/raysan5/raylib/blob/a7ac5c2823d767002ca644b24708353c628cad08/src/platforms/rcore_desktop_sdl.c#L2167-L2172

But according to [SDL migration guide](https://wiki.libsdl.org/SDL3/README-migration):

```
SDL_JoystickInstanceID() => SDL_GetJoystickID()
```
```
SDL_GameControllerGetJoystick() => SDL_GetGamepadJoystick()
```
and
```
SDL_GameControllerOpen() => SDL_OpenGamepad()
```

### Environment

*Provide your Platform, Operating System, OpenGL version, GPU details where you experienced the issue.*
SDL3, Linux (CachyOS), OpenGL 4.6, GPU is irrelevant i guess but controller was [Thrustmaster T.FLIGHT STICK X](https://shop.thrustmaster.com/nl_nl/t-flight-stick-x.html) 

**Would you like a PR to fix this?**

0xFishbyte:
Also most of `rcore_desktop_sdl.c` is using old API. Even `PlatformData` looks like this:
https://github.com/raysan5/raylib/blob/a7ac5c2823d767002ca644b24708353c628cad08/src/platforms/rcore_desktop_sdl.c#L98-L112

[...]

PR #5948, opened 2026-07-02
[rcore] Fix SDL3 backend using wrong joystick functions

0xFishbyte:
Fixes what was discussed in #5946

This works well on both SDL2 and SDL3 backends. Works for both hot-plugging and starting raylib with joystick already attached. But keep in my mind that's on my machine just with one controller.

Thanks.


raysan5:
@0xFishbyte thanks for the review, it looks good. As a side note, it seems those functions were added in SDL 3.2.0.

PR #5951, opened 2026-07-03
[build][Zig] 0.17 migration 

miker2049:
Zig 0.17 is not out yet, but it (supposedly) will be soon.  There was recently some [breaking changes to the build system](https://codeberg.org/ziglang/zig/pulls/35428), causing nightly zig to not work with raylib's `build.zig`.  

Because the aforementioned changes are not in a major release yet, this probably should not be merged yet.  But it doesn't seem like there will be many further breaking changes before the major release, so this can be considered a tentative PR for when it happens.  One further consideration here is that this branch uses an [unmerged branch of zemscripten](https://github.com/zig-gamedev/zemscripten/pull/23) which updates it for 0.17.  So, hopefully that can be merged before this should be!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -69,7 +69,7 @@ pub const emsdk = struct {
         raylib.root_module.addIncludePath(emsdk_dep.path("upstream/emscripten/cache/sysroot/include"));
         wasm.root_module.addIncludePath(emsdk_dep.path("upstream/emscripten/cache/sysroot/include"));
 
-        const emcc_step = zemscripten.emccStep(b, wasm, options);
+        const emcc_step = zemscripten.emccStep(b, &.{}, &.{wasm}, options);
         emcc_step.dependOn(activate_emsdk_step);
 
         return emcc_step;
@@ -77,7 +77,7 @@ pub const emsdk = struct {
 
     pub fn emrunStep(
         b: *std.Build,
-        html_path: []const u8,
+        html_path: std.Build.LazyPath,
         extra_args: []const []const u8,
     ) *std.Build.Step {
         return zemscripten.emrunStep(b, html_path, extra_args);
@@ -718,13 +718,16 @@ fn addExamples(
                 .preload_paths = emcc_examples_preloads_map.get(filename) orelse &.{},
                 .shell_file_path = b.path("src/shell.html"),
                 .install_dir = install_dir,
+                .out_file_name = wasm.name,
             });
             b.getInstallStep().dependOn(emcc_step);
 
-            const html_filename = try std.fmt.allocPrint(b.allocator, "{s}.html", .{wasm.name});
+            const install_file = b.addInstallFile(.{ .cwd_relative = b.fmt("{s}/{s}/{s}.html", .{ module, filename, wasm.name }) }, "web");
+            b.getInstallStep().dependOn(&install_file.step);
+
             const emrun_step = emsdk.emrunStep(
                 b,
-                b.getInstallPath(install_dir, html_filename),
+                install_file.source,
                 &.{},
             );
 
--- a/build.zig
+++ b/build.zig
@@ -92,7 +92,7 @@ pub fn linkWindows(mod: *std.Build.Module, opengl: bool, comptime shcore: bool) 
 }
 
 fn findWaylandScanner(b: *std.Build) void {
-    _ = b.findProgram(&.{"wayland-scanner"}, &.{}) catch {
+    _ = b.findProgram(.{ .names = &.{"wayland-scanner"} }) orelse {
         std.log.err(
             \\ `wayland-scanner` may not be installed on the system.
             \\ You can switch to X11 in your `build.zig` by changing `Options.linux_display_backend`
@@ -648,7 +648,7 @@ fn addExamples(
     const all = b.step(module, "All " ++ module ++ " examples");
     const module_subpath = b.pathJoin(&.{ "examples", module });
 
-    var dir = try b.build_root.handle.openDir(b.graph.io, module_subpath, .{ .iterate = true });
+    var dir = try b.root.openDir(b.graph.io, module_subpath, .{ .iterate = true });
     defer dir.close(b.graph.io);
 
     var iter = dir.iterate();
@@ -706,9 +706,10 @@ fn addExamples(
             const emcc_flags = emsdk.emccDefaultFlags(b.allocator, .{ .optimize = optimize });
             const emcc_settings = emsdk.emccDefaultSettings(b.allocator, .{ .optimize = optimize });
 
-            const EmccExamplesPreloadMap = std.static_string_map.StaticStringMap([]const emsdk.zemscripten.EmccFilePath);
-            const EmccExamplesPreloadKV = struct { []const u8, []const emsdk.zemscripten.EmccFilePath };
-            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @import("examples/example_resources.zon");
+            const EmccExamplesPreloadMap = std.static_string_map.StaticStringMap([]const emsdk.zemscripten.ResourceFile);
+            const EmccExamplesPreloadKV = struct { []const u8, []const emsdk.zemscripten.ResourceFile };
+            const emcc_examples_preloads_ = @import("examples/example_resources.zon");
+            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @ptrCast(@alignCast(&emcc_examples_preloads_));
             const emcc_examples_preloads_map = EmccExamplesPreloadMap.initComptime(emcc_examples_preloads);
 
             const emcc_step = emsdk.emccStep(b, raylib, wasm, .{
@@ -762,7 +763,7 @@ fn waylandGenerate(
     comptime waylandDir: []const u8,
     comptime source: bool,
 ) !void {
-    const dir = try b.build_root.handle.openDir(b.graph.io, waylandDir, .{ .iterate = true });
+    const dir = try b.root.openDir(b.graph.io, waylandDir, .{ .iterate = true });
     defer dir.close(b.graph.io);
 
     var iter = dir.iterate();
--- a/build.zig
+++ b/build.zig
@@ -63,14 +63,11 @@ pub const emsdk = struct {
     }
 
     pub fn emccStep(b: *std.Build, raylib: *std.Build.Step.Compile, wasm: *std.Build.Step.Compile, options: zemscripten.StepOptions) *std.Build.Step {
-        const activate_emsdk_step = zemscripten.activateEmsdkStep(b);
-
         const emsdk_dep = b.dependency("emsdk", .{});
         raylib.root_module.addIncludePath(emsdk_dep.path("upstream/emscripten/cache/sysroot/include"));
         wasm.root_module.addIncludePath(emsdk_dep.path("upstream/emscripten/cache/sysroot/include"));
 
         const emcc_step = zemscripten.emccStep(b, &.{}, &.{wasm}, options);
-        emcc_step.dependOn(activate_emsdk_step);
 
         return emcc_step;
     }
@@ -797,4 +794,3 @@ fn waylandGenerate(
         }
     }
 }
-
--- a/build.zig
+++ b/build.zig
@@ -695,7 +695,7 @@ fn addExamples(
             exe_mod.addCMacro("PLATFORM_WEB", "");
 
             const wasm = b.addLibrary(.{
-                .name = filename,
+                .name = b.fmt("{s}.html", .{filename}),
                 .root_module = exe_mod,
             });
 
--- a/build.zig
+++ b/build.zig
@@ -721,7 +721,7 @@ fn addExamples(
 
             const emrun_step = emsdk.emrunStep(
                 b,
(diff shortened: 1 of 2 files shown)


raysan5:
@michaelfiber, @Not-Nik, @HaxSam, @shreyassanthu77, @pamarcos, @jn-jairo, @jirizverina, @BrunoDG, @maiconpintoabreu, @beppvis Please, could you take a look to this PR?

_NOTE: I'm pinging all of you in my list of `Zig build system` maintainers, if any of you doesn't want to be pinged for this kind of issues, just let me know and I'll take you out of the maintainers list._

build.zig.zon
```diff
@@ -21,8 +21,8 @@
             .hash = "N-V-__8AAJl1DwBezhYo_VE6f53mPVm00R-Fk28NPW7P14EQ",
```
Not-Nik, line 21: I'm seeing a hash mismatch for this package with the patch on `0.17.0-dev.1158+1d1193aa7`, where it fetches `N-V-__8AALBDi159DH5nlv03KX8rG4GFfY8m-GWnFC9DSQL2`. Either I'm being MITMed or the hash actually changed (through Zig or emsdk itself). Could someone verify (on a clean build/system).

Not-Nik (approved):
Besides the hash mismatch this compiles cleanly. Since I believe our policy is to keep up with master when required, r+

build.zig
```diff
@@ -718,13 +716,13 @@ fn addExamples(
...
                 .install_dir = install_dir,
+                .out_file_name = wasm.name,
             });
             b.getInstallStep().dependOn(emcc_step);
 
-            const html_filename = try std.fmt.allocPrint(b.allocator, "{s}.html", .{wasm.name});
             const emrun_step = emsdk.emrunStep(
                 b,
-                b.getInstallPath(install_dir, html_filename),
+                b.graph.path(.install_prefix, b.fmt("{s}/{s}.html", .{ filename, wasm.name })),
```
jn-jairo: Wrong path, it should be the one from install_dir `web/<module>/<filename>` so `install_dir.custom` instead of `filename` should work.

build.zig
```diff
@@ -718,13 +716,13 @@ fn addExamples(
                 .preload_paths = emcc_examples_preloads_map.get(filename) orelse &.{},
                 .shell_file_path = b.path("src/shell.html"),
                 .install_dir = install_dir,
+                .out_file_name = wasm.name,
             });
             b.getInstallStep().dependOn(emcc_step);
```
jn-jairo: Not what you changed, but this line should be removed, it forces to run the `emcc_step` for every example even when we are not building the examples, the `emcc_step` will run anyway when building one or more examples because of the `run_step` and `all` step.

jn-jairo (changes requested):
The new `emcc` is not building the files.

This is with the new code

Image

And this is with the old code from raylib master

Image

As the note about the install path says, this code is using the wrong path to run, trying to get it from the `zig-out/<filename>` instead of the `zig-out/web/<module>/<filename>`.

An old version of the `build.zig` was duplicating the installed files at the root `zig-out/<filename>` but it was already fixed, you must have a dirty install with old files already built, so you are not running from files built with the new code.

The only error I get is

```sh
textures_bunnymark
└─ run dependency
   └─ install generated/
      └─ run dependency (textures_bunnymark) w
shared:INFO: (Emscripten: Running sanity checks)
failed command: /mnt/hd/projects/code/raylib-pr/zig-pkg/N-V-__8AAJl1DwBezhYo_VE6f53mPVm00R-Fk28NPW7P14EQ/upstream/emscripten/emcc.py -O3 -sASYNCIFY -sTOTAL_MEMORY=134217728 -sFORCE_FILESYSTEM=1 -sMALLOC=emmalloc -sUSE_GLFW=3 -sEXPORTED_RUNTIME_METHODS=ccall ./.zig-cache/o/d879f429e89efb52350f0b58e42915cc/libtextures_bunnymark.a ./.zig-cache/o/667685bf2c596541996f084e5ef321b6/libraylib.a -o ./.zig-cache/o/a995d324b9810090a6b25a0ae46d4312/textures_bunnymark --shell-file /mnt/hd/projects/code/raylib-pr/src/shell.html
```

I am using `Kubuntu 24.04.4 LTS` and besides the web build, the desktop build is working on my machine.

Tested with the most recent zig version available at the moment `0.17.0-dev.1417+20befa4e6` as well the minimum version set in this PR `0.17.0-dev.1158+1d1193aa7`.

build.zig.zon
jn-jairo: I ran it on a clean build, cleaned all caches, with both the most recent zig version available at the moment `0.17.0-dev.1417+20befa4e6` as well the minimum version set in this PR `0.17.0-dev.1158+1d1193aa7` and I did not get any hash error, the hash downloaded is the `N-V-__8AAJl1DwBezhYo_VE6f53mPVm00R-Fk28NPW7P14EQ`.

Not-Nik: Cleaned up all my cached files as well, and the mismatch is gone. Strange it occurred at all though..

build.zig
miker2049: Fixed in 23e73da

build.zig
miker2049: [f1dc836](https://github.com/raysan5/raylib/pull/5951/commits/f1dc83603c5fd7e973a7550154b6122a1f35d53c)

miker2049:
I realized that resource asset files were not working for the emscripten examples.  It is blocked by [this issue](https://github.com/zig-gamedev/zemscripten/issues/24) with upstream zemscripten, where the build file there does not actually support virtual paths.

jn-jairo:
> I realized that resource asset files were not working for the emscripten examples. It is blocked by [this issue](https://github.com/zig-gamedev/zemscripten/issues/24) with upstream zemscripten, where the build file there does not actually support virtual paths.

Ok, let's see what we get here, `zig-gamedev/zemscripten` didn't merge the zig 0.17 PR, so you changed the repo here to point to the PR branch at `Yinameah/zemscripten` and also there is a bug with `preload_paths` not using the `ResourceFile.get()` they added.

And as you said it is temporary while they don't merge it.

> One further consideration here is that this branch uses an [unmerged branch of zemscripten](https://github.com/zig-gamedev/zemscripten/pull/23) which updates it for 0.17. So, hopefully that can be merged before this should be!

So, we are waiting for the `zig-gamedev` team to do something about the zig upgrade and the `preload_paths` bug.

The last response from `zig-gamedev` team was a month ago at the zig 0.17 issue where they said:

> Thanks for reaching out. It's difficult to stay on top of maintenance churn of all the zig-gamedev libraries so any help is very welcome. If you want to discuss anything in a more informal forum please feel free to join us on Discord https://discord.gg/JrYcW5sUD

And by looking at other repos from `zig-gamedev` they don't upgrade zig to a new version until way later. Like the [zaudio upgrade to zig 0.16](https://github.com/zig-gamedev/zaudio/pull/18) that got merged just last week.

So, will we wait for `zig-gamedev` to fix things up or what?

build.zig
```diff
@@ -698,17 +695,18 @@ fn addExamples(
...
             const emcc_flags = emsdk.emccDefaultFlags(b.allocator, .{ .optimize = optimize });
             const emcc_settings = emsdk.emccDefaultSettings(b.allocator, .{ .optimize = optimize });
 
-            const EmccExamplesPreloadMap = std.static_string_map.StaticStringMap([]const emsdk.zemscripten.EmccFilePath);
-            const EmccExamplesPreloadKV = struct { []const u8, []const emsdk.zemscripten.EmccFilePath };
-            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @import("examples/example_resources.zon");
+            const EmccExamplesPreloadMap = std.static_string_map.StaticStringMap([]const emsdk.zemscripten.ResourceFile);
+            const EmccExamplesPreloadKV = struct { []const u8, []const emsdk.zemscripten.ResourceFile };
+            const emcc_examples_preloads_ = @import("examples/example_resources.zon");
+            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @ptrCast(@alignCast(&emcc_examples_preloads_));
```
jn-jairo, line 709: Besides the `preload_paths` bug on `zemscripten` this part is not working, the new `src_path` is a `std.Build.LazyPath`, this cast is not filling it up. Here is what I did:

```zig
            const ResourceDesc = struct {
                src_path: []const u8,
                virtual_path: []const u8,
            };

            const EmccExamplesPreloadMap = std.StaticStringMap([]const ResourceDesc);
            const EmccExamplesPreloadKV = struct { []const u8, []const ResourceDesc };
            const emcc_examples_preloads: []const EmccExamplesPreloadKV = @import("examples/example_resources.zon");
            const emcc_examples_preloads_map = EmccExamplesPreloadMap.initComptime(emcc_examples_preloads);

            const preload_paths_: []const ResourceDesc = emcc_examples_preloads_map.get(filename) orelse &.{};

            var preload_paths = std.ArrayList(emsdk.zemscripten.ResourceFile).empty;
            defer preload_paths.deinit(b.allocator);

            for (preload_paths_) |preload_path| {
                try preload_paths.append(b.allocator, .{
                    .src_path = b.path(preload_path.src_path),
                    .virtual_path = preload_path.virtual_path,
                });
            }

            const emcc_step = emsdk.emccStep(b, raylib, wasm, .{
                .optimize = optimize,
                .flags = emcc_flags,
                .settings = emcc_settings,
                .preload_paths = preload_paths.items,
                .shell_file_path = b.path("src/shell.html"),
                .install_dir = install_dir,
                .out_file_name = wasm.name,
            });
```

> also there is a bug with `preload_paths` not using the `ResourceFile.get()` they added.

I tried to fix it using your proposed solution `emcc.addArg(path.get(b))` but turns out the whole `ResourceFile.get()` is wrong, as it was never used looks like zig never bothered to validate this function, so the `self.src_path.path(b, "")` it is using just returns another `LazyPath` instead of a string, but the function needs a string. This is from a commit after the one locked in the master raylib so that is why it works now in raylib, after they changed to that `ResourceFile` things didn't work anymore, but they didn't realize 🙄.

This hole is deeper than we thought.

jn-jairo (changes requested):
From here it should be enough, but the zemscripten is bugged anyway, maybe we will need to change something more here later.

jn-jairo:
> I realized that resource asset files were not working for the emscripten examples. It is blocked by [this issue](https://github.com/zig-gamedev/zemscripten/issues/24) with upstream zemscripten, where the build file there does not actually support virtual paths.

I fixed it in my [fork](https://github.com/jn-jairo/zemscripten/commit/1ed408542ad464f0347ff7405dd14b07074e9ce9) try it with the other changes I proposed in the review.

maiconpintoabreu:
My 2 cents here, I do not agree to keep it updated with zig master as we never had a zig release without breaking changes anyway.

raysan5:
As mentioned, we should probably wait until a more stable state bfore considering merging it...

jn-jairo:
For my personal projects I rather use the zig stable version, but I like to use raylib master to keep updated with the fixes and new features, so I vote to wait until zig 0.17 is released and the dependencies are upgraded to it, but it's good to have the PR ready for it.

miker2049:
Yes to be clear I am happy to turn this into a draft @raysan5 , this was just for me but thought it might be helpful generally _eventually_.  And fwiw, 0.17 is coming in not more than a few months from what I can tell by them.  

In that, yes @jn-jairo I changed to the zemscripten fork attached to the zemscripten 0.17 PR, presumably hoping that it could be changed back when this is merged.   But yeah not sure about zemscripten maintenance at the moment I guess!

One modest proposal here: the zemscripten library is not large, and raylib is not using much of it. It seems like one could just take the necessary types/functions in the zemscripten `build.zig` and have them directly in this repo without adding too many LoC.  It could include the fix with preload paths.  The majority of the actual emscripten runtime stuff that zemscripten is worried about exposing to zig is not used/will not be used by raylib anyway.  Unless, eg, the raylib examples one day want to use `emscripten_set_main_loop`.  

I am happy to take a shot in removing this dependency if there is interest here (in this same PR or elsewhere).

jn-jairo:
> One modest proposal here: the zemscripten library is not large, and raylib is not using much of it. It seems like one could just take the necessary types/functions in the zemscripten `build.zig` and have them directly in this repo without adding too many LoC.

Good idea, I didn't realize zemscripten was that simple until now, some of it we already extend here like the flags and settings. It's mostly the activation and emcc steps that are missing.

We don't use any of the `zemscripten.zig` as it is mostly extern functions and zig already has these at `std.os.emscripten`.

> Unless, eg, the raylib examples one day want to use `emscripten_set_main_loop`.

It uses at `projects/Zig/src/core_basic_window.zig` but it uses the one from zig std, by the way, we need to also update the `projectc/Zig`, it must be just a few lines at the `build.zig`.

> I am happy to take a shot in removing this dependency if there is interest here (in this same PR or elsewhere).

It can be in this PR, so, few more lines in the `emsdk` and we can remove the dependency on zemscripten.

sgulgas:
with build\emscripten related changes - would it make sense to additionally tweak it so emscripten dependencies are lazy loaded so it's only pulled down if targeting emscripten os? 

not a huge deal but it's less stuff to pull down (and vendor) if it's not your target

Issue #5950, opened 2026-07-03
[rmodels] Build failure in 'LoadIQM' on GCC 16.1


eloj:
### Issue description

The IQM object loader code is assigning an `unsigned int` from `iqmHeader->num_joints` to `boneCount` in `struct ModelSkeleton`, which is an `int`. The compiler detects that this can potentially overflow into a negative value, which is then passed to the allocator as a `size_t` potentially resulting in a huge allocation.

Tested at bb96b920d34eed6

### The Error

```shell
[ 36%] Linking C executable models_orthographic_projection
In function 'LoadIQM',
    inlined from 'LoadModel.constprop' at /home/eddy/dev/EXT/raylib/src/rmodels.c:1110:52:
/home/eddy/dev/EXT/raylib/src/rmodels.c:5014:38: error: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
 5014 |     model.currentPose = (Transform *)RL_CALLOC(model.skeleton.boneCount, sizeof(Transform));
      |                                      ^
/usr/include/stdlib.h: In function 'LoadModel.constprop':
/usr/include/stdlib.h:675:14: note: in a call to allocation function 'calloc' declared here
  675 | extern void *calloc (size_t __nmemb, size_t __size)
      |              ^
In function 'LoadIQM',
    inlined from 'LoadModel.constprop' at /home/eddy/dev/EXT/raylib/src/rmodels.c:1110:52:
/home/eddy/dev/EXT/raylib/src/rmodels.c:5015:36: error: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
 5015 |     model.boneMatrices = (Matrix *)RL_CALLOC(model.skeleton.boneCount, sizeof(Matrix));
      |                                    ^
/usr/include/stdlib.h: In function 'LoadModel.constprop':
/usr/include/stdlib.h:675:14: note: in a call to allocation function 'calloc' declared here
  675 | extern void *calloc (size_t __nmemb, size_t __size)
      |              ^
In function 'LoadIQM',
    inlined from 'LoadModel.constprop' at /home/eddy/dev/EXT/raylib/src/rmodels.c:1110:52:
[...]

PR #5955, opened 2026-07-05
[rmodels] Change boneCount field type from int to unsigned int

lauchimoon:
Addresses issue #5950.


raysan5:
@lauchimoon this change breaks th alignment harmony of the variabl inside the struct, that's important to me... but anyway...

PR #5956, opened 2026-07-06
[rcore] Fix MaximizeWindow() not resizing the screen

marre8009-debug:
When tinkering with raylib I noticed that MaximizeWindow() not resizing the screen. After printing the attribute value of GLFW_RESIZABLE, it printed 0.

### **Code before fix:**
image

The solution sets the GLFW_RESIZABLE to GLFW_TRUE for cases where GLFW_RESIZABLE happens to be false. The solution was tested by running the example codes that exists in the raylib repo such as core_3d_camera_fps.c and core_input_keys.c in Windows 11

### **Code after fix:** 
image

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -334,11 +334,12 @@ void ToggleBorderlessWindowed(void)
 // Set window state: maximized, if resizable
 void MaximizeWindow(void)
 {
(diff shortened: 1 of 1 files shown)


marre8009-debug:
Hey Ray. I saw that you closed the PR with no comment. Could you please give me feedback as to what might have went wrong. It's my first time making a PR so I'm new when it comes to large codebases but I'm open to learn and improve.

PR #5957, opened 2026-07-06
fix UB in ComputeSHA-1

TheFissk:
Fixes UB related to integer promotion in `ComputeSHA1`.

When compiling with UBSans on (like the zig compiler does) unsigned chars are promoted to integers, so any value in `msg` > 128 will cause the 32nd bit to be set and technically triggering a signed integer overflow. This shouldn't cause issues in most compilers but it is techincally UB. explicitly casing the msg to an unsigned int fixes the problem

the fastest way to replicate the above issue to run `zig build core_compute_hash` and press the `compute input data hashes button`


raysan5:
@TheFissk Thanks for the review!

PR #5964, opened 2026-07-10
[rlsw] Fix alpha tracking for persistent current colors

### LINKED ISSUE #5961: [rlsw] Persistent translucent current color loses blending after first primitive
**Affected:** `external/rlsw.h` v1.5 (`GRAPHICS_API_OPENGL_SOFTWARE`).
Verified against raylib master `caadb48e`; also present in `65abee1c`.
Platform-independent - reproduced on x86 (Windows/MinGW, both SSE and x87 FP)
and on m68k/AmigaOS.

## Summary

In rlsw immediate mode, a translucent current color set once before multiple
primitives is only considered for the first completed primitive. After the
primitive flush, `hasColorAlpha` is reset even though the current color
persists, so the blend-elision path disables `SW_STATE_BLEND` for later
untextured primitives.

## Minimal reproduction (pure rlsw)

```c
swEnable(SW_BLEND); swBlendFunc(SW_SRC_ALPHA, SW_ONE_MINUS_SRC_ALPHA);
// framebuffer cleared to white

swBegin(SW_TRIANGLES);
swColor4ub(135, 60, 190, 153);      // set ONCE, alpha ~0.6 (GL current color)
swVertex2f(100,100); swVertex2f(100,200); swVertex2f(200,100);  // triangle 1
swVertex2f(200,100); swVertex2f(100,200); swVertex2f(200,200);  // triangle 2
swEnd();
```

**Expected:** both triangles blend identically (~60% violet over white).
**Actual:** triangle 1 is blended; triangle 2 renders unblended/source-like
(pure vertex color, as if `SW_BLEND` were off). Illustrative readback from my
harness: triangle 1 interior ~= (183,138,216), triangle 2 interior = (135,60,190)
(exact byte order depends on the configured framebuffer output format; the
qualitative "first blended, second source-like" split is the point).

## Reproduction at raylib level

With `SUPPORT_QUADS_DRAW_MODE=0`, `DrawRectangleRec(rec, Fade(VIOLET, 0.6f))`
emits one `rlColor4ub` followed by two triangles: the result shows a hard
diagonal seam (first triangle blended, second opaque).
`DrawRectangleLinesEx(rec, 6, Fade(BLACK, 0.3f))` renders partly opaque black
borders. This is how we noticed (raylib's `shapes_colors_palette` example,
hover state).

Note on why the default desktop config rarely shows it: with
`SUPPORT_QUADS_DRAW_MODE=1` most shapes bind the shapes/font texture, whose
GRAY_ALPHA format keeps the `boundTexture->alpha != SW_PIXEL_ALPHA_NONE`
escape in the elision check alive, so blending survives. Functions that
re-submit the color per primitive (e.g. `DrawPoly`, which calls `rlColor4ub`
per segment) are also unaffected. The bug needs an untextured path with a
single persistent translucent color - which is standard GL immediate-mode
usage.

## Root cause

`RLSW.primitive.hasColorAlpha` is used as the basis for the blend-elision
optimization in `sw_immediate_push_vertex()`:

```c
else if ((RLSW.blendFlags & SW_BLEND_FLAG_NEEDS_ALPHA) && (!RLSW.primitive.hasColorAlpha))
    ... state &= ~SW_STATE_BLEND;
```

The flag is OR-ed in by `sw_immediate_set_color()`, but it does not conserve
the persistent current color across primitive boundaries: it is hard-reset to
`false` both in `sw_immediate_begin()` and after each primitive flush in
`sw_immediate_push_vertex()`. Since the current color persists across
primitives and across `swEnd()` (GL semantics) and seeds every following
vertex, all primitives after the first lose `SW_STATE_BLEND` unless the
application happens to re-submit the color.

## Suggested fix

Re-derive the flag from the persisting current color instead of clearing it,
at both reset sites:

```c
// sw_immediate_begin(), and after the primitive flush in sw_immediate_push_vertex():
RLSW.primitive.hasColorAlpha = (RLSW.primitive.color[3] < 1.0f);
```

The `|=` in `sw_immediate_set_color()` stays as is, so a translucent color
submitted mid-primitive still marks the whole primitive.

Verified with this change: the two-triangle rect is pixel-uniform, the
`DrawRectangleLinesEx` borders come out at the exact 30% blend, and a
regression sweep over the plain / blend / depth / blend+depth raster paths
shows no behavior change for opaque colors.

Found while porting raylib to AmigaOS 3.x (m68k, rlsw backend); first
suspected our cross-compiler, but an x86 host harness with vanilla rlsw
reproduced it 1:1.

iztanos:
Fixes #5961.

This follows the root-cause analysis from @HurricanVD in #5961.

The issue suggested preserving `hasColorAlpha` from the persistent current color
at primitive boundaries. This PR takes a slightly narrower approach: it updates
`hasColorAlpha` when a vertex copies the current color.

That keeps the flag tied to the primitive contents themselves. A primitive is
marked as containing alpha only if one of its submitted vertices actually has
alpha below `1.0f`, while persistent translucent colors still remain correctly
tracked across multiple primitives.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -3909,7 +3909,6 @@ static void sw_immediate_set_color(const float color[4])
     RLSW.primitive.color[2] = color[2];
     RLSW.primitive.color[3] = color[3];
 
-    RLSW.primitive.hasColorAlpha |= (color[3] < 1.0f);
 }
 
 static void sw_immediate_set_texcoord(const float texcoord[2])
@@ -3947,6 +3946,9 @@ static void sw_immediate_push_vertex(const float position[4])
     // Copy the attributes in the current vertex
     for (int i = 0; i < 4; i++) vertex->color[i] = RLSW.primitive.color[i];
     for (int i = 0; i < 2; i++) vertex->texcoord[i] = RLSW.primitive.texcoord[i];
+
+    // Track whether any vertex in this primitive has alpha < 1.0
+    RLSW.primitive.hasColorAlpha |= (vertex->color[3] < 1.0f);
 
     // Immediate rendering of the primitive if the required number is reached
     if (RLSW.primitive.vertexCount == SW_PRIMITIVE_VERTEX_COUNT[RLSW.drawMode])


raysan5:
@Bigfoot71 please, could you takee a look to this PR?

src/external/rlsw.h
Bigfoot71: I propose that we also remove `RLSW.primitive.hasColorAlpha |= (color[3] < 1.0f);` from `sw_immediate_set_color()`, and doing this check only in `sw_immediate_push_vertex()`

src/external/rlsw.h
```diff
@@ -3947,6 +3947,7 @@ static void sw_immediate_push_vertex(const float position[4])
     // Copy the attributes in the current vertex
     for (int i = 0; i < 4; i++) vertex->color[i] = RLSW.primitive.color[i];
     for (int i = 0; i < 2; i++) vertex->texcoord[i] = RLSW.primitive.texcoord[i];
+    RLSW.primitive.hasColorAlpha |= (vertex->color[3] < 1.0f);
```
Bigfoot71, line 3951: And while we're at it, we could also make the intent explicit, for example:
```c
    // Track whether any vertex in this primitive has alpha < 1.0
    RLSW.primitive.hasColorAlpha |= (vertex->color[3] < 1.0f);
```

Bigfoot71 (changes requested):

Bigfoot71:
Yes, there is indeed a logic issue, and the proposed fix is correct.

I just suggested removing a line that becomes unnecessary with this change, and perhaps adding a comment to make the intent explicit for everyone.

Otherwise, looks good to me.

raysan5:
@iztanos please, could you implement and test proposed changes? thanks!

iztanos:
Updated, thanks for the review @Bigfoot71.

I removed the old `hasColorAlpha` update from `sw_immediate_set_color()` and kept the check only when a vertex is actually pushed. That makes the flag describe the primitive being built, instead of the current color state.

I also added the small comment you suggested near the new check.

Rebased on current `master` and reran the small `rlsw.h` regression test in Docker. The persistent translucent color case still renders both triangles blended.

raysan5:
@iztanos thanks for the improvement and thanks @Bigfoot71 for the review

PR #5967, opened 2026-07-11
[rcore][glfw] Fix initial macOS window size

loukamb:
While developing a game on macOS, I found that requesting a window taller than the visible work area caused the top of the framebuffer to be clipped and mouse input to be vertically offset. AppKit constrains the window during creation before raylib installs its resize callbacks, so raylib keeps the requested dimensions instead of the actual Cocoa content size and initializes an oversized screen and render area. This change reads the window size back from GLFW after creation on macOS and before render dimensions are initialized, keeping raylib's screen and framebuffer sizes synchronized with the window AppKit created.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1693,6 +1693,14 @@ int InitPlatform(void)
             return -1;
         }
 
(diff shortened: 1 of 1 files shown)


raysan5:
@loukamb Good catch! Thanks for the review!

PR #5970, opened 2026-07-11
[rcore] Add an optional render callback to handle updates during window events

JeffM2501:
This PR adds a 100% optional render callback and a RunGameLoop function to call that callback in a game loop.
This render callback is stored and provided to the platform so it can be used to do game loop updates while processing events, such as move and resize on Windows OSs. This means the game still runs while you resize.

Additionally this allows the same code to work well with Emscripten backends, since the render callback can be sent to emscripten_main_loop.

This is an experimental and tentative PR. The structure is not set in stone. Feedback and comments are welcome.


raysan5:
@JeffM2501 The idea is good and I see the usefulness but I find it a bit complex to follow, mostly intended for advance users... not sure about it... keeping it open for further review...

JeffM2501:
> @JeffM2501 The idea is good and I see the usefulness but I find it a bit complex to follow, mostly intended for advance users... not sure about it... keeping it open for further review...

Yes, I am also not sure about the implementation. There are several ways it could be done inside.

raysan5:
I'm afraid I'm not merging this change, I find it confusing and I think most users do not really need it.

JeffM2501:
Then I suggest that we add a note to the readme that raylib does not support rendering during resizing or window movement on windows platforms.

raysan5:
@JeffM2501 added note

PR #5975, opened 2026-07-12
Fix TextSubtext() returning empty string when position is 0

### LINKED ISSUE #5973: TextSubtext() returns an empty string for position == 0 (breaks the text_writing_anim example)
**Module:** rtext
**raylib version:** master (`f00317be`, `6.0-218`); also present on current HEAD
**Platform/compiler:** platform-independent (logic bug in `rtext.c`; found while porting to an AmigaOS software renderer, but the code path is not platform-specific)

### Description

`TextSubtext(text, position, length)` guards its copy with `(position > 0)`, so a call with `position == 0` (the start of the string) falls through and returns an empty string. Position 0 is a valid start offset, and it is exactly what raylib's own example `examples/text/text_writing_anim.c` uses:

```c
DrawText(TextSubtext(message, 0, framesCounter/10), 210, 160, 20, MAROON);
```

As a result, that example's typewriter reveal never draws anything — only the two static hint lines appear. Any code that reveals a string from its beginning via `TextSubtext(s, 0, n)` is affected.

### Root cause

`src/rtext.c`, `TextSubtext()`:

```c
if ((text != NULL) && (position > 0) && (length > 0))   // position > 0 drops the valid position 0
```

The `(position > 0)` guard was introduced in the recent text-API hardening series. Before that, the function only checked `(text != NULL)` and handled position 0 correctly.

### Fix

```diff
-    if ((text != NULL) && (position > 0) && (length > 0))
+    if ((text != NULL) && (position >= 0) && (length > 0))
```

`(length > 0)` is fine (length 0 -> empty substring is correct); the inner `if (position < textLength)` already bounds the upper end. Only the lower bound needs to include 0.

### Steps to reproduce

1. Build and run `examples/text/text_writing_anim`.
2. Expected: the MAROON message is revealed character by character.
3. Actual: the MAROON message never appears (empty substring); only the two hint lines render.

Minimal check:

```c
const char *s = TextSubtext("Hello", 0, 3);   // expected "Hel", actual "" on master
```

thiagomucci:
### Description
This PR fixes a bug in `TextSubtext()` where passing a `position` of `0` (the start of the string) would incorrectly return an empty string. The recent API hardening introduced a strict `position > 0` check, which broke valid use cases like the `text_writing_anim` example. Changing the guard to `position >= 0` fixes the issue

Closes #5973

_Thanks to @HurricanVD for finding the root cause and providing the logic fix in the issue description!_

### DIFF
--- a/src/rtext.c
+++ b/src/rtext.c
@@ -1699,7 +1699,7 @@ const char *TextSubtext(const char *text, int position, int length)
     static char buffer[MAX_TEXT_BUFFER_LENGTH] = { 0 };
     memset(buffer, 0, MAX_TEXT_BUFFER_LENGTH);
 
-    if ((text != NULL) && (position > 0) && (length > 0))
+    if ((text != NULL) && (position >= 0) && (length > 0))
     {
         int textLength = TextLength(text);
 


raysan5:
@thiagomucci Good catch, thanks for the fix

PR #5980, opened 2026-07-14
[rshapes] Fix DrawRectangleRoundedLinesEx not scaling with transform when thick=1

SamaelHeaven:
In the DrawRectangleRoundedLinesEx function, if the line thickness is <= 1, it uses RL_LINES to draw the lines, otherwise, it uses quads. However, this makes the two behave differently, if we multiply the matrix by a scale or zoom via a camera, RL_LINES stays the same size, whereas the quads are scaled properly.


raysan5:
@SamaelHeaven thanks for the review

Issue #5945, opened 2026-07-01
[rcore] When mouse is outside window Raylib gives stale mouse information


largenumberhere:
> Please, before submitting a new issue verify and check:
> - [x] I tested it on latest raylib version from master branch
> - [x] I checked there is no similar issue already reported
> - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
> - [x] My code has no errors or misuse of raylib

#### Background
I've been messing around making my own wrappers for raylib drawing logic lately. I have a window where I want to detect which section of the window is hovered over and respond to it correctly. I noticed through this process that raylib does not seem to clearly report when I move the mouse outside of the window. At first I assumed `GetMousePosition` was returning some default value, and adjusted my logic accordingly and still had strange behaviour. 

#### Details
When moving the mouse cursor outside of the window, `GetMousePosition` returns stale values, and no other window function seems to notice (`IsWindowState`, `IsWindowFocused`, etc). 
In particular, `GetMousePosition` returns a value somewhat close to where the mouse exited the window. How accurate it is, depends on how fast I move the mouse out of the window. This makes it wholly un-useful information.

#### Proposal
For a simple solution, I think that `GetMousePosition` should return `0,0` when the cursor is outside the window. It is extremely unlikely that any user will move their cursor to exactly this position. It is also sane default that can be ignored when not important, and has an intuitive meaning of initialized but probably not meaningful. It is also internally consistent because it matches the default cursor position when the window has been launched but not yet hovered over.

As for implementation, as starting point glfw has [cursor enter/leave callbacks](https://www.glfw.org/docs/3.3/input_guide.html#cursor_enter). They can be used to observe these events and update state without much hassle.

My intuition says that `IsWindowFocused()` should also handle this case by returning false, even if the window manager would still technically classify the window as focused. This however, depends on the intended precise semantics of what 'Focused' should mean in raylib.

### Environment
Os: `Ubuntu 24.04.4 LTS distribution.`
Raylib: `6.0.0`
OpenGl: 
```
INFO: GL: OpenGL device information:
INFO:     > Vendor:   Intel
INFO:     > Renderer: Mesa Intel(R) Iris(R) Xe Graphics (ADL GT2)
INFO:     > Version:  4.6 (Core Profile) Mesa 25.2.8-0ubuntu0.24.04.2
INFO:     > GLSL:     4.60
```

Gpu
```
  *-display                 
       description: 3D controller
[...]

PR #5981, opened 2026-07-16
[rcore] Fix for stale mouse position on web platform

largenumberhere:
This is a follow up to [issue 5945](https://github.com/raysan5/raylib/issues/5945) that applies a fix for web.
> When moving the mouse cursor outside of the window, GetMousePosition returns stale values, and no other window function seems to notice

This fix is similar to [commit 5e37421](https://github.com/raysan5/raylib/commit/5e374218eec7f137fb8f28c75fd97a639628efa1) to solve the same bug on another platform. I am using this platform's cursor enter/leave callback to zero the stored position the same way.

## Checks
- The wasm and windows build pipelines still work
- Follows the same conventions as similar fix
- No similar pr has been opened recently.
- Solves the bug in a test project
- Synchronized with the latest commit of the main branch.

### DIFF
--- a/src/platforms/rcore_web_emscripten.c
+++ b/src/platforms/rcore_web_emscripten.c
@@ -1507,8 +1507,16 @@ static EM_BOOL EmscriptenMouseCallback(int eventType, const EmscriptenMouseEvent
 {
     switch (eventType)
     {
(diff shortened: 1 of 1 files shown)


raysan5:
@largenumberhere thanks for the review!

PR #5982, opened 2026-07-16
[rtext] Fix TextInsertAlloc buffer overflow

GideonSerf:
### Problem

`TextInsertAlloc()` allocates `textLen + insertLen + 1` bytes, but its suffix-copy loop iterates until `textLen + insertLen + position`.

For any insertion position greater than zero, this reads beyond the source string and writes beyond the allocated result buffer.

### Fix

Remove the duplicated `position` term from the suffix-copy loop bound.

### Validation

- Reproduced the original overflow with MSVC AddressSanitizer


raysan5:
@GideonSerf thanks for the review

PR #5984, opened 2026-07-16
[rtextures][rlgl] Fix pixel data size regressions

GideonSerf:
### Problem
Recent pixel-size hardening added 64-bit variables to safely check image data sizes before converting them to `int`. Some multiplication operands remain `int`, so the intermediate calculation may still occur at 32-bit width before assignment.

Explicitly promoting the first operand ensures the calculation consistently uses 64-bit arithmetic across compilers.

`ImageRotate()` used the total destination image size as its per-pixel stride, causing out-of-bounds access for images larger than one pixel.

### Fix

- Promote pixel-size multiplication operands before evaluation in both `GetPixelDataSize()` and `rlGetPixelDataSize()`.
- Keep `ImageRotate()` allocation size and per-pixel stride as separate values.

### Validation

- Reproduced the `ImageRotate()` overflow with MSVC AddressSanitizer.
- Verified large valid and over-limit pixel-size calculations.
- Verified zero-degree RGBA8 image rotation preserves dimensions and data.
- Built raylib and all examples with MSVC x64 Release.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -5237,7 +5237,7 @@ static int rlGetPixelDataSize(int width, int height, int format)
         {
             int blockWidth = (width + 3)/4;
             int blockHeight = (height + 3)/4;
-            unsigned long long dataSizeBytes = blockWidth*blockHeight*8;
+            unsigned long long dataSizeBytes = (unsigned long long)blockWidth*blockHeight*8;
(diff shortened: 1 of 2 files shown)


raysan5:
@GideonSerf Oh! Thanks for the review!

GideonSerf:
@raysan5 No problem 😊!

PR #5985, opened 2026-07-16
English comments for android_mative_app_glue

penggrin12:
Seems that those two files were copied from https://github.com/chicken-mobile/android-ndk/tree/master/android-native-app and then (i think machine) translated to Russian

Doesn't touch any mismatching code


raysan5:
@penggrin12 thanks for reporting! That sample project was added 5 years ago and it's unmaintained so probably better just remove it completely.

PR #5990, opened 2026-07-17
[build] Fix static CMake link dependencies

### LINKED ISSUE #5926: [build][CMake] Linker errors when building raylib as static library and using OpenGL ES 2.0
 - [] I checked there is no similar issue already reported: maybe #4648
 - [x] My code has no errors or misuse of raylib: no code needed

### Issue description
I'm trying to compile raylib with the pair `PLATFORM=Desktop` and `OPENGL_VERSION="ES 2.0"`. I know from discussions in #4648 that this is not a priority, but this issue exposes unexpected behaviour in the cmake build system.

If I configure raylib to be a static library, with `PLATFORM=Desktop` and `OPENGL_VERSION="ES 2.0"`, I install it and load it from an external cmake project with `find_package(raylib)` I get linker errors describing missing opengl symbols like `glClear`

If I configure raylib to be a shared library instead, the linker errors are solved and the build completes.

### Environment
This has been tested on linux machines:
  * Raspberry PI 1B+ (armv6l)
  * Radxa (aarch64) - Debian bookworm
  * x86 machine - Ubuntu 24.04 LTS
I will explain why I don't think this is an environment problem

### Code Example
To reproduce the build, we will use the provided CMake project

Static library build, fails with linker errors for unresolved opengl symbols
```cmake
cd raylib
mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_EXAMPLES:BOOL=OFF -DPLATFORM=Desktop -DOPENGL_VERSION="ES 2.0"
make -j
sudo make install

cd ../projects/CMake
mkdir build && cd build
cmake .. -DPLATFORM=Desktop
make
```

Dynamic library build, all good
```cmake
cd raylib
mkdir build && cd build
cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_EXAMPLES:BOOL=OFF -DPLATFORM=Desktop -DOPENGL_VERSION="ES 2.0" -DBUILD_SHARED_LIBS:BOOL=ON"
make -j
sudo make uninstall # this is important for reprod
sudo make install

cd ../projects/CMake
mkdir build && cd build
cmake .. -DPLATFORM=Desktop
make
```

There is a catch because the installation may not rewrite some files, so it's better to uninstall the previous version before installing raylib once again

### Solution
I have come up with a number of solution in the static library case, maybe this issue can be a reference for other users.

One can import the raylib package and at the moment of linking just add the libraries needed
```cmake
set(RAYLIB_VERSION 6.0)
find_package(raylib ${RAYLIB_VERSION} REQUIRED)

add_executable(main core_basic_window.c)
target_link_libraries(${PROJECT_NAME} raylib GL) # or whatever the user wants, maybe it's more proper to use GLESv2
```

but this solution goes against the spirit of a cmake build. The user should not be thinking about the dependencies of raylib.

The other solution involves modifying the cmake build system to:
  1. properly include the opengl es2 packages (and maybe we can also use the occasion to work on supporting opengl es3)
  2. change the way in which dependencies are exported

for (2) I have found that the file `src/CMakeLists.txt` at line 107 says
```cmake
target_link_libraries(raylib PRIVATE $<BUILD_INTERFACE:${LIBS_PRIVATE}>)
```

changing it to
```cmake
target_link_libraries(raylib PRIVATE ${LIBS_PRIVATE})
```

solves the problem for me

### Elephant in the room
I'm trying to use ES2.0, but I'm linking with GL. Yes, I have also tried to manually add GLESv2 to the `LIBS_PRIVATE` variable in `cmake/LibraryConfigurations.cmake` at line 80, but that didn't solve my problem. If this issue gets picked I can dig more into it

### Conclusions
This issue is not blocking for me, as I have applied duct tape. It could be an enhancement in the cmake experience of the project. I understand the desire to keep the system free from complications.

meta-legend:
Fixes #5926.

When raylib is installed as a static library, downstream CMake consumers need raylib's private platform/link dependencies to be present in the exported target metadata. Those dependencies were previously wrapped in "BUILD_INTERFACE", so the installed "raylib-targets.cmake" did not carry them forward.

This caused projects using "find_package(raylib)" to configure successfully but fail at link time with unresolved OpenGL/GLES symbols when raylib was built with "PLATFORM=Desktop" and 'OPENGL_VERSION="ES 2.0"'.

This keeps the existing build-only private link behavior, but adds install-only "LINK_ONLY" entries for static-library consumers. Bundled GLFW is excluded from the install interface when it is built as raylib object sources, because it is not an exported target.

I tested on Linux (WSL):

- I built and installed static raylib with "PLATFORM=Desktop", 'OPENGL_VERSION="ES 2.0"', "BUILD_SHARED_LIBS=OFF"
- I also built projects/CMake against the installed package
- Before (issue reproduced): consumer link failed with unresolved GL symbols such as "glViewport" and "glClear"
- After (with fix): consumer linked successfully

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -107,6 +107,20 @@ endif ()
 target_link_libraries(raylib PRIVATE $<BUILD_INTERFACE:${LIBS_PRIVATE}>)
 target_link_libraries(raylib PUBLIC ${LIBS_PUBLIC})
 
+# Static libraries must export their private link dependencies for installed consumers.
+# LINK_ONLY keeps those dependencies out of the compile interface.
+if (NOT BUILD_SHARED_LIBS)
+    set(raylib_install_private_libs ${LIBS_PRIVATE})
+
+    if (NOT glfw3_FOUND)
+        list(REMOVE_ITEM raylib_install_private_libs glfw)
+    endif()
+
+    foreach(lib IN LISTS raylib_install_private_libs)
+        target_link_libraries(raylib INTERFACE $<INSTALL_INTERFACE:$<LINK_ONLY:${lib}>>)
+    endforeach()
+endif()
+
 # Sets some compile time definitions for the pre-processor
 # If CUSTOMIZE_BUILD option is on you will not use config.h by default
 # and you will be able to select more build options


raysan5:
@Peter0x44 @manuel5975p @VictorHM @JohnnyCena123, @ninadsachania, @arodriguez22, @pamarcos, @Shakai-Dev, @SajendraN, @heliumsneakers, @nukelash, @RobLoach Please, could you take a look to this PR?

_NOTE: I'm pinging all of you in my list of `CMake build system` maintainers, if any of you doesn't want to be pinged for this kind of issues, just let me know and I'll take you out of the maintainers list._

Peter0x44:
Why did this change almost every line in the file? Did you change the line endings by accident or something?

I can't review it when the diff looks like that.

raysan5:
Just note that the user seems to be an AI, checking its history of PRs... most probably this PR has been barely reviewed by an human and even less tested (by a human) on real devices... the magik of AI slop...

meta-legend:
> Why did this change almost every line in the file? Did you change the line endings by accident or something?
> 
> I can't review it when the diff looks like that.

Yes, I screwed up the line endings (mixing WSL and Windows edits on a file together is most likely the cause); I could fix it if you want.

> Just note that the user seems to be an AI, checking its history of PRs... most probably this PR has been barely reviewed by a human and even less tested (by a human) on real devices... the magik of AI slop...

I'm not an "AI" and if you looked at my PRs, you would notice that they are very trivial (that is why there are so many), like this one (ig they are also a bit random, but idk what to tell you abt that). This was really tested with WSL, but you could just fix the issue yourself if you prefer (it's just a 12-line change).

I attached both WSL test logs with the unfixed and fixed versions to show the difference between my fix being applied and not.

[wsltestfix.log](https://github.com/user-attachments/files/30135752/wsltestfix.log)
[wsltestunfixed.log](https://github.com/user-attachments/files/30135753/wsltestunfixed.log)

raysan5:
@meta-legend thanks for clarifying, no intention to offend, just cheked your profile and the projects you contributed with barely any previous contributions... it looked suspicious... anyway... thanks for your contribution.

meta-legend:
> @meta-legend thanks for clarifying, no intention to offend, just cheked your profile and the projects you contributed with barely any previous contributions... it looked suspicious... anyway... thanks for your contribution.

no problem, do you want me to fix the line endings?

raysan5:
> do you want me to fix the line endings?

yes, please

meta-legend:
Line endings should hopefully be normal now 🙏

src/CMakeLists.txt
```diff
@@ -107,6 +107,18 @@ endif ()
 target_link_libraries(raylib PRIVATE $<BUILD_INTERFACE:${LIBS_PRIVATE}>)
 target_link_libraries(raylib PUBLIC ${LIBS_PUBLIC})
 
+if (NOT BUILD_SHARED_LIBS)
```
RobLoach, line 112: Would be nice to have a code comment about what and why this is doing its thing. Similar to the other blocks.

Peter0x44:
Now that the line endings are fixed and I can see what changed exactly, it looks reasonable to me. Approved.

src/CMakeLists.txt
meta-legend: i added:
```
# Static libraries must export their private link dependencies for installed consumers
# LINK_ONLY keeps those dependencies out of the compile interface
```

RobLoach (approved):
Thanks much. With this, I was able to build without issue.

raysan5:
@meta-legend thanks for the imprivement. @Peter0x44 @RobLoach thanks for the quick review! 💯

heliumsneakers:
I've just seen this now, after taking a look and running some tests in a container I think it's worth noting that this PR does fix the shared case. But it also exports raw `LIBS_PRIVATE` entries. Specifically external GLFW is exported as`$<LINK_ONLY:glfw>`, which becomes `-lglfw`. That fails for an external static GLFW CMake install, where the installed library is `libglfw3.a`, and it can also resolve the wrong GLFW if multiple versions are present on the user's machine. Though this isn't exactly a breaking change it's important note that users may run into this issue if they are trying to use a static build of GLFW with raylib, if they'd like to stick to some pinned version for example.

PR #5991, opened 2026-07-17
[rtextures] Fix: ImageFromImage() heap-buffer-overflow on compressed formats

### LINKED ISSUE #5989: [rtextures] `ImageFromImage()` heap-buffer-overflow on compressed formats
`ImageFromImage()` computes bytes-per-pixel with `GetPixelDataSize(1, 1, image.format)`,
which returns a *block minimum* for compressed formats rather than a per-pixel size. The
resulting row stride is 16-32x too large and the copy loop reads past the end of the source
image data. Reproduced on current master with ASan.

### Environment
- raylib: master @ <sha>
- OS: Gentoo Linux, GCC 15, ASan (libasan.so.8)

### Reproduction

```sh
magick input.png -resize 256x256! -define dds:compression=dxt1 test.dds
cd src && make clean && make CUSTOM_CFLAGS="-fsanitize=address -g"
cc test_dds.c -o test -fsanitize=address -g -I../../src -L../../src -lraylib -lGL -lm -lpthread -ldl -lrt -lX11
```

```c
#include "raylib.h"

int main(void)
{
    Image img = LoadImage("test.dds");
    Image piece = ImageFromImage(img, (Rectangle){ 0, 0, 64, 64 });
    UnloadImage(piece);
    UnloadImage(img);
    return 0;
}
```

No need for a window or opengl context.

### ASan output

```
==29458==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 512 at 0x7e8cfe5f7400 thread T0
    #0 memcpy
    #1 ImageFromImage rtextures.c:1248
    #2 main test_dds.c:6
0x7e8cfe5f7400 is located 1366 bytes after 43690-byte region
allocated by thread T0 here:
    #1 rl_load_dds_from_memory external/rltexgpu.h:385
```

### Cause

`GetPixelDataSize()` contains a block-minimum special case:

```c
// Most compressed formats works on 4x4 blocks,
// if texture is smaller, minimum dataSize is 8 or 16
if ((width < 4) && (height < 4))
{
    if ((format >= PIXELFORMAT_COMPRESSED_DXT1_RGB) && (format < PIXELFORMAT_COMPRESSED_DXT3_RGBA)) dataSize = 8;
    else if (...) dataSize = 16;
}
```

Called as `GetPixelDataSize(1, 1, fmt)`, `1 < 4` always holds, so `bytesPerPixel` becomes
8 (DXT1) or 16 (DXT3/DXT5/ETC/PVRT/ASTC_4x4) instead of the true 0.5-1 bytes/pixel.
`PIXELFORMAT_COMPRESSED_ASTC_8x8_RGBA` falls through both branches and yields 0, so the
function allocates nothing, copies nothing, and returns a zeroed `Image` with no warning.

The 256x256 DXT1 test image has 9 mipmaps; `LoadDDS` allocates the whole chain as one
43690-byte block. With a stride of `256 * 8 = 2048`, row 22 reads at offset 45056, i.e.
1366 bytes past the end, matching ASan exactly.

There's 2 failure modes from the same wrong stride:
- **Small images**: OOB read (above).
- **Large images**: the read stays inside the allocation but interprets mipmap or unrelated
  pixel data as base-level pixels. Returns a plausible-looking `Image` full of garbage with
  no crash or warning, a 3830x2119 DXT1 source produced no ASan report at all.

### Note: adjacent functions already guard against this

`ImageFromImage()` appears to be the only user of the `GetPixelDataSize(1, 1, fmt)` idiom
that doesn't gate on format:

- `ImageCrop()` warns, `else { ...copy... }`
- `ImageFlipVertical()` warns, `else { ...copy... }`
- `ImageResize()` explicit whitelist of uncompressed formats

It also lacks the `(image.data == NULL) || (image.width == 0) || (image.height == 0)` check
those functions have, though a zeroed `Image` is already rejected by the existing rectangle
validation.

### Suggested fix

Mirror `ImageCrop()`. Cropping block-compressed data at arbitrary pixel offsets isn't
meaningful, so refusing seems right:

```c
Image ImageFromImage(Image image, Rectangle rec)
{
    Image result = { 0 };

    // Security check to avoid program crash
    if ((image.data == NULL) || (image.width == 0) || (image.height == 0)) return result;

    if (image.format >= PIXELFORMAT_COMPRESSED_DXT1_RGB)
    {
        TRACELOG(LOG_WARNING, "IMAGE: Image manipulation not supported for compressed formats");
        return result;
    }
    ...
```

Happy to send a PR if this approach looks right.

Minor: the existing warning in this function reads `"Rectangle provided for ImageToImage not valid"`, should presumably be `ImageFromImage`.

Possibly related, unverified: `ImageClearBackground()` uses the same idiom with no format gate.

gummyniki:
Fixes https://github.com/raysan5/raylib/issues/5989

ImageFromImage() used GetPixelDataSize(1, 1, format) as bytes-per-pixel, which returns a
block minimum for compressed formats rather than a per-pixel size, producing a 16-32x
oversized row stride and an out-of-bounds read (ASan trace in https://github.com/raysan5/raylib/issues/5989).

Now warns and skips for compressed formats, matching ImageCrop(). Verified with ASan:
the DDS repro is clean, and uncompressed images are unaffected.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -1230,23 +1230,31 @@ Image ImageFromImage(Image image, Rectangle rec)
 {
     Image result = { 0 };
 
+    // Security check to avoid program crash
+    if ((image.data == NULL) || (image.width == 0) || (image.height == 0)) return result;
+
     // Basic rectangle validation: size smaller than image size
     if ((rec.x >= 0) && (rec.y >= 0) && (rec.width > 0) && (rec.height > 0) &&
         (((int)rec.x + (int)rec.width) <= image.width) &&
         (((int)rec.y + (int)rec.height) <= image.height))
     {
-        int bytesPerPixel = GetPixelDataSize(1, 1, image.format);
 
-        result.width = (int)rec.width;
-        result.height = (int)rec.height;
-        result.data = RL_CALLOC((int)rec.width*(int)rec.height*bytesPerPixel, 1);
-        result.format = image.format;
-        result.mipmaps = 1;
+	if (image.format >= PIXELFORMAT_COMPRESSED_DXT1_RGB) TRACELOG(LOG_WARNING, "IMAGE: Image manipulation not supported for compressed formats");
+	else
+	{
+            int bytesPerPixel = GetPixelDataSize(1, 1, image.format);
 
-        for (int y = 0; y < (int)rec.height; y++)
-        {
-            memcpy(((unsigned char *)result.data) + y*(int)rec.width*bytesPerPixel, ((unsigned char *)image.data) + ((y + (int)rec.y)*image.width + (int)rec.x)*bytesPerPixel, (int)rec.width*bytesPerPixel);
-        }
+            result.width = (int)rec.width;
+            result.height = (int)rec.height;
+            result.data = RL_CALLOC((int)rec.width*(int)rec.height*bytesPerPixel, 1);
+            result.format = image.format;
+            result.mipmaps = 1;
+
+            for (int y = 0; y < (int)rec.height; y++)
+            {
+                memcpy(((unsigned char *)result.data) + y*(int)rec.width*bytesPerPixel, ((unsigned char *)image.data) + ((y + (int)rec.y)*image.width + (int)rec.x)*bytesPerPixel, (int)rec.width*bytesPerPixel);
+            }
+	}
     }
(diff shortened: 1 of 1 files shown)


raysan5:
@gummyniki thanks for the review

Issue #5993, opened 2026-07-18
[rlsw] Decision needed on which double buffering approach to merge for removing buffer transfer overhead


NighthowlerStudios:
On #5827 Jensroth mentioned how the ESP32 port enabled double buffering to allow dual-core microcontrollers to transmit the buffer smoothly over SPI, but didn't actually submit a PR for it.

> > > (Larger frame-time numbers in the original write-up include a separate
> > > async-double-buffer change in the platform layer that's not part of this PR.)
> > 
> > 
> > Even though ESP32 is not "natively" supported by raylib, I think it would be interesting to consider how to implement the necessary components for this in rlsw.
> 
> This is how we solved it in the [esp32 repo](https://github.com/jensroth-git/raylib-esp32).
> 
> Double-buffer the color framebuffer and run the display transfer on a dedicated FreeRTOS task pinned to core 0, overlapping it with rendering on core 1.
> 
> **rlsw.h changes:**
> 
> `sw_default_framebuffer_t` now holds `sw_texture_t color[2] + int colorIndex` `sw_default_framebuffer_alloc/free` manage both buffers New public function `swSwapColorBuffer()` toggles the render target and returns a pointer to the just-rendered buffer
> 
> **Port layer changes:**
> 
> * `raylib_port_flush_async(src, w, h)` — stores buffer pointer, notifies
>   flush task, returns immediately
> * `raylib_port_wait_flush()` — blocks until the previous transfer completes
> * Flush task: persistent FreeRTOS task on core 0 (4 KB stack), sleeps via
>   `ulTaskNotifyTake`, runs vflip + byte-swap + DMA, signals completion
> 
> **SwapScreenBuffer flow:**
> 
> ```
> wait_flush()          // ensure previous transfer finished
> rendered = swSwapColorBuffer()  // swap render target, get finished buffer
> flush_async(rendered) // kick off transfer on core 0
> ```
> 
> **Memory cost:** ~115 KB PSRAM (second color buffer) + 4 KB task stack.
> 
> Since render time (~42 ms) > transfer time (~15 ms), the wait never blocks in steady state. Effective swap cost drops to ~30 µs.

This is not exactly how it's done on other repositories.  And for most RLSW users currently (which are mostly Linux users seeking sandboxing and experiments rather than embedded developers) they don't need double buffering at all, which makes it sensible for this to not exist for them.  

Currently, the ESP32 approach uses a methodology that is not optional.  This is why it wasn't added to that PR.

[...]

PR #5994, opened 2026-07-18
[rlsw] First batch of embedded optimizations and compiler requirements Pico 2

NighthowlerStudios:
# Backbuffer for RLSW
This implements the backbuffer approach being talked about in #5993 

The "secret option" `RLSW_BACKBUFFER` has been thoroughly tested and confirmed working, as well as confirmed a serious performance boost, similar findings to the ESP32's approach, though this one has less to it, and a bit more manual control needed by the embedded developer to take hold of the buffer and transmit it themselves at the right time.

For example, the developer is still required to add in their own mutex methodology to prevent their video transmission from overlapping with the new buffer pointer submission.

# Use pico_rand.h to get true random seed on startup, fixes compiler error
The Pico 2 does not include a hardware clock that measures real time.  Because of this, the Pico SDK only includes an incomplete stub of `time.h`.  When compiling to Pico 2, if left alone, `time(NULL)` will actually throw `unsupported relocation` on GCC, as it cannot link everything inside of `time()` needed.

With the Pico 2 we have an alternate approach available.  The RP2350 has a hardware "True Random Unit."  This true random unit listens to electrical noise surrounding the chip (with Ring Oscillators) and uses that to seed itself.  This would be much unlike getting "time since startup".

So this commit suggests using the hardware True Random Unit to get the first seed.  Then we can let Raylib take control of the random maths from there, to inline the behaviour with how Raylib runs on other platforms by keeping Raylib's actual algorithm active.

Note that the first fetch costs about 20,000 cycles as the ring oscillators need time to spin up.  But this is in trade-off for getting the seed completely randomized instead of predictable (start up time each reboot would be about the same give or take a few microseconds).  It doesn't make sense for us to use the hardware randomizer to replace Raylib's Xoshiro implementation to fetch the actual values, because the hardware TRNG has a limitation where the seventh read will always trigger a recycle of the oscillators to refresh the values, the previously mentioned 20,000 cycle penalty.  Way too heavy compared to Xoshiro's average of 10 to 20 cycles. 

In short, TRNG once to seed, then existing Xoshiro for fastest random streaming.


src/external/rlsw.h
```diff
@@ -677,6 +677,9 @@ SWAPI bool swResize(int w, int h);
 SWAPI void swReadPixels(int x, int y, int w, int h, SWformat format, SWtype type, void *pixels);
 SWAPI void swBlitPixels(int xDst, int yDst, int wDst, int hDst, int xSrc, int ySrc, int wSrc, int hSrc, SWformat format, SWtype type, void *pixels);
 SWAPI void *swGetColorBuffer(int *width, int *height); // Restored for ESP-IDF compatibility
+#ifdef RLSW_BACKBUFFER
+SWAPI void swSwapColorBuffers(void); // Swap software renderer color/back buffers by pointer only
```
Bigfoot71, line 680: It would be better to keep this function declared and make it a no-op when `RLSW_BACKBUFFER` is disabled.

Also, I think @raysan5 might appreciate a warning log if the function is called while the feature is not enabled.

src/external/rlsw.h
```diff
@@ -1061,6 +1067,9 @@ typedef struct {
 
     sw_handle_t boundFramebufferId;                             // Framebuffer currently bound
     sw_texture_t *colorBuffer;                                  // Color buffer currently bound
+#ifdef RLSW_BACKBUFFER
+    sw_texture_t *colorBackBuffer;
+#endif
```
Bigfoot71: This member appears to be assigned but never actually used.
I think `backBuffer` should only exist in the default framebuffer structure.
The context pointers should simply point to the buffers currently being written to.

Bigfoot71 (changes requested):

Bigfoot71:
As for the changes to `rlsw`, the addition looks good to me.

@raysan5 Please, if this PR is going to be merged, merge it before my other PR #5996 in case there are any conflicts to resolve.

Especially since I asked a question [there](https://github.com/raysan5/raylib/pull/5996#issuecomment-5015808759) about the consistency of the configuration define prefixes, and `RLSW_BACKBUFFER` is being added here:

src/external/rlsw.h
NighthowlerStudios: adjusted with 92ab11901322f3212bf25035d496c30926bf4a6f

src/external/rlsw.h
NighthowlerStudios: adjusted with 661b1c102407e79493b7b47fe32b8ac425202b33

src/external/rlsw.h
NighthowlerStudios: @raysan5 if that `SW_LOG` is okay with you I can resolve this.

raysan5:
@Bigfoot71 thanks for the review, it looks good to me. I can merge it before the following PR.

@NighthowlerStudios Please, could you rename the flag to `SW_DOUBLE_BUFFERING`?

NighthowlerStudios:
> @NighthowlerStudios Please, could you rename the flag to `SW_DOUBLE_BUFFERING`?

@raysan5 Sure thing.  https://github.com/raysan5/raylib/pull/5994/commits/c64ed578751ef21bdb03e1f5a95ccd0058c1298d

raysan5:
@NighthowlerStudios thanks for the improvement! Also answered on the issue.

PR #5997, opened 2026-07-19
Fix DrawSphereWires

oscujic:
In addition to significantly simplifying the code, the following bugs were fixed:

- Top "cap" of the sphere was double-covered
- Sphere had an additional latitude division than specified


raysan5:
@oscujic no plans to changw function signature, not following raylib conventions

PR #6000, opened 2026-07-20
[rmodels] `DrawSphereWires()`, code optimization

oscujic:
Hello Ray! Hopefully this is better. 👍

Summary:
- Eliminated some redundancies
- Fixed a bug(?) where the sphere would have 1 more "ring" (latitude division) than expected
- Fixed a bug where the top ring was double-covered
- Added comments denoting which code draws which lines

### DIFF
--- a/src/rmodels.c
+++ b/src/rmodels.c
@@ -542,30 +542,31 @@ void DrawSphereWires(Vector3 centerPos, float radius, int rings, int slices, Col
         rlBegin(RL_LINES);
             rlColor4ub(color.r, color.g, color.b, color.a);
 
-            for (int i = 0; i < (rings + 2); i++)
+            for (int i = 0; i < rings; i++)
             {
                 for (int j = 0; j < slices; j++)
                 {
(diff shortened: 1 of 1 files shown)


raysan5:
@oscujic Are the generated lines aligned with triangles generated with `DrawSphereEx()`? Please note that this is part of an API, so it should be consistent with the other functions.

Please, could you provide some screenshots of the issue it is trying to solve and the result of this implementation?

oscujic:
DrawSphereEx also had the +1 ring quirk, so I fixed it to be the same as the new DrawSphereWires.
Also, I replaced the algorithm of DrawSphereWires with the one used in DrawSphereEx, so it should be much faster.

Here are some spheres drawn with those changes made (left is old, right is new):
Screenshot 2026-07-20 094108
Both spheres were specified with 6 rings and 12 slices, however you can see the old one has an additional latitude division compared to the new one.
I also gave the lines some transparency so you can see how the top ring on the old one is double covered by the fact that it appears more opaque.

oscujic:
After testing, the new algorithm is indeed much faster.
Old algo: 60 wire spheres drawn before fps drops below 60.
New algo: 700 wire spheres drawn before fps drops below 60.
(used 8 rings and 16 slices in both cases)

raysan5:
@oscujic thanks for the improvement

PR #6005, opened 2026-07-22
[build][Zig] Update to raygui 5.0

jn-jairo:
Update Zig build dependency to raygui 5.0

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -12,8 +12,8 @@
             .lazy = true,
         },
         .raygui = .{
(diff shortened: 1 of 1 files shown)


raysan5:
@jn-jairo Updated! But note that raylib 5.0 had a bug with GuiColorPicker(), it should probably point to master...

jn-jairo:
> @jn-jairo Updated! But note that raylib 5.0 had a bug with GuiColorPicker(), it should probably point to master...

It's pointed to the latest commit in master.

PR #6006, opened 2026-07-23
[build][Zig] Fix web examples building all the time

jn-jairo:
Similar to #5869 but for the web target, it was running the `emcc_step` for each example even when requesting to just build raylib.


raysan5:
@jn-jairo thanks for the review

PR #6007, opened 2026-07-23
[rmodels] Fix GLTF animation interpolation

lassade:
Hi

In the wild some animations use different sampler interpolation (linear/step) for differents channels (translate/rotate/scale) of the same bone, but raylib defaults to a single interpolation type based on the last channel found, so if we get translate (linear) rotate (step) scale(step) the translation will be sampled using the step interpolator.

This patch simples use the channel sampler interpolation fixing the issue


raysan5:
@lassade Thanks for the review!

PR #5949, opened 2026-07-03
[rtext] Adding support for Em Box font size metric

ObedH:
I added the function `void SetFontSizeMetric(FontSizeMetric metric)` to allow users to select which font size metric to use for loading fonts.
Fixes #5678. The user can now call `SetFontSizeMetric(FONT_SIZE_METRIC_EM_BOX)` before calling `LoadFontEx()` when loading pixel fonts that are designed around the em box.
This function modifies a global static variable in `rtext.c` named `fontSizeMetric`. It will cause subsequent calls to the following functions to calculate the font scale differently:
+ `LoadFontFromMemory`
+ `LoadFontEx`

`LoadFontData()` now calculates the font scale depending on the selected `fontSizeMetric`.
+ `FONT_SIZE_METRIC_EM_BOX`: uses `stbtt_ScaleForMappingEmToPixels`
+ `FONT_SIZE_METRIC_ASCENT`: uses `stbtt_ScaleForPixelHeight()`

`FONT_SIZE_METRIC_ASCENT` remains the default option, so as to preserve existing behavior. Existing applications are unaffected  unless the user explicitly selects `FONT_SIZE_METRIC_EM_BOX`.
I created `examples/text/text_font_size_metric.c` to test this functionality, which displays three pixel fonts which render correctly when `FONT_SIZE_METRIC_EM_BOX` is selected, and three vector fonts which render correctly with the default settings.

text_font_size_metric

PR #6008, opened 2026-07-23
[rtext] Add support for Em Box font size metric (Supersedes #5949)

### LINKED ISSUE #5678: [rtext] Incorrect rendering of pixel font
### Issue description

Hello! I'm using the m5x7 pixel font(https://managore.itch.io/m5x7) in my game, with a 480x270 render target (point texture filter) that is rendered then to the screen (960,540), it currently looks like this (using point texture filter as well):
Image

I noticed that it is rendering fine in [Foster framework](https://github.com/FosterFramework/Foster) and they are also using stb_truetype, but instead of 
`stbtt_ScaleForPixelHeight` https://github.com/raysan5/raylib/blob/8a685877eb321fe0eaeabab7b8ec69d2f8a0064d/src/rtext.c#L638
they are using:
`stbtt_ScaleForMappingEmToPixels` https://github.com/FosterFramework/Foster/blob/0b814e1ee312d216008c1fd2291f0e0b12295c91/Framework/Internal/ThirdParty/StbTrueType.cs#L48

Compiled latest raylib with this change and looks like the output is correct now:
Image

I'm not sure if this is a breaking change, tried the `text_font_filters` example with the change and seems to be ok aswell, but not sure yet, currently going through the stb_truetype docs to find out more info https://github.com/nothings/stb/blob/28d546d5eb77d4585506a20480f4de2e706dff4c/stb_truetype.h#L766

Also this issue might be relevant: https://github.com/nothings/stb/issues/689

Found similar issues: https://www.reddit.com/r/raylib/comments/1g0pwr5/pixel_font_blurriness_with_specific_fonts/
https://www.reddit.com/r/raylib/comments/1g0cbny/pixel_font_distortion/
### Environment

Windows 11

thiagomucci:
Fixes #5678. Supersedes #5949 with C99 coding conventions, TRACELOG parameter fix, and open-source font assets.


src/raylib.h
raysan5, line 1163: Please, review those lines, they were updated in master

raysan5:
@thiagomucci It seems required font assets and its licenses info are not included

thiagomucci:
Updated text_font_size_metric.c to use standard raylib font assets comparing EM_BOX vs ASCENT metrics side-by-side. Unlicensed font assets from PR #5949 have been removed

raysan5:
@thiagomucci oh! great! thanks for the review! could you add the .png screenshots por the example? It can be captured with F12 while running the example.

raysan5:
I'm reviewing this change but I got some concerns, raylib expects to use font size in pixels for the font loading but also for all the text drawing functions. `SetFontSizeMetric()` could confuse library users to switch to `EM_BOX` to all font size definitions, including text drawing, creating undesired results.

One solution could be checking it on all text drawing functions but I think it will be even more confusing... Not sure about this change...

raysan5:
I decided not to merge this change. raylib uses pixels for font size, it's up to the user to do the conversion to follow that convention instead of trying to provide mechanisms to support all use cases.

thiagomucci:
its ok, this pr, and the other are a bit messy

PR #6011, opened 2026-07-23
Update rcore.c

LPBeaulieu:
Added #include <ctype.h> to rcore.c, which is required for isalpha()


raysan5:
@LPBeaulieu redesigned the function to avoid `isalpha()`

PR #6013, opened 2026-07-23
[build][Zig] Improve project example

jn-jairo:
The current code is not wrong, but it is not the Zig way of doing it, and as raylib aims to help people who are learning, it's better if the example project reflects what will be expected from someone working with Zig.

In Zig the de-initialization comes right after the initialization using [defer](https://ziglang.org/documentation/0.16.0/#defer) to execute the expression unconditionally at scope exit.

With this pattern it's less likely that one would forget the de-initialization as it's easier to check if the matching de-initialization is missing and also ensures it will run even if an error occurs in this scope.

### DIFF
--- a/projects/Zig/src/core_basic_window.zig
+++ b/projects/Zig/src/core_basic_window.zig
@@ -31,6 +31,7 @@ pub fn main() void {
     // Initialization
     //--------------------------------------------------------------------------------------
     rl.InitWindow(screenWidth, screenHeight, "raylib [core] example - basic window");
+    defer rl.CloseWindow(); // Close window and OpenGL context
 
     if (builtin.os.tag == .emscripten) {
         std.os.emscripten.emscripten_set_main_loop(UpdateDrawFrame, 0, 1);
@@ -44,11 +45,6 @@ pub fn main() void {
             UpdateDrawFrame();
         }
     }
-
-    // De-Initialization
-    //--------------------------------------------------------------------------------------
-    rl.CloseWindow(); // Close window and OpenGL context
-    //--------------------------------------------------------------------------------------
 }
 
 //----------------------------------------------------------------------------------
@@ -63,11 +59,10 @@ fn UpdateDrawFrame() callconv(.c) void {
     // Draw
     //----------------------------------------------------------------------------------
     rl.BeginDrawing();
+    defer rl.EndDrawing();
 
     rl.ClearBackground(rl.RAYWHITE);
 
     rl.DrawText("Congrats! You created your first window!", 190, 200, 20, rl.LIGHTGRAY);
-
-    rl.EndDrawing();
     //----------------------------------------------------------------------------------
 }


raysan5:
@jn-jairo I'd prefer to keep both solutions, properly comented

jn-jairo:
I added both solutions with comments.

Let me know if I should change anything.

projects/Zig/src/core_basic_window.zig
```diff
@@ -63,11 +72,15 @@ fn UpdateDrawFrame() callconv(.c) void {
     // Draw
     //----------------------------------------------------------------------------------
     rl.BeginDrawing();
+    // End drawing using defer to execute the expression unconditionally at scope exit
+    // See: https://ziglang.org/documentation/master/#defer
+    defer rl.EndDrawing();
```
raysan5, line 62: I think this can lead to future issues, imagine this situation:
```
rl.BeginTextureMode();
defer rl.EndTextureMode();

rl.BeginDrawing();
defer.rlEndDrawing();
// Using RenderTexture
```

Same issue with most `rl.Begin*()`/`rl.End*()` functions, many users can have problems if using defer on those cases, including `InitWindow()`/`CloseWindow()` in cases you need to create one window after another.

raysan5:
@jn-jairo Added some comments that make the use of defer problematic for users on some situations, I think it's better to keep the example aligned with C to avoid those situations.

jn-jairo:
Stupid github just made me lost the long message I spent a lot of time writing... I hate microsoft.

Right now, I don't have the time or patience to rewrite it. I'll send you a reply regarding your concerns later.

jn-jairo:
Ok lets try it again, now I'm writing it on a file in my computer so github doesn't screw me again.

Looks like I failed to explain how it works properly, so let's try it again with a longer message.

- Defer expressions are evaluated in reverse order
- Error in Zig are just special enums you can return like `return error.MyErrorName`
- Scope in Zig is not just function scope like C, in Zig any block `{ }` is a scope, which includes things like `while` and `if` blocks
- Scope exit is when the execution leaves the scope by any means, not just by reaching the end of the scope, like `return` `continue` `break`

So let's look this example:

Note: I made this example with `GuiButton` but I found a bug that destroys the static texts you pass to it, making it blank when you run the function a second time. Probably caused by `GuiDrawText` as `DrawText` works fine.

```zig
const std = @import("std");
const rl = @import("raylib");

fn selectOption() !i32 {
    rl.InitWindow(800, 600, "raylib");
    defer rl.CloseWindow();

    rl.SetTargetFPS(60);

    while (!rl.WindowShouldClose()) {
        rl.BeginDrawing();
        defer rl.EndDrawing();

        rl.ClearBackground(rl.WHITE);

        rl.DrawText("[O]k or [E]rror", 10, 10, 30, rl.BROWN);

        if (rl.IsKeyPressed(rl.KEY_O)) {
            return 1;
        }

        if (rl.IsKeyPressed(rl.KEY_E)) {
            return error.MyError;
        }

        // rl.EndDrawing();
    }

    // rl.CloseWindow();

    return 0;
}

pub fn main() !void {
    for (0..4) |_| {
        const option: i32 = selectOption() catch -1;
        const option_text = switch (option) {
            0 => "ESC",
            1 => "Ok",
            -1 => "Error",
            else => unreachable,
        };
        std.debug.print("...... Option: {} - {s}\n", .{option, option_text});
    }
}
```

I recorded a video running this code with both Zig and C style to show the difference:

https://github.com/user-attachments/assets/0ae2310a-217b-4436-9ae1-eb7b77da6a0b

As you can see, with `defer` the `CloseWindow` is executed no matter what happens, but using the C style, the `CloseWindow` is only executed when pressing `ESC` to leave the function by executing until the end of it and reaching the `CloseWindow` call.

In this example I returned the error directly but in a normal case it would be some function that can return a error like:

```zig
fun myFunction() !void {
    // ...

    if (something) {
        return error.MyError;
    }

    // ...
}
```

And inside the `while` loop it would be executed like `try myFunction();` that is a short for `myFunction() catch |err| return err;` and that is the normal way of coding in Zig.

As you can see `defer` is a essential part of Zig, that prevents bugs and makes the life of the programmer easier.

---

Another point, I found that the `raylib-zig` has this same [example](https://github.com/raylib-zig/raylib-zig/blob/devel/examples/core/basic_window_web.zig) and it's just like my code.

You may have noticed that I removed the last commit that added the comments, I did it because this commit was an error.

Basic features of a programming language should not be explained with comments like I did, and bad code, like not using defer, should not be shown as an option.

The correct code is like it was before, plain and simple, that is why I removed the commit.

Other uses of `defer` with different drawing modes can be seen at `raylib-zig` like the [3d_camera_free](https://github.com/raylib-zig/raylib-zig/blob/devel/examples/core/3d_camera_free.zig) where `{ }` is used to create a new block.

---

When I made this PR I expected it to be quick, but somehow it took a lot of our time.

I'm sorry I made you spend so much time on this.

So, it's for the best that we resolve this without me taking more of your time.

My final proposal is for the code as I made originally, which is the current code. For the reason explained above I think it's the best way.

Therefore, feel free to close this PR if you have a different opinion, or to accept it if I have convinced you.

jn-jairo:
[More examples of defer in raylib-zig](https://github.com/search?q=repo%3Araylib-zig%2Fraylib-zig%20defer&type=code)

raysan5:
I decided to avoid this change. Zig users already know about the defer possibility, keeping this example aligned with raylib example avoids potential issues.

PR #6014, opened 2026-07-24
[rshapes] Correct `DrawPolyLinesEx()` line thickness.

MatthewRoush:
Follow up to #6004. The previous formula would produce inconsistent results for different polygons with the same `thick` parameter.


raysan5:
@MatthewRoush thanks for the review! BTW, the example looks really good, maybe it can be added as a new example

MatthewRoush:
Thanks for liking the example! I'll submit a PR with an example that's more consistent with the example template.

PR #6015, opened 2026-07-24
[rcore][GLFW] Fix window landing off-screen when it's bigger than the monitor workarea

### LINKED ISSUE #6002: [rcore][GLFW] Bad window placement with portrait displays
Hi, just tried updating from raylib 5.5 to 6.0 and ran into a bug only happening with 6.0.
If I have screen 3 as Windows' main display, InitWindow() will just flash a window and then it disappears.
It shows in the taskbar with a full white screenshot/preview, but my game is mostly black. It seems to popup outside the screens.
If i shift+win+arrow left/right maybe 5 times, i can get it to suddenly show up on one of the screens and with correct graphics, but window dimensions far off what i had picked. mine were like 16:9 but the window shows up at maybe 9:16 it seems.
My screen 3 is in portait mode.
If I make screen 2 (or 1) my main display, everything works as it did with raylib 5.5.
If I run a SetWindowPosition(0,0); right after InitWindow(), the window shows up with right dimensions correctly, also like it was in 5.5. Even on screen 3. So that's a workaround for the broken default behavior.
Windows 10, latest updates, screens on the side are 2560x1440 and the portrait one is 1080x1920. it's a normal 1920x1080 screen i've rotated, so the setting in Windows is "Portrait" to rotate it 90deg.

Image

Tested with both 6.0 master branch as of today https://github.com/raysan5/raylib/commit/1c3a7748ea665a35b881f991783a998e385a0e33 and 6.0 source zip.
i used the vs2022 sln. did a debug static lib build.

Gooh456:
Fixes #6002.

`InitPlatform()` centers the window with `monitorX + (monitorWidth - width)/2`. There's no check for the window being wider or taller than the monitor. When it is, that goes negative and the window gets placed outside the virtual desktop. On Windows this isn't just "off-center", `glfwSetWindowPos`/`GetWindowPosition` come back with garbage coordinates and the window can stay invisible until you drag it back with win+shift+arrow, which matches what's described in #6002 (portrait primary monitor, landscape window, window vanishes, `SetWindowPosition(0,0)` right after `InitWindow()` "fixes" it).

`SetWindowMonitor()` already handles this correctly, it anchors to the workarea origin instead of centering when the window doesn't fit. This PR just applies the same guard in `InitPlatform()`.

How I checked it: wrote a small repro that calls `InitWindow()` with a width bigger than my monitor and reads `GetWindowPosition()` right after.

Before:
```
RESULT window position: x=2147483392 y=66
```
After:
```
RESULT window position: x=0 y=0
```

Also checked a normal window size still centers the same as before (800x450 on a 1920x1032 workarea gives x=560 y=291, matches the math).

I don't have a real portrait-monitor multi-display setup to reproduce the exact report in #6002, only a single landscape monitor here, so I forced the same code path by requesting a window wider than my monitor instead. Same math, same branch, just couldn't test the literal portrait-monitor case.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -1791,12 +1791,28 @@ int InitPlatform(void)
 
         // Center window into current monitor
     #if defined(__APPLE__)
-        CORE.Window.position.x = monitorX + (monitorWidth - CORE.Window.screen.width)/2;
-        CORE.Window.position.y = monitorY + (monitorHeight - CORE.Window.screen.height)/2;
+        const int windowWidthForCentering = CORE.Window.screen.width;
+        const int windowHeightForCentering = CORE.Window.screen.height;
     #else
-        CORE.Window.position.x = monitorX + (monitorWidth - CORE.Window.render.width)/2;
-        CORE.Window.position.y = monitorY + (monitorHeight - CORE.Window.render.height)/2;
+        const int windowWidthForCentering = CORE.Window.render.width;
+        const int windowHeightForCentering = CORE.Window.render.height;
     #endif
+
+        // NOTE: If the window is as large or larger than the monitor workarea on either axis
+        // (e.g. a landscape window on a portrait primary monitor), centering it can push part
+        // of it past the edge of the virtual desktop, which some window managers (Windows in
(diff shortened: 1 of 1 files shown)


raysan5:
This is a small detail, I'm not making the code more complicated.

PR #6016, opened 2026-07-25
[rtextures] Replaces `rlUnloadTexture` in UnloadRenderTexture with `UnloadTexture`

lamweilun:
Replaces `rlUnloadTexture` in UnloadRenderTexture with `UnloadTexture`.

- Fixes a missing log when calling UnloadRenderTexture.

### DIFF
--- a/src/rtextures.c
+++ b/src/rtextures.c
@@ -4470,11 +4470,7 @@ void UnloadRenderTexture(RenderTexture2D target)
 {
     if (target.id > 0)
     {
-        if (target.texture.id > 0)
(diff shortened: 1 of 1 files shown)


raysan5:
> Fixes a missing log when calling UnloadRenderTexture.

Actually, that was the reason why `rlUnloadTexture` is used instead of `UnloadTexture`. There could be some cases that a `RenderTexture` is recreated every frame and consequently log is spammed with the unload attachment, so I solved that issue with current implementation.

PR #6018, opened 2026-07-25
[rlgl] Add render batch utility functions to control the current depth

rgebee:
Allows existing draw functions to be drawn out of order with different depths. I can set the depth when making custom draw functions using `rlVertex3f` but it made more sense to solve from rlgl rather than reimplement every draw function I want to use for this. I also looked at `rlTranslatef` but that adds additional matrix multiplication that is not needed here and it also applies on top of the current depth which I would need the value of to offset it.

The use case is drawing a variety of entities in a game in an isometric grid where entities may appear behind buildings, other entities etc. They use a few different draw functions and while entities are currently sorted to draw in order, there are a large number of them so I would like to draw them out of order using depth testing to see if it improves the performance.

The main change is a utility function to get the active render batch to match the existing `rlSetRenderBatchActive`. This allows access to the default render batch that is created which allows me to change the depth and debug the render batch state.

I also added utility functions specifically for getting/setting the depth value. Wasn't sure if that would be useful to have as well.

Changes:
* Add `rlGetRenderBatchActive` to match `rlSetRenderBatchActive`.
* Add `rlGetRenderBatchDepth` to get the depth for the active render batch
* Add `rlSetRenderBatchDepth` to set the depth for the active render batch

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -730,9 +730,13 @@ RLAPI int *rlGetShaderLocsDefault(void);                // Get default shader lo
 RLAPI rlRenderBatch rlLoadRenderBatch(int numBuffers, int bufferElements); // Load a render batch system
 RLAPI void rlUnloadRenderBatch(rlRenderBatch batch);    // Unload render batch system
 RLAPI void rlDrawRenderBatch(rlRenderBatch *batch);     // Draw render batch data (Update->Draw->Reset)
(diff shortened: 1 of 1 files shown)


raysan5:
@rgebee I'm afraid I prefer to avoid adding those functions, they modify internal depth and can open the door to future issues. I recommend doing the sorting at user side.

PR #6021, opened 2026-07-25
Bug fix in text insertion logic in rtext.c

SewerynKaminski:
Fixed nasty bug, text after insertion was deleted.


raysan5:
@SewerynKaminski ouch! good catch! thanks for the review!

PR #6022, opened 2026-07-26
Bug fix 2 in text insertion logic in rtext.c

SewerynKaminski:
Fixed nasty bug, text after insertion is deleted


raysan5:
@SewerynKaminski thansk for the review

PR #6026, opened 2026-07-28
[rcore][GLFW] Fix `ToggleBorderlessWindowed()` entering exclusive fullscreen instead of windowed

### LINKED ISSUE #5854: [rcore] Unable to alt-tab/switch windows when `fullscreen` or `borderless-windowed` modes are enabled

tillua467:
Fixes #5854 

`glfwSetWindowMonitor()` was called with a non NULL monitor handle when entering borderless windowed mode. Per GLFW's documented behavior, a non NULL monitor negotiates exclusive fullscreen, which is the same call `ToggleFullscreen()` makes not what borderless windowed is supposed to do. The windowed restore branch of this same function already calls `glfwSetWindowMonitor()` with NULL for this reason, this makes the enter branch consistent with it.

Checked on Windows 11 works as intended, although it's not a issue in linux i did checked on debian gnome and wslg, no issue there as well.

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -287,7 +287,7 @@ void ToggleBorderlessWindowed(void)
                 CORE.Window.screen.height = mode->height;
 
                 // Set screen position and size
-                glfwSetWindowMonitor(platform.handle, monitors[monitor], CORE.Window.position.x, CORE.Window.position.y,
+                glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y,
                     CORE.Window.screen.width, CORE.Window.screen.height, mode->refreshRate);
 
                 // Refocus window
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -288,7 +288,7 @@ void ToggleBorderlessWindowed(void)
 
                 // Set screen position and size
                 glfwSetWindowMonitor(platform.handle, NULL, CORE.Window.position.x, CORE.Window.position.y,
-                    CORE.Window.screen.width, CORE.Window.screen.height, mode->refreshRate);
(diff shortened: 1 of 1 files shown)


raysan5:
@tillua467 thanks for the review, all `windowing` reviews are very sensitive and platform dependant. Could you test on macOS? Did you test it with multiple monitors?

tillua467:
@raysan5 hello thanks for looking at my pr! i did tested on multiple monitors and tested and it works fine! the windows did minimize at the monitor it was opened, and also sadly i don't own a mac device so i can't really test that sorry

tillua467:
@raysan5 i found a weird bug after my commit, it seems when the window is minimized it doesn't minimized to the original, i happens because we call `glfwSetWindowAttrib()` and it changes CORE.Window.screen i am not sure why but here is my way of fixing it:
https://github.com/tillua467/raylib/commit/6727062944707df18d6c37ecffeb2623cb54fe75

Haven't tested it in my Debian I will test tomorrow

raysan5:
this issue should be already addressed by another PR

PR #6027, opened 2026-07-29
[rshapes] Fix `DrawRectangleLinesEx()`

MatthewRoush:
Previously `DrawRectangleLinesEx()` would draw lines overlapping each other when they were thick enough (which was only noticeable if the color's alpha was lower than 255.)


raysan5:
@MatthewRoush thansk for the  review. Note I'm reviewing it to avoid the early return, in raylib I try to avoid/minimize early-retruns when possible

MatthewRoush:
Good to know, I was unaware of the early return policy.

PR #6028, opened 2026-07-29
[rlsw] Internal functions declarations review

Bigfoot71:
Rework the internal functions declarations by omitting `inline` entirely and relying solely on `SW_INLINE` to force inlining only on specific cases.

`SW_INLINE` is only used for trivial functions, short generated functions called from a single location, and simple overloads. Some short functions are left to the compiler to inline based on the call site.

Some other trivial functions are intentionally not force-inlined, as they are primarily called through function pointers.

As originally suggested here: #5996


raysan5:

PR #6029, opened 2026-07-29
[rlsw] Rename public compile-time option prefix to `RLSW_`

Bigfoot71:
Rename the prefix of all public compile-time options from `SW_` to `RLSW_`, as suggested in #5996.

Previously, both prefixes were mixed, `RLSW_` for `RLSW_USE_SIMD_INTRINSICS` and `SW_` for the rest.

I think using `RLSW_` for compile-time options while reserving `SW_` for internal macros and definitions provides a clearer distinction.

### DIFF
--- a/src/external/rlsw.h
+++ b/src/external/rlsw.h
@@ -51,13 +51,13 @@
 *       rlsw capabilities could be customized defining some internal
 *       values before library inclusion (default values listed):
 *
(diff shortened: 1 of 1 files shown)


raysan5:
@Bigfoot71 sounds good to me! thanks for the review!

PR #6031, opened 2026-07-30
[raudio][W.I.P] Separate audio devices from context, support other types other than playback

AllMeatball:
### Reason
I have decided since miniaudio (what raylib uses under the hood) can use (and enumerate) separate audio devices for initialization, raylib should implement these too. 

A lot of games made today have sound configuration menus for input (capture) and output (playback) devices.

This is a draft due not having all of the features I want (some might be added already) (see _**Feature list**_)

### What it does
Currently it just lists out sound devices, no option for individual device setup at the moment.

#### Functions
- rcore: `StringDuplicate`
- raudio: `GetAudioDeviceList`, `GetAudioDeviceList`
#### Config Flags
- `SUPPORT_POSIX_STRDUP` (is 1 by default and may want to be change for compat reasons)
#### Types
- raudio (internal): `rAudioDeviceID`
- raudio (external): `AudioDevice`, `AudioDeviceList`, `AudioDeviceType`

**NOTE: the `StringDuplicate` function is used for copying the audio device names. if you want me to remove this or make it private to raylibs code, I am perfectly willing to do so.** 

### Feature list

- [x]  * Ability to list devices
- [ ]  * Individual/specific device setup (separating miniaudio context from device)
- [ ]  * Capture device / microphone support

### Stuff to improve

- I think that `rAudioDeviceID` is a _bit_ _**too**_ abstracted, a string format like "PulseAudio:hex" could be a better system
- `InitAudioDevice` may change into the following

- before:
```c
// ... init window and what not
InitAudioDevice();
// ... close everything
CloseAudioDevice();
```
- after:
```c
// ... init window and what not
InitAudioContext();

#ifdef USE_DEFAULT_AUDIO_DEVICE
SetAudioDevice(NULL); /* picks the default audio device */
#else
AudioDevice device = {};
/* get specific audio device from `GetAudioDeviceList` */
SetAudioDevice(&device);
#endif

// ... close everything
CloseAudioContext();
```

Let me know if there's anything here that I need to fix or change.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -123,6 +123,10 @@
     //  - SUPPORT_FILEFORMAT_JPG
     #define SUPPORT_CLIPBOARD_IMAGE         1
 #endif
+#ifndef SUPPORT_POSIX_STRDUP
+    // Use posix strdup or built-in one
+    #define SUPPORT_POSIX_STRDUP            1
+#endif
 
 // rcore: Configuration values
 // NOTE: Below values are already defined inside [rcore.c] so there is no need to be
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -47,6 +47,9 @@
 *           - Mixing channels support
 *           - Raw audio context support
 *
+*       Levi Spencer (github: @AllMeatball) (July. 2026)
+*           - Audio device listing (playback and capture)
+*
 *
 *   LICENSE: zlib/libpng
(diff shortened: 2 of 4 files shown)


raysan5:
@AllMeatball thanks but no plans to follow this route.

PR #6032, opened 2026-07-31
[examples] Add core_scissor_shape

davidbuzatto:
Adds a raylib example showing how to clip drawing to an arbitrary shape (circle, polygon, ...) instead of just an axis-aligned rectangle, using the stencil buffer via a couple of raw GL calls (rlgl has no stencil wrapper).

As suggested by @raysan5 in #2964, this is proposed as an example rather than new core/rlgl API.


raysan5:
@davidbuzatto I prefer to keep examples using direct OpenGL calls as others, to minimize potential issues with build systems and multi-platform. Please, could you rename it to `gl_scissor_shapes` and move to `others` directory?

raysan5:
@davidbuzatto thanks for the addition, note that name could be further changed and example moved somewhere else, I'm trying to keep examples minimal and focused for maintenance reasons so `others` could be moved to `raylib-extras` repo at some moment.

PR #6039, opened 2026-08-06
[raudio] Fix crash on CloseAudioDevice: uninit device before mutex

get-charlie:
On Android (AAudio) the callback thread could still run when the mutex was destroyed, causing a SIGABRT (pthread_mutex_lock on a destroyed mutex). Uninit the device first so its callback thread stops before ma_mutex_uninit.

### DIFF
--- a/src/raudio.c
+++ b/src/raudio.c
@@ -533,8 +533,9 @@ void CloseAudioDevice(void)
 {
     if (AUDIO.System.isReady)
     {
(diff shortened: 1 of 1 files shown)


raysan5:
@mackron Please, could you take a look to this PR?

mackron:
This change is good. Recommend merging.

raysan5:
@get-charlie thanks for the improvement and thanks @mackron for the quick review

PR #6040, opened 2026-08-06
[rcore][Android] Fix Android relaunch hang after destroy while paused

get-charlie:
When the activity is destroyed while the app is paused, PollInputEvents kept looping after setting CORE.Window.shouldClose: while paused, ALooper_pollAll is called with an infinite timeout (-1), so on the next iteration it blocked forever waiting for an event that never arrives. main() never returned, the native thread stayed alive, and the following launch hung inheriting that thread.


raysan5:
@get-charlie thanks for the review!

PR #6041, opened 2026-08-06
[examples] shaders shadowmap rendering camera fix 

JeffM2501:
The camera pos was being sent to the shader before UdpateCamera, so it was using the position from the last frame, now it updates the camera first so it is the current frame.


raysan5:
@JeffM2501 It seems all spaces were replaced by tabs, please, could you review it?

JeffM2501:
> @JeffM2501 It seems all spaces were replaced by tabs, please, could you review it?

Yeah sorry about that, Did the change on a different machine than normal, I'll get it fixed up.

raysan5:

PR #6044, opened 2026-08-09
[rcore][RGFW] Fix window Y position reading X in RGFW move callback

latent-9:
In `RGFW_cb_windowmovefunc`, the stored window Y position reads the window's X coordinate:

```c
CORE.Window.position.x = platform.window->x;
CORE.Window.position.y = platform.window->x;
```

The second line should read `platform.window->y`. This is confirmed by the equivalent assignment later in the same file (window setup), which correctly reads both coordinates:

```c
CORE.Window.position.x = platform.window->x;
CORE.Window.position.y = platform.window->y;
```

and by `GetWindowPosition`, which reads both `platform.window->x` and `platform.window->y`. As-is, handling a window-move event stores the X value into `CORE.Window.position.y`, so the reported Y position is wrong after moving the window.

### DIFF
--- a/src/platforms/rcore_desktop_rgfw.c
+++ b/src/platforms/rcore_desktop_rgfw.c
@@ -613,7 +613,7 @@ static void RGFW_cb_windowmovefunc(const RGFW_event *e)
     if (e->common.win != platform.window) return;
 
     CORE.Window.position.x = platform.window->x;
-    CORE.Window.position.y = platform.window->x;
+    CORE.Window.position.y = platform.window->y;
 }
 static void RGFW_cb_keycharfunc(const RGFW_event *e)
 {


raysan5:
@CrackedPixel, @refinism, @pamarcos, @strangezakary, @Shakai-Dev, @jfarrell303, @jasn-armstrng, @justinhhorner, @heliumsneakers, @Ne0nWinds Please, could you take a look to this PR?

jasn-armstrng:
Verified on (Manjaro) Linux(x11, RGFW backend). Built both the master and this branch. The fix is correct. Line 616 should read `platform.window->y` matching the assignment at line 1843.

Note: However, I could not observe the bug through the public API, seems as tho it's latent: `GetWindowPosition()` queries RGFW live via `RGFW_window_getPosition()` rather than reading the CORE.Window.position, and the only readers of the cached value `ToggleFullScreen` (line 770) and `ToggleBorderlessWindowed` (line 802) overwrite it from `previousPosition` first.

latent-9:
Thanks for building and checking it. Good point that it is latent through the public API since GetWindowPosition reads RGFW live. Still worth keeping the cached value correct for any future reader.

heliumsneakers:
The fix is correct, I was able to verify on MacOS Arm64. The cached window coordinates now read back normally.

heliumsneakers (approved):
This is a valid one line typo fix. Cached window coordinates now read back correctly from the move event handler when updating `CORE.Window.position`.

jfarrell303 (approved):
Correct one line fix. I was also able build and test this locally (CachyOS) and verified that the cached value now reads back correctly.

raysan5:
@latent-9 thanks for the improvement and review!

latent-9:
@raysan5 ywc legend

PR #6045, opened 2026-08-09
[rshapes] Fix drawing circle sectors

MatthewRoush:
Functions that draw a sector of a circle would keep drawing segments past a full rotation, which looked bad.

Example of previous behavior from `DrawCircleSector()` (the brighter red area is where it's overlapping itself due to not stopping at 360 degrees.)
circle_overlapping

Additionally, `DrawCircleSectorLines()` and `DrawRingLines()` drew the cap lines even when drawing a full circle, which seems undesirable (though that could just be a matter of taste.)

With caps
with_caps

Without caps
without_caps

### DIFF
--- a/src/rshapes.c
+++ b/src/rshapes.c
@@ -1576,8 +1576,13 @@ void DrawCircleSectorLines(Vector2 center, float radius, float startAngle, float
         endAngle = tmp;
     }
 
+    bool showCapLines = true;
     // Drawing a whole circle, things get weird without limiting the circle to 360 degrees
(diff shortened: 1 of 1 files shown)


raysan5:
@MatthewRoush thanks for the improvement!

 > Additionally, DrawCircleSectorLines() and DrawRingLines() drew the cap lines even when drawing a full circle, which seems undesirable (though that could just be a matter of taste.)

Definitely, better to remove those extra lines on full circles.

PR #6051, opened 2026-08-10
Update GLFW to 3.5.1

M374LX:
Note: this new GLFW version drops Windows XP and Vista support.

The macOS build was previously failing, but the inclusion of the QuartzCore framework seems to have fixed the problem. As I don't own a Mac, I suggest someone who does to test and ensure everything (including examples) can be built using all supported build systems.


raysan5:
@M374LX thanks for working on this update!

Just for reference, some of the big changes of this update:

```
[Win32] Removed support for Windows XP and Vista (#2505)
[Cocoa] Added QuartzCore framework as link-time dependency
[Cocoa] Removed support for OS X 10.10 Yosemite and earlier (#2506)
```

raysan5:
@M374LX Some of the libraries added could not be required, afaik, they are only intended for examples, they can be removed:

image

raysan5:
@M374LX thanks for the update, it looks good, let's see if something breaks, it's a sensitive change...

PR #6058, opened 2026-08-12
Use unmodified GLFW

M374LX:
Turns out that we can use an unmodified GLFW and thus free raylib from this burden when updating the library.

We only need to temporarily rename four functions using ``#define`` and ``#undef`` and define an empty function named ``_glfwConnectNull()`` in ``rglfw.c``.


raysan5:
@M374LX that's nice, thanks!

PR #6061, opened 2026-08-13
[rcore_web_emscripten] Fix time calculation in GetTime function

andzdroid:
emscripten_get_now returns milliseconds, need to divide not multiply to get seconds.

Also, should I even be using this platform? I'm confused about the web platforms.


raysan5:
@andzdroid wow! good catch! thanks for the review!

This platform backend is functional but incomplete, it removes `libglfw.js` dependency, still needs a bit more work.

PR #6067, opened 2026-08-16
[rcore_web_emscripten] Handle emscripten keyboard events

andzdroid:
Added key press and char press handling to the emscripten platform.

Works on the input keys example and keyboard testbed in firefox, chrome and safari.

The text input example did not work, but that's due to mouse handling not working properly (can't hover over the input).

### DIFF
--- a/src/platforms/rcore_web_emscripten.c
+++ b/src/platforms/rcore_web_emscripten.c
@@ -42,6 +42,7 @@
 **********************************************************************************************/
 
 #include <emscripten/emscripten.h>      // Emscripten functionality for C
+#include <emscripten/dom_pk_codes.h>    // Emscripten physical keyboard codes
 #include <emscripten/html5.h>           // Emscripten HTML5 library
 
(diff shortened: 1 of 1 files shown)


raysan5:
@andzdroid thanks for working on this improvement!

tarmo888:
Out of curiosity, how can `rcore_web_emscripten` even be used for building?
I tried to replace `rcore_web.c` with `rcore_web_emscripten.c` at https://github.com/raysan5/raylib/blob/b98c4e031d3a7c6fab4d62182924489f9e8b7cab/src/rcore.c#L529 and use software rendering. While keyboard and mouse seem to work now, the mouse clicks still go to 0,0.

Is `rcore_web_emscripten` not plugged into anything because it's still half-baked (`rcore_web.c` with software rendering is even more broken than this) or I just don't understand how to use it.

raysan5:
> Is `rcore_web_emscripten` not plugged into anything because it's still half-baked

Yeap, that's it. Once ready, it will replace current `rcore_web` that depends on `libglfw.js` and doesn't support software rendering

PR #6068, opened 2026-08-16
add raylib binding for modula 2 and 3

RobertFlexx:
works for both modula 2 and 3


raysan5:
@RobertFlexx thanks for the addition

PR #6069, opened 2026-08-17
[rAduio] Add user data to Audio Stream callbacks

JeffM2501:
The current audio stream callback does not provide a way for the calling app to pass any user data to the callback that is run by the audio thread. This means that a calling app must have one compile time callback per audio stream since there is no way for it to know what stream the callback goes with.
This PR adds an EX version of the Add Stream Callback function that takes a void* that is passed to the callback, this way users can pass state data to the callback and make it more useful. 

This is a pretty standard way of passing user data into callbacks that the advanced users who are using the callbacks should understand.

This IS a breaking change, all existing callbacks will need a new argument, but it can be ignored. I feel the feature is not used commonly so the added functionality is worth the hassle of making current users perform a small update to the code. This will also help future proof the API.

### DIFF
--- a/examples/audio/audio_stream_callback.c
+++ b/examples/audio/audio_stream_callback.c
@@ -39,10 +39,10 @@ typedef enum {
 //------------------------------------------------------------------------------------
 // Module Functions Declaration
 //------------------------------------------------------------------------------------
-static void SineCallback(void *framesOut, unsigned int frameCount);
-static void SquareCallback(void *framesOut, unsigned int frameCount);
-static void TriangleCallback(void *framesOut, unsigned int frameCount);
-static void SawtoothCallback(void *framesOut, unsigned int frameCount);
+static void SineCallback(void *framesOut, unsigned int frameCount, void* userData);
+static void SquareCallback(void *framesOut, unsigned int frameCount, void* userData);
+static void TriangleCallback(void *framesOut, unsigned int frameCount, void* userData);
+static void SawtoothCallback(void *framesOut, unsigned int frameCount, void* userData);
 
 static int waveFrequency = 440;
 static int newWaveFrequency = 440;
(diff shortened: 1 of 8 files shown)


JeffM2501:
I just realized that I could make this compatible with the current API, with two callback types, but I feel that would be ugly and more confusing. If we are willing to accept this small API breakage I think it is the better path moving forward.
But if you'd like to see one that preserves the existing API, let me know and I can write that up.

Jagholin:
This was my idea.

I'm making a video player/library that uses raylib for UI and sound, and I realized some time ago that playing a single video file at a time works fine, and that's enough for a video player. But if I'm also making a library out of it, then there is no reason why the limitation of "only playing 1 file at a time" should exist, and currently it doesn't (in theory). Except for the `AudioStream`s API. If a user wants to start playing multiple videos, they should connect each instance of a "video player" to the corresponding audio stream, and in my case, I do it via an `AudioCallback`. Now here is the problem: in the current API there is no way for `AudioCallback`s invocation to know, which audio data to send. In order for it to know, it should be able to receive some extra user-specified data(like an ID of the "video player" struct).

I considered using `UpdateAudioStream` for this, and I don't think that's a better solution. I would probably need to place it in a separate thread so that it doesn't hurt main thread's performance, and we'll have an extra unneeded thread context switch for every audio data transfer. Also the rest of my project works better with audio being "pull"ed rather than "push"ed so that's an additional argument against `UpdateAudioStream`.

That's where this PR will greatly help.

Thanks again to @JeffM2501 for a quick response. :1st_place_medal:

raysan5:
@JeffM2501 I prefer to avoid that breaking change at this moment. If at some moment audio API is reviewed, it will be considered.

JeffM2501:
I can build a version that does not break the API, would you prefer that or to wait until breaking changes are more acceptable? I understand the desire to get 6.2 out the door with minimal changes, and this does NOT need to be part of 6.2

raysan5:
@JeffM2501 Let's wait for 6.2 first, there is already too much to process...

JeffM2501:
> @JeffM2501 Let's wait for 6.2 first, there is already too much to process...

Sounds good

PR #6070, opened 2026-08-17
[rcore_web_emscripten] Update emscripten mouse handling

andzdroid:
Emscripten deprecated canvasX/canvasY and don't fill them in anymore.


raysan5:
@andzdroid thanks for the update, did you test it works on scaled canvas?

andzdroid:
Yes, the normalized position is based on the css size of the element, not the canvas size, so it works no matter how the canvas itself is scaled.

PR #6071, opened 2026-08-17
[rcore][Emscripten] Full screen handling

andzdroid:
Fullscreen: doesn't resize framebuffer
Borderless window: resizes framebuffer

The canvas size is kept in sync with the framebuffer size. Framebuffer also adapts to changes to canvas css size.

Tested with standard renderer and software renderer, and regular DPI and high DPI mode.

### DIFF
--- a/src/platforms/rcore_web_emscripten.c
+++ b/src/platforms/rcore_web_emscripten.c
@@ -61,6 +61,8 @@
 //----------------------------------------------------------------------------------
 // Types and Structures Definition
 //----------------------------------------------------------------------------------
+typedef void (*EmscriptenResizeObserverCallback)(void);
+
 typedef struct {
     char canvasId[64];                  // Current canvas id
     EMSCRIPTEN_WEBGL_CONTEXT_HANDLE glContext; // OpenGL context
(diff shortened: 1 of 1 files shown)


raysan5:
@andzdroid This PR should be carefully reviewed, does it maintain same behaviour as `rcore_web`? Did you test it with [`core_highdpi_testbed`](https://github.com/raysan5/raylib/blob/master/examples/core/core_highdpi_testbed.c) example? Note that `FLAG_WINDOW_RESIZABLE` should allow a resizable framebuffer if enabled.

andzdroid:
It doesn't quite behave like rcore_web, I can fix it.

However, imo rcore_web's handling of display sizes is incorrect. It's mixed up render size and screen size, and is taking the CSS window size as the DPR-adjusted value incorrectly. In css, sizes are not dpr-adjusted.

Example of what I'm talking about:
When window.innerWidth is 1200, rcore_web is setting the render size to 1200, then calculating the screen size by dividing by DPR. If DPR is 1.2, then the screen size gets set to 1000.

In this example, the screen size should be 1200 and render size scaled to 1440.

Should I submit a PR to fix this in rcore_web first? Or should we keep this behaviour and I'll update this PR to match?

andzdroid:
I've checked this behaviour on other platforms. Glfw, win32, rgfw all increase the render size by dpi scale, rather than decreasing the screen size. rcore_web is the exception that decreases the screen size instead of increasing the render size.

https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_win32.c#L386
https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_glfw.c#L1740
https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_desktop_rgfw.c#L512

https://github.com/raysan5/raylib/blob/master/src/platforms/rcore_web.c#L1526

PR #6072, opened 2026-08-18
[rshapes] Add some missing `Draw*LinesEx()` functions and new example `shapes_outlines_testbed`

MatthewRoush:
Added functions:
- `DrawEllipseLinesEx()`
- `DrawTriangleLinesEx()`
- `DrawCircleSectorLinesEx()`
- `DrawRingLinesEx()`

`DrawCircleSectorLinesEx()` and `DrawRingLinesEx()` are quite large and complicated (when compared to other shape drawing functions), so they may be outside the scope of raylib.

Also added a new example, `shapes_outlines_testbed`, for testing the previous and newly added functions.


MatthewRoush:
@raysan5 it seems like the two checks that failed aren't related to the changes in this PR. They both failed during "Setup Environment" after six hours, which seems like an unrelated issue. Rerunning the checks should make them pass. If there's something I'm missing though, please let me know.

raysan5:
@MatthewRoush yeah, I know, I checked it. Just thinking about if adding all those functions or not, it's a big change, maintenance  is always a concern and not sure if many users could need those functions. But I like them, so it's a difficult decision.

MatthewRoush:
That's reasonable, they certainly are quite big. Just wanted to make sure I wasn't missing something I was supposed to do.

raysan5:
@MatthewRoush Still have lots of doubts about this addition, almost 2000 new lines of code to maintain... this is huge... but adding it for now, hopefully functions are mostly self-contained and can be just removed if not possible to maintain...

Thanks for the addition.

PR #6077, opened 2026-08-21
[build] Updated examples/Makefile after testing on OSX

WilliamIsMe1:
I may be wrong, but the best way to install ray lib is through home-brew, and I'd expect that that should be a search path for others. Also I tested on MacOS arm64 and it works :D

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -314,6 +314,9 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW)
     ifeq ($(PLATFORM_OS),BSD)
         LDFLAGS += -Lsrc -L$(RAYLIB_LIB_PATH)
     endif
+	ifeq ($(PLATFORM_OS),OSX)
+		LDFLAGS += -L/opt/homebrew/lib -L$(RAYLIB_LIB_PATH)
+	endif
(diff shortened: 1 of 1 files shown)


examples/Makefile
```diff
@@ -418,7 +418,7 @@ ifeq ($(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW)
     ifeq ($(PLATFORM_OS),OSX)
         # Libraries for OSX 10.9 desktop compiling
         # NOTE: Required packages: libopenal-dev libegl1-mesa-dev
-        LDLIBS = -lraylib -framework OpenGL -framework Cocoa -framework IOKit -framework CoreAudio -framework CoreVideo -framework QuartzCore
+        LDLIBS = -L/opt/homebrew/lib -lraylib -framework OpenGL -framework Cocoa -framework IOKit -framework CoreAudio -framework CoreVideo -framework QuartzCore
```
raysan5: This should be added to `LDFLAGS`, not `LDLIBS`, please review it.

WilliamIsMe1: Done!

raysan5:
@WilliamIsMe1 thanks for the improvement and quick review!

PR #6081, opened 2026-08-21
[examples] New example shaders lights bloom

PanicTitan:
new example showing bloom effect on lights


raysan5:
@PanicTitan Nice example addition, thanks!

PR #6082, opened 2026-08-21
[examples] New example textures portal window

PanicTitan:
New example showing a portal like effect


raysan5:
@PanicTitan Very nice example, thanks

PR #6086, opened 2026-08-22
[cmake] Propagate Cocoa backend frameworks to consumers of static raylib on macOS

### LINKED ISSUE #6080: [cmake] Installed static raylib on macOS doesn't propagate Cocoa/IOKit frameworks to consumers
## Description

When raylib is built as a **static** library on macOS, installed via CMake, and consumed with `find_package(raylib)`, linking any consumer fails with undefined Objective-C symbols:

```
  "_objc_retain", referenced from:
      __glfwCreateStandardCursorCocoa in libraylib.a(cocoa_window.m.o)
ld: symbol(s) not found for architecture x86_64
```

The installed package config only propagates `OpenGL.framework`:

```
$ grep INTERFACE_LINK_LIBRARIES lib/cmake/raylib/raylib-targets.cmake
INTERFACE_LINK_LIBRARIES ".../OpenGL.framework;..."
```

`Cocoa`, `IOKit`, `QuartzCore` are missing, so every consumer must add the frameworks manually.

## Reproduction

```
git clone https://github.com/raysan5/raylib && cd raylib
mkdir build && cd build
cmake .. -DBUILD_SHARED_LIBS=OFF -DBUILD_EXAMPLES=OFF && make -j3
cmake --install . --prefix /tmp/raylib-stage
cd ../examples && mkdir build && cd build
cmake .. -DCMAKE_PREFIX_PATH=/tmp/raylib-stage && make core_basic_window
# -> ld: symbol(s) not found (objc_retain ...)
```

Building the same example through the root project (`-DBUILD_EXAMPLES=ON`) works, because directory-scoped link flags reach the examples target — the problem is only in what gets *installed/exported*.

## Suggested fix

Declare the required frameworks on the `raylib` target so they land in `INTERFACE_LINK_LIBRARIES`. For shared builds the frameworks are already linked into the dylib privately, so exporting them publicly would be redundant noise; a generator expression keyed on `BUILD_SHARED_LIBS` could keep the exported interface clean.

Found while verifying #6079.

zaveshaa:
Closes #6080

## Problem

When raylib is built as a **static** library and consumed via `find_package(raylib)` on macOS, linking fails with undefined Objective-C symbols:

```
  "_objc_retain", referenced from:
      __glfwCreateStandardCursorCocoa in libraylib.a(cocoa_window.m.o)
ld: symbol(s) not found for architecture x86_64
```

Root cause chain:
1. Bundled GLFW links its Cocoa backend frameworks (`Cocoa`, `IOKit`, `QuartzCore`) as **PRIVATE**
2. The static-install export block in `src/CMakeLists.txt` removes `glfw` from exported dependencies (correct — its symbols are inside the archive)
3. But the frameworks GLFW needed are not re-exported either, so the installed package only carries `OpenGL.framework`

## Fix

In the static-only export block, when the bundled GLFW is used on `APPLE`, append the same three frameworks GLFW's Cocoa backend declares, resolved via `find_library()`. Shared builds are untouched (the block is guarded by `NOT BUILD_SHARED_LIBS`).

## Verification (macOS 15, Intel, Apple clang 14)

Full standalone flow, before → after:

| Step | Before | After |
|---|---|---|
| `cd examples && cmake .. -DCMAKE_PREFIX_PATH=<stage>` | configures (after #6079) | ✅ |
| `make core_basic_window` | ❌ ld: objc symbols not found | ✅ links |

Repro script:
```
cmake .. -DBUILD_SHARED_LIBS=OFF -DBUILD_EXAMPLES=OFF && make -j3
cmake --install . --prefix /tmp/raylib-stage
cd ../examples && cmake .. -DCMAKE_PREFIX_PATH=/tmp/raylib-stage && make core_basic_window
```

Note: this PR depends on #6079 for the examples to configure standalone; the library-side fix here is independent.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,5 +1,17 @@
 # Setup the project and settings
+cmake_minimum_required(VERSION 3.22)
 project(examples)
+
+# Include raylib's cmake helper functions, so the examples can also be
+# configured standalone (`cd examples && cmake .`) and not only as part
+# of the raylib project.
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/../cmake")
+include(AddIfFlagCompiles)
+
+# Default to the Desktop platform when configured standalone.
+if (NOT DEFINED PLATFORM)
+    set(PLATFORM "Desktop" CACHE STRING "Platform to build examples for")
+endif ()
 
 # Directories that contain examples
 set(example_dirs
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -117,8 +117,9 @@ if (NOT BUILD_SHARED_LIBS)
 
         # Bundled GLFW links its Cocoa backend frameworks privately, so they
         # must be propagated to consumers of the static library, whose final
-        # link line would otherwise miss them.
-        if (APPLE)
+        # link line would otherwise miss them. Only the Desktop platform uses
+        # bundled GLFW; other platforms (e.g. Memory) have no Cocoa/GLFW symbols.
+        if (APPLE AND PLATFORM STREQUAL "Desktop")
             find_library(COCOA_FRAMEWORK Cocoa)
             find_library(IOKIT_FRAMEWORK IOKit)
             find_library(QUARTZCORE_FRAMEWORK QuartzCore)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -114,6 +114,19 @@ if (NOT BUILD_SHARED_LIBS)
 
     if (NOT glfw3_FOUND)
         list(REMOVE_ITEM raylib_install_private_libs glfw)
+
+        # Bundled GLFW links its Cocoa backend frameworks privately, so they
+        # must be propagated to consumers of the static library, whose final
+        # link line would otherwise miss them.
+        if (APPLE)
+            find_library(COCOA_FRAMEWORK Cocoa)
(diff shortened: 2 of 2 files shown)


raysan5:
@Peter0x44 @manuel5975p @VictorHM @JohnnyCena123, @ninadsachania, @arodriguez22, @pamarcos, @Shakai-Dev, @SajendraN, @heliumsneakers, @nukelash, @RobLoach Please, could you review this PR?

heliumsneakers (changes requested):
This is a good change, but it needs to be gated only to the desktop platform. For example, if I build on MacOS with `-DPLATFORM=Memory` Cocoa/IOKit/QuartzCore are now exported even though that platform uses the software framebuffer backend and has no Cocoa/GLFW symbols. 

Since bundled GLFW is only added for desktop I recommend gating this change there:

``` cmake
# Instead of only
if (APPLE)

# Do something like
if (APPLE AND "${PLATFORM}" STREQUAL "Desktop")
```

raysan5:
@heliumsneakers thanks for the review; @zaveshaa please, could you apply proposed changes?

zaveshaa:
@heliumsneakers good catch, thanks — pushed a fix that gates the framework propagation with `APPLE AND PLATFORM STREQUAL "Desktop"` (matching the existing check further down), so platforms like Memory no longer export Cocoa/IOKit/QuartzCore.

raysan5:
@zaveshaa thanks for the improvement!

PR #6088, opened 2026-08-22
[external] Small fix to fix_win32_compatibility.h for newer windows SDK versions.

JeffM2501:
This PR makes a small change to the compatibility header. Newer versions of windows SDKs removed PlaySound from windows.h when WIN32_LEAN_AND_MEAN is defined, so to ensure it is picked up by the replacement macro we must include the sound header specifically.

This prevents a link error when including windows.h and raylib in the same translation unit.


raysan5:

PR #6089, opened 2026-08-23
[tools] Fix crash when adding a new textures or audio examples

PanicTitan:
rexm could crash and corrupt examples_list.txt when adding a new example in the textures category. This happened because the tool searched for the category name as plain text, and "text" happens to be part of "textures" (and shows up inside names like core_render_texture), so it sometimes found the wrong spot and inserted the new example in the middle of an unrelated line.

This fix makes the search look for the exact category boundary instead of just any matching text, so it always finds the right place. It also fixes a small out-of-bounds issue that could happen specifically with the audio category, since it's the last one in the list.

Tested by creating a new textures example and a new audio example, both now work correctly and the examples list stays intact.

I found this when making the models_mesh_uv_painting example, it was textures_mesh_uv_painting but I faced this bug only in this example then I renamed it to test and it worked.


raysan5:
@PanicTitan thanks for the review

PR #5968, opened 2026-07-11
[rcore][desktop] Fix clipboard image memory leak

LuisR385:
# PR Title

Fix memory leak in `GetClipboardImage()`

---

# Summary

This PR fixes a memory leak in `GetClipboardImage()`.

`SDL_GetClipboardData()` returns a heap-allocated buffer that must be released with `SDL_free()`. The current implementation loads the image from the returned buffer but never frees it, resulting in a memory leak.

This PR simply calls `SDL_free(fileData)` immediately after `LoadImageFromMemory()`.

The goal of this PR is to keep the change as small as possible and consistent with raylib's design philosophy.

---

# Environment

- Windows 11
- x64
- MSVC
- Visual Studio 2022

---

# Reproduction

The following sample reproduces the issue.

```cpp
#include <raylib.h>
#include <crtdbg.h>

int main(void)
{
    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);

    const int screenWidth = 800;
    const int screenHeight = 450;

    InitWindow(screenWidth, screenHeight, "raylib - clipboard image memory leak");

    SetTargetFPS(60);

    Texture2D clipboardTexture = { 0 };
    bool hasClipboardTexture = false;

    while (!WindowShouldClose())
    {
        const bool controlPressed =
            IsKeyDown(KEY_LEFT_CONTROL) ||
            IsKeyDown(KEY_RIGHT_CONTROL);

        if (controlPressed && IsKeyPressed(KEY_V))
        {
            Image clipboardImage = GetClipboardImage();

            if (IsImageValid(clipboardImage))
            {
                if (hasClipboardTexture)
                {
                    UnloadTexture(clipboardTexture);
                }

                clipboardTexture = LoadTextureFromImage(clipboardImage);
                hasClipboardTexture = IsTextureValid(clipboardTexture);

                UnloadImage(clipboardImage);
            }
        }

        BeginDrawing();

        ClearBackground(RAYWHITE);

        DrawText("Copy an image, then press Ctrl + V", 20, 20, 20, DARKGRAY);

        if (hasClipboardTexture)
        {
            DrawTexture(clipboardTexture, 20, 60, WHITE);
        }

        EndDrawing();
    }

    if (hasClipboardTexture)
    {
        UnloadTexture(clipboardTexture);
    }

    CloseWindow();

    return 0;
}
```

---

# Reproduction Images

## Example 1

Copy this image:

[...]

PR #6091, opened 2026-08-25
[rcore][win32] Implement GetClipboardImage()

omaribrahimsaad:
GetClipboardImage() was still a stub on the native Win32 backend — it just logged a warning, and the message was wrong anyway (it said "GetClipboardText not implemented").

src/external/win32_clipboard.h is already in the repo and already used by the GLFW and RGFW backends, so this just wires it up the same way.

I followed the RGFW pattern rather than the GLFW one, because rcore_desktop_win32.c already includes windows.h and needs the WINUSER/WINBASE/WINGDI_ALREADY_INCLUDED guards to avoid redefinitions.

Also frees the buffer returned by Win32GetClipboardImageData() after decoding — LoadImageFromMemory() copies into its own allocation, so the raw BMP bytes are orphaned otherwise. Same fix as [#5968](https://github.com/raysan5/raylib/pull/5968) did for the SDL backend. Note that GLFW and RGFW still have this leak; happy to do those in a separate PR if you want them consistent.

Tested on Windows 10 with MSVC:

pasting an image works, including repeatedly
empty clipboard and text-in-clipboard both warn and return an empty Image, no crash
builds with SUPPORT_CLIPBOARD_IMAGE set to both 0 and 1
default GLFW build unaffected, no new warnings

Found this while testing the textures_clipboard_image example. Its Ctrl+V path doesn't currently work on this backend since the modifier keys aren't mapped yet in GetKeyFromWparam(), so I tested with a local tweak — looking at that separately.

### DIFF
--- a/src/platforms/rcore_desktop_win32.c
+++ b/src/platforms/rcore_desktop_win32.c
@@ -1129,12 +1129,37 @@ const char *GetClipboardText(void)
     return NULL;
 }
 
+#if SUPPORT_CLIPBOARD_IMAGE && SUPPORT_MODULE_RTEXTURES
+    #define WIN32_CLIPBOARD_IMPLEMENTATION
+    #define WINUSER_ALREADY_INCLUDED
+    #define WINBASE_ALREADY_INCLUDED
+    #define WINGDI_ALREADY_INCLUDED
+    #include "../external/win32_clipboard.h"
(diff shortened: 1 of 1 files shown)


raysan5:
@omaribrahimsaad thanks for the addition

Issue #5664, opened 2026-03-16
[examples] Examples collection testing and issues reporting for `raylib 6.0`


raysan5:
Opening this issue to get non-properly-working examples reported. I tried compiling examples for Windows, Linux and Web, also updated `raylib.com/examples` with latest version.

All the automatic tests have passed, including `rexm` tool tests. Still there can be examples not working as expected on some platforms. Opening this issue to get those examples reported. Please report them as:
```
[example_name] - Platform
```
Platform being one of: `Windows`, `Linux/X11`, `Linux/Wayland`, `macOS`, `Web`, `DRM`, `Android`, other (specify)

For example:
```
audio_stream_callback - Web, macOS
```

dan-hoang:
audio_module_playing - Web

This example works for me on desktop but not web, neither locally nor on raylib.com. I tried on both Firefox and Safari (iOS).

There is this error: ```WARNING: FILEIO: [resources/mini1111.xm] Music file could not be opened```

raysan5:
Examples on **`PLATFORM_WEB`** testing results:

 - [ ] `core_window_letterbox` - Canvas scaling, FLAG_WINDOW_RESIZABLE issues
 - [ ] `core_viewport_scaling` - Canvas scaling, FLAG_WINDOW_RESIZABLE issues
 - [ ] `core_highdpi_testbed` - Canvas scaling, FLAG_WINDOW_RESIZABLE issues
 - [ ] `shapes_ball_physics` - Memory out of bounds
 - [ ] `textures_tiled_drawing` - Canvas scaling, FLAG_WINDOW_RESIZABLE issues
 - [x] `textures_cellular_automata` - UI issues
 - [ ] `textures_framebuffer_rendering` - Controls/camera issues
 - [ ] `text_font_loading` - BMFont support not enabled
 - [x] `models_yaw_pitch_roll` - Texture issues with model (WRAP mode support on Web?)
 - [ ] `models_decals` - Memory out of bounds
 - [ ] `models_point_rendering` - Points not supported on Web?
 - [ ] `shaders_basic_lighting` - Issues with `lighting` shader loading, possibly related to server, not shader itself
 - [ ] `shaders_mesh_instancing` - Issues with `lighting` shader loading, possibly related to server, not shader itself
 - [ ] `shaders_deferred_rendering` - Issues with `lighting` shader loading, possibly related to server, not shader itself
[...]

PR #6092, opened 2026-08-25
[examples] Fix crash when model resources fail to load in animation examples

zaveshaa:
Fixes the segfault reported in #5664 for `models_animation_blending`: when `resources/models/gltf/robot.glb` can not be opened, `LoadModelAnimations()` returns NULL with `animCount = 0`, and the example still indexes `anims[10]`.

While checking I found the same pattern in three more animation examples:

- `models_animation_gpu_skinning`: `(animIndex + 1) % animCount` divides by zero on key press, then reads `anims[animIndex]`
- `models_animation_timing`: update loop is guarded, but the UI drawing reads `anims[animIndex]` every frame unconditionally
- `models_animation_blend_custom`: index clamping does not help because the array itself is NULL

All four now exit early with a warning when the model or its animations could not be loaded, unloading what was created. All four still compile.

### DIFF
--- a/examples/models/models_animation_blend_custom.c
+++ b/examples/models/models_animation_blend_custom.c
@@ -81,6 +81,16 @@ int main(void)
     int animCount = 0;
     ModelAnimation *anims = LoadModelAnimations("resources/models/gltf/greenman.glb", &animCount);
 
+    // WARNING: Exit if example resources could not be loaded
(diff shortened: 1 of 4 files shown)


raysan5:
It's up to the user to manage it, I prefer to avoid additional code

PR #6095, opened 2026-08-25
Updated BINDINGS.md

overdev:
for Python raylib-py v6.0 release


raysan5:
@overdev thanks for the update!

PR #6096, opened 2026-08-25
Fix unused variable warnings (SDL3)

M374LX:
This fixes warnings about variables named ``monitorCount`` in different functions being unused when using SDL3.

Example build command:
```
make CUSTOM_CFLAGS=-DUSING_SDL3_PROJECT PLATFORM=PLATFORM_DESKTOP_SDL SDL_INCLUDE_PATH=/usr/include
```


raysan5:
@M374LX thanks for the improvement!

Issue #6097, opened 2026-08-26
[rtext] Memory leak at LoadFontFromImage when failing to compute spacing's.


amuerta:
I did not test latest commit, but I have not found concrete and exact problem mentioned by briefly looking onto most `LoadFontFromImage` issues.

 - [ ] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [x] My code has no errors or misuse of raylib

### Issue description

In process of attempting to make my own default font from a `.bdf` font, I accidentally found a mistake in `LoadFontFromImage` that causes a substantial memory leak that was caught by memory address sanitizer present in clang.

I converted `.bdf` font to `.png` which looks like so:

<img width="112" height="182" alt="Image" src="https://github.com/user-attachments/assets/2676bdb1-b291-440d-b4ea-e1749339b3d9" />

When loading it i got sanitizer leak complaints. I started seeking the source.

The problem occurs at https://github.com/raysan5/raylib/blob/master/src/rtext.c#L441
13 lines before that:
 ```c
Color *pixels = LoadImageColors(image);
```
makes an new allocation and then checks in a loop for some pixel miss match? I'd guess we count empty pixels before
we see a symbol pixel. For some reason this font atlas fails there and quits with `x` and `y` begin zero which is handled
by returning a default raylib font.

AHA! We forgot to free `pixels`! 

Adding a free fixes it.
```c
    if ((x == 0) || (y == 0)) {
        free(pixels); // amu: added this to prevent leak.
        return font; // ray: Security check
    }
```

### Environment  
- I use ArchLinux.
- Usually have vendor copies of older raylib (5.0, 5.5, 6.0)
- This problem persist between 5.5 and 6.0, I didn't check 5.0. 

[~] › sudo pacman -Si raylib
Repository      : extra
Name            : raylib
Version         : 6.0-1
Description     : Simple and easy-to-use game programming
                  library
Architecture    : x86_64
URL             : https://www.raylib.com
Licenses        : Zlib
Groups          : None
Provides        : None
Depends On      : None
Optional Deps   : None
Conflicts With  : None
Replaces        : None
Download Size   : 774.06 KiB
Installed Size  : 2381.61 KiB
Packager        : Alexander F. Rødseth
                  <xyproto@archlinux.org>
Build Date      : Sat 25 Apr 2026 01:49:53 PM EEST
[...]

PR #6100, opened 2026-08-26
Fix memory leak in rtext LoadFontFromImage method

ApexNadir:
Fix for
[rtext] Memory leak at LoadFontFromImage when failing to compute spacing's

https://github.com/raysan5/raylib/issues/6097

Issue is memory is being allocated for the image but it's not being added in the returned font due to the early exit. So when the font is ultimately destroyed, this memory is not freed.


raysan5:
@ApexNadir good catch! thanks for the review!

PR #6101, opened 2026-08-26
[rexm] Leave environment setup up to the user.

MatthewRoush:
rexm used to assume certain paths existed and then `_putenv()` them into `PATH`, which is very error prone. `make` was also assumed, though `mingw32-make` seems to be more common on Windows. Since this shifts things to expect the user to setup their environment, I added some explanation to the README as well.

`testlog` command still uses a hardcoded path when doing web testing (`BUILD_TESTING_WEB`,) unfortunately.

### DIFF
--- a/tools/rexm/rexm.c
+++ b/tools/rexm/rexm.c
@@ -70,6 +70,12 @@
 #define REXM_MAX_RESOURCE_PATHS         256
 #define REXM_MAX_RESOURCE_PATH_LENGTH   256
 
+#if defined(_WIN32) && defined(USE_MINGW_MAKE)
+    #define MAKE_COMMAND "mingw32-make"
+#else
+    #define MAKE_COMMAND "make"
+#endif
+
 // Create local commit with changes on example renaming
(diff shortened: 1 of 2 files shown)


tools/rexm/rexm.c
```diff
@@ -70,6 +70,12 @@
 #define REXM_MAX_RESOURCE_PATHS         256
 #define REXM_MAX_RESOURCE_PATH_LENGTH   256
 
+#if defined(_WIN32) && defined(USE_MINGW_MAKE)
+    #define MAKE_COMMAND "mingw32-make"
```
raysan5, line 74: Actually, latest `w64devkit` versions already use `make`, I prefer to avoid that `mingw32-make` and keep a common path on all platforms.

raysan5:
@MatthewRoush Thanks for the review, definitely it's better to avoid hardcoded paths. `rexm` evolved from some quick code to automate examples building and there are some parts that could be improved.

raysan5:
I'm merging and reviewing that detail myself.

PR #6102, opened 2026-08-26
Add raylib binding for solod

nalgeon:
Please add this raylib binding for the Solod language. Thank you for raylib!


raysan5:
@nalgeon Very nice! Thanks for the addition! 👍😄

PR #6103, opened 2026-08-27
[rcore][desktop] Fix clipboard image memory leak (GLFW, RGFW)

omaribrahimsaad:
Follow-up to [#6091](https://github.com/raysan5/raylib/pull/6091). Win32GetClipboardImageData() returns an RL_MALLOC'd buffer, and LoadImageFromMemory() decodes into its own allocation, so the raw BMP bytes are leaked on every paste. Same fix [#5968](https://github.com/raysan5/raylib/pull/5968) applied to the SDL backend.

Tested on Windows with both backends — repeated pastes work, no change in behaviour.

Unrelated, but noticed while in there: the RGFW X11 branches at lines 1310 and 1334 check defined(DRGFW_X11), but the macro is RGFW_X11 (src/Makefile:513 passes -DRGFW_X11). Looks like the -D got into the name, so those paths are dead on Linux. Left it alone since I can't test X11 — happy to patch if you want.


raysan5:
@omaribrahimsaad Good catch! Thanks for the fix!

PR #6104, opened 2026-08-27
[rcore][Emscripten] Invert emscripten scroll wheel direction

andzdroid:
Emscripten's wheel down is positive, unlike glfw and raylib: https://emscripten.org/docs/api_reference/html5.h.html#c.EmscriptenWheelEvent


raysan5:
@andzdroid thanks for the review

PR #6105, opened 2026-08-28
feat(models): add opt-in u32 mesh index support

jvkabum:
## Summary

Adds opt-in 32-bit mesh index support while preserving the existing 16-bit index path and API behavior.

`Mesh.indices` remains `unsigned short *` for backward compatibility. Meshes that require indices above `65535` can use the new `indices32` field together with `MESH_INDEX_UINT32`.

## Changes

- Added `MeshIndexType` with `MESH_INDEX_UINT16` and `MESH_INDEX_UINT32`.
- Added `Mesh.indices32` and `Mesh.indexType`.
- Appended the new `Mesh` fields to preserve the offsets of existing public fields.
- Updated mesh upload and draw paths to select the correct OpenGL index type:
  - `GL_UNSIGNED_SHORT` for `u16`
  - `GL_UNSIGNED_INT` for `u32`
- Updated indexed and instanced drawing paths.
- Updated mesh cleanup, validation, export, and collision-related index access.
- Updated the glTF loader to preserve `UNSIGNED_INT` index accessors instead of narrowing them to `u16`.
- Added a regression example containing an index value of `65536`.

## Compatibility

Existing applications and models using `Mesh.indices` continue to use 16-bit indices unchanged.

The new format is opt-in:

```c
mesh.indices32 = indices;
mesh.indexType = MESH_INDEX_UINT32;

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -695,6 +695,7 @@ MODELS = \
     models/models_loading_m3d \
     models/models_loading_vox \
     models/models_mesh_generation \
+    models/models_mesh_u32_indices \
     models/models_mesh_picking \
     models/models_orthographic_projection \
     models/models_point_rendering \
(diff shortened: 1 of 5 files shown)


raysan5:
@jvkabum thanks but I prefer to avoid two paths to maintain

PR #6106, opened 2026-08-28
[rexm] Fix small oversight.

MatthewRoush:
The leading double quotes were accidentally deleted in the `make` command strings.


raysan5:
Ouch! Thanks! 😅

Issue #6056, opened 2026-08-12
[rcore][drm] `SUPPORT_DRM_CACHE` stops presenting after 2 page flips on Mali (RK3326)


jtakakura:
Please, before submitting a new issue verify and check:

 - [x] I tested it on latest raylib version from master branch
 - [x] I checked there is no similar issue already reported
 - [x] I checked the documentation on the wiki
 - [x] My code has no errors or misuse of raylib

### Issue description

Follow-up to #4988, where more testing and benchmarking was asked for before making the cached path the default. I have an R36S, which uses ARM's Mali blob rather than mesa, so the results are different from the Raspberry Pi it was developed on.

With `SUPPORT_DRM_CACHE` defined, the screen never updates. Not "runs and then freezes" — nothing appears after the first couple of frames. The game loop is unaffected: my log shows it running at 39-60 fps, button presses still change scenes, and the program exits cleanly. Only presentation is dead.

I interposed on libdrm and gbm to see what happens:

    addfb      call=1 rc=0 handle=1
    addfb      call=2 rc=0 handle=2
    addfb      call=3 rc=0 handle=4
    lock_front call=1 bo=0x512a790 NEW distinct=1
    lock_front call=2 bo=0x52390d0 NEW distinct=2
    lock_front call=3 bo=0x5238b20 NEW distinct=3
    lock_front call=4 bo=0x51ab430 NEW distinct=4
    pageflip   call=1 rc=0
    pageflip   call=2 rc=0

After this there are no further `drmModeAddFB` and no further `drmModePageFlip` calls for the rest of the session, while `gbm_surface_lock_front_buffer()` and `gbm_surface_release_buffer()` keep being called every frame. `lock_front` never returns NULL, so this is not buffer exhaustion.

`gbm_surface_lock_front_buffer()` hands out a 4th distinct `gbm_bo` pointer on this driver. `fbCache` holds `MAX_DRM_CACHED_BUFFERS = 3` and `GetOrCreateFbForBo()` has no eviction, so from the 4th buffer on it takes

    if (fbCacheCount >= MAX_DRM_CACHED_BUFFERS) return 0; // FB cache full

and `SwapScreenBuffer()` releases the buffer and returns before the page flip, every frame, permanently. `errCnt[2]` counts exactly this, but the block that prints the counters is commented out.

I would guess mesa reuses the same three buffers, which would explain why this does not show up on a Pi.

[...]

PR #6107, opened 2026-08-28
[rcore][drm] Raise MAX_DRM_CACHED_BUFFERS to 4 and warn when the cache fills

jtakakura:
Fixes the `SUPPORT_DRM_CACHE` freeze in #6056, the way @rob-bits suggested
there.

`fbCache` has no eviction, so when a driver rotates a fourth buffer,
`GetOrCreateFbForBo()` returns 0 for it and keeps doing so, and
`SwapScreenBuffer()` returns before the page flip on every frame after that.
Mesa rotates three buffers, ARM's Mali blob on RK3326 rotates four.

Verified on an R36S: four buffers get framebuffers, page flips continue past
1900 instead of stopping at 2, and `SetTargetFPS` holds 39, 60 and 78 fps.
Traces are in #6056.

The warning covers a driver that rotates more still, since that case is a
permanent freeze rather than a dropped frame and `errCnt[2]` is never printed.

### DIFF
--- a/src/platforms/rcore_drm.c
+++ b/src/platforms/rcore_drm.c
@@ -79,7 +79,11 @@
     #include <poll.h>       // Required for: drmHandleEvent() poll
     #include <errno.h>      // Required for: EBUSY, EAGAIN
 
-    #define MAX_DRM_CACHED_BUFFERS  3
+    // NOTE: Some drivers rotate more buffers than others, mesa uses 3 but the
+    // ARM Mali blob on RK3326 uses 4. A buffer beyond this count can not get a
+    // framebuffer and the display stops updating for good, so this must not be
+    // lower than any driver in use
+    #define MAX_DRM_CACHED_BUFFERS  4
 #endif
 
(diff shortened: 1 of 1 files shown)


src/platforms/rcore_drm.c
```diff
@@ -624,7 +628,19 @@ static uint32_t GetOrCreateFbForBo(struct gbm_bo *bo)
     }
 
     // Create new entry if cache not full
-    if (fbCacheCount >= MAX_DRM_CACHED_BUFFERS) return 0; // FB cache full
+    if (fbCacheCount >= MAX_DRM_CACHED_BUFFERS)
+    {
+        // NOTE: Once this happens no page flip is ever scheduled again and the
+        // screen freezes while the game keeps running, so say it out loud once
+        static bool cacheFullReported = false;
```
raysan5: Please, could you move this global to PlatformData?

raysan5:
@jtakakura thanks for the review!

PR #6111, opened 2026-08-29
[rexm] Improve `.log` file saving for `test` command.

MatthewRoush:
Before, the `test` command used javascript (client) to save the generated log file, which doesn't allow you to specify a path to save the file to (for security reasons.)

`test` was already using a python server to run the web example, which we can take advantage of. Using a python script (`rexm_web_test_server.py`) we can control how we handle POST requests. With that, we can use javascript to send a POST request to the server, and the server can then save the log file (avoiding client restrictions.)

As an added bonus, we don't need `FileSaver.js` anymore.

## Miscellaneous changes

- Fix `tools/rexm/Makefile` so that it uses `SHELL=cmd` on Windows.
- Change the `test` command server port number from `8080` to `38080` to reduce conflicts with other programs.

### DIFF
--- a/src/shell.html
+++ b/src/shell.html
@@ -194,23 +194,16 @@ jwE50AGjLCVuS8Yt4H7OgZLKK5EKOsLviEWJSL/+0uMi7gLUSBseYwqEbXvSHCec1CJvZPyHCmYQffaB
 
     <textarea id="output" rows="8"></textarea>
 
-    <script type='text/javascript' src="https://cdn.jsdelivr.net/gh/eligrey/FileSaver.js/dist/FileSaver.min.js"> </script>
     <script type='text/javascript'>
-        function saveFileFromMEMFSToDisk(memoryFSname, localFSname)     // This can be called by C/C++ code
+        // Save a file from memory onto the server
+        function saveFileFromMEMFSToDisk(memoryFSname, localFSpath)     // This can be called by C/C++ code
         {
-            var isSafari = false; // Not supported, navigator.userAgent access is being restricted
-            //var isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
             var data = FS.readFile(memoryFSname);
-            var blob;
 
-            if (isSafari) blob = new Blob([data.buffer], { type: "application/octet-stream" });
-            else blob = new Blob([data.buffer], { type: "application/octet-binary" });
-
-            // NOTE: SaveAsDialog is a browser setting. For example, in Google Chrome,
-            // in Settings/Advanced/Downloads section you have a setting:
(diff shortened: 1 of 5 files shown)


src/shell.html
```diff
@@ -194,23 +194,16 @@
...
 
-            if (isSafari) blob = new Blob([data.buffer], { type: "application/octet-stream" });
-            else blob = new Blob([data.buffer], { type: "application/octet-binary" });
-
-            // NOTE: SaveAsDialog is a browser setting. For example, in Google Chrome,
-            // in Settings/Advanced/Downloads section you have a setting:
-            // 'Ask where to save each file before downloading' - which you can set true/false.
-            // If you enable this setting it would always ask you and bring the SaveAsDialog
-            saveAs(blob, localFSname);
+            var xhr = new XMLHttpRequest();
```
raysan5, line 203: Does it require the Python script?

raysan5:
@MatthewRoush I don't like `FileSaver.js` but I prefer to avoid adding a Python script dependency. The `XMLHttpRequest` proposed requires the Python script?

MatthewRoush:
Yes, because we have to implement our own `POST` handling logic. The primary annoyance with the previous implementation was that it couldn't save a file to a specific path because of browser security restrictions. This means that the file is automatically (or not, if the user has the setting on that makes it prompt every time) saved to the "default" save location, which may not be the traditional "Downloads" directory.

The way to bypass the client-side file saving restrictions is to not save the file from the client, but instead from the server (which is also just the client computer, because it's a `localhost://`.) The problem is that the default server from running `python http.server` (which is what we previously did to run the web exmaples) doesn't handle `POST` requests. That means we have to write our own Python script to implement `POST` request handling. We also have to handle `GET` requests since we're using our own request handling logic, as far as I know.

The `XMLHttpRequest()` sends the `POST` request from the client to the server, and the Python script (our server) handles that `POST` request. We can send a `POST` request containing the `.log` data, with the URL being the path (relative to the server root) we want to save the `.log` file to (`exBasePath/exCategory/logs/exName.log`). Then, on the server, we can handle that `POST` request by saving the `.log` data we received to the file path specified in the URL.

It is unfortunate to have the extra Python script, but it solves the "how do we save the '.log' file to the specific path that we want in an automated way without requiring the user to just do it themselves" problem that we have.

raysan5:
@MatthewRoush thanks for the detailed explanation but still I prefer let the user manage it manually than adding a Python script dependency.

EDIT: Not sure if some of the parts of this PR could also be implemented without the script addition.

MatthewRoush:
Yeah, the two small changes (fixing the Makefile and changing the port number) are unrelated to the script, everything else relies on the script.

Issue #6110, opened 2026-08-29
[rlgl] Batching system grouping incompatible drawing commands under rare circumstances


MikiZX1:
### Issue description

Both DrawBillboard and DrawPlane use batch system to draw geometry and they both use QUADs to add geometry to the batch system. At the moment batch system will batch the two geometries together even though they will not necessarily be compatible due to different textures used by each. 
Issue arises when billboard is added to the batch first (as initial geometry in that batch) - adding a plane after the billboard causes the plane to be drawn using billboard's texture rather than raylib's default 1x1 white texture.
If a plane is added first to the batch, followed by a billboard then both the plane and the billboard are drawn correctly; if then another plane is added to the batch, that second plane is drawn (again) using the texture provided for the billboard.

After the billboard has been added to the batch and the batch is rendered using rlDrawRenderBatchActive(), drawing plane works well.

### Environment

Linux, OpenGL version 3.3, GPU details unknown

### Issue Screenshot

Issue:
<img width="742" height="429" alt="Image" src="https://github.com/user-attachments/assets/3115b98c-f345-4f13-a570-3f9affa9aa72" />

Expected (No Issue):
<img width="705" height="418" alt="Image" src="https://github.com/user-attachments/assets/296dc661-65c7-4f24-bf56-1f4f166642a7" />

### Code Example
```

int main()
{
InitWindow(800, 450, "raylib - test batching");

Camera camera;
camera.fovy=90;
camera.position=(Vector3){0,10,-10};
camera.target=Vector3Zero();
camera.up=(Vector3){0,1,0};
camera.projection=CAMERA_PERSPECTIVE;

Image image1=GenImageColor(128,128,ORANGE);
Texture texture1=LoadTextureFromImage(image1);
UnloadImage(image1);

 while (!WindowShouldClose()) 
 {
    BeginDrawing();
      ClearBackground(GRAY);
      BeginMode3D(camera);
      DrawBillboard(camera,texture1,(Vector3){7.0f,5.0f,6.0f}, 3.0f, WHITE);
      //rlDrawRenderBatchActive();    // makes drawing OK
      DrawPlane((Vector3){0.0f,-10.0f,0.0f},(Vector2){10,10}, WHITE);
      EndMode3D();
    EndDrawing();
 }

UnloadTexture(texture1);
CloseWindow();
return 0;
}
```

raysan5:
[...]

PR #6112, opened 2026-08-29
[rlgl] Fix batching system grouping incompatible drawing commands (#6110)

jxpitwr333:
#6110 

rlBegin() was only checking primitive mode changes and skipping draw call splits when textureId changed (for example, DrawPlane after DrawBillboard). Now checks both mode and textureId against RLGL.State.currentTextureId before appending vertices.

### DIFF
--- a/src/rlgl.h
+++ b/src/rlgl.h
@@ -1452,7 +1452,8 @@ void rlBegin(int mode)
 {
     // Draw mode can be RL_LINES, RL_TRIANGLES and RL_QUADS
     // NOTE: In all three cases, vertex are accumulated over default internal vertex buffer
-    if (RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].mode != mode)
+    if ((RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].mode != mode) ||
+        (RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].textureId != RLGL.State.currentTextureId))
     {
         if (RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].vertexCount > 0)
         {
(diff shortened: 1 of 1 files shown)


raysan5:
@jxpitwr333 this is an extremely sensitive change and it has broken raylib batching several times, please, could you test with RenderDoc that multiple draw calls do not break the batching? You can test bunnymark example and also different shapes drawing. Post screenshots, please.

jxpitwr333:
@raysan5 Sorry, I forgot to submit test screenshots.

This is the issue pre fix, using the code from the example provided in the issue:
issue_before_change
Only one draw call, but both calls in the code are different and the second one inherits the first one.

This is the issue post-fix:
issue_example
As you can see, separate draw calls, no wrongful inheritance from batching.

Bunnymark before:
bunnymark_before_change

Bunnymark after:
bunnymark_after_change

Please forgive the fact that the number of bunnies is not exactly the same in both screenshots, for me, the left mouse button behaves weirdly inside qrenderdoc vs when opening the application without it. But they're in the same ballpark and the batch calls appear correct.

If more screenshots are needed, I will submit them in short notice.

src/rlgl.h
```diff
@@ -1476,7 +1477,7 @@ void rlBegin(int mode)
 
         RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].mode = mode;
         RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].textureId = RLGL.State.currentTextureId;
-        RLGL.State.currentTextureId = RLGL.State.defaultTextureId;
+        RLGL.currentBatch->draws[RLGL.currentBatch->drawCounter - 1].vertexCount = 0;
```
raysan5, line 1480: Please, could you explain why those two lines changed?

CrackedPixel: @jxpitwr333 is this a mistake?

jxpitwr333: hi hello i honestly forgot to answer this, because when I was retesting the changes I tried some repeated shapes and did not like/understand what was happening, so I stopped. my change worked for the bug reported in the issue, and for bunnymark, but i cannot say for certain it works under the other test i tried, or if i'm missing something within the test and the test itself is wrong.

the last line resetting the vertex is a no-op anyway. i added it because i thought it was needed, as rlSetTexture and other adjacent functions also do it. but upon inspecting it i realized it is redundant. it can be safely removed.

still, I think this PR should be closed until I or someone smarter can properly ensure this has no weird side-effects.

raysan5:
This change is too sensible and could potentially break many things for many users.

PR #6114, opened 2026-09-01
[rTextures] Add LoadRenderTextureEx with user defined format

JeffM2501:
This PR adds LoadRenderTextureEx that takes the color buffer pixel format as an argument, allowing the user to define a FBO other than RGBA. This is useful for render textures that are used for simple things like masks or fullscreen processing where a full RGBA buffer is not needed, and can save a lot of vram.
Only uncompressed formats are supported, invalid formats emit a trace log message.

### DIFF
--- a/src/raylib.h
+++ b/src/raylib.h
@@ -1461,6 +1461,7 @@ RLAPI Texture2D LoadTexture(const char *fileName);                              
 RLAPI Texture2D LoadTextureFromImage(Image image);                                                       // Load texture from image data
 RLAPI TextureCubemap LoadTextureCubemap(Image image, int layout);                                        // Load cubemap from image, multiple image cubemap layouts supported
 RLAPI RenderTexture2D LoadRenderTexture(int width, int height);                                          // Load texture for rendering (framebuffer)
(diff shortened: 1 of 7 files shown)


src/raylib.h
```diff
@@ -1461,6 +1461,7 @@ RLAPI Texture2D LoadTexture(const char *fileName);
 RLAPI Texture2D LoadTextureFromImage(Image image);                                                       // Load texture from image data
 RLAPI TextureCubemap LoadTextureCubemap(Image image, int layout);                                        // Load cubemap from image, multiple image cubemap layouts supported
 RLAPI RenderTexture2D LoadRenderTexture(int width, int height);                                          // Load texture for rendering (framebuffer)
+RLAPI RenderTexture2D LoadRenderTextureEx(int width, int height, int pixelFormat);                       // Load texture for rendering (framebuffer), with specific format
```
raysan5: Please, could it use just `format` for parameter, to keep it aligned with `LoadImageRaw()`, `GetPixelColor()`, `SetPixelColor()`... I know is not tha explicit but it's also shorter. Thanks.

JeffM2501: updated

raysan5:
@JeffM2501 thanks for the addition

PR #6116, opened 2026-09-02
implemented integer render textures: R32I, R32UI, R16I, R16UI

alee724:


raysan5:
Those formats are not required by many users and add maintenance work

PR #6123, opened 2026-09-03
[rexm] Fix `Makefile` and change port number.

MatthewRoush:
- Fix `Makefile`: The `make` program that comes with `w64devkit` uses a UNIX-style shell (`w64devkit/bin/sh.exe`) so we can't use `del` command (we could do `SHELL=cmd` on Windows, but this is easier, and the expectation is that `w64devkit` is being used anyway.)
- Change port number: `8080` is a pretty common port number, changing it helps to prevent conflicts with other running programs.


raysan5:
@MatthewRoush thanks for the review!

PR #6124, opened 2026-09-04
[rexm] Add `clean` command.

MatthewRoush:
`build`, `test`, and `testlog` generate lots of files (`.exe`, `.html`, etc.) when executed. When you build/test enough examples, it's kind of annoying manually deleting the files. The generated reports (`.md` files) are not deleted.


raysan5:
@MatthewRoush looks good! thanks for the addition!

PR #6129, opened 2026-09-05
[raudio] Add functions for retrieving music and audio volume

Mystia-Izakaya:
We have SetMusicVolume and SetAudioVolume, so why not also add GetMusicVolume and GetAudioVolume?


raysan5:
No need, because those are set by user and can keep the reference.

PR #6134, opened 2026-09-09
[rcore_desktop_glfw.c] Do not reset mouse position user when user is off window

jestarray:
See this comment and Jeffs as to why
https://github.com/raysan5/raylib/issues/5945#issuecomment-5595141064

### DIFF
--- a/src/platforms/rcore_desktop_glfw.c
+++ b/src/platforms/rcore_desktop_glfw.c
@@ -2207,7 +2207,6 @@ static void CursorEnterCallback(GLFWwindow *window, int entered)
     else
     {
         CORE.Input.Mouse.cursorOnScreen = false;
-        CORE.Input.Mouse.currentPosition = (Vector2){ 0 };
     }
 }
 


largenumberhere:
I don't agree with this reversion. It means that it is now impossible to tell if the mouse is not on the window. It leaves the values undefined and platform-specific. 

With the other implementation it is trivial to simply keep track of which object was last hovered over yourself. In this specific example, my solution would be to check for 0,0 before updating the camera. In my opinion it should not be the library's responsibility to silently make this decision for you because very often it is unwanted behavior and there is no way to opt-out without this signalling mechanism.

(Re: https://github.com/raysan5/raylib/issues/5945#issuecomment-5595106183)
I also advise against using (-1, -1) as the sentinel value. While it seems clearer conceptually, it is likely to cause unexpected results in most mathematical calculations. (0,0) is easy to check for, is a more common C convention and unlikely to be returned under any other circumstances.

Another thing to consider - as shown by the camera snapping complaint above, (0, 0) shows up obviously and non-destructively in testing as a edge case that should be considered and solved. Whereas with this reversion, it will go back to being a silent unintended quirk in many applications. It took me some time to track it down this cause of my unintended behavior in my own application.

In summary, I believe the solution implemented in https://github.com/raysan5/raylib/issues/5945#issuecomment-5595141064 to be more obvious, more useful to the average person, more flexible, and more correct.

largenumberhere:
You seem to have combined this with a commit that clears out some metadata files?
I'm not familiar with that part of the repo, but it does look like it should be in a separate PR.

largenumberhere:
Oh, I should also say. Thank you for the review.

raysan5:
@jestarray why rlparser is broken? actually, why executes? 🤔

jestarray:
@raysan5 I honestly have no idea why it executed. Alright I removed it from this PR

raysan5:
@jestarray @largenumberhere despite this was changed a while ago, I think that forcing a { 0, 0 } when out of screen could be missleading and it's better to let thee underlying library set the default value for that case.

PR #6138, opened 2026-09-11
[raymath.h] Add lerp angle and normalized direction functions

xyzpw:
Adds `LerpAngle` for linear interpolation along shortest angular path.
Also adds normalized direction functions for `Vector2` and `Vector3`.


raysan5:
@xyzpw those functions can be easely created at user side when needed

PR #6141, opened 2026-09-11
[Bindings] Add recomendations for binding authors

JeffM2501:
This PR adds a list of recommendations for binding authors to help them provide a well supported and easy to use binding for raylib.


raysan5:
@JeffM2501 thanks, for some reason, `rlparser` workflow is running and I don't know why, it should only run when `raylib.h` is changed, [as stated in the workflow itself](https://github.com/raysan5/raylib/blob/master/.github/workflows/parse_api.yml#L7)... it can be some GitHub issue... I'm investigating it...

raysan5:
@JeffM2501 Anyway, just fixing it the hard way...

PR #6142, opened 2026-09-11
Update h-raylib version in BINDINGS.md

Anut-py:
Updated Haskell binding version from 5.5-dev to 6.1-dev.


raysan5:
@Anut-py thanks for the update

PR #6144, opened 2026-09-12
[rcore] Fix ComputeSHA256() wrong hash when message length mod 64 is 56..59

### LINKED ISSUE #6143: [rcore] ComputeSHA256() returns wrong hash for messages of length 56..59 mod 64

 - [X] I checked there is no similar issue already reported
 - [X] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)

### Issue description

`ComputeSHA256()` returns an incorrect hash whenever `dataSize % 64` is 56, 57, 58 or 59. It fails the FIPS 180-4 published two-block test vector.

The padding reserve is computed with the size of the wrong variable:

```c
const unsigned long long bitLen = 8ULL*dataSize;
unsigned long long paddedSize = dataSize + sizeof(dataSize);   // reserves 4 bytes
paddedSize += (64 - (paddedSize%64));
...
buffer[dataSize] = 0x80;                                        // 1 byte marker
for (int i = 1; i <= sizeof(bitLen); i++)                       // 8 byte length
{
    buffer[(paddedSize - sizeof(bitLen)) + (i - 1)] = ...;
}
```

`dataSize` is `int`, so `sizeof(dataSize)` reserves 4 bytes, but SHA-256 needs 9: the `0x80` marker plus a 64-bit big-endian message length. The write a few lines below correctly uses `sizeof(bitLen)` (8 bytes), so the reserve and the write disagree.

When `(dataSize + 4) % 64` leaves only 5 to 8 bytes of room, the length field is written on top of the `0x80` marker and the tail of the message, and the digest is wrong. The padded size is also one block short of the canonical size in those cases. There is no out-of-bounds access — every write stays inside the allocation — so the failure is silent.

Measured against a reference SHA-256 implementation on current master:

| Test | Result |
|---|---|
| FIPS 180-4 / NIST vectors (0, 3, 56, 112, 1000000 bytes) | 4 pass, **1 fail** (the 56 byte vector) |
| All lengths 0..4096 | **256 wrong** of 4097 |
| 3000 random messages, random lengths | **210 wrong** |

The failing lengths are exactly `dataSize % 64 ∈ {56, 57, 58, 59}`, which is 4/64 = **6.25%** of all input lengths. It is deterministic, not data dependent: within those four residue classes every length fails. Comparing the padding arithmetic against the canonical padded size over `dataSize = 0..10,000,000` gives 625,000 mismatches, all in the same four classes.

`ComputeSHA1()` is not affected: it reserves the full 8 bytes with `((((dataSize + 8)/64) + 1)*64)` and verifies clean against reference SHA-1 for all lengths 0..300. (Its separate length-encoding bug was fixed in #5397; that PR did not change this reserve.) `ComputeMD5()`, `ComputeCRC32()`, `EncodeDataBase64()` and `DecodeDataBase64()` were also checked against reference implementations over the same length range and are correct.

Changing the reserve to `sizeof(bitLen)` makes all 5 NIST vectors pass, all 4097 sweep lengths and all 3000 random messages match the reference, and makes the padded size equal the canonical size for all 10,000,001 lengths tested. A PR follows.

### Environment

- raylib 6.1-dev, master @ `2f768eb67052ca7aaa5849282342a1685bf35b20`
- Ubuntu 24.04.4 LTS, x86_64
- `PLATFORM_DESKTOP_GLFW`, `GRAPHICS_API_OPENGL_33`
- GCC 16.0.1

Platform independent: this is integer arithmetic in `rcore.c`, unrelated to GPU or windowing.

### Issue Screenshot

Not applicable, console output is included below.

### Code Example

```c
#include "raylib.h"
#include <stdio.h>
#include <string.h>

int main(void)
{
    // FIPS 180-4 two-block test vector, 56 bytes (56 % 64 == 56)
    const char *msg = "abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq";

    unsigned int *hash = ComputeSHA256((const unsigned char *)msg, (int)strlen(msg));

    printf("raylib   ");
    for (int i = 0; i < 8; i++) printf("%08x", hash[i]);
    printf("\n");
    printf("expected 248d6a61d20638b8e5c026930c3e6039a33ce45964ff2167f6ecedd419db06c1\n");

    return 0;
}
```

Output on master @ `2f768eb6`:

```
raylib   aafff98beb5f808f8be1cf96605775628e87f9acade749e7ffd2208d6d4ac859
expected 248d6a61d20638b8e5c026930c3e6039a33ce45964ff2167f6ecedd419db06c1
```

`sha256sum` agrees with the expected value:

```console
$ printf 'abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq' | sha256sum
248d6a61d20638b8e5c026930c3e6039a33ce45964ff2167f6ecedd419db06c1  -
```

killerdevildog:
Fixes #6143.

`ComputeSHA256()` reserves the padding tail with `sizeof(dataSize)`. `dataSize` is `int`, so that reserves 4 bytes, but SHA-256 needs 9: the `0x80` marker plus a 64-bit big-endian message length. The write a few lines below already uses `sizeof(bitLen)` (8 bytes), so the reserve and the write disagreed.

When only 5 to 8 bytes of room remained, the length field was written over the `0x80` marker and the tail of the message, and the padded size was one block short of the canonical size. Every write stayed inside the allocation, so the digest was silently wrong rather than crashing. This affected `dataSize % 64 ∈ {56, 57, 58, 59}`, which is 4/64 = 6.25% of input lengths, including the FIPS 180-4 two-block test vector.

Reserving `sizeof(bitLen)` fixes it. With an 8 byte reserve the existing round-up line produces exactly the canonical padded size for every length: writing `r = (dataSize + 8) mod 64`, when `r != 0` the result is the smallest multiple of 64 at or above `dataSize + 9`, and when `r == 0` (`dataSize ≡ 56 mod 64`) it adds the extra block the standard requires. Room after the message is `8 + (64 - r) >= 9` in all cases.

This PR changes only one line in `src/rcore.c`.

Validation: the 56 byte FIPS 180-4 vector fails before the change and passes afterward. All 5 NIST vectors pass (lengths 0, 3, 56, 112 and 1000000). Every length from 0 to 4096 matches a reference SHA-256, covering all 64 residue classes mod 64 sixty-four times each, as do 3000 pseudo-random messages at pseudo-random lengths up to 70000 bytes; before the change those same runs produced 256 and 210 wrong digests respectively. Comparing the padding arithmetic against the canonical padded size over lengths 0 to 10,000,000 gives 0 mismatches, and `paddedSize % 64 == 0` with room `>= dataSize + 9` holds throughout; the old expression gave 625,000 mismatches over that range. The full harness runs clean under AddressSanitizer and UndefinedBehaviorSanitizer. `ComputeSHA1()`, `ComputeMD5()`, `ComputeCRC32()`, `EncodeDataBase64()` and `DecodeDataBase64()` were re-checked against reference implementations and are unchanged and correct. `rcore.c` compiles with no warnings under `-Wall -Wextra`; the only warning in a full build is the pre-existing `'mi' may be used uninitialized` in bundled `external/m3d.h`, reached via `rmodels.c`.

The test harnesses are local and are not included in this PR. Built and tested on Linux with GCC (`PLATFORM_DESKTOP_GLFW`, `GRAPHICS_API_OPENGL_33`); not tested on Windows, macOS or Web, though the change is platform independent integer arithmetic. Note that digests returned for the affected lengths change with this fix, which is the intent: they now match the standard.

### DIFF
--- a/src/rcore.c
+++ b/src/rcore.c
@@ -3535,7 +3535,7 @@ unsigned int *ComputeSHA256(const unsigned char *data, int dataSize)
     hash[7] = 0x5be0cd19;
 
     const unsigned long long bitLen = 8ULL*dataSize;
-    unsigned long long paddedSize = dataSize + sizeof(dataSize);
+    unsigned long long paddedSize = dataSize + sizeof(bitLen); // Reserve room for the 64 bit message length appended at the end
     paddedSize += (64 - (paddedSize%64));
     unsigned char *buffer = (unsigned char *)RL_CALLOC(paddedSize, sizeof(unsigned char));
 


raysan5:
@killerdevildog thanks for the review

PR #6146, opened 2026-09-12
Fix make shell detection

### LINKED ISSUE #6135: [build][Make] `make clean` not working correctly for me when running from mingw64 shell
**WARNING: Please, read this note carefully before submitting a new issue:**

It is important to realise that **this is NOT A SUPPORT FORUM**, this is for reproducible BUGS with raylib ONLY.

There are lots of generous and helpful people ready to help you out on [raylib Discord forum](https://discord.gg/raylib) or [raylib reddit](https://www.reddit.com/r/raylib/).

Remember that asking for support questions here actively takes developer time away from improving raylib.

---

 - [ x ] I tested it on latest raylib version from master branch
 - [ x ] I checked there is no similar issue already reported
 - [ x ] I checked the documentation on the [wiki](https://github.com/raysan5/raylib/wiki)
 - [ x ] My code has no errors or misuse of raylib

### Issue description

*Briefly describe the issue you are experiencing (or the feature you want to see added to raylib). Tell us what you were trying to do and what happened instead. Remember, this is not the best place to ask questions. For questions, go to [raylib Discord server](https://discord.gg/raylib).*

Msys2 comes with a raylib package but I wanted to build it myself and use that instead, so i cloned the repo and built it. Then I noticed that release mode is only using -O1 by default so I changed it to -O2 and added a couple extra settings and tried to run `make clean && make` so i could rebuild only to be dumped into a cmd session from the mingw64 terminal with no actual cleaning being done. Typing exit a few times got me back to mingw64 but the old .o files are still there and raylib doesn't rebuild. I assume this is because `$(PLATFORM_SHELL)` might always return cmd on windows even if you're using an msys2 shell, though it might also be something wrong with my mingw64 environment though i don't know what that could be.

### Environment

Windows 11 25H2, OpenGL 4.6, NVidia RTX4050 Laptop GPU

### Issue Screenshot

*If possible, provide a screenshot that illustrates the issue. Usually an image is better than a thousand words.*

Image

### Code Example

*Provide minimal reproduction code to test the issue. Please, format the code properly and try to keep it as simple as possible, just focusing on the experienced issue.*

`make clean && make`

Peter0x44:
Fixes #6135.

  The Makefiles assume Windows means cmd, but GNU Make can select sh
  instead. This causes `make clean` to use the wrong shell commands.

  Detect the recipe shell using `echo "test"`: cmd preserves the quotes,
  while sh removes them. Apply this to the library, examples and tools,
  and remove the forced cmd shell from library cleanup. Explicit
  PLATFORM_SHELL overrides remain supported.

  Also fix Win32 clean recipe indentation and,
 pass rlparser's recursive make variables as command-line arguments
  so those calls work under both cmd and sh.

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -122,6 +122,14 @@ BUILD_WEB_RESOURCES_PATH  ?= $(dir $<)resources@resources
 # WARNING: Requires raylib compiled with GRAPHICS_API_OPENGL_ES3
 BUILD_WEB_WEBGL2      ?= FALSE
 
+# Detect the shell used by make: cmd preserves quotes, while sh removes them.
+ifndef PLATFORM_SHELL
+    PLATFORM_SHELL = sh
+    ifeq ($(shell echo "test"),"test")
+        PLATFORM_SHELL = cmd
+    endif
+endif
+
 # Determine PLATFORM_OS when required
 ifeq ($(TARGET_PLATFORM),$(filter $(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW PLATFORM_DESKTOP_SDL PLATFORM_DESKTOP_RGFW PLATFORM_WEB PLATFORM_WEB_RGFW))
     # No uname.exe on MinGW!, but OS=Windows_NT on Windows!
@@ -784,7 +792,11 @@ endif
 clean:
 ifeq ($(TARGET_PLATFORM),PLATFORM_DESKTOP_GLFW)
     ifeq ($(PLATFORM_OS),WINDOWS)
+        ifeq ($(PLATFORM_SHELL),cmd)
 		del *.o *.exe /s
+        else
+		find . -type f '(' -name '*.o' -o -name '*.exe' ')' -exec rm -f {} +
+        endif
     endif
     ifeq ($(PLATFORM_OS),BSD)
 		find . -type f -perm -ugo+x -delete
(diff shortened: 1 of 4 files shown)


emecdelam:
For curiosity, are there any reasons to use echo "test" instead of just finding the executable? something like

```
ifeq ($(OS),Windows_NT)
    ifeq ($(findstring sh.exe,$(SHELL)),sh.exe)
        DETECTED_SHELL := sh
    else
        DETECTED_SHELL := cmd
    endif
else
    DETECTED_SHELL := sh
endif
```

Peter0x44:
Actually, my blogpost is a little wrong on this point. Make *DOES* report sh.exe in SHELL (it's not unset), but it's not consistent.

This is maybe a GNU make bug, but if you run `make` without an sh.exe on your PATH it will still report that it using sh (despite that it is NOT). Probably could be patched, but these buggy makes are of course going to be very widespread.

I still think it's the correct approach for this PR.

Peter0x44:
```
   Configuration         $(SHELL)       $(shell echo "test")
  ━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━
   sh on PATH            C:/…/sh.exe    test
  ────────────────────  ─────────────  ──────────────────────
   No sh on PATH         sh.exe         "test"
  ────────────────────  ─────────────  ──────────────────────
   Explicit SHELL=sh     sh             test
  ────────────────────  ─────────────  ──────────────────────
   Explicit SHELL=cmd    cmd            "test"
```

raysan5:
@Peter0x44 thanks for the review and the explanation!

PR #6147, opened 2026-09-13
[examples] Fix texture slot state leak in shadowmap rendering example

Ne0nWinds:
This PR fixes the shadowmap rendering example on web. The shadows were not rendering due to an opengl state leak bug, which primarily affected WebGL.

The issue was reported here: #5664

## Problem Details
On the first iteration of the main game loop, the example code binds the shadowmap texture to texture slot 10 (textureActiveSlot variable). It does not get unbound from the texture slot at any point, so it stays active at texture slot 10 forever after this point.

This becomes problematic on the second iteration of the main game loop. On line 155, `BeginTextureMode(shadowMap)` is called, which binds the shadowmap as a drawable framebuffer. However, it's still bound as a readable texture on slot 10.

This is an invalid state in WebGL. The same underlying texture is bound as both a readable texture and as a writable framebuffer for the same draw call. Desktop drivers (which are more permissive than WebGL) will allow this because the shadowmap texture is not actually read from/written to in the same shader, but nonetheless, it's technically an invalid state on desktop too.

This can be clearly seen by running the web example and taking a capture with Spector.js (browser extension for diagnosing WebGL issues). The first frame renders the shadows correctly, but every subsequent frame does not.

## Solution
The fix is to unbind the shadowmap texture from slot 10 after it's used to draw the main scene. This prevents the texture slot state leak described above.

## Platforms Tested
Firefox 155, Chrome 153 - Shadows now render properly; confirmed fixed
Windows 10 - Shadows continue to work; no regression


raysan5:
@Ne0nWinds thanks for the review

PR #6148, opened 2026-09-14
[docs] Replace Youtube Subscriber Badge with Static Badge

openblocki:
Shields.io sunsetted their YouTube badges about a month ago -> https://github.com/badges/shields/issues/12102
Having the static badge looks better than the "broken" badge imo.


raysan5:
@openblocki Oh! I didn't  know! Thanks for the update!

PR #6149, opened 2026-09-15
[rtext] glyph lookup update

emecdelam:
In the wishlist for the 7.0 version, I see that there is a potential improvement for glyph search https://github.com/raysan5/raylib/discussions/4513 , I also saw an attempt with a hashmap https://github.com/raysan5/raylib/pull/5249 that changed some structs from the API.

In this PR, the search attempts direct lookup for continuous fonts (as could be used for ASCII) otherwise, it attempts a binary search, the glyphs are sorted when loading the font and there is no change to existing structs definition.

I also read somewhere that you didn't like to add glibc calls so I mark this as a draft in case you prefer to remove the binary search part or `qsort` calls.

Also I noticed while reviewing that the code before used a macro definition just before a #if , I'm not sure as to why it's been done this way. Anyway I'm always open to changing the code if you have any suggestion or improvements :)


raysan5:
@emecdelam thanks but no plans to work on that for now.

emecdelam:
@raysan5 no problem, I think it would be nice to remove this item from the wishlist then 
image https://github.com/raysan5/raylib/discussions/5710

PR #6150, opened 2026-09-15
[rtextures] added support for webp format

CrackedPixel:
using the [simplewebp](https://github.com/MikuAuahDark/simplewebp) single-header library, it's very simple:

BSD-3 license: https://opensource.org/license/BSD-3-clause

then it's invisible to the user, just like any other format:
```c
Texture texBunny = LoadTexture("resources/raybunny.webp");
```
image


src/external/simplewebp.h
```diff
@@ -0,0 +1,5651 @@
...
+ * WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+ * USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ * 
+ * Additional IP Rights Grant (Patents)
+ * ------------------------------------
+ * 
+ * "These implementations" means the copyrightable works that implement the WebM codecs distributed by Google as part
+ * of the WebM Project.
+ * 
+ * Google hereby grants to you a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable (except as
```
raysan5, line 5640: I'm afraid there could bee concerns with those Patents Rights if included on raylib...

raysan5:
@CrackedPixel I reallly like that WebP singlee-file header-only loader! Great discovery! But I'm afraid there could be some concerns with the IP Pattents listed, so I prefer to avoid merging this PR. 

Users with WebP support required could use the header directly on their projects.

src/external/simplewebp.h
CrackedPixel: > I'm afraid there could bee concerns with those Patents Rights if included on raylib...

"perpetual, non-exclusive, royalty free, irrevocable" is pretty good :sweat_smile: but i did see the other message about it being a large library for just 1 format, thats fine. at least if someone needs a reference we have one :sunglasses:

PR #6152, opened 2026-09-16
Add gcl native raylib module entry to BINDINGS.md

gnuchanos:
What I did before was extremely amateurish, and I deleted the repo out of frustration. This time, I’m trying to be more professional and built a system that works with more native built-in modules. The goal is entirely education-focused. Since the system includes both Lua and Python as well, the main focus will of course be on my own C-like language.


raysan5:
@gnuchanos merging but not sure if it will be kept... bindings usually point to a binding, not a language repo...

gnuchanos:
it's not simple binding raylib is build in module for my language and extra lua and python wrapper
i write excidently 6.0 but it's 6.1-dev

image
image
image

image
image @raysan5

PR #6154, opened 2026-09-16
[examples] update example to current version of raylib

macsmol:
Updated textures_mouse_painting example to use the new gesture function.


raysan5:
@macsmol thanks for the review!

PR #6156, opened 2026-09-16
[rexm] Make search boundaries explicit.

MatthewRoush:
Avoids accidentally modifying examples (i.e. modifying `shapes_logo_raylib_anim` when you're trying to work with `shapes_logo_raylib`.)


raysan5:
@MatthewRoush thanks for the review

PR #6158, opened 2026-09-17
[rcore] Split WaitTime OS-sleep into SleepTime()

Skellybob56:
I extracted the system halting from `WaitTime(double seconds)` into a new public `SleepTime(double seconds)`.

`WaitTime` is a partial busy wait to ensure accuracy and low CPU usage. Some use cases are only interested in low CPU usage and not accuracy. It can be difficult to get sleep directly from the OS in some languages and on some platforms. raylib already contains a platform-agnostic system halt so there is no reason to force the user to make their own.

This should have no effect on any pre-existing functionality, only add one new `SleepTime` function.

I have decided against adding a symmetrical `BusyWaitTime(double seconds)` function that always busy waits because it is trivial for the user to implement in a platform-agnostic manner.

I have tested `WaitTime(0.5)` and `SleepTime(0.5)`: `WaitTime` measured 0.500000s while `SleepTime` measured 0.500588s.
I have also tested a while loop with `WaitTime(0.001)` and a while loop with `SleepTime(0.001)`. Task Manager reports 12.5% CPU usage for the `WaitTime(0.001)` loop (with a measured frequency of 999.97Hz) and 0.0% CPU usage for the `SleepTime(0.001)` loop (with a measured frequency of 502.43Hz).
While `SleepTime` is almost twice as slow as a true 1ms wait, it is massively more performant which can be the more important factor.


raysan5:
@Skellybob56 Thanks but I prefer to avoid it.

Skellybob56:
May I ask why?
This function would be very useful to me and others with the same requirements.

raysan5:
> May I ask why?

Nobody requested it in 13 years, so it's not that needed

PR #6159, opened 2026-09-17
[rshapes] Change `DrawRectangleRounded*()` segment auto-calculation threshold.

MatthewRoush:
* Change `DrawRectangleRounded*()` segment auto-calculation threshold.

Auto-calculating rounded rectangle segments when there are fewer than 4 segments seems overly aggressive (a user may actually want 1, 2, or 3 segments.) If a user wants the segments to be auto-calculated, they can still pass in 0 (or any value smaller than 1) for the segments.

Using 4 as the minimum segments may have been copied from circle drawing logic (where fewer than 4 segments may be undesirable.) For rounded rectangles though, the segments are drawn per corner, so the total drawn segments are `segments*4`.

* Refactor `DrawRectangleRounded*()` functions.

Move some logic around and make formatting more consistent.


raysan5:
@MatthewRoush thanks for the update, makes sense. Just note that in raylib I try to avoid defining variables as `const`, I prefer just to define regular variables, it's more readable and consistent along the library; the benefit of `const` is minimal.

MatthewRoush:
I'll make sure to keep that in mind! It may be worth adding that to `CONVENTIONS.md`.

PR #6160, opened 2026-09-18
[rshapes] fix out of date docs. QUADS are now the default

jestarray:


raysan5:
@jestarray Good catch! Thanks for the update!

jestarray:
I think it's been the default for quite a while, and I was wondering when would someone want to turn this off. If no one is using it, in 7.0+ I suggest deleting support for triangles and remove this flag

PR #6162, opened 2026-09-18
external/qoa: update vendored qoa.h with upstream out-of-bounds fixes (fixes #6161)

### LINKED ISSUE #6161: Update vendored src/external/qoa.h — it predates two upstream OOB fixes (heap + stack overflow via LoadWaveFromMemory)
raylib bundles a copy of qoa.h from before phoboslab/qoa added two bounds fixes, so a small crafted .qoa file passed to the public LoadWaveFromMemory() writes out of bounds. Both are already fixed upstream, so this is a re-vendor request rather than a raylib-code patch.

Heap write: qoa_decode() sizes the sample buffer from the file header (total_samples = qoa->samples * qoa->channels, qoa.h:659-660), but qoa_decode_frame() writes decoded samples using the frame header's own samples field (:640) without re-checking it fits the allocation. The only bound present (max_total_samples) is derived from the attacker's own frame_size, not from the allocation. A 40-byte file whose header says one sample and whose frame declares twenty gives a 2-byte allocation and a write past it at qoa.h:640.

Stack write: qoa_decode_frame() indexes qoa->lms[], sized QOA_MAX_CHANNELS, using the frame-header channel count with no upper bound, so a 232-byte file declaring nine channels writes past that array at qoa.h:615.

SUPPORT_FILEFORMAT_QOA is default-enabled, and LoadWave/LoadWaveFromMemory/LoadSound are public RLAPI functions, so this is reachable from an untrusted .qoa file through the ordinary API.

Both defects are fixed upstream in phoboslab/qoa commit af36bdf (2026-06-07). qoa ships no releases, so the fix lives on its default branch, and raylib has not re-vendored it.

Reproduce: PoC bundle at https://drive.google.com/file/d/15N7onM7uyOhKNyseLxrwIfT_K3e1fZzO/view . run_poc.sh drives both defects through the public LoadWaveFromMemory(): ASan reports a heap-buffer-overflow WRITE at qoa.h:640 for the 40-byte file, and a stack-buffer-overflow at qoa.h:615 for the 232-byte nine-channel file. A version-flip control compiles the same harness against phoboslab/qoa HEAD, feeds the same files, and both are refused (qoa_decode returns NULL for the nine-channel file) while that copy still decodes valid audio, which shows it is the stale vendored copy. A one-byte control and a raylib-encoded positive control both decode cleanly.

Fix: update src/external/qoa.h to phoboslab/qoa HEAD, which contains af36bdf. It is worth checking whether other vendored headers in src/external/ are similarly behind their upstreams.

Affected: raylib 6.0 (newest release) and master (vendored qoa.h byte-identical); 5.5 and 5.0 also.

Found by Claude and verified by Calif.io's human engineers. Our reference is ANT-2026-YJZR448E.

thaidn:
raylib's bundled `src/external/qoa.h` predated two upstream bounds fixes in phoboslab/qoa, so a crafted `.qoa` file passed to the public `LoadWaveFromMemory()` could write out of bounds (reported in #6161):

- **Heap:** `qoa_decode()` sized the sample buffer from the file header, but `qoa_decode_frame()` wrote decoded samples using the frame header's own `samples`/`channels` fields with no cross-check against the allocation (`qoa.h:640`). A header declaring 1 sample with a frame declaring 20 produced a 2-byte allocation and a 40-byte write.
- **Stack:** `qoa_decode_frame()` indexed `qoa->lms[]` (sized `QOA_MAX_CHANNELS`) using the frame-header channel count with no upper bound (`qoa.h:615`), so a file declaring `channels=9` wrote past the array.

This re-vendors `qoa.h` from current `phoboslab/qoa` (commit `f73b4a36`), which adds the missing checks:

- `qoa_decode_header()` now rejects `channels > QOA_MAX_CHANNELS` (fixes the stack write).
- `qoa_decode_frame()` now derives `num_slices` from `samples`, bounds it against `QOA_SLICES_PER_FRAME` and the frame's data size, and guards `frame_size < header_size` (fixes the heap write).
- `qoa_decode()` rounds the allocation up to full frames with an overflow guard.

raylib's copy was otherwise an unmodified, older upstream copy, so the diff is exactly these fixes. Verified against the reproduction in #6161: both crafted inputs are now refused (`qoa_decode` returns `NULL` for the nine-channel file) while valid QOA audio still decodes correctly.

Fixes #6161.

🤖 Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01JfEozSeCHNgikWTQMDta5U


raysan5:
@thaidn thanks for the update

PR #6163, opened 2026-09-18
[gh actions] clang tidy linting: needless casts and assignments in if

jestarray:
This PR adds a gh actions file that runs some clang lints to keep the code tidy:
```
/home/runner/work/raylib/raylib/src/rtext.c:1245:67: warning: redundant explicit casting to the same type 'float' as the sub-expression, remove this casting [readability-redundant-casting]
/home/runner/work/raylib/raylib/src/rtext.c:1316:67: warning: redundant explicit casting to the same type 'float' as the sub-expression, remove this casting [readability-redundant-casting]
/home/runner/work/raylib/raylib/src/rtext.c:1810:31: warning: redundant explicit casting to the same type 'char *' as the sub-expression, remove this casting [readability-redundant-casting]
```
This will make it so future PRs will fail if they don't fix needless casting:
```c
// rectangle already stores position in floats so its needless
(float)rect.x;
```
and assignments in if statements:
```c
// NOTE: Used by DrawLineBezier() only
static float EaseCubicInOut(float t, float b, float c, float d)
{
    float result = 0.0f;

    if ((t /= 0.5f*d) < 1) result = 0.5f*c*t*t*t + b; // t/= 0.5 really should be in a new line
    else
    {
        t -= 2;
        result = 0.5f*c*(t*t*t + 2.0f) + b;
    }

    return result;
}
```


raysan5:
@jestarray thanks for the improving, I'm meerging but not forcing it execution, unfortunately I'm in runners quota limits and I can't affort more minutes

PR #6164, opened 2026-09-18
changed MAX_FILEPATH_LENGTH to be the same as MAX_PATH

randomnumbersandletters:
just to make things consistent


raysan5:
@randomnumbersandletters thanks for the review

PR #6165, opened 2026-09-18
[examples] Added `shapes_drag_puzzle`

gabriel-piangers:
Added a drag and drop puzzle example implementing 3 different shapes, Rectangle, Circle and Triangle.

Added for the discussion [#5209](https://github.com/raysan5/raylib/discussions/5209).

I have tested it on Windows and seems to work fine. Only thing that bothered me a little is that the drawing order for the shapes is fixed instead of being based on the order of latest pick. Not really sure this could be done easily since all objects are from different types. But I still made it so the dragged object always draws last.

I have used the `rexm` tool to add the example as suggested.

core_shapes_drag_puzzle


raysan5:
@gabriel-piangers please, review code formating to follow raylib conventions, braces should be aligned

EDIT: Please, rename to just `shapes_drag_puzzle` and move to `shapes` category, it does not belong to `core`.

gabriel-piangers:
> @gabriel-piangers please, review code formating to follow raylib conventions, braces should be aligned
> 
> EDIT: Please, rename to just `shapes_drag_puzzle` and move to `shapes` category, it does not belong to `core`.

I changed the code formatting accordingly to the raylib conventions and switched it to shapes category.

raysan5:
@gabriel-piangers thanks! great addition! 👍😄

PR #6166, opened 2026-09-19
[rtext] Add `TextCompare()` to compare text strings

davidbuzatto:
This PR adds `TextCompare()`, a three-way text comparison function that returns `<0`, `0` or `>0` like `strcmp()`.

### Motivation

`TextIsEqual()` only reports equality, so there is no way to order text strings (e.g. sorting a list of names, or using it as a `qsort()` comparator) with the raylib text API alone, users have to fall back to `strcmp()` directly.

I ran into this while prototyping a game with a simple scripting language used to move a robot: while writing the lexer for it I needed to compare strings lexicographically, and raylib's text functions could only tell me whether two strings were equal.

### Scope

I know raylib aims to keep a compact API, so I'm proposing this only for consistency with the existing basic text functions: raylib already provides counterparts for the common string operations (`TextCopy()`, `TextLength()`, `TextIsEqual()`, `TextFindIndex()`...) and this is the one that was missing to complete that set. It is a single small function, and I deliberately did not add variants (case-insensitive, length-limited...).

### Behaviour

- Non-NULL strings: same result as `strcmp()`. Comparison is byte-wise, case-sensitive and not locale-aware.
- NULL is considered smaller than any string and `NULL == NULL`, so the result is always a valid ordering (safe to use as a `qsort()` comparator) and no special error value is needed. It follows the NULL-safety added to `TextIsEqual()` in #2121, but note that `TextIsEqual(NULL, NULL)` returns `false` while `TextCompare(NULL, NULL)` returns `0`.
- Only the sign of the result is meaningful.

The only libc dependency is `strcmp()`, already required by `TextIsEqual()`.

### Testing

- Built the static library on Windows 10 (`PLATFORM_DESKTOP`, MinGW gcc 15.2, `-std=c99 -Wall`) with no new warnings.
- Linked a small test program against it checking NULL/NULL, NULL/string, equal, less, greater and prefix cases, plus a `qsort()` of an array containing NULLs.
- Not tested on other platforms.


raysan5:
@davidbuzatto thanks but there is no need for this function at the moment, as library grows I have to be more strict with additions

PR #6170, opened 2026-09-20
[examples] Fix the documented Zig build command

Hugohong258:
The examples README says `zig build` compiles all examples, but the default step only installs the raylib library. Document `zig build examples` instead and clarify that the Zig commands run from the repository root.

Validated on Windows with Zig 0.16.0:
- `zig build --help` lists the separate `examples` step.
- `zig build -j4 --prefix <default-output> --summary all` succeeds and installs the library and headers, without example executables.
- `zig build examples -j4 --prefix <examples-output> --summary all` succeeds (461/461 steps) and installs 225 example executables.
- `git diff --check` passes.

The examples were compiled, not launched. No build scripts or example sources are changed.


raysan5:
@Hugohong258 thanks for the review!

PR #6172, opened 2026-09-20
[rshapes] Don't draw a circle when `radius <= 0`.

MatthewRoush:
A circle shouldn't be drawn if the radius is zero (and a negative radius probably doesn't make sense...)


raysan5:
@MatthewRoush thanks for the review

PR #6173, opened 2026-09-20
[cmake] Fix install export with vendored SDL

### LINKED ISSUE #6157: [build] SDL backend: using vendored SDL causes error in raylib's InstallConfigurations.cmake logic
Whenever I try to use the raylib SDL backend with a vendored SDL subproject unpatched, I get an error like this:

```
CMake Error in thirdparty/raylib/src/CMakeLists.txt:
  install(EXPORT "raylib-targets" ...) includes target "raylib" which
  requires target "SDL3-shared" that is not in any export set.
```

This is the CMakeLists.txt that I use:

```cmake
project(raysdl)
cmake_minimum_required(VERSION 3.16)

if (APPLE)
    enable_language(OBJC)
endif()

add_executable(${PROJECT_NAME})

add_subdirectory(src)
add_subdirectory(thirdparty) # adds SDL and raylib subdirectories

target_link_libraries(${PROJECT_NAME} SDL3::SDL3)
target_link_libraries(${PROJECT_NAME} raylib)

if (APPLE)
    target_link_libraries(${PROJECT_NAME} "-framework IOKit")
    target_link_libraries(${PROJECT_NAME} "-framework Cocoa")
    target_link_libraries(${PROJECT_NAME} "-framework QuartzCore")
    target_link_libraries(${PROJECT_NAME} "-framework OpenGL")
endif()
```

The quick and dirty solution was patching InstallConfigurations.cmake by wrapping it with an if(PROJECT_IS_TOP_LEVEL) condition. After that, it works fine. Perhaps that fix could be integrated?

RobLoach:
Excludes vendored (add_subdirectory) SDL targets from the static library install interface so `install(EXPORT raylib-targets)` no longer fails. Fixes #6157

Similar thing with glfw.


src/CMakeLists.txt
```diff
@@ -112,6 +112,11 @@ target_link_libraries(raylib PUBLIC ${LIBS_PUBLIC})
 if (NOT BUILD_SHARED_LIBS)
     set(raylib_install_private_libs ${LIBS_PRIVATE})
 
+    # SDL from add_subdirectory() is not in raylib's export set, so it cannot
+    # be referenced from the install interface. SDL found via find_package()
+    # is in LIBS_PUBLIC and handled through find_dependency() instead.
```
RobLoach:
``` suggestion
    # SDL is handled through LIBS_PUBLIC and find_dependencies(), so
    # we avoid having them privately.
```

raysan5:
@RobLoach thanks for the review!

Issue #6174, opened 2026-09-20
vox_loader: out-of-bounds write in Vox_SetVoxel, bounds checks are commented out


AvAl4nch:
`Vox_SetVoxel` has both bounds checks commented out, and voxel coords from the `XYZI` chunk are never checked against the `SIZE` chunk:

```c
int offset = (chX * pvoxarray->ChunkFlattenOffset) + (chZ * pvoxarray->chunksSizeY) + chY;

//if (offset > voxarray->arrayChunksSize)
//{
//	TraceLog(LOG_ERROR, "Out of array");
//}

CubeChunk3D* chunk = &pvoxarray->m_arrayChunks[offset];   // offset comes from the file
...
chunk->m_array = (unsigned char *)VOX_MALLOC(size);       // OOB write
chunk->arraySize = size;                                  // OOB write
```

A 64-byte .vox with `SIZE=(1,1,1)` and one voxel at `(255,255,255)` gives a heap OOB write through `LoadModel()`. ASAN reports `WRITE of size 8` at vox_loader.h:369 and `WRITE of size 4` at :370, with the 16-byte `m_arrayChunks` allocation indexed 224 bytes out.

`Vox_GetVoxel` directly below already has the checks the setter is missing:

```c
if (x < 0 || y < 0 || z < 0) return 0;
if (x >= pvoxarray->sizeX || y >= pvoxarray->sizeY || z >= pvoxarray->sizeZ) return 0;
```

Adding those two lines to `Vox_SetVoxel` fixes it.

Two smaller things in the same function: the chunk-walk loop checks `fileDataPtr < endfileDataPtr` once per iteration and then reads 12+ bytes unchecked, and the `XYZI` loop trusts `numVoxels` from the file without comparing against the buffer end.

Tested at 7bb0345.

raysan5:
too much AI text, no time to read

AvAl4nch:
Sorry about the length. Short version:

`Vox_SetVoxel` has both bounds checks commented out, and voxel coords from `XYZI` are never checked against the `SIZE` chunk:

```c
int offset = (chX * pvoxarray->ChunkFlattenOffset) + (chZ * pvoxarray->chunksSizeY) + chY;
//if (offset > voxarray->arrayChunksSize)
CubeChunk3D* chunk = &pvoxarray->m_arrayChunks[offset];   // offset comes from the file
```

[...]

PR #6175, opened 2026-09-20
[vox_loader] Add bounds checks to Vox_SetVoxel and chunk parsing

AvAl4nch:
Fixes the memory safety issue from #6174.

`Vox_SetVoxel` had both bounds checks commented out, and the coords it gets come straight from the `XYZI` chunk with nothing checking them against the `SIZE` chunk:

```c
int offset = (chX * pvoxarray->ChunkFlattenOffset) + (chZ * pvoxarray->chunksSizeY) + chY;
//if (offset > voxarray->arrayChunksSize)
CubeChunk3D* chunk = &pvoxarray->m_arrayChunks[offset];
```

This restores those two checks and adds the same range test `Vox_GetVoxel` already does.

The chunk walk had the same problem: it checked `fileDataPtr < endfileDataPtr` once, then read 12 bytes of header plus up to 12 more of content. That also trips on valid files, because the RGBA branch reads 255 entries while the chunk holds 256, leaving 4 bytes that get parsed as another header. Each branch now checks it has the bytes first.

Tested with ASAN:

| Input | before | after |
|---|---|---|
| 64-byte .vox, voxel outside SIZE | heap OOB write, vox_loader.h:369 | loads, voxel skipped |
| truncated .vox | heap OOB read, vox_loader.h:602 | loads |
| .vox with RGBA palette | heap OOB read | loads |
| normal .vox | ok | identical verts/idx/colors |

Also fuzzed the patched loader for a few minutes, no crashes.

One thing I did not touch: `Vox_AllocArray` computes `int size = sizeof(CubeChunk3D) * chx * chy * chz` in `int`, which can overflow on a large SIZE chunk. Left it alone to keep this focused, happy to add it if you want.


raysan5:
@AvAl4nch thanks for the improvement

AvAl4nch:
Thanks for merging.

Since this one was a heap out-of-bounds write reachable from a `.vox` file, downstream users would benefit from a CVE to pin against. It affects 4.0.0 through 6.0, and the fix isn't in a tagged release yet.

GitHub can assign one directly: Security > Advisories > New draft advisory > Request CVE ID. Takes a couple of minutes and you keep control of the text. Happy to supply the affected range, CWE and CVSS if useful.

If you'd rather not, no problem, I'll request it from MITRE instead.

raysan5:
@AvAl4nch Thanks, no plans to do that but please, if the intention of this AI-generated PR was only try to farm this kind of "medals", I'd would appreciate you don't send any other PR to my project. Thanks for your understanding.

Considering reverting this merge...

AvAl4nch:
hello brother its now the real me talking, to be clear about my intentions im doing a research on how getting a cve in this time is not an indication of skill, so ive build a dedicated setup to find security issues and fix them fully automatically while also requesting a cve, im sorry for bothering you and i really appreciate your time and help. you wont see any other PRs from my side
thanks for understanding and have a good rest of your day.

AvAl4nch:
note: if you really want me to revert this issue i would do that, but my AI only submits an issue if it test it locally and builds an exploit for it so i would recommend you keep it if you can.

PR #6177, opened 2026-09-21
[raylib.h] ConfigFlags:  reordered numerically by the hex value

jestarray:
Just a simple re-order so in case we add future flags, we can easily see which flags are already taken


raysan5:
@jestarray I prefer to keep current order

PR #6180, opened 2026-09-22
[rcamera] fix Camera_Free gamepad drift

Andre-NPC01:
The PR fixes gamepad drift by rescaling the axis values so they smoothly start from 0 after the deadzone.


raysan5:
@Andre-NPC01 I prefer to keep current code, it's simpler to read.

Andre-NPC01:
@raysan5  Thanks for the review.  Ya it's was a lot code compared original 6 line of code.   Could the value that the gamepad movements are checked against be exposed like the camera draw distance was.?

PR #6181, opened 2026-09-22
[rcore][win32] Copy Mouse.currentButtonState to Mouse.previousButtonState in PollInputEvents

Ne0nWinds:
`IsMouseButtonPressed()` should only return `true` once per mouse click. However, when compiling with the new win32 backend, it would return true every single frame that the mouse button was held down, effectively functioning like `IsMouseButtonDown()`. 

The solution is to copy all values of `CORE.Input.Mouse.currentButtonState` to `CORE.Input.Mouse.previousButtonState` in `PollInputEvents`. This is exactly what the glfw backend does.

I tested by compiling the `core_input_mouse` example with the win32 backend. `IsMouseButtonPressed()` now only fires once every mouse click.


raysan5:
@Ne0nWinds Good catch! Thanks for the improvement! Win32 backend is still under development and it could have some bugs and things to review or implement

